Issue #2, opened 2017-03-01
Support for CMA meta-data


arporter:
CMA operators require an extension to PSyclone to generate calls in the PSy layer into the operator infrastructure. New metadata to describe this has been added. The associated LFRic ticket is 536.

This ticket will add support for parsing the new metadata.

arporter:
Chris thinks that it is plausible that we can drop the proposed meta-funcs and identify the purpose of a kernel using its arguments alone. I'll continue to explore this option.

arporter:
Pull request (#5) merged to master.
56c9734 - updated changelog and top-level psyclone.pdf on master branch.
Closing issue.

PR #5, opened 2017-03-03
Add support for CMA metadata

arporter:
This pull request is to bring in work done in #2 to add meta-data support for Column Matrix Assembly operators.


doc/dynamo0p3.rst
```diff
@@ -182,8 +190,11 @@ in the :ref:`kernel-layer` section. This section explains the
 dynamo0.3-specific rules for kernels and then goes on to describe
 their metadata and subroutine arguments.
 
-Rules
-+++++
+General Rules
++++++++++++++
+
+In the following, 'operator' refers to both LMA and CMA operator
```
rupertford: Is LMA the same as a traditional operator?  The above line makes it sound like there are only LMA and CMA operators. If so, we need to use the names consistently in the documentation i.e. it should be referred to either as an operator or as LMA. If not, then operator refer to a "standard" operator (should this have a TLA too?) as well as LMA and CMA.

doc/dynamo0p3.rst
```diff
@@ -230,6 +241,44 @@ Rules
     does not require values beyond the level-1 halo. If it does then
     PSyclone will abort.
 
+Rules for Kernels that work with CMA Operators
```
rupertford: If I understand correctly, we no longer have a single set of rules for kernels, we now have 2 main types of kernels
1) traditional kernels
2) CMA kernels
where CMA kernels can be split into 3 types.
The rules appear to be different for each case.
Therefore we should have a general rule section (which we do but I don't think is general as it includes rules for traditional kernels that do not apply to certain CMA kernels.
We should also have a standard kernel section with rules specific to these types of kernels.

doc/dynamo0p3.rst
```diff
@@ -230,6 +241,44 @@ Rules
...
+
+Matrix-Matrix
+^^^^^^^^^^^^^
+
+A kernel that has only column-wise operators as arguments is identified
+as performing a matrix-matrix operation. In this case, exactly one of the
+arguments must be written to while the others must be read-only.
+
 Metadata
 ++++++++
```
rupertford: The metadata section currently has no examples of the use of metadata to describe CMA (and LMA if this is a new concept) for the different uses of CMA in kernels. There should be at least an example of the metadata required in the 3 scenarios.

I would also be useful to clarify what these iterate over (cells I presume).

doc/dynamo0p3.rst
```diff
@@ -61,6 +61,14 @@ Operator
```
rupertford: At the start of the Kernel section I think it should be explained that there are different types of Kernel with their own rules and that these will be explained in the following sections.

doc/dynamo0p3.rst
```diff
@@ -408,22 +458,24 @@ need not be on the same space.
...
+GH_INTEGER        	n/a                             GH_SUM (Built-ins only)
+GH_REAL           	n/a                             GH_SUM (Built-ins only)
+GH_FIELD                Discontinuous (w3)              GH_WRITE
+GH_FIELD                Continuous (not w3)             GH_INC
+GH_OPERATOR             Any for both 'to' and 'from'    GH_WRITE
+GH_COLUMNWISE_OPERATOR  Any for both 'to' and 'from'    GH_WRITE
+======================  ============================    =======================
 
 Note that only Built-ins may modify scalar arguments. There is no
 restriction on the number and function-spaces of other quantities that
```
rupertford: The Subroutine section also needs looking at. Presumably we now have different rules for the different types of Kernels and we have new datatypes that we need to give examples of, in terms of their dimensions. If this information is not yet known then there should at least be placeholders.

Also, do the new kernel types have any different behavior when it comes to parallelization, loop fusion etc. If so, this information needs to be captured.

arporter:
I've now made quite a few changes. I've attempted to make it clear that what we previously just called an 'operator' is in fact a LMA operator. I've changed one of the meta-data examples to use a CMA operator but I don't think adding more examples would help very much (they'd look very similar to the existing ones.) I've fully populated the detailed rules on the expected kernel interfaces. These differ slightly from those in the tests that Chris and Eike have constructed because I've attempted to be consistent with the way we order arguments (e.g. dofmaps) in other kernels.

There are still some open questions:
  * Can a CMA kernel use field vectors?
  * Presumably stencil operations are forbidden for CMA kernels?
  * Are there any issues around parallelising CMA kernels?

src/dynamo0p3.py
```diff
@@ -674,6 +712,124 @@ def _validate(self, need_evaluator):
...
+                            mutable_cma_op.function_space_to or
+                        lop.function_space_from !=
+                            mutable_cma_op.function_space_from):
+                        raise ParseError(
+                            "When assembling a column-wise operator from "
+                            "LMA operators the to and from function "
+                            "spaces must match but this is not the case "
+                            "for kernel {0}".format(self.name))
+                # The kernel must not write to any CMA operators
+                lma_ops = psyGen.args_filter(self._arg_descriptors,
```
arporter: Done.

src/dynamo0p3.py
```diff
@@ -565,6 +599,10 @@ class DynKernMetadata(KernelType):
     def __init__(self, ast, name=None):
         KernelType.__init__(self, ast, name=name)
 
+        # The type of CMA operation this kernel performs (empty string
+        # signifies that no CMA operators are involved)
+        self._cma_operation = ""
```
arporter: Yes; done.

src/dynamo0p3.py
```diff
@@ -3436,8 +3592,9 @@ def iteration_space_arg(self, mapping=None):
 
         # Since we always compute operators out to the L1 halo we first
         # check whether this kernel writes to an operator
-        op_args = self.args_filter(arg_types=["gh_operator"],
-                                   arg_accesses=["gh_write", "gh_inc"])
+        op_args = psyGen.args_filter(self._args,
+                                     arg_types=["gh_operator"],
```
arporter: I'd only touched this line because of changing ``args_filter()`` to be a module function rather than a method. However, you're right so I've now changed the ``arg_types`` argument too.

eikehmueller:
In answer to Andrew's questions:

- I assume by 'apply' kernels you mean the one in ```columnwise_op_app_kernel_mod.F90```? This implements to operation y += A.x where A is a CMA and y, x are fields. So yes, apply kernels should always take in one field and write (or increment) another field. They should also read a CMA.
- an assembly kernel (see ```columnwise_op_asm_kernel_mod.F90```) assembles a CMA from a LMA. So at the moment at least it should read a LMA and write a CMA. There might be cases where it reads more than one LMA (when assembling the diagonal part of D^T.M_{u,h,inv}.D) and it might also be useful to pass in fields and global constants, so I would leave those options open. But the only thing that ever gets written in an assembly kernel is a CMA.
- my understanding of stencil operations is that those are operations which involve data in neighbouring columns. So then you are correct, stencil operations are forbidden (a CMA should only operate on data in one column).
- it is probably easiest to assume that CMA kernels always have vertical dependencies (i.e. you can't parallelise in the vertical direction). There are never any horizontal dependencies in CMA operations, i.e. columns can be processed independently
- that depends on the matrix-matrix kernel. If that kernel adds two matrices C = A + B, then: to[A] = to[B] = to[C], from[A] = from[B]=from[C]. If it multiplies two matrices, C = A*B, then to[C] = to[A], from[A] = to[B], from[B] = from[C]. But you could also imagine a mixture of the two: C = A+B*D. So I think it will be difficult to come up with a general set of rules without kernel-code introspection. Can we just leave them out? The number of those kernels is small (probably just C=A+B, C=A*B), and those kernels are never written by a science user.

arporter:
That helps a lot @eikehmueller , thank you. One follow-up question: when I said "field vector" I meant whatever quantity it is that we denote in PSyclone meta-data using e.g. ``GH_FIELD*n`` where ``n`` is some integer (often 3 for x, y and z). Should we permit such quantities as arguments to kernels that apply CMA operators to fields?

eikehmueller:
@arporter: Could you point me to an example where ```GH_FIELD``` is used? Thanks!

arporter:
@eikehmueller : There's ``src/test/kernel/columnwise_op_app_kernel_mod_test.pf`` on Chris' ``r7170_columnwise_operator`` branch. i.e. the example of applying a column-wise operator to a field.
If you meant an example of the use of a field vector then see ``Dynamo_trunk/src/dynamo/kernel/weighted_m3_rho_kernel_mod.F90`` for an example.

eikehmueller:
@arporter: if by "field-**vector**" you mean ```GH_FIELD*n``` as opposed to a single ```GH_FIELD```, then I can't currently think of an example where this would be the case. The apply kernel should operator on single ```GH_FIELD``` objects, though, as the ```columnwise_op_app_kernel_type``` kernel in the test you mention.

rupertford:
@eikehmueller yes field-vector is the name the Met Office (and therefore we) use for gh_field*n. I was going to point you to the documentation on it but that section still needs to be addded :-( https://psyclone.readthedocs.io/en/stable/dynamo0p3.html#algorithm

arporter:
Since stencils may only be specified on read-only fields, they are already forbidden from appearing in the meta-data for matrix-matrix kernels. I've now added checks and tests that they are forbidden in CMA assembly and application kernels.

arporter:
Have moved existing check that there are no vector args (or stencil accesses) up to top level of `_identify_cma_op` so that it is applied to any kernel that has a CMA operator argument.

arporter:
Back to Rupert for 2nd review...

arporter:
1. Fixed - thanks.
2. Fixed.
3. Done. pylint gives 10/10 :-)

PR #28, opened 2017-04-26
Dependence analysis

rupertford:
Adds dependence analysis to PSyclone and a move transformation to test it out with.


doc/psy_layer.rst
```diff
@@ -108,10 +108,10 @@ API
...
 
-A PSy **Schedule** object consists of a tree of objects which can be
-used to describe the required schedule for a PSy layer
-subroutine. This subroutine is called by the Algorithm layer and
-itself calls one or more Kernels and/or implements any required
+A PSy **Schedule** object consists of a tree of objects (called Nodes
+in PSyclone) which can be used to describe the required schedule for a
+PSy layer subroutine. This subroutine is called by the Algorithm layer
+and itself calls one or more Kernels and/or implements any required
 Built-in operations. These objects can currently be a **Loop**, a
```
arporter: s/These objects/The Node objects/ because of the intervening sentence describing the PSy layer subroutine.

doc/psy_layer.rst
```diff
@@ -131,3 +131,66 @@ merging, inlining, OpenMP parallelisation etc.). The tree could be
...
+   >>> schedule.dag(file_name="lovely", file_format="png")
+
+.. note:: The dag method can be called from any node and will will
+          output the dag for that node and all of its children.
+
+if we were to look at the dynamo eg6 example we would see the
+following image:
+
+.. image:: dag.png
+	   :width: 256
```
arporter: Nice. However, when I generate the pdf version of the docs, this image extends below the bottom of the page!

src/transformations.py
```diff
@@ -1064,3 +1064,74 @@ def apply(self, node, const_bounds=True):
...
+        # This also checks the location and position arguments
+        if not node.is_valid_location(location, position=position):
+            raise TransformationError(
+                "In the Move transformation apply method, data dependencies "
+                "forbid the move to the new location")
+
+    def apply(self, node, location, position="before"):
+        '''Move the node represented by :py:obj:`node` before location
+        :py:obj:`location` by default and after if the optional
+        `position` argument is set to 'after' '''
```
arporter: Also, worth emphasizing that this method uses the dependency analysis to verify that the requested transformation is valid.

doc/getting_going.rst
```diff
@@ -112,6 +114,43 @@ you can instruct pip to do a user-local install:
...
+::
+   > sudo apt install graphviz
+
+Should you wish to, uninstalling is simply performed by doing:
+::
+   > sudo apt remove graphviz
+
+If you are using a different operating system or version of linux that does
+not support ``apt`` please see the download instructions here
+http://www.graphviz.org/Download..php
```
rupertford: OK, I've reordered.

doc/getting_going.rst
```diff
@@ -112,6 +114,43 @@ you can instruct pip to do a user-local install:
...
+
+If you do not have sufficient privileges for a system-wide install then
+you can instruct pip to do a user-local install:
+::
+    > pip install --user graphviz
+
+If graphviz itself is not installed on your system then it can be
+simply installed using ``apt``
+::
+   > sudo apt install graphviz
```
rupertford: Please see the comments for line 152 which address this.

examples/dynamo/eg6/omp_script.py
```diff
@@ -22,5 +22,6 @@ def trans(psy):
 
     # take a look at what we've done
     schedule.view()
+    schedule.dag(file_format="png")
```
rupertford: Fixed pep8 error.

rupertford: I agree, I will modify the implementation so that nothing is output if the python graphviz bindings are not available. We should write out a warning to a log file in this case. I will add a comment in the code to state this.

arporter:
After our chat today, this is just a note to remind you to check whether any x-failing tests should now be fixed. I've also just spotted that `test_fuse_colour_loops` could make use of the `move` transformation.

src/dynamo0p3.py
```diff
@@ -1301,13 +1301,15 @@ def __init__(self, alg_invocation, idx):
                         # 1 to the vector size which is what we
                         # require in our Fortran code
                         for idx in range(1, halo_field.vector_size+1):
-                            exchange = DynHaloExchange(
-                                halo_field, parent=loop, vector_index=idx,
-                                inc=inc)
+                            exchange = DynHaloExchange( halo_field,
```
rupertford: Oh yes, oops. Now fixed.

src/dynamo0p3.py
```diff
@@ -1301,13 +1301,15 @@ def __init__(self, alg_invocation, idx):
...
+                                                        parent=loop.parent,
+                                                        vector_index=idx,
+                                                        inc=inc)
                             loop.parent.children.insert(loop.position,
                                                         exchange)
                     else:
-                        exchange = DynHaloExchange(halo_field, parent=loop,
+                        exchange = DynHaloExchange(halo_field,
```
rupertford: Yes, I found this bug when adding tests in this issue. I could argue that it is therefore now covered, as one of the new tests will fail if the bug is reintroduced. However, I think you are right that there should be a specific test that halo exchanges have the correct parent set up. I will therefore add a test.

src/psyGen.py
```diff
@@ -701,6 +664,182 @@ def gen_code(self, parent):
 class Node(object):
     ''' baseclass for a node in a schedule '''
 
+    def dag(self, file_name='dag', file_format='svg'):
+        '''Create a dag of this node and it's children'''
```
rupertford: Done.

src/psyGen.py
```diff
@@ -701,6 +664,182 @@ def gen_code(self, parent):
...
+        import graphviz as gv
```
rupertford: Agreed. Already done as a response to an earlier comment.

src/psyGen.py
```diff
@@ -701,6 +664,182 @@ def gen_code(self, parent):
...
+        try:
+            graph = gv.Digraph(format=file_format)
+        except ValueError:
+            raise GenerationError(
+                "unsupported graphviz file format '{0}' provided".
+                format(file_format))
+        self.dag_gen(graph)
+        graph.render(filename=file_name)
+
+    def dag_gen(self, graph):
```
rupertford: Good point. It's not obvious. I have added comments.

src/psyGen.py
```diff
@@ -701,6 +664,182 @@ def gen_code(self, parent):
...
+        implementation assumes the node has no directly associated
+        arguments (i.e. is not a Call class or subclass). Arguments of
+        any of this nodes descendents are considered to be
+        associated. '''
+        args = []
+        for call in self.calls():
+            args.extend(call.args)
+        return args
+
+    def backward_dependence(self):
```
rupertford: Done.

src/psyGen.py
```diff
@@ -701,6 +664,182 @@ def gen_code(self, parent):
...
+        dependence = None
+        for arg in self.args:
+            dependent_arg = arg.forward_dependence()
+            if dependent_arg:
+                node = dependent_arg.call
+                while node.depth > self.depth:
+                    node = node.parent
+                if self.sameParent(node):
+                    if not dependence:
+                        dependence = node
```
rupertford: I believe you understand the logic now but for the record we are looking for the closest dependence. There may be more than one and the closest is the one that constrains us. Therefore we do not return when we find the first dependence, we search through all of them.

rupertford: I have added comments.

src/psyGen.py
```diff
@@ -783,7 +922,7 @@ def _find_position(self, children, position):
             position += 1
             if child == self:
                 return True, position
-            if isinstance(child, Loop):
+            if child.children:
```
rupertford: Yes, the routine was meant to return a unique id for each node but was coded to only recurse into loops with children (as they were the only examples at the time). But we now have directives as well and may add other things in the future. I changed the code so that it works generically. The new tests for dependence analysis test the corrected code as we use the global id when checking for forward or backward dependencies.

src/psyGen.py
```diff
@@ -1257,7 +1426,29 @@ class GlobalSum(Node):
     manipulated in, a schedule. '''
     def __init__(self, scalar, parent=None):
         Node.__init__(self, children=[], parent=parent)
-        self._scalar = scalar
+        import copy
+        self._scalar = copy.copy(scalar)
+        if scalar:
+            # update scalar values appropriately
+            # HACK:TODO: update mapping to readwrite when it is supported
```
rupertford: Yes we do zero sums. However, this is the global sum object which is a different beast. This takes partial sums from each mpi process (which indeed are each zero'ed), sums them up then updates the partial sums on each mpi process with the summed value.

src/psyGen.py
```diff
@@ -1272,14 +1463,40 @@ class HaloExchange(Node):
...
+        ''' Return the name to use in a dag for this node'''
+        name = ("haloexchange({0})_".format(self._field.name) +
+                str(self.position))
+        if self._check_dirty:
+            name = "check" + name
+        return name
+
+    @property
+    def args(self):
+        '''Return the list of arguments associated with this node. Overide the
```
rupertford: Done.

src/psyGen.py
```diff
@@ -1291,6 +1508,15 @@ def view(self, indent):
 class Loop(Node):
 
     @property
+    def dag_name(self):
+        ''' Return the name to use in a dag for this node'''
+        if self.loop_type:
+            name = "loop_[{0}]_".format(self.loop_type) + str(self.position)
```
rupertford: Well spotted. Modified a test to include this line.

src/psyGen.py
```diff
@@ -1819,6 +2051,11 @@ def __init__(self):
         self._fs_descriptors = None
         self._reduction = None
 
+    @property
+    def dag_name(self):
+        ''' Return the name to use in a dag for this node'''
+        return "builtin_" + str(self.abs_position)
```
rupertford: I did this then realized we are using the kernel name for builtins (hence they are all called xx_code) when we should be using the metadata type name i.e. the one we specify in the algorithm layer. I had to change a few tests but all tests are passing now. I guess I could have done this as a different issue but I thought I might as well while I'm here!

src/psyGen.py
```diff
@@ -1864,16 +2101,31 @@ def iteration_space_arg(self, mapping={}):
                               "we assume there is at least one writer, "
                               "reader/writer, or increment as an argument")
 
-    def set_dependencies(self):
-        for argument in self._args:
-            argument.set_dependencies()
-        # TODO create a summary of dependencies
+    def args_filter(self, arg_types=None, arg_accesses=None):
```
rupertford: Whilst I agree that this (and presumably args_filter in the Loop class) should be replaced by psyGen.args_filter, this appears to still be used on my branch by args_filter in the loop class which is then used in dynamo0p3.py and transformations.py so I don't think it is simply a case of removing it. args_filter already existed so I'm not sure why it is showing up as a change in this issue.

src/transformations.py
rupertford: OK, I've hopefully done this although I did not want to make too big a deal about the dependence checking.

src/tests/dynamo0p3_transformations_test.py
```diff
@@ -3350,3 +3351,156 @@ def test_list_multiple_reductions():
...
+    move_trans.apply(schedule.children[initial_index],
+                     schedule.children[target_index],
+                     position="after")
+
+    new_arg = schedule.children[target_index+1]
+    assert orig_arg == new_arg
+
+
+def test_move_forward():
+    '''Test that MoveTrans moves the node forewards to the expected
```
rupertford: Done.

src/tests/dynamo0p3_transformations_test.py
```diff
@@ -3350,3 +3351,156 @@ def test_list_multiple_reductions():
...
+    move_trans.apply(schedule.children[initial_index],
+                     schedule.children[target_index])
+
+    new_arg = schedule.children[target_index-1]
+    schedule.view()
+    assert orig_arg == new_arg
+
+
+def test_move_forward_after():
+    '''Test that MoveTrans moves the node forewards to the expected
```
rupertford: Done.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,794 @@ def test_invalid_reprod_pad_size():
...
+    schedule = invoke.schedule
+    arg_f1_inc_1 = schedule.children[0].children[0].arguments.args[0]
+    arg_f1_inc_2 = schedule.children[2].children[0].arguments.args[0]
+    arg_f2_read_1 = schedule.children[0].children[0].arguments.args[2]
+    arg_f2_inc = schedule.children[1].children[0].arguments.args[0]
+    arg_f2_read_2 = schedule.children[2].children[0].arguments.args[1]
+    # different names returns False
+    assert not arg_f2_inc._dependent_arg(arg_f1_inc_1)
+    # same name both reads returns False
+    assert not arg_f2_read_1._dependent_arg(arg_f2_read_2)
```
rupertford: _dependent_arg just looks at two argument objects and specifies whether there is a dependence between them or not based on their name and their access. Different names do not depend on each other and if both accesses are read then they do not depend on each other. The logic here is quite simple and is what all the more complex methods are built on. Could you clarify what you think needs changing as it looks OK to me (but of course I'm naturally blinkered about my own code).

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,794 @@ def test_invalid_reprod_pad_size():
...
+    # same name both writes (the 4.5 example only uses inc) returns True
+    _, invoke_info = parse(
+        os.path.join(BASE_PATH, "15.1_builtin_and_normal_kernel_invoke.f90"),
+        distributed_memory=False, api="dynamo0.3")
+    psy = PSyFactory("dynamo0.3", distributed_memory=False).create(invoke_info)
+    invoke = psy.invokes.invoke_list[0]
+    schedule = invoke.schedule
+    arg_f1_write_1 = schedule.children[0].children[0].arguments.args[1]
+    arg_f1_write_2 = schedule.children[1].children[0].arguments.args[1]
+    assert arg_f1_write_1._dependent_arg(arg_f1_write_2)
```
rupertford: I had this as False for quite a while then realised that you should never have two writers following each other so thought it should probably raise an exception or warning. In the end I went with the simplest solution which was to allow it for the moment. Once you allow it then there is a dependence as by changing the order of writers you may get different results for any subsequent read. For example

a=1
a=2
b=a

The first and second writes to "a" depend on each other as we can not reorder them because b will get a different result.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,794 @@ def test_invalid_reprod_pad_size():
...
+    result = kern_asum_arg._find_argument(schedule.children)
+    assert result == glob_sum_arg
+    # b) kern arg depends on globalsum arg
+    result = glob_sum_arg._find_argument([schedule.children[3].children[0]])
+    assert result == kern_asum_arg
+
+
+@pytest.mark.xfail(reason="gh_readwrite not yet supported in PSyclone")
+def test_globalsum_arg():
+    '''Check that the globalsum argument is defined as gh_readwrite and
```
rupertford: gh_sum data is zero'ed first. The globalsum object takes partial sums from each MPI process, sums them together and updates them with the summed value on all MPI processes. It therefore has gh_readwrite access.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,794 @@ def test_invalid_reprod_pad_size():
...
+    # b) previous
+    assert prev_dep_loop_node.backward_dependence() == loop3
+    # 3: haloexchange dependencies
+    _, invoke_info = parse(
+        os.path.join(BASE_PATH, "4.5_multikernel_invokes.f90"),
+        distributed_memory=True, api="dynamo0.3")
+    psy = PSyFactory("dynamo0.3", distributed_memory=True).create(invoke_info)
+    invoke = psy.invokes.invoke_list[0]
+    schedule = invoke.schedule
+    schedule.dag()
```
rupertford: Agreed. This should not have been there.

examples/dynamo/eg6/omp_reprod_script.py
```diff
@@ -28,5 +28,6 @@ def trans(psy):
 
     # take a look at what we've done
     schedule.view()
+    schedule.dag()
 
     return psy
```
rupertford: Fixed.

doc/psy_layer.rst
rupertford: Done.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,794 @@ def test_invalid_reprod_pad_size():
...
+    for idx, arg in enumerate(all_args):
+        assert arg == schedule_args[idx]
+    # 2) Loop1
+    loop1_args = loop1.args
+    for idx, arg in enumerate(kern1.arguments.args):
+        assert arg == loop1_args[idx]
+    # 3) Loop2
+    loop2_args = loop2.args
+    for idx, arg in enumerate(kern2.arguments.args):
+        assert arg == loop2_args[idx]
```
rupertford: Good point. I've added a test for that.

doc/psy_layer.rst
rupertford: That is weird. It is fine for me. Could you try again in case any minor changes to the documentation text have fortuitously sorted out the layout.

rupertford:
I've just checked the xfailing tests and there are two associated with dependence analysis. However, these tests are associated with determining the function space of arguments to kernels when they are specified as any_space (or something else generic). Whilst we have dependence analysis in this pull request we have not yet written the logic to try to determine an arguments function space (or check the validity of its use). I have therefore raised a new issue #30 which will do this.

rupertford:
Very well spotted about the use of explicit move's in the dynamo0p3_transformations_test.py test_fuse_colour_loops test. There were quite a few others in the file as well. I've updated them all to use MoveTrans instead. One of them was actually invalid :-).

rupertford:
I've updated the system specific installation documentation as that did not include installing graphviz.

I think we are ready for second review from @arporter.

examples/dynamo/eg6/omp_script.py
arporter: I've uninstalled the graphviz python bindings and successfully run the examples and the tests. Great.

doc/getting_going.rst
```diff
@@ -112,6 +114,41 @@ you can instruct pip to do a user-local install:
...
+graphviz
+^^^^^^^^
+
+The data dependencies of a PSy-layer schedule (see Section
+:ref:`psy-layer-schedule`) determine the validity of changes to a
+schedule. PSyclone supports the visualisation of these dependencies as
+a graph using graphviz. This visualisation is not needed to use
+PSyclone but will cause one of the tests to fail if graphviz is not
+installed.
+
```
arporter: Happily this is no longer true so this text needs updating.

doc/psy_layer.rst
```diff
@@ -131,3 +131,66 @@ merging, inlining, OpenMP parallelisation etc.). The tree could be
...
+dot format is output with the name ``dag`` and a file in svg format is
+output with the name ``dag.svg``. The file name can be changed using
+the ``file_name`` optional argument and the output file format can be
+changed using the ``file_format`` optional argument. The file_format
+value is simply passed on to graphviz so the graphviz documentation
+should be consulted for valid formats if svg is not required.
+::
+   >>> schedule.dag(file_name="lovely", file_format="png")
+
+.. note:: The dag method can be called from any node and will will
```
arporter: Repeated "will"

doc/psy_layer.rst
arporter: If I look at the top-level psyclone.pdf that you have generated then it is fine. If I generate my own psyclone.pdf then I still have the problem. I think we can say that it is my problem rather than yours though.

doc/psy_layer.rst
```diff
@@ -131,3 +131,66 @@ merging, inlining, OpenMP parallelisation etc.). The tree could be
...
+if we were to look at the dynamo eg6 example we would see the
+following image:
+
+.. image:: dag.png
+	   :width: 256
+
+In the image, all nodes (Psyclone's generic name for objects in the
+schedule) with children are split into a start vertex and an end
+vertex (for example the schedule node has a `schedule_start` and a
+`schedule_end` vertex). Blue arrows indicate that there is a parent to
```
arporter: suggest: (for example the Schedule node has both `schedule_start` and `schedule_end` vertices).

doc/psy_layer.rst
```diff
@@ -131,3 +131,66 @@ merging, inlining, OpenMP parallelisation etc.). The tree could be
...
+child relationship (from a start node) or a child to parent
+relationship (to an end node). Green arrows indicate that there is a
+forward dependence. Therefore the OMP parallel loop must complete
+before the globalsum is performed. Red arrows indicate that there is a backward dependence. However
+the direction of the red arrows are reversed to improve the flow of the dag
+layout. In this example the forward and backward dependence is the
+same, however this is not always the case. The two built-ins do not
+depend on each other, so they have no associated green or red arrows.
```
arporter: Please could you explain what you mean by "forward dependence" and "backward dependence." I was thinking that I should know but I don't and a quick google doesn't help much. Therefore I think we should have some explanation here.

doc/system_specific_setup.rst
```diff
@@ -103,6 +103,19 @@ may be downloaded from https://github.com/stfc/fparser/releases. Simply
 unpack the tarball and ensure that the resulting
 ``fparser-x.y.z/src/fparser`` directory is in your PYTHONPATH.
 
+PSyclone supports the ability to output a schedule dependency graph
+using the graphviz package. This is optional and the associated
+routine will silently return if the graphviz bindings are not
+installed. To output a graph you need to install the graphviz package
+::
+
+   > sudo apt-get install graphviz
```
arporter: This is package-manager specific. Can you simply point to the previous location in the documentation where you cover this?

examples/dynamo/README
```diff
@@ -57,7 +57,10 @@ algorithm layer.
     python ../../../src/generator.py alg.f90
 
 The sixth example illustrates the use and implementation of
-reductions. The example may be run sequentially:
+reductions. It also demonstrates the generation of a schedule's
+dependence graph in svg format (see dag.svg). The example may be run
+sequentially:
+
```
arporter: Need to note what happens if graphviz is not installed.

src/psyGen.py
```diff
@@ -701,6 +664,255 @@ def gen_code(self, parent):
...
+        '''If this Node can be moved to the new_node
+        (where position determines whether it is before of after the
+        new_node) without breaking any data dependencies then return True,
+        otherwise return False. '''
+        # First perform correctness checks
+        # 1: check new_node is a Node
+        if not isinstance(new_node, Node):
+            raise GenerationError(
+                "In the psyGen Call class is_valid_location() method the "
+                "supplied argument is not a Node")
```
arporter: Please add the type of the supplied argument to the error message.

doc/getting_going.rst
rupertford: Updated.

doc/psy_layer.rst
rupertford: Fixed.

doc/psy_layer.rst
rupertford: Fixed.

doc/psy_layer.rst
rupertford: OK, I've explained.

doc/system_specific_setup.rst
rupertford: Yes it is but that is on purpose as we are in the Ubuntu-specific set up section which supports apt

examples/dynamo/README
rupertford: Done.

src/psyGen.py
rupertford: Done.

rupertford:
@arporter I'm not sure if you've finished your second review but I've fixed the issues you have identified so far.

src/psyGen.py
arporter: Much better - thanks.

src/psyGen.py
arporter: Great.

src/psyGen.py
arporter: Aha, thanks for explaining.

src/psyGen.py
```diff
@@ -1272,14 +1537,40 @@ class HaloExchange(Node):
 
     def __init__(self, field, halo_type, halo_depth, check_dirty, parent=None):
         Node.__init__(self, children=[], parent=parent)
-        self._field = field
+        import copy
+        self._field = copy.copy(field)
```
arporter: I was initially concerned that `copy.copy(None)` might not work but I've just checked and it is OK.

src/psyGen.py
arporter: I've just deleted it and all tests pass. Would you mind trying that?

src/psyGen.py
```diff
@@ -1930,43 +2266,74 @@ def call(self):
...
+                for argument in node.arguments.args:
+                    if self._dependent_arg(argument):
+                        return argument
+            elif isinstance(node, HaloExchange):
+                argument = node.field
+                if self._dependent_arg(argument):
+                    return argument
+            elif isinstance(node, GlobalSum):
+                argument = node.scalar
+                if self._dependent_arg(argument):
```
arporter: Could use the new `args` method of `GlobalSum` and `HaloExchange` here. e.g.

    elif isinstance(node, HaloExchange) or isinstance(node, GlobalSum):
        for argument in node.args:
            if self._dependent_arg(argument):
                return argument

src/psyGen.py
```diff
@@ -701,6 +664,256 @@ def gen_code(self, parent):
...
+        if remote_node:
+            # this node has a backward dependence.
+            remote_name = remote_node.dag_name
+            if remote_node.children:
+                # the remote node has children so I will connect to
+                # its end vertex
+                remote_name += end_postfix
+            # Create the backward dependence edge in red.
+            graph.edge(remote_name, local_name, color="red")
+        elif self.parent:
```
arporter: The above if-block (`if remote_node`) is now not covered by the test suite (I'm running with graphviz installed and eg6 produces dag output for me).

src/psyGen.py
```diff
@@ -701,6 +664,256 @@ def gen_code(self, parent):
...
+        if remote_node:
+            # this node has a forward dependence
+            remote_name = remote_node.dag_name
+            if remote_node.children:
+                # the remote node has children so I will connect to
+                # its start vertex
+                remote_name += start_postfix
+            # Create the forward dependence edge in green
+            graph.edge(local_name, remote_name, color="green")
+        elif self.parent:
```
arporter: The above if-block (`if remote_node`) is not covered by the test suite.

src/psyGen.py
arporter: In fact, I've just realised there's a `Call.args()` method too so the code can be further simplified.

src/tests/dynamo0p3_transformations_test.py
```diff
@@ -578,9 +580,8 @@ def test_colouring_multi_kernel():
             # can be removed
             del schedule.children[8:13]
             # f is required but can be moved before the first loop
-            schedule.children.insert(6, schedule.children.pop(7))
-            # In the future we will have transformations to move
-            # elements around with checks for validity.
+            schedule, _ = mtrans.apply(schedule.children[7],
+                                       schedule.children[6])
             index = 7
```
arporter: Nice :-)

src/tests/dynamo0p3_transformations_test.py
```diff
@@ -3350,3 +3369,156 @@ def test_list_multiple_reductions():
...
+    _, info = parse(os.path.join(BASE_PATH,
+                                 "4.2_multikernel_invokes.f90"),
+                    api=TEST_API)
+    psy = PSyFactory(TEST_API).create(info)
+    invoke = psy.invokes.invoke_list[0]
+    schedule = invoke.schedule
+    move_trans = MoveTrans()
+    with pytest.raises(TransformationError) as excinfo:
+        move_trans.apply(None, schedule.children[0])
+    assert ("In the Move transformation apply method the "
```
arporter: We should also check that something sensible happens when the second argument to the `apply()` method is not a `Node`.

src/tests/psyGen_test.py
arporter: My trouble is that my understanding of dependencies was (and is) coloured by the Habakkuk work where a variable has a dependency on *another* variable if the latter appears on the RHS of an expression that assigns to the former.
I also know that I don't know (to quote Al Gore?) the terminology used in Computer Science when describing this type of thing. I'm assuming that you do and that's what you've used. I would find it slightly easier to understand if `_depends_arg` (which returns a boolean rather than an argument) was renamed, e.g. to `_depends_upon()`.
It would also help if you elaborated on what you mean by "dependency between" in the doc-string of that method.
Apologies if I'm being dense but I figure if that if it's not very clear to me then it won't be to (at least some) other people either. (I find this kind of thing very hard - how much knowledge do we assume and what do we spell out?)

src/tests/psyGen_test.py
arporter: We could follow in the footsteps of compiler writes and silently optimise-out the second write :-) Anyway, I understand now. Again this highlights my misunderstanding of the use of the word "dependence". I would argue that, strictly speaking, it is `b=a` that depends on `a=2`. The two assignment statements do not depend on each other.

src/tests/psyGen_test.py
arporter: Ah yes. OK.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,843 @@ def test_invalid_reprod_pad_size():
...
+                               same_space=True)
+    loop = schedule.children[0]
+    kern1 = loop.children[0]
+    kern2 = loop.children[1]
+    loop_args = loop.args
+    kern_args = kern1.arguments.args
+    kern_args.extend(kern2.arguments.args)
+    for idx, arg in enumerate(kern_args):
+        assert arg == loop_args[idx]
+
```
arporter: Great.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,843 @@ def test_invalid_reprod_pad_size():
...
+    psy = PSyFactory("dynamo0.3",
+                     distributed_memory=False).create(invoke_info)
+    invoke = psy.invokes.invoke_list[0]
+    schedule = invoke.schedule
+    my_file = tmpdir.join('test')
+    schedule.dag(file_name=my_file.strpath)
+    assert not os.path.exists(my_file.strpath)
+    if keep:
+        sys.modules['graphviz'] = keep
+
```
arporter: Great.

src/tests/psyGen_test.py
```diff
@@ -810,3 +815,843 @@ def test_invalid_reprod_pad_size():
...
+def test_node_dag(tmpdir):
+    '''test that dag generation works correctly. Skip the test if
+    graphviz is not installed'''
+    graphviz_installed = False
+    try:
+        import graphviz
+        graphviz_installed = True
+    except ImportError:
+        pass
+    if graphviz_installed:
```
arporter: Could you just return here (1633) and then you wouldn't have to have the rest of the test inside an `if` block?

doc/system_specific_setup.rst
arporter: Oops, sorry.

arporter:
@rupertford I've finished my second review. There are some minor bits of tidying, a couple of sections missing from code coverage (I think they were covered when I did my first review, either that or I missed them) and a discussion to be had about the meaning of "has a dependency" :-)

src/psyGen.py
rupertford: You are quite right. args_filter in the loop class uses psyGen.args_filter, not this one. have removed it.

src/tests/psyGen_test.py
rupertford: I agree about changing the name, so have done so.

The terminology is not standard as there does not appear to be terminology for what I need. What I did was borrow the terminology from loop dependence analysis where a forward dependence is one from a future loop iteration and a backward dependence is one from a previous loop iteration. I've hopefully explained this in enough detail in the doc string as requested.

src/tests/psyGen_test.py
rupertford: Actually write-after-write (waw) is a recognized form of dependence :-).

I agree that two consecutive writes does not make sense and could be optimized out. You can write a transformation for that if you like!

I think it would be better to issue a warning and/or raise an exception in the future.

doc/psy_layer.rst
rupertford: Phew. If I make the image smaller the quality starts degrading (when generated on my system) so I didn't want to do that!.

src/psyGen.py
rupertford: I hadn't thought of that as a possibility so I'm pleased it works!

src/psyGen.py
rupertford: There is an args method for all nodes :-). I have changed the code as requested.

src/psyGen.py
rupertford: Lines are now covered.

src/psyGen.py
rupertford: Lines are now covered.

src/tests/dynamo0p3_transformations_test.py
rupertford: The second argument is passed into the node's is_valid_location() method and is checked there. The associated test is test_node_is_valid_location() in psyGen_test.py

src/tests/psyGen_test.py
rupertford: Quite right. I have done as suggested.

rupertford:
@arporter, I have hopefully addressed the comments from your second review so am ready for third review.

src/tests/dynamo0p3_transformations_test.py
arporter: OK.

src/psyGen.py
```diff
@@ -2302,26 +2281,43 @@ def _find_argument(self, nodes):
...
+        where a forward dependence indicates a dependence in a future
+        loop iteration and a backward dependence indicates a
+        dependence on a previous loop iteration. Note, we currently
+        assume that any read or write to an argument results in a
+        dependence i.e. we do not consider the internal structure of
+        the argument (e.g. it may be an array). However, this
+        assumption is OK as all elements of an array are typically
+        accessed. However, we may need to revisit this when we change
+        the iteration spaces of loops e.g. for overlapping
+        communication and computation. '''
```
arporter: That's great - even I understand it now :-)

Issue #32, opened 2017-05-11
Build generated code when testing


arporter:
At the moment we do not build any code when performing PSyclone tests. Therefore we do not know if the code will compile or not, we just trust this by inspection. This will help reduce mistakes by us before shipping new versions to users. In this ticket we will look at adding the building of code as part of our testing system i.e. within py.test. This was previously 469 on SRS.

arporter:
Created `test_compilation` branch.

arporter:
In order to be able to (nicely) generate a list of the Use statements in the PSy layer, I've had to re-jig things so that we store the f2pygen AST within the PSy object - previously we just created it when `gen()` was called and then immediately disposed of it. I've then added a `walk()` method to the f2pygen `BaseGen` class so that we can walk down the AST and find nodes of a particular type. This then allows us to pull out all of the `UseGen` nodes.
I've got a hard-coded list of 'infrastructure' modules (e.g. `field_mod`). I've done it this way so that we can ensure they are compiled in the correct order to satisfy dependencies between them.

I've added the new compilation-check to two tests so far and both fail. One because our `testkern.F90` doesn't have the necessary use statements for things like `kernel_type`. The other fails because my current stub infrastructure modules are missing key elements.

arporter:
Now have compilation test working for a built-in example. I think I can move the `walk` method I added to `PSy` and put it as a module routine in `utils`. I also need to add a test that the user-supplied compiler and flags work.

arporter:
Now have working compilation tests for both a built-in and a user-supplied kernel. I've also added a few tests for the new code that performs these tests.
Next step is to remove commented-out code from the infrastructure modules.

rupertford:
Pull request #39 has been merged to master. Closing this issue.

PR #39, opened 2017-05-19
Add ability to compile generated code during testing

arporter:
This PR is for work done under #32.


arporter:
Hi @rupertford, not sure whether we should merge this on before or after our freeze for pypi.

arporter:
@rupertford I've brought the associated branch up-to-date with master so that it now has the new directory structure (this then lets me use it in other branches without too much pain).

arporter:
Realised that we can add command-line options to py.test using `addopts`. I have therefore set-up some fixtures in `conftest.py`. A test that involves compilation should now have the `@utils.compile` decorator. Such tests are then only performed if the `--compile` flag is passed to py.test. The default compiler is `gfortran` but this, and the compile flags, can be changed using the `--f90=my_compiler` and `--f90flags="some flags"` options.

As part of this change I have removed the look-up of the F90 and F90FLAGS environment variables.

src/psyclone/tests/utils.py
```diff
@@ -28,3 +50,141 @@ def count_lines(root, string_name):
...
+    if api not in supported_apis:
+        raise CompileError("Unsupported API in code_compiles. Got {0} but "
+                           "only support {1}".format(api, supported_apis))
+
+    if api == "dynamo0.3":
+        base_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                                 "test_files", "dynamo0p3")
+        from dynamo0p3_build import INFRASTRUCTURE_MODULES as module_files
+
+    kernel_modules = []
```
rupertford: I think kernel_modules can be found simply by psy_gen.invokes.names. This would remove the need for the psy_module change to dynamo0p3.py and psyGen.py and the walk function above

src/psyclone/tests/utils.py
```diff
@@ -28,3 +50,141 @@ def count_lines(root, string_name):
         if string_name in line:
             count += 1
     return count
+
+
+def walk(parent, my_type):
```
rupertford: I think this can be removed if the changes proposed below are possible (see next comment).

src/psyclone/tests/utils.py
```diff
@@ -28,3 +50,141 @@ def count_lines(root, string_name):
...
+        print output
+        if error:
+            print "========="
+            print error
+        raise CompileError(output)
+    else:
+        return True
+
+
+def code_compiles(api, psy_ast, tmpdir, f90, f90flags):
```
rupertford: If we can integrate the build tests with existing correctness tests then it might be better to pass in the code as a string here rather than the ast (psy_ast), as it will already have been generated within the test. If we are going for different correctness and compile tests then there is no need to change.

src/psyclone/tests/utils.py
arporter: Well spotted. It's not quite as simple as you suggest - I have to dig down to the kernel objects and then get their associated `module_name` - but it's nicer.

src/psyclone/dynamo0p3.py
```diff
@@ -922,27 +922,36 @@ def gen(self):
         :rtype: ast
 
         '''
+        # return the AST for the generated code
+        return self.psy_module.root
+
+    @property
```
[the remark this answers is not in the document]
arporter: Agreed - I've removed it.

src/psyclone/psyGen.py
```diff
@@ -214,6 +214,7 @@ def __init__(self, invoke_info):
 
         self._name = invoke_info.name
         self._invokes = None
+        self._psy_module = None
```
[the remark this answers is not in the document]
arporter: Removed.

src/psyclone/psyGen.py
```diff
@@ -231,6 +232,11 @@ def gen(self):
         raise NotImplementedError("Error: PSy.gen() must be implemented "
                                   "by subclass")
 
+    @property
```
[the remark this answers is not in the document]
arporter: Removed.

src/psyclone/tests/dynamo0p3_build.py
```diff
@@ -0,0 +1,23 @@
+''' Module containing configuration required to build code generated
```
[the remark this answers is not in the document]
arporter: Oh yes - added.

src/psyclone/tests/utils.py
arporter: Now that we look for USE'd modules using the `psy_ast` we cannot replace it with a string. In fact this means that my optimisation to store the generated AST is now valid again but, as it is an optimisation, I'll leave it as it is for now.

src/psyclone/tests/utils.py
arporter: Removed.

arporter:
I've added `utils.COMPILE_TEST` which is set according to whether or not the `--compile`` flag is passed to py.test. This can then be used within existing tests to decide whether or not to test the compilation of the generated code. I've modified the two existing examples of testing compilation to use this rather than having duplicate tests.
I feel the need to document this new feature but this is very much something for a developer and not a user. I could begin a new Developing chapter in the existing documentation?

As suggested, I've removed all extraneous comments and code from the infrastructure stubs (including one file that I'd duplicated!). All copyright/license notices have been modified so that the new MO text comes first followed by ours. I think I've addressed all other comments.

Back to @rupertford for 2nd review...

Issue #45, opened 2017-05-26
Kernel metadata - replace evaluator_shape with gh_shape


arporter:
Following changes in the LFRic infrastructure around evaluators and quadrature, there is a requirement to re-name evaluator_shape to gh_shape.

arporter:
Doing this work on branch `rename_evaluator`.

rupertford:
I accidentally merged the rename_evaluator branch to master.
I then reverted it via the GitHub option.
After that I'm not quite sure what happened - I think I tried to merge master back to rename_evaluator after updating master and ended up with all changed from rename_evaluator removed.
Even if I go back to an earlier version of rename_evaluator I am not able to merge it again to master as it removes any of the rename_evaluator changes.
Apparently the standard solution is to revert the revert (on master I presume) and then merge in any subsequent changes. However, I'm not sure about doing this given that I've already subsequently merged something else to master.
Therefore I've taken a manual approach. I've created a new branch rename_evaluator_2 from the latest version of master and have then copied over the modified files from the original rename_evaluator branch. I will then create a new pull request.

rupertford:
pull request #49 has been merged to master. Closing this issue.

PR #46, opened 2017-06-01
Rename evaluator

arporter:
This pull request completes issue #45 and is simply a matter of re-naming meta-data quantities relating to evaluators/quadrature. It is required by SRS issue 1070.


rupertford:
Argggh, I was reviewing a different pull request and just merged this one by accident

PR #49, opened 2017-06-09
issue #45. Recreated changes from original rename evaluator branch onto this one

rupertford:
Original (closed and reverted) pull request was #46


arporter:
Updated documentation in 43b7a42. Doing this revealed we had documentation for features we don't yet support (different shapes of quadrature) so I've removed that for the time being. Presumably the original assumption was that those features would be implemented in the very near future but that was some time ago now and they haven't...

arporter:
Back to @rupertford for review.

rupertford:
Documentation looks good.
Updated documentation pdf and changelog.

PR #63, opened 2017-07-31
Coloured sched

### LINKED ISSUE #54: Use colours when viewing a schedule
PSyclone already has a dependence on numpy and (deep) within numpy there is support for outputting coloured text to a terminal (see https://docs.scipy.org/doc/numpy/reference/distutils.html). It would therefore be nice to make use of this when `view()`ing a schedule.

arporter:
This pull request closes #54.


arporter:
Add test to cover missed code in psyGen.Schedule.view(). All other modifications are covered.
Need to check for pep8 and pylint next...

arporter:
pep8 and pylint are now all fine. Final step (I think) is to update the documentation to describe the need to install termcolor in order to get coloured schedules.

arporter:
pep8 and pylint are now all fine. Final step (I think) is to update the documentation to describe the need to install termcolor in order to get coloured schedules.

arporter:
Documentation done. Ready for first review by @rupertford

doc/getting_going.rst
```diff
@@ -45,15 +45,24 @@ Dependencies
...
+Package Index (pypi.org) and ``pip``. See
 https://packaging.python.org/installing/ for more information.
 
+In addition to the mandatory dependencies just described, PSyclone
+also has optional dependencies on both ``graphviz`` and ``termcolor``.
+PSyclone can use graphviz to produce a visualisation of a schedule's
+dependency graph. If this is desired then the Python package ``graphviz``
+(for the Python bindings) as well as the graphviz package itself must
+be installed. The Python package ``termcolor`` is used for pretty-printing
+a schedule in terminals that support coloured text. If the package is
```
rupertford: Probably worth saying what happens if graphviz is not available seeing you do so for termcolor.

src/psyclone/dynamo0p3.py
```diff
@@ -707,7 +707,7 @@ def __init__(self, ast, name=None):
     def _validate(self, need_evaluator):
         ''' Check that the meta-data conforms to Dynamo 0.3 rules for a
         user-provided kernel or a built-in '''
-        from dynamo0p3_builtins import BUILTIN_MAP
+        from psyclone.dynamo0p3_builtins import BUILTIN_MAP
```
rupertford: why is adding psyclone.* required in this issue?

.travis.yml
```diff
@@ -35,8 +35,16 @@
 language: python
 python:
   - 2.7
+# The env section allows us to specify multiple, different environments
+# in which the test suite must be run. The environment variables set
+# here are picked up in the bin/install_optional.sh script.
+env:
+  matrix:
+    - WITH_TERMCOLOR=0
```
rupertford: This is nice, although does it mean that all tests get run twice, or just the ones which depend on WITH_TERMCOLOR?

src/psyclone/gocean1p0.py
```diff
@@ -104,7 +104,7 @@ class GOInvokes(Invokes):
     ''' The GOcean specific invokes class. This passes the GOcean specific
         invoke class to the base class so it creates the one we require. '''
     def __init__(self, alg_calls):
-        if False:
+        if False:  # pylint: disable=using-constant-test
```
rupertford: nice pylint tidying

src/psyclone/psyGen.py
```diff
@@ -42,8 +42,20 @@
...
+# We use the termcolor module (if available) to enable us to produce
+# coloured, textual representations of Invoke schedules. If it's not
+# available then we don't use colour.
+try:
+    from termcolor import colored
+except ImportError:
+    # We don't have the termcolor package available so provide
+    # alternative routine
+    def colored(text, color):
+        ''' Returns the supplied text argument unchanged. Is a swap-in
```
rupertford: Please change "Is" to "This is"

src/psyclone/psyGen.py
```diff
@@ -74,6 +86,17 @@
...
 
+# Colour map to use when writing Invoke schedule to terminal. (Requires
+# that the termcolor package be installed. If it isn't then output is not
+# coloured.) See https://pypi.python.org/pypi/termcolor for details.
+SCHEDULE_COLOUR_MAP = {"Schedule": "yellow",
+                       "Loop": "blue",
+                       "GlobalSum": "cyan",
+                       "Directive": "green",
+                       "HaloExchange": "magenta",
+                       "Call": "red",
```
rupertford: Should BuiltIns have a different colour to Kernel calls?

src/psyclone/psyGen.py
```diff
@@ -1178,11 +1208,18 @@ def gen_code(self, parent):
 class Directive(Node):
 
     def view(self, indent=0):
-        print self.indent(indent) + "Directive"
+        ''' Print a text representation of this node to stdout '''
```
rupertford: Added in a comment string. Whoop. You haven't done so in all cases where a view method is changed but it is not a requisite for this issue so any improvement is a bonus.

doc/getting_going.rst
arporter: Done.

src/psyclone/dynamo0p3.py
arporter: It's something I should have done when I re-structured the repository. Now pylint complains if we
import psyclone packages using a relative path so I'm fixing it as I come across it.

.travis.yml
arporter: It means that all tests get run twice. I think this is necessary as I wouldn't want to rely on people cherry-picking which tests to run for a given environment. In theory we could pick-up the value of WITH_TERMCOLOR in the code that sets-up the pytest fixtures. However, the tests are written in such a way that they are valid with or without termcolor being installed - there's no explicit reference to that package within them.

src/psyclone/psyGen.py
arporter: Done, plus, inspired by the RSE conference, I've extended the docstring to fully document the interface.

src/psyclone/psyGen.py
arporter: Well, they could :-) I've taken this to mean you'd like them to and since I am agnostic I've changed the colour map. There are only two other colors (grey and white) available and grey is pretty much invisible on my system. I'm therefore using red and magenta for built-ins and Kernels, blue and cyan for halo swaps and global sums and white for loops.

arporter:
I've addressed all comments and questions. In a fit of enthusiasm I've added docstrings to any view() methods that I've touched and have extended existing docstrings to describe interfaces. The latter is a bid to reduce the number of "what type of thing have I got here?" => `print type(thing), dir(thing)` during PSyclone development. Back to @rupertford for 2nd review.

PR #68, opened 2017-08-17
Fix #67, added test cases.

### LINKED ISSUE #67: Crash when printing Invoke class
A minor issue: while adding some debug prints to study some code, I caused a crash when printing an instance of psygen::Invoke().
Reason is that ``__str__`` is:
```
    def __str__(self):
        return self._name+"("+str(self.unique_args)+")"
```
And ``self.unique_args`` is not defined. And while I have a test case and fix ready for this, I realised that ``self.unique_args`` is never used or set anywhere in PSyclone - looks like a left over. Unfortunately I am not sure what should be used. Maybe just return ``self._name`` in ``__str__()`` (but even then it needs a minor fix in case that ``Invoke(None, None, None)`` is used since ``self._name`` is then still undefined ... set``self._name`` to "Undefined" perhaps in this special case??).

hiker:
Hi,

as discussed here the patch. I added a few tests for the various parameters as well. Unfortunately I couldn't figure out (quickly enough) how to test the following 'elif' case:
```
 # create a name for the call if one does not already exist
        if alg_invocation.name is not None:
            self._name = alg_invocation.name
        elif len(alg_invocation.kcalls) == 1 and \
             alg_invocation.kcalls[0].type == "kernelCall":
```
with a 'custom' test case - I opted for the easy way out and just parsed a file, which then gives me this one test case I needed. I might need either more time or some help to ``create an alg_invocation`` with one kernel call 'manually'.
I also added a 'setName' function to allow to test the branches in the above test - not sure if you like that, or if you prefer to assign directly to the ``._name`` attribute(??) or what ;)

Feedback welcome!
   Joerg


src/psyclone/psyGen.py
```diff
@@ -381,10 +381,12 @@ class Invoke(object):
     ''' Manage an individual invoke call '''
 
     def __str__(self):
-        return self._name+"("+str(self.unique_args)+")"
+        return self._name+"("+", ".join(map(lambda x:str(x), self._alg_unique_args))+")"
 
     def __init__(self, alg_invocation, idx, schedule_class,
                  reserved_names=None):
+        self._name = "invoke"
+        self._alg_unique_args = []
```
arporter: New lines are covered and make sense. However line 384 is not pep8-compliant.

src/psyclone/tests/psyGen_test.py
```diff
@@ -398,6 +399,38 @@ def test_reset():
...
+
+    _, invoke = parse(
+        os.path.join(BASE_PATH, "1.12_single_invoke_deref_name_clash.f90"),
+        api="dynamo0.3")
+
+    alg_invocation = invoke.calls.values()[0]
+    inv = Invoke(alg_invocation, 0, DynSchedule)
+    assert inv.__str__() == "invoke_0_testkern_type(a, f1_my_field, f1%my_field, m1, m2)"
+
+
```
arporter: Tests are good although might as well be merged into a single test unless there's a good reason for having '1' and '2' separate. Line 431 is not pep8-compliant.

arporter: Also, to keep pylint happy please add `  # pylint: disable=invalid-name` after the test name.

src/psyclone/psyGen.py
arporter: Having said that, pylint thinks the `lambda` is unnecessary. Perhaps use a list comprehension:

    ", ".join([str(arg) for arg in self._alg_unique_args])

hiker:
I merged in the other branches so that this branch now works with recent fparser/psyclone, and fixed the raised pep8/pylint issues. Pylint still has quite a few complains (too many branches), but I did not introduce this :)

src/psyclone/psyGen.py
```diff
@@ -415,10 +415,13 @@ class Invoke(object):
     ''' Manage an individual invoke call '''
 
     def __str__(self):
-        return self._name+"("+str(self.unique_args)+")"
+        return self._name+"("+", ".join([str(arg) for arg in
```
arporter: This line is not pep8-compliant (sorry).

src/psyclone/psyGen.py
```diff
@@ -415,10 +415,13 @@ class Invoke(object):
...
+                                         self._alg_unique_args])+")"
 
     def __init__(self, alg_invocation, idx, schedule_class,
                  reserved_names=None):
```
arporter: Please could you make a start on documenting the interface to this constructor. If you don't know what all the arguments are then please just leave placeholders.

src/psyclone/psyGen.py
hiker: What is the error message? My pylint does not complain:
```
joerg@localhost:~/work/psyclone/src/psyclone> pylint ./psyGen.py 
No config file found, using default configuration
************* Module psyclone.psyGen
...
R:337, 4: Too many branches (13/12) (too-many-branches)
E:467,20: Method 'schedule' has no 'calls' member (no-member)
R:421, 4: Too many branches (15/12) (too-many-branches)
```

arporter: Oh, my mistake! I thought the lack of space around operators would be a problem but maybe it isn't because it's an argument to `return`? Ignore me :-)

src/psyclone/psyGen.py
hiker: OK, I tried :)

Issue #82, opened 2017-09-29
DAG-generation test fails


arporter:
On my VM (graphviz 2.36.0-0ubuntu3.2 and graphviz python bindings 0.8) test_node_dag fails because the formatting of the graphviz output has changed:

    digraph {
	    schedule_start
	    schedule_end
	    loop_0_start
	    loop_0_end
		    loop_0_end -> loop_1_start [color=green]
		    schedule_start -> loop_0_start [color=blue]

compared to:

    digraph {
	    schedule_start
  	    schedule_end
	    loop_0_start
	    loop_0_end
	    loop_0_end -> loop_1_start [color=green]
	    schedule_start -> loop_0_start [color=blue]

On my desktop (graphviz-2.38.0-39.fc25.x86_64 and graphviz python bindings 0.6) this test passes.
Since the differences are only white space it is probably possible to change this test to cope with either case.

arporter:
I've done the fix in `dag_gen_test_fix`. It was straightforward - I replaced the simple string comparison with a regex. The test now passes on both my VM and my desktop. I've checked for pep8 and pylint.
The modified test does not generate code so no compilation step required.

rupertford:
pr #83 has been merged to master. Closing issue.

PR #83, opened 2017-09-29
Dag gen test fix

arporter:
This resolves issue #82 by changing a test to use a regex.


arporter:
Ready for review by @rupertford (sorry but it is only tiny)

Issue #53, opened 2017-07-12
Coded version.py rather than a generated one


rupertford:
At the moment we generated a version.py file after we have run setup.py. This then acts as a central place for documentation etc to pull the correct version.

I would prefer there to be a pre-existing version.py file which is read by setup.py as well as other parts of the distribution as we would then not need to run setup.py before anything else. This issue is observed when trying to build the documentation.

Andy has pointed out that there is a bit of chicken and egg here as setup.py will not be able to see version.py until the distribution is installed, which is done by setup.py.

However, he also said that you could play with the python path within setup.py to point to the correct place (the relative location is known). I would prefer to do this.

arporter:
I had a bit of a look around at how best to do this. This page summarises the options:

https://packaging.python.org/guides/single-sourcing-package-version/

I've gone for a Python version.py module file (means we don't have to worry about parsing/regex) but, in order to avoid having to have psyclone installed, I read that file using execfile() in both conf.py and setup.py.

arporter:
Just need to do pep8/pylint on the three modified files now...

rupertford:
pr #88 is now on master so closing this issue

PR #88, opened 2017-10-03
Coded version number

arporter:
This PR is for work done under #53. It introduces version.py as a static file rather than having it generated during install (in setup.py). This means that PSyclone does not have to have been installed in order to build the documentation.


.gitignore
```diff
@@ -9,4 +9,3 @@ dag
 dag.png
 dag.svg
 dist
-src/psyclone/version.py
```
rupertford: removed the file from the list of files to ignore in git. Good.

doc/conf.py
```diff
@@ -3,29 +3,33 @@
...
 #
 # Note that not all possible configuration values are present in this
 # autogenerated file.
 #
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os
+import sys
+import os
```
rupertford: Changes to make pep8 compliant. Good, but sys is not needed, as specified by pylint. There are a few pylint issues leading to a negative score but it does not really matter that much for this file.

doc/conf.py
```diff
@@ -47,48 +51,56 @@
...
+# it (which would require that PSyclone be installed first), we read it
+# using execfile().
 # 'version' is the short X.Y version and 'release' is the full version,
 # including any alpha/beta/rc tags.
-from psyclone.version import version as release, short_version as version
+# We are in the doc directory but need to read version.py from
+# src/psyclone
+BASE_PATH = os.path.dirname(os.path.abspath(__file__))
+BASE_PATH = os.path.dirname(BASE_PATH)
+execfile(os.path.join(BASE_PATH, "src", "psyclone", "version.py"))
```
rupertford: We now read version information from a file. Much better. The only issue (as already privately discussed) is that execfile is not supported in Python3, but we can cross that bridge when we come to it.

setup.py
```diff
@@ -75,41 +76,15 @@
...
-                                   version=VERSION))
-    finally:
-        vfile.close()
-
+# We read the version number from version.py in the src/psyclone directory.
+# Rather than importing it (which would require that PSyclone already be
+# installed), we read it using execfile().
+BASE_PATH = os.path.dirname(os.path.abspath(__file__))
+execfile(os.path.join(BASE_PATH, "src", "psyclone", "version.py"))
+VERSION = __VERSION__
```
rupertford: Good.

src/psyclone/version.py
```diff
@@ -31,46 +31,15 @@
...
 
-BASE_PATH = os.path.dirname(os.path.abspath(__file__))
+__MAJOR__ = 1
+__MINOR__ = 5
+__MICRO__ = 0
 
+__SHORT_VERSION__ = "{0:d}.{1:d}".format(__MAJOR__, __MINOR__)
+__VERSION__ = "{0:d}.{1:d}.{2:d}".format(__MAJOR__, __MINOR__, __MICRO__)
 
-def test_vn_generation(monkeypatch, tmpdir):
```
rupertford: Removal of a test - yay!

PR #93, opened 2017-11-02
Add swap loop transform

hiker:
A bit of an early pull request for some feedback: it adds a swap-loop transform, e.g.:
```
do j=1, n
  do i=1, m
...
```
becomes
```
do i=1, m
  do j=1,n
```
Typically an unwanted transformation, but I need it for bit-reproducible results (in a 1-time only loop, so no real performance impact).

Problems: I get plenty of warnings for lines too long and trailing whitespaces - but that's all in an assignment of Fortran code to a string. Do I really need to split this into shorter strings and concatenate them, or can I just ignore the warning?
Any feedback welcome ;)


src/psyclone/transformations.py
```diff
@@ -1096,11 +1096,11 @@ def name(self):
         ''' Returns the name of this transformation as a string '''
         return "Move"
 
-    def _validate(self, node, location, position):
+    def _validate(self, node, location, position):  # pylint: disable=no-self-use
```
arporter: I wouldn't worry about these pylint warnings since putting in the disable clause then makes the line too long for pep8!

src/psyclone/tests/test_files/gocean1p0/test27_loop_swap.f90
```diff
@@ -0,0 +1,58 @@
+!-------------------------------------------------------------------------------
```
arporter: I suspect that you can cut this example down a lot - think of the bare minimum a compiler would accept (even if the compiled code would not execute). We support compiling code as part of the test suite for LFRic and ideally we'll extend this to the GOcean API. Therefore, any test code should be compilable (but not necessarily linkable).

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,196 @@ def test_module_inline_warning_no_change():
...
+    ''' Testing correct loop swapping transform. Esp. try first, middle, and
+    last invokes to make sure the inserting of the inner loop happens at
+    the right place.'''
+
+    psy, invoke = get_invoke("test27_loop_swap.f90", 0)
+    schedule = invoke.schedule
+
+    # First make sure to throw an early error if the source file
+    # single_invoke_two_kernels.f90 should have been changed
+    assert str(psy.gen) == """  MODULE psy_test27_loop_swap
```
arporter: I wouldn't bother testing against the whole MODULE here - just the body of the subroutine. And yes, we are picky about white space and long lines :-) See some of the dynamo tests for examples.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,196 @@ def test_module_inline_warning_no_change():
...
+
+    psy, invoke = get_invoke("test27_loop_swap.f90", 0)
+
+    schedule = invoke.schedule
+    swap = LoopSwapTrans()
+
+    # Test error if given node is not the outer loop of an at least
+    # double nested loop:
+    with pytest.raises(TransformationError):
+        swap.apply(schedule.children[0].children[0])
```
arporter: Please add checks that it is the correct TransformationError that has been raised by doing e.g.:

    with pytest.raises(TransformationError) as excinfo:
        swap.apply(...)
    assert "node is not a loop" in str(excinfo)

hiker:
Please don't review yet. I am using the fparser improved-error-messages branch, and it triggers failures in 4 more test cases (one is fixed in ae59de5). I think that previously errors in Fortran code were not diagnosed and went undetected, but I need to check.

hiker:
OK, nearly all done - by I might need some help/discussion:
1) One of the DAG tests fails (test_node_dag(tmpdir)) - it tests for 'red', 'blue' etc , but the dag created on my platform uses #ff0000/#00ff00/#0000ff ... as far as I can see. Any suggestions what to do? Remove the tests for colour, or test for #ff0000 or "red"  ?
2) This patch very likely needs my fparser improved-error-messages branch merged. It catches a lot of typos in your Fortran tests (see bc1273d ), but I assume that some tests will fail (see 5c123357b88efd9099ea861429fc883452867ef7), since the error messages are different now). Again, I'll leave it to you to decide how to handle this

hiker:
Yes, I am reasonable certain that the travis failures are due to using the old fparser. I don't have time to backport this (and re-port again later), so I suggest to leave this pull request till a new fparser version is ready to be used that includes my changes.

arporter:
re. the "red"/"blue" problem I'd extend the test so that it's happy with either the string or the hex value.
re. the fparser problem, that is harder. Your fixes to fparser are now on master and do reveal errors in the PSyclone test suite, as you say. Presumably you've fixed those problems here but then that causes test failures when v. 0.0.5 of fparser is used (as it is in Travis)? So, we need to make a new release of fparser and then this branch can be merged? I'll review it as it is so that we can be ready to go once we're ready to make a new release of fparser.

hiker:
All issues should be fixed now, but it needs a new fparser release. It contains all fixes from #107 (i.e. fixes for current fparser) and #108 (updates to work with next fparser result). Please hold off till a new fparser is released, and #107 and #108 are merged in.

hiker:
Should be ready to go now.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,98 @@ def test_module_inline_warning_no_change():
...
     _, _ = inline_trans.apply(kern_call, inline=False)
+
+
+def test_loop_swap_correct():
+    ''' Testing correct loop swapping transform. Esp. try first, middle, and
+    last invokes to make sure the inserting of the inner loop happens at
+    the right place.'''
+
+    psy, invoke = get_invoke("test27_loop_swap.f90", 0)
+    invoke = psy.invokes.get("invoke_loop1")
```
arporter: You assign to `invoke` twice here. If the first one isn't needed then you can use `_` instead which keeps pylint happy.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,98 @@ def test_module_inline_warning_no_change():
...
+    # test27_loop_swap.f90 should have been changed
+    expected_schedule = '''Loop[]: j= lower=2,jstop,1
+Loop[]: i= lower=2,istop,1
+kern call: bc_ssh_code'''
+    assert expected_schedule in  schedule_str
+
+    expected_schedule = '''Loop[]: j= lower=1,jstop+1,1
+Loop[]: i= lower=1,istop,1
+kern call: bc_solid_u_code'''
+    #assert expected_schedule in  schedule_str
```
arporter: Please remove commented-out assert. In fact, should this be commented out as the value just assigned to `expected_schedule` is not used before that variable is assigned a different value?

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,98 @@ def test_module_inline_warning_no_change():
...
+    swapped2, _ = swap.apply(swapped1.children[1])
+    psy.invokes.get('invoke_loop1').schedule = swapped2
+    schedule_str = str(swapped2)
+
+    expected_schedule = '''Loop[]: i= lower=1,istop,1
+Loop[]: j= lower=1,jstop+1,1
+kern call: bc_solid_u_code'''
+    assert expected_schedule in schedule_str
+
+    # Now swap the middle loops
```
arporter: Comment should say "the last loops"

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -9,14 +9,14 @@
 ''' Module containing tests of Transformations when using the
```
arporter: Please add yourself as an author of this file.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,98 @@ def test_module_inline_warning_no_change():
...
+Loop[]: j= lower=1,jstop,1
+kern call: bc_solid_v_code'''
+    assert expected_schedule in schedule_str
+    return
+
+
+def test_loop_swap_errors():
+    ''' Test loop swapping transform with incorrect parameters. '''
+
+    psy, invoke = get_invoke("test27_loop_swap.f90", 1)
```
arporter: s/psy/_/

src/psyclone/transformations.py
```diff
@@ -1361,3 +1363,83 @@ def apply(self, loop, depth=None):
...
+        from .psyGen import Loop
+        if not isinstance(node_outer, Loop):
+            raise TransformationError("Error in LoopSwap transformation. "
+                                      "Given node is not a loop.")
+
+        node_inner = node_outer.children[0]
+        # Check that the supplied Node is a Loop
+        if not isinstance(node_inner, Loop):
+            raise TransformationError("Error in LoopSwap transformation. "
+                                      "First inner statement is not a loop.")
```
arporter: Please make this error message more informative. Something like:
"... transformation. Supplied node must be the outer loop of a loop nest but the first inner statement is not a loop, got blah" where, if possible, "blah" should be something that will help the user identify the thing that is not a loop that is causing the problem.

(We try to ensure that our error messages assist the user as much as possible. It's work for us but they do appreciate it :-) )

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1184,98 @@ def test_module_inline_warning_no_change():
...
+    with pytest.raises(TransformationError) as error:
+        swap.apply(schedule.children[0].children[0])
+    assert "First inner statement is not a loop" in str(error.value)
+
+    # Not a loop:
+    with pytest.raises(TransformationError) as error:
+        swap.apply(schedule.children[0].children[0].children[0])
+    assert "Given node is not a loop" in str(error.value)
+
+    # Now create an outer loop with more than one inner statements
```
arporter: Should this say "...with more than one inner loop?" i.e. they're not perfectly nested?

src/psyclone/tests/test_files/gocean1p0/test27_loop_swap.f90
```diff
@@ -0,0 +1,27 @@
...
```
arporter: Please could you use the same copyright header as in the latest dynamo test files? Also, you are the author, not me :-)

src/psyclone/transformations.py
```diff
@@ -1133,6 +1133,7 @@ def name(self):
         ''' Returns the name of this transformation as a string '''
         return "Move"
 
+    # pylint: disable=no-self-use
```
arporter: Nice tweak for pylint - thanks!

src/psyclone/transformations.py
```diff
@@ -1361,3 +1363,83 @@ def apply(self, loop, depth=None):
...
+        >>> from parse import parse
+        >>> from psyGen import PSyFactory
+        >>> ast,invokeInfo=parse("dynamo.F90")
+        >>> psy=PSyFactory("dynamo0.1").create(invokeInfo)
+        >>> schedule=psy.invokes.get('invoke_v3_kernel_type').schedule
+        >>> schedule.view()
+        >>>
+        >>> from transformations import LoopSwapTrans
+        >>> swap=LoopSwapTrans()
+        >>> new_schedule,memento=swap.apply(schedule.children[0],
```
arporter: Please update this example to match the API of the `apply` method - i.e. it only takes a single Node as argument.

src/psyclone/transformations.py
```diff
@@ -1361,3 +1363,83 @@ def apply(self, loop, depth=None):
...
+                                             schedule.children[1])
+        >>> new_schedule.view()
+    '''
+
+    def __str__(self):
+        return "Exchange the order of two nested loops: inner becomes " + \
+               "outer and vice versa"
+    @property
+    def name(selfself):
```
arporter: Is `selfself` what you meant? (Update: I guess not because pylint doesn't like it)

src/psyclone/transformations.py
```diff
@@ -1361,3 +1363,83 @@ def apply(self, loop, depth=None):
...
+        '''Returns the name of this transformation as a string'''
+        return "LoopSwap"
+
+    def _validate(self, node_outer):  # pylint: disable=no-self-use
+        ''' validity checks for input arguments '''
```
arporter: Please document the interface to this routine.

src/psyclone/transformations.py
```diff
@@ -1361,3 +1363,83 @@ def apply(self, loop, depth=None):
...
+
+        from .psyGen import Loop
+        if not isinstance(node_outer, Loop):
+            raise TransformationError("Error in LoopSwap transformation. "
+                                      "Given node is not a loop.")
```
arporter: Please add to this message to say what the supplied node actually is to help the user identify where they're going wrong.

src/psyclone/transformations.py
```diff
@@ -1361,3 +1363,83 @@ def apply(self, loop, depth=None):
...
+        if not isinstance(node_inner, Loop):
+            raise TransformationError("Error in LoopSwap transformation. "
+                                      "First inner statement is not a loop.")
+
+        if len(node_outer.children) != 1:
+            raise TransformationError("Error in LoopSwap transformation. "
+                                      "More than one statement in outer loop.")
+
+    def apply(self, outer):  # pylint: disable=arguments-differ
+        ''' Swaps the loop by :py:obj:`inner`with its inner loop. '''
```
arporter: Please correct this docstring and add documentation for the interface to this routine.

src/psyclone/tests/gocean1p0_transformations_test.py
hiker: *grin* There is actually a reason for that: get_invoke() is broken - it should return the first invoke (index 0), but doesn't (because internally it uses keys(), but keys() does not guarantee an order). I might commit a proper fix as well (but I did use a _) :)

src/psyclone/transformations.py
hiker: Not sure what the best way is to show the data, but I did add something - feedback welcome.

src/psyclone/transformations.py
hiker: Done, but I am not sure how to properly specify rtype a tuple of a schedule and momento.

hiker:
I hope I addressed everything - ready for review.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -5,18 +5,20 @@
 # whose members are identified at https://puma.nerc.ac.uk/trac/GungHo/wiki
```
arporter: Sorry to be a pain but I've just realised that the `LoopSwapTrans` is a general transformation but we don't have any tests for it with any other API. In fact, for the dynamo0p3 API I think it is not applicable since the only time we have nested loops there is when a loop over cells is coloured. We only do that if we're parallelising with OpenMP but I think the attempt to do that would fail once the loops are swapped (because we check that any loop over colours is performed sequentially). I think therefore that we have three choices: alter `SwapLoopTransform` to check that the API=="GOcean", rename it to e.g. `GOSwapLoopTransform` or add a test for Dynamo0p3 with the proviso that it's not actually useful at the moment. My preference would be for the latter because, although it's not useful right now, it is clearly a general concept and should be treated as such.

hiker: I had a look and I can't even find a (valid) example where I could swap loops in dynamo0p3 - as you said the only nested loops are coloured OMP loops, which can't be swapped (and while I didn't check, I wouldn't be surprised that the OMP statement in the schedule will make the swap transform to abort). So I am not even sure how to create a valid/useful testcase for that. I would suggest to do the first two: check the API, and rename the transform *and those transforms really should be split into different classes: one file with the base transforms, and then one for each API ... but you know that probably ;) ).

hiker:
Everything fixed except the naming / dynamo API support. Please let me know if you are ok with my suggestion.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1184,3 +1186,116 @@ def test_module_inline_warning_no_change():
...
+                     "has 2 inner statements, the first two being .* and .*",
+                     str(error.value)) is not None
+
+    # Now remove the body of the first inner loop, and pass the first
+    # inner loop --> i.e. a loop with an empty body
+    del fused.children[0].children[0].children[0]
+
+    with pytest.raises(TransformationError) as error:
+        swap.apply(fused.children[0].children[0])
+    assert re.search("Supplied node .* must be the outer loop of a loop nest "
```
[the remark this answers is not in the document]
hiker: Re psy: that variable is used further down, so I can't remove it. Maybe there is a better way that avoids this, but I am not aware of any.

src/psyclone/tests/gocean1p0_transformations_test.py
arporter: OK, let's do the first two. (And yes, you're right about splitting the file but that's something for another issue :-) )

src/psyclone/tests/gocean1p0_transformations_test.py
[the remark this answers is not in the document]
arporter: Ah, if it's used further down (and I see that it is) then that's fine!

src/psyclone/tests/gocean1p0_transformations_test.py
hiker: Done - I couldn't find an easy way to test the API (since a transform doesn't easily have access to it), but I tested the loop type to make sure it's GOLoop, which should be sufficient.
I noticed that three gocean transforms in transformations.py are called GOceanXXX, not GOXXX - but I followed your suggestion to use GOSwap...

arporter:
Code changes are good, examples all work, tests all OK. Only thing missing is an update to the documentation - you just need to add a bit to the Transformations section of `gocean1p0.rst` and check that the resulting auto-generated text is formatted correctly (I was going to be naughty and do this for you so I had a quick go and it isn't).

arporter:
(Also, I hadn't realised that we were already inconsistent with our naming convention :-( Something for another issue.)

arporter:
Hi @hiker, It all *looks* OK but when I build the documentation, I get an error saying that it can't find the GOLoopSwapTrans:

    autodoc: failed to import class u'GOLoopSwapTrans' from module u'psyclone.transformations'

I suspect it must be my setup but I can't work out what's wrong. When you build the docs are they all OK?

arporter:
Changes are all good. There are a couple of lines that are too long but I'll fix those when I update the changelog.

arporter:
Changelog and top-level doc updated. I've also fixed those long lines in `transformations.py`.

hiker:
Hi @arporter ,

thanks for fixing the lines. It confused me a bit, since pylint also checks line length. But it turns out that it's default value is 100 characters, not 80 *sigh* So sorry for missing that.

Do you know if there is any benefit of running pep8 vs pyling with max-line-length=80? I think it would help avoiding mistakes if we would only need to run 1 tool. Hmm - https://stackoverflow.com/questions/1428872/pylint-pychecker-or-pyflakes writes: "[pep8]has found several syntax no-nos that PyLint did not. But PyLint found stuff that was not specifically linked to PEP8 but interesting.". 

Still, should I add a .pylintrc that sets the line length to 80 for pylint?

arporter:
Yes, we've found that using both tools is better as it catches more things. I hadn't realised that pylint's default line length is 100 chars so yes, a .pylintrc that fixes that would be great.

Issue #99, opened 2017-11-03
alphabeticise author list in documentation 


rupertford:
To be fair to all documentation contributors, we have decided to make the documentation author list alphabetic.

arporter:
Pull request merged. closing issue.

PR #100, opened 2017-11-03
issue #99. Modified order of documentation names to be alphabetic.

rupertford:
Names have been reordered. Documentation builds OK. Ready for review.


arporter:
Changes look fine. Documentation builds fine. Made me realise that we still have the version down as "1.5.0" even though we're now working on changes that are not in 1.5.0. I'm sure there must be official ways of handling this situation but that's not something for this issue.

Issue #103, opened 2017-11-14
Remove halo exchange when read to annexed dofs precedes write to halo l1


rupertford:
At the moment a halo exchange is added when a fields annexed dofs are read after a field has its l1 halo written and this is not required. The dynamo/eg8 test on master assumes such a halo exchange is not produced and therefore currently does not work.

rupertford:
created annexeddofshaloexchange branch

rupertford:
The associated pull request #104 has been merged to master. Closing this issue.

PR #104, opened 2017-11-21
issue #103. Remove unnecessary halo exchange for annexed dofs

rupertford:


src/psyclone/dynamo0p3.py
```diff
@@ -3934,9 +3935,34 @@ def create_halo_exchanges(self):
...
+                    # Previous dependence is not a halo exchange.
+                    #
+                    # There is still one case where we do not want to
+                    # add in a halo exchange. This is when the field
+                    # is continuous but only its annexed dofs are read
+                    # and the previous write is to the level 1 halo
+                    # (or more).
+                    if self._upper_bound_name == "ncells":
+                        # Only this fields annexed dofs are read. We also
+                        # now know that this is a continuous field.
```
arporter: Please explain (in this comment) how we know that this is a continuous field.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -6926,3 +6926,39 @@ def test_halo_req_no_read_deps(monkeypatch):  # pylint: disable=invalid-name
         _, _ = halo_exchange.required()
     assert ("Internal logic error. There should be at least one read "
             "dependence for a halo exchange" in str(excinfo.value))
+
+
+def test_no_halo_exchange_annexed_dofs():  # pylint: disable=invalid-name
+    '''If a kernel writes to a discontinuous field and also reads from a
+    continuous field then that field reads to its annexed dofs. If the
+    continuous field is written to, into its level1 halo in a previous
```
arporter: Suggest: "...then that field's annexed dofs are read. If the previous write to this continuous field included these annexed dofs then no halo exchange is required."
If I understand correctly, the level-1 halo includes more than just annexed dofs - hence the fact that looping to 'ncell' is sufficient.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -6926,3 +6926,39 @@ def test_halo_req_no_read_deps(monkeypatch):  # pylint: disable=invalid-name
...
+    result = str(psy.gen)
+    print result
+    assert "CALL f1_proxy%halo_exchange" in result
+    assert "CALL f2_proxy%halo_exchange" not in result
+
+
+def test_halo_ex_inv_ub(monkeypatch):  # pylint: disable=invalid-name
+    '''If kernel writes to a discontinuous field and also reads from a
+    continuous field then that field reads to its annexed dofs. If the
+    continuous field is written to, we then check the upper bound of
```
arporter: suggest: "...then that read will include the annexed dofs of that field."

src/psyclone/tests/test_files/dynamo0p3/14.7.1_halo_annexed.f90
```diff
@@ -0,0 +1,27 @@
+program single_invoke_annexed
```
arporter: This file is missing the copyright/license header.

src/psyclone/dynamo0p3.py
```diff
@@ -3934,9 +3935,34 @@ def create_halo_exchanges(self):
...
                 if not isinstance(prev_node, DynHaloExchange):
-                    # previous dependence is not a halo exchange so
-                    # create a new halo exchange
-                    self._add_halo_exchange(halo_field)
+                    # Previous dependence is not a halo exchange.
+                    #
+                    # There is still one case where we do not want to
+                    # add in a halo exchange. This is when the field
+                    # is continuous but only its annexed dofs are read
+                    # and the previous write is to the level 1 halo
```
arporter: I think it is sufficient that the previous write be to owned cells since that includes annexed dofs. The level-1 halo is beyond that.

src/psyclone/dynamo0p3.py
arporter: oh, and s/field/field's/

src/psyclone/dynamo0p3.py
```diff
@@ -3934,9 +3935,34 @@ def create_halo_exchanges(self):
...
+                        if upper_bound in ["ncells", "cell_halo"]:
+                            # we do not require a halo exchange as the
+                            # writer ensures annexed dofs are clean
+                            pass
+                        elif upper_bound == "ndofs":
+                            # we require a halo exchange as the writer
+                            # does not write annexed dofs
+                            self._add_halo_exchange(halo_field)
+                        else:
+                            raise GenerationError(
```
arporter: Please update the docstring for this routine to include the fact that it can raise GenerationError.
I'm concerned that this code won't work (well, will trigger the GenerationError) if the previous loop has been transformed for OpenMP (which the MO do as a matter of course?). Please could you add a test for this case.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -6926,3 +6926,39 @@ def test_halo_req_no_read_deps(monkeypatch):  # pylint: disable=invalid-name
...
+
+def test_no_halo_exchange_annexed_dofs():  # pylint: disable=invalid-name
+    '''If a kernel writes to a discontinuous field and also reads from a
+    continuous field then that field reads to its annexed dofs. If the
+    continuous field is written to, into its level1 halo in a previous
+    loop then no halo exchange is required'''
+    _, invoke_info = parse(os.path.join(BASE_PATH, "14.7.1_halo_annexed.f90"),
+                           api="dynamo0.3")
+    psy = PSyFactory("dynamo0.3").create(invoke_info)
+    schedule = psy.invokes.invoke_list[0].schedule
```
arporter: Please extend this test to include (optional) compilation of the generated PSy layer.

src/psyclone/dynamo0p3.py
rupertford: I believe what I am doing is correct here and the comment is misleading. I'll update it. Anyway, below is the logic. Obviously let me know if I've made a mistake.

Annexed dofs only exist for continuous fields. Continuous fields can be modified in two ways 1: incremented when the parent loop iterates over cells 2: updated when the parent loop iterates over dofs. In case 1: we always (at least) iterate to the l1 halo as this ensures that annexed dofs are correct for any subsequent read. In case 2: we can currently either a) write to owned cells, which does not include annexed dofs, or b) write to a particular halo depth (or the full halo depth) via the redundant computation transformation. So, what you say would be true if we supported iterating over dofs up to annexed dofs (and perhaps will be true at some point), but is not at the moment.

rupertford: Looking at the code again, I think it would make more sense for it to be moved to the required() method for a halo exchange so I will do that before next review.

src/psyclone/dynamo0p3.py
rupertford: This code has now been removed.

src/psyclone/dynamo0p3.py
rupertford: This code has now been removed

src/psyclone/tests/dynamo0p3_test.py
rupertford: Quite right. Done.

src/psyclone/tests/test_files/dynamo0p3/14.7.1_halo_annexed.f90
rupertford: Added.

src/psyclone/tests/dynamo0p3_test.py
rupertford: This test no longer exists.

src/psyclone/tests/dynamo0p3_test.py
rupertford: I've modified the text. As discussed earlier it is not possible to just compute owned and annexed dofs, you can either compute owned, or owned +l1 halo. One reason for this is to ensure that annexed dofs are always correct by redundantly computing into the l1 halo by default.

rupertford:
OK, I think it's ready for second review. I've moved the logic so it is tested in the halo exchange required() method. This feels much nicer as the annexed dofs information now appears in the right places. Sorry for the modification. Back to @arporter.

arporter:
Code now makes much more sense to me.
All changes are covered.
Tests (including compilation) all pass/xfail.
All examples run OK.
Documentation does not require updating.
Code changes are pep8 and pylint clean.

PR #108, opened 2017-11-24
Updated code to handle new AnalyzeError raised by the

hiker:
improved-error-message branch of fparser.

Not to be merged before a new release of fparser is done that includes the above branch (till then some tests will fail).


hiker:
I properly fixed some issues in my latest update, but this will need a new fparser release before it can be merged.

hiker:
Merged into #107 since fparser was updated by now, so am closing this,

Issue #91, opened 2017-10-13
Remove unecessary __init__.py files


arporter:
It turns out that I was over zealous in adding `__init__.py` files when I re-structured PSyclone. In fact there should only be such files in directories that contain packages. That means that we should have just one such file in `src/psyclone`.

rupertford:
pr #111 is now on master. Closing this issue.

PR #111, opened 2017-12-04
Remove superfluous __int_.py files

arporter:
This PR is for work done under #91. It is a small change that just removes those `__init__.py` files that are not required. In theory this would leave us one such file in `src/psyclone` as that is where all our packages are. However, the test suite also imports a deliberately-broken package from `src/psyclone/tests/test_files` so I've had to leave one in there too. I'm not sure if there's a better way of doing this but I figure this is an improvement over what is currently on master.


arporter:
Note that some of the tests on travis will fail because, by default, the latest version of fparser gets installed. When I 'fixed' this (by specifying version 0.0.5 in `setup.py`) all tests passed.

Issue #114, opened 2017-12-08
Add interface to CLAW


arporter:
We need to be able to apply CLAW transformations to user-supplied kernels. In this Issue we will start to add the necessary wrappers and documentation.

arporter:
In contrast to PSyclone's native transformations which operate on the
PSy AST, CLAW operates on Fortran source code. In this case this will
be one or more of the kernels called from the PSy layer.
PSyclone already has the location of these kernels since the meta-data
they contain must be parsed. This location information must now also be passed
to the transformation script in case it wishes to modify the kernels.

Naming of the transformed kernel - just name mangle and output it in the
same location as the generated psy code? However, LFRic has naming
rules which means the name of the kernel is intimately linked with the
name of the file which contains it. This means that if we name mangle
then we have to do this consistently through the kernel file (i.e.
the module, type declaration and kernel subroutine will all have to be
renamed). If we don't change the name of the kernel then making sure
we get the right kernel at compile time becomes the problem of the build system so
long as we don't over-write the original kernels. The latter problem
can be avoided by allowing the output directory to be specified as an
argument to PSyclone. This will then have to be passed down to the 
`trans()` function as an additional argument. Therefore, the proposed interface to `trans()` is now:

    def trans(psy, kernel_search_path, out_dir):
        '''
        :param psy: root of PSy AST
        :param kernel_search_path: List of directories to search for kernels
        :param out_dir: directory in which to put any transformed kernels
        '''

rupertford:
There is another practical issue here. If PSyclone is run separately for two different algorithm files which each make use of the same kernel then what do we do? I guess we could probably just say it is a build problem. However to help the build system we might want to have various configuration options for the case where an existing output file is found I.e. 1: overwrite, 2: keep original 3: raise an exception.

There might be a use case where the hpc expert wants to create a modified version of the kernel but also keep the original (or some other variant). In this case I'm not sure it can be done though different directories due to the module names clashing, so we might need to be able to create a unique file and module name that is also used by PSyclone.

rupertford:
OK, I'm going to argue against the proposed 2nd argument to trans() and suggest it should be a dictionary of Kernel AST's, indexed by module name. We then need to decide if each should include its original file name, or whether we create the file name from the module name.

[...]

PR #120, opened 2018-01-02
Implement Claw interface

arporter:
This PR is for work done under #114.


arporter:
While starting work on introducing OpenACC directives with CLAW, I discovered test failures in this branch. I don't understand why. Maybe I've updated the version of various things on my laptop. The most problematic failure was to do with losing the stderr from where I 'shell out' to run CLAW. I've fixed that but accidentally committed it to the wrong branch so I've used `git cherry-pick` to apply it to the branch for this PR.

arporter:
I made the mistake of working on (and so breaking) my branch of CLAW that this PR depends upon. I've now cut a new `psyclone` branch in stfc/claw-compiler which should be used with this PR.

doc/claw.rst
```diff
@@ -0,0 +1,212 @@
...
+
+The version of the CLAW compiler used by PSyclone is available from
+https://github.com/stfc/claw-compiler.  See the INSTALL.md file for
+installation instructions.  Once it is installed, you will need to
+manually copy the Python interface module into the `lib` directory of
+the Claw installation:
+
+::
+
+    cp claw-compiler/omni-cx2x/src/claw/python/ClawTransform.py <claw-install-root>/lib
```
rupertford: I could not find a file called ClawTransform.py or even a directory called omni-cx2x.

arporter: My apologies - this goes back to the fact that Valentin has re-structured CLAW and I got carried-away and pulled his changes into the stfc version of CLAW. The file you need is now:

    cx2t/src/claw/python/ClawTransform.py

(On the `psyclone` branch of stfc/claw-compiler)

rupertford:
I have a potentially disruptive suggestion. In your implementation the PSyclone script passes a list of kernels and a script file. What I propose is for CLAW transformations to be called in the same way PSyclone transformations are called. So, for example

```
def trans(psy):
    import claw
    ...
    kernel = psy......
    ...
    claw_openacc_trans = claw.trans.openacckern()
    new_kernel = claw_openacc_trans.apply(kernel)
    ...
```

Do you think this is a) feasible and, if so b) worth doing now in this pr?

arporter:
I think it is feasible and probably even sensible for specific cases such as the one your example suggests - that of adding the necessary OpenACC `routine` directive to a kernel. However, I'm concerned that in general, this would be equivalent to requiring PSyclone to support a set of transformations that operate on a whole PSy layer, rather than letting the user build-up their recipe from fundamental transformations (such as loop fuse). For specific, simple cases it could be powerful but I think, in general, allowing users to write their own script is better. e.g. if we think of transforming the matrix-vector kernel in the various ways that Karthee/Sergi have done, then that would need a script (I think) rather than a one-size fits all transformation.

rupertford:
Ah, I see your point. The PSyclone script, as it stands, does not know about the innards of a Kernel.

Given that you are writing a script that does know about the innards of a Kernel (a "CLAW" script, rather than a "PSyclone" script) and given that both are just Python, is there any particular reason for formally separating the two i.e. is there any reason why the contents of a "CLAW" script could not be in a "PSyclone" script?

arporter:
Nope, there's no reason and in fact I have written the examples so that both the PSyclone and CLAW scripts are routines in the same file (we have a `trans()` and a `claw_trans()`). I thought it best to separate the two routines as one is run from CLAW and only operates on a CLAW AST whereas the original works on our PSyclone schedule.

doc/getting_going.rst
```diff
@@ -60,9 +60,13 @@ example in openSUSE 42.2:
 
 and then use pip2.7 instead of pip.
 
+PSyclone supports the use of the CLAW compiler
+(https://github.com/C2SM-RCM/claw-compiler) to perform kernel
```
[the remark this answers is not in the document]
arporter: Done.

rupertford:
Andy and I have discussed the limitations of the current approach to integrating CLAW into PSyclone i.e. the way it has been implemented in this PR. The fact that we need to use jython, which only works for Python2, that we have to separate out scripts working on CLAW and scripts working on the PSyclone IR and that it is difficult to make use of existing CLAW functionality by only driving CLAW via a script are quite strong limitations.

We feel that a potentially better approach would be for PSyclone to drive CLAW by adding directives to Kernel code, then running CLAW. The Python transformation interface could remain, it would just be interfacing with CLAW in a different way. The advantages of this approach is that we don't need to modify CLAW with a Python scripting interface, we don't need to use Jython, we can mix PSyclone and CLAW transformations in whatever way we like and we can potentially make use of existing CLAW functionality e.g. optimizing for a GPU as a black-box high-level optimization. We can also have access to the OMNI XML Fortran representation allowing us to modify code outside of CLAW if required, e.g. when inlining.

If @arporter is happy I'll therefore close this pull request and add the above information to the original issue.

rupertford:
Closing this pull request as we expect to use a different approach

PR #122, opened 2018-01-10
Add pylintrc file to fix #118

### LINKED ISSUE #118: Adding pylintrc to tests to disable common warnings
There are around 240 'disable invalid names' directives in the psyclone test directory, 30% of the files are over 1000 lines (which triggers a warning). We seem to have:
- 259 locally disabled messages (mostly invalid names)
- 92 invalid-name errors
- 79 duplicated code errors

Are you ok if I add a pylintrc files that disables three tests (invalid name, duplicated code, file length) by default? I'd say that they are not useful for our tests.

hiker:
Since I had no one disagreeing with #118, here the pull request. Before review I'd suggest to first look at my changed default for .pylintrc in https://github.com/hiker/PSyclone/commit/a252cda83e69b4d4501435f6092de9d159d74739#diff-491f94e68f2ed0d28f2abc2c098399b4 and check the defaults I've picked and let me know if you are ok with my choices.


arporter:
Hi @hiker, I did mean to respond to #118 but never got there. I like the idea of disabling some warnings but I don't think invalid-name and duplicate-code should be included. I'd rather trade the slight inconvenience of adding the explicit disables for the improved readability of the tests. After all, there is a lot of code in the tests and that carries a maintenance burden, even though it's not the actual PSyclone code.

We need to re-structure the directories containing the tests (in fact, I realise I need to create an Issue for that). Once that's done a lot of our very-long test names can be shortened as we can push a lot of the information they contain into the directory structure and file names.

I'm not comfortable with the very long list of things that we're disabling warnings for in pylintrc. It seems to me we should start small and then add to them if we all agree. e.g. In the current list I see things like `import-star-module-level` and `old-raise-syntax` which are things I would want to be warned about.

hiker:
1. The long list: this is mostly pylint's default. The link above shows the diff, i.e. which (non-default) options I added, which are only: ```invalid-name,duplicate-code,no-self-use,too-many-locals,too-many-lines,protected-access,locally-disabled,locally-enabled.```. The two options you mention (import-star-module-level and old-raise-syntax ) are disabled - but with a .pylintrc file it would be easy to enable them, and I am happy to add this to this pull request :)
2. Re readability: adding #pylint disable= sometimes followed by the enable statement does not improve readability at all, it just means one can see less of the actual code on the screen. 259 locally-disabled messages show imho clearly that the defaults are just not correct for the kind of code in that directory. Atm we are nearly forcing everyone to write two or three lines for each function just because the descriptive name is too long (not sure if we need a pylint enable or only disable) . If in the future we split the test better (which I am all in favour of) and can remove those pylint warnings, it is again less work to just remove the default instead having to modify (by then probably) 500 directives :)
3. Re duplicated code: I had a quick look - a lot of duplicated code is Fortran code (and even comments in Fortran code), and it sometimes changes minor things like int vs real or so. I don't think refactoring this helps with readability or maintainability.

OK, I won't start an endless discussion, so: imho tests are a necessity, but they should not burden the developers too much or costs too much time. I would pretty much prefer all of us to spend time on writing interesting transforms and other features than refactoring test cases. At this stage test cases fulfill their task.

Besides, this pull requests does not prevent any of this, it is just ... like an 80% improvement now on pylint issues, and if someone has time to do the remaining 20%, just change the psyclone defaults back to what they were before ;)

So, did I change any minds or not? If you want  invalid-name and duplicate-code enabled, just let me know and I'll see what remains of this pull request ;)

TeranIvy:
I would agree with @hiker that disabling/enabling invalid-name warnings in .pylintrc file instead of locally in tests is a better solution. When the tests are restructured enabling invalid-name check in .pylintrc file would be a good way to check simultaneously for all of them whether the restructuring is done properly instead of removing statements one by one.

hiker:
I've enabled all pylint tests that are disabled by default except for print (there are over 350 print statements in the tests, since I am not sure how useful these print statements are, I'll leave it for someone else to decide), and suppressed-messages (since we suppress messages on purpose ;) ). I then fixed all warnings, in a few cases by disabling/enabling the corresponding test if I thought that refactoring does not make any sense, or it's a clear false positive.

Review please :)

hiker:
I just realised that I missed the point of making the .pylintrc minimal. I just fixed that, so now ready for review.

src/psyclone/tests/expression_test.py
```diff
@@ -10,14 +11,18 @@ def my_test(name, parser, test_string, names=None):
     preserved by the parsing operation, so the test_string must conform to
     the whitespace conventions of the unparser for the test to succeed.'''
     # These imports are required in order for the exec in the code below
-    # to work
-    from psyclone.expression import Grouping, BinaryOperator, FunctionVar, \
-        Slicing, LiteralArray, NamedArg
+    # to work. Pylint complains about those as unused variables
+    # pylint: disable=unused-variable
```
arporter: Nice tweak to warnings - thanks.

Issue #145, opened 2018-03-06
the big picture


rupertford:
It would be useful to capture existing and future developments in PSyclone so people can see what's been done and what could be looked at. A concept graph or mindmap view would probably make most sense.

rupertford:
Created bigpicture branch

rupertford:
Created and added a mindmap outlining the big picture as I see it.
The mindmap is created in Linux using vym (view your mind). I chose this as it is a freely available Linux tool.

rupertford:
The Met Office use freemind version 0.9.0 (as kindly confirmed by @TeranIvy). This has advantages over vym, which is what we are currently using. Therefore we will migrate. This requires a file format change. There is a conversion tool https://sourceforge.net/projects/vym2fm/ and a separate xslt script here http://freemind.sourceforge.net/wiki/index.php/Import_and_export. I will try to convert automatically.

arporter:
Closing this issue has the branch was merged a while ago.

PR #146, opened 2018-03-06
issue #145. Add initial mindmap

rupertford:
Added an initial "big picture" mindmap which is a first attempt to list what we've done, what we're working on and what could be done in the future. I've used a mindmap and created in using vym which is a free Linux tool. This document can and should be updated by others but only under revision control.


arporter:
It's a pretty complete picture so I'm merging to master. We should really have a readme or similar to explain what the .vym file is but we can live without that for now.

PR #153, opened 2018-03-23
Added --version (-v) flag, and added PSyclone version to help message.

hiker:
I recently had the problem with Justin that we couldn't easily identify which psyclone version was installed - we had to look at the installation directory to see this (afaik ... unless I missed something).

So this pull requests add a -v option (which admittedly is limited useful, since ```psyclone -v``` ALONE does not work, the argument parsing requires a filename to be specified, so you need to do ```psyclone -v somefile```). But the version number is now also listed in the -h output.

I also fixed some pylint messages.

### DIFF
--- a/src/psyclone/generator.py
+++ b/src/psyclone/generator.py
@@ -7,6 +7,8 @@
 # whose members are identified at https://puma.nerc.ac.uk/trac/GungHo/wiki
 # -----------------------------------------------------------------------------
 # Author R. Ford STFC Daresbury Lab
+# Modified work Copyright (c) 2018 by J. Henrichs, Bureau of Meteorology
+
 
 '''
     This module provides the PSyclone 'main' routine which is intended
@@ -36,6 +38,12 @@ def handle_script(script_name, psy):
     :type script_name: string
     :param psy: The psy layer to which the script is applied.
     :type psy: :py:class:`psyclone.psyGen.PSy`
+    :raises IOError: If the file is not found.
+    :raises GenerationError: if the file does not have .py extension
+        or can not be imported.
+    :raises GenerationError: if trans() can not be called.
+    :raises GenerationError: if any exception is raised when trans()
+        was called.
     '''
     sys_path_appended = False
     try:
--- a/src/psyclone/generator.py
+++ b/src/psyclone/generator.py
@@ -29,9 +29,76 @@ from psyclone.line_length import FortLineLength
 from psyclone.version import __VERSION__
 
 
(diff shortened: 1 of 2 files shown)


hiker:
Oops - I missed that there were already tests for generate - my apologies.

But I totally disagree that adding the 4 disable statements are the way to go ... since this is what happened before this pull request: I added those statements. Then noticed that they are actually correct, that code needs refactoring ... so I refactored it. Then I realised that I would need test cases in order to get this patch in (as I've said, I missed the existing test cases - I did grep, must have been a typo :( ). So I thought that it's not worth to go through all that effort for one command line option, and reverted my changes, but did not add the disable (since the code should be improved, so ignoring this warning is wrong) :)

OK, so long story short: new patch, with proper test cases, and refactored generate() function (which only needed one pylint disable now).

src/psyclone/generator.py
```diff
@@ -21,16 +21,84 @@
...
+from psyclone.version import __VERSION__
+
+
+def handle_script(script_name, psy):
+    '''Loads and applies the specified script to the given psy layer.
+    The 'trans' function of the script is called with psy as parameter.
+    :param script_name: Name of the script to load.
+    :type script_name: string
+    :param psy: The psy layer to which the script is applied.
+    :type psy: :py:class:`psyclone.psyGen.PSy`
```
rupertford: Could you please add the exceptions as part of the function signature.

src/psyclone/generator.py
```diff
@@ -21,16 +21,84 @@
 import sys
```
rupertford: Please add your name and copyright information to the top of the file. Note, the license information is actually incorrect but we can change that separately.

src/psyclone/tests/generator_test.py
```diff
@@ -102,6 +102,7 @@ def test_similar_kernel_name():
```
rupertford: Please add your name and copyright info to the top of the file. Again the license information needs changing but that can be done separately.

src/psyclone/generator.py
hiker: I've added the exception details about the ones explicitly raised in the subroutine. But it also has a generic catch-all exception handling, re-raising the original exception - and I have no idea how to document this (and didn't see any example either).

hiker:
All done (I hope). What is actually the best way to check that the parameter documentation is correct. Some functions are documented in the PSyclone manual, but e.g. handle_script is not included - what do I need to do to check my formatting works as expected?

Issue #121, opened 2018-01-02
Installation and configuration files


arporter:
Up until now there has only been one configuration file associated with PSyclone (`config.py`) and it is not something that the average user ever has to edit. As such it is just installed as part of the PSyclone package when using setuptools.

However, with the work to interface Claw and PSyclone (#114) it is now necessary to have another configuration file with details of the Claw/Omni installation. Unlike `config.py` this *will* have to be edited as part of the installation procedure. We therefore need to decide where configuration files should be installed to and then edit `setup.py` in order to achieve this. (See e.g. https://docs.python.org/2/distutils/setupscript.html#installing-additional-files for details of how this should be done.)

arporter:
Since we treat `config.py` as part of the PSyclone package, we currently just `import` it whenever we need to. However, this approach won't work if we have configuration files that are not on our PYTHONPATH. We may need extra code to load a configuration file and make the resulting data available.

arporter:
As @MatthewHambley is not yet a collaborator, I shall bring him into this discussion like so...

MatthewHambley:
I'm not sure how Python normally handles these matters but I do know the expectations on a Posix system.

Normally there would be a hidden configuration file in the users home directory. Maybe a `~/.psyclone` file. If we expect multiple configuration files it could be a directory.

There might also be a site-wide "default" configuration in either `/etc/psyclone` or`'/usr/local/etc/psyclone`.

The contents of these files is largely down to the owning application. It is commonly something human readable though.

Obviously different expectations exist for Windows systems and probably MacOS and Linux systems too despite being somewhat Posixy.

hiker:
There should also be a project specific file (e.g. $PROJECTDIR/.psyclone).

MatthewHambley:
> There should also be a project specific file (e.g. $PROJECTDIR/.psyclone).

That's a good idea but what is `$PROJECTDIR` in this case. Do you mean "current directory"? Effectively `./.psyclone`?

arporter:
[...]

PR #166, opened 2018-04-18
For #121 - add a proper config file

arporter:


arporter:
I've just realised that my much-vaunted use of virtualenv breaks the way I search for installed configuration files :-( In fact, I don't know where the install process will put the configuration file if I'm running pip inside a virtual environment.

arporter:
Actually, when I do pip uninstall I see that it was put in:

    /home/memyself/MyInstalls/python_envs/py2/lib/python2.7/site-packages/home/memyself/.psyclone
    /psyclone.cfg

which isn't exactly ideal.

TeranIvy:
My experience so far with `pip install` in virtual environments is that it installs fparser and PSyclone (providing I'm running Python 2.7 on RedHat Linux) in the `lib/python2.7/site-packages/` subdirectory of the virtualenv installation directory. `which python` returns the Python (virtualenv if used) installation directory path (let's call it `$VIRTENV_DIR`). If I want to install packages elsewhere I need to run e.g.

`pip install --install-option="--prefix=$PSYCLONE_INSTALLDIR" fparser`

or optionally with a release number 

`pip install --install-option="--prefix=$PSYCLONE_INSTALLDIR" fparser==0.0.7`

but before that I need to add the path to any required packages to `PYTHONPATH`, e.g.

`export PYTHONPATH=$VIRTENV_DIR/lib/python2.7/site-packages:$PYTHONPATH`

otherwise it will try to install required packages as well.

`pip install` with Anaconda Python on Windows works in similar way, with some tweaks to paths due to different structure of subdirectories.

Shall I try this branch on MO machines to compare results?

arporter:
Thanks for the info @TeranIvy but don't worry about testing this branch just yet. I'll do some more work and then it would be really helpful if you could test it with your various MO environments.

arporter:
Travis test failures are because pip is pulling in fparser 0.0.7 and this branch only works with <=0.0.6.
However, it all works for me now - if we detect that we are running in a virtual environment (virtualenv or Py3's venv) then the configuration file is installed to the base directory of that environment. This ensures that different virtual environments remain isolated from one another. In a virtual environment, the order in which we search for the config file is now:

1. ${PWD}/.psyclone/
2. <base-dir-of-virtual-env>/
3. ${HOME}/.psyclone/
4. /etc/

@TeranIvy - would you be kind enough to try this branch on the environments that you use and check that the tests (config_test.py in particular) run OK after installation?

TeranIvy:
Hi @arporter! I'm sorry for the delay in testing the branch. Since we keep PSyclone and fparser separate from the rest of Python environments, I usually install them by getting archives, unpacking them and using
`pip install --install-option="--prefix=$PSYCLONE_INSTALLDIR" .`
or 
`python setup.py install --prefix=$PSYCLONE_INSTALLDIR`
if pip plays up.
from the unpacked directory.

In this case I made an archive of your branch and testes installations on Spice and Crays (XC40/XCS/MONSooN). I left the options in the added `psyclone.cfg` as they are. There is a minor annoyance of having to get `configparser` tarball to send it to Crays (not on Spice as `pip install` gets dependencies automatically) and installing it before PSyclone. However, I can just add the relevant line to build scripts.

To conclude: everything works OK :-)

arporter:
Thanks very much for testing this @TeranIvy. So, to clarify, when you install PSyclone you do so as some other user but not one that has write access to `/etc`?

TeranIvy:
Yes, @arporter, that's correct. I have two options of installing PSyclone:
a) My `.local` directory as a user,
b) Designated directories for the LFRic build systems on Spice/Crays where I have admin privileges.

Neither of those includes modifying `/etc`, as they are used for MO-wide configurations and managed by the relevant MO support. I do not think they will allow me to make modifications there for a package that only we use.

arporter:
After a fair bit of faffing I've decided to just install the config file to `<python-base>/share/psyclone/psyclone.cfg`. This simplifies the installation and thus is more robust. However, if the user specifies the `--prefix=` option when doing the install, then PSyclone won't be able to find the configuration file. In this case it will then be up to the user to either a) copy the config file to one of the locations searched by PSyclone or b) supply the location of the config file when invoking PSyclone. I haven't implemented support for option b) yet but I will...

TeranIvy:
Both `--prefix` and `--user` options can create a `share/` subdirectory in the installation directory for any package data files required. Is this what you meant by option b)?

arporter:
@TeranIvy - I think so, yes. The installation will itself now create `share/psyclone/psyclone.cfg`, even if the directory doesn't initially exist. I've added support for a `PSYCLONE_CONFIG` environment variable - if this is set and it points to an existing file then that is used as the configuration file.

I'm currently getting test failures, I think related to the way I manage whether DM is enabled. More investigation required...

TeranIvy:
Hi @arporter, I am looking into pip installation through a requirement file where the installation prefix can be specified.  I thought of two options of using the information from a requirement file:

- Create the `psyclone.pth` file which would contain the path to PSyclone installation base directory and place it into `lib/pythonx.x/site-packages` together with the PSyclone package itself,
- `configuration.py` could just parse the edited requirements file to find out the installation director path.

In this way specifying an environment variable for the configuration file could be avoided. What do you think about this?

arporter:
I'm not sure that using a .pth file will help - we will still ultimately need to know where the installed configuration file ended up. The 'proper' solution to this seems to be to specify `package_data` in the call to `setup()` and then use the resource-management API of `pkg_resources` to find it. However, this seems complicated (I can't get it to work) and still leaves us with the problem of where to put a 'system-wide' configuration file. For simplicity, I'm therefore going to stick with the environment variable approach.

MatthewHambley:
I'm a little late to the party but would `<python root>/etc` not make a good place to put site configuration, as opposed to user configuration?

TeranIvy:
`<python root>/etc` could work for users who a) modify their Python installation, b) do not use `--prefix` to install PSyclone somewhere else. Neither of these work for us on all platforms we use.

arporter:
Arrgh! I now get test failures and, on closer inspection, the shebangs in the scripts that are installed by pip (`psyclone` and `genkernelstub`) end up with the wrong python - they have the system python instead of that associated with the virtualenv in which I've installed PSyclone.

@MatthewHambley - it seems that `share` is the more common pythonic location so I propose to stick with that.

arporter:
However, if I install in a virtualenv on my desktop (instead of my laptop) then the shebang in the installed scripts is correct. Desktop is Fedora Core, laptop is Linux Mint.

arporter:
Steve has pointed us to this which nicely sums-up the situation:
https://imgs.xkcd.com/comics/python_environment.png

arporter:
I think all new lines are covered.
All modified routines have docstrings.
All examples run OK.
Documentation is updated.
Actual functionality of PSyclone is unchanged so no need to test with LFRic.
Ready for first review...

rupertford:
I'll hold off performing a second review until annexed dof computation #165 is on master (and this branch has been updated) as #165 adds a new config option which would need to be reflected here.

config/psyclone.cfg
```diff
@@ -0,0 +1,47 @@
+# -----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
arporter: I think you're safe - #165 is pretty much done :-)

doc/configuration.rst
```diff
@@ -0,0 +1,104 @@
...
+formatted accordingly. It currently consists of a required, DEFAULT
+section e.g.:
+::
+
+    [DEFAULT]
+    SUPPORTEDAPIS = gunghoproto, dynamo0.1, dynamo0.3, gocean0.1, gocean1.0
+    DEFAULTAPI = dynamo0.3
+    SUPPORTEDSTUBAPIS = dynamo0.3
+    DEFAULTSTUBAPI = dynamo0.3
+    DISTRIBUTED_MEMORY = true
```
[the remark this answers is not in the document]
arporter: Done.

doc/generator_script.rst
```diff
@@ -91,8 +91,8 @@ API. The default API, along with the supported API's can be seen by
 running the generator.py script with the -h option.
 
 If you use a particular API frequently and it is not the default then
-you can change the default by editing the config.py file in the
-<PSYCLONEHOME>/src directory.
+you can change the default by editing the ``psyclone.cfg`` file.
```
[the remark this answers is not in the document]
arporter: Updated.

src/psyclone/configuration.py
```diff
@@ -0,0 +1,343 @@
...
+
+class ConfigFactory(object):
+    '''
+    Creates a singleton instance of our Config object.
+    '''
+    _instance = None  # Our single Config object
+
+    def __init__(self, config_file=None):
+        '''
+        Create our singleton Config object. If config_name is specified
```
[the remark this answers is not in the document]
arporter: Oh yes.

src/psyclone/gen_kernel_stub.py
```diff
@@ -47,9 +47,12 @@
 from psyclone.dynamo0p3 import DynKern, DynKernMetadata
 from psyclone.psyGen import GenerationError
 from psyclone.parse import ParseError
-from psyclone.config import SUPPORTEDSTUBAPIS, DEFAULTSTUBAPI
+from psyclone.configuration import ConfigFactory
```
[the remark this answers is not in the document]
arporter: No kidding :-) but done. I figure it's pretty good if you're complaining about things like that ;-)

src/psyclone/tests/config_test.py
```diff
@@ -0,0 +1,343 @@
...
+        # Test when (we appear to be) both inside and outside a virtual
+        # environment
+        for inside_venv in [True, False]:
+            monkeypatch.setattr(
+                "psyclone.virtual_utils.within_virtual_env",
+                lambda: inside_venv)  # pylint: disable=cell-var-from-loop
+            with pytest.raises(ConfigurationError) as err:
+                _ = Config.find_file()
+            err_msg = str(err)
+            print err_msg
```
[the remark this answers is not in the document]
arporter: Done.

arporter:
I've done the rebase and am in the process of sorting out the problems. First off, the new 'compute_annexed_dofs' option is really dynamo0.3-specific and deserves its own section in the config file. I was going to sub-class `Config` and add the dynamo0.3-specific options that way. However, that would require us knowing which API we are using when we create the `Config` object and we don't know that because we get the default API from the configuration file. I think therefore that I'll just have to add methods to the `Config` class that deal with parsing the API-specific sections.

arporter:
All tests now pass in py2 and py3. I need to address some code coverage issues and then tackle the documentation.

arporter:
I think we're now ready for 2nd review...

src/psyclone/configuration.py
```diff
@@ -0,0 +1,393 @@
...
+    @staticmethod
+    def create():
+        '''
+        :returns: the singleton Config instance
+        :rtype: :py:class:`psyclone.config.Config`
+        '''
+        return ConfigFactory._instance
+
+
+class Config(object):
```
rupertford: I have a slight concern that this class provides information for all API's irrespective of which API we have chosen. I understand there is a chicken and egg issue here but could the specific sections have their own classes, or would that make things worse? I am happy to stick with what we have at the moment as there are not many parameters but might there be problems in the future e.g. if two different api-specific sections have different defaults for the same thing?

src/psyclone/dynamo0p3.py
```diff
@@ -4095,8 +4110,16 @@ def upper_bound_halo_depth(self):
         return self._upper_bound_halo_depth
 
     def _lower_bound_fortran(self):
-        ''' Create the associated fortran code for the type of lower bound '''
-        if not config.DISTRIBUTED_MEMORY and self._lower_bound_name != "start":
+        '''
+        Create the associated fortran code for the type of lower bound.
```
rupertford: consistency in case for '[Ff]ortan' please

src/psyclone/dynamo0p3.py
```diff
@@ -4140,11 +4163,11 @@ def _upper_bound_fortran(self):
...
 
-        # We only require a mesh object if distributed memory is enabled
-        # and the loop is over cells
-        if config.DISTRIBUTED_MEMORY and \
+        # We only require a mesh object to get upper loop bounds if
+        # distributed memory is enabled and the loop is over cells
+        if _CONFIG.distributed_memory and \
            self._upper_bound_name in ["ncells", "cell_halo"]:
-            if self._kern.is_intergrid:
+            if self._kern and self._kern.is_intergrid:
```
rupertford: Could you please explain the reason for adding self._kern here

src/psyclone/dynamo0p3.py
```diff
@@ -4449,10 +4472,10 @@ def gen_code(self, parent):
         depending on the loop type and then call the base class to
         generate the code.
 
-        :param parent: an f2pygen object that will be the parent of
+        :param parent: an f2pygen object that will be the parent of \
         f2pygen objects created in this method
         :type parent: :py:class:`psyclone.f2pygen.BaseGen`
-        :raises GenerationError: if a loop over colours is within an
+        :raises GenerationError: if a loop over colours is within an \
```
rupertford: OK, I have to ask. What's with the '\' - are new versions of pycodestyle or pylint complaining?

src/psyclone/gen_kernel_stub.py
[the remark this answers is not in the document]
rupertford: Agreed, thanks.

src/psyclone/generator.py
```diff
@@ -27,10 +53,14 @@
 from psyclone.parse import parse, ParseError
 from psyclone.psyGen import PSyFactory, GenerationError
 from psyclone.algGen import NoInvokesError
-from psyclone.config import SUPPORTEDAPIS, DEFAULTAPI, DISTRIBUTED_MEMORY
 from psyclone.line_length import FortLineLength
 from psyclone.profiler import Profiler
 from psyclone.version import __VERSION__
+from psyclone.configuration import ConfigFactory
+
+
```
rupertford: Do we need two blank lines here?

src/psyclone/parse.py
```diff
@@ -1,7 +1,8 @@
 # -----------------------------------------------------------------------------
 # BSD 3-Clause License
 #
-# Modifications copyright (c) 2017, Science and Technology Facilities Council
+# Modifications copyright (c) 2017-2018, Science and Technology
```
rupertford: Actually I think Imperial agreed to pass copyright to STFC, but we obviouly keep authorship information.

src/psyclone/parse.py
```diff
@@ -879,15 +889,21 @@ def parse(alg_filename, api="", invoke_name="invoke", inf_name="inf",
     >>> ast,info=parse("argspec.F90")
 
     '''
+    _config = configuration.ConfigFactory().create()
```
rupertford: better coding. Good.

src/psyclone/psyGen.py
```diff
@@ -194,24 +202,27 @@ def __str__(self):
...
-
-    def __init__(self, api="", distributed_memory=config.DISTRIBUTED_MEMORY):
+    '''
+    Creates a specific version of the PSy. If a particular api is not
+    provided then the default api, as specified in the psyclone.cfg
+    file, is chosen.
+    '''
+    def __init__(self, api="", distributed_memory=None):
         '''Initialises a factory which can create API specific PSY objects.
         :param api: Name of the API to use.
```
rupertford: This should be 'param str api'?

src/psyclone/psyGen.py
```diff
@@ -1524,14 +1535,24 @@ def _encloses_omp_directive(self):
 
 
 class OMPDoDirective(OMPDirective):
+    '''
+    Class representing an OpenMP DO directive in the PSyclone AST.
+
+    :param list children: list of Nodes that are children of this Node.
+    :param parent: the Node in the AST that has this directive as a child.
```
rupertford: datatype should be here?

src/psyclone/tests/config_test.py
```diff
@@ -0,0 +1,402 @@
...
+
+@pytest.fixture(scope="module",
+                params=["DISTRIBUTED_MEMORY",
+                        "REPRODUCIBLE_REDUCTIONS",
+                        "COMPUTE_ANNEXED_DOFS"])
+def bool_entry(request):
+    '''
+    Parameterised fixture that will cause a test that has it as an
+    argument to be run for each boolean member of the configuration file
+    :param request: Object through which to access current parameter value
```
rupertford: type of request should be specified?

src/psyclone/tests/config_test.py
```diff
@@ -0,0 +1,402 @@
...
+    return request.param
+
+
+@pytest.fixture(scope="module",
+                params=["REPROD_PAD_SIZE"])
+def int_entry(request):
+    '''
+    Parameterised fixture that returns the names of integer members of the
+    configuration file.
+    :param request: Object through which to access current parameter value
```
rupertford: and again?

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -57,7 +57,17 @@
...
+ROOT_PATH = os.path.dirname(os.path.dirname(os.path.dirname(
+    os.path.dirname(os.path.abspath(__file__)))))
+# Construct the path to the default configuration file
+DEFAULT_CFG_FILE = os.path.join(ROOT_PATH, "config", "psyclone.cfg")
+
+TEST_API = "dynamo0.3"
 
+# Our configuration object
+from psyclone.configuration import ConfigFactory
+_CONFIG = ConfigFactory().create()
```
rupertford: Should this be here? Actually, looking at the code, the question has turned into - will there be any side effects when running tests in parallel due to having this here, rather than a local create per test (even though it should be the same instance). If not, please ignore.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -6748,8 +6757,7 @@ def test_arg_discontinuous(monkeypatch, annexed):
 
     # 1 discontinuous field returns true
     # Check w3, wtheta and w2v in turn
-    import psyclone.config
-    monkeypatch.setattr(psyclone.config, "COMPUTE_ANNEXED_DOFS", annexed)
+    monkeypatch.setattr(_CONFIG, "_compute_annexed_dofs", annexed)
```
rupertford: Create one here rather than at top of the file? There could be side effects between different tests in parallel?

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -7170,18 +7177,17 @@ def test_no_halo_exchange_annex_dofs(tmpdir, f90, f90flags, monkeypatch,
     fewer halo exchange call generated.
 
     '''
-    import psyclone.config
-    monkeypatch.setattr(psyclone.config, "COMPUTE_ANNEXED_DOFS", annexed)
+    monkeypatch.setattr(_CONFIG, "_compute_annexed_dofs", annexed)
```
rupertford: Again, create new reference to config singleton here?

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -7191,36 +7197,37 @@ def test_no_halo_exchange_annex_dofs(tmpdir, f90, f90flags, monkeypatch,
...
-    config.COMPUTE_ANNEXED_DOFS is False, as a previous writer may
+    Config.compute_annexed_dofs is False, as a previous writer may
     have iterated over dofs and only written to its own dofs, leaving
     the annexed dofs dirty. This test checks these two cases. Note the
     former case should currently never happen in real code as a halo
     exchange would not be added in the first place.
     '''
-    import psyclone.config
-    monkeypatch.setattr(psyclone.config, "COMPUTE_ANNEXED_DOFS", False)
+    monkeypatch.setattr(_CONFIG, "_compute_annexed_dofs", False)
```
rupertford: Perhaps ignore my suggestion as 3 uses is on the edge, as long as there are no side effects.

src/psyclone/tests/dynamo0p3_transformations_test.py
```diff
@@ -59,6 +59,10 @@
 BASE_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                          "test_files", "dynamo0p3")
 
+# Our configuration object
+from psyclone.configuration import ConfigFactory
+_CONFIG = ConfigFactory().create()
```
rupertford: Again should this be created in each test so tests don't interfere?. If not then leave as it is. Especially as it i used alot!

src/psyclone/tests/psyGen_test.py
```diff
@@ -927,19 +928,19 @@ def test_named_invoke_name_clash():
...
-    '''Check that we raise an exception if the pad size in config.py is
+def test_invalid_reprod_pad_size(monkeypatch):
+    '''Check that we raise an exception if the pad size in psyclone.cfg is
     set to an invalid value '''
-    from psyclone import config
-    keep = config.REPROD_PAD_SIZE
-    config.REPROD_PAD_SIZE = 0
+    # Make sure we monkey patch the correct Config object
+    from psyclone import psyGen
+    monkeypatch.setattr(psyGen._CONFIG, "_reprod_pad_size", 0)
```
rupertford: Thanks for fixing/improving

src/psyclone/tests/test_files/dummy_config.cfg
```diff
@@ -0,0 +1,53 @@
...
+DEFAULTAPI = dynamo0.3
+SUPPORTEDSTUBAPIS = dynamo0.3
+DEFAULTSTUBAPI = dynamo0.3
+DISTRIBUTED_MEMORY = true
+REPRODUCIBLE_REDUCTIONS = false
+# Ammount to pad the local summation array when REPRODUCIBLE_REDUCTIONS is true
+REPROD_PAD_SIZE = 8
+
+[dynamo0.3]
+OMNI_MODULE_PATH = /some/directory
```
rupertford: Should this be here in this PR? What about annexed_dofs?

src/psyclone/tests/test_files/dummy_config.cfg
```diff
@@ -0,0 +1,53 @@
...
+
+[gocean1.0]
+OMNI_MODULE_PATH = /some/directory
```
rupertford: Should this be here?

src/psyclone/virtual_utils.py
```diff
@@ -0,0 +1,52 @@
...
+# COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: R. W. Ford and A. R. Porter, STFC Daresbury Lab
```
rupertford: I don't think my name is warranted here.

src/psyclone/configuration.py
arporter: Yes, I agree. I've therefore changed the implementation so that Dynamo 0.3-specific settings are now stored in their own object (which itself belongs to the Config object). I've written some developer documentation about this too.

src/psyclone/dynamo0p3.py
arporter: Comment added (built-ins have self._kern == None).

src/psyclone/dynamo0p3.py
arporter: Done.

src/psyclone/dynamo0p3.py
arporter: I learnt this from Joerg - without it the generated documentation for the interface is not formatted correctly.

src/psyclone/generator.py
arporter: No - done.

src/psyclone/parse.py
arporter: Done.

src/psyclone/psyGen.py
arporter: Just checking you were on the ball :-)

src/psyclone/psyGen.py
arporter: Oh, yes.

src/psyclone/tests/config_test.py
arporter: Done and done.

src/psyclone/tests/dynamo0p3_test.py
arporter: No, there should be no side effects because I'm careful to only monkeypatch this Config object. I do create local Config objects in some of the tests when I want to initialise them with duff data.

arporter: Different tests in parallel each have their own python process and therefore their own Config object.

src/psyclone/tests/dynamo0p3_test.py
arporter: Monkeypatch will keep things local.

src/psyclone/tests/test_files/dummy_config.cfg
arporter: Oops, that was a hangover from old code. I've removed it and put in annexed_dofs instead.

arporter:
Ready for next round...

src/psyclone/configuration.py
```diff
@@ -0,0 +1,426 @@
...
+    Create our singleton Config object. If config_file is specified
+    then we throw-away the old Config and create a new one.
+
+    :param str config_file: Specific configuration file to use when \
+                            creating the Config object.
+    '''
+    _instance = None  # Our single Config object
+
+    def __init__(self, config_file=None):
+        if not ConfigFactory._instance or config_file:
```
rupertford: For consistency you should have the arguments described here not in the class description

src/psyclone/configuration.py
```diff
@@ -0,0 +1,426 @@
...
+
+    :param config: The 'parent' Config object.
+    :type config: :py:class:`psyclone.configuration.Config`
+    :param section: The entry for the dynamo0.3 section of \
+                    the configuration file, as produced by ConfigParser.
+    :type section:  :py:class:`configparser.SectionProxy`
+
+    '''
+    def __init__(self, config, section):
+
```
rupertford: Again for consistency with other classes please move argument descriptors from class description to a docstring for this method.

src/psyclone/configuration.py
```diff
@@ -0,0 +1,426 @@
...
+        '''
+        return self._config_file
+
+
+class DynConfig(object):
+    '''
+    Dynamo0.3-specific Config sub-class. Holds configuration options specific
+    to the Dynamo 0.3 API.
+
+    :param config: The 'parent' Config object.
```
rupertford: I note, you have started using the "describe __init__ args in the class description". I thought you didn't like this - and in fact this is not the way you have done it in the past? I think we can ignore for the moment as we don't seem to have one agreed style.

doc/system_specific_setup.rst
```diff
@@ -165,9 +165,24 @@ Cloning PSyclone from git and setting up your environment is done as follows::
...
+immediately.
+
+.. warning::
+
+   On OpenSUSE it is necessary to add `$HOME/.local/bin` to
+   your `$PATH` if you have done a user-local install.
+
+.. warning::
+
+   when installing in 'editable' mode, pip does
```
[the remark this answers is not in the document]
arporter: Good point. Done.

arporter:
Assuming CI is all happy, this is ready for review again...

Issue #119, opened 2018-01-02
Add support for insertion of monitoring hooks


arporter:
This was previously ticket 748 on the MO SRS (https://code.metoffice.gov.uk/trac/lfric/ticket/748).

In the discussion on that ticket it was agreed that we should support a generic interface that takes
the names of both the calling procedure and module as arguments:

    call profile_start( module_name, procedure_name )
    ...
    call profile_stop( module_name, procedure_name )

arporter:
Rather than have to do string matching in the `profile_stop()` call, I would advocate that the `profile_start()` call return an integer identifier for the region being created. This could then be passed to `profile_stop()`:

    call profile_start(module_name, procedure_name, region_idx)
    ...
    call profile_stop(region_idx)

Having said that though, the granularity with which we would expect to use this profiling API should mean that there will be significant work between the start and stop calls so timing overhead should only be a small percentage.

arporter:
For want of a better way of bringing people in, @rupertford, @MatthewHambley were involved in the discussion on the original ticket.

hiker:
Not sure how you envision the profiling calls to be added - you mention a modification to gen(). While I understand the reason in https://code.metoffice.gov.uk/trac/lfric/ticket/748 to only support one API in psyclone, I would be interested in supporting TAU, but the tau API requires a custom data type as additional parameter, which I am not sure can be easily handled:
```
subroutine F1()
integer profiler(2) / 0, 0 /
save    profiler
call TAU_PROFILE_TIMER(profiler,'f1()')
call TAU_PROFILE_START(profiler)
...
call TAU_PROFILE_STOP(profiler)
end
```
Your approach would at least force me to re-implement parts that are already in tau (mapping of a data type ('profiler' in my example, 'region_idx' in your example) to a region/subroutine name) - and then to also dynamically manage the tau data structure for each region index.

My approach would have been to use a transform, since this would allow me to also add the declaration, so it would appear to be less work. I admit I can't easily evaluate how to best support tau (a custom transform would give more flexibility, e.g. if you want to measure individual kernels in one invoke, the transform could add the start/stop calls around all outer loops).

arporter:
Thanks for your input @hiker - I have to admit I'm not familiar with the details of using TAU but is there anything stopping us from implementing the data structure within the interface to TAU? e.g. something like:

    module profiling
      integer, parameter :: MAX_TIMERS = 100
      integer :: profiler(2, MAX_TIMERS)
[...]

Issue #170, opened 2018-05-15
OpenACC support for GOcean 1.0


arporter:
OpenACC 2.6 (https://www.openacc.org/sites/default/files/inline-files/OpenACC.2.6.final.pdf) now supports "deep-copying" of members of Fortran derived types from the host to the device.
This makes adding support to PSyclone for OpenACC much more straightforward.
In this issue we will implement this support with the GOcean 1.0 API as the first target.

arporter:
In fact, requiring that modules (that PSyclone does not currently touch) have OpenACC mark-up added to them is something that would have to be taken on by the infrastructure developer. It's very likely that other changes to infrastructure will be required to get OpenACC working anyway (e.g. I've added the `data_on_device` attribute to GOcean field objects).

arporter:
The original work for OpenACC support was done under ticket 322 on the MetOffice SRS on branch https://github.com/stfc/PSyclone/tree/PSyclone_r3720_OpenACC_directives/src. However, that was 2 years ago so I'm going to start again with a fresh `openacc` branch.

arporter:
I can get the basics working quite quickly but this involves putting GOcean API-specific code inside `psyGen.py`. This is because `AccParallelDirective.var_list()` creates a list of just the field names required by a given kernel (`Kernel.arguments.args[idx].name`) . For the GOcean 1.0 API we must append "%data" to get the actual array being passed to the kernel. For the OpenACC `copyin` directive I think I need the name of the field object *and* the data array within it.

arporter:
Although this issue is explicitly targeting the GOcean API, I do need to design with other APIs in mind.
The Dynamo 0.3 code that we would be putting inside a parallel region might look something like:

      DO cell=1,mesh%get_last_halo_cell(1)
        !
        CALL matrix_vector_mm_code(cell, nlayers, v_proxy%data, s_proxy%data, &
                mm_proxy%ncell_3d, mm_proxy%local_stencil, ndf_any_space_1_v, &
                undf_any_space_1_v, map_any_space_1_v(:,cell))
      END DO

arporter:
The kernel arguments for Dynamo 0.3 are marshalled by the `dynamo0p3.ArgOrdering` class. This would probably be the place to add an `acc_var_list()` method or similar. We will also need to deal with the fact that flagging that a field is on a device will be API specific (e.g. in GOcean we set `fld_obj%data_on_device = .true.`).

arporter:
[...]

PR #171, opened 2018-05-16
Monitoring hooks

hiker:
A first pull request for #119 (adding support for monitoring hooks).

Documentation added, three wrapper libraries (dummy example, simple timer, and dl_timer ), test cases seem to cover all of the new code.

Review please ;)


hiker:
As far as I can see it's apparently additional colouring codes in the output. I can't reproduce the issue:
```
$ WITH_TERMCOLOR=1 coverage run --source=psyclone -m py.test  ./profile_test.py
============================================================================= test session starts =============================================================================
platform linux2 -- Python 2.7.14, pytest-3.2.3, py-1.4.34, pluggy-0.4.0
rootdir: /home/joerg/work/psyclone, inifile:
plugins: pylint-0.7.1, pep257-0.0.5, pep8-1.0.6, flakes-2.0.0, cov-2.5.1
collected 9 items                                                                                                                                                              

profile_test.py .........

========================================================================== 9 passed in 21.52 seconds ==========================================================================
```
What do I need to do to enable the colour codes in the output? WITH_TERMCOLOR by itself seems not to now, but I can't find anything else in .travis.yml

arporter:
That's weird. You just need to install the `termcolor` package...

arporter:
(it's in the documentation :-) )

hiker:
*grin* but not in the part that lists all packages to install - but I admit the 'installation' and 'getting started' chapters have been on my todo list for a while :)

Anyway, I need more time - while I could fix some of the failures, they colour codes break re syntax , so either I need to think of something to make the colour codes valid RE, or remove the color codes from the psyclone output, or completely avoid using re (which is easiest tbh) ... tomorrow.

hiker:
OK, I've fixed the colourisation related test failures, and also the usage of the singleton namespace manager (which is used for variable names) to ensure we create unique region names. Tests also added.

Some comments (probably not that relevant for this ticket);
1. IMHO tests should fail if termcolor is not available but WITH_TERMCOLOR is 1. If you agree, am happy to open a ticket.
2. I don't think that the colourisation tests add anything to the test cases, they just become more complicated. What about adding a 'no_colour' option to view, that could be used when testing?
3. Finding out about this colour support, I also added a colour to the ProfileNode schedule view. Afaik termcap only supports 8 colour names, so I reused green (well, technically black is still available, but I though that was not really a colour ;) ).

If the gods of CI agree, ready for review.

hiker:
OK, am only blaming my own stupidity here. Bow the gods of CI agree, and it is indeed ready for review.

arporter:
Thanks for this @hiker, I shall do the review today. The environment variable you found (WITH_TERMCOLOR) is *only* for the purposes of controlling the test environment on Travis. We need to test the case when termcolor is installed in a user's environment as well as when it's not (since we support both cases). Probably this should be documented somewhere - the upcoming developers' documentation (part of Rupert's annexed-dofs PR) would be the place for this.

arporter:
As we are now taking responsibility for distributing some Fortran code with PSyclone, we need to consider what level of support we are providing for that. Although not something for this issue, I think we should (somehow) extend the test suite to include checks for the various `profile_mod` implementations. Having said that, building a given implementation requires that we have the profiling library that it wraps installed. Therefore, maybe we can only do it for @hiker's simple example (although we could bring in dl_timer as a git submodule)?

src/psyclone/transformations.py
```diff
@@ -1669,3 +1669,118 @@ def apply(self, outer):  # pylint: disable=arguments-differ
...
+        # We need to make sure that the nodes are consecutive children,
+        # otherwise code might get moved in an incorrect order
+        prev_position = -1
+        for child in node_list:
+            if child.parent is not node_parent:
+                raise TransformationError(
+                    "Error in {0} transformation: supplied nodes "
+                    "are not children of the same Schedule/parent."
+                    .format(str(self)))
+            if prev_position>=0 and prev_position+1 != child.position:
```
arporter: Need white space around operators.

src/psyclone/transformations.py
```diff
@@ -1669,3 +1669,118 @@ def apply(self, outer):  # pylint: disable=arguments-differ
...
+        # the ProfileNode astheir parent. Use a slice
+        # of the list of nodes so that we're looping over a local
+        # copy of the list. Otherwise things get confused when
+        # we remove children from the list.
+        for child in node_list[:]:
+            # Remove child from the parent's list of children
+            node_parent.children.remove(child)
+            child.parent = profile_node
+
+        # Add the OpenMP region directive as a child of the parent
```
arporter: Please update this comment. If this transform has a lot in common with the OpenMP one could they share code somehow? Actually, I've just looked and I don't think there is much in the way of re-use.

doc/profiling.rst
```diff
@@ -0,0 +1,239 @@
...
+
+    from psyclone.profiler import Profiler
+
+    ...
+
+    _, INVOKEINFO = parse("shallow_alg.f90", api=API)
+
+    Profiler.set_options([Profiler.INVOKES, Profiler.KERNELS])
+    PSY = PSyFactory(API).create(INVOKEINFO)
+
```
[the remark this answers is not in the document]
hiker: OK, I've removed it. I thought it could be useful for devs, but the code is also in the profile test files.

doc/transformations.rst
```diff
@@ -49,6 +49,10 @@ API-specific sections).
     :members:
     :noindex:
 
+.. autoclass:: psyclone.transformations.ProfileRegionTrans
```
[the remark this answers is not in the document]
hiker: I just tried, and it worked for me, both pdf and html :(

src/psyclone/profiler.py
```diff
@@ -0,0 +1,212 @@
...
+    @staticmethod
+    def add_profile_nodes(schedule, loop_class):
+        '''This function inserts all required Profiling Nodes (for invokes
+        and kernels, as specified on the command line) into a schedule.
+        :param loop_class: The loop class (e.g. GOLoop, DynLoop) to instrument.
+        "type loop_class: :py::class::`psyclone.psyGen.Loop` or derived class.
+        '''
+
+        from psyclone.transformations import ProfileRegionTrans
+        if Profiler.profile_kernels():
```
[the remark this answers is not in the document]
hiker: I am not entirely sure here - invoke profiling will very likely work in most (all?) cases. And when I played around with existing scripts, they all aborted with appropriate error messages. Yes, if you are using scripts chances are that kernel profiling will not work (or better: break the scripts). We could not allow a script when enabling automatic kernel profiling?

src/psyclone/profiler.py
```diff
@@ -0,0 +1,212 @@
...
+        # should be only one kernel, but if Profile is invoked after e.g.
+        # a loop merge more kernels might be there
+        region_name = "unknown-kernel"
+        for k in self.walk(self.children, Kern):
+            region_name = k.name
+            break
+
+        region_name = Profiler.create_unique_region(region_name)
+
+        # Find the enclosing subroutine statement for declaring variables
```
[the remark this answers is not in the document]
hiker: OK, I thought I had a look at that previously, and hit some problems - but can't remember now, and all test cases still work :)

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,572 @@
...
+
+from psyclone.generator import GenerationError
+from psyclone.gocean1p0 import GOKern, GOSchedule
+from psyclone.parse import parse
+from psyclone.profiler import Profiler, ProfileNode
+from psyclone.psyGen import Loop, PSyFactory
+from psyclone.transformations import ProfileRegionTrans, TransformationError
+
+
+def get_invoke(api, algfile, key):
```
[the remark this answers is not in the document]
hiker: Done - it would be great if:
1) the new function would also allow to supply a name, since this works around problem 2:
2) in general getting the index the way it is implemented atm does not work (keys of a dictionary in general do not have to have the right order). I found this problem in some other test cases I've implemented ... let's hope #170 fixes this properly ;)

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,572 @@
...
+
+    # Insert a profile call between outer and inner loop.
+    # This forces the profile node to loop up in the tree
+    # to find the subroutine node (i.e. we are testing
+    # the while loop in the ProfileNode).
+    new_sched, _ = prt.apply(invoke.schedule.children[0]
+                             .children[0].children[0])
+
+    new_sched.view()
+    out, _ = capsys.readouterr()
```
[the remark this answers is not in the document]
hiker: OK, now I have to ask: you or Rupert requested tests that check for the schedule, and not only for the generated Fortran code. I had (admittedly a very brief) look - can you point me to a test example to show what kind of tests you want here?

This is (iirc) the only point not implemented yet.

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,572 @@
...
+        {1}[type='outer',field_space='ct',it_space='all_pts']
+            {1}[type='inner',field_space='ct',it_space='all_pts']
+                {2} bc_ssh_code(ncycle,p_fld,tmask) '''
+        '''[module_inline=False]'''
+    ).format(coloured_schedule, coloured_loop, coloured_kern, coloured_profile)
+    assert correct in out
+
+    # ... but only if we do call the actual invoke now - but no need
+    # to test the result.
+    invoke.gen()
```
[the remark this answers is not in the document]
hiker: Neither do I ... now. Sorry, the implementation and tests have gone through quite a few changes, so I am not entirely sure why I needed it in the past. Now it is gone :)

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,572 @@
...
+
+    Profiler.set_options(None)
+
+
+# -----------------------------------------------------------------------------
+def test_module_name_not_found():
```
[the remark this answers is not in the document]
hiker: Yes, indeed - test removed.

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,572 @@
...
+                and i._call.items[0] == "\"unknown module\"":
+            found = True
+
+    assert found
+
+    Profiler.set_options(None)
+
+
+# -----------------------------------------------------------------------------
+def test_profile_errors2():
```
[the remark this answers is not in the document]
hiker: Yes (though there is still one useful test in this subroutine).

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,572 @@
...
+    assert re.search(correct_re, code, re.I) is not None
+    Profiler.set_options(None)
+
+
+# -----------------------------------------------------------------------------
+def test_profile_kernels_dynamo0p3():
+    '''Check that all kernels are instrumented correctly
+    '''
+    Profiler.set_options([Profiler.KERNELS])
+    _, invoke = get_invoke("dynamo0.3", "1_single_invoke.f90", 0)
```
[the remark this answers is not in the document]
hiker: Turns out this test case was actually doing more or less the same thing as the next one ... so I've removed it.

src/psyclone/transformations.py
hiker: Yes - tbh, the openmp transform is broken (as far as I can see: call it with (schedule.children[2], schedule.children[1]) as parameter, and as far as I can see it will silently swap the two nodes without any test. I intend to open a separate bug (and pull request with an x-failing test), and once this pull request is merged in, refactor the common code into a new function or baseclass. But I didn't want to include this whole issue in this pull request.

hiker:
Thanks a lot for the very detailed review - it has helped finding heaps of errors - old code, copy&paste.

The only outstanding issue are the tests for the schedule (where I've added the colour codes) - as indicated above: could you please show me an example what kind of test you expect? Then I'll fix the remaining test cases and pass it back to you :)

hiker:
OK, am giving up for now - my python3 environment is somewhat borked and doesn't work. And I admit I don't get the latest failures - in https://github.com/stfc/PSyclone/pull/171/commits/92831fcd955ce7dde3342e039c77330117f078ce I added a ```list()``` around the output of keys() in line 75, but the CI output at https://travis-ci.org/stfc/PSyclone/jobs/388817172 shows the same line 75 without the list(). I must be missing something.

The failures in psyGen_test.py are (very likely) followup problems because the profile tests aborted and did not reset the profiling flags. I'll add a tear-down function tomorrow to make sure that failures in profile do not affect other tests.

hiker:
OK, I think I got it all sorted out: un-borked my python 3 installation, python3 tests now all working, and also replaced the colour-dependent test to one without colour - at least I hope this was what you wanted me to do.
I left one colour-dependent test in place, thinking that it would improve code coverage, but just found out apparently it doesn't - am happy to replace this with a simple str(schedule) test as well.
Back to you ... finally :)

src/psyclone/psyGen.py
```diff
@@ -255,6 +261,7 @@ class PSy(object):
 
     '''
     def __init__(self, invoke_info):
+        # pylint:disable=dangerous-default-value
```
arporter: Unless you have a good reason for disabling this, I would prefer to leave it alone (or fix the dangerous default value :-))

src/psyclone/tests/generator_test.py
```diff
@@ -380,6 +380,49 @@ def test_main_version(capsys):
...
+                  '--profile', 'invokes', filename])
+    assert Profiler.profile_kernels()
+    assert Profiler.profile_invokes()
+
+    # Check for missing parameter (argparse then
+    # takes the filename as parameter for profiler):
+    with pytest.raises(SystemExit):
+        main(options+["--profile", filename])
+    _, outerr = capsys.readouterr()
+    assert "invalid choice" in outerr
```
arporter: Please check that this error message now helpfully tells the user what their options are for the `--profile` flag.

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,547 @@
...
+    assert re.search(correct_re, code, re.I) is not None
+
+    _, invoke = get_invoke("gocean1.0", "single_invoke_"
+                           "two_kernels.f90", 0)
+
+    # Convert the invoke to code, and remove all new lines, to make
+    # regex matching easier
+    code = str(invoke.gen()).replace("\n", "")
+
+    correct_re = ("subroutine invoke.*"
```
arporter: Please add a comment describing what this test is for. Please also extend the test to check that the arguments to the profiling-calls are correct.

src/psyclone/tests/profile_test.py
```diff
@@ -0,0 +1,547 @@
...
+                  "do j.*"
+                  "do i.*"
+                  "call.*"
+                  "end.*"
+                  "end.*"
+                  r"call ProfileEnd\((?P=profile2)\)")
+    groups = re.search(correct_re, code, re.I)
+    assert groups is not None
+    # Check that the variables are different
+    assert groups.group(1) != groups.group(2)
```
arporter: Again, please also check that the actual arguments passed to the profiling API are correct.

src/psyclone/tests/test_files/gocean1p0/single_invoke_two_identical_kernels.f90
```diff
@@ -0,0 +1,55 @@
+!-------------------------------------------------------------------------------
```
arporter: Please could you replace this copyright notice with the full one, see e.g. `test28_invoke_kernel_stencil.f90`

src/psyclone/transformations.py
```diff
@@ -1669,3 +1669,118 @@ def apply(self, outer):  # pylint: disable=arguments-differ
...
+    >>> ast,invokeInfo=parse(filename,api=api,invoke_name="invoke")
+    >>> psy=PSyFactory(api).create(invokeInfo)
+    >>>
+    >>> from psyclone.psyGen import TransInfo
+    >>> t=TransInfo()
+    >>> p_trans= t.get_trans_name('ProfileRegionTrans')
+    >>>
+    >>> schedule=psy.invokes.get('invoke_0').schedule
+    >>> schedule.view()
+    >>> new_schedule=schedule
```
arporter: You don't need this line (1692)

src/psyclone/transformations.py
```diff
@@ -1669,3 +1669,118 @@ def apply(self, outer):  # pylint: disable=arguments-differ
...
+
+    @property
+    def name(self):
+        ''' Returns the name of this transformation as a string '''
+        return "ProfileRegionTrans"
+
+    def apply(self, nodes):
+        # pylint: disable=arguments-differ
+        '''Apply this transformation to a subset of the nodes within a
+        schedule - i.e. enclose the specified Loops in the
```
arporter: s/Loops/Nodes/ since the region will include e.g. kernel calls too.

arporter:
Just had another thought: please could you add tests that the profiling transformation plays nicely with e.g., the OpenMP transformations? I know it *should* but we should check...

hiker:
I have most things fixed.

> Finally, I think it makes sense to prevent the user from requesting auto profiling of kernels if they are also passing an optimisation script (-s argument). Better that PSyclone refuses to try than silently fails to add any profiling hooks.

Well, more likely the user script would fail imho, but agreed. I added a warning message in this case - only a warning since I think there might be cases in which it is just fine (e.g. invoke profiling should work, and a carefully designed transform script can handle inserted profile nodes as well). Alternative I am happy to abort in this case, but then would prefer another option (--force-profile??) to still go ahead.

> re. my concern about thread safety, I think we have three options:
>     we could print a warning (log message really) if profiling is within a threaded region;
>     we could add an is_threadsafe() query function (or even just a public logical parameter) to the profiling wrapper API and generate a call to this;
>     since we know which profilers we support and we know whether or not they're thread-safe, we could get the user to tell us which profiler they're building for.
> 

IMHO I wouldn't worry, since profiling will be used by 'advanced' user who know what they are doing. Printing a warning is (Imho) annoying (I really prefer warning-free compilations - and that would include PSyclone ;) ). What should threadsafe() do? Like a lock? And it is not possible (in general) to see if a library is threadsafe or not: tau can be installed with or without thread support - you would have to check an environment variable to see if the strings contains omp somewhere. 
But you tell me what you prefer, am happy to implement one way or another ;)

And no idea where this "disable=dangerous-default-value" came from ... I suspect a incorrect copy&paste ;)

That leaves the OMP tests - now I unfortunately hit a problem in some of my tests - trying to insert a profile region inside an omp-do loop did not work as expected:
```
      !$omp parallel default(shared), private(j,i)
      !$omp do schedule(static)
      CALL ProfileStart("boundary_conditions_ne_offset_mod", "bc_ssh_code", profile)
      DO j=2,jstop
        DO i=2,istop
          CALL bc_ssh_code(i, j, 1, t%data, t%grid%tmask)
        END DO 
      END DO 
      !$omp end do
      CALL ProfileEnd(profile)
      !$omp end parallel
```
This looks like the same issue I have with MPI - the 'setDirty' calls are included in my ProfileRegions. So obviously you have two bugs - one in OpenMP directives, and one in the HaloExchange calls :)

OK, I'll have a look what I am doing wrong there ;)  But in between we can agree on the other two issues above, so I can fix them as expected.

arporter:
Thanks @hiker. re. the first problem, since the automatic profiling is applied *after* any transformation scripts, I don't think there will be any failures - just no profiling output. I think aborting by default and allowing some sort of "force it" flag is the best bet.

re. the second problem, (thread safety), the query function would just return a bool and we would generate code to abort the run as necessary. I was basing that suggestion on the fact that we have to implement a profile-wrapper library for each profiler anyway so could easily include whether or not it was thread safe. However, I hadn't realised that a given profiler might or might not be thread safe depending on the build options used :-( All in all, I think I agree with you that we won't worry. Please could you add a warning in the documentation about the need for a thread-safe profiler if profiling within thread-parallel regions?

hiker:
The OMP problem shown in the example above is actually somewhat complicated: the code created by putting a ProfileRegion inside a omp-do-directive is incorrect (only loops are allowed). IMHO the omp-do-directive should not allow a child that is not a loop, but since this has potentially far reaching consequences, I opened #178 for this, and for now added a test to the ProfileTransform to prevent this from happening. The incorrect order of statements above (#!omp end do and ProfileEnd were swapped) is actually a consequence of the omp-enddo directive moving itself 'up' in the tree to be just after the loop - which might be worth a separate discussion ;)

Otherwise openmp tests were added as requested, I have addresses the other concerns as much as I thought it would be useful (see my previous commits), but feedback welcome.

Ready for next review.

hiker:
I believe all issues have been address ;) Back for review.

doc/profiling.rst
```diff
@@ -198,6 +198,11 @@ As an example::
     newschedule, _ = p_trans.apply(schedule.children[1:3])
     newschedule.view()
 
+.. warning::
+
+    It is the responsibility of the user to make sure that a profile
```
arporter: Great. Please could you also extend section 14.2 (Profiling Command Line Options) to mention the ability/need to force PSyclone to do auto-profiling if an optimistion script is also supplied.

src/psyclone/gocean1p0.py
```diff
@@ -173,11 +175,24 @@ class GOInvoke(Invoke):
...
+        '''
+
         if False:  # pylint: disable=using-constant-test
             self._schedule = GOSchedule(None)  # for pyreverse
         Invoke.__init__(self, alg_invocation, idx, GOSchedule)
 
+        # Add profiling nodes to schedule if automatic profiling has been
+        # requested.
+        from psyclone.profiler import Profiler
+        Profiler.add_profile_nodes(self.schedule, GOLoop)
```
arporter: I've given this some more thought as it seems to me that automatic profiling shouldn't require API-specific code. In fact, the way you've written `add_profile_nodes` already means that it will work for all APIs if we pass `Loop` instead of the API-specific sub-class. However, we can only instrument a Schedule once it is complete and that part *is* API specific. Therefore, how about moving the code to add profiling into the constructor of the `Invokes` base class? e.g.:

        for idx, alg_invocation in enumerate(alg_calls.values()):
            my_invoke = Invoke(alg_invocation, idx)
            # Automatically instrument this Invoke for profiling (if enabled)
            Profiler.add_profile_nodes(my_invoke.schedule, Loop)

src/psyclone/tests/generator_test.py
```diff
@@ -380,6 +381,126 @@ def test_main_version(capsys):
...
+    assert Profiler.profile_kernels()
+    assert Profiler.profile_invokes()
+
+    # Check for missing parameter (argparse then
+    # takes the filename as parameter for profiler):
+    with pytest.raises(SystemExit):
+        main(options+["--profile", filename])
+    _, outerr = capsys.readouterr()
+
+    correct_re = "invalid choice.*choose from 'invokes', 'kernels'"
```
arporter: Great.

arporter:
I was really hoping to to give this the green light this time but there are a few tweaks needed I think. Please see the comments in-line but mainly:

1. We need to document the 'force' option in the documentation (14.2)
2. Tighten-up some of the error messages/warnings
3. I think we should re-structure the code slightly so that automatic profiling is handled in the base class rather than having to be implemented for each API.

For 3., `add_profile_nodes` will already work if just passed `Loop` instead of an API-specific sub-class of it. The API-specifc issue is that we can only add profiling once the construction of the Schedule is complete. We can get around this by adding to the `Invokes` base class instead of the `Invoke` class.
e.g. once an `Invoke` has been constructed, add profiling:

        for idx, alg_invocation in enumerate(alg_calls.values()):
            my_invoke = Invoke(alg_invocation, idx)
            Profiler.add_profile_nodes(my_invoke.schedule, Loop)
            ...

What do you think?

src/psyclone/gocean1p0.py
hiker: I quickly tested it, and it seems to work - I somehow dismissed the Invokes base class too quickly ;)

But I also noticed that my regular expression tests probably don't detect duplicated ProfileStart etc. calls (i.e. your new code AND my old code together still passes all tests ... which is strange).
I'll test this and add the other mentioned issues as well.

src/psyclone/tests/profile_test.py
```diff
@@ -566,3 +573,81 @@ def test_transform_errors(capsys):
...
+    prt = ProfileRegionTrans()
+    omp_loop = GOceanOMPLoopTrans()
+    omp_par = OMPParallelTrans()
+
+    # Parallelise the first loop:
+    sched1, _ = omp_loop.apply(schedule.children[0])
+    sched2, _ = omp_par.apply(sched1.children[0])
+    sched3, _ = prt.apply(sched2.children[0])
+
+    # \x20 is used instead of a space to avoid pep8 warning
```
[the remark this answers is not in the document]
hiker: Nice? I consider this an ugly hack, and we should fix psyclone to remove this space :)

hiker:
I had some issues with the tests when creating the profile nodes in Invokes - many tests failed because they couldn't import 'colored' - I guess I introduced a circular dependency somehow?? Anyway, I had to import profile locally in the function, not at the top of the file. But implementing the profile node creation in Invokes is indeed a great solution, thanks!

Everything else should have been addressed I think.

src/psyclone/tests/profile_test.py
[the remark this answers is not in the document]
rupertford: It's fparser that adds a space :-)
I usually go with ...
```
correct = (
    " xxxx \n"
    "yyy")
```

doc/profiling.rst
```diff
@@ -0,0 +1,246 @@
...
+with various performance measurement tools. The profiling can
+be enabled automatically using command line parameters like::
+
+    psyclone --profile kernels ...
+
+Or more fine grained by applying a profiling transformation in
+a transformation script.
+
+PSyclone can be used with a variety of existing profiling tools.
+It currently supports TAU, Dr Hook, dl_timer, and even comes
```
arporter: Please correct this line to remove reference to TAU and Dr Hook.

hiker:
Disabling the whitespace warning for pylint does not disable the same warning for pep8 - so I reformatted the whole string to use \n, as mentioned by rupert.

And I was briefly tempted to just add the tau library into this pull request (since it's already done) ... but then this story here will never end :)

Back to you :)

arporter:
New changes are good. Tests all pass on my machine now.

PR #180, opened 2018-06-21
Make gocean loop specification more flexible

hiker:
Atm it is not possible to specify a single-iteration loop (e.g. only a single halo row or columns), e.g. 
```
do i=1, 1
```
or
```
do j=x_max, x_max
``` 
This pull request makes the loop boundary specification more flexible to allow to specify a loop 'nest' that only covers one (or two, ...) rows/columns:
```
        self._bounds_lookup['offset_ne']['ct']['s_halo'] = \
            {'inner': {'start': "1", 'stop': "{stop}+1"},
             'outer': {'start': '{stop}+1', 'stop':"{stop}+1"}}
```
This will specify an outer loop on a single halo row only.

No new iteration space is actually added in this pull request, though a few pep8/pylint issues were fixed. I'll actually add support for 1-d iteration space in a separate pull request (which then will also add tests).

### DIFF
--- a/src/psyclone/gocean1p0.py
+++ b/src/psyclone/gocean1p0.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 # -----------------------------------------------------------------------------
 # BSD 3-Clause License
 #
@@ -297,9 +298,9 @@ class GOSchedule(Schedule):
 
     def view(self, indent=0):
         ''' Print a representation of this GOSchedule '''
-        print(self.indent(indent) + self.coloured_text + "[invoke='" + \
-            self.invoke.name + "',Constant loop bounds=" + \
-            str(self._const_loop_bounds) + "]")
+        print(self.indent(indent) + self.coloured_text + "[invoke='" +
+              self.invoke.name + "',Constant loop bounds=" +
+              str(self._const_loop_bounds) + "]")
         for entity in self._children:
             entity.view(indent=indent + 1)
 
@@ -360,6 +361,7 @@ class GOSchedule(Schedule):
         self._const_loop_bounds = obj
 
 
+# pylint: disable=too-many-instance-attributes
 class GOLoop(Loop):
     ''' The GOcean specific Loop class. This passes the GOcean specific
         single loop information to the base class so it creates the one we
@@ -457,6 +459,7 @@ class GOLoop(Loop):
                     {'inner': {'start': "1", 'stop': "{stop}+1"},
                      'outer': {'start': "1", 'stop': "{stop}+1"}}
 
+    # pylint: disable=too-many-branches
     def _upper_bound(self):
         ''' Returns the upper bound of this loop as a string '''
         schedule = self.ancestor(GOSchedule)
@@ -476,9 +479,12 @@ class GOLoop(Loop):
                 stop = schedule.jloop_stop
 
             if index_offset:
-                d_bounds = self._bounds_lookup[index_offset][self.field_space] \
-                    [self._iteration_space][self._loop_type]
-                stop = d_bounds["stop"].format(start='', stop=stop)
+                # This strange line splitting was the only way I could find
+                # to avoid pep8 warnings: using [..._space]\ keeps on
+                # complaining about a white space
+                bounds = self._bounds_lookup[index_offset][self.field_space][
+                    self._iteration_space][self._loop_type]
+                stop = bounds["stop"].format(start='', stop=stop)
             else:
                 stop = "not yet set"
         else:
@@ -512,6 +518,7 @@ class GOLoop(Loop):
                     stop += "%ystop"
         return stop
 
+    # pylint: disable=too-many-branches
     def _lower_bound(self):
         ''' Returns a string containing the expression for the lower
         bound of the loop '''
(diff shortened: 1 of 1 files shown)


src/psyclone/gocean1p0.py
```diff
@@ -396,67 +398,68 @@ def __init__(self, parent=None,
 
         # Loop bounds for a mesh with NE offset
         self._bounds_lookup['offset_ne']['ct']['all_pts'] = \
-            {'inner': {'start': "1", 'stop': "+1"},
-             'outer': {'start': "1", 'stop': "+1"}}
+            {'inner': {'start': "1", 'stop': "{stop}+1"},
```
arporter: It feels better/more consistent to have `{stop}+1` rather than just `+1` - thanks.

src/psyclone/gocean1p0.py
```diff
@@ -525,9 +533,17 @@ def _lower_bound(self):
...
+                stop = schedule.jloop_stop
             if index_offset:
-                start = (self._bounds_lookup[index_offset][self.field_space]
-                         [self._iteration_space][self._loop_type]["start"])
+                # This strange line splitting was the only way I could find
+                # to avoid pep8 warnings: using [..._space]\ keeps on
+                # complaining about a white space
+                bounds = self._bounds_lookup[index_offset][self.field_space][
+                    self._iteration_space][self._loop_type]
+                start = bounds["start"].format(start='', stop=stop)
```
arporter: It seems wrong/unnecessary to have to specify stop when getting a lower bound. Is it really necessary? My reading of the code is that the string stored under the "start" key won't have a `stop` format identifier?

src/psyclone/gocean1p0.py
```diff
@@ -297,9 +298,9 @@ def __init__(self, alg_calls):
 
     def view(self, indent=0):
         ''' Print a representation of this GOSchedule '''
```
arporter: Please add interface information to this docstring.

src/psyclone/gocean1p0.py
```diff
@@ -360,6 +361,7 @@ def const_loop_bounds(self, obj):
         self._const_loop_bounds = obj
 
 
+# pylint: disable=too-many-instance-attributes
 class GOLoop(Loop):
```
arporter: Please add information on the interface to the constructor to the class docstring.

src/psyclone/gocean1p0.py
```diff
@@ -396,67 +398,68 @@ def __init__(self, parent=None,
...
             for itspace in VALID_ITERATES_OVER:
                 self._bounds_lookup[offset]['every'][itspace] = \
-                    {'inner': {'start': "1", 'stop': "+1"},
-                     'outer': {'start': "1", 'stop': "+1"}}
+                    {'inner': {'start': "1", 'stop': "{stop}+1"},
+                     'outer': {'start': "1", 'stop': "{stop}+1"}}
 
+    # pylint: disable=too-many-branches
     def _upper_bound(self):
         ''' Returns the upper bound of this loop as a string '''
```
arporter: Please alter this docstring so as to describe the interface to this routine.

src/psyclone/gocean1p0.py
```diff
@@ -511,6 +518,7 @@ def _upper_bound(self):
                     stop += "%ystop"
         return stop
 
+    # pylint: disable=too-many-branches
     def _lower_bound(self):
         ''' Returns a string containing the expression for the lower
```
arporter: Please alter this docstring to describe the interface to this routine.

src/psyclone/gocean1p0.py
hiker: No, when you need a loop over a 1d halo region you will need both start and stop in upper and lower boundaries, e.g. for a (future) loop only for the eastern halo:
```
        self._bounds_lookup['offset_ne']['ct']['e_halo'] = \
            {'inner': {'start': '{stop}+1', 'stop': '{stop}+1'},
             'outer': {'start': "2", 'stop': "{stop}"}}
```
So in this case we actually only need an outer loop (across all rows, i.e. only j loop), the inner 'loop' is ```do i=istop+1, istop+1```, which on gen() will be replaced with ```i=istop+1``` (well, at least it will be like this once I have further pull requests in ;) ).

So we do need start and stop for both upper and lower boundaries.

arporter: Ah yes - thanks for explaining.

src/psyclone/gocean1p0.py
```diff
@@ -525,9 +557,17 @@ def _lower_bound(self):
             if go_kernels:
                 index_offset = go_kernels[0].index_offset
 
+            if self._loop_type == "inner":
```
arporter: Is this part going to be OK in the presence of a GOLoopSwap transformation? I guess it is because I've had a quick look at the code for that transformation and AFAICS it doesn't modify  the loop_type of the loops being transformed. Is that correct?

arporter:
Thanks for answering my questions and updating the doc strings. pylint and pycodestyle are happy. I just have one more question about what happens if someone applies a loop swap transformation (see inline). Once that is resolved I'll merge this on.

src/psyclone/gocean1p0.py
hiker: Yes, it just changes the tree, not the content of any of the nodes. So the former outer loop is after the transform the inner loop, but it is still marked as 'outer' loop (maybe instead of inner and outer, we should call it dim1 and dim2 or so ;) ). Besides, the swap transformation tests would have detected this.

Issue #179, opened 2018-06-18
Python3 compilation tests failing


rupertford:
Andy has reported that "a few of the compilation tests fail with Python 3 (due to a bug in the CompilationError constructor)"

TeranIvy:
Merged PR #183 so closing this issue.

PR #183, opened 2018-07-02
#179 fix problem in CompileError constructor in py3

arporter:
Tiny bug fix for compilation tests run under Python 3.


TeranIvy:
Happy to take this.

PR #189, opened 2018-07-18
Fix different profile names

### LINKED ISSUE #188: Profile interface creates different variable names on each gen_code call
From Andy:

> I'm just doing some re-factoring work and in the course of that I've discovered that if you call gen() on the same invoke more than once then the names of the ProfileData variables changes! The first time through it's named 'profile', the second time it becomes 'profile_1'. I guess in practice we'll get away with this but it's something that needs fixing. e.g. adding the line `print(str(invoke.gen())`
> in test_profile_invokes_gocean1p0 breaks the test! Since you're going to be doing more work on the profiling interface I thought it best to just email you about it rather than open another issue. 
>

hiker:
Fixes #188


src/psyclone/tests/profile_test.py
```diff
@@ -137,10 +137,16 @@ def test_profile_basic(capsys):
                              .children[0].children[0])
 
     new_sched_str = str(new_sched)
+
+    # Note that the first profile node is created using the get_invoke
```
arporter: You could make `coloured_text` into a static method and then you wouldn't need to create a ProfileNode just to get the coloured text? Alternatively you could use `invoke.schedule.children[0]` as that is a ProfileNode.

src/psyclone/tests/profile_test.py
```diff
@@ -197,6 +203,11 @@ def test_profile_invokes_gocean1p0():
                   r"call ProfileEnd\(profile\)")
     assert re.search(correct_re, code, re.I) is not None
 
+    # Check that if gen() is called more than once the same profile
+    # variables and region names are created:
+    code_again = str(invoke.gen()).replace("\n", "")
+    assert code == code_again
```
arporter: Great.

src/psyclone/profiler.py
```diff
@@ -148,12 +147,14 @@ def __init__(self, children=None, parent=None):
             :type parent: A :py::class::`psyclone.psyGen.Node`.
         '''
         Node.__init__(self, children=children, parent=parent)
-        self._namespace = NameSpace()
+        self._var_name = NameSpaceFactory().create().create_name("profile")
```
arporter: Please add comments describing the purpose of these three variables.

arporter:
Changes are good. I just have a couple of small requests around comments/doc strings.
Code is pylint and pycodestyle clean. This is a bug fix so no documentation change required.
Coverage of the diff is 100%.

src/psyclone/profiler.py
```diff
@@ -148,12 +147,14 @@ def __init__(self, children=None, parent=None):
...
+        self._region_name = None
+        self._module_name = None
 
     # -------------------------------------------------------------------------
     def __str__(self):
         ''' Returns a name for the ProfileNode. '''
```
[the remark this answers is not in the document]
hiker: Ha - view() and str() ... I have been wanting to address this for a long time. But it's out of scope for this ticket, so I'll send a separate email or ticket ... and have updated the docs for now to what str() actually does (and not what imho it should be doing).

src/psyclone/tests/profile_test.py
hiker: While I like static, it would be inconsistent with every other use of coloured_text - so I just use your clever invoke.schedule.children[0] idea - thanks!

PR #197, opened 2018-07-25
Use gen_invoke from utils.py everywhere

hiker:
OK, I admit that this change got a bit out of hand :)  I started with fixing the one todo (use gen_invoke from utils in profiler_test) ... and then I ended up doing:
- Update utils.gen_invoke so that there is no need to add the api-specific directory in the caller, instead set the directory based on API (leaving out gungho, since afaik that's deprecated)
- use the new gen_invoke everywhere, and remove all the os.path.join that previously added the api-specific subdirectory name
- Then I also didn't like the 'import utils' (imho this is just asking for trouble sooner or later), so I did a proper from psyclone.tests.utils import ... (which needed a __init__ file in tests), and replaced all utils.XXX uses appropriately.

### DIFF
--- a/src/psyclone/tests/gocean1p0_transformations_test.py
+++ b/src/psyclone/tests/gocean1p0_transformations_test.py
@@ -49,7 +49,7 @@ from psyclone.transformations import TransformationError, \
     GOceanOMPLoopTrans, KernelModuleInlineTrans, GOceanLoopFuseTrans, \
     ACCParallelTrans, ACCDataTrans, ACCLoopTrans
 from psyclone.generator import GenerationError
-from utils import count_lines, get_invoke
+from psyclone.tests.utils import count_lines, get_invoke
 
 # The version of the PSyclone API that the tests in this file
 # exercise
@@ -60,9 +60,8 @@ def test_const_loop_bounds_not_schedule():
     ''' Check that we raise an error if we attempt to apply the
     constant loop-bounds transformation to something that is
     not a Schedule '''
-    _, invoke = get_invoke(
-        os.path.join("gocean1p0",
-                     "test11_different_iterates_over_one_invoke.f90"), API, 0)
+    _, invoke = get_invoke("test11_different_iterates_over_one_invoke.f90",
+                           API, idx=0)
     schedule = invoke.schedule
     cbtrans = GOConstLoopBoundsTrans()
 
@@ -73,9 +72,8 @@ def test_const_loop_bounds_not_schedule():
 def test_const_loop_bounds_toggle():
     ''' Check that we can toggle constant loop bounds on and off and
     that the default behaviour is "on" '''
-    psy, invoke = get_invoke(
-        os.path.join("gocean1p0",
-                     "test11_different_iterates_over_one_invoke.f90"), API, 0)
+    psy, invoke = get_invoke("test11_different_iterates_over_one_invoke.f90",
+                             API, idx=0);
     schedule = invoke.schedule
     cbtrans = GOConstLoopBoundsTrans()
 
@@ -118,9 +116,8 @@ def test_const_loop_bounds_invalid_offset():
     ''' Test that we raise an appropriate error if we attempt to generate
     code with constant loop bounds for a kernel that expects an
     unsupported grid-offset '''
-    psy, invoke = get_invoke(
-        os.path.join("gocean1p0",
-                     "test26_const_bounds_invalid_offset.f90"), API, 0)
+    psy, invoke = get_invoke("test26_const_bounds_invalid_offset.f90",
+                             API, idx=0)
     cbtrans = GOConstLoopBoundsTrans()
     schedule = invoke.schedule
     newsched, _ = cbtrans.apply(schedule, const_bounds=True)
@@ -132,9 +129,8 @@ def test_const_loop_bounds_invalid_offset():
 def test_loop_fuse_different_iterates_over():
     ''' Test that an appropriate error is raised when we attempt to
     fuse two loops that have differing values of ITERATES_OVER '''
-    _, invoke = get_invoke(
-        os.path.join("gocean1p0",
-                     "test11_different_iterates_over_one_invoke.f90"), API, 0)
+    _, invoke = get_invoke("test11_different_iterates_over_one_invoke.f90",
+                           API, idx=0)
     schedule = invoke.schedule
     lftrans = LoopFuseTrans()
     cbtrans = GOConstLoopBoundsTrans()
@@ -155,9 +151,7 @@ def test_loop_fuse_different_iterates_over():
 
 def test_loop_fuse_unexpected_error():
     ''' Test that we catch an unexpected error when loop fusing '''
-    _, invoke = get_invoke(
-        os.path.join("gocean1p0",
-                     "test14_module_inline_same_kernel.f90"), API, 0)
+    _, invoke = get_invoke("test14_module_inline_same_kernel.f90", API, idx=0)
     schedule = invoke.schedule
 
     lftrans = GOceanLoopFuseTrans()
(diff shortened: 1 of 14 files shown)


hiker:
I need some help with compiling tests. ATM it fails with:
```
>       monkeypatch.setattr(psyclone.tests.utils, "TEST_COMPILE", value=False)
E       NameError: global name 'psyclone.tests.utils' is not defined
```
And same issue if I use only utils. What do I need to do here?

arporter:
Have you imported either psyclone in the former case or utils in the latter?

hiker:
Thanks!
I had the corresponding ```from ... import ...```. It looks like monkey-patch needs an ```import ...```. So I'll just add this import this in test_compile_string now - update patch coming, I just need to do the pep and pylint and tests and so on :)

arporter:
In reviewing Rupert's updates to fparser2, I've realised that you don't need to import anything for monkeypatch, you can just do it all in a string:

    monkeypatch.setattr("psyclone.tests.utils.TEST_COMPILE", value=False)

hiker:
OK, I modified the patch to use the same technique. Note that the syntax was wrong, correct is:
```
  monkeypatch.setattr("psyclone.tests.utils.TEST_COMPILE", False)
```
(no ```value=```).

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -60,9 +60,8 @@ def test_const_loop_bounds_not_schedule():
     ''' Check that we raise an error if we attempt to apply the
     constant loop-bounds transformation to something that is
     not a Schedule '''
-    _, invoke = get_invoke(
-        os.path.join("gocean1p0",
-                     "test11_different_iterates_over_one_invoke.f90"), API, 0)
+    _, invoke = get_invoke("test11_different_iterates_over_one_invoke.f90",
+                           API, idx=0)
```
arporter: Nice simplification.

src/psyclone/tests/utils.py
```diff
@@ -282,9 +283,20 @@ def get_invoke(algfile, api, idx=None, name=None):
...
+    if api == "gocean1.0":
+        dir_name = "gocean1p0"
+    elif api == "dynamo0.3":
+        dir_name = "dynamo0p3"
+    elif api == "gocean0.1":
+        dir_name = "gocean0p1"
+    elif api == "dyanmo0.1":
+        dir_name = "dynamo0p1"
+    else:
+        assert False
```
arporter: Please put in a verbose exception here to aid a future developer who adds to the available APIs.

hiker:
I don't know enough about the test coverage stats and how they are affected, but I would otherwise suggest to rename this file to test_utils.py and move it from .../psyclone/tests to .../psyclone - then we can import it with the full name. Though ... this then would require quite a few new unit tests for test_utils :(

hiker:
I had a look, and it appears that we can specify files to be excluded from py.test coverage runs - so can we tell it to ignore ```tests``` for coverage? I think there is even a config file we could use to do that. Am happy to try, but what is the command that produces the statistics:
```
py.test  --cov=psyclone  --cov-report term-missing  --cov-report=html 
```
?

hiker:
I tried to ignore the tests directory from the report, and while this appears to be possible with coverage.py, it does not appear possible with py.test - at least I give up now ;)

So, I guess we are down to either moving utils.py from tests to src/psyclone, or removing the __init__.py file. Considering that I don't feel like even trying to cover utils.py with tests, I'd suggest that I just undo the modifications to the import statement.

arporter:
Upon reflection over the weekend, I think we should not have the tests directory as part of the PSyclone package (i.e. it should not have an __init__.py file). I think the best solution is to rename "utils.py" to "psyclone_test_utils.py" and then use unqualified imports, e.g.:

    from psyclone_test_utils import COMPILE

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -289,10 +335,11 @@ def get_invoke(algfile, api, idx=None, name=None):
         dir_name = "dynamo0p3"
     elif api == "gocean0.1":
         dir_name = "gocean0p1"
-    elif api == "dyanmo0.1":
+    elif api == "dynamo0.1":
         dir_name = "dynamo0p1"
     else:
-        assert False
+        raise RuntimeError("The API '{0}' is not supported by get_invoke.".
```
arporter: Thanks for improving this bit. Please could you extend the error message to tell the user which APIs are supported.

arporter:
Thanks for re-naming the utils module and updating all of the imports. Almost there now. I think that for a test to be useful it must be picked-up automatically (because we will almost certainly forget to make sure it gets run otherwise :-) ). How about changing `compile_test.py` to `utilities_test.py` and moving the new test into there?

hiker:
I thought I had posted that it's ready for review ... apparently I forgot. So, ready for review.
Only potential problem: I added support for gungho (so all supported APIs now work), but I could not get any existing test case for gungho to work - so I commented the test for gungho out for now

arporter:
Thanks @hiker. I'd drop the test for gungho altogether if I were you. (Really we should remove all mention of it from PSyclone but that's definitely another issue.) Final change - please could you correct the module docstring for the `utilities_test.py` to say that it tests various utilities, not just those used for compilation.

Once that's done I'll merge it.

src/psyclone/tests/utilities_test.py
```diff
@@ -114,31 +115,59 @@ def test_find_fortran_file(tmpdir):
...
+    '''Tests get_invokes. '''
+
+    # First test all 4 valid APIs - we only make sure that no exception
+    # is raised, so no assert required
+    _, _ = get_invoke("openmp_fuse_test.f90", "gocean0.1", idx=0)
+    _, _ = get_invoke("test14_module_inline_same_kernel.f90",
+                      "gocean1.0", idx=0)
+    _, _ = get_invoke("algorithm/1_single_function.f90", "dynamo0.1", idx=0)
+    _, _ = get_invoke("1_single_invoke.f90", "dynamo0.3", idx=0)
+    # It seems that none of the gungho tests actually work :(
```
arporter: Please delete this commented-out test for the 'gungho' API.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -338,3 +301,58 @@ def string_compiles(code, tmpdir, f90, f90flags):
...
+
+    if api == "gocean1.0":
+        dir_name = "gocean1p0"
+    elif api == "dynamo0.3":
+        dir_name = "dynamo0p3"
+    elif api == "gocean0.1":
+        dir_name = "gocean0p1"
+    elif api == "dynamo0.1":
+        dir_name = "dynamo0p1"
+    elif api == "gunghoproto":
```
arporter: Please remove support for "gunghoproto" here since the attempts to test it revealed more problems...

arporter: Oh, you'll have to update the message for the `RuntimeError` below too (sorry).

arporter:
Almost, almost there - you just need to remove the support for "gunghoproto" from `get_invoke()` and the associated, commented-out test code.

hiker:
After my first commit I changed my mind, and implemented a better / more flexible fix for this. hope that's all good now. I wonder if it might be worth to add a mapping api->directory_name to the config file (e.g. change support_api_list to be a dictionary, and add a function to access the mapping). This way we would have only one place to change when adding/removing APIs. But that's not part of this pull request ;) Let me know if you think that's a good idea, then I'll add a ticket (or maybe even add this as part of my 'add iteration spaces' pull request).

I've opened #207 to remove gungho, but now this pull request should be independent of this.

Ready for review ... again.

hiker:
> Given that the mapping between API and directory name is something that only the test suite needs to worry about I don't think it should be exposed in the PSyclone source (if that's what you're suggesting?

Not directly: I don't want to expose the name of the test directories, but define a mapping of API name to a 'sensible' potential file- or directory-name (e.g. sensible meaning dot free). The advantages would be in the future, e.g. if we restructure the file layout so that each API has its own directory, we could get rid of code like this:
```
if self._type == "gunghoproto":
    return GHProtoKernelType(ast, name=name)
elif self._type == "dynamo0.1":
    return DynKernelType(ast, name=name)
...
```
and just import KernelType from an API specific directory (though that would need an eval() as much as I can tell). And we then would use this 'sensible' name as test directory names as a side effect :)

Anyway, doesn't really matter - I just felt a need to write that I did not want to simply move test-related code into PSyclone ;)

Thanks a lot for your patience with my pull requests - highly appreciated!

PR #200, opened 2018-07-27
Fix for #176: OMP transform can create invalid code

### LINKED ISSUE #176: Incorrect parameters to OMP transforms can lead to invalid code and python exceptions
A x-failing test case is in:

https://github.com/hiker/PSyclone/blob/openmp-test/src/psyclone/tests/gocean1p0_transformations_test.py
(test_omp_region_with_slice_change_order, lines 290)

1) the omp transform can change the order of statements
2) a ValueError can be raised if the same child is listed more than once (imho it should be a TransformationError).

Note that my monitoring hooks have a fixed that should solve those issues, and I am happy once the monitoring hooks are merged to master to fix this issue properly.

hiker:
This fixes #176 (making sure that OMPParallelTrans does not create incorrect or invalid code or crashes).


hiker:
I added a new validate functions as a 'mixin' class, i.e. multiple inheritance is used to get access to the function. Advantage: the class definition now points to CheckChildrenList, so it's more obvious that the dependency exists, and it can use 'self' to get the name of the transform (and in an OO environment I don't like global functions :) ). But on the other hand, it is actually only a small function, and we need a pylint disable directive (since pylint complains that self.name does not exist).

Alternative: I can easily just turn it into a simple function. Or I would make this based on Transform and use CheckChildrenList as base class instead of transform (but I am not sure if this is possible, the OMP transform classes depend somewhat on each other, I am not sure if this would work).

Anyway, let me know what you prefer ;)

src/psyclone/transformations.py
```diff
@@ -1474,6 +1513,7 @@ def _validate(self, node_list):
 
         # Now call the general validation checks
         super(OMPParallelTrans, self)._validate(node_list)
+        CheckChildrenList._validate(self, node_list)
 
 
 class ACCParallelTrans(ParallelRegionTrans):
```
arporter: Although this Issue was originally to do with OpenMP, I expect that the OpenACC implementation that has made it on to master in the interim suffers from the same flaw (since I wrote both of them :-(). Please could you extend your fix to the ACC classes.

I'm also thinking that it would probably be nice if we had some sort of `RegionTrans` base class that could be used by OpenACC/OpenMP/Profiling/PSyKE. Maybe @TeranIvy could consider this as she works on PSyKE?

TeranIvy: `RegionTrans` base class sounds like a good technical solution, @arporter. Happy to implement it in PSyKE branch!

hiker: I had a look at the inheritance diagram, and the existing ParallelRegionTrans is at the right location. I'll move the additional tests into _validate of that class.

hiker:
Oops, sorry - I missed that part, I only read the comments since my previous commit.

As you requested I put it in the developers section, but I am not sure if this ideal: there appear to be three potential sections for documentation: api, transform, and developers. Assuming that the idea is that the normal user (who is not going to write transforms) only needs to know how to use them, moving this into developers is of course correct. I am just not sure what the api section is supposed to be (well, for whom should this section be written) - but I leave this for a separate cleanup later.

I would also make a case that _validate() should be called validate(), since imho it should be a protected member (a class must call the validate() function of its base class, so validate() is not a private/internal function, which a leading _  otherwise indicates in python). I don't consider this part of this pull request, but as a start I added an abstract _validate function with generic documentation for everything to Transform, so it can be added to developers.rst (and added ":private-members:" so that _validate is actually included in the docs).

I also tried to address the various pylint complains about incorrect number of parameters (for apply and _validate). So as a start I added a *args argument to the abstract methods in Transform, so it is obvious in the developers doc of the base class that there will be other arguments.

Unfortunately *args does not solve the pylint warning. I could remove those by adding *_ as last parameter of all apply and validate functions (since then pylint considers it to be consistent with the method in Transformation) - but this then means we need to add tests in each apply and validate function to make sure that those additional arguments are actually empty (and then add test cases for that) - not a nice way.

So I guess the only option is then to disable pylint's test for number of arguments everywhere, which I do not consider part of this request :)

Ready for review.

arporter:
Thanks @hiker, that's done the trick. I think Developers' is the right place as this is implementation detail, and should not be in the user guide. (The Developers' Guide is in the User Guide because that's the only way we've found to get readthedocs to include it.)

PR #209, opened 2018-08-24
Nemo API

arporter:
Initial support for the NEMO API.


arporter:
Ready for first review. This PR is by no means "complete" but it's already big enough. I haven't added much in the way of documentation and the functionality is incomplete. However, there is full test coverage and I've added two new examples.

src/psyclone/psyGen.py
```diff
@@ -1914,6 +1928,49 @@ def _encloses_omp_directive(self):
...
+        '''
+        from fparser.common.readfortran import FortranStringReader
+        from fparser.two.Fortran2003 import Comment
+        # Check that we haven't already been called
+        if self._ast:
+            return
+        # Find the locations in which we must insert the begin/end
+        # directives...
+        # Find the children of this node in the AST of our parent node
+        try:
```
rupertford: I'm sure you don't need me to tell you that this insertion is a bit messy and we need to better support this sort of thing in fparser2, especially adding directives.

src/psyclone/psyGen.py
```diff
@@ -1914,6 +1928,49 @@ def _encloses_omp_directive(self):
...
+        text = "!$omp parallel default(shared), private({0})".format(
+            ",".join(self._get_private_list()))
+        startdir = Comment(FortranStringReader(text,
+                                               ignore_comments=False))
+
+        # Create the end directive and insert it after the node in
+        # the AST representing our last child
+        enddir = Comment(FortranStringReader("!$omp end parallel",
+                                             ignore_comments=False))
+        if end_idx == len(self._parent._ast.content) - 1:
```
rupertford: It's unclear to me why you have this if clause here

src/psyclone/psyGen.py
```diff
@@ -2081,6 +2137,67 @@ def gen_code(self, parent):
...
+            return
+        # Since this is an OpenMP (parallel) do, it can only be applied
+        # to a single loop.
+        if len(self._children) != 1:
+            raise GenerationError(
+                "An OpenMP PARALLEL DO can only be applied to a single loop "
+                "but this Node has {0} children: {1}".
+                format(len(self._children), self._children))
+
+        # Find the locations in which we must insert the begin/end
```
rupertford: Good comments. Thanks.

src/psyclone/psyGen.py
```diff
@@ -1264,6 +1271,13 @@ def is_openmp_parallel(self):
     def gen_code(self):
         raise NotImplementedError("Please implement me")
 
+    def update(self):
```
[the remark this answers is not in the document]
rupertford: You add support for OMP PARALLEL as well as OMP PARALLEL DO but you don't use the former in an example. I've not checked the tests yet.

src/psyclone/psyGen.py
```diff
@@ -2325,8 +2442,10 @@ def loop_type(self):
 
     @loop_type.setter
     def loop_type(self, value):
-        assert value in self._valid_loop_types, \
-            "Error, loop_type value is invalid"
+        if value not in self._valid_loop_types:
```
rupertford: raise generation error rather than an assert. Good.

src/psyclone/psyGen.py
```diff
@@ -3555,3 +3674,54 @@ def name(self):
 
     def apply(self):
         return None, None
+
+
+class IfBlock(Node):
```
rupertford: Added if and elseif as nodes. Good.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        result += "EndLoop"
+        return result
+
+    @property
+    def kernel(self):
+        ''' Returns the kernel object if one is associated with this loop,
+        None otherwise. '''
+        kernels = self.walk(self.children, NemoKern)
+        if kernels:
+            # TODO cope with case where loop contains >1 kernel (e.g.
```
rupertford: We should raise an exception here, not just blindly continue!

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+    syntax).
+
+    :param ast: the part of the fparser2 AST representing the loop
+    :type ast: :py:class:`fparser.two.Fortran2003.Assignment_Stmt`
+    :param parent: the parent of this Loop in the PSyclone AST
+    '''
+    def __init__(self, ast, parent=None):
+        from fparser.common.readfortran import FortranStringReader
+        Loop.__init__(self, parent=parent,
+                      valid_loop_types=VALID_LOOP_TYPES+["unknown"])
```
rupertford: perhaps unknown should be added to the main list instead of adding it as a string in more than one place (which is dangerous).

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        subsections = walk_ast(ast.items, [Fortran2003.Section_Subscript_List])
+        # A Section_Subscript_List is a tuple with each item the
+        # array-index expressions for the corresponding dimension of the array.
+        for idx, item in enumerate(subsections[0].items):
+            if isinstance(item, Fortran2003.Subscript_Triplet):
+                # If an array index is a Subscript_Triplet then it is a range
+                # and thus we need an explicit loop for this dimension.
+                outermost_dim = idx
+
+        # TODO allow for implicit loops with specified bounds (e.g. 2:jpjm1)
```
rupertford: If this is not supported then we should raise an exception if this situation arises.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        self._start = 1
+        self._step = 1
+        if outermost_dim == 0:
+            # TODO ensure no name clash is possible with variables that
+            # already exist in the NEMO source.
```
rupertford: Good point. Please add to the list of issues.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        for idx, item in enumerate(subsections[0].items):
+            if isinstance(item, Fortran2003.Subscript_Triplet):
+                # If an array index is a Subscript_Triplet then it is a range
+                # and thus we need an explicit loop for this dimension.
+                outermost_dim = idx
+
+        # TODO allow for implicit loops with specified bounds (e.g. 2:jpjm1)
+        self._start = 1
+        self._step = 1
+        if outermost_dim == 0:
```
rupertford: We should parameterise the ordering to allow for different ones i.e. have a lookup based on a config file.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        prog_unit = self.root.invoke._ast
+        spec = walk_ast(prog_unit.content, [Fortran2003.Specification_Part])
+        if not spec:
+            names = walk_ast(prog_unit.content, [Fortran2003.Name])
+            # TODO create a Specification_Part
+            raise InternalError("No specification part found for routine {0}!".
+                                format(names[0]))
+        # TODO check that this variable has not already been declared
+        # Requires that we capture all variable declarations in the routine
+        # in some sort of management class.
```
rupertford: the namespace manager can do this if I understand correctly.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        for subsec in subsections:
+            # A tuple is immutable so work with a list
+            indices = list(subsec.items)
+            # Replace the colon with our new variable name
+            indices[outermost_dim] = name
+            # Replace the original tuple with a new one
+            subsec.items = tuple(indices)
+
+        # Create an explicit loop
+        text = '''\
```
rupertford: It's a small thing but I prefer using
(
"xxx\n"
"yyy\n"
)
as it allows us to inline the code more nicely.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+do {0}=1,{1},{2}
+  replace = me
+end do
+'''.format(self._variable_name, self._stop, self._step)
```
rupertford: Actually we need a higher level (f2pytrans style) approach to create a loop, or extend the fparser API. But that is clearly future work.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        :returns: True if the node does represet an implicit loop.
+        :rtype: bool
+        '''
+        if not isinstance(node, Fortran2003.Assignment_Stmt):
+            return False
+        # We are expecting something like:
+        #    array(:,:,jk) = some_expression
+        # so we check the left-hand side...
+        lhs = node.items[0]
+        if isinstance(lhs, Fortran2003.Part_Ref):
```
rupertford: Again I think that whether a statement uses array syntax or not should be provided by a library.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+            start_idx = clause_indices[idx]
+            # No need to subtract 1 here as Python's slice notation means
+            # that the end_idx'th element is excluded
+            end_idx = clause_indices[idx+1]
+            ast.content[start_idx]._parent = ast  # Retrofit parent info
+            self.addchild(NemoIfClause(ast.content[start_idx:end_idx],
+                                       parent=self))
+
+    def gen_code(self):
+        ''' Override abstract method of base class. '''
```
rupertford: raise exception if this is called?

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+            raise InternalError(
+                "Unrecognised member of if block: '{0}'. Expected one of "
+                "Else_Stmt or Else_If_Stmt.".format(type(ast_nodes[0])))
+        # Continue on down the AST
+        self.process_nodes(parent=self,
+                           nodes=ast_nodes[1:],
+                           nodes_parent=self._ast._parent)
+
+    def gen_code(self):
+        ''' Override abstract method of base class. '''
```
rupertford: raise exception?

doc/nemo.rst
```diff
@@ -0,0 +1,96 @@
...
+      
+.. _nemo-api:
+
+NEMO API
+========
+
+In contrast to the other APIs supported by PSyclone, the NEMO API is
+designed to work with source code that does *not* follow the PSyKAl
+separation of concerns. Instead, the NEMO source code is treated as if
+it were a manually written PSy layer with all kernels in-lined. This
```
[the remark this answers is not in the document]
arporter: Not really. In PSyKAl the Algorithm layer contains calls to `invoke()` that are then turned into calls to generated PSy layer routines. In NEMO we are treating each subroutine as an `invoke` that consists of zero or more (inlined) kernel calls. As such we never consider the algorithm.

doc/nemo.rst
```diff
@@ -0,0 +1,96 @@
...
+         Loop[type='levels',field_space='None',it_space='None']
+             Loop[type='lat',field_space='None',it_space='None']
+                 KernCall[Implicit]
+         Loop[type='levels',field_space='None',it_space='None']
+             Loop[type='lat',field_space='None',it_space='None']
+                 KernCall[Implicit]
+         Loop[type='levels',field_space='None',it_space='None']
+             Loop[type='lat',field_space='None',it_space='None']
+                 Loop[type='lon',field_space='None',it_space='None']
+                     KernCall[]
```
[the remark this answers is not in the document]
arporter: Created #222.

doc/nemo.rst
```diff
@@ -0,0 +1,96 @@
...
+:cite:`nemo_code_conv` in order to reason about the code being
+processed. Rather than construct a Schedule for the PSy layer from
+scratch (as is done for other APIs), the Schedule is constructed by
+parsing the supplied Fortran code and generating a higher-level
+representation.
+
+Algorithm
+---------
+
+Since NEMO source is treated as a pre-existing PSy layer, this API
```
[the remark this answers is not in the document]
arporter: Halo exchanges only exist in in the PSy layer so I think that's correct. Control code is more difficult but I don't see how we could create an Algorithm layer?

examples/nemo/eg1/runme_openmp.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
+## -------------------------------------------------------------------------
```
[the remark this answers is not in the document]
arporter: Oops! Done.

examples/nemo/eg1/runme_openmp.py
```diff
@@ -0,0 +1,53 @@
...
+
+    print("Invokes found:")
+    print(psy.invokes.names)
+    
+    sched = psy.invokes.get('tra_adv').schedule
+    sched.view()
+
+    TRANS_INFO = TransInfo()
+    print(TRANS_INFO.list)
+    FUSE_TRANS = TRANS_INFO.get_trans_name('LoopFuse')
```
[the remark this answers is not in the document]
arporter: Oops. Removed.

examples/nemo/eg1/runme_openmp.py
```diff
@@ -0,0 +1,53 @@
...
+    OMP_TRANS = TRANS_INFO.get_trans_name('OMPParallelLoopTrans')
+
+    for loop in sched.loops():
+        kernel = loop.kernel
+        if kernel:
```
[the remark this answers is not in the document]
arporter: Good point.

examples/nemo/eg1/runme_openmp.py
```diff
@@ -0,0 +1,53 @@
...
+            if loop.loop_type == "levels":
+                sched, _ = OMP_TRANS.apply(loop)
```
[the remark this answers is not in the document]
arporter: I've extended the documentation and added READMEs.

examples/nemo/eg1/runme_openmp.py
```diff
@@ -0,0 +1,53 @@
...
+
+    sched.view()
+
+    psy.invokes.get('tra_adv').schedule = sched
+    print(psy.gen)
```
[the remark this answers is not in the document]
arporter: There's an x-failing test for the first problem. I've created a new `Limitations` section in the NEMO API documentation.

examples/nemo/eg2/omp_levels_trans.py
```diff
@@ -0,0 +1,60 @@
...
+    from psyclone.psyGen import PSyFactory, TransInfo
+    # Get the Schedule of the target routine
+    sched = psy.invokes.get('tra_ldf_iso').schedule
+    # Get the transformation we will apply
+    ompt = TransInfo().get_trans_name('OMPParallelLoopTrans')
+    # Apply it to each loop over levels containing a kernel
+    for loop in sched.loops():
+        if loop.kernel and loop.loop_type == "levels":
+            sched, _ = ompt.apply(loop)
+    psy.invokes.get('tra_ldf_iso').schedule = sched
```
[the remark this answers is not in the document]
arporter: Oops, yes. (Although, since we modify the psy we're passed a reference to, it isn't strictly necessary.)

examples/nemo/eg2/runme_openmp.py
```diff
@@ -0,0 +1,80 @@
...
+
+    print("Invokes found:")
+    print(psy.invokes.names)
+    
+    sched = psy.invokes.get('tra_ldf_iso').schedule
+    sched.view()
+
+    TRANS_INFO = TransInfo()
+    print(TRANS_INFO.list)
+    FUSE_TRANS = TRANS_INFO.get_trans_name('LoopFuse')
```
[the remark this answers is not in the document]
arporter: Oops. Tidied.

examples/nemo/eg2/runme_openmp.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
arporter: I've added the CASE issue to the list of limitations. The `else` *is* a child of the `if` since it cannot exist in isolation.

src/psyclone/transformations.py
```diff
@@ -979,6 +979,34 @@ def apply(self, node):
         return OMPParallelLoopTrans.apply(self, node)
 
 
+class NemoOMPParallelLoopTrans(OMPParallelLoopTrans):
```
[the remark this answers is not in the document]
arporter: Removed.

src/psyclone/parse.py
```diff
@@ -906,6 +906,12 @@ def parse(alg_filename, api="", invoke_name="invoke", inf_name="inf",
     else:
         check_api(api)
 
+    if api == "nemo":
+        # For this API we just parse the NEMO code and return the resulting
+        # fparser2 AST
+        ast = parse_nemo(alg_filename)
+        return ast, ast
```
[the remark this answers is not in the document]
arporter: Good point. I've changed it to return None for the algorithm AST.

src/psyclone/psyGen.py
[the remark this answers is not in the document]
arporter: I've just checked and I do have tests for it.

src/psyclone/psyGen.py
arporter: Agreed.

src/psyclone/psyGen.py
arporter: Because I assumed that `some_list.insert(len(some_list), my_obj)` would be an error but it isn't - I've removed the if clause.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+from fparser.two.utils import walk_ast
+from fparser.two import Fortran2003
+
+# The base colour map doesn't have CodeBlock as that is currently
+# a NEMO-API-specific entity.
+NEMO_SCHEDULE_COLOUR_MAP = copy.deepcopy(_BASE_CMAP)
+NEMO_SCHEDULE_COLOUR_MAP["CodeBlock"] = "red"
+
+# The loop index variables we expect NEMO to use
+NEMO_LOOP_VARS = ["ji", "jj", "jk", "jt", "jn"]
```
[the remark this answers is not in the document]
arporter: My view of config files has changed a bit after various discussions with Joerg. Although I agree we want to code for future flexibility and adaptability, I'm not sure that moving these into a config file would significantly improve either of those. Do we expect the average user to want to modify these quantities or is it more of a PSyclone developer issue?

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+NEMO_LOOP_TYPE_MAPPING = {"ji": "lon", "jj": "lat", "jk": "levels",
+                          "jt": "tracers", "jn": "tracers"}
+
+
+class ASTProcessor(object):
+    '''
+    Mixin class to provide functionality for processing the fparser2 AST.
+    '''
+    @staticmethod
+    def add_code_block(parent, statements):
```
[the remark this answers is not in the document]
arporter: I've changed it to `nodes_to_code_block()`

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        '''
+        Create a NemoCodeBlock for the supplied list of statements
+        and then wipe the list of statements.
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+                                   parent=parent)
+        parent.addchild(code_block)
+        del statements[:]
+        return code_block
+
+    # TODO remove nodes_parent argument once fparser2 AST contains
+    # parent information (fparser/#102).
+    def process_nodes(self, parent, nodes, nodes_parent):
+        '''
+        Create the PSyclone AST representation of the code represented
```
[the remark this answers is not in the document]
arporter: Yes. I've added documentation for the interface to this routine too.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+
+        code_block = NemoCodeBlock(statements,
+                                   parent=parent)
+        parent.addchild(code_block)
+        del statements[:]
+        return code_block
+
+    # TODO remove nodes_parent argument once fparser2 AST contains
+    # parent information (fparser/#102).
+    def process_nodes(self, parent, nodes, nodes_parent):
```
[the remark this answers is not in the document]
arporter: Thanks - the benefit of implementing this several times is that I finally got to something good :-)

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+            elif NemoImplicitLoop.match(child):
+                # An implicit loop marks the end of any current
+                # code block
+                self.add_code_block(parent, code_block_nodes)
+                parent.addchild(NemoImplicitLoop(child, parent=parent))
+            elif NemoIfBlock.match(child):
+                self.add_code_block(parent, code_block_nodes)
+                parent.addchild(NemoIfBlock(child, parent=parent))
+            elif not isinstance(child, (Fortran2003.End_Do_Stmt,
+                                        Fortran2003.Nonlabel_Do_Stmt)):
```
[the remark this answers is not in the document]
arporter: What about it? :-) I've added this to the list of limitations.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+
+    :param ast: node in fparser2 AST representing the program unit.
+    :type ast: :py:class:`fparser.two.Fortran2003.Main_Program` or \
+               :py:class:`fparser.two.Fortran2003.Module`
+    :param str name: the name of this Invoke (program unit).
+    '''
+    def __init__(self, ast, name):
+        self._schedule = None
+        self._name = name
+        self._psy_unique_vars = ["a_variable"]
```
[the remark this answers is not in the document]
arporter: I don't know! Removed.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+
+    def update(self):
+        ''' Walk down the tree and update the underlying fparser2 AST
+        to reflect any transformations. '''
+        for invoke in self.invoke_list:
+            invoke.update()
+
+
+class NemoPSy(PSy):
+    ''' The NEMO 0.1-specific PSy class. This creates a NEMO-specific
```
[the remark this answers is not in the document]
arporter: Oh yes, and interface to constructor documented.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+class NemoCodeBlock(Node):
+    ''' Node representing some generic Fortran code that PSyclone
+    does not attempt to manipulate. As such it is a leaf in the PSyclone
+    AST and therefore has no children. '''
+
+    def __init__(self, statements, parent=None):
+        Node.__init__(self, parent=parent)
+        # Store a list of the parser objects holding the code
+        # associated with this block
+        self._statements = statements[:]
```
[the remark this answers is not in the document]
arporter: The contents of the list need to be copied because the list itself is temporary and is deleted (in the renamed `add_code_block()` routine). Possibly this indicates something sub-optimal in the way I've implemented code block generation but I can't quite see what it is.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        #print "OpenMP shared vars: " + ",".join(self._shared_vars)
+        #print "OpenMP private vars: " + ",".join(self._private_vars)
+        #print "OpenMP first-private vars: " + \
+        #    ",".join(self._first_private_vars)
+        return
+
+    def _load_from_implicit_loop(self, loop, parent=None):
+        ''' Populate the state of this NemoKern object from an fparser2
+        AST for an implicit loop (Fortran array syntax) '''
+        # TODO implement this method!
```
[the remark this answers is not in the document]
arporter: Added to the list of issues in the doc.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,822 @@
...
+        self._kernel_type = "Implicit"
+        return
+
+    def local_vars(self):
+        '''Return a list of the variable (names) that are local to this loop
+        (and must therefore be e.g. threadprivate if doing OpenMP)
+
+        '''
+        return []
```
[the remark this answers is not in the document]
arporter: Added to the list in the doc.

src/psyclone/nemo.py
arporter: Good point. Done.

src/psyclone/nemo.py
arporter: Done + test.

src/psyclone/nemo.py
arporter: Done.

src/psyclone/nemo.py
arporter: Good point. I've added it as a dict at the top of the module for now. As mentioned earlier, I'm unsure about whether this belongs in a config file.

src/psyclone/nemo.py
arporter: Oh yes, that's an idea. I was thinking I might want to store more information than just the name (e.g. type, rank etc) but I'm not sure yet...

src/psyclone/nemo.py
arporter: Done.

src/psyclone/nemo.py
arporter: Absolutely.

src/psyclone/nemo.py
arporter: Done + test.

src/psyclone/nemo.py
arporter: Done + test.

arporter:
OK, I think I've addressed all comments. I've also gone through and made sure that all the docstrings are up-to-date in the new module.
Ready for 2nd review...

examples/nemo/eg2/README
```diff
@@ -0,0 +1,52 @@
...
+
+This will output PSyclone's Schedule view of the Fortran routine followed by
+the generated Fortran code with the OpenMP directives added.
+
+The second script, omp_levels_trans.py, is intended to be provided to
+PSyclone as an optimisation script:
+
+    $ psyclone -api "nemo" -s ./omp_levels_trans.py traldf_iso.F90
+
+Again, the generated Fortran will be written to stdout.
```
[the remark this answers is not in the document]
arporter: Done.

.travis.yml
```diff
@@ -45,12 +45,16 @@ env:
...
-  # Specify "fparser" (without quotes) as an argument to install_optional.sh
-  # in order to run the tests with the version of fparser pointed to by
-  # the git submodule.
+  # Specify "fparser_submodule" (without quotes) as an argument to
+  # install_optional.sh in order to run the tests with the version of fparser
+  # pointed to by the git submodule.
   - ./bin/install_optional.sh fparser_submodule
   - pip install coveralls
   - pip install codecov
+# TODO remove following two lines (which install the master branch of
```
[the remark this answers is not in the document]
arporter: Added #235 and put this in the comment above and in .travis.yml

examples/nemo/eg2/omp_levels_trans.py
```diff
@@ -0,0 +1,60 @@
...
+# -----------------------------------------------------------------------------
+# Authors: R. W. Ford and A. R. Porter, STFC Daresbury Lab
+
+'''A simple transformation script for the introduction of OpenMP with PSyclone.
+In order to use it you must first install PSyclone. See README.md in the
+top-level psyclone directory.
+
+Once you have PSyclone installed, this script may be used by doing:
+
+ >>> psyclone -api "nemo0.1" -s ./omp_levels_trans.py traldf_iso.F90
```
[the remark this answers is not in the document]
arporter: Done.

arporter:
* If we do not monkeypatch the Assignment_Stmt in parse_test then we are fine
* If we monkeypatch but do not actually run `get_integer_array()` then we are fine
* If I put a print/pdb inside the patched function it is only called from within parse_test, not subsequently (which is right because the function I've used to patch goes out of scope)

arporter:
* Explicitly calling the `undo()` method of the monkeypatch object makes no difference.
Created a new test file containing the key test from parser_test and the first failing nemo test.
If I override the base-class `__new__` method and put in a pdb call there, I can see that it gets called if I run the 2nd (failing) test on its own. If I run both tests then it does not get called in the second test.

rupertford:
Travis is happy.

src/psyclone/nemo.py
```diff
@@ -0,0 +1,954 @@
...
+            raise InternalError("Expecting Nonlabel_Do_Stmt as first child "
+                                "of Block_Nonlabel_Do_Construct but "
+                                "got {0}".format(type(loop.content[0])))
+        self._body = []
+        for content in loop.content[1:]:
+            if isinstance(content, End_Do_Stmt):
+                break
+            self._body.append(content)
+
+        #  I could get this by walking back up the tree and counting how
```
rupertford: So presumably we want to know if we have loop nesting of 2 or 3? Where would this be used and is this part of the todo list?

src/psyclone/nemo.py
```diff
@@ -0,0 +1,954 @@
...
+        # kernel_dag = dag_of_code_block(loop, "nemo_kernel")
+        # inputs = kernel_dag.input_nodes()
+        # outputs = kernel_dag.output_nodes()
+        # print "Kernel has {0} outputs: ".format(len(outputs)) + \
+        #     ",".join([node.variable.orig_name for node in outputs])
+        self._shared_vars = set()
+        self._first_private_vars = set()
+        self._private_vars = set()
+        #  If there are scalar variables that are inputs to the DAG (other than
+        #  the loop counters) then they must be declared first-private.
```
rupertford: Please add "in an OpenMP loop directive."

src/psyclone/nemo.py
```diff
@@ -0,0 +1,954 @@
...
+        #  something that is not a loop...
+        # if len(self._loop_vars) == 2:
+        #     self._kernel_type = "2D"
+        # else:
+        #     self._kernel_type = "3D"
+
+        # TODO bring habakkuk up-to-date with changes to fparser and then
+        # uncomment this code.
+
+        #  Analyse the loop body to identify private and shared variables
```
rupertford: for use by OpenMP?

src/psyclone/psyGen.py
```diff
@@ -253,6 +254,11 @@ def create(self, invoke_info):
         elif self._type == "gocean1.0":
             from psyclone.gocean1p0 import GOPSy
             return GOPSy(invoke_info)
+        elif self._type == "nemo":
```
rupertford: Please add argumet information to the docstring in the parent method.

src/psyclone/psyGen.py
```diff
@@ -1058,6 +1064,7 @@ def __init__(self, children=None, parent=None):
         else:
             self._children = children
         self._parent = parent
+        self._ast = None  # Reference into fparser2 AST (if any)
```
rupertford: Please add docstring argument information for the method (either in the __init__ or the class docstring).

src/psyclone/psyGen.py
```diff
@@ -2321,8 +2436,10 @@ def loop_type(self):
 
     @loop_type.setter
     def loop_type(self, value):
-        assert value in self._valid_loop_types, \
-            "Error, loop_type value is invalid"
+        if value not in self._valid_loop_types:
```
rupertford: Docstring argument info required.

src/psyclone/psyGen.py
```diff
@@ -3697,3 +3814,54 @@ def name(self):
 
     def apply(self):
         return None, None
+
+
+class IfBlock(Node):
+    '''
+    Class representing an if-block within a Schedule.
+    '''
```
rupertford: Need to describe parent argument in a docstring.

src/psyclone/psyGen.py
```diff
@@ -3697,3 +3814,54 @@ def name(self):
...
+        return "If-block: "+self._condition
+
+    @property
+    def coloured_text(self):
+        '''
+        Return text containing the (coloured) name of this node type
+
+        :return: the name of this node type, possibly with control codes
+                 for colour
+        :rtype: string
```
rupertford: Isn't it str? And could you include this as part of the return seeing as it' s a simple type?

src/psyclone/psyGen.py
```diff
@@ -3697,3 +3814,54 @@ def name(self):
...
+        '''
+        return colored("If", SCHEDULE_COLOUR_MAP["If"])
+
+    def view(self, indent=0):
+        ''' Print representation of this node to stdout '''
```
rupertford: document indent arg.

src/psyclone/psyGen.py
```diff
@@ -3697,3 +3814,54 @@ def name(self):
...
+        print(self.indent(indent) + self.coloured_text + "[" +
+              self._condition + "]")
+        for entity in self._children:
+            entity.view(indent=indent + 1)
+
+
+class IfClause(IfBlock):
+    '''
+    Represents a sub-clause of an If block - e.g. an "else if()".
+    '''
```
rupertford: document parent argument.

src/psyclone/psyGen.py
```diff
@@ -3697,3 +3814,54 @@ def name(self):
...
+        self._clause_type = ""  # Whether this is an else, else-if etc.
+
+    @property
+    def coloured_text(self):
+        '''
+        Return text containing the (coloured) name of this node type
+
+        :return: the name of this node type, possibly with control codes
+                 for colour
+        :rtype: string
```
rupertford: str not string? and include as part of return?

src/psyclone/nemo.py
```diff
@@ -0,0 +1,954 @@
...
+        raise NotImplementedError("The NemoPSy.inline method has not yet "
+                                  "been implemented!")
+
+    @property
+    def gen(self):
+        '''
+        Generate the (updated) fparser2 AST for the NEMO code represented
+        by this NemoPSy object.
+
+        :rtype: :py:class:`fparser.two.Fortran2003.Main_Program` or \
```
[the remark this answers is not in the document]
arporter: Oops, so I do.

src/psyclone/nemo.py
arporter: I'm not sure of the relevance of this anymore - the way we analyse the code is more sophisticated than it was when I first wrote this (I think). The kernel type is in fact currently only used in the `view` method of the kernel. I've changed this to set `_kernel_type = "Explicit"` for the time being.

src/psyclone/nemo.py
arporter: Done.

src/psyclone/nemo.py
arporter: Added.

src/psyclone/psyGen.py

src/psyclone/psyGen.py
arporter: Done. I've also moved the constructor (and `__str__`) up to the start of the class definition as that feels more logical.

src/psyclone/psyGen.py
arporter: Oh yes. Done.

src/psyclone/psyGen.py
arporter: It is (str) and Stackoverflow says I can't :-( (http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists). Updated.

src/psyclone/psyGen.py
arporter: As before: yes and no.

arporter:
The documentation problem was because of a bug in sphinx. As a workaround I've renamed the file containing the bibitems to be "zz_references.bib" so that sphinx processes it *after* all of the other files.

arporter:
In going through and checking pylint I've realised that in this issue I've added an `_ast` property to the `Node` class, without an associated getter. However, due to the kernel-transformation support, the `psyGen.Kern` class already has an `_fp2_ast` property and an `ast` getter. Fixing that here is going to prolong the agony of this Issue so I've created #241 and added this to the list of know issues/restrictions in nemo.rst.

arporter:
Travis is happy. Ready for another review...

src/psyclone/psyGen.py
rupertford: Oops, rtfm. Although it would be nicer.

arporter:
I'd misunderstood the sphinxcontrib-bibtex workaround - I should have made sure that the rst file containing the `.. bibliography` command was processed last, not the file containing the bib entries.
Getting one (and only one) "Bibliography" section heading in both latex and html output was also tricky as latex makes such a heading automatically but html does not. I've got around that by using a `.. only:: html` to protect the heading entry in the rst file.

Issue #256, opened 2018-12-19
Represent with PSyIRe all NemoLite2D kernels that need to be converted to OpenCL


sergisiso:
This includes adding `ASTProcessor` handlers to all necessary fparser2 node types. Maybe adding new `Node` subclasses. And adding C and OpenCL generator methods to all needed PSyIRe classes.

sergisiso:
NemoLite2D needs the addition of PSyIR nodes for:
- If statements
- Return Statements
- Unary Expressions

PR #300, opened 2019-02-18
Issue #256 (partially) Add Return Stmt and UnaryOperations to PSyIR

sergisiso:
See issue #256
- [x] Return Statements
- [x] Unary Expressions


sergisiso:
@arporter @rupertford This is ready for review now, I reduced the scope of this PR to just the addition of Return Statements and Unary Operations to the PSyIR.

src/psyclone/psyGen.py
```diff
@@ -4928,6 +4928,19 @@ def _if_stmt_handler(self, node, parent):
                            nodes_parent=node)
         return ifblock
 
+    def _return_handler(self, node, parent):
+        '''
+        Transforms an fparser2 Return_Stmt to the PSyIRe representation.
```
arporter: Please s/PSyIRe/PSyIR/ throughout.

src/psyclone/psyGen.py
```diff
@@ -4928,6 +4928,19 @@ def _if_stmt_handler(self, node, parent):
...
+
+        :param child: node in fparser2 AST.
```
arporter: Should 'child' here (and on the line below) be 'node'?

src/psyclone/psyGen.py
```diff
@@ -4947,6 +4960,26 @@ def _assignment_handler(self, node, parent):
 
         return assignment
 
+    def _unary_op_handler(self, node, parent):
+        '''
+        Transforms an fparser2 UnaryOpBase to the PSyIRe representation.
+
+        :param child: node in fparser2 AST.
```
arporter: Again, s/child/node/ here?

src/psyclone/psyGen.py
```diff
@@ -5636,6 +5669,67 @@ def gen_c_code(self, indent=0):
         return self._reference
 
 
+class UnaryOperation(Node):
+    '''
+    Node representing a UnaryOperation expression. As such it has one operand
+    as children 0, and an attribute with the operator type.
```
arporter: s/children/child/

src/psyclone/psyGen.py
```diff
@@ -5636,6 +5669,67 @@ def gen_c_code(self, indent=0):
...
+
+    :param str operator: string representing the unary operator.
+    :param parent: the parent node of this UnaryOperation in the PSyIR.
+    :type parent: :py:class:`psyclone.psyGen.Node`
+    '''
+    def __init__(self, operator, parent=None):
+        super(UnaryOperation, self).__init__(parent=parent)
+        self._operator = operator
```
arporter: If possible I think it would be good to check here that the supplied operator is something valid.

src/psyclone/psyGen.py
```diff
@@ -5636,6 +5669,67 @@ def gen_c_code(self, indent=0):
...
+
+        :param int indent: Depth of indent for the output string.
+        :return: C language code representing the node.
+        :rtype: str
+        :raises GenerationError: if the node or its children are invalid.
+        '''
+        if len(self.children) != 1:
+            raise GenerationError("UnaryOperation malformed or "
+                                  "incomplete. It should have exactly 1 "
+                                  "children, but it found {0}."
```
arporter: s/but it found/but found/

src/psyclone/psyGen.py
```diff
@@ -5636,6 +5669,67 @@ def gen_c_code(self, indent=0):
...
+                                  "".format(str(len(self.children))))
```
arporter: I don't think you need the `str()` here - I think the format will take care of it.

src/psyclone/psyGen.py
```diff
@@ -5815,3 +5910,47 @@ def gen_c_code(self, indent=0):
         :rtype: str
         '''
         return self._value
+
+
+class Return(Node):
+    '''
+    Node representing a Return statement (subroutine break without return
```
arporter: Fortran is unusual in that it does not permit a return value to be supplied to the return statement. Since other languages do and this is an IR, should we future-proof ourselves here and allow for an optional return value or did you have some other plan for that functionality?

src/psyclone/tests/psyGen_test.py
```diff
@@ -3691,6 +3754,45 @@ def test_fparser2astprocessor_handling_binaryopbase(f2008_parser):
...
+    processor.process_nodes(fake_parent, [fparser2unary_operation], None)
+    # Check a new node was generated and connected to parent
+    assert len(fake_parent.children) == 1
+    new_node = fake_parent.children[0]
+    assert isinstance(new_node, UnaryOperation)
+    assert len(new_node.children) == 1
+    assert new_node._operator == '-'
+
+
+def test_fparser2astprocessor_handling_retrun_stmt(f2008_parser):
```
arporter: s/_retrun_/_return_/ :-)

src/psyclone/tests/psyGen_test.py
```diff
@@ -3691,6 +3754,45 @@ def test_fparser2astprocessor_handling_binaryopbase(f2008_parser):
...
+    ''' Test that fparser2 Return_Stmt is converted to the expected PSyIR
+    tree structure.
+    '''
+    from fparser.common.readfortran import FortranStringReader
+    from fparser.two.Fortran2003 import Execution_Part
+    reader = FortranStringReader("return")
+    return_stmt = Execution_Part.match(reader)[0][0]
+
```
arporter: Probably safest to check that `return_stmt` is the expected fparser2 node here before proceeding with the rest of the test.

src/psyclone/tests/psyGen_test.py
```diff
@@ -3691,6 +3754,45 @@ def test_fparser2astprocessor_handling_binaryopbase(f2008_parser):
...
 
+def test_fparser2astprocessor_handling_unaryopbase(f2008_parser):
+    ''' Test that fparser2 UnaryOpBase is converted to the expected PSyIR
+    tree structure.
+    '''
+    from fparser.common.readfortran import FortranStringReader
+    from fparser.two.Fortran2003 import Execution_Part
+    reader = FortranStringReader("x=-4")
+    fparser2unary_operation = Execution_Part.match(reader)[0][0].items[2]
+
```
arporter: Please add a check here that `fparser2unary_operation` is what you expect it to be.

sergisiso:
@arporter For return values, my rational was that it is just a matter of convention, you can convert a C return value to: assignment to an output Symbol + PSyIR return break . And in the other direction a PSyIR kernel can always be implemented to C/Fortran passing the output as an argument (or a C with return/Fortran function if one checks for restrictions and implement the special case).

How you end up returning the value is a language-specific detail that I think it is worth abstracting from PSyIR and deciding in the gen_* for the specific language. For instance note that in OpenCL there is this restriction "The function using the __kernel qualifier can only have return type void in the source code."

src/psyclone/psyGen.py
sergisiso: done

src/psyclone/psyGen.py
sergisiso: Ups, that is wrong in all handlers. Fixed all.

src/psyclone/psyGen.py

src/psyclone/psyGen.py
sergisiso: done

src/psyclone/psyGen.py
sergisiso: I am aware of this, and a proper solution is needed for nemolite2D (it is one of the reasons I decided to break this PR in smaller pieces).

I just created issue #339 and I will add a TODO in this line as a temporary notice.

src/psyclone/psyGen.py
sergisiso: done, also s/children/child/

sergisiso: Assignment and BianryOperator had similar messages, I updates all of them to keep consistency.

src/psyclone/psyGen.py
sergisiso: removed str()

src/psyclone/psyGen.py
sergisiso: As you say, Fortran does not have return with value and neither OpenCL (__kernels are always void type), so we don't need it for now. My feeling is that if we parse a language that has return value syntax we could specify this in the symbol table and then use an assignment to this return value + the routine exit return.

src/psyclone/tests/psyGen_test.py
sergisiso: Good catch

src/psyclone/tests/psyGen_test.py
sergisiso: done

src/psyclone/tests/psyGen_test.py
sergisiso: done

sergisiso:
@arporter I addressed your in-line comments.

Regarding documentation, the functionality introduced on this PR is documented on the new classes docstrings. Do you want me to add them somewhere in the developer's guide (I am not sure because other nodes are not there at the moment)? or will this be solved by the new split out developer's guide (#340) which I understand will have all PSyIR Nodes?

arporter:
@sergisiso I've had another look at the Developers' Guide and agree with you that there's no obvious place to put in documentation on individual nodes. That being so I think we'll stick with the fact that they are at least internally documented.

src/psyclone/psyGen.py

src/psyclone/psyGen.py
```diff
@@ -4990,12 +4990,25 @@ def _if_stmt_handler(self, node, parent):
                            nodes_parent=node)
         return ifblock
 
+    def _return_handler(self, node, parent):
```
arporter: pylint complains that `node` is unused and also that this could be a static function.

src/psyclone/tests/nemo/nemo_psy_test.py
```diff
@@ -163,8 +163,6 @@ def test_complex_code():
     sched = psy.invokes.invoke_list[0].schedule
     loops = sched.walk(sched.children, nemo.NemoLoop)
     assert len(loops) == 5
-    cblocks = sched.walk(sched.children, CodeBlock)
```
arporter: pylint complains that `CodeBlock` is now unused. Would you mind fixing the other things it complains about while you're here even though they're not your fault?

src/psyclone/tests/psyGen_test.py
```diff
@@ -3756,6 +3819,47 @@ def test_fparser2astprocessor_handling_binaryopbase(f2008_parser):
...
+    assert isinstance(return_stmt, Return_Stmt)
+
+    fake_parent = Node()
+    processor = Fparser2ASTProcessor()
+    processor.process_nodes(fake_parent, [return_stmt], None)
+    # Check a new node was generated and connected to parent
+    assert len(fake_parent.children) == 1
+    new_node = fake_parent.children[0]
+    assert isinstance(new_node, Return)
+    assert len(new_node.children) == 0
```
arporter: pylint prefers `assert not new_node.children`

src/psyclone/psyGen.py
sergisiso: I knew about this one (`ignore_handler` had the same problem) but I am unsure what is the best fix. The problem here is that the handler(node, parent) is an interface equal for all handlers and it will be called with this two arguments by resolving dynamically which handler it calls.

One solution is to use the `_` as first argument, but then I am not sure how to write the docstring (is `:param _ :` ok?). The second is to add the `pylint: disable=unused-argument` (currently in `ignore_handler`)

The static function message is even more difficult to solve, it also happens just in a few handlers because most of them do use self, but I don't think I can convert just a few of them to static functions because the call would be different (and it just looks odd). Should I maybe add a pylint disable comment here as well?

src/psyclone/tests/nemo/nemo_psy_test.py
sergisiso: Fixed all but: `Unable to import  ***` and ` Access to a protected member ***` warnings.

src/psyclone/tests/psyGen_test.py
sergisiso: fixed

src/psyclone/psyGen.py
arporter: Ah, yes, I wondered whether this was conforming to some uniform interface. I think we should use `_` for the argument and then not document them at all in the docstring, apart from a comment to say that they are there for compatibility with XXX interface.
As for the static function, I think a pylint disable is the way to go. BTW, I don't think the call would be any different because you can still just do `some_obj.some_static_method()`.

sergisiso: That works for the `return_handler(_, parent)`, but the `ignore_handler(_, _)` which also is not using parent complains with:
```
 SyntaxError: duplicate argument '_' in function definition
```
Pylint accepts ` def _ignore_handler(self, _1, _2):` is this readable enough? I will add the docstring note anyway.

sergisiso: `def _ignore_handler(self, *_):`  is also accepted by pylint for a list of ignored arguments

arporter: That combined with a docstring note seems like a nice solution.

sergisiso:
@arporter Ok, I think it is ready for next review.

Issue #281, opened 2019-01-23
Support compilation testing for the GOcean API


arporter:
Although we support compiling generated code (in the test suite) for the dynamo0.3 API, we don't yet have that functionality for GOcean1.0. We will add that here.

hiker:
Quick update:
- I've found one bug in the current handling of f90 flags (a flag like "-I /somewhere" is passed as ONE parameter to the compiler, which then tries to compile "-I /somewhere"). I've already fixed that locally.
- To proper support both APIs some refactoring is necessary. This should also allow us to avoid recompiling the whole infrastructure for each test case. I've already asked via email, but just in case: is anyone else touching test compilations atm?
-I need to supply additional compiler flag(s?) if I want to avoid compiling infrastructure libraries all the time: at least I need to add the include path so that .mod files are found (I _think_ all Fortran compilers will search for .mod files in the include path). Can I just assume that  "-I PATH" will work? Or is there a need to make this more flexible (which would mean to define compiler specific options? E.g. the cray compiler has a "-p" option where module-only search paths can be specified, but it will also search any path specified with -I, so while there could be some use to using -p instead of -I, both will work)

arporter:
I *think* that if you use a fixture then you can get the test system to build the infrastructure just once for all the tests (the 'scope' attribute of a fixture). You then have control of where to put it etc.

hiker:
Atm I am favouring something along the line:
```
    db = Dynamo0p3Build(f90, f90flags, tmpdir)
    assert db.code_compiles(psy)
```
This API-specific compilation flag can then keep track if the infrastructure files have been compiled (in a class variable), and could only be created once per file (in a setup function that's called automatically), or even all the time - its creation is cheap.

arporter:
That looks nice. I've just thought that we might need to be careful about race conditions when running the tests in parallel.

hiker:
Good point, I need to think about this. Even though the compilation has to happen in a fixed order (I don't want to support makefile-like logic in PSyclone), if a file is compiled twice at the same time we could end up with an invalid .mod or .o file. We could add a 'lock' file to the infrastructure dir, but that's kind of messy. Or maybe each process could compile the infrastructure library once in a process-specific directory.

Or worst case - the infrastructure library must be (manually) compiled first :(

arporter:
I think a process-specific directory sounds like the simplest solution.

hiker:
Leaving the parallel compilation to later, one questions:

[...]

PR #297, opened 2019-02-13
Improved profiling support

hiker:

PR #313, opened 2019-03-12
Compile gocean1.0 tests

hiker:
This is for #281.

Note that I don't have easy access to an opencl compiler, so the opencl compilation part is not tested.

I also did not properly address the fact that various combinations of compilers and options might be required for a thorough compilations test (e.g. to test openmp, openacc, opencl) code. I will open a separate issue to find out how we want to support this best.

This is a refactor of the compile environment, wrapping compilation for various APIs into separate classes. The infrastructure libraries will be compiled (once per process if compiling in parallel to avoid race conditions). Oh, I might have sneaked a sentence or two about coverage reporting into the docs while I was updating it for the compilation handling.


arporter:
Hi @hiker - is this one ready for review?

hiker:
Oops - I was certain I had set the 'ready for review' flag ... only I didn't ;)  Yes, it is ready for a first round :)

src/psyclone/tests/dynamo0p3_build.py
```diff
@@ -32,34 +32,101 @@
...
+        Dynamo0p3Build._compilation_path = str(self._tmpdir)
+
+        try:
+            # Compile each infrastructure file
+            for fort_file in Dynamo0p3Build.INFRASTRUCTURE_MODULES:
+                name = self.find_fortran_file([self._infrastructure_path],
+                                              fort_file)
+                # Abort if there should be a problem
+                if not self.compile_file(name):
+                    break
```
arporter: `Dynamo0p3Build._infrastructure_built` will still get set to `True` if this break statement is executed. Should we raise an Exception of some sort instead?

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
+
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2017-2018, Science and Technology Facilities Council
```
arporter: Possibly this should just be 2019 as this is a new file?

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: J. Henrichs, Bureau of Meteorology
+
+
+''' Module containing configuration required to build code generated
+for the Dynamo0p3 API '''
```
arporter: Gocean1p0!

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+
+from __future__ import absolute_import, print_function
+import os
+from psyclone_test_utils import Compile, CompileError
+
+
+class GOcean1p0Build(Compile):
+    '''Build class for compiling test files for the GOcean1.0 api. It
+    uses dl_esm_inf as included in the PSyclone distribution.
```
arporter: "... distribution (as a git submodule)."

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+    '''
+
+    _infrastructure_built = False
```
arporter: Could these be in the base class?

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+    def __init__(self, tmpdir):
+        super(GOcean1p0Build, self).__init__(tmpdir)
+
+        base_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                                 "test_files", "gocean1p0")
+        self.base_path = base_path
+
+        # On first instantiation (triggered by conftest.infra_compile)
+        # compile the infrastructure library files.
+        if Compile.TEST_COMPILE and not GOcean1p0Build._infrastructure_built:
```
arporter: What about TEST_OPENCL_COMPILE?

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+            self._build_infrastructure()
+            GOcean1p0Build._infrastructure_path = str(tmpdir)
```
arporter: It would make sense for `_build_infrastructure` to be responsible for storing the infrastructure path.

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+        for gocean1p0. Each parameter must be a separate entry
+        in the list, e.g.: ["-I", "/some/path"] and not ["-I /some/path"].
+        :returns: A list of strings with the compiler flags required.
+        :rtpe: list
+        '''
+        return ["-I", self._infrastructure_path]
+
+    def _build_infrastructure(self):
+        '''Compile dl_esm_inf.
+        '''
```
arporter: Please document the return values and the exceptions that can be raised.

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+
+        import subprocess
+        dl_esm_inf_path = \
+            os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                         "../../../external/dl_esm_inf/finite_difference/src")
+
+        arg_list = ["make", "F90={0}".format(self._f90),
+                    "F90FLAGS={0}".format(self._f90flags),
+                    "-f", "{0}/Makefile".format(dl_esm_inf_path)]
+        try:
```
arporter: The remainder of this method looks very similar to the equivalent in the Dynamo0p3 version of this class. Could the running of the compilation itself and the following error checking be moved into the base class? However, if that would be complicated then don't worry about it.

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+            print("Error was: ", str(err))
+            GOcean1p0Build._infrastructure_built = False
+            raise CompileError(str(err))
+        finally:
+            old_pwd.chdir()
+
+        # Check the return code
+        stat = build.returncode
+        if stat != 0:
+            print(output)
```
arporter: Write to stderr?

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+    '''
+
+    def code_compiles(self, psy):
+        '''Attempts to build the OpenCL Fortran code supplied as an AST of
+        f2pygen objects. Returns True for success, False otherwise.
+        If no Fortran compiler is available then returns True. All files
+        produced are deleted.
+
+        :param psy_ast: The AST of the generated PSy layer
+        :type psy_ast: Instance of :py:class:`psyGen.PSy`
```
arporter: s/psyGen.PSy/psyclone.psyGen.PSy/

src/psyclone/tests/gocean1p0_build.py
```diff
@@ -0,0 +1,139 @@
...
+        :return: True if generated code compiles, False otherwise
+        :rtype: bool
+        '''
+
+        if not Compile.TEST_COMPILE_OPENCL:
+            return True
+
+        # Call the internal _code_compile function, since it does
+        # not
```
arporter: "since it does not" ....what? :-)

src/psyclone/tests/gocean1p0_opencl_test.py
```diff
@@ -176,3 +207,4 @@ def test_set_arg_const_scalar():
         otrans.apply(sched)
     assert ("Cannot generate OpenCL for Invokes that contain kernels with "
             "arguments passed by value" in str(err))
+    assert GOcean1p0OpenCLBuild(tmpdir).code_compiles(psy)
```
arporter: You don't need this line as the point of this test is that we refuse to generation OpenCL code.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1242,10 +1266,28 @@ def test_module_inline_same_kernel():
     # check that the subroutine has only been inlined once
     count = count_lines(psy.gen, "SUBROUTINE time_smooth_code(")
     assert count == 1, "Expecting subroutine to be inlined once"
+    # No compilation test here, see test_module _inline_and_compile
+
+
+@pytest.mark.xfail(reason="Inline function uses a module variable")
```
arporter: Please could you create an associated Issue for this - e.g. "Validate method of kernel-inline transformation does not check for module variables" - and reference it here.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1257,7 +1299,7 @@ def test_module_inline_warning_no_change():
     _, _ = inline_trans.apply(kern_call, inline=False)
 
 
-def test_loop_swap_correct():
+def ctest_loop_swap_correct(tmpdir):
```
arporter: s/ctest/test/ otherwise this test won't be picked up by pytest.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1405,6 +1447,8 @@ def test_ocl_apply():
 
     gen = str(psy.gen)
     assert "USE clfortran" in gen
+    # ATM no support for opencl compilation
```
arporter: Is that true? Actually, I guess it is because OpenCL support in dl_esm_inf is not yet on master? If so, please could you create an issue for this and reference it here with a TODO.

src/psyclone/tests/test_files/gocean1p0/kernel_stencil.f90
```diff
@@ -67,11 +67,13 @@ module kernel_stencil
   !===================================================
 
   !> Compute the mass flux in the x direction at point (i,j)
-  subroutine compute_cu_code(i, j, cu, p, u)
+  subroutine compute_cu_code(i, j, cu, p, u, area)
```
arporter: This kernel doesn't access the `area` array?

hiker:
...
> gocean1p0_transformations_test.py:1738: AssertionError
> --------------------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------------------
> b"/home/me/Projects/GungHo/PSyclone/src/psyclone/tests/test_files/gocean1p0/kernel_scalar_float.f90:71:8:\n\n use model_mod, only: rdt\n 1\nFatal Error: Can't open module file \xe2\x80\x98model_mod.mod\xe2\x80\x99 for reading at (1): No such file or directory\ncompilation terminated.\n"

Now that one is actually a complete pain. OK, the bug itself is easy to fix, there is a 'use model_mod, only: rdt' statement in kernel_scalar_float, and 'rdt' is unused. So we can easily remove the use statement and all is good.

And it compiled for me, because (from previous work) I had some left-over .mod files in the gocean Fortran test_files directory, and 'model_mod' was one of them. 

After deleting the .mod file I got the same failed test... good :)

But then I checked where else `model_mod` is used, and found `kernel_scalar_int.f90`, which is used by `test_module_inline_with_sub_use `(parsing `single_invoke_scalar_int_arg.f90`). Surprisingly, this test still worked ... because it didn't try to compile in the first place. So I added a compile test here (which then of course failed, since it still could not find model_mod.mod), removed the 'use' statement, and replaced the variable 'used' with a simple constant. While this should compile now, the actual test fails before compilation, since it asserts that this 'use' statement is still in the produced code. And I'd say that this is an important test to make sure we are not losing some use statements somewhere.

So to properly solve this problem, I started to add code to search the kernels for other use statements, which could then be added to the list of modules that need to be compiled (see `kernel_modules` that must be compiled) before the psy-layer can be. But then the problem of course is that a kernel also 'uses' dl_esm_inf modules, which it then tries to compile (it can't easily know if a module is coming from dl_esm_inf, or the application). 

As you can see, things are getting rather complicated then (we would need to only compile 'use'd modules that are not part of dl_esm_inf, so more tests there, ...).

My suggestion: we just do not compile files that must have a 'use' statement in them (like kernel_scalar_int.f90). In many cases where the 'use' is not essential for the test I will just remove the 'use' statement and replace any variable with a simple constant or so.

Is that ok? I'd say our compilation tests will be quite thorough anyway, if there is a handful(?) of Fortran files that are not tested for compilation, I'd think that's acceptable.

> 
> I'm also struggling to remember why we thought that OpenCL compilation support needed to be separate. The FortCL library is just Fortran. OpenCL only becomes an issue at link time. However, once @sergisiso has OpenCL kernel generation working we will need to be able to verify the compilation of those. Perhaps that's what we were thinking of?

The OpenCL Fortran files have a 'use fortcl ..' statement, which we can't compile. Since I would suspect that many users have a setup similar to mine (gfortran, no opencl), I wanted to make sure that this setup works.

The rest is work in progress :)

arporter:
> My suggestion: we just do not compile files that must have a 'use' statement in them (like kernel_scalar_int.f90). In many cases where the 'use' is not essential for the test I will just remove the 'use' statement and replace any variable with a simple constant or so.
> 
> Is that ok? I'd say our compilation tests will be quite thorough anyway, if there is a handful(?) of Fortran files that are not tested for compilation, I'd think that's acceptable.
> 
Yes, that sounds fine to me. We need to tackle the whole issue of use statements in kernels anyway so it's OK to punt it for the moment.

> 
> The OpenCL Fortran files have a 'use fortcl ..' statement, which we can't compile. Since I would suspect that many users have a setup similar to mine (gfortran, no opencl), I wanted to make sure that this setup works.
> 
Ah, I see. FortCL is a submodule of dl_esm_inf but again, that's probably not yet on master of dl_esm_inf. Once it is I think compilation should be straightforward, at least for the Fortran driver stuff. The transformed kernels will be a different matter.

doc/developers.rst
```diff
@@ -90,14 +90,26 @@ you would do::
...
+showing the line numbers which are not covered. By using ``--cov`` more than once
+you can report on more than one file. You can also request
+only selected tests to be run by specifying the file names on the command line.
+Additionally html output can be created by adding the option ``--cov-report html``::
+
+  > cd <PSYCLONEHOME>/src/psyclone/tests
+  > py.test --cov-report term-missing --cov-report html --cov psyclone.dynamo0p3 ./dynamo0p3_basis_test.py ./parse_test.py
+
+The html output can be viewed with a browser at ``file:///.../tests/htmlcov/index.html``
+and it highlights all source lines in red that are not covered by at least one test.
```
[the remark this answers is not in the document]
hiker: OK, if the used library for opencl will be part of dl_esm_inf, we can fix the compiles. But on the other hand, we need to handle opencl separate anyway, so I'd recommend to leave the separate command line option in place, and handle the rest later (e.g. as part of #314),

doc/developers.rst
```diff
@@ -147,10 +166,30 @@ specified by further command-line flags::
...
+library, compilation tests must have access to compiler specific
+.mod files. For dynamo0.3 a stub implemenation of the required functions
+from the LFRic infrastructure is included in 
+``tests/test_files/dynamo0p3/infrastructure``. When compilation tests
+are requested, the stub files are compiled to create the .mod files. If you 
+compile in parallel (see :ref:`parallel_execution` section) each
+process will compile its own version of the files to avoid race conditions.
+This happens only once per process in each test session.
+
+For the gocean1.0 API a complete copy of the dl_esm_inf library is included in
```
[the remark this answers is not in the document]
hiker: Oops - fixed :)

doc/profiling.rst
```diff
@@ -220,9 +220,16 @@ As an example::
...
 ----------------------------------------
 PSyclone comes with wrapper libraries to support usage of
-dl_timer and a simple non-thread-safe timing
-library. Support for DrHook and TAU will be added in the
-future.
+Dr Hook, dl_timer and a simple non-thread-safe timing
+library. Support for further profiling libraries will be
+added in the future. To compile the wrapper libraries,
+change into the directory ``lib/profiling`` of PSyclone
+and type ``make`` to compile all wrapper. If only some
```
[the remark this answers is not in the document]
hiker: Yes, I fixed the typo, but that paragraph will be changed in #297.

src/psyclone/tests/conftest.py
```diff
@@ -86,6 +91,28 @@ def have_graphviz():
     return "graphviz" in sys.modules
 
 
+@pytest.fixture(scope="session", autouse=True)
```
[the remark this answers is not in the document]
hiker: *grin* I spent a lot of time reading about fixtures :)

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -107,206 +98,261 @@ def print_diffs(expected, actual):
...
+
+        # Build the command to execute. Note that the f90 flags are a string
+        # and so must be split into individual parts for popen (otherwise
+        # "-I /some/path" will result in the compiler trying to compile the
+        # file "-I /some/path").
+        arg_list = [self._f90, filename] + self._f90flags.split()
+        if not link:
+            arg_list.append("-c")
+        if self.get_infrastructure_flags():
+            arg_list += self.get_infrastructure_flags()
```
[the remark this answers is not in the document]
hiker: Hmm - yes and no ;) In my original implementation get_infrastructure_flags was abstract, but later I found that utilities_test best uses Compile directly, which was why I needed the if test. I now changed Compile.get_infrastructure_flags() to return [], then I could remove the if test.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -107,206 +98,261 @@ def print_diffs(expected, actual):
...
+        If no Fortran compiler is available or compilation testing is not
+        enabled then it returns True. All files produced are deleted.
+
+        :param str code: The code to compile. Must have no external
+               dependencies.
+        :return: True if generated code compiles, False otherwise
+        :rtype: bool
+
+        '''
+        if not Compile.TEST_COMPILE:
```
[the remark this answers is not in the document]
hiker: Good question :) I think they all end up using compile_file, which tested both, but it is of course much better to test earlier.

src/psyclone/tests/gocean1p0_build.py
hiker: I prefer them to be there, since it will make it easier (for #314) to handle the case where only one of the APIs will be compiled. Also, if that variable would be shared, I can't see how both infrastructure libraries/wrapper) will be compiled.

src/psyclone/tests/gocean1p0_build.py
hiker: The code in Gocean1p0Build._build_infrastructure shares some commonality with Compile._compile_file, not with Dynamo0p3Build - hmm - well, maybe a little bit, catching errors, but I don't see how this could easily be done in one base class (gocean calls popen with build, dynamo calls compile_file in a loop). Maybe a ``print_error`` function or so depending on the exception, but this would likely apply to a lot of other code as well, way beyond the scope of this pull request. So tbh, I'd say for now to just leave it the way it is.

src/psyclone/tests/gocean1p0_transformations_test.py
hiker: Done, #315

src/psyclone/tests/gocean1p0_transformations_test.py
hiker: Now, that's a bit of an issue then: I've just added support for opencl compilation, and assumed the failures were caused by me not having an opencl compiler ... while it is apparently because of #316. Should I now go and remove --compileopencl support etc?? That would be quite a bit of work, which we would have to add back in again when doing #314. I've added a reference to #316 in this one case, but I am not sure what to do with then gocean1p0_opencl_test.py - should we comment out all those tests then?

src/psyclone/tests/test_files/gocean1p0/kernel_stencil.f90
hiker: No, it doesn't - but the kernel declaration contained an additional argument, and the test case needs the fourth parameter. I've added a comment explaining why area is there as a paramter.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -107,206 +98,261 @@ def print_diffs(expected, actual):
...
+                if os.path.isfile(str(name)+"."+suffix):
+                    name += "." + suffix
+                    return name
+        raise IOError("Cannot find a Fortran file '{0}' with suffix in {1}".
+                      format(name, FORTRAN_SUFFIXES))
+
+    def compile_file(self, filename, link=False):
+        ''' Compiles the specified Fortran file into an object file (in
+        the current working directory) using the specified Fortran compiler
+        and flags. Raises a CompileError if the compilation fails.
```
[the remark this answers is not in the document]
hiker: It turns out that compile_file ALWAYS returns true (or raises an exception). So I've removed the return value completely here and in all tests.

hiker:
I got stuck on a train today due to signal failure ... 1.5 hours later I have some unplanned, minor additional improvements, I am just not sure if you want them included here or in a separate request. The first part is strongly related to compilation, and the second part is strongly related to the first part, but not to compilation 😄   Not sure if this is a good enough argument though to include them both here - so you decide.

1) I removed the directory cleanup after compilation. *VERY* often I had to look at the actual files to see what the reason for a test failure is, or compare with a previous run - and always got stuck because the cleanup had removed those files. So I ended up constantly commenting out the cleanup. Note that py.test manages the temporary directories for you, it deletes all but the last three ones when you start it. And I admit I don't understand the comment in the sources that the directory cleanup will allow the same directory to be used several times - even if the files in there you can compile several times(??) - what am I missing? Anyway, I have a patch that removes the cleanup (and also removes the test cases that check if the directories are cleaned up).
2) I replaced all use of tempfiles in tests with tmpdir fixtures (and then also removed the deletion of those files). In case you haven't noticed, there were at least some tempfile tests that leaked files in /tmp ... I had 260 files there 😏  Now everything is under the py.test temporary directory, and automatically cleaned up.

src/psyclone/tests/gocean1p0_transformations_test.py
arporter: I'm happy to leave that flag in because, as you say, it's a lot of work and we will need it in the future.

arporter:
Hi @hiker. I thought that kind of signal failure only happened here - it's kind of reassuring that it's not just a British thing :-) Those changes you describe sound like they fit OK with this PR. TBH, I can't remember quite what I was worrying about when I said that a test could safely re-use a directory. Possibly I was (mistakenly) thinking of a make-type system that would not bother compiling a file if the .o already existed. Then again, maybe it was something else. Either way I'm happy with your proposed solution as it uses pytest functionality and leaves the generated files for debugging purposes.
Is this PR ready for review again or was that just a comment checking whether or not to include these new changes in it?

hiker:
A British thing?? No way, I am sure the Australians invented signal failures!

OK, I've added the temporary file handling improvements. And I need the updated dl_esm_inf (https://github.com/stfc/dl_esm_inf/pull/24) so that all opencl tests will actually work.
*grin* While submodule are really nice, it makes for a lot of work to get an update in: I needed an update to FortCL (to compile in separate directory), which then required an update to dl_esm_inf (to compile fortcl and itself in separate directory), so that I can then finally update psyclone to compile the dl_esm_inf per process. I feel like Marco Polo on my way East ... it's a long trip ;)

arporter:
Thinking about it, maybe we introduced signal failures into Australia :-) It is, as you say, a long trip. I think that's just because we have a lot of dependencies and many of them are in a state of flux. If you can think of any way of improving things then all suggestions are welcome :-)

hiker:
OK, ready for a review. Note that I had several problems with the update, likely because of the nested submodules we now have (could also be because I am using a windows file system mounted on a linux VM - issues with permissions). If a 'git status' aborts with some kind of porcelain errors, try a mixture of `git submodule init`, and `git submodule update --init --recursive`. At a certain stage it worked for me.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -160,27 +152,28 @@ def base_path(self, base_path):
...
 
-    @abc.abstractmethod
     def get_infrastructure_flags(self):
-        '''Returns a list with the required flag to use the required
+        '''Returns a list with the required flags to use the required
         infrastructure library. This is typically ["-I", some_path] so that
         the module files of the infrastructure can be found.
         :returns: A list of strings with the compiler flags required.
-        :rtpe: str
+        :rtpe: list
```
arporter: s/rtpe/rtype/

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -107,206 +98,261 @@ def print_diffs(expected, actual):
...
+        self._tmpdir = tmpdir
+        # pylint: disable=no-member
+        self._f90 = pytest.config.getoption("--f90")
+        self._f90flags = pytest.config.getoption("--f90flags")
+        # pylint: enable=no-member
+        self._base_path = None
+
+    @property
+    def base_path(self):
+        '''Returns the base path of all test files for the API. Needs to
```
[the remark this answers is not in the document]
arporter: Sorry to be a pain but it's still not clear to me what this means. Is it the location of the Fortran infrastructure files that need to be compiled?

src/psyclone/tests/psyclone_test_utils.py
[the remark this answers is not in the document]
arporter: Great.

src/psyclone/tests/gocean1p0_build.py
arporter: Oh yes, good point.

src/psyclone/tests/gocean1p0_build.py
arporter: OK.

src/psyclone/tests/gocean1p0_transformations_test.py
arporter: Thanks. Please could you put the ref to #315 inside the `reason` string of the xfail.

src/psyclone/tests/test_files/gocean1p0/kernel_stencil.f90
arporter: Oh yes; the meta-data says it needs the cell area. Don't know why I didn't see that last time around.

arporter:
Oh yes, I too hit that problem with git - thanks for the tip. Please could you add it to the developers' guide in the relevant section?

src/psyclone/tests/psyclone_test_utils.py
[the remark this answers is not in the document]
arporter: Actually, from a sub-class I think it's the location of the compiled infrastructure files? Please put this in the docstring.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -340,17 +326,12 @@ def string_compiles(self, code):
         with open(filename, 'w') as test_file:
             test_file.write(code)
 
+        success = True
         try:
-            success = self.compile_file(filename)
+            self.compile_file(filename)
```
arporter: Now that you've simplified this, you could simplify it further by just doing `return False` in the `except CompileError` part and `return True` at the end of the routine.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -272,29 +265,22 @@ def _code_compiles(self, psy_ast):
...
                                                str(self._tmpdir)], fort_file)
-                success = self.compile_file(name)
+                self.compile_file(name)
 
             # Finally, we can build the psy file we have generated
-            success = self.compile_file(psy_filename)
+            self.compile_file(psy_filename)
         except CompileError:
             # Failed to compile one of the files
             success = False
```
arporter: Just `return False` and then `return True` below?

arporter: Also, pylint now complains that old_pwd is not used.

src/psyclone/tests/psyclone_test_utils.py
arporter: (and pylint complains that old_pwd is not used)

src/psyclone/tests/dynamo0p3_build.py
```diff
@@ -42,9 +42,11 @@
 
 
 class Dynamo0p3Build(Compile):
-    '''Build class for compiling test files for the Dyanmo0.3 api.
-    using the wrapper library contained in the test_files.
+    '''Build class for compilation of test files for the Dynamo0.3 api.
+    Is uses the wrapper library form test_files/dynamo0p3/infrastructure
```
arporter: s/Is uses the wrapper library form/It uses the wrapper library from/

src/psyclone/tests/psyclone_test_utils.py
hiker: Actually ... the warning correctly indicates that I am not changing back to the original directory. While I don't think this will break anything (afaik any code using a temporary directory will chdir first anyway), but as long as we don't have this as a requirement, I think it's better that I add a finally and chdir back to tmpdir - which then also means keeping the success variable around.

hiker:
OK, ready again ... I hope :)

src/psyclone/tests/psyclone_test_utils.py
arporter: OK, I'm happy with that - it seems the safest thing to do.

arporter:
Thanks for making those changes and updating the documentation.
All comments have been addressed.
Documentation builds OK and looks OK.
No xfailing tests are affected.
Coverage of the diff is NA in this case (only the testing framework is affected).

Issue #317, opened 2019-03-19
alggen does not support functions containing invokes


rupertford:
The new version of alggen being added in PR #285 has not been coded to support Fortran functions containing invokes as I didn't realise this was done. Turns out that this is used in LFRic so  support needs to be added.

The issue can be addressed once #285 makes it onto master.

rupertford:
PR #285 is now on master.

rupertford:
Created branch alg_fp2_func

rupertford:
#330 is now on master. Closing this issue.

PR #330, opened 2019-03-29
issue #317. use can not be added to a function.

rupertford:
This PR adds support for adding use statements to functions. This should lift the restriction of PSyclone not supporting algorithm code with invokes in functions.


rupertford:
Turned out to be a trivial fix (it already worked, I just needed to remove the restriction).
Ready for review from @arporter or @TeranIvy

rupertford:

Issue #319, opened 2019-03-20
Kernel file search should be case insensitive


rupertford:
@TeranIvy has reported the following bug:

Not able to find kernel because of upper case "dA" in the algorithm and lower case "da" in kernel name.
"calc_da_at_w2_kernel_mod.F90" is in gungho/source/kernel directory, just the case "dA" is different.
When "dA" is replaced with "da" in the algorithm the error goes away, however fparser should be case-insensitive.

[ikavcic@eld470 gungho]$ psyclone -d source/kernel/ source/algorithm/runtime_constants_mod.x90
"Parse Error: Kernel file 'calc_dA_at_w2_kernel_mod.[fF]90' not found in /net/home/h06/ikavcic/PSyclone_work/fparser2_work/r17724_test_fparser2/gungho/source/kernel"

rupertford:
Created branch fix_kernel_search

rupertford:
The match is performed by `get_kernel_filepath` in `parse/kernel.py` using `fnmatch.filter(filenames, search_string)`. I think the problem is that this filter is case sensitive (in a case sensitive os). There doesn't appear to be a case insensitive option so one option would be to revert to a regex ...

`options = [filename for filename in filenames if re.search(search_string, filename, re.IGNORECASE)]`

TeranIvy:
Merged #331 so closing the issue.

PR #331, opened 2019-03-31
issue #319. Case insensitive match for kernel file names.

rupertford:
LFRic code was failing with the new version of PSyclone that uses fparser2 to parse algorithm files. One of the errors was due to algorithm files being specified in one case and the actual filename of the file where the kernel code resides is in another case as PSyclone was performing a case sensitive match. Note, I don't understand why this was not also a problem before the change to fparser2 (for algorithms) in PSyclone. The code should be changed to perform a case insensitive match.


rupertford:
Ready for first review from @arporter or @TeranIvy

TeranIvy:
Happy to take so @arporter can focus on refactoring of kernel arguments :-)

Issue #326, opened 2019-03-27
Fix pytest deprecated warnings.


arporter:
Running the test suite with more recent versions of pytest gives (many) deprecation warnings, e.g.:

    psyclone_test_utils.py:115
      /home/me/Projects/GungHo/PSyclone/src/psyclone/tests/psyclone_test_utils.py:115:   PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
    TEST_COMPILE = pytest.config.getoption("--compile")

arporter:
This looks like it may be tricky to solve. The recommended way to get at pytest configuration settings is now via the `request` object. AFAICS, only fixture functions can be provided with this `request` object (https://docs.pytest.org/en/latest/fixture.html#request-context). Therefore, we would need a new test fixture that maybe provides the `Compile` object to a test.

arporter:
e.g. a test would now look like:

    def test_field(GOcean1p0Builder):
        ...
        psy = PSyFactory(...)
        assert GOcean1p0Builder.code_compiles(psy)

where:

    @pytest.fixture
    def GOcean1p0Builder(tmpdir, request):
        test_compile = request.config.getoption("--compile")
        ...
        return GOcean1p0Build(tmpdir)

What do you think @hiker (you know more about fixtures than I do :-) )?

hiker:
I had a look at this, and you are right. This unfortunately means that it is not easy :(

pytest imports the tests before calling a session fixture in conftest. This means that we can't use a command line option to specify the value for COMPILE (since COMPILE will be defined too late). But this is the only supported option to get access to the config option (and therefore to the command line) - so, a bit of a problem :(

ATM it works because when importing a test file, it will import psyclone_test_utilities, which sets @Compile.COMPILE based on pytest.config.

My suggested solution:
we remove the decorator that will cause some test to be skipped, and instead add a test inside the function:
```
@Compile.COMPILE
def test_compiler_works(tmpdir):
```
would become:
```
def test_compiler_works(tmpdir):
   If not Compile.COMPILE:
    pytest.skip("Compilation not enabled")
```
(untested, but it is supposed to work according to the documentation).  Am happy to do this, let me know.

hiker:
[...]

PR #335, opened 2019-04-02
Fix pytest deprecated

hiker:
Fix for #326. Should be all ready to go.


src/psyclone/tests/conftest.py
```diff
@@ -92,10 +80,15 @@ def have_graphviz():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def infra_compile(tmpdir_factory):
-    '''A per-session initialisation function that makes sure that the
-    infrastructure files for the dynamo0p3 and gocean1p0 APIs are compiled.
+def infra_compile(tmpdir_factory, request):
+    '''A per-session initialisation function that sets the compilation flags
+#    in the Compile class based on command line options for --compile,
```
arporter: Why the comment lines here? Please fix...

src/psyclone/tests/conftest.py
```diff
@@ -92,10 +80,15 @@ def have_graphviz():
...
+#    --compileopencl, --f90, --f90flags. Then makes sure that the
+    infrastructure files for the dynamo0p3 and gocean1p0 APIs are compiled
+    (if compilation was enabled).
     '''
+    from psyclone_test_utils import Compile
+    Compile.store_compilation_flags(request.config)
```
arporter: Nice.

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -109,29 +113,42 @@ class Compile(object):
...
+
+    # Class variable to store if opencl compilation is enabled
+    # (--compileopencl).
+    TEST_COMPILE_OPENCL = False
+
+    # Class variable to store the chosen Fortran compiler (--f90, default
+    # gfortran).
+    F90 = "gfortran"
+
+    # Class variable to store the choosen f90 compiler flags (--f90flags).
```
arporter: "chosen"

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -109,29 +113,42 @@ class Compile(object):
...
+    F90FLAGS = ""
+
+    @staticmethod
+    def store_compilation_flags(config):
+        '''This function is called from the conftest session fixture
```
arporter: Should this say "...session fixture infra_compile" or something?

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -109,29 +113,42 @@ class Compile(object):
...
+        Compile.F90 = config.getoption("--f90")
+        Compile.F90FLAGS = config.getoption("--f90flags")
 
     def __init__(self, tmpdir=None):
         self._tmpdir = tmpdir
-        # pylint: disable=no-member
-        self._f90 = pytest.config.getoption("--f90")
-        self._f90flags = pytest.config.getoption("--f90flags")
-        # pylint: enable=no-member
+        # Take the compiler and compile flags from the statis variables.
```
arporter: s/statis/static/

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -109,29 +113,42 @@ class Compile(object):
...
+        # This allows a specific instance to use different compiler options
+        # which is used in some of the compilation tests.
```
arporter: Should we therefore be careful to take copies, e.g.:

    self._f90 = Compile.F90[:]

or is this OK? i.e. we're not just storing a reference back to the static variable are we?

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -163,6 +180,24 @@ def get_infrastructure_flags(self):
         '''
         return []
 
+    @staticmethod
+    def skip_if_compilation_disabled():
+        '''This function is used in all tests that should only run
+        if compilation is enabled. It calls pytest.skip if compilation
+        is not enabled.'''
+        if not Compile.TEST_COMPILE:
+            return pytest.skip("Need --compile option to run")
```
arporter: pylint points out that this probably shouldn't be a `return`?

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -163,6 +180,24 @@ def get_infrastructure_flags(self):
...
+
+    @staticmethod
+    # pylint: disable=invalid-name
```
arporter: pylint also says that this suppression is "useless" :-)

src/psyclone/tests/conftest.py
hiker: No idea ... done :)

src/psyclone/tests/psyclone_test_utils.py
hiker: Done.

src/psyclone/tests/psyclone_test_utils.py
hiker: Yes indeed, forgot to update that comment when I had to combine the two functions. Done

src/psyclone/tests/psyclone_test_utils.py
hiker: Oops

src/psyclone/tests/psyclone_test_utils.py
hiker: Those are just string, and string copy does not create a reference.

src/psyclone/tests/psyclone_test_utils.py
hiker: Correct indeed - now here start the pylint saga, since my pylint did not complain about this, till I updated pylint :)

src/psyclone/tests/psyclone_test_utils.py
hiker: That seems to depend on the pylint version - my previous version was ok with that :)

After the update I now have different pylints for python2 and 3 (1.9.4 and 2.3.1 resp). 

Python3's pylint seems to be ok with the name, but then complains for me about "Class 'Compile' inherits from object, can be safely removed from bases in python3", while python2 is not ok with the name ("doesn't conform to snake_case naming style" - which seems to indicate that the name is too long), but happy with inheritance from object.

Since you don't mention the inheritance from 'object', I've left this unchanged, and hope that you now get a clean pylint output. But it might be worth that we all agree on a certain pylint version, and update it at the same time?

src/psyclone/tests/psyclone_test_utils.py
arporter: Aha!

src/psyclone/tests/psyclone_test_utils.py
arporter: I'm not too worried about the details - e.g. the names thing. The inheritance from 'object' is, I think, a Python 2/3 thing and we will have to live with that until we can drop Python 2. In fact we could probably suppress that warning altogether but we can ignore it for now.

arporter:
All comments/requests have been addressed.
None of the remaining review checklist items are relevant as this change only affects the mechanics of the test suite.

arporter:
I will merge this once the final Travis checks have run.

PR #341, opened 2019-04-04
337 fix match error in the algorithm layer

rupertford:
The algorithm parsing code ('parse/algorithm.py') was not matching kernel/builtin names with declared 'use' names if their case was different. This PR fixes this error.


rupertford:
Ready for review by @arporter or @TeranIvy

PR #347, opened 2019-04-09
Issue #256 (partially). fparser2ASTProcessor support for If Constructs

sergisiso:
This is already implemented on the Nemo API, the intent of this PR is to create a generalized version. I also propose the following changes.

In the Nemo API there is the following class hierarchy: `Node <- IfBlock <- IfClause`
where `IfBlock` has a self._condition with the condition expression 'out-of-tree' and the self._children are the contents of the if body. `IfClause` is the same for 'else' and 'else if' with a self._clause_type to differentiate between them.

I propose to use just 1 sub-class (eliminate IfClause) and have all content on the tree. For this, the IfBlock
will need 2 mandatory children: 1) Condition, 2) If-body and a third optional child 3) else-body. The if-body and else-body are always Schedule nodes, so it can contain multi-statement chunks of code.

'else if' statements can be encoded as as nested ifBlocks, which it is functionally the same. In order to recreate the original code I propose to add a `self.hint_was_elseif = True/False` attribute. If an else-body contains a Schedule with just one IfBlock with hint_was_elseif == True, then the gen_* routine can write it as and 'else if' statement.


sergisiso:
This PR is ready for the first review, I made some changes from the PR initial description:
- Instead of using hint_* attributes, I use a single annotations attribute, which contains string keyword tags.
- I included a `Case_Construct_handler` that manages simple `SELECT CASE` constructs (no ranges or DEFAULT for now).
- I already implemented the gen_c_code(), but I decided to ignore the annotations (I believe it is not so important to keep the syntax when going from language to language, and probably I would end up modifying for performance latter on when looking to optimize OpenCL).

sergisiso:
@arporter Also, because I removed the `NemoIf*` classes in favor of the `IfBlock` which has a different tree structure, even if it passes the unit-test and the ./check_examples generation, it would be good to double check that the output of the nemo examples still makes sense.

src/psyclone/psyGen.py
```diff
@@ -5356,13 +5538,112 @@ def _if_stmt_handler(self, node, parent):
...
+                selector = child.items[0]
+            if isinstance(child, (Fortran2003.Case_Stmt,
+                                  Fortran2003.End_Select_Stmt)):
+                clause_indices.append(idx)
+                # Case Default and value Ranges not supported yet
+                if 'DEFAULT' in str(child):
+                    raise NotImplementedError("Case Default Statement")
+                if ':' in str(child):
+                    raise NotImplementedError("Case Value Range Statement")
+
```
arporter: Please could you add an `else` here and raise an InternalError that we've found a child of a Select_Case_Stmt that we weren't expecting.

src/psyclone/psyGen.py
```diff
@@ -5356,13 +5538,112 @@ def _if_stmt_handler(self, node, parent):
...
+                        currentparent.addchild(elsebody)
+                        elsebody.addchild(ifblock)
+                        # Keep pointer to fpaser2 AST
+                        elsebody._ast = node.content[start_idx]
+                        ifblock._ast = node.content[start_idx]
+                        elsebody._ast_end = node.content[end_idx]
+                        ifblock._ast_end = node.content[end_idx]
+                    else:
+                        rootif = ifblock
+                        rootif._ast = node  # Keep pointer to fpaser2 AST
```
arporter: Is this correct? It doesn't seem to agree with how you set `ifblock._ast` just above. Also, please could you set `rootif._ast_end`. Can this setting of `ifblock._ast/_ast_end` be done outside the `if rootif` block?

src/psyclone/psyGen.py
```diff
@@ -6079,6 +6360,15 @@ def __init__(self, reference_name, parent):
         super(Reference, self).__init__(parent=parent)
         self._reference = reference_name
 
+    @property
+    def ref_name(self):
```
arporter: How about just calling this `name` or is there some reason why that's not a good idea?

src/psyclone/tests/psyGen_test.py
```diff
@@ -2716,8 +2716,124 @@ def test_codeblock_gen_c_code():
...
+    assert "If[]\n" in str(ifblock)
+    assert "condition1" in str(ifblock)  # Test condition is printed
+    assert "Return[]" in str(ifblock)  # Test if_body is printed
+
+
+def test_ifblock_properties():
+    '''Test that an IfBlock node properties can be retrieved'''
+    ifblock = IfBlock()
+
+    # Condition can't be retrieved before is added as a children.
```
arporter: /children/child/

src/psyclone/tests/psyGen_test.py
```diff
@@ -2716,8 +2716,124 @@ def test_codeblock_gen_c_code():
...
+    with pytest.raises(InternalError) as err:
+        _ = ifblock.condition
+    assert("IfBlock malformed or incomplete. It should have "
+           "at least 1 children, but found 0." in str(err.value))
+
+    ref1 = Reference('condition1', parent=ifblock)
+    ifblock.addchild(ref1)
+
+    # If_body can't be retrieved before is added as a children.
```
arporter: /children/child/

src/psyclone/tests/psyGen_test.py
```diff
@@ -2716,8 +2716,124 @@ def test_codeblock_gen_c_code():
...
+    assert "if (condition1) {\n" in ifblock.gen_c_code()
+    assert "return;\n" in ifblock.gen_c_code()
+    assert "}" in ifblock.gen_c_code()
+    assert "else" not in ifblock.gen_c_code()
+
+    sch = Schedule(parent=ifblock)
+    ifblock.addchild(sch)
+    ret = Return(parent=sch)
+    sch.addchild(ret)
+    assert "else" in ifblock.gen_c_code()
```
arporter: Please could you make this last check a little more thorough to check that the correct form is being produced.

arporter: We may also want to start thinking about adding support for compiling generated C code as part of the test suite...please could you create a new Issue for this.

src/psyclone/tests/psyGen_test.py
```diff
@@ -3792,6 +3908,245 @@ def test_fparser2astprocessor_handling_if_stmt(f2008_parser):
...
+            branch3 = 1
+        endif''')
+    fparser2if_construct = Execution_Part.match(reader)[0][0]
+
+    fake_parent = Node()
+    processor = Fparser2ASTProcessor()
+    processor.process_nodes(fake_parent, [fparser2if_construct], None)
+
+    # Check a new node was properly generated and connected to parent
+    assert len(fake_parent.children) == 1
```
arporter: Could you add checks that the various `._ast` and `._ast_end` are set correctly here too?

src/psyclone/transformations.py
```diff
@@ -120,9 +121,9 @@ def _validate(self, node_list):
 
         # Check that we aren't attempting to include any else/else if's without
         # their parent 'if'.
-        from psyclone.psyGen import IfClause
+        from psyclone.psyGen import IfBlock
         for node in node_list:
-            if isinstance(node, IfClause):
+            if isinstance(node, Schedule) and isinstance(node.parent, IfBlock):
```
arporter: I know we discussed this and we have tests that check that it works but I don't understand it! Please could you add a comment explaining. Is it that, for the list of nodes to be valid, they must have a `Schedule` as their parent?

sergisiso:
@arporter The `_ast_end` attribute is just assigned and never used (besides in one test). Do you think it is something we could remove?, or do you expect it to be needed in the future?

```
laptop PSyclone (ExtendIfBlocks) $ grep -r "_ast_end" src/
src/psyclone/psyGen.py:        self._ast_end = None
src/psyclone/psyGen.py:        return self._ast_end
src/psyclone/psyGen.py:            self._ast_end = directive
src/psyclone/psyGen.py:        self._ast_end = enddir
src/psyclone/psyGen.py:        self._ast_end = enddir
src/psyclone/psyGen.py:                    elsebody._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:                    newifblock._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:                ifbody._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:                elsebody._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:                    ifbody._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:                        elsebody._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:                        ifblock._ast_end = node.content[end_idx]
src/psyclone/psyGen.py:            self._ast_end = self._statements[-1]
src/psyclone/psyGen.py:            self._ast_end = None
src/psyclone/tests/nemo/transformations/openacc/data_directive_test.py:    assert isinstance(datadir._ast_end, Fortran2003.Comment)
src/psyclone/tests/nemo/transformations/openacc/data_directive_test.py:    assert str(datadir._ast_end).lower() == "!$acc end data"
```

sergisiso:
Oh, I forget about the ast_end property, although it is used only twice in contiguous lines, I should explore this further.

arporter:
Hi @sergisiso, `_ast_end` is used by some of the transformations to ensure that a closing directive gets put in the right place. I suppose it's really a workaround for the non-intuitive parse tree that fparser produces.

doc/developers.rst
```diff
@@ -303,6 +303,25 @@ The Symbol Table has the following interface:
 .. autoclass:: psyclone.psyGen.SymbolTable
     :members:
 
+Control Flow Nodes
+==================
+
+PSyIR has two control flow nodes: IfBlock and Loop. This nodes represent
```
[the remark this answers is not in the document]
sergisiso: fixed

doc/developers.rst
```diff
@@ -303,6 +303,25 @@ The Symbol Table has the following interface:
...
+the canonical structure to build conditional branching constructs and
```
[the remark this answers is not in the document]
sergisiso: Changed to: "structure with which conditional branching constructs and iteration constructs are build."

doc/developers.rst
```diff
@@ -303,6 +303,25 @@ The Symbol Table has the following interface:
...
+iteration constructs. Additional language-specific syntax for branching
+and iteration will be normalized to use these same constructs.
+For example, Fortran has the additional branching constructs `ELSE IF`
+and `CASE`, when a Fortran code is translated to PSyIR, Psyclone will
```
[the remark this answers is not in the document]
sergisiso: done

doc/developers.rst
```diff
@@ -303,6 +303,25 @@ The Symbol Table has the following interface:
...
+build a semantic equivalent implementation using IfBlocks.
```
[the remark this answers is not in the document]
sergisiso: done

doc/developers.rst
```diff
@@ -303,6 +303,25 @@ The Symbol Table has the following interface:
...
+However, the necessary nodes in the new tree structure will be annotated
+with information to recreate the language-specific syntax.
```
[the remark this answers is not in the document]
sergisiso: done

src/psyclone/psyGen.py
```diff
@@ -4657,17 +4668,72 @@ def apply(self):
 
 class IfBlock(Node):
     '''
-    Class representing an if-block within the PSyIR.
+    Class representing an if-block within the PSyIR. It has two mandatory
+    children: the first one represented the if-condition and the second one
```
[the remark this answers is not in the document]
sergisiso: done

src/psyclone/psyGen.py
```diff
@@ -4657,17 +4668,72 @@ def apply(self):
...
+    the if-body; and an optional third child representing the else-body.
 
     :param parent: the parent of this node within the PSyIR tree.
     :type parent: :py:class:`psyclone.psyGen.Node`
+    :param str annotation: Tags that provide additional information about \
+        the node. The node should still be functionally correct when \
+        ignoring this tags.
```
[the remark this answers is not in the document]
sergisiso: done

src/psyclone/psyGen.py
```diff
@@ -4657,17 +4668,72 @@ def apply(self):
...
+    :raises InternalError: when initialised with invalid parameters.
     '''
-    def __init__(self, parent=None):
+    valid_annotations = ('was_elseif', 'was_single_stmt', 'was_case')
```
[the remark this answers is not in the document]
sergisiso: Added a description of each element in the annotation docstring so we also have this on the developer's guide.

src/psyclone/psyGen.py
```diff
@@ -4657,17 +4668,72 @@ def apply(self):
...
+        of this IfBlock.
+
+        :return: IfBlock conditional expression.
+        :rtype: :py:class:`psyclone.psyGen.Node`
+        :raises InternalError: If the condition is not attached to the tree.
+        '''
+        if len(self.children) < 1:
+            raise InternalError(
+                "IfBlock malformed or incomplete. It should have at least 1 "
+                "children, but found {0}.".format(len(self.children)))
```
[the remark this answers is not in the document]
sergisiso: Updated to 2, I had 1 because you can still get the condition of a partially malformed IF (just one child) but it is probably more robust to always check it is fully formed.

src/psyclone/psyGen.py
```diff
@@ -4657,17 +4668,72 @@ def apply(self):
...
-    def __str__(self):
-        return "If-block: "+self._condition
+    @property
+    def condition(self):
+        ''' Return the PSyIR Node representing the conditional expression
+        of this IfBlock.
+
+        :return: IfBlock conditional expression.
+        :rtype: :py:class:`psyclone.psyGen.Node`
+        :raises InternalError: If the condition is not attached to the tree.
```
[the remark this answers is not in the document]
sergisiso: I like the suggestion, and I can use the same docstring in the if_body property :)

src/psyclone/psyGen.py
```diff
@@ -4686,34 +4752,48 @@ def view(self, indent=0):
...
-        Return text containing the (coloured) name of this node type.
+        Generate a string representation of this node using C language.
 
-        :returns: the name of this node type, possibly with control codes \
-                  for colour.
+        :param int indent: Depth of indent for the output string.
+        :return: C language code representing the node.
         :rtype: str
+        :raises GenerationError: If any mandatory children of the IfBlock \
+            node is missing.
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -4686,34 +4752,48 @@ def view(self, indent=0):
...
         '''
-        Return text containing the (coloured) name of this node type.
+        Generate a string representation of this node using C language.
 
-        :returns: the name of this node type, possibly with control codes \
-                  for colour.
+        :param int indent: Depth of indent for the output string.
+        :return: C language code representing the node.
         :rtype: str
+        :raises GenerationError: If any mandatory children of the IfBlock \
```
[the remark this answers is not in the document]
sergisiso: Agreed

src/psyclone/psyGen.py
```diff
@@ -5345,6 +5426,107 @@ def _ignore_handler(self, *_):
...
+        for idx in range(num_clauses):
+            start_idx = clause_indices[idx]
+            end_idx = clause_indices[idx+1]
+            clause = node.content[start_idx]
+
+            if isinstance(clause, (Fortran2003.If_Then_Stmt,
+                                   Fortran2003.Else_If_Stmt)):
+                # If its an 'IF' clause just create an IfBlock, otherwise
+                # it is an 'ELSE' clause and it needs an IfBlock annotated
+                # with 'was_elseif' inside an Schedule.
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -5356,13 +5538,112 @@ def _if_stmt_handler(self, node, parent):
...
+        for idx, child in enumerate(node.content):
+            child._parent = node  # Retrofit parent info
+            if isinstance(child, Fortran2003.Select_Case_Stmt):
+                selector = child.items[0]
+            if isinstance(child, (Fortran2003.Case_Stmt,
+                                  Fortran2003.End_Select_Stmt)):
+                clause_indices.append(idx)
+                # Case Default and value Ranges not supported yet
+                if 'DEFAULT' in str(child):
+                    raise NotImplementedError("Case Default Statement")
```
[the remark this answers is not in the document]
sergisiso: Note that handler's NotImplementedErrors are always catch by process_node() method which generates a CodeBlock, it is not an Error Message that will raise to the user.
Would you still want more verbose messages in this case?

src/psyclone/psyGen.py
sergisiso: `else` is valid case, I reordered the code with the `clause_indices.append(idx)` inside the else to make this easier to read.

src/psyclone/psyGen.py
sergisiso: Yep,  `name` is better.

src/psyclone/tests/psyGen_test.py
sergisiso: fixed

src/psyclone/tests/psyGen_test.py
sergisiso: fixed

src/psyclone/tests/psyGen_test.py
sergisiso: Now full if-else expression is checked.

sergisiso: Created issue #372

src/psyclone/psyGen.py
```diff
@@ -5345,6 +5426,107 @@ def _ignore_handler(self, *_):
...
+        currentparent = parent
+        num_clauses = len(clause_indices) - 1
+        for idx in range(num_clauses):
+            start_idx = clause_indices[idx]
+            end_idx = clause_indices[idx+1]
+            clause = node.content[start_idx]
+
+            if isinstance(clause, (Fortran2003.If_Then_Stmt,
+                                   Fortran2003.Else_If_Stmt)):
+                # If its an 'IF' clause just create an IfBlock, otherwise
```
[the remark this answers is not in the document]

src/psyclone/transformations.py
sergisiso: I deleted this altogether, I think the original code was just trying to avoid transformations to IfClause but was allowing transformation to the body of the clause, now the IfClause doesn't exist anymore.

src/psyclone/psyGen.py
sergisiso: I updated this code appropriately, now `ifblock._ast/_ast_end` is common

src/psyclone/tests/psyGen_test.py
sergisiso: Added some `_ast/_ast_end` checks on `test_fparser2astprocessor_handling_if_construct`

sergisiso: Also added checks in `test_fparser2astprocessor_handling_Case_construct`

sergisiso:
@arporter It is ready for next review, I fixed the minor typos and added better checks for the fparser2 ast pointers.

doc/developers.rst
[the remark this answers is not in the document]
arporter: Almost! You need "...are built."

doc/developers.rst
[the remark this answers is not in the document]
arporter: Almost :-) Please could you put a colon after `CASE`.

src/psyclone/psyGen.py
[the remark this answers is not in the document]
arporter: Great!

src/psyclone/psyGen.py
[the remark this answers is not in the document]
arporter: You've changed the description in the docstring but the code still actually raises a `GenerationError` :-)

src/psyclone/psyGen.py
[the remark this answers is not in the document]
arporter: Ah, OK. No, it's fine as it is in that case (but perhaps add a comment saying that NotImplementedErrors are caught by process_node()).

src/psyclone/psyGen.py
arporter: Thanks although what I meant was we should protect ourselves against the `child` *not* being an instance of either `Case_Stmt` or `End_Select_Stmt`. Of course this "should not" happen but you never know...

Also, now I look at this again, is checking that str(child) contains 'DEFAULT' robust enough. e.g. do you need to worry about upper/lowercase and what if someone had a parameter named `default` and thus did `case(default)`?

src/psyclone/psyGen.py
arporter: Great, thanks.

src/psyclone/tests/psyGen_test.py
arporter: Great, thanks.

src/psyclone/transformations.py
arporter: Great.

src/psyclone/psyGen.py
```diff
@@ -5364,13 +5556,110 @@ def _if_stmt_handler(self, node, parent):
...
+            if isinstance(child, Fortran2003.Select_Case_Stmt):
+                selector = child.items[0]
+            if isinstance(child, (Fortran2003.Case_Stmt,
+                                  Fortran2003.End_Select_Stmt)):
+                # Case Default and value Ranges not supported yet
+                if 'DEFAULT' in str(child):
+                    raise NotImplementedError("Case Default Statement")
+                elif ':' in str(child):
+                    raise NotImplementedError("Case Value Range Statement")
+                else:
```
arporter: pylint says this `else:` is unnecessary (because it follows a `raise`)

src/psyclone/transformations.py
```diff
@@ -111,22 +111,14 @@ def _validate(self, node_list):
 
         # Check that the proposed region contains only supported node types
         for child in node_list:
-            flat_list = [child] + child.walk(child.children, object)
+            flat_list = [item for item in child.walk([child], object)
+                         if type(item) != Schedule]
```
arporter: pylint says you should use isinstance instead of type()

doc/developers.rst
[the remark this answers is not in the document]
sergisiso: done

doc/developers.rst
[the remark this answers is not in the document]
sergisiso: Sure :)

src/psyclone/psyGen.py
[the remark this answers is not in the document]
sergisiso: Ups, fixed now.

src/psyclone/psyGen.py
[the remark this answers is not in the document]
sergisiso: added comment

src/psyclone/psyGen.py
sergisiso: I am not sure about your first point, we are searching select_case_stmt, case_stmt and end_slelect_stmt to record the indices and selector name, everything else (like the case body statements) should just pass to the next iteration without raising an error.

Your second concern is right, I will convert the str(child) to lower now.

src/psyclone/nemo.py
```diff
@@ -709,138 +707,3 @@ def match(node):
                     return False
         return True
```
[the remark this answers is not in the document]
sergisiso: deleted

src/psyclone/nemo.py
```diff
@@ -43,7 +43,7 @@
 from __future__ import print_function, absolute_import
 import copy
 from psyclone.psyGen import PSy, Invokes, Invoke, InvokeSchedule, Node, \
-    Loop, Kern, InternalError, IfBlock, IfClause, NameSpaceFactory, \
+    Loop, Kern, InternalError, IfBlock, NameSpaceFactory, \
```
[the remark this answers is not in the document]
sergisiso: deleted

src/psyclone/psyGen.py
```diff
@@ -4657,17 +4672,79 @@ def apply(self):
...
+    def else_body(self):
+        ''' Return children of the Schedule executed when the IfBlock
+        evaluates to False.
+
+        :return: Statements to be executed when IfBlock evaluates to False.
+        :rtype: list of :py:class:`psyclone.psyGen.Node`
+        '''
+        if len(self._children) == 3:
+            return self._children[2]._children
+        else:
```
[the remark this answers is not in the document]
sergisiso: deleted

src/psyclone/psyGen.py
sergisiso: deleted

src/psyclone/transformations.py
sergisiso: done

src/psyclone/psyGen.py
```diff
@@ -5353,6 +5445,106 @@ def _ignore_handler(self, *_):
...
+
+            if isinstance(clause, (Fortran2003.If_Then_Stmt,
+                                   Fortran2003.Else_If_Stmt)):
+                # If it's an 'IF' clause just create an IfBlock, otherwise
+                # it is an 'ELSE' clause and it needs an IfBlock annotated
+                # with 'was_elseif' inside a Schedule.
+                newifblock = None
+                if isinstance(clause, Fortran2003.If_Then_Stmt):
+                    ifblock = IfBlock(parent=currentparent)
+                    ifblock._ast = node  # Keep pointer to fpaser2 AST
```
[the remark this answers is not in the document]
sergisiso: I created Node setters for `ast/ast_end` instead, as otherwise I would need to modify multiple constructors and I would like to keep the PSyIR code as uncoupled from fparser as possible.

sergisiso:
@arporter I addressed your comments and pylint issues. It is ready for next review.

src/psyclone/psyGen.py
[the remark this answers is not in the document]

src/psyclone/psyGen.py
arporter: Ah, I see - I'd not fully understood the code. Thanks for adding the case insensitivity. Is this now robust to someone having a variable named 'default' and using that in a `case` statement, e.g.:

    integer, parameter :: default = 4
    ...
    select case(my_var)
        case(0)
            call do_this()
        case(default)
            call do_that()
        default
            call error()
    end select

src/psyclone/psyGen.py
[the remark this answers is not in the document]
arporter: OK, good idea.

arporter:
Everything is now done apart from the check for the 'default' clause of a `select case` - I think it needs to be made more robust. It might be as simple as doing `strip(str(...)) == "default"` but I'll let you investigate.

src/psyclone/psyGen.py
sergisiso: The default alone is not syntactically correct and produces a compile-time error (or fparser2 syntax error).  I tried the other cases `case (default)`, `case (DEFAULT)`, `case default`, `case DEFAULT` and all seem to work with the current implementation.

sergisiso: Last comment is wrong, the parenthesis are meaningful and `CASE (default)` should be comparing the selector with a symbol named 'default'.

sergisiso: I fixed it by using fparser2 child type comparisons instead of string comparisons, it should work properly and be more robust now.

sergisiso:
@arporter Case select should be more robust now and I also brought it to master. It is ready for next review.

arporter:
New code is much more robust and a test has been added for the case (no pun intended) I was concerned about. I've taken the liberty of fixing one minor pycodestyle error that had been introduced.

arporter:
Once Travis has passed, I will merge this to master.

Issue #323, opened 2019-03-21
Capture symbols from use stmts in the SymbolTable


arporter:
Capture any symbols accessed via module use statements (in kernels) in the SymbolTable. This is a first step towards being able to convert those symbols that are variables into kernel arguments instead.

arporter:
I've begun adding code to handle any USE statements in the parse tree. For the moment I think it would be best to postpone attempting to resolve any symbols brought in via USE statements until we know we need to do so. That way we only parse additional code if we need to.

arporter:
In #347 @sergisiso is adding support for annotations (via a new `Node._annotations` list of `str`). Although we'll need annotations here I think these will live in the `SymbolTable` and will hold the name of the module from which the symbol is imported.

arporter:
Just realised that my new scope name of `global_use` is Fortran specific :-( Maybe it should just be `global`?

sergisiso:
#366 added support for symbols with `use module, only: xxx` statements. I keep this issue open as other symbols inside modules are not resolved yet.

sergisiso:
#518 is on master, it is now possible to extend the #366 to `use` statements without `only:`. This is different than #190 where `use` (with or without `only`) are found in a kernel and the generated kernel /programming modes does not support `use` statements. However, both solutions need to parse the `use`.

rupertford:
@arporter, can this issue be closed now?

arporter:
Yes, I think so.

PR #366, opened 2019-04-26
Add symbols from module use statements to the SymbolTable (for #323)

arporter:
This is a small PR that simply puts any symbols accessed via `use xxx, only yyy` statements into the SymbolTable. Rather than attempt to work out what type these symbols are, we mark them as having 'deferred' type. If a Kernel is the target of a transformation then any such symbols must be resolved at that point but that will be the subject of another PR.


arporter:
Ready for review from either @rupertford or @sergisiso. It might be good for all three of us to discuss whether we're happy with the changes in this PR before doing a full review.

arporter:
We've had our discussion and agreed:
  1. Since `annotation` implies something that is not strictly required, we've agreed to have `interface`;
  2. Instead of `global_argument` and `global` we'll have just `global` with `interface` specifying whether the symbol is an argument or has some other interface;
  3. If we don't know how a symbol is used then we should leave it as `inout`.
  4. We need a slight change of wording regarding the meaning of `is_input`/`is_output` - these should capture how data is used within the code under consideration, not the lifetime of the associated data.

arporter:
Test failures are because I haven't updated `setup.py` with the new Enum dependence. I haven't done that because Joerg is already doing that in #371.

arporter:
@sergisiso Would you mind reviewing this as, unfortunately,  it's a bit of a change to the SymbolTable interface? There are two missed lines at the mo and I'll add tests for those shortly.

sergisiso:
@arporter  Note that PR #374 from @rupertford is almost ready and touches some of the same Symbol and SymbolTable lines that you are updating here. Basically it goes from
`symbol_table.declare("name",[], scope=..,is_input=..., is_output=...)` to `symbol_table.declare(Symbol("name",[], scope=..,is_input=..., is_output=...))`
and you are then modifying `scope=..,is_input=..., is_output=...` to `interface=...`. The changes are simple but I expect git could have some issues merging.

src/psyclone/psyGen.py
```diff
@@ -5290,6 +5291,24 @@ def iterateitems(nodes):
                 return nodes.items
             return [nodes]
 
+        # Look at any USE statments
+        for decl in walk_ast(nodes, [Fortran2003.Use_Stmt]):
+
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -5832,12 +5845,129 @@ def _number_handler(self, node, parent):
...
+
 class Symbol(object):
     '''
     Symbol item for the Symbol Table. It contains information about: the name,
-    the datatype, the shape (in column-major order), the scope and for
-    global-scoped symbols whether the data is already defined and/or survives
-    after the kernel.
+    the datatype, the shape (in column-major order), the scope and, for
+    global-scoped symbols, the interface to those symbols (i.e. the mechanism
```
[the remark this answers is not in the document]
arporter: I know that feeling when you look at something you previously wrote :-) Updated.

src/psyclone/psyGen.py
```diff
@@ -5848,30 +5978,26 @@ class Symbol(object):
...
-                      by argument.
-    :param bool is_input: Whether the symbol represents data that exists \
-                          before the kernel is entered and that is passed \
-                          into the kernel.
-    :param bool is_output: Whether the symbol represents data that is passed \
-                           outside the kernel upon exit.
+    :param interface: Object describing the interface to this symbol (i.e. \
+                      whether it is passed as a routine argument or accessed \
+                      in some other way) or None if the symbol is local.
+    :type interface: :py:class:`psyclone.psyGen.SymbolInterface`
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -5848,30 +5978,26 @@ class Symbol(object):
...
-    valid_scope_types = ('local', 'global_argument')
     # Tuple with the valid datatypes.
-    valid_data_types = ('real', 'integer', 'character', 'boolean')
+    valid_data_types = ('real',  # Floating point
+                        'integer',
+                        'character',
+                        'boolean',
+                        'deferred'  # Type of this symbol has not yet
+                                    # been determined
+    )
```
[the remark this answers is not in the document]
arporter: Oh yes. Fixed.

src/psyclone/psyGen.py
```diff
@@ -5989,34 +6069,32 @@ def shape(self):
...
-        if new_scope not in Symbol.valid_scope_types:
-            raise ValueError("Symbol scope attribute can only be one of {0}"
-                             " but got '{1}'."
-                             "".format(str(Symbol.valid_scope_types),
-                                       str(new_scope)))
+        return self._interface
 
-        self._scope = new_scope
+    @interface.setter
+    def interface(self, value):
```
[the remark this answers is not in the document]
arporter: Oops, added.

src/psyclone/psyGen.py
```diff
@@ -6070,6 +6148,8 @@ def __str__(self):
             ret = ret[:-2] + "]"  # Deletes last ", " and adds "]"
         else:
             ret += "Scalar"
+        if self.interface:
+            ret += ", " + str(self.interface)
```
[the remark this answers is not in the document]
arporter: Oh yes, that's nicer.

src/psyclone/psyGen.py
```diff
@@ -5832,12 +5845,129 @@ def _number_handler(self, node, parent):
...
+        :raises TypeError: if the supplied value is not of the correct type.
+        '''
+        if not isinstance(value, SymbolAccess):
+            raise TypeError("SymbolInterface.access must be a SymbolAccess "
+                            "but got {0}".format(type(value)))
+        self._access = value
+
+    def __str__(self):
+        if self.is_argument:
+            return "Argument={0}".format(self.is_argument)
```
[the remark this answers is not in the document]
arporter: Good point. I didn't want to lose the "Argument" information though so I've changed it to output Argument=True/False as appropriate.

src/psyclone/psyGen.py
```diff
@@ -6108,46 +6187,43 @@ def declare(self, name, datatype, shape=[], scope='local',
...
             if symbol.scope == 'local':
-                symbol.scope = 'global_argument'
-                symbol.is_input = True
-                symbol.is_output = True
+                # TODO For discussion with the reviewer. The PSyIR should be
+                # language independent but obviously
+                # the line below assumes we're dealing with Fortran. Should
+                # we have some annotation somewhere that says which language
+                # we're coming from?
+                # A Fortran argument has intent(inout) by default
```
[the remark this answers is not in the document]
arporter: Oh yes, that makes sense. I've altered the interface to take a SymbolInterface object. If one isn't supplied then we create one with an access of UNKNOWN. This allows the Fortran-specific front-end to specify that the default is actually READWRITE.

On a separate note, I think I've just discovered a problem with `specify_argument_list()` - it is written (and tested) so as to be called more than once. However, each time it is called it resets `self._argument_list` to [] before constructing it afresh. Therefore, if it is called more than once, e.g.:

    symbol_table.specify_argument_list(["var1"])
    symbol_table.specify_argument_list(["var2"])

then `symbol_table._argument_list` will only contain `var2` but the Symbol for `var1` remains modified to say that it is a routine argument. I think the fix for this would be to make sure that it doesn't reset `self._argument_list` and maybe call it something like `specify_arguments`/`mark_arguments` to try and make it clear that its operation is cumulative.

src/psyclone/psyGen.py
```diff
@@ -5832,12 +5845,129 @@ def _number_handler(self, node, parent):
...
+    READ = 1
+    # The first access of the symbol in the scoping block is a write and
+    # therefore any value that it may have had upon entry is discarded.
+    WRITE = 2
+    # The first access of the symbol in the scoping block is a read but
+    # it is subsequently written to.
+    READWRITE = 3
+    # The way in which the symbol is accessed in the scoping block is
+    # unknown
+    UNKNOWN = 4
```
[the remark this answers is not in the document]
arporter: Oh yes, having it nested inside `Symbol` seems like a good idea (if that's a Pythonic thing to do?). Yep, happy to chat about READ/WRITE IN/OUT choices :-)

src/psyclone/psyGen.py
```diff
@@ -5832,12 +5845,129 @@ def _number_handler(self, node, parent):
...
+        Setter for the access type of this symbol.
+
+        :param value: the new access type.
+        :type value: :py:class:`psyclon.psyGen.SymbolAccess`
+
+        :raises TypeError: if the supplied value is not of the correct type.
+        '''
+        if not isinstance(value, SymbolAccess):
+            raise TypeError("SymbolInterface.access must be a SymbolAccess "
+                            "but got {0}".format(type(value)))
```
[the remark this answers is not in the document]

arporter:
I think I've addressed all comments (or at least attempted to respond to them). Ready for 2nd review or for discussion tomorrow...

src/psyclone/psyGen.py
```diff
@@ -5832,12 +5861,127 @@ def _number_handler(self, node, parent):
...
+            return super(FortranInterface, self).__str__()
+
+    @property
+    def module_name(self):
+        '''
+        :returns: the name of the Fortran module from which the symbol is \
+                  imported or None if it is not a module variable.
+        :rtype: str or None
+        '''
+        return self._module_name
```
sergisiso: With the current `Interface` and `FortranInterface` is not very clear when you read the user code that we are creating arguments or globals. At the moment we have:
``` 
Symbol('a', 'real', [], interface=FortranInterface(access=...))
Symbol('b', 'real', [], interface=FortranInterface(access=..., module_use=...))
```
So we just know that `a` is an Argument because it is the default of FortranInterface and `b` is a global because it has the `module_use` argument.  I think a more readable alternative would be to have `Interface` as an Abstract base class with the `is_argument` and `is_global` methods defined, and then implement Argument and FortranGlobal as subclasses. Also I believe the access parameter is not needed for FortranGlobals as they are always READWRITE, right? An then optionally this can also be a nested (just has meaning for Symbol). With all that we could get some user interface like this:
``` 
Symbol('a', 'real', [], interface=Symbol.Argument(access=...))
Symbol('b', 'real', [], interface=Symbol.FortranGlobal(module_use=...))
```
What do you think?

arporter: I think that's a good idea and have re-factored things accordingly.

arporter:
Ready for next review now once Travis has reported. I don't know why GitHub is saying the branch is out-of-date - I've just tried merging with the latest master on the command line only to be told that the branch is already up-to-date.

rupertford:
> Ready for next review now once Travis has reported. I don't know why GitHub is saying the branch is out-of-date - I've just tried merging with the latest master on the command line only to be told that the branch is already up-to-date.

I've just merged #380 onto master which is probably why it's saying out of date.

src/psyclone/psyGen.py
```diff
@@ -6060,8 +6196,7 @@ def __init__(self, kernel=None):
         # Reference to KernelSchedule to which this symbol table belongs.
         self._kernel = kernel
 
-    def declare(self, name, datatype, shape=[], scope='local',
-                is_input=False, is_output=False):
+    def declare(self, name, datatype, shape=[], interface=None):
```
sergisiso: Not your fault but pylint says: "[dangerous-default-value] Dangerous default value [] as argument"
We can use `shape=None` and then give the empty list if it is None.

src/psyclone/psyGen.py
```diff
@@ -6075,46 +6210,53 @@ def declare(self, name, datatype, shape=[], scope='local',
...
-                          information about the sharing mechanism, at the \
-                          moment just 'global_argument' is available for \
-                          variables passed in/out of the kernel by argument.
-        :param bool is_input: Whether the symbol represents data that exists \
-                              before the kernel is entered and that is passed \
-                              into the kernel.
-        :param bool is_output: Whether the symbol represents data that is \
-                               survives outside the kernel upon exit.
-        :raises KeyError: The provided name can not be used as key in the
+        :param interface: If not None then indicates that the symbol exists \
```
sergisiso: s/symbol/data/ ?

src/psyclone/psyGen.py
```diff
@@ -6075,46 +6210,53 @@ def declare(self, name, datatype, shape=[], scope='local',
...
+        :param interface: Interface describing how the supplied arguments \
+                          are accessed in the kernel or None.
+        :type interface: :py:class:`SymbolInterface` or NoneType.
+
+        :raises InternalError: if interface is supplied but is not a \
+                               SymbolInterface.
+
         '''
+        if interface and not isinstance(interface, SymbolInterface):
+            raise InternalError(
```
sergisiso: Would `TypeError` fit better?

src/psyclone/psyGen.py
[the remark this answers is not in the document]
sergisiso: You are right, when a symbol leaves the argument list  the interface should be coherent with it.

Also, the `interface` argument is just used when arguments symbols do not have an interface yet, so some are defined in the fparser2ASTProcessor and some here inside the SymbolTable.

I think I was trying to solve 2 things with this method: 1) We need to keep the order of the arguments, 2) in Fortran in some cases we just know a symbol is a Argument if it is in the argument list. Maybe it was an error to try to solve both together.

What do you think if we instead:
- Move the `for name in argument_name_list:` loop that solves (2) outside of the method into the fparser2ASTProcessor `process_declarations()` line where the method is called. (It is quite a Fortran-specific behavior anyway)
- Remove the `interface` argument as `specify_argument_list` will just be in charge of (1)
- Add 2 error conditions into `specify_argument_list`: a) All symbol names passed exist and have a Symbol.Argument interface. 2) There aren't any other symbols in the SymbolTable with interface of type Symbol.Argument.
- Add the same 2 conditions to the property that returns the argument_list. So if at some point the user adds or removes a new Symbol.Argument interface to/from a symbol, it will invalidate the `symbol_table.argument_list` until a new list order with the updated symbols is provided.

src/psyclone/psyGen.py
sergisiso: You can probably ignore this one depending how you chose to solve the issue below.

src/psyclone/psyGen.py
arporter: Yes, I've made it more explicit.

src/psyclone/psyGen.py
arporter: Oh yes - I'd already fixed that in the `Symbol` constructor so fixing it here is trivial :-)

src/psyclone/psyGen.py
```diff
@@ -5889,54 +6057,11 @@ def datatype(self):
...
-        :raises TypeError: Provided parameters have invalid error type.
-        :raises ValueError: 'new_is_output' contains an invalid value.
-        '''
-        if not isinstance(new_is_output, bool):
-            raise TypeError("Symbol 'is_output' attribute must be a boolean.")
-        if self.scope == 'local' and new_is_output is True:
-            raise ValueError("Symbol with 'local' scope can not have "
-                             "'is_output' attribute set to True.")
-        self._is_output = new_is_output
+    def access(self):
```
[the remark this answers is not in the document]
arporter: Oops. Fixed.

src/psyclone/psyGen.py
```diff
@@ -5815,30 +5894,120 @@ class Symbol(object):
...
+            :param str value: the name of the Fortran module.
+
+            :raises TypeError: if the supplied value is not a str.
+            :raises ValueError: if the supplied string is not at least one \
+                                character long.
+            '''
+            if not isinstance(value, str):
+                raise TypeError("module_name must be a str but got '{0}'".
+                                format(type(value)))
+            if len(value) < 1:
```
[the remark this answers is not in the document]
arporter: Oh yes, I knew that really ;-)

arporter:
I think I've addressed all of the reviewer's comments. I just need to manually bring the branch up-to-date with master since I lost the race with Rupert :-)

rupertford:
I sympathise :-)

arporter:
That was quite a painful merge in the end. Anyway, all done now and ready for next review...

src/psyclone/tests/psyGen_test.py
```diff
@@ -3258,74 +3258,82 @@ def test_symbol_scalar_array():
...
+            in str(sym3))
 
-def test_symbol_is_input_setter():
-    '''Test that a Symbol's is_input method can be set if given a new
-    valid value, otherwise it raises a relevant exception.'''
+    sym2._shape.append('invalid')
+    with pytest.raises(InternalError) as error:
+        _ = str(sym2)
+    assert ("Symbol shape list elements can only be 'Symbol', 'integer' or "
+            "'None', but found") in str(error.value)
```
sergisiso: This last 10 lines did not merge as expected, they belong to the `test_symbol_can_be_printed()` test

src/psyclone/tests/psyGen_test.py
```diff
@@ -3258,74 +3258,82 @@ def test_symbol_scalar_array():
...
+    with pytest.raises(ValueError) as err:
+        _ = Symbol.FortranGlobal("")
+    assert "module_name must be one or more characters long" in str(err)
+    with pytest.raises(TypeError) as err:
+        _ = Symbol.FortranGlobal(1)
+    assert "module_name must be a str but got" in str(err)
 
     sym3 = Symbol("s3", "integer", constant_value=12)
-    assert "s3: <integer, local, Scalar, constant_value=12>" in str(sym3)
+    assert "s3: <integer, Scalar, local, constant_value=12>" in str(sym3)
```
sergisiso: This 2 lines have also jumped from the `test_symbol_can_be_printed()`

src/psyclone/tests/psyGen_test.py
```diff
@@ -4057,8 +4125,49 @@ class UnrecognizedType(object):
...
+    reader = FortranStringReader("use my_mod, only: some_var\n"
+                                 "use this_mod\n"
+                                 "use other_mod, only: var1, var2\n")
+    fparser2spec = Specification_Part(reader)
+    processor.process_declarations(fake_parent, fparser2spec.content, [])
+    for var in ["some_var", "var1", "var2"]:
+        assert fake_parent.symbol_table.lookup(var).name == var
+        assert fake_parent.symbol_table.lookup(var).scope == "global"
+    assert fake_parent.symbol_table.lookup("some_var").interface.module_name \
+            == "my_mod"
```
sergisiso: pycodestyle says `continuation line over-indented for visual indent`

sergisiso:
Also, do you think we should add another consistency check in the `specify_argument_list` and `argument_list` property that iterates over the whole symbol_table items and checks that all the items that have interface as `symbol.Argument` are part of the argument_list?

This will automatically invalidate the `argument_list` if we add a new symbol with `Symbol.Argument` in the symbol table after calling `specify_argument_list`

src/psyclone/psyGen.py
```diff
@@ -5241,6 +5242,8 @@ def process_declarations(self, parent, nodes, arg_list):
         :type arg_list: :py:class:`fparser.Fortran2003.Dummy_Arg_List`
         :raises NotImplementedError: The provided declarations contain
                                      attributes which are not supported yet.
+        :raises GenerationError: if the parse tree for a USE statement does \
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -5806,56 +5848,200 @@ def _number_handler(self, node, parent):
...
 
     '''
-
-    # Tuple with the valid values for the access attribute.
-    valid_scope_types = ('local', 'global_argument')
-    # Tuple with the valid datatypes.
-    valid_data_types = ('real', 'integer', 'character', 'boolean')
-    # Mapping from supported data types for constant values to
-    # internal Python types
+    ## Tuple with the valid datatypes.
```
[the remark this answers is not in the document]
arporter: Hmm. Not sure what to do about that. The reason I've added the '##' is because that's Doxygen mark-up and it means our comments get picked up in the Doxygen-generated reference guide.
I think we should discuss this in the telco tomorrow. Perhaps we can disable that particular pycodestyle error?

src/psyclone/psyGen.py
```diff
@@ -5878,18 +6066,16 @@ def __init__(self, name, datatype, shape=[], scope='local',
...
         self._shape = shape
+        self._interface = None
 
-        # The following attributes have setter methods (with error checking)
-        self._scope = None
+        if interface:
+            # If an interface is specified for this symbol then the data with
+            # which it is associated must exist outside of this kernel.
+            self.interface = interface
+        # The following attribute has a setter method (with error checking)
```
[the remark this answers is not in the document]
arporter: Oh yes, done. (And I've removed the check that `interface` is not None as the setter now allows that.)

src/psyclone/psyGen.py
```diff
@@ -6260,25 +6419,32 @@ def swap_symbol_properties(self, symbol1, symbol2):
...
-    def specify_argument_list(self, argument_name_list):
+    def specify_argument_list(self, argument_symbols):
         '''
-        Keep track of the order of the arguments and provide the scope,
-        is_input and is_output information if it was not available on the
-        variable declaration.
+        Sets-up the internal list storing the order of the arguments to this
+        kernel.
+
+        :param list argumentssymbols: Ordered list of the Symbols representing\
```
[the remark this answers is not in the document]
arporter: Oops.

src/psyclone/tests/psyGen_test.py
arporter: Oops again. Fixed.

src/psyclone/tests/psyGen_test.py
arporter: Oh yes. Apologies for missing that.

src/psyclone/tests/psyGen_test.py
arporter: Done.

arporter:
Assuming Travis is happy then this is ready for another review.

arporter:
(I've followed your suggestion of making the internal consistency checks more thorough. I had to update a few tests as a result but I think the code is better now.)

src/psyclone/psyGen.py
```diff
@@ -5840,55 +5882,200 @@ def _number_handler(self, node, parent):
...
+            return True
+
+        def __str__(self):
+            return "Argument(pass-by-value={0})".format(self._pass_by_value)
+
+    class FortranGlobal(SymbolInterface):
+        '''
+        Describes the interface to a Fortran Symbol representing data that
+        exists outside of the local scope but is not supplied as a routine
+        argument.
```
sergisiso: Instead of "but is not supplied as a routine argument." why not use "is supplied as a Fortran global variable."
In the future we may want to add more SymbolInterface classes.

src/psyclone/psyGen.py
```diff
@@ -5840,55 +5882,200 @@ def _number_handler(self, node, parent):
...
+                                 "long")
+            self._module_name = value
+
+        @property
+        def is_argument(self):
+            '''
+            :returns: Whether or not this symbol is a routine argument.
+            :rtype: bool
+            '''
+            return False
```
sergisiso: Again, if we add more SymbolInterface sub-classes in the future, maybe we want this is_argument->False in the base class and just redefine it at the `Argument` class.

src/psyclone/psyGen.py

src/psyclone/psyGen.py
arporter: I've removed `is_argument` altogether as we only ever use it in the tests :-)

src/psyclone/tests/psyGen_test.py
```diff
@@ -3559,28 +3556,125 @@ def test_symboltable_can_be_printed():
...
+        sym_table._validate_arg_list([sym_table.lookup("var3")])
+    # Manually put that symbol into the argument list
+    sym_table._argument_list = [sym_table.lookup("var3")]
+    pattern = (r"Symbol \'var3.*\' is listed as a kernel argument but has an "
+               r"interface of type \'.*\.FortranGlobal\'>")
+    assert re.search(pattern, str(err)) is not None
+    # Check that the argument_list property converts this error into an
+    # InternalError
+    with pytest.raises(InternalError) as err:
+        _ = sym_table.argument_list()
```
sergisiso: pylint says `sym_table.argument_list` is not callable, maybe without paranthesis? Also, is the str(err) comparison assert missing?

src/psyclone/tests/psyGen_test.py
```diff
@@ -3593,7 +3687,7 @@ def test_symboltable_contains():
     assert "var1" in sym_table
     assert "var2" in sym_table
     assert "var3" not in sym_table
-
+
```
sergisiso: Trailing whitespace

src/psyclone/tests/gocean1p0_opencl_test.py
```diff
@@ -307,6 +318,7 @@ def test_opencl_kernel_gen_wrong_kernel():
     ''' Tests that gen_ocl method raises the proper error when the
     GOKernelSchedule does not represent a proper GOcean kernel.
     '''
+    from psyclone.psyGen import Symbol
```
sergisiso: pylint says Symbol is redefined from outer scope, we have it already in the file imports

src/psyclone/tests/psyGen_test.py
arporter: Oops and no, I'd just not bothered :-) I've added it here and above now.

src/psyclone/tests/psyGen_test.py
arporter: Removed.

src/psyclone/tests/gocean1p0_opencl_test.py
arporter: Oops, fixed.

arporter:
Apologies for missing those pylint errors. Should be all OK now.

Issue #239, opened 2018-11-14
Use fparser2 for parsing kernel meta-data


arporter:

rupertford:
Created branch 239_kernels_fparser2

sergisiso:
I have started implementing this

Issue #138, opened 2018-02-28
Disallow GH_WRITE access for continuous fields


TeranIvy:
Although `GH_WRITE` access is officially not allowed for continuous fields in the Dynamo0.3 API there is no explicit restriction and test to reject this combination in PSyclone.
This issue will add the restriction, similar to how it is being done in #25 for `GH_READWRITE` access for continuous fields (`class DynArgDescriptor03(Descriptor):` in `dynamo0p3.py` and related tests in `dynamo0p3_test.py`).

The instances of continuous `GH_WRITE` access in PSyclone code and LFRic kernels will have to be replaced with `GH_INC`. To test for any undesirable side effects in LFRic I replaced `GH_WRITE` continuous instances (only `W0` and `W2` were found) with `GH_INC` in a working copy of the trunk. The good news is that all tests successfully complete (compilation, runs and KGOs checks). It would perhaps be good to warn LFRic developers not to add any more continuous `GH_WRITE` until the PSyclone restriction is in place - more work to remove them later :)

**Update**
It was agreed that this approach will be adopted for the user-supplied kernels that operate over `cells`. Built-ins can use `GH_WRITE` and `GH_READWRITE` on `ANY_SPACE` as they loop over DoFs (see #471).

TeranIvy:
Closing the issue as this will be properly implemented in #471.

TeranIvy:
Reopening as there is enough work for continuous function spaces without any_space and built-in accesses to do in one issue.

rupertford:
I think this should stay open even though #818 is now on master. Is that right @TeranIvy?

TeranIvy:
It can be safely closed now, @rupertford, as the rest of the field access logic for `any_space` and DoF looping is being done in isssue #471 / PR #823. I will link this issue to PR #818 as it should have been from the start, so that will close it.

TeranIvy:
Well, linking PR didn't automatically close the issue - since you are the reviewer, @rupertford, are you happy to close it?

rupertford:
Closed as suggested by @TeranIvy

PR #371, opened 2019-04-30
Fix 329

### LINKED ISSUE #355: Wrong working directory ... or: PSyclone install does not install psyclone.cfg in a local install
When I do a ``pip install --user -e .``, it appears that no psyclone.cfg is installed in a local directory, while a system-wide install will copy this file to /usr/share/psyclone.

So far this worked for me because I have a custom config file in psyclone/tests/.psyclone, which is used in most tests. But I just stumbled across some tests do a chdir to a temporary directory, and then NOT a chdir back to the original directory. Which means that the next test that is run will have the previous test's temporary directory as working directory. If additionally the latter test reloads the config file (since that is what it is supposed to test), now psyclone cannot find my .../psyclone/tests/,psyclone config file anymore (since it only tests .psyclone/psyclone.cfg), so the tests fail with config file not found. Doing a system-wide install fixes the problem (since psyclone always checks /usr/share/psyclone).

As a first solution I have fixed up the tests in my current branch to do a chdir back to the original directory, but that only solves the problem partially: if a test in a temporary directory should fail an assert, the chdir back is not executed, and the next test may fail because it has an unexpected working directory again :(

Two solutions that I can see:
1) we fix the install process in case of a local installation so that the psyclone.cfg file is copied to $HOME/.local/share/psyclone. --> All tests would find this config file since it is always tested for in config find_file().
2) We add a fixture in conftest that is executed for ALL tests (or at least the ones that chdir to a directory) that will make sure the chdir back to the original directory at the end (or at the beginning of each test).

The first one seems to be the easier solution, but I don't understand why it is not installed in the first place :)

hiker:
This patch does not fix _all_ global variables (as in: variables that e.g. dynamo0p3 sets in psyGen), but it should fix all access related global variables. Since the patch is already reasonable big, I think it's better to limit the scope to access variables for now, and leave some of the other global variables for a future patch.

A side effect of this patch was that I renamed some of the global variables to have a GH_ in front of it in dynamo0p3 (that simplified my incremental work when I had some variables being the api-specific name, some the new api-independent access type), a future patch should remove them all similar to this patch removing FIELD_ACCESS_MAP , MAPPING_REDUCTIONS etc.

As discussed previously I added the access_type to a new directory '.../psyclone/core' (with tests in psyclone/core/tests), as a simple start to split those huge files into smaller ones. The idea of having tests under core/tests is that it is easy if you work in core to only run the tests closely related to that directory. Am happy to move that directory (there would be some case to be made to have it in .../src/psyclone/tests/core/ - mostly the usage for the conftest.py file under tests).

I also moved some of the constant lists (e.g. READ_ACCESSES) as functions into AccessType.

It also fixes #355.


arporter:
In case I've missed it, @TeranIvy - is the MO environment OK if we gain a dependence on enum?

src/psyclone/dynamo0p3.py
```diff
@@ -8102,25 +8112,24 @@ def unique_fs_names(self):
...
-    def iteration_space_arg(self, mapping=None):
+    def iteration_space_arg(self):
         '''
         Returns an argument we can use to dereference the iteration
         space. This can be a field or operator that is modified or
         alternatively a field that is read if one or more scalars
         are modified. If a kernel writes to more than one argument then
         that requiring the largest iteration space is selected.
 
-        :param dict mapping: un-used argument. Retained for consistency
```
arporter: Whoop!

src/psyclone/gocean0p1.py
```diff
@@ -51,8 +52,20 @@ class GODescriptor(Descriptor):
     metadata. Currently this class is just a wrapper around the base
     class.
 
+    :param str access: whether argument is read/write etc.
+    :param str space: which function space/grid-point type argument is \
+    on
+    :param dict stencil: type of stencil access for this \
+    argument. Defaults to None if the argument is not supplied.
+    :param string mesh: which mesh this argument is on. Defaults to \
```
arporter: s/string/str/

src/psyclone/psyGen.py
```diff
@@ -203,7 +199,7 @@ def args_filter(arg_list, arg_types=None, arg_accesses=None, arg_meshes=None,
     :param arg_types: List of argument types (e.g. "GH_FIELD")
     :type arg_types: list of str
     :param arg_accesses: List of access types that arguments must have
-    :type arg_accesses: list of str
+    :type arg_accesses: List of :py:class:`psyclone.core.access_type`.
```
arporter: s/access_type/AccessType/ I think?

src/psyclone/psyGen.py
```diff
@@ -2278,13 +2268,16 @@ def view(self, indent=0):
 
     def _get_reductions_list(self, reduction_type):
         '''Return the name of all scalars within this region that require a
-        reduction of type reduction_type. Returned names will be unique. '''
+        reduction of type reduction_type. Returned names will be unique.
+        :param: reduction_type
```
arporter: Please complete this line documenting `reduction_type` :-)

src/psyclone/psyGen.py
```diff
@@ -2278,13 +2268,16 @@ def view(self, indent=0):
...
+        :type reduction_type: :py:class:`psyclone.core.AccessType`
+        '''
+        print("Called")
```
arporter: and remove this line ;-)

src/psyclone/psyGen.py
```diff
@@ -3124,26 +3117,24 @@ def __str__(self):
...
-            "must be provided"
         for kern_call in self.kern_calls():
             for arg in kern_call.arguments.args:
-                if arg.access.lower() == mapping["inc"]:
+                if arg.access == AccessType.INC:
                     return True
         return False
 
-    def unique_modified_args(self, mapping, arg_type):
+    def unique_modified_args(self, arg_type):
```
arporter: Please document the interface to this routine.

src/psyclone/psyGen.py
```diff
@@ -3615,51 +3608,48 @@ def gen_arg_setter_code(self, parent):
...
 
-    def incremented_arg(self, mapping={}):
+    def incremented_arg(self):
         ''' Returns the argument that has INC access. Raises a
         FieldNotFoundError if none is found.
 
-        :param mapping: dictionary of access types (here INC) associated \
-                        with arguments with their metadata strings as keys
-        :type mapping: dict
-        :returns: a Fortran argument name.
```
arporter: Please put back the `:returns:` line in the docstring.

src/psyclone/psyGen.py
```diff
@@ -3615,51 +3608,48 @@ def gen_arg_setter_code(self, parent):
...
                 return arg
+
+        api_config = Config.get().api_conf()
+        rev_access_mapping = api_config.get_reverse_access_mapping()
+
         raise FieldNotFoundError("Kernel {0} does not have an argument with "
                                  "{1} access".
-                                 format(self.name, mapping["inc"]))
+                                 format(self.name,
+                                        rev_access_mapping[AccessType.INC]))
```
arporter: I've definitely not understood things yet. What's the difference between using `rev_access_mapping` and `api_name`? I'd naively have thought that they gave you the same information?

src/psyclone/psyGen.py
```diff
@@ -3615,51 +3608,48 @@ def gen_arg_setter_code(self, parent):
...
                     return arg
+        api_config = Config.get().api_conf()
+        rev_access_mapping = api_config.get_reverse_access_mapping()
         raise FieldNotFoundError("Kernel {0} does not have an argument with "
                                  "{1} or {2} access".
-                                 format(self.name, mapping["write"],
-                                        mapping["readwrite"]))
+                                 format(self.name,
+                                        rev_access_mapping[AccessType.WRITE],
+                                        rev_access_mapping
```
arporter: Strange line-break. Please could you adjust the formatting.

src/psyclone/psyGen.py
```diff
@@ -3965,25 +3955,20 @@ def names(self):
...
         :raises GenerationError: if none such argument is found.
 
         '''
-        assert mapping != {}, "psyGen:Arguments:iteration_space_arg: Error "
-        "a mapping needs to be provided"
         for arg in self._args:
-            if arg.access.lower() == mapping["write"] or \
-               arg.access.lower() == mapping["readwrite"] or \
-               arg.access.lower() == mapping["inc"]:
+            if arg.access == AccessType.WRITE or \
```
arporter: Could just do something like:

    if arg.access not in AccessType.write_accesses():

perhaps?

src/psyclone/psyGen.py
```diff
@@ -4199,21 +4184,13 @@ def __init__(self, call, arg_info, access):
...
-        # MAPPING_ACCESSES specified in the dynamo0p3 file which
-        # overide the default ones in this file.
-        self._write_access_types = [MAPPING_ACCESSES["write"],
-                                    MAPPING_ACCESSES["readwrite"],
-                                    MAPPING_ACCESSES["inc"],
-                                    MAPPING_REDUCTIONS["sum"]]
-        self._read_access_types = [MAPPING_ACCESSES["read"],
-                                   MAPPING_ACCESSES["readwrite"],
-                                   MAPPING_ACCESSES["inc"]]
+        self._write_access_types = [AccessType.WRITE,
```
arporter: Could these lists just be defined in AccessType?

src/psyclone/psyGen.py
```diff
@@ -4241,9 +4218,28 @@ def access(self):
 
     @access.setter
     def access(self, value):
-        ''' set the access type for this argument '''
+        '''Set the access type for this argument.
+        :param value: New access type.
+        :type value: :py:class:`psyclone.core.access_type`.
```
arporter: s/access_type/AccessType/

src/psyclone/psyGen.py
```diff
@@ -4241,9 +4218,28 @@ def access(self):
...
+        '''
+        if not isinstance(value, AccessType):
+            raise ValueError("Invalid access type: '{0}'.".format(value))
+
         self._access = value
 
+    def is_read(self):
+        '''Checks if the argument is read, i.e. has access
+        type read, readwrite, or inc.'''
+        return self._access in [AccessType.READ, AccessType.READWRITE,
```
arporter: `in self._read_access_types` perhaps?

src/psyclone/psyGen.py
```diff
@@ -4241,9 +4218,28 @@ def access(self):
...
+                                AccessType.INC]
+
+    def is_written(self):
+        '''Checks if the argument is written, i.e. has access
+        type write, readwrite, or inc.'''
+        return self._access in [AccessType.WRITE, AccessType.READWRITE,
```
arporter: Again, `in self._write_access_types`?

src/psyclone/psyGen.py
```diff
@@ -4241,9 +4218,28 @@ def access(self):
...
+        :type value: :py:class:`psyclone.core.access_type`.
+        :raises ValueError if value is not an AccessType.
+        '''
+        if not isinstance(value, AccessType):
+            raise ValueError("Invalid access type: '{0}'.".format(value))
+
         self._access = value
 
+    def is_read(self):
+        '''Checks if the argument is read, i.e. has access
```
arporter: Please document the return value.

src/psyclone/psyGen.py
```diff
@@ -4241,9 +4218,28 @@ def access(self):
...
+        type read, readwrite, or inc.'''
+        return self._access in [AccessType.READ, AccessType.READWRITE,
+                                AccessType.INC]
+
+    def is_written(self):
```
arporter: Please document the return value.

src/psyclone/tests/conftest.py
```diff
@@ -79,6 +79,28 @@ def have_graphviz():
...
+    import os
+    config_file = Config.get_repository_config_file()
+
+    # In case that PSyclone is installed and tested (e.g. travis),
+    # the 'repository' config file does not exist (since it is
+    # installed in a different directory). In that case the standard
+    # search path of the Configuration object will find the right
+    # config file. So only overwrite the default search path behaviour
+    # if the repository config file is actually found:
+    if os.path.isfile(config_file):
```
arporter: I love it when a plan comes together :-)

.travis.yml
```diff
@@ -50,12 +50,14 @@ before_install:
   # fparser pointed to by the git submodule.
   - ./bin/install_optional.sh
   - pip install codecov
+  - python -V
+  # python 2.7.14 on travis does not have enum, so in this case install
+  # a compatible enum implementation.
+  - if [ "$(python --version 2>&1)" == "Python 2.7.14" ]; then pip install enum34; fi
```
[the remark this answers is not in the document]
hiker: I admit I never looked at what setup.py does :) But agreed, and done - sees to work fine.

src/psyclone/configuration.py
```diff
@@ -121,6 +121,18 @@ def get(do_not_load_file=False):
...
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def get_repository_config_file():
+        '''This function returns the absolute path to the config file included
+        in the PSyclone repository. It is used by the testing framework to make
+        sure all tests get the same config file (see tests/config_tests for the
+        only exception).
+        '''
+        this_dir = os.path.dirname(os.path.abspath(__file__))
+        return os.path.join(this_dir, "..", "..", "config",
```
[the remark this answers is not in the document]
hiker: While I am not aware of anything not supporting ".." (and still running PSyclone), but using dirname twice indeed works.

src/psyclone/configuration.py
```diff
@@ -335,7 +363,11 @@ def find_file():
               <base-dir-of-virtual-env>/share/psyclone/
         - ${HOME}/.local/share/psyclone/
         - <system-install-prefix>/share/psyclone/
+        - ../../config/
 
+        The last entry is used for development when no installation was done.
+        It will then pick the psyclone.cfg file that is part of the
+        repository.
```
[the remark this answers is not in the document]
hiker: Hangover (and I don't say whose)

src/psyclone/configuration.py
```diff
@@ -574,7 +606,104 @@ def get_default_keys(self):
...
+        self._access_mapping = {"read": "read", "write": "write",
+                                "readwrite": "readwrite", "inc": "inc",
+                                "sum": "sum"}
+        # Get the mapping and convert it into a directory. The input is in
+        # the format: key1:value1, key2=value2, ...
+        mapping = section.get("ACCESS_MAPPING")
+        if mapping:
+            from psyclone.core.access_type import AccessType
+            self._access_mapping = APISpecific.create_dict_from_string(mapping)
+            # Now convert the string type ("read" etc) to AccessType
```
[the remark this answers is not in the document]
hiker: Very good catch ... fixed, and I added a special test for this as well.

src/psyclone/core/access_type.py
```diff
@@ -0,0 +1,125 @@
...
+    READWRITE = 4
+    SUM = 5
+
+    @staticmethod
+    def get_size():
+        ''':returns: The number of elements in this emnumerator.
+        :rtype int'''
+        return len(AccessType)
+
+    def __str__(self):
```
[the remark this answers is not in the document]
hiker: __str__() returns (e.g. ) "INC", by default it would "AccessType.INC". I think it's worth keeping __str__(), I prefer it :)

src/psyclone/core/tests/access_type_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_generic():
+    '''Generic tests.'''
+    assert AccessType.get_size() == 5
+
+
+def test_str():
+    '''Tests conversion to an (API specific) string'''
+
```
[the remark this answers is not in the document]
hiker: Indeed, copy&paste error mixed the two test cases. Should be all fixed now.

src/psyclone/core/tests/access_type_test.py
```diff
@@ -0,0 +1,82 @@
...
+    assert str(AccessType.INC) == "INC"
+    assert str(AccessType.WRITE) == "WRITE"
+    assert str(AccessType.READ) == "READ"
+    assert str(AccessType.READWRITE) == "READWRITE"
+    assert str(AccessType.SUM) == "SUM"
+
+
+def test_api_name():
+    '''Tests api_name(). '''
+    assert AccessType.INC.api_name() == "gh_inc"
```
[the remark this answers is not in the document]
hiker: Same copy&paste error - the setting of the config file was in the wrong test :(

src/psyclone/dynamo0p3.py
```diff
@@ -582,20 +562,32 @@ def __init__(self, arg_type):
                 "not get to here")
 
         # The 2nd arg is an access descriptor
-        if arg_type.args[1].name not in VALID_ACCESS_DESCRIPTOR_NAMES:
+        self._access_descriptor = arg_type.args[1]
+
+        # Now convert from GH_* names to the generic access type:
+        api_config = Config.get().api_conf("dynamo0.3")
+        access_mapping = api_config.get_access_mapping()
+        rev_access_mapping = api_config.get_reverse_access_mapping()
```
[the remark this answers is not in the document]
hiker: No, it is used ... just way below. I've moved the assignment of rev_access_mapping to the location where it is actually used.

src/psyclone/dynamo0p3.py
```diff
@@ -696,26 +688,28 @@ def __init__(self, arg_type):
...
                 raise ParseError(
                     "In the dynamo0.3 API each meta_arg entry must have 2 "
                     "arguments if its first argument is gh_{{r,i}}scalar, but "
                     "found {0} in '{1}'".format(len(arg_type.args), arg_type))
             # Test allowed accesses for scalars (read_only or reduction)
-            if self._access_descriptor.name not in ["gh_read"] + \
-               VALID_REDUCTION_NAMES:
+            if self._access_descriptor.name not in [AccessType.READ] + \
+               AccessType.get_valid_reduction_modes():
+                api_name = rev_access_mapping[self._access_descriptor.name]
```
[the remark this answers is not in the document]
hiker: I really prefer to have 'api' in the name, since access_name() does not indicate that AccessType.INC.access_name() can be different depending on the API loaded. What about:
api_specific_name()  ?

src/psyclone/dynamo0p3.py
```diff
@@ -4203,25 +4200,39 @@ def __init__(self, alg_invocation, idx):
...
+                        arg_accesses=AccessType.get_valid_reduction_modes(),
+                        unique=True):
                     global_sum = DynGlobalSum(scalar, parent=loop.parent)
                     loop.parent.children.insert(loop.position+1, global_sum)
 
+    def unique_declarations(self, datatype, access=None):
+        if access and not isinstance(access, AccessType):
+            api_config = Config.get().api_conf("dynamo0.3")
+            valid_names = api_config.get_valid_accesses_api()
+            raise GenerationError(
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyGen.py
hiker: We were both wrong, as far as I can tell the right way should be: psyclone.core.access_type.AccessType.

Is there actually a way of testing this? I remembered that I saw a doxyfile in the past, but that has been reverted.

src/psyclone/psyGen.py
hiker: No, you are spot on ... just that code was written before I decided to add the convenience function api_(specific)_name. I've updated this to use api_name() (and will change it once we agree on the final name for api_name()).

src/psyclone/psyGen.py
hiker: OK, now I had that problem in several places: why is 'sum' missing in that list? AccessType has a 'all_write_access()' function, but it includes the reductions (i.e. sum). If we can add SUM here, then it's easy. If not, we need to find better names, which might get quite confusing ('all_write_access', 'all_write_non_reduction_accesses', ...).

src/psyclone/psyGen.py
hiker: Sure (there already is 'all_read_access()'). Do you want to keep the self._XXXX_access_types variables, or should I replace them everywhere with corresponding functions in AccessType?

src/psyclone/psyGen.py
hiker: See above :)

src/psyclone/psyGen.py
hiker: The two is_read/write functions are actually a left over before I decided to use enums. So I have removed the completely.

hiker:
@arporter : Than you very much for your effort in this review!!

I've also done all other corrections you mentioned (I thought I better save a dozen of emails just stating 'done' by just typing this once).

Outstanding before another review can be done:
1) Better name for api_name ... I suggested api_specific_name
2) Keep self._write_access_types etc, or replace their usage everywhere with an AccessType function call.

src/psyclone/dynamo0p3.py
```diff
@@ -582,20 +562,32 @@ def __init__(self, arg_type):
...
-        if arg_type.args[1].name not in VALID_ACCESS_DESCRIPTOR_NAMES:
+        self._access_descriptor = arg_type.args[1]
+
+        # Now convert from GH_* names to the generic access type:
+        api_config = Config.get().api_conf("dynamo0.3")
+        access_mapping = api_config.get_access_mapping()
+        rev_access_mapping = api_config.get_reverse_access_mapping()
+        try:
+            self._access_descriptor.name = \
+                access_mapping[self._access_descriptor.name]
```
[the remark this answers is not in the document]
hiker: Oops, I missed commenting on this point. Yes, I was surprised, too - and felt VERY bad when assigning an AccessType to the name. The problem is that _access_descriptor is a FunctionVar, which is used for the argument type:
```
    if arg_type.args[0].name not in GH_VALID_ARG_TYPE_NAMES:
...
            self._type = arg_type.args[0].name
...
        # The 2nd arg is an access descriptor
        self._access_descriptor = arg_type.args[1]
```
So I don't think that I can easily rename that attribute. I could add one more attribute perhaps, though ... it's odd to add this to a FunctionVar :(

The best option I see would be to create special classes for that would replace the FunctionVar there. I can't really estimate the scope of that, I am not deeply familiar with that code :( But I don't think it's a trivial change :(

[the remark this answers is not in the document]
hiker: Dang - I was too eager. Turns out while the full FunctionVar is stored, only the name attribute of it was actually used, which made it quite easy to fix this properly (a199a225ac52828930dd1a02fb03e5184aa69fe1). Sorry for missing that.

src/psyclone/core/access_type.py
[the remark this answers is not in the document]
arporter: OK.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
arporter: How about `api_access_name`?

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
arporter: Oh good. This code will probably also be touched by @rupertford 's work on switching the meta-data parsing over to fparser2 (#239).

src/psyclone/psyGen.py
arporter: I think you'd have to modify the documentation source files to say that you wanted Sphinx to generate the docs for this class. One day we will have an API reference guide and we would check in there that it was generated OK.

src/psyclone/psyGen.py
arporter: Great!

src/psyclone/psyGen.py
arporter: Off the top of my head I don't know the answer. Perhaps @rupertford can remember? I suspect we could put SUM in the list and see what happens...

src/psyclone/psyGen.py
arporter: I think it would be nice to have everything in one place (`AccessType`) if that works for you.

arporter:
@hiker, you're welcome :-) I think I've addressed the outstanding questions.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
hiker: While I have to admit to prefer api_specific_name (the typical usage of this functions is:
```
  access.api_specific_name()   # vs:
  access.api_access_name()
```
But it doesn't really mattter, so api_access_name() it is now.

src/psyclone/psyGen.py
hiker: I started to have a look at this. From the docs it would indeed appear that a reduction should not be in that list (it is used to find a field to iterate over - in case of a SUM I'd guess the argument would be a scalar). So for now I am using existing AccessType functions:
```
            if arg.access in AccessType.all_write_accesses() and \
                    arg.access not in AccessType.get_valid_reduction_modes():
                return arg
```
I also tried to add a test case for this, but it turns out that dynamo3 does not use this function (it overwrites it with its own function and does not call the function in psyGen), and of course gocean does not support SUM in the first place. I'll just left it unchanged (except for the lines above, which I think will indeed increase portability later).

src/psyclone/psyGen.py
hiker: OK, all _{read,write}_access_types are gone now :)

hiker:
I believe it is ready for another review.

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -1280,11 +1294,13 @@ def test06_kernel_invalid_access():
     ''' Check that we raise an error if a kernel's meta-data specifies
     an unrecognised access type for a kernel argument (i.e. something
     other than READ,WRITE,READWRITE) '''
-    with pytest.raises(ParseError):
+    with pytest.raises(ParseError) as err:
         parse(os.path.join(os.path.dirname(os.path.abspath(__file__)),
                            "test_files", "gocean1p0",
                            "test06_invoke_kernel_wrong_access.f90"),
               api="gocean1.0")
+    assert "compute_cu: argument access  is given as 'wrong' but must be one "\
```
arporter: Thanks for adding.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1421,7 +1428,12 @@ def test_go_loop_swap_errors():
                          "test_files", "dynamo0p3",
                          "1.0.1_single_named_invoke.f90"),
                     api="dynamo0.3")
+    # TODO: that still doesn't work, we have inconsistent values
```
arporter: Is this still true? If so it sounds like something that needs addressing...

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1421,7 +1428,12 @@ def test_go_loop_swap_errors():
...
+    # in some psygen objects (read vs gh_read)
+    Config.get().api = "dynamo0.3"
```
arporter: Candidate for monkeypatching?

src/psyclone/tests/kernel_transformation_test.py
```diff
@@ -182,7 +182,7 @@ def test_new_kernel_file(tmpdir, monkeypatch):
     assert str(names[0]) == "continuity{0}_type".format(tag)
     # TODO #281 check compilation of code (needs Joerg's extension of
```
arporter: As you've touched this test, please could you add the compilation testing and remove this TODO?

src/psyclone/tests/psyGen_test.py
```diff
@@ -864,7 +872,7 @@ def test_acc_dir_view(capsys):
     acclt = ACCLoopTrans()
     accdt = ACCEnterDataTrans()
     accpt = ACCParallelTrans()
-
+    Config.get().api = "gocean1.0"
```
arporter: monkeypatch?

src/psyclone/tests/psyGen_test.py
```diff
@@ -914,6 +924,7 @@ def test_globalsum_view(capsys):
     then call view() on that.'''
     from psyclone.psyGen import colored, SCHEDULE_COLOUR_MAP
     from psyclone import dynamo0p3
+    Config.get().api = "dynamo0.3"
```
arporter: Is this line still needed despite the fixture?

src/psyclone/tests/psyGen_test.py
```diff
@@ -1013,6 +1025,7 @@ def test_reduction_var_error():
 def test_reduction_sum_error():
     '''Check that we raise an exception if the reduction_sum_loop()
     method is provided with an incorrect type of argument'''
+    Config.get().api = "dynamo0.3"
```
arporter: Is this line needed?

src/psyclone/tests/psyGen_test.py
```diff
@@ -1999,13 +2013,15 @@ def test_node_is_valid_location():
 def test_node_ancestor():
     ''' Test the Node.ancestor() method '''
     from psyclone.psyGen import Loop
+    Config.get().api = "gocean1.0"
```
arporter: monkeypatch?

src/psyclone/tests/psyGen_test.py
```diff
@@ -2091,6 +2107,7 @@ def test_acc_dag_names():
     from psyclone.psyGen import ACCEnterDataDirective
     from psyclone.transformations import ACCEnterDataTrans, ACCParallelTrans, \
         ACCLoopTrans
+    Config.get().api = "gocean1.0"
```
arporter: monkeypatch?

src/psyclone/tests/psyGen_test.py
```diff
@@ -2556,6 +2574,7 @@ def test_find_w_args_multiple_deps(monkeypatch, annexed):
 def test_loop_props():
     ''' Tests for the properties of a Loop object. '''
     from psyclone.psyGen import Loop
+    Config.get().api = "gocean1.0"
```
arporter: monkeypatch/use fixture?

src/psyclone/tests/psyGen_test.py
```diff
@@ -2564,11 +2583,14 @@ def test_loop_props():
         loop.loop_type = "not_a_valid_type"
     assert ("loop_type value (not_a_valid_type) is invalid. Must be one of "
             "['inner', 'outer']" in str(err))
+    Config.get().api = "dynamo0.3"
 
 
 def test_node_abstract_methods():
     ''' Tests that the abstract methods of the Node class raise appropriate
     errors. '''
+    Config.get().api = "gocean1.0"
```
arporter: monkeypatch/fixture?

src/psyclone/tests/psyGen_test.py
```diff
@@ -2581,18 +2603,21 @@ def test_node_abstract_methods():
...
         Node.view(loop)
     assert "BaseClass of a Node must implement the view method" in str(err)
+    Config.get().api = "dynamo0.3"
 
 
 def test_kern_ast():
     ''' Test that we can obtain the fparser2 AST of a kernel. '''
     from psyclone.gocean1p0 import GOKern
     from fparser.two import Fortran2003
+    Config.get().api = "gocean1.0"
```
arporter: monkeypatch/fixture?

src/psyclone/dynamo0p3.py
```diff
@@ -6528,13 +6525,13 @@ def gen_stub(self):
...
         INC access.  '''
-        return Kern.incremented_arg(self, FIELD_ACCESS_MAP)
+        return Kern.incremented_arg(self)
 
     @property
     def written_arg(self):
         ''' Returns the argument corresponding to a field or operator that has
         WRITE access '''
-        return Kern.written_arg(self, FIELD_ACCESS_MAP)
+        return Kern.written_arg(self)
```
arporter: This line is not covered but, since all we're doing is calling the `written_arg` method of the parent class, this method can be removed altogether. In fact, the same applies for `incremented_arg` above.

src/psyclone/dynamo0p3.py
```diff
@@ -646,31 +638,31 @@ def __init__(self, arg_type):
...
-               and self._access_descriptor.name.lower() == "gh_inc":
+               and self._access_descriptor.name == AccessType.INC:
                 raise ParseError(
                     "It does not make sense for a field on a discontinuous "
                     "space ({0}) to have a 'gh_inc' access".
                     format(self._function_space1.lower()))
             # TODO: extend for "gh_write"
             if self._function_space1.lower() in CONTINUOUS_FUNCTION_SPACES \
-               and self._access_descriptor.name.lower() == "gh_readwrite":
+               and self._access_descriptor.name == AccessType.READWRITE:
```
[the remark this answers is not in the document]
TeranIvy: It did not, in fact we discourage using it, see issue #138.

src/psyclone/tests/psyGen_test.py
hiker: @arporter , there are several places where I set a default api in a test function, mostly in 'generic' (non-api-specific) tests like alggen_test etc. By now in nearly all places this is not necessary for the tests anymore. Short explanation:
Originally I used Config.get().api_conf() in dynamo* and gocean*. For this to work I indeed needed to change the default API in the tests to make sure the right api-specific config will be picked up. But then I realised that in e.g. dynamo0p3 I should always get the dynamo0.3 config, so I changed the code to use: config.get().api_conf("dynamo0.3"). Since the API-files (dynamo*, gocean*) are responsible for translating the api-specific strings to the AccessType, there is no need to have a default API specified ... unless there is an error. In case of an error (especially if detected by a baseclass e.g. in psygen) a default API is needed to translate AccessType back to the api-specific string.

As summary: in all test cases that do not trigger an (AccessType-related) error, there is no need to change the API anymore, only the tests which test (AccessType-related) exception handling need a default being defined in order for the right error message.

So I could basically remove nearly all settings of an API.  The fixture in each API-file still makes sense (since e.g. a gocean1p0 test will likely test gocean1p0 related errors).

So would you be ok if I just remove the unnecessary settings of the api (with the only exception being tests that raise accesstype-related errors) - and since we have a fixture, that would only leave some non-API-specific tests (psyGen_test, alggen_test) where it is actually required.

arporter: Yes, that sounds good to me!

src/psyclone/configuration.py
```diff
@@ -121,6 +121,18 @@ def get(do_not_load_file=False):
                 Config._instance.load()
         return Config._instance
 
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def get_repository_config_file():
+        '''This function returns the absolute path to the config file included
+        in the PSyclone repository. It is used by the testing framework to make
+        sure all tests get the same config file (see tests/config_tests for the
+        only exception).
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/configuration.py
```diff
@@ -574,7 +606,104 @@ def get_default_keys(self):
         return self._config.defaults()
 
 
-class DynConfig(object):
+# =============================================================================
+class APISpecific(object):
+    '''A base class for functions that each API-specific class must provide.
+    At the moment that is a function access_mapping().
+    :param section: :py:class:`configparser.SectionProxy`
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
hiker: If I understood you correct, you actually agreed to api_specific_name for the function, and then renaming the local variables. I've done that.

src/psyclone/dynamo0p3.py
```diff
@@ -4203,25 +4200,39 @@ def __init__(self, alg_invocation, idx):
...
                 for scalar in loop.args_filter(
-                        arg_types=VALID_SCALAR_NAMES,
-                        arg_accesses=VALID_REDUCTION_NAMES, unique=True):
+                        arg_types=GH_VALID_SCALAR_NAMES,
+                        arg_accesses=AccessType.get_valid_reduction_modes(),
+                        unique=True):
                     global_sum = DynGlobalSum(scalar, parent=loop.parent)
                     loop.parent.children.insert(loop.position+1, global_sum)
 
+    def unique_declarations(self, datatype, access=None):
```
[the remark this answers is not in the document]
hiker: Outdated, the whole function in dynamo0p3 has been deleted, it is all handled in psyGen now.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
hiker: Outdated, function removed. psyGen now raises InternalError in this case.

src/psyclone/psyGen.py
hiker: The risk of "you touch it, you document it", since I don't really know the dynamo API - sorry, I've applied your changes.

src/psyclone/configuration.py
```diff
@@ -574,7 +606,106 @@ def get_default_keys(self):
...
+    def create_dict_from_string(input_str):
+        '''Takes an input string in the format:
+        key1:value1, key2:value2, ...
+        and creates a dictionary with the key,value pairs.
+        Spaces are removed.
+        :param str input_str: The input string.
+        :returns: A dictionary with the key,value pairs from the input string.
+        :rtype: dict.
+        :Raises ConfigurationError if the input string contains an entry
+                that is does not have a ":".
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/configuration.py
```diff
@@ -574,7 +606,106 @@ def get_default_keys(self):
...
+        '''Returns the reverse mapping of a PSyclone internal access type
+        to the API specific string, e.g.: AccessType.READ to 'gh_read'.
+        This is used to provide the user with API specific error messages.
+        :returns: The mapping of access types to API-specific strings.
+        :rtype: Dictionary of strings
+        '''
+        return self._reverse_access_mapping
+
+    def get_valid_accesses_api(self):
+        '''Returns the sorted, API-specific sorted names of all valid access
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/configuration.py
```diff
@@ -574,7 +606,106 @@ def get_default_keys(self):
         return self._config.defaults()
 
 
-class DynConfig(object):
+# =============================================================================
+class APISpecificConfig(object):
+    '''A base class for functions that each API-specific class must provide.
+    At the moment that is a function access_mapping().
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/dynamo0p3.py
```diff
@@ -696,26 +684,29 @@ def __init__(self, arg_type):
...
                     "In the dynamo0.3 API each meta_arg entry must have 2 "
                     "arguments if its first argument is gh_{{r,i}}scalar, but "
                     "found {0} in '{1}'".format(len(arg_type.args), arg_type))
             # Test allowed accesses for scalars (read_only or reduction)
-            if self._access_descriptor.name not in ["gh_read"] + \
-               VALID_REDUCTION_NAMES:
+            if self._access_type not in [AccessType.READ] + \
+               AccessType.get_valid_reduction_modes():
+                rev_access_mapping = api_config.get_reverse_access_mapping()
+                api_name = rev_access_mapping[self._access_type]
```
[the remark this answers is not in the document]
hiker: Done (and as indicated previously, I also renamed the api_name function in AccessType).

src/psyclone/dynamo0p3.py
```diff
@@ -4201,25 +4196,28 @@ def __init__(self, alg_invocation, idx):
...
+        if datatype not in GH_VALID_ARG_TYPE_NAMES:
             raise GenerationError(
                 "unique_proxy_declarations called with an invalid datatype. "
                 "Expected one of '{0}' but found '{1}'".
-                format(str(VALID_ARG_TYPE_NAMES), datatype))
-        if access and access not in VALID_ACCESS_DESCRIPTOR_NAMES:
+                format(str(GH_VALID_ARG_TYPE_NAMES), datatype))
+        if access and not isinstance(access, AccessType):
+            api_config = Config.get().api_conf("dynamo0.3")
+            valid_names = api_config.get_valid_accesses_api()
```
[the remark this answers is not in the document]
hiker: Done (in psyGen, this function in dynamo03 has been removed).

src/psyclone/psyGen.py
```diff
@@ -2278,13 +2279,16 @@ def view(self, indent=0):
 
     def _get_reductions_list(self, reduction_type):
         '''Return the name of all scalars within this region that require a
-        reduction of type reduction_type. Returned names will be unique. '''
+        reduction of type reduction_type. Returned names will be unique.
+        :param: reduction_type The reduction type (e.g. AccessType.SUM) to
```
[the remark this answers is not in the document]

src/psyclone/dynamo0p3.py
```diff
@@ -4201,25 +4196,28 @@ def __init__(self, alg_invocation, idx):
...
+                        unique=True):
                     global_sum = DynGlobalSum(scalar, parent=loop.parent)
                     loop.parent.children.insert(loop.position+1, global_sum)
 
     def unique_proxy_declarations(self, datatype, access=None):
         ''' Returns a list of all required proxy declarations for the
         specified datatype.  If access is supplied (e.g. "gh_write")
         then only declarations with that access are returned. '''
-        if datatype not in VALID_ARG_TYPE_NAMES:
+        if datatype not in GH_VALID_ARG_TYPE_NAMES:
```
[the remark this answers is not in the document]
hiker: Done (outdated)

src/psyclone/core/tests/access_type_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+    assert str(AccessType.INC) == "INC"
+    assert str(AccessType.WRITE) == "WRITE"
+    assert str(AccessType.READ) == "READ"
+    assert str(AccessType.READWRITE) == "READWRITE"
+    assert str(AccessType.SUM) == "SUM"
+
+
+def test_api_name():
+    '''Tests api_name(), i.e. conversion to an
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/alggen_test.py
```diff
@@ -411,25 +419,31 @@ def test_single_invoke_dynamo0p1():
         specified in an invoke call for the dynamo0.1 API. We use the
         generate function as parse and PSyFactory need to be called before
         AlgGen so it is simpler to use this. '''
+    Config.get().api = "dynamo0.1"
```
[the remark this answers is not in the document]
hiker: As discussed elsewhere: it turns out in nearly all cases setting the api is not necessary anymore (since it is actually only used in case of errors), so while I had started using monkeypatch (and finally started to understand what it does ;) ), I could remove all those assignments. It turns out that the error testing is done in functions that set the right API in a fixture.

src/psyclone/tests/alggen_test.py
```diff
@@ -411,25 +419,31 @@ def test_single_invoke_dynamo0p1():
...
     assert "CALL invoke_0_testkern_type(f1, f2, m1)" in gen
+    Config.get().api = "dynamo0.3"
 
 
 def test_zero_invoke_dynamo0p1():
     ''' Test that an exception is raised if the specified file does
         not contain any actual invoke() calls. We use the generate
         function as parse and PSyFactory need to be called before
         AlgGen so it is simpler to use this. '''
+    Config.get().api = "dynamo0.1"
```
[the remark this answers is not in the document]
hiker: Outdated.

src/psyclone/tests/config_test.py
```diff
@@ -117,6 +121,13 @@ def int_entry(request):
     return request.param
 
 
+def test_get_repo_config_file():
+    '''Check that we can find the expected config file in the repository.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/config_test.py
```diff
@@ -491,6 +511,61 @@ def test_incl_path_errors(tmpdir):
...
+
+    mapping = APISpecificConfig.create_dict_from_string("")
+    assert mapping == {}
+
+    # The function only uses the first ":" :
+    mapping = \
+        APISpecificConfig.create_dict_from_string("k1:v1, k2:v2:something")
+    assert mapping == {"k1": "v1", "k2": "v2:something"}
+
+    # Tests errors: '=' is not valid, will
```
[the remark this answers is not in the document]
hiker: Fixed.

doc/configuration.rst
```diff
@@ -113,6 +121,33 @@ REPROD_PAD_SIZE         If generating code for reproducible OpenMP reductions,
...
+access_mapping          This field defines the strings that are used by a
+                        particular API to indicate write, read, ... access. Its
+                        value is a comma separated list of access-string:access
+                        pairs, e.g.:
+
+                        ``gh_read:read, gh_write:write, gh_rw:readwrite,
+                        gh_inc:inc, gh_sum:sum``
+
+                        At this stage these 5 types are defined for read, write,
+                        read+write, increment and summation access by PSyclone.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p1_transformations_test.py
```diff
@@ -10,13 +10,21 @@
...
 from psyclone.parse.algorithm import parse
 from psyclone.psyGen import PSyFactory
 from psyclone.transformations import OMPParallelTrans
 
 TEST_API = "dynamo0.1"
 
 
+@pytest.fixture(scope="module", autouse=True)
+def setup():
+    '''Make sure that all tests here use dynamo0.3 as API.'''
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p3_lma_test.py
```diff
@@ -42,6 +42,7 @@
 import os
 import pytest
 from fparser import api as fpapi
+from psyclone.core.access_type import AccessType
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p3_stubgen_test.py
```diff
@@ -98,12 +98,12 @@ def test_dynscalars_err(monkeypatch):
     arg = kernel.arguments.args[1]
     arg._type = "invalid-scalar-type"
     # Monkeypatch the list of supported scalar types to include this one
-    monkeypatch.setattr(dynamo0p3, "VALID_SCALAR_NAMES",
+    monkeypatch.setattr(dynamo0p3, "GH_VALID_SCALAR_NAMES",
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1807,13 +1814,27 @@ def test_invoke_uniq_declns_invalid_access():
...
+    with pytest.raises(InternalError) as excinfo:
         psy.invokes.invoke_list[0].unique_declarations("gh_field",
                                                        access="invalid_acc")
     assert 'unique_declarations called with an invalid access type' \
         in str(excinfo.value)
 
 
+def test_invoke_uniq_declns_valid_access():
+    ''' tests that valid access modes ('read', 'write') are
+    accepted by  Invoke.unique_declarations() is called.'''
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1807,13 +1814,27 @@ def test_invoke_uniq_declns_invalid_access():
...
+    _, invoke_info = parse(os.path.join(BASE_PATH,
+                                        "1.7_single_invoke_2scalar.f90"),
+                           api=TEST_API)
+    psy = PSyFactory(TEST_API, distributed_memory=True).create(invoke_info)
+    print("type is", type(psy.invokes.invoke_list[0]))
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1807,13 +1814,27 @@ def test_invoke_uniq_declns_invalid_access():
...
+    psy.invokes.invoke_list[0].unique_declarations("gh_field",
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -3018,17 +3027,17 @@ def test_arg_descriptor_init_error():
     arg_type = field_descriptor._arg_type
     # Now try to trip the error by making the initial test think
     # that GH_INVALID is actually valid
-    from psyclone.dynamo0p3 import VALID_ARG_TYPE_NAMES, DynArgDescriptor03
+    from psyclone.dynamo0p3 import GH_VALID_ARG_TYPE_NAMES, DynArgDescriptor03
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/gocean0p1_test.py
```diff
@@ -39,11 +39,12 @@
...
 import pytest
+from fparser.api import parse as parse1
+from psyclone.core.access_type import AccessType
 from psyclone.psyGen import PSyFactory
 from psyclone.gocean0p1 import GODescriptor, GOKernelType
 from psyclone.parse.algorithm import parse
 from psyclone.parse.utils import ParseError
-from fparser.api import parse as parse1
 
 API = "gocean0.1"
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -1114,6 +1122,12 @@ def test_raw_arg_list_error(monkeypatch):
         _ = kern.arguments.raw_arg_list()
     assert ("Kernel compute_z_code, argument z_fld has unrecognised type: "
             "'broken'" in str(err))
+    # Also test that assigning a non-AccessType value to a kernel argument
```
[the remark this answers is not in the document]
hiker: Done.
The strange error message is caused by a difference in output between python2 and 3. Similar tests exist elsewhere (psyGen_test.py: 'assert "Unrecognized attribute type " in str(error.value)'), but I have replaced this test with a regular expression tests that also checks for 'str' in the output (i.e. independent of 'class' (python3 iirc) or 'type' (python2) in the string).

src/psyclone/tests/gocean1p0_transformations_test.py
hiker: Done.

src/psyclone/tests/gocean1p0_transformations_test.py
hiker: Outdated.

src/psyclone/tests/kernel_transformation_test.py
hiker: Done.

src/psyclone/tests/psyGen_test.py
hiker: Outdated.

src/psyclone/tests/psyGen_test.py
hiker: Done.

src/psyclone/tests/psyGen_test.py
hiker: Nope :)

src/psyclone/tests/psyGen_test.py
hiker: Outdated.

src/psyclone/tests/psyGen_test.py
hiker: Outdated.

src/psyclone/tests/psyGen_test.py
hiker: Outdated.

src/psyclone/tests/psyGen_test.py
hiker: Outdated.

src/psyclone/tests/psyGen_test.py
hiker: OK, I changed this everywhere, I've added 'outdated' to all the corresponding comments here.

src/psyclone/dynamo0p3.py
hiker: Done (except that those functions here had @property, which psyGen did not have ... but since I don't like @property anyway, I've added '()' to all accesses of the former property).

hiker:
Codecov/patch still shows only 99.13% target hit ... but I have to admit I can't figure out how to see which new lines were not hit. I see dynamo0p3 in red, but when I click on the red rectangle, I get: "File/folder was not found on base commit. No compare to display. " And for dynamo0.3 I see a +0.07% - in short: I have no idea how to figure out what might be wrong there. So I'll mark this as ready for review, but if there is something that I missed, please enlighten me how I can test this myself.

src/psyclone/configuration.py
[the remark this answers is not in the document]
arporter: Almost! You need a colon after "ConfigurationError" :-)

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
arporter: Great - thank you.

src/psyclone/psyGen.py
arporter: Ah, I see. I hadn't thought of that drawback before. Mind you, the chances are that nobody else is much the wiser either at this point (since if it hasn't got an interface description then we haven't touched it in quite a while).

src/psyclone/core/tests/access_type_test.py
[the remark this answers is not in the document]
arporter: Great. Thanks for taking the trouble to change it all again.

src/psyclone/tests/alggen_test.py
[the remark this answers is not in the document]
arporter: I'm a bit confused here - I see that you've changed the test to set the API using monkeypatch BUT it's being set to 'gocean1.0' (lines 422 and 438). I guess that the errors that those tests are provoking happen before the kernel meta-data is parsed so we get away with it?

src/psyclone/tests/alggen_test.py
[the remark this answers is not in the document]
arporter: See previous comment - it's being set to "gocean1.0" instead of "dynamo0.1"

src/psyclone/tests/gocean1p0_test.py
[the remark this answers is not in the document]
arporter: Great, thanks. Picky, but could you remove the "Also" from the comment on line 1136?

src/psyclone/tests/alggen_test.py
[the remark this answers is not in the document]
arporter: Having got through to the end of all the comments I think I finally understand what's going on :-) Are this line and the one I comment on below just hangovers that should be removed?

arporter:
All tests, including compilation, pass.

re. Coverage, what you want is the 'Diff' tab on CodeCov. This then shows (if you scroll down) that the code that is missed is the `updated_arg` method in dynamop3.py. It looks as though this is never called which is a bit suspicious. Maybe it's completely unnecessary?

src/psyclone/dynamo0p1.py
```diff
@@ -379,6 +380,7 @@ def gen_code(self, parent):
                        position=["before", position])
 
 
+
```
arporter: pycodestyle now complains that there are too many blank lines.

src/psyclone/dynamo0p3.py
```diff
@@ -734,24 +726,24 @@ def function_space_to(self):
         ''' Return the "to" function space for a gh_operator. This is
         the first function space specified in the metadata. Raise an
         error if this is not an operator. '''
-        if self._type in VALID_OPERATOR_NAMES:
+        if self._type in GH_VALID_OPERATOR_NAMES:
```
arporter: Not strictly modified by you but since you've altered the `if`, please could you fix pylint's complaint about the unnecessary `else` below.

src/psyclone/dynamo0p3.py
```diff
@@ -734,24 +726,24 @@ def function_space_to(self):
...
-                "this is a '{1}'".format(VALID_OPERATOR_NAMES, self._type))
+                "this is a '{1}'".format(GH_VALID_OPERATOR_NAMES, self._type))
 
     @property
     def function_space_from(self):
         ''' Return the "from" function space for a gh_operator. This is
         the second function space specified in the metadata. Raise an
         error if this is not an operator. '''
-        if self._type in VALID_OPERATOR_NAMES:
+        if self._type in GH_VALID_OPERATOR_NAMES:
```
arporter: Again, pylint doesn't like the `else` below.

src/psyclone/dynamo0p3.py
```diff
@@ -760,9 +752,9 @@ def function_space(self):
         are 2 function spaces, return function_space_from. '''
         if self._type == "gh_field":
             return self._function_space1
-        elif self._type in VALID_OPERATOR_NAMES:
+        elif self._type in GH_VALID_OPERATOR_NAMES:
```
arporter: This can just be `if ...` and same again below.

src/psyclone/dynamo0p3.py
```diff
@@ -776,10 +768,10 @@ def function_spaces(self):
         spaces, we return both. '''
         if self._type == "gh_field":
             return [self.function_space]
-        elif self._type in VALID_OPERATOR_NAMES:
+        elif self._type in GH_VALID_OPERATOR_NAMES:
```
arporter: Again, this can just be `if` and same again below.

src/psyclone/dynamo0p3.py
```diff
@@ -5790,11 +5800,7 @@ def _upper_bound_fortran(self):
     def has_inc_arg(self, mapping=None):
         ''' Returns True if any of the Kernels called within this loop
         have an argument with INC access. Returns False otherwise. '''
-        if mapping is not None:
-            my_mapping = mapping
-        else:
-            my_mapping = FIELD_ACCESS_MAP
-        return Loop.has_inc_arg(self, my_mapping)
+        return Loop.has_inc_arg(self)
```
arporter: Can we get rid of the `mapping` argument to `has_inc_arg` now?

src/psyclone/dynamo0p3.py
```diff
@@ -5830,18 +5836,18 @@ def _halo_read_access(self, arg):
                     "currently unsupported for kernels with stencil "
                     "accesses. Found '{0}'.".format(self._upper_bound_name))
             return self._upper_bound_name in ["cell_halo", "ncells"]
-        if arg.type in VALID_SCALAR_NAMES:
+        if arg.type in GH_VALID_SCALAR_NAMES:
```
arporter: Please replace the `elif`s that are after `return`s with just `if`

src/psyclone/dynamo0p3.py
```diff
@@ -8495,19 +8491,21 @@ def intent(self):
                  by the kernel argument metadata
         :rtype: str
         '''
-        if self.access == "gh_read":
+        if self.access == AccessType.READ:
```
arporter: Please simplify the `elif`s to `if`s where possible to keepy pylint happy.

src/psyclone/configuration.py
```diff
@@ -574,7 +606,108 @@ def get_default_keys(self):
         return self._config.defaults()
 
 
-class DynConfig(object):
+# =============================================================================
+class APISpecificConfig(object):
+    '''A base class for functions that each API-specific class must provide.
+    At the moment this is just the function 'access_mapping' that maps between
+    API-specific access-descriptor strings and the PSyclone internal
```
arporter: pylint complains about trailing white space here.

src/psyclone/tests/psyGen_test.py
```diff
@@ -2569,6 +2579,7 @@ def test_loop_props():
 def test_node_abstract_methods():
     ''' Tests that the abstract methods of the Node class raise appropriate
     errors. '''
+    from psyclone.psyGen import Node
```
arporter: pylint complains that `Node` has already been imported.

arporter:
There are a handful of pylint errors to address (most of them aren't strictly yours) and a few remaining threads from previous reviews that we need to resolve. Once that's done then this can be merged.

hiker:
> the code that is missed is the `updated_arg` method in dynamop3.py. It looks as though this is never called which is a bit suspicious. Maybe it's completely unnecessary?

Yes, I agree that it appears to be unused, so I removed it which then made another function unused (except for in tests, so I removed the tests as well).

src/psyclone/dynamo0p1.py
hiker: Done.

src/psyclone/dynamo0p3.py
hiker: Done. I admit I noticed those warnings, but I usually prefer to have patches as 'clean' as possible, so that it is easy for the reviewer to see the important changes. I was thinking of a separate pull request that improves coding style ;) But, if you want it, you get it :)  Done.

src/psyclone/dynamo0p3.py
hiker: Done.

src/psyclone/dynamo0p3.py
hiker: Done.

src/psyclone/dynamo0p3.py
hiker: Done.

src/psyclone/dynamo0p3.py
hiker: I removed the whole function, which is not necessary anymore since it only calls the function in the base class.

src/psyclone/dynamo0p3.py
hiker: Done.

src/psyclone/dynamo0p3.py
hiker: Done.

src/psyclone/configuration.py
hiker: And I just learned that if you use pycharm, and 'go to the end of a line' ... it goes to the last non-space character in a line (i.e. before any trailing white-space) ... strange. Anyway, fixed :)

src/psyclone/tests/psyGen_test.py
hiker: Fixed.

hiker:
> re. Coverage, what you want is the 'Diff' tab on CodeCov. 
OK, that is good to know ... though as far as I can tell it still means you have to manually scroll through hundreds of lines, trying to spot the ones that are red (which are often and have a "+" (i.e. were added/modified in the patch). But ok, I learned something :) I wonder if it would be worth to add some of the details to either the dev-guide, or the wiki (code review part) at least.

src/psyclone/configuration.py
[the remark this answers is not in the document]
hiker: Gee, that was embarrassing (then again ... good case for being able to create an API guide, since it would allow a dev to easily check the correctness of markups). Done.

src/psyclone/tests/alggen_test.py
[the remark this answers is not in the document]
hiker: Yes, left over ... and I am wildly cursing here: I grep'ed to find all settings of the api that I had added to the tests, but used:
```
grep "Config.get().api" *py
```
but the monkeypatched lines were not detected by the grep :( Note to self: `grep "Config.get().*api"` :)

src/psyclone/tests/alggen_test.py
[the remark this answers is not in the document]
hiker: Removed.

src/psyclone/tests/gocean1p0_test.py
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/psyGen_test.py
hiker: Done.

hiker:
Ready to go again.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
arporter: Yes, I did - thanks!

src/psyclone/tests/alggen_test.py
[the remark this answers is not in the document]
arporter: :-)

src/psyclone/configuration.py
arporter: I don't even know what pycharm is :-)

arporter:
I'll merge this once the final Travis checks have completed.

PR #389, opened 2019-05-22
Nemo config

hiker:
Fixes #386: the variable name and related information are stored in the config file now, making the nemo API easy to adjust for other codes that use different variable names. E.g. in ROMS I now get lat and lon loops identified in a schedule. Note that I do not check if the values 'lat', 'lon' etc are used, as far as I can tell it should be able to change e.g. 'lat' to 'latitude' in the config file (except for the current test cases, that test for 'lat' etc.

PR #394, opened 2019-05-27
Fix graphviz detection

### LINKED ISSUE #392: Availability of graphiz is tested incorrectly
ATm conftest.py uses the following code to test if graphviz is available:
```
import sys
return "graphviz" in sys.modules
```
This only works if graphviz has been imported. If only a subset of the tests is run, this might not be the case. I think the correct solution is to try-import here.

hiker:
Fixes #392 (test if graphviz is available).


hiker:
Oops - I just reliased that I had this branched off the nemo-config branch. So #389 must be merged first.

Issue #364, opened 2019-04-25
Re-structure class hierarchy for Kernels


arporter:
Currently we have:

    Node <-- Call <-- Kern <-- NemoKern
                 \        \____DynKern
                  \ BuiltIn
                        
but actually, all of our current `BuiltIn`s are implemented in-line and not as calls. Similarly, `NemoKern`s are currently in-line code. Therefore we propose to re-structure like so:

    Node <-- Kern <--BuiltIn
                 \___CodedKern

where `Kern` now has an attribute that captures whether or not it is in-lined or a call.

TeranIvy:
`CodedKern` is a bit confusing name for what is essentially still a subroutine **call** and not really "coding" Kernel contents. Kernel Stub and Kernel Generation from #393 would be good examples of "coding" (adding content) to a Kernel subroutine.

arporter:
@TeranIvy `CodedKern` is so named because it is a kernel that is implemented in code. Whether it is called or is in-line is an attribute (and may change as the result of transformations. Does that make sense?

arporter:
All `dynamo0p3_*` tests pass now.

PR #404, opened 2019-06-10
Re-jig the Kernel class hierarchy (for #364)

arporter:


arporter:
I need to grep through the Developers' guide (and maybe the source too) and check for existing refs to `Call` and `Kern`.

arporter:
I've now done the Developers' Guide, the examples and the source code. In the latter I've taken the opportunity to tidy up some of the docstrings that I was touching.
This is now ready for first review.

src/psyclone/psyGen.py
```diff
@@ -3476,37 +3478,46 @@ def arguments(self):
     @property
     def name(self):
         '''
-        :returns: the name of the kernel associated with this call.
+        :returns: the name of the kernel.
```
sergisiso: Capitalise

sergisiso: well, looking at other docstrings they are not always capitalised, so I guess is more of a question, should they be?

src/psyclone/tests/gocean1p0_config_test.py
```diff
@@ -52,12 +52,24 @@ def setup():
...
-def teardown_function():
-    '''This teardown function is called at the end of all tests and makes
-    sure that we wipe the Config object so we get a fresh/default one
-    for any further test (and not a left-over one from a test here).
+@pytest.fixture(scope="function", autouse=True)
+def clear_config_instance():
+    ''' The tests in this module all assume that there is no pre-existing
+    Config object, so this fixture ensures that the config instance is
+    deleted before and after each test function. The latter makes sure that
+    any other test executed next will automatically reload the default
```
sergisiso: Does it reload the default one? The `setup()` fixture is only executed on module entry, I think this function just uninitialized the Config but it doesn't seem to reload the 'default'

src/psyclone/psyGen.py
arporter: Strictly speaking, I don't think they should as the contents appear after a ":" in the generated documentation (https://psyclone-ref.readthedocs.io/en/latest/autogenerated/psyclone.dynamo0p3.html#psyclone.dynamo0p3.DynKernMetadata). I must admit I'm not too worried but we should probably be consistent at least.

src/psyclone/dynamo0p1.py
```diff
@@ -300,7 +300,16 @@ def __init__(self):
...
+        '''
+        Load this DynKern object with state pulled from the call object.
+
+        :param call: details of the Algorithm-layer call of this Kernel.
+        :type call: :py:class:`psyclone.parse.algorithm.KernelCall`
+        :param parent: parent of this kernel node in the PSyIR.
+        :type parent: :py:class:`psyclone.dynamo0p1.DynLoop` or NoneType.
+
+        '''
+        super(DynKern, self).__init__(DynKernelArguments, call, parent)
```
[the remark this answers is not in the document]
arporter: I don't know why it's this way. We have an issue to remove dynamo0p1 altogether though :-)

src/psyclone/tests/gocean1p0_config_test.py
arporter: The use of `yield` means that control will return to this fixture after each test has been executed. We don't need to load the 'default' config. because that will happen automatically if `_instance` is `None`.

arporter:
I think that the documentation you refer to is correct but I agree that it is not very clear that it is talking about objects used to capture information from the parsed "call invoke" (algorithm) and parsed kernel meta-data. Unfortunately we seem to have at least three classes that capture information about Kernels although only one of them sub-classes `Node`. I've tweaked the documentation a bit to try to make the distinction clearer.

arporter:
Whether or not a (coded) kernel is called or in-lined will become a property of the `CodedKern` object but I've not done that in this PR.

Well spotted on the `calls()` and `kern_calls()` methods. I did think about changing to having a single interface where you could specify what sort of kernels you want to find but then I thought that might clash with #411 so I've simply re-named the existing methods.

arporter:
Assuming Travis is happy then I think this is ready for 2nd review.

PR #400, opened 2019-06-03
Towards #385. PSyIR Fortran front-end Do_Constructs handler.

sergisiso:
This PR:
- Adds a generic Fortran do loop handler to the ASTProcessor based on the implementation of the NemoLoop parser.
- start/stop/step expressions of Loop are now in-tree. (Just the do_loop bounds)
```
Loop[type='None', field_space='None', it_space='None']
        Literal[value:'1']   # Child 1 is the Start bound
        Reference[name:'un'] # Child 2 is the Stop bound
        Literal[value:'1'] # Child 3 is the Stepping value
        Schedule[]  # Child 4 is the Loop body (always a Schedule)
            Assignment[]
                Reference[name:'rtmp1']
                Literal[value:'3']
```
This is still not able to represent different types of Loops, it is limited to Fortran Do Loops.

PR #424, opened 2019-07-04
Variable Access API

hiker:
This pull requests add functionality to determine variable accesses in the PSyIR. As discussed the access data stored now includes a pointer to the original node.

The API is used to properly determine thread-private variables in omp transforms.


src/psyclone/dynamo0p3.py
```diff
@@ -6156,6 +6157,25 @@ def __init__(self):
...
+            # the access mode defined. So take it from the arg_descriptor:
+            if arg.vector_size > 0:
+                # It's an array, so add an arbitrary index value for the
+                # stored indices (which is at this stage the only way to
+                # indicate an array access).
+                var_accesses.add_access(arg.name, arg.access, self, [1])
+            else:
+                var_accesses.add_access(arg.name, arg.access, self)
+        super(DynKern, self).reference_accesses(var_accesses)
+        var_accesses.next_location()
```
arporter: Please could you explain what this line is for (in a comment).

src/psyclone/gocean1p0.py
```diff
@@ -883,6 +883,17 @@ def __init__(self):
         # Get a reference to the namespace manager
         self._name_space_manager = NameSpaceFactory().create()
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. All accesses are marked
+        according to the kernel declaration.
+        :param var_accesses: \
```
arporter: Again, please document the purpose of this argument as well as its type.

src/psyclone/nemo.py
```diff
@@ -352,6 +353,14 @@ def view(self, indent=0):
         '''
         print(self.indent(indent) + self.coloured_text + "[]")
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. It calls the corresponding
+        kernel schedule function.
+        :param var_accesses: \
```
arporter: Blank line before `:param:` please and please document the purpose of `var_accesses` as wel as its type.

src/psyclone/nemo.py
```diff
@@ -472,6 +481,31 @@ def __str__(self):
         result += "EndLoop"
         return result
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. The loop variable is
+        set as READ and WRITE. Then the loop body's access is added.
+        TODO: The start, stop and step values are only strings, so we
```
arporter: Please include the issue number for this TODO (and create said issue if needed).

src/psyclone/nemo.py
```diff
@@ -472,6 +481,31 @@ def __str__(self):
...
+        can't get access information. It might then also be possible to
+        just fall back to Loop.reference_accesses (which then should work).
+        :param var_accesses: \
```
arporter: Blank line and please document what `var_accesses` is for as well as its type.

src/psyclone/nemo.py
```diff
@@ -472,6 +481,31 @@ def __str__(self):
...
+        var_accesses.next_location()
+        # TODO: atm start, stop and step are just strings, so we can't
+        # get any variable information
+        # self._start.reference_accesses(var_accesses)
+        # self._stop.reference_accesses(var_accesses)
+        # self._step.reference_accesses(var_accesses)
+        # We also can't call the base class, since it will accesses
+        # start stop etc :(
+        # super(NemoLoop, self).reference_accesses(var_accesses)
+        for child in self.children:
```
arporter: Nice :-)

src/psyclone/psyGen.py
```diff
@@ -1532,6 +1532,17 @@ def update(self):
         for child in self._children:
             child.update()
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. The default implementation
+        just recurses down to all children:
```
arporter: picky (I know; surprising) but full stop rather than colon please.

src/psyclone/psyGen.py
```diff
@@ -2447,24 +2458,40 @@ def _get_private_list(self):
...
-                if variable_name.lower() not in result:
-                    result.append(variable_name.lower())
-        return result
+                result.add(variable_name.lower())
+
+        # Now determine scalar variables that must be private:
+        var_accesses = VariablesAccessInfo()
+        self.reference_accesses(var_accesses)
+        for var_name in var_accesses.get_all_vars():
+            accesses = var_accesses.get_varinfo(var_name).get_all_accesses()
```
arporter: Rather than having to call `get_varinfo` would it be worth having some sort of shortcut since this is going to be a very common operation. c.f. Sergi's work on the PSyIR, could we implement things such that `var_accesses[var_name]` does what we want?

src/psyclone/psyGen.py
```diff
@@ -3174,6 +3202,32 @@ def __str__(self):
         result += "EndLoop"
         return result
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. It combines the data from
+        the loop bounds (start, stop, end step), as well as the loop body.
+        The loop variable is marked as READWRITE, start, stop, step as READ.
+        :param var_accesses: \
```
arporter: Blank line and please document purpose of `var_accesses` as well as its type.

src/psyclone/psyGen.py
```diff
@@ -3174,6 +3202,32 @@ def __str__(self):
...
+            :py:class:`psyclone.core.access_info.VariablesAccessInfo`
+        '''
+
+        # It is important to first add the WRITE access, since this way
+        # the dependency analysis for declaring openmp private variables
+        # will automatically declare the loop variables to be private
+        # (write access before read)
+        var_accesses.add_access(self.variable_name, AccessType.WRITE, self)
+        var_accesses.add_access(self.variable_name, AccessType.READ, self)
+        # TODO: self._start/stop/step are not defined at this stage (at least
```
arporter: Please include an Issue number here. Also, the loop bounds might be arithmetic expressions (e.g. `jpi+1`). I guess we're waiting on the proper loop support that  @sergisiso is implementing in #400.

src/psyclone/psyGen.py
```diff
@@ -3326,6 +3380,16 @@ def view(self, indent=0):
         for entity in self._children:
             entity.view(indent=indent + 1)
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. The API specific classes
+        add the accesses to the arguments. So the code here only calls
+        the baseclass, and increases the location.
+        :param var_accesses: \
```
arporter: Blank line and description please.

src/psyclone/psyGen.py
```diff
@@ -4808,6 +4878,25 @@ def __str__(self):
             result += str(entity)
         return result
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. It combines the data from
+        the condition, if-body and (if available) else-body. This could
+        later be extended to handle cases where a variable is only written
+        in of the two branches.
```
arporter: s/in of/in one of/

src/psyclone/psyGen.py
```diff
@@ -4808,6 +4878,25 @@ def __str__(self):
...
+        :param var_accesses: \
```
arporter: Blank line and description please.

src/psyclone/psyGen.py
```diff
@@ -5863,6 +5952,29 @@ def __str__(self):
             result += str(entity)
         return result
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information from this node. The assigned-to
+        variable will be set to 'WRITE'.
+        :param var_accesses: \
```
arporter: Blank line and description please.

src/psyclone/psyGen.py
```diff
@@ -5863,6 +5952,29 @@ def __str__(self):
...
+            :py:class:`psyclone.core.access_info.VariablesAccessInfo`
+        '''
+
+        # It is important that a new instance is used to handle the LHS,
+        # since an assert in 'change_read_to_write' makes sure that there
```
arporter: s/an assert/a check/

src/psyclone/psyGen.py
```diff
@@ -5863,6 +5952,29 @@ def __str__(self):
...
+        self.lhs.reference_accesses(accesses_left)
+
+        # Now change the (one) access to the assigned variable to be WRITE:
+        var_info = accesses_left.get_varinfo(self.lhs.name)
+        var_info.change_read_to_write()
+
+        # Merge the data (that shows now WRITE for the variable) with the
+        # parameter to this function:
+        var_accesses.merge(accesses_left)
+        self.rhs.reference_accesses(var_accesses)
```
arporter: Does this work OK for e.g. `my_var = my_var + 1`? i.e. `my_var` is read before it is written. Would the `location` for the read and write be the same in this situation?

src/psyclone/psyGen.py
```diff
@@ -5927,6 +6039,14 @@ def view(self, indent=0):
     def __str__(self):
         return "Reference[name:'" + self._reference + "']\n"
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information from this node. The assigned-to
+        variable will be set to 'WRITE'.
+        :param var_accesses: \
```
arporter: Blank line and description please.

src/psyclone/psyGen.py
```diff
@@ -6265,6 +6385,28 @@ def __str__(self):
...
+        # does not recurse to the children, so at this stage no
+        # index information has been stored:
+        list_indices = []
+        for child in self._children:
+            child.reference_accesses(var_accesses)
+            list_indices.append(child)
+
+        if list_indices:
+            var_info = var_accesses.get_varinfo(self._reference)
+            var_info.get_all_accesses()[0].set_indices(list_indices)
```
arporter: I'm not comfortable with the fact that the first entry in the list of accesses is "special" and is the only one with information that the variable is an array. Ideally I think we'd solve this with a link to the appropriate entry in the SymbolTable but we don't yet use that in the NEMO API. Maybe `var_info` should just have a `set_indices` method which would then hide the fact that we're storing it in element 0 of the list?

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+                                "i: READ, j: READ, x: READ, y: READ"
+
+    # Increment operation: c(i) = c(i)+1
+    increment_access = schedule.children[2]
+    assert isinstance(increment_access, Assignment)
+    var_accesses = VariablesAccessInfo()
+    increment_access.reference_accesses(var_accesses)
+    assert str(var_accesses) == "c: READWRITE, i: READ"
+
+    # Using an intrinsic (looks like an array access): d(i, j) = sqrt(e(i, j))
```
arporter: Not any more! (Sorry.) Please bring this branch up-to-date and fix...

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+                                 end program test_prog''')
+    ast = parser(reader)
+    psy = PSyFactory(API).create(ast)
+    schedule = psy.invokes.get("test_prog").schedule
+
+    indirect_addressing = schedule.children[0]
+    assert isinstance(indirect_addressing, Assignment)
+    var_accesses = VariablesAccessInfo()
+    indirect_addressing.reference_accesses(var_accesses)
+    assert str(var_accesses) == "a: READ, g: WRITE, h: READ, i: READ"
```
arporter: I'm impressed!

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+                                "q: READWRITE, r: READ"
+    # Test that the two accesses to 'q' indeed show up as
+    q_accesses = var_accesses.get_varinfo("q").get_all_accesses()
+    assert len(q_accesses) == 2
+    assert q_accesses[0].get_access_type() == AccessType.READ
+    assert q_accesses[1].get_access_type() == AccessType.WRITE
+    assert q_accesses[0].get_location() < q_accesses[1].get_location()
+
+
+@pytest.mark.xfail(reason="Calls in nemo not yet supported")
```
arporter: Please mark this with an Issue number.

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+    var_accesses = VariablesAccessInfo()
+    call_stmt.reference_accesses(var_accesses)
+    assert str(var_accesses) == "a: UNKNOWN, b: UNKNOWN"
+
+
+def test_do_loop(parser):
+    ''' Check the handling of do loops.
+    TODO: this only tests a nemo do loop. At this stage the loop
+    boundaries in nemo are only strings (not instances of Reference or so),
+    so in case of a loop like: "do jj=1, n", N would
```
arporter: Issue number please and your docstring ends rather abruptly ;-)

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+
+    do_loop = schedule.children[0]
+    assert isinstance(do_loop, nemo.NemoLoop)
+    var_accesses = VariablesAccessInfo()
+    do_loop.reference_accesses(var_accesses)
+    assert str(var_accesses) == "ji: READWRITE, jj: READWRITE, s: WRITE, "\
+                                "t: READ"
+
+
+@pytest.mark.xfail(reason="Nemo converts all loop limits to strings")
```
arporter: Issue number please (might be #400) and strictly it should be "NEMO API converts..."

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+    do_loop = schedule.children[0]
+    assert isinstance(do_loop, nemo.NemoLoop)
+    var_accesses = VariablesAccessInfo()
+    do_loop.reference_accesses(var_accesses)
+    # TODO: n is not reported at the moment
+    assert str(var_accesses) == "ji: READWRITE, jj: READWRITE, n: READ, "\
+                                "s: WRITE, t: READ"
+
+
+@pytest.mark.xfail(reason="Gocean loops boundaries are also strings")
```
arporter: Issue number again please.

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+    do_loop.reference_accesses(var_accesses)
+    print(var_accesses)
+    assert str(var_accesses) == "cu_fld: WRITE, i: READWRITE, j: READWRITE, "\
+                                "p_fld: READ, u_fld: READ"
+    # TODO: atm the return value starts with:  ": READ, cu_fld: WRITE ..."
+    # The empty value is caused by not having start, stop, end of the loop
+    # defined at this stage.
+
+
+@pytest.mark.xfail(reason="Dynamoloops boundaries are also strings")
```
arporter: Issue number please.

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+def test_dynamo():
+    '''Test the handling of a dynamo0.3 loop. Note that the variable accesses
+    are reported based on the user's point of view, not the code actually
+    created by PSyclone, e.g. it shows a dependency on 'some_field', but not
+    on some_field_proxy etc. Also the dependency is at this stage taken
+    from the kernel declaration, not the actual kernel usage.
```
arporter: s/declaration/metadata/

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+                                 "test_files", "dynamo0p3",
+                                 "1_single_invoke.f90"),
+                    api="dynamo0.3")
+    psy = PSyFactory("dynamo0.3", distributed_memory=False).create(info)
+    invoke = psy.invokes.get('invoke_0_testkern_type')
+    schedule = invoke.schedule
+
+    var_accesses = VariablesAccessInfo()
+    schedule.reference_accesses(var_accesses)
+    # TODO: atm there is an empty entry (": READ") at the beginning of the
```
arporter: Issue number please.

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,329 @@
...
+                                 enddo
+                                 a = b
+                                 end program test_prog''')
+    ast = parser(reader)
+    psy = PSyFactory(API).create(ast)
+    schedule = psy.invokes.get("test_prog").schedule
+
+    var_accesses = VariablesAccessInfo()
+    schedule.reference_accesses(var_accesses)
+    print(var_accesses)
```
arporter: Please rm this print().

hiker:
Hi @arporter 

thanks for the feedback. While I am working on this two questions/comments:
1) In several places you ask to specify the scope ("in the current Schedule"? Please quantify the scope.", "Please clarify what "corresponding code section" means - code associated with a given sub-tree in the PSyIR?). The issue is that in general we can't know what the scope is. Certainly it is typically a subtree, but it could have been created by a list of statements (each of which would be a node). I can't provide a real example of this atm, but maybe an algorithm like this could be used:
```
   accesses = VariablesAccessInfo()
   while 1:
       next_statement.reference_accesses(accesses)    # Add the variable accesses of the next statement
       if not can_be_parallelised(accesses):    # Stop when the next statement can not be parallelised anymore
          break
       list_of_parallelisable_statements.append(next_statement)
       next_statement = next_statement.next()
```
   No idea if this code example makes sense in reality - I just want to make a case to leave the comments as vague as they are, since it is up to the user to know and define what information is stored in this class. My suggestion: instead of changing the comment, I'll add more documentation to the developers guide, e.g. adding a normal and 'not so normal' example (like the one above)?

2) What is the correct coding style for docstrings? My understanding is that it is now: 
```
def func(v1, v2):
   '''Comment

   :param v1: ...
   :type v1: ...
   :param v2: ...
   :type v2: ...

   :return: ...
   :rtype: ...

   :raises ...: ...
   :raises ...: ...

   '''
   here_starts = the_code_section
```
TBH, I don't like the last part (newline in comment, not between comment and code): imho the new line between docstring and code is to make it easier to see where the code starts. While this doesn't change much with the ''' in the next line, my 'concern' is that we add a new line to the comment, which does not belong in the comment. Having said that, I'll fix this, but can't help wondering if there is a reason (doxygen??)? I am also happy to start a coding style section in the developers guide (since I have to admit I checked other code, and the usage of closing ''' and newline is ... very varied: some functions have both (new line, ''', new line), none, newline,''', or ''',newline :) Also to be sure: no newline before the comment?

Thanks!
   Joerg

arporter:
Hi @hiker, I'm happy with your suggestion for 1. As for two, I prefer the blank line inside the docstring but I take your point that it's not really a part of the docstring. Depending on what @hiker, @sergisiso, @TeranIvy and @rupertford think, we could change to your suggestion. As for documenting the rules, that's a great idea :-) (I thought I'd already started a wiki page for that but maybe that's in fparser.)

hiker:
I just found a corner case, that is not supported. It will be a bit of a refactoring to support this:
```
   g(g(1)) = g(g(1)) + 1
```
Main issue is that the LHS now has two accesses to the variable, which triggers the InternalErrors (previously asserts) that I have in the code. I can't see any real code using this kind of construct, but I guess we should handle this properly :(

hiker:
After writing the above, I remembered that we agreed that certain code styles would not be supported ... though I can neither find the details now nor do I remember (maybe related to parallelising a loop that would affect more than one dimension?? ``h(j+i,j) = a(i,j) + b(j,i)`` or so? Two j's on the LHS when parallelising across j)? 

So I turn my previous statement into a question: do we allow this kind of statement, or are we ok if PSyclone will raise an exception?

hiker:
I've created #443 to discuss the coding style. Till this is decided I will just follow @arporter's comments, and potentially fix it up when #443 gets done.

arporter:
I think we're OK if PSyclone raises an exception - as you say this is an edge case so as long as the error message is clear then I'm happy :-)

doc/developer_guide/developers.rst
```diff
@@ -530,6 +530,67 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
 `Arguments` class.
 
+Variable Accesses
+=================
+
+Especially in the NEMO API, it is not possible to rely on pre-defined
+kernel information to determine dependencies between loops. So an additional,
+somewhat lower-level API is being implemented that can be used to determine
+variable accesses (READ and WRITE), which is based on the PSyIR information only,
+not on any kernel information. The information about all variable usage
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/developers.rst
```diff
@@ -530,6 +530,67 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+all accesses for that variable using ``psyclone.config.access_info.AccessInfo``
+instances:
+
+.. autoclass:: psyclone.core.access_info.VariableAccessInfo
+    :members:
+
+.. autoclass:: psyclone.core.access_info.AccessInfo
+    :members:
+
+Here a simple example on how to use this API. This is from the
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/developers.rst
```diff
@@ -530,6 +530,67 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+``psyclone.psyGen.OMPParallelDirective`` (so ``self`` is an instance of this
+transformation), and this code is used to determine a list with all scalar
```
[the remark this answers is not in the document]
hiker: Done

doc/developer_guide/developers.rst
```diff
@@ -530,6 +530,67 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+variables that must be declared as thread-private::
+
+  var_accesses = VariablesAccessInfo()
+  self.reference_accesses(var_accesses)
+  for var_name in var_accesses.get_all_vars():
+      accesses = var_accesses.get_varinfo(var_name).get_all_accesses()
+      # Ignore variables that have indices, we only look at scalar
+      # This tests only the first access, the assumption is that all
+      # accesses are either using an array access, or none.
+      if accesses[0].get_indices() is not None:
```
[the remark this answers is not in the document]
hiker: While I totally agree,I am not sure if the SymbolTable is ready to be used ... and how are we going to handle the case if we should not have any declarations? Not sure if the latter case could actually happen in reality (it will affect all test cases, since I never declared anything there ;) ). I'll bring this up in our telco tonight.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author J. Henrichs, Bureau of Meteorology
+# -----------------------------------------------------------------------------
+
+'''This module management of variable access information.'''
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+    pattern of one variable (e.g. variable is read at a certain location).
+    A location is a number which can be used to compare different access
+    (i.e. if one access is happening before another). Each consecutive
+    location will have an increasing location number, but read and write
+    accesses in the same statement will have the same location number.
+    If the variable accessed is an array, this class will also store
+    the indices used in the access.
+    Note that the name of the variable is not stored in this instance, it
+    is typically available using VariableAccessInfo class (which stores all
+    AccessInfo objects for a variable).
```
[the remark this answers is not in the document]
hiker: As discussed, I've added more clarification in the documentation, since at this stage the object is not restricted to a schedule, or even a single PSyIR node.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+    :param access: The access type.
+    :type access_type: :py:class:`psyclone.core.access_type.AccessType`
+    :param int location: A number used in ordering the accesses.
+    :param indices: Indices used in the access, defaults to None
+    :type indices: list of :py:class:`psyclone.psyGen.Node` instances \
+        (e.g. Reference, ...)
+    :param node: Node in PSyIR in which the access happens, defaults to None.
+    :type node: :py:class:`psyclone.psyGen.Node` instance.
+    '''
+
```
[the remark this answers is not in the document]
hiker: I've done all of them (I won't confirm them all), but we also have #443 to discuss this.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+            self._indices = indices[:]
+        else:
+            self._indices = None
+
+    def change_read_to_write(self):
+        '''This changes the access mode from READ to WRITE.
+        This is used for processing assignment statements,
+        where the LHS is first considered to be READ,
+        and which is then changed to be WRITE.'''
+        assert self._access_type == AccessType.READ
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def set_indices(self, indices):
+        '''Sets the indices for this AccessInfo instance.
+
+        :param indices: List of indices used in the access.
+        :type indices: list of :py:class:`psyclone.psyGen.Node` instances.
+        '''
+        self._indices = indices[:]
+
+    def get_indices(self):
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+        '''
+        return self._indices
+
+    def get_access_type(self):
+        ''':returns: the access type.
+        :rtype: :py:class:`psyclone.core.access_type.AccessType`'''
+        return self._access_type
+
+    def get_location(self):
+        ''':returns: the location information for this access.
```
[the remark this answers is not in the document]
hiker: As above, updated documentation.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+
+    :param str var_name: Name of the variable.
+    '''
+
+    def __init__(self, var_name):
+        self._var_name = var_name
+        # This is the list of AccessInfo instances for this variable.
+        self._accesses = []
+
+    def get_var_name(self):
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def change_read_to_write(self):
+        '''This function is only used when analysing an assignment statement.
+        The LHS has first all variables identified, which will be READ.
+        This function is then called to change the assigned-to variable
+        on the LHS to from READ to WRITE. Since the LHS is stored in a separate
+        VariableAccessInfo class, it is guaranteed that there is only
+        one entry for the variable.
+         '''
+        assert len(self._accesses) == 1
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+    '''This class stores all VariableAccessInfo instances for all variables
+    in the corresponding code section. It maintains a 'location' information,
+    which is an integer number that is increased for each new statement. It
+    can be used to easily determine if one access is before another.
+
+    :param int location: The index for the first statement, default is 0.
+    '''
+
+    def __init__(self, location=0):
+        self._var_to_varinfo = {}
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def next_location(self):
+        '''Increases the location number.
+        '''
+        self._location = self._location + 1
+
+    def add_access(self, var_name, access_type, node, indices=None):
+        '''Adds access information for the specified variable.
+
+        :param access_type: The type of access (READ, WRITE, ...)
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/dynamo0p3.py
```diff
@@ -6156,6 +6157,25 @@ def __init__(self):
...
         self._is_intergrid = False  # Whether this is an inter-grid kernel
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. All accesses are marked
+        according to the kernel declaration.
+        :param var_accesses: \
+            :py:class:`psyclone.core.access_info.VariablesAccessInfo`
+        '''
+        for arg in self.arguments.args:
+            # If arg is a grid type argument, it does not have
```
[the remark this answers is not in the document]
hiker: Good catch, that comment is outdated (in a previous version I had to access the arg data or metadata information ... and that comment was copied from somewhere :) ). But with my fix for the vector_size this comment is now outdated.

src/psyclone/dynamo0p3.py
hiker: Done

src/psyclone/psyGen.py
hiker: Good idea, done. I also added a __getitem__ for the list of access information.

src/psyclone/psyGen.py
hiker: Yes, it works, and yes, same location, and I have a test case for that (and actually a minor fix to make sure that the read is coming before the write).

src/psyclone/psyGen.py
hiker: Agreed - that line was actually wrong, the 0 must be a -1 to modify the last element that was just added before. I've fixed it and added a comment to describe this.

src/psyclone/tests/dependency_test.py
hiker: Luckily it works out of the box for now, it just sets e as read. The one issue that we might need to solve if we have intrinsics that modify a parameters.

src/psyclone/tests/dependency_test.py
hiker: Done.

doc/developer_guide/developers.rst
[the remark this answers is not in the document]
hiker: I had a go at this issue, and I am not sure if I might have misunderstood :)  I tried to add a symbol table to the VariablesAccessesInfo class, and then add a function 'is_array()' - but it somehow feels wrong. Imho this class should really only store the information, not the interpretation - the script using the access information should decide what to use (e.g. symbol table or if there is an index in the access). 

So I re-read the ticket, I am wondering if you actually only meant: in this example in the docs we should use the symbol table (if available)? Here how I have rewritten the example code in the docs:
```
  for var_name in var_accesses.all_vars:
      accesses = var_accesses[var_name].all_accesses
      # Ignore variables that are arrays, we only look at scalar ones.
      # If we do have a symbol table, use the shape of the variable
      # to determine if the variable is scalar or not
      if symbol_table:
          if len(symbol_table.lookup(var_name).shape) > 0:
              continue
      else
          # There is no symbol table, check instead if the first access of
          # the variable has indices, and assume it is an array if it has:
          if accesses[0].indices is not None:
              continue

```

Is just the change of the example in the documentation what you wanted me to change? Or do you really want to have support for an optional symbol table inside the variable access information?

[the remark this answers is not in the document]
arporter: I think I meant in the specific example and I'm happy with your suggested change. I don't think we should be adding a SymbolTable to the `VariablesAccessInfo` class. I think probably what is needed is that each Symbol in the PSyIR contain a ptr to its entry in the SymbolTable (if any) but that's not something for this Issue.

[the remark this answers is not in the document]
hiker: Puh - glad to hear that (and glad that I stopped adding the SymbolTable :grin:). I've added this to the example.

hiker:
@arporter Thanks for the answer. I believe I have addressed all comments now. Note that my last commit had a [skip ci], since it only updated the documentation.

src/psyclone/core/access_info.py
```diff
@@ -0,0 +1,364 @@
...
+        This is used for processing assignment statements,
+        where the LHS is first considered to be READ,
+        and which is then changed to be WRITE.
+
+        :raises InternalError: if the variable originally does not have\
+            READ access.
+        '''
+        if self._access_type != AccessType.READ:
+            from psyclone.psyGen import InternalError
+            raise InternalError("Trying to change variable to 'WRITE' "
```
arporter: Please could you include the variable name in the error message.

src/psyclone/core/access_info.py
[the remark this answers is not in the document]
arporter: Great. Please could you add "Please see the Developers' Guide for more information." or some such.

src/psyclone/dynamo0p3.py
```diff
@@ -6156,6 +6157,28 @@ def __init__(self):
...
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. All accesses are marked
+        according to the kernel metadata
+
+        :param var_accesses: VariablesAccessInfo instance that stores the\
+            information about variable accesses.
+        :type var_accesses: \
+            :py:class:`psyclone.core.access_info.VariablesAccessInfo`
+        '''
+        for arg in self.arguments.args:
```
arporter: I started to wonder whether we couldn't use the same `reference_accesses` routine for both gocean1.0 and dynamo0.3 and then I realised that you're using the `vector_size` property to decide whether a kernel argument is an array. I suspect this isn't actually what you want to do since the naming is a bit confusing. In Kernel meta-data a user may specify that a kernel takes a 'vector' of fields by doing, e.g. "GH_FIELD*3". This translates into the kernel having 3 field arguments, each of which is then an array. I think you should really check on the `type` property of the argument - if it is anything other than a scalar (`gh_real` or `gh_integer` in dynamo0p3) then you have an array.

src/psyclone/gocean1p0.py
```diff
@@ -883,6 +883,20 @@ def __init__(self):
...
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. All accesses are marked
+        according to the kernel metadata.
+
+        :param var_accesses: VariablesAccessInfo instance that stores the\
+            information about variable accesses.
+        :type var_accesses: \
+            :py:class:`psyclone.core.access_info.VariablesAccessInfo`
+        '''
+        for arg in self.arguments.args:
```
arporter: Again, you need to check whether the args are scalars or not (I think). It may be possible to do this using the same code as for dynamo0p3.

src/psyclone/psyGen.py
```diff
@@ -4797,6 +4878,28 @@ def __str__(self):
             result += str(entity)
         return result
 
+    def reference_accesses(self, var_accesses):
+        '''Get all variable access information. It combines the data from
+        the condition, if-body and (if available) else-body. This could
+        later be extended to handle cases where a variable is only written
+        in one the two branches.
```
arporter: Almost! "in one of the two branches."

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,358 @@
...
+    call_stmt.reference_accesses(var_accesses)
+    assert str(var_accesses) == "a: UNKNOWN, b: UNKNOWN"
+
+
+def test_do_loop(parser):
+    ''' Check the handling of do loops.
+    TODO #400: this only tests a nemo do loop. At this stage the loop
+    boundaries in nemo are only strings (not instances of Reference or so),
+    so in case of a loop like: "do jj=1, n", `n` is not be listed as
+    a READ access.
```
arporter: Thanks for clarifying. I think this test should therefore be altered to check that `n` *is* listed as a READ access and then marked as x-fail. Ah, but then I see it would just be a duplicate of the test below. Is there a reason for having the two separate tests?

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,358 @@
...
+
+    do_loop = schedule.children[0]
+    assert isinstance(do_loop, nemo.NemoLoop)
+    var_accesses = VariablesAccessInfo()
+    do_loop.reference_accesses(var_accesses)
+    assert str(var_accesses) == "ji: READWRITE, jj: READWRITE, s: WRITE, "\
+                                "t: READ"
+
+
+@pytest.mark.xfail(reason="NEMO API converts all loop limits to strings")
```
arporter: Please could you add the issue number reference to the `reason` string.

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,358 @@
...
+    # TODO #400: n is not reported at the moment
+    assert str(var_accesses) == "ji: READWRITE, jj: READWRITE, n: READ, "\
+                                "s: WRITE, t: READ"
+
+
+@pytest.mark.xfail(reason="Gocean loops boundaries are also strings #400/#444")
+def test_goloop():
+    ''' Check the handling of non-NEMO do loops.
+    TODO #400/#444: Does not work atm, GOLoops also have start/stop as
+    strings, which are even not defined. Only after genCode will they be
```
arporter: "after gen_code() is called will they..."

src/psyclone/core/access_info.py
hiker: *grin* I left it out on purpose, since the AccessInfo object does not have the variable name (it seems a bit of a waste to store the variable name for each access, while de-facto a user will always go via the VariableAccessInfo class, which has the variable name.

So two options:
1) I catch this error in the `VariableAccessInfo` class (since all calls to the `AccessInfo.change_read_to_write` will come from `VariableAccessInfo.change_read_to_write` anyway), and then raise the InternalError with the VariableName.
2) I add the variable name to `AccessInfo`.

My preference would actually be 1) tbh, it seems a waste to add the variable name everywhere ... just more stuff to keep in sync.

doc/developer_guide/developers.rst
```diff
@@ -568,6 +568,155 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+      if len(accesses) == 1:
+          continue
+
+      # We have at least two accesses. If the first one is a write,
+      # assume the variable should be private:
+      if accesses[0].access_type == AccessType.WRITE:
+          result.add(var_name.lower())
+
+
+The next, hypothetical example shows how the VariablesAccessInfo class can
```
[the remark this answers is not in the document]
hiker: Nope :) I actually did look while writing this section, but only noticed sections which did not highlight class names (etc) at all. In most cases it is actually in italics (single backtick), so I am changing this, and try to update the other non-italics uses I have noticed. I am also going to add a section about writing documentation to the developers guide (will be part of the coding style documentation update).

doc/developer_guide/developers.rst
```diff
@@ -568,6 +568,155 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+
+  var_accesses = VariablesAccessInfo()
+  self.reference_accesses(var_accesses)
+  for var_name in var_accesses.all_vars:
+      accesses = var_accesses[var_name].all_accesses
+      # Ignore variables that are arrays, we only look at scalar ones.
+      # If we do have a symbol table, use it the shape of the variable
+      # to determine if the variable is scalar or not
+      if symbol_table:
+          if len(symbol_table.lookup(var_name).shape) == 0:
```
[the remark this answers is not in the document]
hiker: Oops - good catch!

doc/developer_guide/developers.rst
```diff
@@ -568,6 +568,155 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+``psyclone.psyGen.OMPParallelDirective`` (so ``self`` is an instance of this
+node), and this code is used to determine a list of all the scalar
+variables that must be declared as thread-private::
+
+  var_accesses = VariablesAccessInfo()
+  self.reference_accesses(var_accesses)
+  for var_name in var_accesses.all_vars:
+      accesses = var_accesses[var_name].all_accesses
+      # Ignore variables that are arrays, we only look at scalar ones.
+      # If we do have a symbol table, use it the shape of the variable
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyGen.py
hiker: Gee ... (frantically looking for an excuse) ... my keyboard is broken?  😊

src/psyclone/tests/dependency_test.py
hiker: Well, I think it's a good idea to have a passing test - it might check more of the code (e.g. taking a different branch in the code in case of the x-fail problem). I've marked the passing test with #400 and that it can be removed once the other (currently x-failing) test works as expected.

src/psyclone/tests/dependency_test.py
hiker: Done.

src/psyclone/tests/dependency_test.py
hiker: Done.

src/psyclone/core/access_info.py
arporter: :-) I noticed it was there in other error messages. I'm happy with option 1 as well.

doc/developer_guide/developers.rst
```diff
@@ -568,6 +568,155 @@ DataAccess class i.e. the `_field_write_arguments()` and
...
+      # We have at least two accesses. If the first one is a write,
+      # assume the variable should be private:
+      if accesses[0].access_type == AccessType.WRITE:
+          result.add(var_name.lower())
+
+
+The next, hypothetical example shows how the VariablesAccessInfo class can
+be used iteratively. Assume that you have a function that determines
+if the given variable accesses can be parallelised, and the aim is to
+determine the largest consecutive block of statement that can be
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/core/access_info.py
hiker: Done.

src/psyclone/dynamo0p3.py
hiker: Let me retell the odyssey here ... sit back, grab some popcorn 🍿 and relax ...

I liked the idea of unifying this ... so I added a special `is_scalar()` function (since scalars have a different `type` depending on API), which could then be used API independent. 

Then I discovered that in GOcean there are two different classes for arguments (GOKernelArgument and GOKernelGridArgument), and I didn't have `is_scalar()` for the latter. So I extended the data structure GRID_PROPERTY_DICT to included scalar/array information and added the third `is_scalar()`. While doing this and changing my test case to include a grid property, it turns out that (at least) the opencl code creation is incorrect if e.g. go_grid_x_min_index is used (the grid_property_dict maps this to "subdomain%internal%xstart" ... I assume Andy can see where this goes wrong then 😢 ). So I fixed that, too, and updated the test cases.
Next, I then looked at the variable access information, which was also not correct (using e.g. `xstart` as read variable in the example above, which makes no sense at all), so  I handled grid properties separately to instead mark `<some_field>%data%subdomain%internal%xstart` as READ (I'll leave the user defined type handling for latter - you can obviously argue that we should mark field, field%data, ... all as READ??).

So as a result of all this refactoring, the two implementations are still different and not in the base class, since I can't see an easy way of achieving this due to the special handling for grid properties (unless we do a lot more refactoring of the Argument classes, which imho should be a separate pull request).

Am happy to discuss this 😅

src/psyclone/gocean1p0.py
hiker: All fixed, but see above.

src/psyclone/tests/dependency_test.py
hiker: I actually added more 'duplicated' tests - main problem is that if a test is marked as xfail, we can't distinguish the reason for the fail anymore. And with only one xfailing test for (say) gocean, any typo etc in gocean code that should be detect during testing would go unnoticed, since the tests would still xfail (and yes, that exactly happened to me ... I had a typo in my new is_scalar function, but no indication of any unexpected test failures 😢  ) . So I added a duplicated test for gocean and dynamo that is not xfailing, with enough comments to indicate that this should be removed once the xfailing tests work correctly.

hiker:
OK, I think it is ready for the next review. I see two lines not covered in the coverty tests, but they are (as far as I can tell) existing failures, and I just happened to modify code close to it (e.g. to avoid a warning) ... oh, it actually says that the diff is 100% covered 😅

doc/developer_guide/developers.rst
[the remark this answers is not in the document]
arporter: Thanks!

src/psyclone/gocean1p0.py
```diff
@@ -80,25 +80,29 @@
...
-                      "go_grid_dy_v": "dy_v",
-                      "go_grid_lat_u": "gphiu",
-                      "go_grid_lat_v": "gphiv",
-                      "go_grid_dx_const": "dx",
-                      "go_grid_dy_const": "dy",
-                      "go_grid_x_min_index": "subdomain%internal%xstart",
-                      "go_grid_x_max_index": "subdomain%internal%xstop",
-                      "go_grid_y_min_index": "subdomain%internal%ystart",
-                      "go_grid_y_max_index": "subdomain%internal%ystop"}
+# properties of the grid to their names and type in the Fortran grid_type.
```
arporter: Perhaps add:

    # i.e. they can be accessed in Fortran by doing grid%<name>, e.g. grid%area_t or grid%subdomain%internal%xstart.

src/psyclone/gocean1p0.py
```diff
@@ -892,8 +896,25 @@ def reference_accesses(self, var_accesses):
         :type var_accesses: \
             :py:class:`psyclone.core.access_info.VariablesAccessInfo`
         '''
+
+        # Grid properties are accessed using one of the fields. This stores
+        # the field used to avoid repeatedly determining the best field:
+        grid_for_property = None
```
arporter: Strictly speaking, this should be `field_for_grid_property` or somesuch.

src/psyclone/gocean1p0.py
```diff
@@ -892,8 +896,25 @@ def reference_accesses(self, var_accesses):
...
         for arg in self.arguments.args:
-            var_accesses.add_access(arg.name, arg.access, self)
+            if arg.type == "grid_property":
+                if grid_for_property is None:
```
arporter: This should be `if not grid_for_property:`

src/psyclone/gocean1p0.py
```diff
@@ -1431,13 +1461,26 @@ def name(self):
...
+
     @property
     def type(self):
         ''' The type of this argument. We have this for compatibility with
             GOKernelArgument objects since, for this class, it will always be
             "grid_property". '''
         return self._type
 
+    def is_scalar(self):
+        ''':return: If this variable is a scalar variable or not.
```
arporter: s/If/whether/

src/psyclone/gocean1p0.py
```diff
@@ -1788,8 +1831,14 @@ def grid_prop(self):
     @property
     def type(self):
         ''' The type of this argument - whether it is a scalar, a field or
-            a grid-property. The latter are special because they must be
-            supplied by the PSy layer. '''
+        a grid-property. The latter are special because they must be
+        supplied by the PSy layer.
+
+        :return; The type of this argument, either 'field', 'scalar', or \
```
arporter: Thanks for updating this docstring. You have a typo in `:return:` though - second colon is currently a semi-colon.

src/psyclone/psyGen.py
```diff
@@ -4630,6 +4629,11 @@ def space(self):
     def stencil(self):
         return self._arg.stencil
 
+    @abc.abstractmethod
+    def is_scalar(self):
+        ''':return: If this variable is a scalar variable or not.
```
arporter: s/If/whether/

src/psyclone/tests/dependency_test.py
```diff
@@ -159,7 +159,7 @@ def test_if_statement(parser):
     var_accesses = VariablesAccessInfo()
     if_stmt.reference_accesses(var_accesses)
     assert str(var_accesses) == "a: READ, b: READ, i: READ, p: WRITE, "\
-                                "q: READWRITE, r: READ"
+                                "q: READ+WRITE, r: READ"
```
arporter: What's the difference between `READWRITE` and `READ+WRITE` and please could you document this in the Developers' Guide?

src/psyclone/tests/dependency_test.py
arporter: All sounds good to me - thanks for explaining.

src/psyclone/tests/dependency_test.py
```diff
@@ -0,0 +1,407 @@
...
+                                "j: READWRITE, " "ssh_fld: READWRITE, "\
+                                "tmask: READ"
+    # TODO #444: atm the return value starts with:  ": READ, cu_fld: WRITE ..."
+    # The empty value is caused by not having start, stop, end of the loop
+    # defined at this stage.
+
+
+def test_goloop_partially():
+    ''' Check the handling of non-NEMO do loops.
+    TODO #400: This test is identical to test_goloop above, but is asserts in a
```
arporter: s/is/it/

src/psyclone/tests/gocean1p0_opencl_test.py
```diff
@@ -190,13 +191,14 @@ def test_set_kern_float_arg(outputdir):
     otrans.apply(sched)
     generated_code = str(psy.gen)
     expected = '''\
-    SUBROUTINE bc_ssh_code_set_args(kernel_obj, nx, a_scalar, ssh_fld, tmask)
+    SUBROUTINE bc_ssh_code_set_args(kernel_obj, nx, a_scalar, ssh_fld, '''+\
```
arporter: This routine no longer matches the OpenCL implementation of the bc_ssh_code kernel (https://github.com/stfc/PSycloneBench/blob/master/ocean/nemo/nemolite2d/kernels/opencl/boundary_conditions_kern.c).

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -276,12 +276,11 @@ def test_scalar_float_arg(tmpdir):
         "      DO j=1,jstop+1\n"
         "        DO i=1,istop+1\n"
         "          CALL bc_ssh_code(i, j, a_scalar, ssh_fld%data, "
-        "ssh_fld%grid%tmask)\n"
+        "ssh_fld%grid%subdomain%internal%xstop, ssh_fld%grid%tmask)\n"
```
arporter: As previously commented, I don't understand why we're now passing `xstop` to this kernel?

src/psyclone/tests/test_files/gocean1p0/kernel_scalar_float.f90
```diff
@@ -14,9 +14,10 @@ module kernel_scalar_float
   !=======================================
 
   type, extends(kernel_type) :: bc_ssh
-     type(go_arg), dimension(3) :: meta_args =              &
+     type(go_arg), dimension(4) :: meta_args =                 &
           (/ go_arg(GO_READ,      GO_R_SCALAR, GO_POINTWISE),  &
              go_arg(GO_READWRITE, GO_CT,       GO_POINTWISE),  &
+             go_arg(GO_READ,      GO_GRID_X_MAX_INDEX),        &
```
arporter: Ah! So you've extended the test to exercise something that was previously broken?

hiker:
OK, so there appear two opencl releated problem? The first one, that you did not see: originally PSyclone would create code like:
```
SUBROUTINE bc_ssh_code_set_args(kernel_obj, ..., subdomain%internal%xstop, tmask)
...
integer(KIND=c_intptr_t), target :: ssh_fld, subdomain%internal%xstop, tmask
```
My patch in this pull request replaces `subdomain%internal%xstop` with just xstop. I also need that fix for more useful dependency handling (I somehow originally got just 'xstop', which does not exist as a stand-alone variable). Now I get at least the "field%grid..." expression - proper support for user defined types can be done later imho.

But it appears that opencl does not handle any scalar grid attributes at all, at least when I used go_grid_dx_const (which has the same issue as using my xstop etc). Since this is unrelated to my branch, I've opened #459 for this, commented out the call to compilation, and added a todo with the issue number. Seems to be the best solution for now.

src/psyclone/core/access_info.py
```diff
@@ -162,7 +164,19 @@ def is_read(self):
         :rtype: bool
         '''
         for access_info in self._accesses:
-            if access_info.access_type == AccessType.READ:
+            if access_info.access_type in AccessType.all_read_accesses():
+                return True
+        return False
+
+    def has_read_write(self):
+        '''Checks if the specified variable name has at least on READWRITE
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/core/access_info.py
```diff
@@ -162,7 +164,19 @@ def is_read(self):
...
+        access.
+
+        :returns: true if the specified variable name is read (at least once).
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/core/access_info.py
```diff
@@ -362,3 +393,16 @@ def is_read(self, var_name):
 
         var_access_info = self[var_name]
         return var_access_info.is_read()
+
+    def has_read_write(self, var_name):
+        '''Checks if the specified variable name has at least one READWRITE
+        access (which is typically only used in a function call)
+
+        :param str var_name: Name of the variable
+        :returns: true if the specified variable name is read (at least \
```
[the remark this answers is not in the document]
hiker: I prefer the 'has at least one READWRITE access' - sounds better than 'is READWRITE at least once'?

src/psyclone/core/access_info.py
```diff
@@ -362,3 +393,16 @@ def is_read(self, var_name):
...
+            once).
+        :rtype: bool
+        :raises: KeyError if the variable names can not be found.'''
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/gocean1p0.py
```diff
@@ -892,8 +896,25 @@ def reference_accesses(self, var_accesses):
...
+
+        # Grid properties are accessed using one of the fields. This stores
+        # the field used to avoid repeatedly determining the best field:
+        grid_for_property = None
         for arg in self.arguments.args:
-            var_accesses.add_access(arg.name, arg.access, self)
+            if arg.type == "grid_property":
+                if grid_for_property is None:
+                    grid_for_property = self._arguments.find_grid_access()
+                var_name = grid_for_property.name + "%data%" + \
```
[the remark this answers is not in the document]
hiker: Oops - indeed.

src/psyclone/gocean1p0.py
```diff
@@ -1397,6 +1417,11 @@ def function_space(self):
             argument as specified by the kernel argument metadata.'''
         return self._arg.function_space
 
+    def is_scalar(self):
+        ''':return: If this variable is a scalar variable or not.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/gocean1p0.py
```diff
@@ -1431,13 +1461,26 @@ def name(self):
             used in the generated code like so: <fld>%grid%name '''
         return self._name
 
+    @property
+    def dereference_name(self):
+        ''' Returns dereference string required to access a property in
+        a dl_esm field (e.g. "subdomain%internal%xstart"). This name is
+        used in the generated code like so: <fld>%grid%name '''
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/dynamo0p3.py
```diff
@@ -8384,6 +8384,11 @@ def type(self):
         ''' Returns the type of this argument. '''
         return self._type
 
+    def is_scalar(self):
+        ''':return: If this variable is a scalar variable or not.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/gocean1p0.py
hiker: Done, and the bug that I also fixed is in my comment at the end of this thread.

src/psyclone/gocean1p0.py
hiker: Indeed ... though it made it necessary to split a line ... 81 characters with the better name

src/psyclone/gocean1p0.py
hiker: I thought I had pylint telling me in the past that I needed 'is' for None comparisons. Afaik there is a slight difference (since 'if not' would also trigger when the value is 0). Doesn't matter here, and indeed pylint does not complain about if not ... so done.

src/psyclone/gocean1p0.py
hiker: Done.

src/psyclone/gocean1p0.py
hiker: Done.

src/psyclone/psyGen.py

src/psyclone/tests/dependency_test.py
hiker: I am adding this to the documentation of the __str__ method (which is included in the developers guide), since there is no separate type 'READ+WRITE'. I am not even sure if this is really required - it allows to distinguish between:
```
a=a+1     ! one read and one write access with same location
b=a; a=7;    ! one read and one write access with different location
call sub(a)    ! with metadata for a indicating READWRITE
```
And I thought it would be useful to see the difference between the first two and the last access in __str__().

src/psyclone/tests/dependency_test.py
hiker: Done

src/psyclone/tests/gocean1p0_test.py
hiker: As discussed, I added this as an additional (originally failing) test case.

src/psyclone/tests/test_files/gocean1p0/kernel_scalar_float.f90
hiker: Yep :+1:

hiker:
Ready for next review.

doc/developer_guide/developers.rst
```diff
@@ -574,8 +574,10 @@ Variable Accesses
 Especially in the NEMO API, it is not possible to rely on pre-defined
 kernel information to determine dependencies between loops. So an additional,
 somewhat lower-level API has been implemented that can be used to determine
-variable accesses (READ and WRITE), which is based on the PSyIR information only,
-not on any kernel metadata information. The information about all variable usage
+variable accesses (READ, WRITE etc.), which is based on the PSyIR information.
+Only exception is if a kernel is called, then the metadata for the kernel
```
arporter: "The only exception to this is that if a kernel..."

src/psyclone/tests/dependency_test.py
arporter: That's great - thanks.

src/psyclone/core/access_info.py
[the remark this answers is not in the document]
arporter: Yep.

src/psyclone/dynamo0p3.py
arporter: Thanks for explaining (in a comment elsewhere).

arporter:
Once Travis has reported back, I'll merge this to master.

PR #426, opened 2019-07-05
Simplify walk

### LINKED ISSUE #410: Why is 'walk' implemented the way it is?
When trying to work around #409, my first thought was to add a walk function for NemoKern that looks in self._kern_schedule instead of self._children, but then I noticed that walk has a very strange implementation:
```
    def walk(self, children, my_type):
        for child in children:
...
            local_list += self.walk(child.children, my_type)
        return local_list
```
Notice that 'self' is only used to determine the walk function used (which as far as I can see is always Node.walk), the parameter children can be something else. That leads to odd calls:
```
nodes = self.root.walk(self.root.children, ...)
```
Why is it not implemented by taking 'children' from self:
```
 def walk(self, my_type):
        for child in self.children:
...
```
then it could be called as
```
nodes = self.root.walk(...)
```
I know there are calls in the form:
```
all_nodes = self.walk(self.root.children, Node)
```
but considering that there appears to be only one walk implementation, I can't see a problem this being changed to:
```
all_nodes = self.root.walk(self.root.children, Node)
```
which would then also allow this to be called as:
```
all_nodes = self.root.walk(Node)
```
with this implementation.

Main reason why I ask: with this implementation I could then overwrite NemoKern.walk to do what I want it to do :to work around #409)

I am happy to discuss this next time we meet, it's not urgent for me.

hiker:
This fixes #410. It's mostly a straight-forward simplification, e.g.:
```
-    if node.walk([node], excluded_node_types):
+    if node.walk(excluded_node_types):
```
Only exception is the function NemoKern.match:
```
        nodes = node.walk((CodeBlock, NemoLoop))
        # 'node' can be returned as part of the walk, remove it:
        if node in nodes:
            nodes.remove(node)
        if nodes:
            # A kernel cannot contain unrecognised code (including IO
            # operations and routine calls) or loops.
            return False
```
The fact that node.walk() can include node itself causes the tests to fail, so this is the one function where I had to handle this explicitly.

And FYI: I also tried some performance optimisation by supplying the result list as parameter to walk (to avoid the frequent creation of local lists as return value, and concatenation of lists) - but while the run time difference was neglectable, the 'old' implementation that concatenates list was always just a bit faster.


hiker:
I just realised that I must have branched that off the variable-access branch (pull request #424). Since I am too lazy to rebase (and this is really not urgent), I'll wait with this till $424 is merged.

And it appears that the nemo test can be changed to:
```
         nodes = node.walk((CodeBlock, NemoLoop))
         # 'node' is always contained in the result of walk. So if there is one
         # more CodeBlock or NemoLoop, return False.
        if len(nodes)>1:
             # A kernel cannot contain unrecognised code (including IO
             # operations and routine calls) or loops.
             return False
```
I admit I am not sure if 'nodes' is indeed always contained in the result of walk tbh - the tests all seems to work, but it should be confirmed - the current code (removing the node if it is) is 'safer'. Let me know which variant you prefer.

hiker:
I am going to rebase this on master and open a new pull request, this was off a very strange branch 😱

Issue #438, opened 2019-07-15
Fix pytest ExceptionInfo strings


sergisiso:
Pytest 5.0 changed the ExceptionInfo objects returned by pytest.raises (https://docs.pytest.org/en/latest/changelog.html).

This breaks many of PSyclone tests that contain a `str(error)` expression after a pytest.raises from a clean pip installation or if you do a pip update.

A quick fix is limiting pytest to version < 5.0, but the proper fix requires updating the tests.

arporter:
I've been naughty and already done this fix as part of work on another Issue. (I hadn't realised it was pytest that had changed - I just went ahead and fixed the failing tests.) However, it would be nicer if we did it as a separate PR so I propose we keep this Issue open.

arporter:
Just to correct what I said earlier - it turns out that the fix I thought I'd done was actually in fparser, not in PSyclone.

rupertford:
Created branch 438_fix_pytest_strings

arporter:
#573 is merged.

PR #439, opened 2019-07-15
Issue #438. Pytest quick fix

sergisiso:
Limits pytest version in order to work with new pip installations. See #438.

(also, I am wondering if this is related with the codecov issues we are having)


sergisiso:
@rupertford @arporter At the moment master is broken if used with a clean pip install or update, this should fix it. This issue was not raised by travis because they use a slightly older version of pip 19.0, but it happens with the virtualenv pip 19.1. It should be ready for review when all test pass.

sergisiso:
I manually submitted a codecov report using the codecov cli tool, if this is in master I belive it will report coverage tests again :)

image

sergisiso:
@arporter I added the Warning in the Getting Going, there are other references to pytest which I haven't modified but I think it's better to keep this simple and leave the issue #438 opened until we fix it properly, then we can remove the version limit and the Warning.

arporter:
Thanks for tweaking the documentation and I agree - there's no point going to town on it (i.e. making sure we cover everywhere py.test is mentioned.

arporter:
Oh, and well done for fixing CodeCov! You'll have to explain to me what you did (tomorrow perhaps).

Issue #433, opened 2019-07-09
psyir fortran backend does not support nemo implicit loops


rupertford:
The nemo api extends the PSyIR to include a NemoImplicitLoop. The PSyIR Fortran back-end should be able to output such loops.

Issue #432, opened 2019-07-09
the psyir fortran back end does not support the nemo_invoke node


rupertford:

arporter:
This should have been closed when #476 was merged.

Issue #429, opened 2019-07-09
Write all lines for multi-line code blocks


rupertford:
At the moment only the first line of a multi-line code block is output from the psyir fortran back end.

rupertford:
Created branch 429_code_blocks

LonelyCat124:
I think this has been solved by #476 so I will close.

Issue #473, opened 2019-08-21
Graphviz Test Failures on master


hiker:
I am getting graphviz/dag related test failures on master:
```
# linux-atdw:~/work/psyclone >
pytest -s -k test_writetoread_dag ./src/psyclone/tests/gocean1p0_test.py
=========================================================================================================== test session starts ============================================================================================================
platform linux -- Python 3.6.5, pytest-4.5.0, py-1.5.2, pluggy-0.11.0
rootdir: /home/joergh/work/psyclone, inifile: setup.cfg
plugins: xdist-1.28.0, forked-1.0.2
collected 49 items / 48 deselected / 1 selected                                                                                                                                                                                            

src/psyclone/tests/gocean1p0_test.py F

================================================================================================================= FAILURES =================================================================================================================
___________________________________________________________________________________________________________ test_writetoread_dag ___________________________________________________________________________________________________________

tmpdir = local('/tmp/pytest-of-joergh/pytest-178/test_writetoread_dag0'), have_graphviz = True

    def test_writetoread_dag(tmpdir, have_graphviz):
        ''' Test that the GOInvokeSchedule::dag() method works as expected when we
        have two kernels with a write -> read dependency '''
        _, invoke_info = parse(os.path.join(os.path.
                                            dirname(os.path.
                                                    abspath(__file__)),
                                            "test_files", "gocean1p0",
                                            "single_invoke_write_to_read.f90"),
                               api=API)
        psy = PSyFactory(API).create(invoke_info)
        invoke = psy.invokes.invoke_list[0]
        old_cwd = tmpdir.chdir()
        invoke.schedule.dag()
        if have_graphviz:
            dot_file = os.path.join(str(tmpdir), "dag")
            assert os.path.isfile(dot_file)
            with open(dot_file, "r") as dfile:
                dot = dfile.read()
            assert dot.startswith("digraph")
            # write -> read means that the second loop can only begin once the
            # first loop is complete. Check that we have the correct forwards
            # dependence (green) and backwards dependence (red).
>           assert ('"loop_[outer]_1_end" -> "loop_[outer]_4_start" [color=red]'
                    in dot or
                    '"loop_[outer]_1_end" -> "loop_[outer]_4_start" '
                    '[color=#ff0000]' in dot)
[...]

PR #476, opened 2019-08-22
issue #429, #432, #433. Improve Fortran PSyIR back end for NEMO

rupertford:
Fixes multi-line code block error. #429
Adds support for nemo implicit loops. #433
Adds support for the nemo invoke node. #432


rupertford:
Ready for review from either @arporter or @sergisiso.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -496,11 +479,74 @@ def codeblock_node(self, node):
...
         :type node: :py:class:`psyclone.psyGen.CodeBlock`
 
         :returns: The Fortran code as a string.
         :rtype: str
 
         '''
-        return self._nindent.join(str(node.ast).splitlines(True))
+        result = ""
+        for statement in node.get_statements:
+            result += "{0}{1}\n".format(self._nindent, str(statement))
```
hiker: I don't like the concatenations like this anywhere - adding strings is very inefficient in python (strings are unmutable). It's much faster to assemble the strings in a list, and then join them. I won't request a change, but leave it to a discuss during our next telco. But this just asks for:
```
result_list = []
for statement in node.get_statements:
   result_list.extend( [self.n_indent, str(statement)] )
result_list.append("")
return "\n".join(result_list)
```
OK, that code wasn't tested, but just to get the idea (maybe we can even return (self._nintend+"\n").join(...), if we set up the list correctly.

src/psyclone/tests/psyGen_test.py
```diff
@@ -2682,8 +2679,24 @@ def test_codeblock_can_be_printed():
...
+
+    For simplicity we use a list of strings rather than an AST.
+
+    '''
+    original = ["hello", "there"]
+    cblock = CodeBlock(original)
+    result = cblock.get_statements
+    assert result == original
+    # Check that the list is a copy not a reference.
+    assert id(result) != id(original)
```
hiker: I learned something here :)

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -39,14 +39,18 @@
...
     FortranWriter
 from psyclone.psyGen import Symbol, Fparser2ASTProcessor, Node, CodeBlock
 
 
+@pytest.fixture(scope="function")
+def fort_writer():
+    '''Create and return a FortanWriter object with default settings.'''
+    return FortranWriter()
+
+
```
hiker: Great improvement that!

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -339,13 +300,12 @@ def test_fw_binaryoperator_unknown(monkeypatch):
     # Remove sign() from the list of supported binary operators
     monkeypatch.delitem(Fparser2ASTProcessor.binary_operators, "sign")
     # Generate Fortran from the PSyIR schedule
-    fvisitor = FortranWriter()
     with pytest.raises(VisitorError) as excinfo:
-        _ = fvisitor(schedule)
+        _ = fort_writer(schedule)
```
hiker: Is this assignment to _ actually necessary? Can't we simply ignore the result? Same for other uses like this further below.

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -199,57 +201,18 @@ class Unsupported(Node):
     # pylint: enable=abstract-method
 
     unsupported = Unsupported()
-    assignment = schedule.children[0]
+    assignment = schedule[0]
     binary_operation = assignment.children[1]
```
hiker: No idea why this is in the left column now ...
Isn't this assigment.rhs?

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -289,14 +251,14 @@ def test_fw_kernelschedule(monkeypatch):
 
     monkeypatch.setattr(schedule, "_name", None)
     with pytest.raises(VisitorError) as excinfo:
-        _ = fvisitor(schedule)
+        _ = fort_writer(schedule)
```
hiker: Do we need to assign to _?

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -616,49 +590,132 @@ def test_fw_codeblock():
...
+    :type parser: :py:class:`fparser.two.Fortran2003.Program`
+    :param str code: the code as a string.
+    :returns: the first schedule in the supplied code.
+    :rtype: :py:class:`psyclone.nemo.NemoInvokeSchedule`
+    '''
+    from psyclone.psyGen import PSyFactory
+    reader = FortranStringReader(code)
+    prog = parser(reader)
+    psy = PSyFactory(api="nemo").create(prog)
+    return psy.invokes.invoke_list[0].schedule
```
hiker: That seems pretty useful - should this perhaps be moved into tests/psyclone_test_utils.py?

hiker:
We discussed the issue with the unnecessary assignment to _ in the telco, and decided to put it to a vote :) There is some documentation aspect to assigning to _ (since it indicates that there is a return value, we are just not interested in it), on the other hand it just looks unnecessary complicated (we had _,_ = some_function_returning_a_tuple() in the past). So this is not required for this pull request.

The way we concatenate strings is a bit of a concern for me - considering the example of concatenating n lines, we have O(n) unnecessary copy operations (line1+line2 to temporary buffer, then this temporary  buffer plus line3 to another temporary buffer, ...) - the amount of data copies grows O(n^2). I am happy to discuss this as a separate issue (with the idea that the visitor pattern does not return a string, but a list - which then just gets "".join()ed).

Changes are covered, no todo for the issues in case left. No documentation change required.
Compile tests still running.

hiker:
Compilation passed as well (except for #473). There is also a minor conflict with master.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -496,11 +479,74 @@ def codeblock_node(self, node):
...
         Fortran string, indenting as appropriate.
 
+        At the moment it is not possible to distinguish between a
+        codeblock that is one or more full lines (and therefore needs
+        a newline added) and a codeblock that is part of a line (and
+        therefore does not need a newline). The current implementation
+        adds a newline irrespective. This is the subject of issue
+        #388.
+
         :param node: A CodeBlock PSyIR node.
```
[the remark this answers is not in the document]
rupertford: I've updated the file, rather than just fixing a few instances.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -496,11 +479,74 @@ def codeblock_node(self, node):
...
+    def nemoinvokeschedule_node(self, node):
+        '''A NEMO invoke schedule is the top level node in a PSyIR
+        representation of a NEMO program unit (program, subroutine
+        etc). It does not represent any code itself so all it needs to
+        to is call its children and return the result.
+
+        :param node: A NemoInvokeSchedule PSyIR node.
+        :type node: :py:class:`psyclone.nemo.NemoInvokeSchedule`
+
+        :returns: The Fortran code as a string.
```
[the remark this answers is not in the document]
rupertford: Done.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -496,11 +479,74 @@ def codeblock_node(self, node):
...
+        return result
+
+    def nemokern_node(self, node):
+        '''NEMO kernels are a group of nodes collected into a schedule
+        so simply call the nodes in the schedule.
+
+        :param node: A NemoKern PSyIR node.
+        :type node: :py:class:`psyclone.nemo.NemoKern`
+
+        :returns: The Fortran code as a string.
```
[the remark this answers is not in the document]
rupertford: Complete

src/psyclone/psyir/backend/fortran.py
```diff
@@ -496,11 +479,74 @@ def codeblock_node(self, node):
...
+        best way to capture their behaviour is decided. This method
+        outputs the Fortran representation of such a loop by simply
+        using the original Fortran ast (i.e. acting in a similar way
+        to a code block). As it is a temporary solution we don't
+        bother fixing the _ast internal access.
+
+        :param node: a NemoImplicitLoop PSyIR node.
+        :type node: :py:class:`psyclone.psyGen.NemoImplicitLoop`
+
+        :returns: The Fortran code as a string.
```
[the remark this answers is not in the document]
rupertford: Finished.

src/psyclone/tests/psyGen_test.py
rupertford: Andy and Sergi were impressed that you learnt something from me until I told them what it was and they pointed out that I should use "is", which is better, so I've changed it. :-)

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: Thanks.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: I'm going to leave this for now and we can discuss at Weds telco. My vote is to keep them as I think it makes it clear what would be returned if the function were successful or required.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: Again, let's defer until the telco.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: I can't get this to work. If you move the function to tests/psyclone_test_utils.py then there is no way to get access to it as the fortran_test.py file is in a subdirectory and neither relative nor absolute imports work.

This highlights the bigger problem that tests/psyclone_test_utils functions are not available to tests in subdirectories. The only way I can see that would work without code replication would be to make psyclone_test_utils part of the PSyclone package. I think this could be another Weds discussion topic.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: Changed to assignment.rhs. You can't modify rhs so I've added a comment before the reference (children[1]) is modified.

src/psyclone/psyir/backend/fortran.py
rupertford: OK, so we'll discuss at the next telco rather than chaning here.

rupertford:
Ready for next review from @hiker.

src/psyclone/tests/psyir/backend/fortran_test.py
hiker: This is covered in #508.

hiker:
I took the liberty of fixing a few minor pylint messages. All changes are now pylint clean. Changelog is updated, no need to rebuild documentation.

This can be merged once the tests are finished (@arporter , @sergisiso feel free to do this, it's getting too late for me, otherwise I'll see if I can do this on Monday),

arporter:
I'll leave it to @rupertford to decide what he wants to do with all the issues associated with this PR.

PR #479, opened 2019-08-27
Fix link to GungHo

kinow:
Hi,

I think the link to the paper is broken. MO has a page for GungHo similar to the one for LFRic, so I think this could be a replacement for the PDF article?

Cheers
Bruno

### DIFF
--- a/doc/user_guide/introduction.rst
+++ b/doc/user_guide/introduction.rst
@@ -10,7 +10,7 @@ started with the aim to support the emerging API in the GungHo project
 for a finite element dynamical core.
 
 The `GungHo project
-<https://www.metoffice.gov.uk/binaries/content/assets/mohippo/pdf/g/p/mosac_16.10.pdf>`_
+<https://www.metoffice.gov.uk/research/foundation/dynamics/next-generation>`_
 was initiated in 2011 to address challenges of weather and climate
 prediction on the next generation of supercomputers. The project ran for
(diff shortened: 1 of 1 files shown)


rupertford:
Thanks Bruno (@kinow), the pdf had more information on the GungHo project than the web page does,, so rather than just accepting your PR, I've asked the LFRic team if they know where the pdf has moved to and also whether they think it would be more robust to just reference the web page you suggest instead.

kinow:
Thanks @rupertford!

rupertford:
Thanks @kinow. I went with your proposed change. Note, I also added your name as a modifier in the relevant file. I was a bit naughty and did this change myself (rather than get you to do it) as it seemed a bit silly to get you to do it and we didn't have the information in the correct format (with copyright info) in any case!

kinow:
All good @rupertford , thanks for taking care of this. I couldn't find a `CONTRIBUTING.md`, but if there are any docs that contributors should read before writing issues/PR's, to give less work to reviewers, just let me know and I will read them before the next PR.

rupertford:
@kinow, that is a good point. I've made issue #484 to add a CONTRIBUTING.md file but in the meantime we have information on our wiki: https://github.com/stfc/PSyclone/wiki which has code review and interface markup guidelines. We also have a developers guide which would hopefully provide some useful information: https://psyclone-dev.readthedocs.io/en/latest/

Issue #495, opened 2019-09-05
Add openCL options for local-size and cmd_queue


sergisiso:
It should be possible to give a local-size parameter and a command_queue value to each openCL kernel.

Note that local-size should be consistent between the kernel code and the kernel calling statement.

Issue #494, opened 2019-09-05
C/OpenCL backend type casting operation is wrong 


sergisiso:
It should be `(type) variable` instead of `type(variable)`

hiker:
Afaik they are both valid C and C++
...
ah yes:
https://stackoverflow.com/questions/1652396/what-is-the-difference-between-typevalue-and-typevalue

sergisiso:
I think in pure C and therefore in OpenCL only the former is valid (at least Xilinx OpenCL compiler was failing with the second one). In C++ you can use either. So is better to always use `(type) variable` to avoid incompatibilities.

The change is quite simple, it is already implemented in the `c.py` backend on #496

PR #496, opened 2019-09-05
Fixes for nemoLite2d OpenCL generation (for #495, #494, ...)

sergisiso:
A few small fixes we need to have nemoLite2d working with the OpenCL back-end:
- [x] Fix C type casting operation #494 
- [x] ~~Kernel renaming with kernel name #499~~ (temporal Quickfix for OpenCL kernels)
- [x] Add local-size and command-queue parameters in each kernel #495
- [x] Remove duplication of width parameter and global_work_size()


sergisiso:
@arporter @rupertford This is ready for its first review. It contains the necessary changes to make the OpenCL generation work automatically (with some caveats which are marked with TODOs) for the nemolite2d application (it has been tested with the PSycloneBench opencl_psykal branch and it produces the same results than the CPU serial version). It also provides a set of options to start exploring OpenCL optimizations.

src/psyclone/psyGen.py
```diff
@@ -1768,15 +1800,21 @@ def gen_code(self, parent):
...
+        if self.opencl and self._opencl_options['end_barrier']:
+
             parent.add(CommentGen(parent,
                                   " Block until all kernels have finished"))
-            parent.add(AssignGen(parent, lhs=flag,
-                                 rhs="clFinish(" + qlist + "(1))"))
+
+            # We need a clFinish for all the queues in the implementation
+            opencl_num_queues = 1
+            for kern in self.coded_kernels():
```
arporter: This bit of code working out the number of queues is replicated from earlier. Could it be a method (or have the result stored somewhere)?

src/psyclone/psyGen.py
```diff
@@ -1768,15 +1800,21 @@ def gen_code(self, parent):
...
+                opencl_num_queues = max(
+                    opencl_num_queues,
+                    kern._opencl_options['queue_number'])
+            for i in range(1, opencl_num_queues + 1):
```
arporter: I suspect pylint/pep8 won't like `i`.

src/psyclone/psyGen.py
```diff
@@ -3773,6 +3812,38 @@ def get_kernel_schedule(self):
             self._kern_schedule = astp.generate_schedule(self.name, self.ast)
         return self._kern_schedule
 
+    def set_opencl_options(self, options):
+        '''
+        Validate and store a set of options associated to the Kernel to
```
arporter: s/associated to/associated with/

src/psyclone/psyGen.py
```diff
@@ -3773,6 +3812,38 @@ def get_kernel_schedule(self):
...
+        tune the OpenCL code generation.
+
+        :param options: a set of options to tune the OpenCL code.
+        :type options: map of <string>:<value>
```
arporter: s/map/dict/

src/psyclone/psyGen.py
```diff
@@ -3773,6 +3812,38 @@ def get_kernel_schedule(self):
...
+                        "integer.")
+            elif key == "queue_number":
+                if not isinstance(value, int):
+                    raise TypeError(
+                        "CodedKern opencl_option 'queue_number' should be an "
+                        "integer.")
+            else:
+                raise AttributeError(
+                    "CodedKern does not support the opencl_option '{0}'."
+                    "".format(key))
```
arporter: Please include the valid options in this error message.

src/psyclone/psyGen.py
```diff
@@ -3773,6 +3812,38 @@ def get_kernel_schedule(self):
...
+
+            self._opencl_options[key] = value
+
+        # If options are valid copy them in the associated KernelSchedule
```
arporter: Picky but s/in the/into the/
Also, why do we need a copy of these properties in the KernelSchedule? Is it possible to just have one single place to keep them?

src/psyclone/psyGen.py
```diff
@@ -3977,7 +4048,23 @@ def rename_and_write(self):
         fdesc = None
         while not fdesc:
             name_idx += 1
-            new_suffix = "_{0}".format(name_idx)
+            new_suffix = ""
+
+            # On the GOcean OpenCL work we need to differentiate between the
```
arporter: "...between kernels generated from the same module file..."

src/psyclone/psyGen.py
```diff
@@ -5963,6 +6052,17 @@ def symbol_table(self):
         '''
         return self._symbol_table
 
+    def set_opencl_options(self, options):
+        '''
+        Set the opencl_options map. Note that KernelSchedule expects to
+        receive options which are already vaildated by the caller.
```
arporter: I can see where you're coming from but I think it's better to be defensive and do the checks here too. Having said that, it would be even better if we only kept these properties in one place.

src/psyclone/psyir/backend/opencl.py
```diff
@@ -162,9 +162,18 @@ def kernelschedule_node(self, node):
         data_args = symtab.data_arguments
 
         # Start OpenCL kernel definition
-        code = self._nindent + "__kernel void " + node.name + "(\n"
+        code = self._nindent
+        if node._opencl_options['local_size'] != 1:
+            loc_size_value = node._opencl_options['local_size']
+            code += "__attribute__((vec_type_hint(double)))\n" \
+                    "".format(loc_size_value)
```
arporter: You're using `.format()` here but there's no e.g. `{}` or `{0}` in the string itself.

src/psyclone/tests/gocean1p0_opencl_test.py
```diff
@@ -327,6 +461,7 @@ def test_symtab_implementation_for_opencl():
         in str(err)
 
 
+@pytest.mark.xfail(reason="Use statment check is bypassed to make OpenCL work")
```
arporter: Please label this with an Issue number.

src/psyclone/transformations.py
```diff
@@ -2285,46 +2285,57 @@ def apply(self, sched, opencl=True):
         # All we have to do here is set the flag in the Schedule. When this
         # flag is True PSyclone produces OpenCL at code-generation time.
         sched.opencl = opencl
+
+        # This transformation stores some options into the InvokeSchedule
+        sched.set_opencl_options(options)
```
arporter: Since this call can raise Exceptions I think we should catch them here and re-raise them as a `TransformationError` for consistency with other transformations.

doc/user_guide/transformations.rst
```diff
@@ -528,15 +528,76 @@ of the FortCL library (https://github.com/stfc/FortCL) to access
...
+.. note:: The generated OpenCL files follow the `--kernel-renaming` argument
+    conventions, but in addition to the modulename they also include the
+    kernelname as part of the filename in the format:
+    `modulename_kernelname_index.cl`
+
+
+
+The ``OCLTrans`` transformation accepts an `options` argument with a
+map of optional parameters to tune the OpenCL host code in the PSy layer.
+This options will be attached to the transformed InvokeSchedule.
```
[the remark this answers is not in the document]
sergisiso: Fixed

doc/user_guide/transformations.rst
```diff
@@ -528,15 +528,76 @@ of the FortCL library (https://github.com/stfc/FortCL) to access
...
+
++--------------+----------------------------------------------+---------+
+| Option       |  Description                                 | Default |
++==============+==============================================+=========+
+| end_barrier  | Whether a synchronization                    | True    |
+|              | barrier should be placed at the end of the   |         |
+|              | Invoke.                                      |         |
++--------------+----------------------------------------------+---------+
+
+Additionally, each individual kernel inside the Invoke that is going
```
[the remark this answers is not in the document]
sergisiso: You meant delete the part inside the parenthesis right? In this case Done.

doc/user_guide/transformations.rst
```diff
@@ -528,15 +528,76 @@ of the FortCL library (https://github.com/stfc/FortCL) to access
...
+to be transformed, also accepts a map of options which are provided by
+the `set_opencl_options()` method of the `Kern` object.
+This can affect both, the driver layer or the OpenCL kernels.
```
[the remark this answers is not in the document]
sergisiso: Fixed

doc/user_guide/transformations.rst
```diff
@@ -528,15 +528,76 @@ of the FortCL library (https://github.com/stfc/FortCL) to access
...
+The current available options are:
+
++--------------+---------------------------------------------+---------+
+| Option       |  Description                                | Default |
++==============+=============================================+=========+
+| local_size   | Number of work-items to compute             | 1       |
+|              | in a single kernel.                         |         |
++--------------+---------------------------------------------+---------+
+| queue_number | Queue identifier number where               | 1       |
```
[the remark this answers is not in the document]
sergisiso: Replaced text

doc/user_guide/transformations.rst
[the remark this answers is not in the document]
arporter: No, sorry, that's not what I meant. I meant put the text "inside the Invoke that is going to be transformed" inside parentheses :-)

doc/user_guide/transformations.rst
```diff
@@ -528,15 +528,76 @@ of the FortCL library (https://github.com/stfc/FortCL) to access
...
++==============+=============================================+=========+
+| local_size   | Number of work-items to compute             | 1       |
+|              | in a single kernel.                         |         |
++--------------+---------------------------------------------+---------+
+| queue_number | Queue identifier number where               | 1       | 
+|              | to enqueue the associated kernel.           |         |
++--------------+---------------------------------------------+---------+
+
+
+See below an example of an PSyclone script that uses an ``OCLTans`` with
```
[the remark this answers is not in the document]
sergisiso: Done

doc/user_guide/transformations.rst
```diff
@@ -528,15 +528,76 @@ of the FortCL library (https://github.com/stfc/FortCL) to access
...
+multiple InvokeSchedule and kernel-specific transformations.
+
+
+.. code-block:: python
```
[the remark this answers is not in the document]
sergisiso: Yes, the `.. literalinclude:: ../../examples/gocean/eg3/ocl_trans.py` command did the trick

src/psyclone/psyGen.py
```diff
@@ -429,23 +429,28 @@ def gen_code(self, parent):
...
             invoke.gen_code(parent)
             # If we are generating OpenCL for an Invoke then we need to
             # create routine(s) to set the arguments of the Kernel(s) it
             # calls. We do it here as this enables us to prevent
             # duplication.
             if invoke.schedule.opencl:
+                generate_ocl_init = True
                 for kern in invoke.schedule.coded_kernels():
                     if kern.name not in opencl_kernels:
+                        opencl_num_queues = max(
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -1661,8 +1668,33 @@ def __init__(self, KernFactory, BuiltInFactory, alg_calls=None):
         Schedule.__init__(self, sequence=sequence, parent=None)
         self._invoke = None
         self._opencl = False  # Whether or not to generate OpenCL
+        # InvokeSchedule opencl_options default values
+        self._opencl_options = {"end_barrier": True}
         self._name_space_manager = NameSpaceFactory().create()
 
+    def set_opencl_options(self, options):
+        '''
+        Validate and store a set of options associated to the InvokeSchedule to
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -1661,8 +1668,33 @@ def __init__(self, KernFactory, BuiltInFactory, alg_calls=None):
...
+        tune the OpenCL code generation.
+
+        :param options: a set of options to tune the OpenCL code.
+        :type options: map of <string>:<value>
```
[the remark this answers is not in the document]
sergisiso: updated docstring

src/psyclone/psyGen.py
sergisiso: They are not exactly the same as the one for the opencl_init gets the maximum of all Invokes to allocate the resources, while this one is the maximum for the specific Invoke because it just needs barriers for the queues it uses.

Is it true that the former could include a call to the latter, but it will still be two max() operations very similar to the current implementations, so I think it's better to keep it this way for now.

I also prefer not to have this value stored as even if it is not implemented yet I like the idea of some 'smart' scheduling transformation that changes the queues of each kernel at its convenience and it is only when we call `gen_code` that this has to be fully determined.

src/psyclone/psyGen.py
```diff
@@ -1661,8 +1668,33 @@ def __init__(self, KernFactory, BuiltInFactory, alg_calls=None):
...
+        # Validate that the options given are supported and store them
+        for key, value in options.items():
+            if key == "end_barrier":
+                if not isinstance(value, bool):
+                    raise TypeError(
+                        "InvokeSchedule opencl_option 'end_barrier' "
+                        "should be a boolean.")
+            else:
+                raise AttributeError(
+                    "InvokeSchedule does not support the opencl_option '{0}'."
```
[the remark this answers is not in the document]
sergisiso: Done. I used the plural tense "Accepted options are: 'end_barrier'." as this may accept more options if the future. Is this right? or should I use the singular form for now?

src/psyclone/psyGen.py
sergisiso: Of course, updated to `queue_number` :)

src/psyclone/psyGen.py
sergisiso: Fixed

src/psyclone/psyGen.py
sergisiso: Fixed

doc/user_guide/transformations.rst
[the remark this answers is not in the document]
sergisiso: I got you

src/psyclone/psyGen.py

src/psyclone/psyGen.py

src/psyclone/psyir/backend/opencl.py
sergisiso: Ups, fixed.

src/psyclone/tests/gocean1p0_opencl_test.py
sergisiso: Done

sergisiso:
@arporter I do agree that the replicated Kern/KernelSchedule opencl_options is not a good design. Looking into it I think they should be in Kern call as not all options are needed in the KSchedule (e.g. num_queue). And in the future we may have multiple calls referencing the same kernel and with distinct opencl_options.

The problem is KernelSchedule does not have a reference to its callee to query the information. In reality we just need to have the 'local_size' (blocking parameter) and there is a few solutions for this: 

1. Add a reference from KernelSchedule to its callee. This is possible because right now each call parses its own instance of the kernel so there is a 1-to-1 match. But I don't like this option as I think each KernelSchedule should be independent of Kern and have the possibility to reuse them.
2.  Add just 'opencl_local_size' as an annotation to KernelSchedule. We would reuse the annotations infrastructure and it fits well with the end result we want which is an '__attribute_(...)' which is kind of a C annotation.
3. When we call the OpenCL visitor from Kern, we can specify an argument `ocl_writer = OpenCLWriter(kernels_local_size=4); ocl_writer(self.get_kernel_schedule())`. This works because we instantiate an OpenCLWriter for each kernel we call.

I am inclined to option 2, but I would like to know your opinion before implementing it.

rupertford:
Would it make sense for a KernelSchedule to have a reference to 1 or more callees i.e. implement option 1 but with a 1..n relationship?

arporter:
I think I'm a bit confused about the relationship between a `Kern` and a `KernelSchedule`. A given `Kern` must have only one `KernelSchedule` but multiple `Kern`s could all refer to the same `KernelSchedule`. However, in the latter case, we could conceivably want different local sizes for different kernel calls? Since that has to be hard-coded at compile time we would then need to generate different versions of the same kernel. That then feels like the kernels are in fact different and this should be reflected by having different `Kern` objects in the Schedule. Is that correct?

rupertford:
I think the idea of multiple Kerns referencing the same KernelSchedule would be a new concept in PSyclone. I think that at the moment we assume a 1-1 relationship internally (but we do something when writing them out to file with the 'single' etc options).

If we were to go down this route then all changes to KernelSchedule would be "seen" by all Kerns that reference it and if we wanted different things to happen in different contexts we would have to have a transformation to split/replicate the KernelSchedule (giving it a different name/instance id).

sergisiso:
That's right, at the moment each Kern generate/parses its associated KernelSchedule and maintains a reference Kern->KernelSchedule, so Kerns pointing to the same PSyKAl kernels will have individual KernelSchedules.

I am not proposing changing this implementation but I want to avoid making a Kern<-KernelSchedule reference as I think it does not fit conceptually even if the implementation would work.

This is why I think options 2 and 3 are better (although they have their own problems).

arporter:
Thanks for clarifying. I'm not keen on option 2 as it feels like a re-purposing of `annotations`: currently we only use them to store information on where Nodes in the AST came from whereas with this change they would affect the generated code. I quite like option 3 - what are the drawbacks you see with that approach?

sergisiso:
After thinking again about this I now also prefer option 3, so it is the one that I will implement.

The good things about 3 is that it is specific of the OpenCL backend, so the data is just where it is needed and as currently implemented the Writer instacnce is short-lived (just created for generating 1 kernel and then destroyed, so the duplication of the opencl options are not permanent).

At first I though it was odd to give a parameter that just the kernel node cares about to the general back-end. But now I think this is because openCL conveniently has the special purpose attribute but if not it could affect multiple nodes.

src/psyclone/transformations.py

src/psyclone/psyGen.py
sergisiso: Removed this method as agreed in the conversation below.

src/psyclone/psyGen.py
sergisiso: Removed these 3 lines to avoid duplicating the Kernel options.

sergisiso:
@arporter This is ready for the next review.

The main change is the removal of the KernelSchedule opencl_options duplication and I added instead a OpenCLWriter argument to provide the 'local_size'. I addressed all other inline comments but the one requesting a method to compute the maximum number of queues, see justification in-line.

src/psyclone/psyGen.py
```diff
@@ -1768,15 +1803,22 @@ def gen_code(self, parent):
...
                                   " Block until all kernels have finished"))
-            parent.add(AssignGen(parent, lhs=flag,
-                                 rhs="clFinish(" + qlist + "(1))"))
+
+            # We need a clFinish for all the queues in the implementation
+            opencl_num_queues = 1
+            for kern in self.coded_kernels():
+                opencl_num_queues = max(
+                    opencl_num_queues,
+                    kern._opencl_options['queue_number'])
```
arporter: pylint complains about this reference to `_opencl_options` - I guess we need a getter.

src/psyclone/tests/gocean1p0_opencl_test.py
```diff
@@ -254,13 +359,43 @@ def test_opencl_kernel_code_generation():
...
+def test_opencl_kernel_output_file(kernel_outputdir):
+    '''Check that an OpenCL file named modulename_kernelname_0 is generated.
+    '''
+    import os
+    psy, _ = get_invoke("single_invoke.f90", API, idx=0)
+    sched = psy.invokes.invoke_list[0].schedule
+    otrans = OCLTrans()
+    otrans.apply(sched)
+    sched.kernels()[0].name = "name"
+    generated_code = str(psy.gen).lower()
```
arporter: pylint complains that `generated_code` is unused. We could just remove the assignment part of this statement altogether (as I think we've agreed/voted to not use `_` in tests unless absolutely necessary).

src/psyclone/tests/gocean1p0_opencl_test.py
```diff
@@ -254,13 +359,43 @@ def test_opencl_kernel_code_generation():
...
+def test_opencl_kernel_output_file_with_suffix(kernel_outputdir):
+    '''Check that an OpenCL file named modulename_kernelname_0 is
+    generated without the _code suffix in the kernelname
+    '''
+    import os
+    psy, _ = get_invoke("single_invoke.f90", API, idx=0)
+    sched = psy.invokes.invoke_list[0].schedule
+    otrans = OCLTrans()
+    otrans.apply(sched)
+    generated_code = str(psy.gen).lower()
```
arporter: Again, `generated_code` is unused.

src/psyclone/tests/gocean1p0_transformations_test.py
```diff
@@ -1475,7 +1475,8 @@ def test_ocl_apply(kernel_outputdir):
     # transformation to something that is not an InvokeSchedule
     with pytest.raises(TransformationError) as err:
         _, _ = ocl.apply(schedule.children[0])
-    assert "the supplied node must be a (sub-class of) Schedule " in str(err)
+    assert "the supplied node must be a (sub-class of) InvokeSchedule " \
+        "" in str(err)
```
arporter: I don't think you need the '""' on this line.

src/psyclone/transformations.py
```diff
@@ -2265,7 +2265,7 @@ def name(self):
         '''
         return "OCLTrans"
 
-    def apply(self, sched, opencl=True):
+    def apply(self, sched, opencl=True, options={}):
```
arporter: I should have picked this up first time around but pylint warns that `{}` is a dangerous default value. Please use None instead and change the method body appropriately.

doc/user_guide/transformations.rst
```diff
@@ -561,33 +561,19 @@ The current available options are:
...
-| queue_number | Queue identifier number where               | 1       | 
-|              | to enqueue the associated kernel.           |         |
+| queue_number | The identifier of the OpenCL Command Queue  | 1       | 
+|              | to which the kernel should be submitted.    |         |
 +--------------+---------------------------------------------+---------+
 
 
-See below an example of an PSyclone script that uses an ``OCLTans`` with
-multiple InvokeSchedule and kernel-specific transformations.
+Below is an example of an PSyclone script that uses an ``OCLTans`` with
```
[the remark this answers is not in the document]
sergisiso: Fixed

src/psyclone/psyGen.py
```diff
@@ -1690,7 +1692,8 @@ def set_opencl_options(self, options):
                         "should be a boolean.")
             else:
                 raise AttributeError(
-                    "InvokeSchedule does not support the opencl_option '{0}'."
+                    "InvokeSchedule does not support the opencl_option '{0}'. "
+                    "The supported options are: 'end_barrier'."
```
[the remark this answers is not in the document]
sergisiso: Done for the 1696, but I am not sure if I should do it for 1688 as the error is specific to that key.

src/psyclone/psyGen.py
```diff
@@ -3835,15 +3839,12 @@ def set_opencl_options(self, options):
                         "integer.")
             else:
                 raise AttributeError(
-                    "CodedKern does not support the opencl_option '{0}'."
-                    "".format(key))
+                    "CodedKern does not support the opencl_option '{0}'. "
```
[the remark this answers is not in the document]

src/psyclone/psyir/backend/opencl.py
```diff
@@ -47,7 +47,38 @@ class OpenCLWriter(CWriter):
...
+        super(OpenCLWriter, self).__init__(
+            skip_nodes, indent_string, initial_indent_depth)
+
+        if not isinstance(kernels_local_size, int):
+            raise TypeError(
+                "kernel_local_size should be an integer but found "
+                "'{0}'.".format(type(kernels_local_size).__name__))
+
+        if kernels_local_size < 1:
+            raise TypeError(
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/backend/opencl.py
```diff
@@ -163,12 +194,10 @@ def kernelschedule_node(self, node):
 
         # Start OpenCL kernel definition
         code = self._nindent
-        if node._opencl_options['local_size'] != 1:
-            loc_size_value = node._opencl_options['local_size']
-            code += "__attribute__((vec_type_hint(double)))\n" \
-                    "".format(loc_size_value)
+        if self._kernels_local_size != 1:
+            code += "__attribute__((vec_type_hint(double)))\n"
```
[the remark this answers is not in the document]
sergisiso: I will delete this line for now, although recommended in the Xilinx User Guide I didn't observe any performance difference.

I explore this further and re-introduce it in the future if necessary.

src/psyclone/tests/gocean1p0_opencl_test.py
sergisiso: This is needed because it generates the OpenCL kernel as side-effect. I removed the `str().lower()`, used `_` and added a comment mentioning the side-effect.

src/psyclone/tests/gocean1p0_opencl_test.py
sergisiso: Same as above

src/psyclone/tests/gocean1p0_transformations_test.py
sergisiso: Ups, deleted.

src/psyclone/transformations.py
sergisiso: Fixed

src/psyclone/psyGen.py
sergisiso: Added a property to access opencl_options

sergisiso:
@arporter If the unit-test passes, this is ready for next review. I also check again that the OpenCL still produces exact same results as the serial version for NemoLite2D after the changes.

src/psyclone/tests/gocean1p0_opencl_test.py
arporter: Great, thanks.

arporter:
All requested changes have been made. All tests still pass/xfail, including with compilation.

arporter:
Once #511 is merged, this is ready to go (needs changelog and UG update).

Issue #388, opened 2019-05-21
CodeBlock needs start and end of line information


rupertford:
When translating an instance of a CodeBlock class back to Fortran the end of the instance may be at the end of a line, or within a line. If it is at the end of a line then there needs to be a newline at the end of the CodeBlock instance (but not if it is not). Either this newline should be kept as part of a CodeBlock instance (it is not now) or a method indicating whether it is at the end of a line needs to be added to a CodeBlock.

SImilarly, if the start of the CodeBlock instance is at the beginning of a line then it may need to be indented. The CodeBlock instance needs a method to indicate whether this is the case (or not).

sergisiso:
Would having a way to differentiate between statement/s and expression work?

At the moment we are fully abstract, which is nice, but the tree accepts meaningless things like having a Loop in the If child representing the condition. We could categorize Loop, If, Assignment ,... as statements and Literal, Reference, Operation, ... as expressions, CodeBlock could be both. Having statement/expression differentiation would allow for better consistency tests and the backend can assume indentation+lb if it is a statement or no indent+lb if it is an expression.

rupertford:
Hi @sergisiso. That might work. At the moment I explicitly add newlines when a rule is also a statement, but the PSyIR hierarchy does not always distinguish (I don't think).

For example, what about the case where a codeblock is an expression but also ends the line, such as
``` a = max(b, c)```
With my current fparser branch the fparser2 parse tree adds an `Intrinsic_Function_Reference` fparser2 node which is not recognised by the fparser2_to_PSyIR front end, so the RHS becomes a code block. There is no parent `statement` node here (I don't think) so the codeblock also terminates the line. If we had a wrapper `statement` node that might work but I can't see how it could be done otherwise (apart from doing what I originally suggested).

**Comment by sergisiso** [comment]
Isn't the Assignment the statement here? So it should print: `<Indent> LHS.gen = RHS.gen \n`

This CodeBlock should be marked as an expression in the front end, and then the back-end should not allow for an `\n` when doing the `str(fparser_ast)`

I wasn't thinking about a wrapper node, I was thinking more in the lines of an attribute of Node which can be set to statement or expression.

**Comment by rupertford** [comment]
You are right about Assignment and that is what I already have ...
``` 
        lhs = self.visit(node.children[0])
        rhs = self.visit(node.children[1])
        result = "{0}{1}={2}\n".format(self._nindent, lhs, rhs)
```
[...]

PR #498, opened 2019-09-06
issue #388. CodeBlock statements and expressions

rupertford:


rupertford:
Ready for review from @sergisiso.

src/psyclone/psyGen.py
```diff
@@ -5984,6 +6001,18 @@ def __init__(self, statements, parent=None):
         else:
             self.ast = None
             self.ast_end = None
+        # Store any Code Block properties.
+        self._structure = structure
```
sergisiso: Do we need type checking to verify the argument `structure` is an element or the Enum `Structure`?

src/psyclone/psyGen.py
```diff
@@ -6598,7 +6627,32 @@ def nodes_to_code_block(parent, statements):
...
+                structure = CodeBlock.Structure.STATEMENT
+            elif isinstance(prev_node, (Reference, Operation, Literal)):
+                # The previous node is an expression, therefore this
+                # code block must be an expression.
+                structure = CodeBlock.Structure.EXPRESSION
+            else:
+                raise InternalError(
+                    "Fparser2ASTProcessor:nodes_to_codeblock: Unsupported "
+                    "node type '{0}' found.".format(type(prev_node).__name__))
+        else:
```
sergisiso: OK until here, but the else condition (Codeblock has no sibling yet) does not imply it is an statement, it can be the first node of an expression,

See my example on the unit-tests below.

src/psyclone/tests/psyGen_test.py
```diff
@@ -5513,6 +5532,87 @@ def test_fparser2astprocessor_do_construct_while(f2008_parser):
...
+def test_fp2astproc_nodes_to_code_block_3(f2008_parser):
+    '''Check that a codeblock that contains an expression has the
+    structure property set to expression.
+
+    '''
+    from fparser.common.readfortran import FortranStringReader
+    # The string "HELLO" is currently a code block in the PSyIR
+    reader = FortranStringReader('''
+        program test
+        if (a == "HELLO") then
```
sergisiso: I think this wont't work if the condition is reversed: `if ("HELLO" == a) then` then it returns a CodeBlock with structure =  STATEMENT

src/psyclone/psyGen.py
```diff
@@ -5967,10 +5967,27 @@ class CodeBlock(Node):
     :param statements: list of fparser2 AST nodes representing the Fortran \
                        code constituting the code block.
     :type statements: list of :py:class:`fparser.two.utils.Base`
+    :param structure: optional argument indicating whether this code \
```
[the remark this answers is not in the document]
rupertford: I made it optional as I thought it might not be something that makes sense for a CodeBlock in a different language. Do you think it is a generic property?

src/psyclone/psyGen.py
```diff
@@ -5967,10 +5967,27 @@ class CodeBlock(Node):
...
+    block is a statement or an expression.
+    :type structure: :py:class:`psyclone.psyGen.CodeBlock.Structure` \
+    or `NoneType`
```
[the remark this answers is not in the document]
rupertford: See above.

src/psyclone/psyGen.py
```diff
@@ -5984,6 +6001,18 @@ def __init__(self, statements, parent=None):
         else:
             self.ast = None
             self.ast_end = None
+        # Store any Code Block properties.
```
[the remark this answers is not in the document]
rupertford: Yes, it was left over from when I was thinking of supporting a list of general properties. I've changed the words.

src/psyclone/psyGen.py
rupertford: I thought about this then decided that it is not something that a user would specify, it is something that is created by a front end so thought it would be OK to assume the inputs are correct. However, I can obviously add tests and checks if you think it makes sense.

src/psyclone/psyGen.py
rupertford: Yes, quite right, the code would never work. I've updated taking your advice i.e. checking whether the parent is a schedule or not (If yes then it's a statement, if not it's an expression), raising an exception if it is a directive (as these require further discussion in my opinion).

src/psyclone/tests/psyGen_test.py
rupertford: Agreed, please see previous comment. I obviously had a serious brain block before.

rupertford:

src/psyclone/psyGen.py
sergisiso: OK, I am happy to assume the inputs are correct in this case.

src/psyclone/tests/psyGen_test.py
```diff
@@ -5513,6 +5532,86 @@ def test_fparser2astprocessor_do_construct_while(f2008_parser):
     assert isinstance(fake_parent.children[0], CodeBlock)
 
 
+# (1/4) fparser2astprocessor::nodes_to_code_block
+def test_fp2astproc_nodes_to_code_block_1(f2008_parser):
+    '''Check that a statement codeblock that is at the "top level" in the
+    PSyiR has the structure property set to statement (as it has a
```
sergisiso: s/PSyiR/PSyIR
Also in the test below.

src/psyclone/tests/psyGen_test.py
```diff
@@ -5513,6 +5532,86 @@ def test_fparser2astprocessor_do_construct_while(f2008_parser):
...
+# (4/4) fparser2astprocessor::nodes_to_code_block
+def test_fp2astproc_nodes_to_code_block_4(f2008_parser):
+    '''Check that a codeblock that has a directive as a parent causes the
+    expected exception.
+
+    '''
+    with pytest.raises(InternalError) as excinfo:
+        _ = Fparser2ASTProcessor.nodes_to_code_block(Directive(), "hello")
+    assert ("A CodeBlock with a Directive as parent is not yet supported."
+            in (str(excinfo.value)))
```
sergisiso: pylint says: [superfluous-parens] Unnecessary parens after 'in' keyword

src/psyclone/psyGen.py
```diff
@@ -5967,10 +5968,27 @@ class CodeBlock(Node):
...
+        may be required when processing.
+
+        '''
+        # The Code Block comprises one or more Fortran statements
+        # (which themselves may contain expressions).
+        STATEMENT = 1
+        # The Code Block comprises one or more Fortran expressions.
+        EXPRESSION = 2
+
+    def __init__(self, statements, structure=None, parent=None):
```
sergisiso: The `statements` and `self._statements` should probably be renamed to something else (nodes?, fp2_nodes?), as they are not just statements. This applies to all the `CodeBlock` class, including docstrings.

src/psyclone/tests/psyGen_test.py
rupertford: Oops.

src/psyclone/tests/psyGen_test.py
rupertford: Well spotted.

src/psyclone/psyGen.py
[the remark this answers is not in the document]
rupertford: OK, I've made the argument mandatory.

src/psyclone/psyGen.py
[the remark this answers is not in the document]
rupertford: Ditto :-)

src/psyclone/psyGen.py
rupertford: OK, I think I've done this. I also modified the names in the nodes_to_code_block function and noticed that the docstrings were not correct so fixed those too.

rupertford:

Issue #502, opened 2019-09-09
Add support for the Fortran SIZE() intrinsic to the PSyIR.


arporter:
In #497 we need to construct loop bounds using the Fortran SIZE intrinsic. The PSyIR does not yet have support for this. I'm unsure how to do this in a language-agnostic way since languages like C don't provide this functionality - it's entirely up to the programmer to keep track of how much memory they allocated to a variable.

sergisiso:
This would also help with #440 and #412

rupertford:
pr #503 is on master so closing this issue.

Issue #440, opened 2019-07-15
Extend Loops to be language independent


sergisiso:
To simplify the #400  PR I left some expressions like  `"SIZE(" + self.field_name + "," + index + ")"` inside `Literals`.  This is a useful workaround for things that are not implemented in the PSyIR but PSyclone was already writing directly. It works for generating Fortran (abeit a bit unsafe) but is not language independent.

This is the list of non-language independent features left in the Loops implementation:

- [ ] SIZE intrinsic
- [ ] self._lower_bound_fortran() and self._upper_bound_fortran() outputs
- [ ] Derived Types (% operator)
- [ ] References should be checked in the Symbol table if this exist
(...)

arporter:
Just to note that support for derived types is the subject of #363

LonelyCat124:
@sergisiso Any idea where we are with this now? I'm looking at a test:
```
def test_gocean_omp_parallel():
    '''Test that an OMP PARALLEL directive in a 'classical' API (gocean here)
    is created correctly.
    '''
    _, invoke = get_invoke("single_invoke.f90", "gocean",
                           idx=0, dist_mem=False)

    omp = OMPParallelTrans()
    omp.apply(invoke.schedule[0])
    print(invoke.schedule[0].debug_string())

    # Now remove the GOKern (since it's not yet supported in the
    # visitor pattern) and replace it with a simple assignment
    # TODO: #440 tracks this
    replace_child_with_assignment(invoke.schedule[0].dir_body)

    # omp_sched is a GOInvokeSchedule, which is not yet supported.
    # So only convert starting from the OMPParallelDirective
    fvisitor = FortranWriter()
    result = fvisitor(invoke.schedule[0])
    correct = '''!$omp parallel default(shared)
a = b
!$omp end parallel'''
    assert correct in result
```

The print I've added outputs:
```
!$omp parallel default(shared) private(i,j)
do j = cu_fld%internal%ystart, cu_fld%internal%ystop, 1
  do i = cu_fld%internal%xstart, cu_fld%internal%xstop, 1
    < kern call: compute_cu_code >
  enddo
enddo
!$omp end parallel
```
Is there some support still missing?

LonelyCat124:
Also replacing the print to output `FortranWriter()(invoke.schedule[0])` gives:
```
!$omp parallel default(shared) private(i,j)
do j = cu_fld%internal%ystart, cu_fld%internal%ystop, 1
[...]

Issue #363, opened 2019-04-25
Represent Fortran derived types in the PSyIR


arporter:
Some sections of NEMO (notably those dealing with sea ice and ice bergs) make use of Fortran derived types. These are currently not handled by the PSyIR and are left as CodeBlocks which then means we can't have Kernels that contain derived types.

arporter:
I can follow what has been described here: https://psyclone-dev.readthedocs.io/en/latest/psyir_symbols.html#datatypes

When processing existing code we will often encounter variables that are declared as being of a certain derived type. At that point we will know that there is such a derived type but not what it consists of (because it will typically be defined in some other Container).

arporter:
Brought Rupert, Sergi and Joerg into the discussion (that I'm currently having with myself) in case they spot anything.

arporter:
Obviously, one can have an array of structures/derived types. Currently the `ArrayType` constructor expects the supplied datatype to have `intrinsic` and `precision` attributes which a derived type won't. However, support for arrays of structures might be a separate Issue - I'll see how it goes.

sergisiso:
>  At that point we will know that there is such a derived type but not what it consists of

Would then declaring it as 'DeferredType' be enough or there is some information that it would miss.

> constructor expects the supplied datatype to have intrinsic and precision attributes which a derived type won't.

I think that was to simplify the modifications on existing code at the time, but I agree an ArrayType should support any kind of DataType as its element.

arporter:
> > At that point we will know that there is such a derived type but not what it consists of
> 
> Would then declaring it as 'DeferredType' be enough or there is some information that it would miss.
> 
Yes, I was wondering that - I'll have a look and see.

arporter:
A derived type isn't just a list of types (and routines), it contains *named* entities, e.g.:

    type :: my_type
      integer :: flag
      real :: g
    end type my_type

contains a member named "flag" that is of type integer. This is precisely the type of information that is captured in the SymbolTable. So, does it make sense for a StructureType to have its own SymbolTable?

arporter:
The scope in which `my_type` is declared must have a symbol for it. I think this means we need a new type of symbol, `TypeSymbol` perhaps?

arporter:
[...]

PR #503, opened 2019-09-11
For #502 - support the SIZE intrinsic

arporter:


arporter:
I think this is ready for review, probably by @sergisiso or @rupertford.

arporter:
I noticed that the Sphinx AutoAPI feature is not working when the 'latest' documentation is built on RTD. However, I've traced this back to the fact that we are using features of fparser that are not yet in a release and RTD installs fparser from PyPI. I therefore propose to just live with this. The 'stable' version of the documentation should always be OK because that is for a release and a release of PSyclone must work with a released version of fparser.

sergisiso:
> I have to say that I do not like the fact there appears to be no way to calculate the equivalent of SIZE in C, as this means that we can easily end up with a PSyIR representation of code that can be output in Fortran but not in C which kind of defeats one of the main benefits of the PSyIR.

@rupertford It depends what we mean by Array in PSyIR, if by Array we mean a container with a pointer and a length (similar to Fortran arrays or C++ std::vector) then I see no problem in writing the C back-end with size() and array operations supported.

src/psyclone/tests/psyir/frontend/size_intrinsic_test.py
```diff
@@ -0,0 +1,61 @@
...
+        Reference, Literal
+    fake_parent = Schedule()
+    processor = Fparser2ASTProcessor()
+    for expr in ["n = SIZE(a, 3)", "n = SIZE(a(:,:,:), 3)"]:
+        reader = FortranStringReader(expr)
+        fp2intrinsic = Execution_Part(reader).content[0]
+        processor.process_nodes(fake_parent, [fp2intrinsic], None)
+        assert isinstance(fake_parent[0], Assignment)
+        assert isinstance(fake_parent[0].rhs, BinaryOperation)
+        fake_parent[0].rhs.view()
```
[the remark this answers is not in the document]
arporter: Oh yes, done.

src/psyclone/tests/psyir/frontend/size_intrinsic_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_size(parser):
+    ''' Basic test that the SIZE intrinsic is recognised and represented
+    in the PSyIR. '''
+    from fparser.two.Fortran2003 import Execution_Part
+    from psyclone.psyGen import Schedule, Assignment, BinaryOperation, \
+        Reference, Literal
+    fake_parent = Schedule()
+    processor = Fparser2ASTProcessor()
+    for expr in ["n = SIZE(a, 3)", "n = SIZE(a(:,:,:), 3)"]:
```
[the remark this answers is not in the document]
arporter: Yes, that's nicer - done.

arporter:
> 
> 
> The comments in gocean0p1.py line 273 need to be addressed somehow. Either make use of the new intrinsic or create a new issue that explains what needs to be done and update the reference (as well as updating #440).

I've done this in both `gocean0p1.py` and `gocean1p0.py` by altering the routines that create the upper/lower bounds to all return PSyIR. There's still the issue of derived types but I've put in a TODO for those.

arporter:
This is now ready for 2nd review from @rupertford...

src/psyclone/tests/psyir/frontend/size_intrinsic_test.py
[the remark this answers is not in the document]
rupertford: Great.

rupertford:
> > I have to say that I do not like the fact there appears to be no way to calculate the equivalent of SIZE in C, as this means that we can easily end up with a PSyIR representation of code that can be output in Fortran but not in C which kind of defeats one of the main benefits of the PSyIR.
> 
> @rupertford It depends what we mean by Array in PSyIR, if by Array we mean a container with a pointer and a length (similar to Fortran arrays or C++ std::array) then I see no problem in writing the C back-end with size() and array operations supported.

Thanks @sergisiso. There was a direct conversation @sergisiso, @arporter and I and @sergisiso has reassured us that all will be fine when translating from PSyIR to different languages via different backends :-).

@sergisiso also pointed out that we will need the size of arrays in general, even if we do not use the SIZE intrinsic so if and when we add a C/C++ front end we might have issues as native C arrays are implemented as pointers without sizes. We would therefore need to constrain C/C++ codes to cases where we could capture the sizes of arrays (via another variable probably) or where the C code uses appropriate array types (e.g. C++ Arrays).

src/psyclone/gocean1p0.py
```diff
@@ -719,8 +735,15 @@ def _lower_bound(self):
         All occurences of {start} and {stop} in _bounds_loopup will
         be replaced with the constant loop boundary variable, e.g.
         "{stop}+1" will become "istop+1" (or "jstop+1" depending on
-        loop type).'''
+        loop type).
 
+        :param parent:
```
rupertford: A little more verbose please for the argument descriptions :-)
And please remove the "]"

src/psyclone/gocean1p0.py
```diff
@@ -782,8 +806,11 @@ def view(self, indent=0):
         '''
         # Generate the upper and lower loop bounds
         # TODO: Issue 440. upper/lower_bound should generate PSyIR
-        self.start_expr = Literal(self._lower_bound(), parent=self)
-        self.stop_expr = Literal(self._upper_bound(), parent=self)
+        self.start_expr = self._lower_bound(parent=self)
```
rupertford: Can the above TODO be removed now seeing as lower and upper bounds do generate PSyIR (albeit with Literal)? The reference to #363 captures future improvement to better support bounds in PSyIR.

src/psyclone/gocean0p1.py
```diff
@@ -289,6 +292,10 @@ def gen_code(self, parent):
                 self.stop_expr = Reference(
                     self.field_space + "%jstop", parent=self)
 
+        # Update our bounds in the PSyIR
+        self.children[0] = self.start_expr
```
rupertford: Isn't self.start_expr the same as self.children[0]? Similarly for self.stop_expr? If so, you don't need these lines.

src/psyclone/gocean1p0.py
```diff
@@ -782,8 +806,11 @@ def view(self, indent=0):
...
+        self.stop_expr = self._upper_bound(parent=self)
+        # Update the loop limits for the loop
+        self.children[0] = self.start_expr
```
rupertford: sn't self.start_expr the same as self.children[0]? Similarly for self.stop_expr? If so, you don't need these lines.

src/psyclone/gocean1p0.py
```diff
@@ -792,13 +819,29 @@ def __str__(self):
 
         # Generate the upper and lower loop bounds
         # TODO: Issue 440. upper/lower_bound should generate PSyIR
-        self.start_expr = Literal(self._lower_bound(), parent=self)
-        self.stop_expr = Literal(self._upper_bound(), parent=self)
+        self.start_expr = self._lower_bound(parent=self)
+        self.stop_expr = self._upper_bound(parent=self)
+        # Update the stored loop limits for the loop
+        self.children[0] = self.start_expr
```
rupertford: sn't self.start_expr the same as self.children[0]? Similarly for self.stop_expr? If so, you don't need these lines.

src/psyclone/gocean1p0.py
```diff
@@ -833,8 +876,11 @@ def gen_code(self, parent):
 
         # Generate the upper and lower loop bounds
         # TODO: Issue 440. upper/lower_bound should generate PSyIR
-        self.start_expr = Literal(self._lower_bound(), parent=self)
-        self.stop_expr = Literal(self._upper_bound(), parent=self)
+        self.start_expr = self._lower_bound(parent=self)
+        self.stop_expr = self._upper_bound(parent=self)
+        # Update the loop limits for the parent loop
+        self.children[0] = self.start_expr
```
rupertford: sn't self.start_expr the same as self.children[0]? Similarly for self.stop_expr? If so, you don't need these lines.

src/psyclone/gocean1p0.py
```diff
@@ -833,8 +876,11 @@ def gen_code(self, parent):
 
         # Generate the upper and lower loop bounds
         # TODO: Issue 440. upper/lower_bound should generate PSyIR
-        self.start_expr = Literal(self._lower_bound(), parent=self)
-        self.stop_expr = Literal(self._upper_bound(), parent=self)
+        self.start_expr = self._lower_bound(parent=self)
```
rupertford: As mentioned earlier, should the above TODO be removed, seeing as you now do generate PSyIR?

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -1073,6 +1077,23 @@ def test_goloop_unmatched_offsets():
         in str(excinfo.value)
 
 
+def test_goloop_bounds_invalid_iteration_space():
+    ''' Check that the _upper/lower_bound() methods raises the expected error
```
rupertford: raises -> raise?

rupertford:
Back to @arporter

src/psyclone/gocean1p0.py
arporter: You're so picky. As it happens, I looked a bit more closely at what `parent` is used for and realised that we can remove it altogether since it is actually always `self`.

src/psyclone/gocean1p0.py
arporter: Yes, good point.

src/psyclone/gocean0p1.py
arporter: These lines are here because our bounds might have changed and therefore we need to correct the appropriate child nodes.

src/psyclone/gocean1p0.py
arporter: As above.

src/psyclone/gocean1p0.py
arporter: As above.

src/psyclone/gocean1p0.py
arporter: As above.

src/psyclone/gocean1p0.py
arporter: Done.

src/psyclone/tests/gocean1p0_test.py
arporter: Done.

arporter:
If Travis is happy then this is ready for review...

src/psyclone/gocean0p1.py
rupertford: As discussed, these are the same thing. I'll remove them rather than requesting a change.

Issue #512, opened 2019-09-17
[PSyIR] Support OMP directives for NEMO


hiker:
I never opened a ticket for what I am working on ... and I only new realised that I was supposed to put the ticket reference in my commits - I admit I didn't know, but will do so now.

@arporter , @sergisiso I just have one question: I run into a few PSyIR issues (i.e. certain structures are not yet supported). I am working around those issues, but need to add # TODO. My question: is there a general 'finish PSyIR' ticket, or should I open a new generic ticket (we could add a todo list in that ticket), or should I open a ticket for each and every node type not working?

In case that you want to have a look: https://github.com/stfc/PSyclone/tree/psyir-support-directives. This is what I do in the directive visitor to handle missing nodes:
```
           try:
                result_list.append(self._visit(child))
            except VisitorError:
                # TODO: Ignore unsupported nodes for now, to allow all
                # tests to work.
                result_list.append("Node type {0} not yet supported\n"
                                   .format(type(child)))  
```

sergisiso:
I am a bit confused about why do you need this try/except instead of just `self._visit(child)`. Wouldn't unsupported features be in Codeblocks. So the Fortran back-end may still generate them without returning a VisitorError and the C back-end should just fail anyway.

If it is just temporal code to make things work before the full PR is finished I don't think the # TODO needs to reference an issue if it is not going into master.

hiker:
I need this in order to test gocean, to see if the visitor pattern works with that API as well, which has unknown pattern. I also have implemented the directives for C, which I want to test, but it has as well missing nodes. For example, the visitor would return:
```
$omp parallel private(i,j)
Node type <class 'psyclone.gocean1p0.GOLoop'> not yet supported - Visitor Error: Unsupported node 'GOKern' found: method names attempted were ['gokern_node', 'codedkern_node', 'kern_node', 'node_node'].
!$omp end parallel
```
(though renaming 'nemokern_node' to 'kern_node' seems to fix that problem :) I still need to verify that )

So it's more of a problem with C:
```
#pragma omp parallel private(i,j)
{
Node type <class 'psyclone.gocean1p0.GOLoop'> not yet supported - Visitor Error: Unsupported node 'GOLoop' found: method names attempted were ['goloop_node', 'loop_node', 'node_node'].
}
```
And while I am tempted to just implement a few additional basic C functions, this is not exactly high priority for me, and I might be interfering with what you (and others) are doing, so I won't :)

[...]

Issue #514, opened 2019-09-19
[PSyIR] Add support for reductions and 'omp parallel do'


hiker:
This extends #512. Reductions and 'omp parallel do' are not needed for my ROMS work, but it would complete the omp directive support in PSyIR, so I opened a separate issue for this.

arporter:
At some point since this was opened, we have added support for `OMPParallelDoDirective`. However, support for reductions is still required. @jwallwork23 says:

In the latest version of our PSyclone-driven OpenACC GPU port, reduction loops get loop seq. Performance is improved significantly if these are manually edited to instead use loop independent reduction(...). Would it be possible to support reduction clauses in PSyclone?

arporter:
Alan Hewitt at the Met Office seems to have taken up the UKCA torch and is also asking about OpenACC reductions. Could be a good one for @hiker or @LonelyCat124 to take on perhaps?

arporter:
Could be a duplicate of #1812 and #2446?

LonelyCat124:
I think not quite duplicate of #1812 is more an issue on how we declare reductions (as clause objects instead of strings which current reductions are).

This issue is in theory discussing OMPParallelDoDirectives (i think? Its not totally clear) vs #2446 focusing GPU directives - I think what we need for those directives is the same so its close enough to duplicate (though 2446 extends the behaviour to support atomics).

PR #515, opened 2019-09-19
Psyir support directives

hiker:
This pull requests adds support for 'omp parallel' and 'omp do' directives (see #512), with private variable declarations. This is the required for ROMS-psyclone (which does not need/support reductions, and 'omp parallel do'). It also adds support for Loop nodes with the C visitor.

Note that I refactored some code so that C and Fortran visitor can share code that is in the base class, instead of a some code duplication between the classes.

I already had to fight several conflicts, so I would prefer to get this merged in before I add 'omp parallel do', and support for reductions (which I track under #514). And ... I of course forgot to add the issue number in my commits. I'll try to improve this :)

This needs (or includes) #511.


src/psyclone/psyGen.py
```diff
@@ -1297,12 +1297,10 @@ def list(self, indent=0):
...
-        result = ""
-        for idx, value in enumerate(my_list):
-            result += str(value)
-            if idx < (len(my_list) - 1):
-                result += ","
-        return result
+        ''':returns: a comma-delimited string of all members.
+        :rtype: str
+        '''
+        return ",".join(my_list)
```
arporter: Gosh, that must have been an old routine - we know a lot more Python now! Please document the `my_list` argument to the function.

src/psyclone/psyGen.py
```diff
@@ -2433,6 +2431,39 @@ def gen_code(self, parent):
             for call in reprod_red_call_list:
                 call.reduction_sum_loop(parent)
 
+    def begin_string(self):
+        '''Returns the beginning statement of this directive, i.e.
+        "omp parallel". The visitor is responsible for adding the
+        correct directive begin.
+
+        :returns: the beginning statement oforthis directive.
```
arporter: maybe s/beginning/opening/ and s/oforthis/of this/.

src/psyclone/psyGen.py
```diff
@@ -2433,6 +2431,39 @@ def gen_code(self, parent):
...
+        return result
+
+    def end_string(self):
+        '''Returns the end (or closing) statement of this directive, i.e.
+        "omp end parallel". The visitor is responsible for adding the
+        correct directive begin.
+
+        :returns: the end statement for this directive.
+        :rtype: str
+        '''
```
arporter: Blank line inside docstring rather than after it please.

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
+        directive contains the opening "}", or "!$" in Fortran. The
+        string {0} will be replaced with the end directive content.
+        It must be implemented by any visitor.
+        :return: the end of a directive.
+        :rtype: str
+        '''
+
+    def ompdirective_node(self, node):
+        '''This method is called when an OMPDirective instance is found in
+        the PSyIR tree. It returns the begin and end directive, and the
```
arporter: s/begin and end directive/opening and closing directives/

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
+        statements in between as a string (depending on the language).
+
+        :param node: a Directive PSyIR node.
+        :type node: :py:class:`psyclone.psyGen.Directive`
+
+        :returns: the Fortran code as a string.
+        :rtype: str
+        '''
+
+        result_list = [self.directive_start.format(node.begin_string())]
```
arporter: Despite all your feelings against properties, I would have said that `node.begin_string` and `node.end_string` are really cases where they make sense...

src/psyclone/psyir/backend/c.py
```diff
@@ -408,4 +413,40 @@ def codeblock_node(self, node):
         :raises VisitorError: The CodeBlock can not be translated to C.
 
         '''
+        # pylint: disable=no-self-use,unused-argument
```
arporter: So `node` should be replaced with `_`...

src/psyclone/psyir/backend/c.py
```diff
@@ -408,4 +413,40 @@ def codeblock_node(self, node):
...
+    def directive_start(self):
+        ''':return: "#pragma" - the start of a directive in C.
+        :rtype: str
+        '''
+        # Note that {{ is replaced with a single { in the format call
+        return "#pragma {0}\n{{\n"
+
+    @property
+    def directive_end(self):
+        ''':return: "#pragma" - the start of a directive in C.
```
arporter: Cut-n-paste error in docstring.

src/psyclone/psyir/backend/c.py
```diff
@@ -408,4 +413,40 @@ def codeblock_node(self, node):
...
+        1: loop variable
+        2: first loop iteration
+        3: last loop iteration
+        4: step size
+        5: body of the loop
+
+        :return: the format of a loop.
+        :rtype: str
+        '''
+        return "{0}for({1}={2}-1; {1}<{3}; {1}+={4})\n"\
```
arporter: I don't think there should be a `-1` in here, or, at least, we should document that this founction returns the first loop iteration minus one. I don't know enough about the C backend but maybe @sergisiso can comment - what's our strategy in generating C when thinking about the fact that array accesses are zero-indexed (and, consequently, loop bounds are adjusted c.f. those in Fortran)?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -572,3 +562,17 @@ def nemoimplicitloop_node(self, node):
...
+    @property
+    def directive_start(self):
+        ''':returns: "!${0}" - the start of a directive in Fortran
+        :rtype: str
+        '''
+        return "!${0}\n"
+
+    @property
+    def directive_end(self):
+        ''':returns: "!${0}" - the end of a directive in Fortran
```
arporter: Probably lost in translation but obviously "!$" is not the end of a directive. Maybe using "closing directive" instead?

src/psyclone/tests/psyclone_test_utils.py
```diff
@@ -36,18 +36,24 @@
...
 # The various file suffixes we recognise as being Fortran
 FORTRAN_SUFFIXES = ["f90", "F90", "x90"]
 
 
 class CompileError(Exception):
     '''
     Exception raised when compilation of a Fortran source file
-    fails.
 
+    fails.
```
arporter: Strange line break?

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
```diff
@@ -0,0 +1,239 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author J. Henrichs, Bureau of Meteorology
+# -----------------------------------------------------------------------------
+
+'''Performs pytest tests on the psyclond.psyir.backend.fortran module'''
```
arporter: s/psyclond/psyclone/ and it also seems to be tests of both the Fortran and C support for the OpenMP directives?

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
```diff
@@ -0,0 +1,239 @@
...
+        end subroutine tmp
+        end module test'''
+    schedule = create_schedule(code)
+    from psyclone.transformations import OMPParallelTrans
+
+    # Now apply a parallel transform
+    omp_par = OMPParallelTrans()
+    omp_par.apply(schedule[0])
+
+    fvisitor = FortranWriter()
```
arporter: Strictly speaking, if it's just the Writers that you want to test you could construct the PSyIR manually but I'll let it pass.

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
```diff
@@ -0,0 +1,239 @@
...
+    :type node: :py:class:`psyclone.psyGen.Node`
+    '''
+
+    # Create a simple 'a=b' assignment statement for all tests
+    assignment = Assignment(parent=node)
+    lhs = Reference('a', assignment)
+    rhs = Reference('b', assignment)
+    assignment.addchild(lhs)
+    assignment.addchild(rhs)
+    del node.children[0]
```
arporter: It might be more robust to simply replace `node.children[0]` with the new `assignment` object - if the `parent` has more than one child then this will produce consistent results.

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
```diff
@@ -0,0 +1,239 @@
...
+    a = (2 * i);
+    b[i] = (b[i] + a);
+  }
+}'''
+    assert correct in result
+
+
+# ----------------------------------------------------------------------------
+def test_gocean_omp_do():
+    '''Test that an OMP PARALLEL directive in a 'classical' API (gocean here)
```
arporter: s/PARALLEL/DO/

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
```diff
@@ -0,0 +1,239 @@
...
+    from psyclone.tests.psyclone_test_utils import get_invoke
+
+    _, invoke = get_invoke("single_invoke.f90", "gocean1.0",
+                           idx=0)
+    omp = OMPLoopTrans()
+    omp_sched, _ = omp.apply(invoke.schedule[0])
+
+    # Now remove the GOKern (since it's not yet supported in the
+    # visitor pattern) and replace it with a simple assignment.
+    # While this is invalid usage of OMP (omp must have a loop,
```
arporter: Unfinished comment?

sergisiso:
> I am slightly concerned about the "implicit" transformation (-1) of the loop bounds in the C writer and the potential implications for array accesses. I may be worrying over nothing here and @sergisiso may have this in hand?

Tbh, I haven't though too much about this.

Until now I didn't use any implicit transformation in the indexes. Note that gocean does not have loops in the kernels and the iteration variables are provided by the opencl get_global_ids. These are inferred from the invocation of the OpenCL kernel which in turn uses the c_iso_bindings to make the call. I am not sure if the c_iso_bindings or FortCL make any indexing conversion?

I am also worried about not translating the PSyIR exactly as found as this could lead to lots of implicit transformations (e.g. if conditions that have the iteration index as element should also be corrected?) so it could be pretty hard to be consistent. Although not using 'implicit' -1 may mean we don't use the 0 row/column in C and use a bit more space. My first impression is that is better to not use implicit conversions at all, but probably we should think more carefully what is the best solution.

setup.py
```diff
@@ -42,7 +42,11 @@
 
 BASE_PATH = os.path.dirname(os.path.abspath(__file__))
 SRC_PATH = os.path.join(BASE_PATH, "src")
-PACKAGES = find_packages(where=SRC_PATH)
+PACKAGES = find_packages(where=SRC_PATH,
```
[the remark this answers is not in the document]
hiker: Complicated question ... I first moved the one ... create_schedule or so function into utilities, before I noticed the side effects (tests were installed etc), at which stage I created a separate branch just for that and created #511, then moved this to be based on #511 (since it already had code to move that function). A bit messy, but most of the work will be in merging, which I will handle - that's what I meant with "I had some merge conflicts", basically changes to the function that I moved into psyclone_utilities (which has now been renamed *sigh* merging will be a pain). Still, we should get #511 in first.

src/psyclone/psyGen.py
hiker: I forgot to ask this: I actually wanted to suggest to remove this function, and replace the (iirc) three calls to it with just ",".join() - is that ok?

src/psyclone/psyGen.py
```diff
@@ -2433,6 +2431,39 @@ def gen_code(self, parent):
...
+        "omp parallel". The visitor is responsible for adding the
+        correct directive begin.
+
+        :returns: the beginning statement oforthis directive.
+        :rtype: str
+        '''
+
+        # pylint: disable=no-self-use
+        result = "omp parallel"
+        # TODO: not yet working with NEMO
```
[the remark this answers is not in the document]
hiker: Hmm - I think that was why I opened #512 - but I'll either use #512, or create a new ticket then.

src/psyclone/psyir/backend/base.py
hiker: I have to admit it originally was, till I realised that (at least) the ``omp parallel (do)`` directives do a significant amount of computation (determining private variable, in the future handling reductions), so according to pep8 (https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance: "Note 3: Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.") this **must** be a function, not a property.

I obviously try to appear important and following python standards and quoting stuff as proof, highlighting stuff in bold - willfully ignoring that "expensive" is somewhat subjective, and just trying to confuse you into agreeing ;) So you let me know, if you really want to contradict the official python style guide and convert this into a property :grin: Though I seriously think it should be a function, and then for consistency's sake end_string should be, too.

src/psyclone/psyir/backend/c.py
hiker: I can't really agree with that one. It's the base class, and even though it does not use it, it's abstract and and classes inheriting should be able to use this as a template - it's just better documentation. So I really prefer to disable the pylint warnings (or do you want to use _ instead of self as well :grinning: ? We obviously need self in order to overwrite that function).

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
hiker: Done, though ideally I would say that addchild should take care of existing parent relationship and adjust them as well - so I would prefer addchild. But since addchild at this stage does not to anything else, I've changed it.

hiker:
> I am also worried about not translating the PSyIR exactly as found as this could lead to lots of implicit transformations (e.g. if conditions that have the iteration index as element should also be corrected?) so it could be pretty hard to be consistent. Although not using 'implicit' -1 may mean we 

Yes, you are of course correct. Unless we adjust any usage of the loop variables outside of indices with (index+1) - which sounds just stupid, we should indeed not adjust the loop bounds. I've removed this.

> don't use the 0 row/column in C and use a bit more space. My first impression is that is better to not use implicit conversions at all, but probably we should think more carefully what is the best solution.

but then I don't agree with that: typically the data allocation is done in Fortran, so we can't just simple ignore a row/column or so. Imho when using a loop variable as an index, we need to use 'variable -1', e.g.:
```
            # (e.g. A[3,5,2] -> A[3 * ALEN2 * ALEN1 + 5 * ALEN1 + 2])
```
should become:

```
            # (e.g. A[3,5,2] -> A[(3-1) * ALEN2 * ALEN1 + (5-1) * ALEN1 + (2-1)])
```
likely in array_node:
```
code = code + "("+ self._visit(child)+"-1)"      # Convert index 1 to n to be in (0 to n-1)
```
But indeed this is not an issue we need to solve right now. I only added simple loop support since it makes for very nice, symmetric testing.  My suggestion would be to add a #TODO to sort out loop index usage in c later.

sergisiso:
>  typically the data allocation is done in Fortran, so we can't just simple ignore a row/column or so.

@hiker That's true.

I think the problem is that we have not designed the C back-end yet. What we have is just a byproduct of the OpenCL back-end that I generalized because I though it would be useful in the future. I wouldn't worry about this until somebody takes the task of 'implementing a C backend for one of the APIs'. I am OK with the TODO warnings for now.

src/psyclone/psyGen.py
arporter: Yes, I nearly suggested that myself.

setup.py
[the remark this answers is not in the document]
arporter: OK :-)

src/psyclone/psyGen.py

src/psyclone/psyGen.py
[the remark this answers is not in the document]
hiker: OK, #514 it is - done (and I closed #512).

src/psyclone/psyGen.py

src/psyclone/psyGen.py
```diff
@@ -2662,6 +2693,28 @@ def gen_code(self, parent):
...
+        :returns: the beginning statement for this directive.
+        :rtype: str
+        '''
+
+        return "omp do schedule({0})".format(self._omp_schedule)
+
+    def end_string(self):
+        '''Returns the end (or closing) statement of this directive, i.e.
+        "omp end do". The visitor is responsible for adding the
+        correct directive begin.
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -2662,6 +2693,28 @@ def gen_code(self, parent):
...
                    position=["after", position])
 
+    def begin_string(self):
+        '''Returns the beginning statement of this directive, i.e.
+        "omp do ...". The visitor is responsible for adding the
+        correct directive begin.
+
+        :returns: the beginning statement for this directive.
+        :rtype: str
+        '''
```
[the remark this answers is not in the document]

src/psyclone/psyGen.py
```diff
@@ -2662,6 +2693,28 @@ def gen_code(self, parent):
         parent.add(DirectiveGen(parent, "omp", "end", "do", ""),
                    position=["after", position])
 
+    def begin_string(self):
+        '''Returns the beginning statement of this directive, i.e.
+        "omp do ...". The visitor is responsible for adding the
+        correct directive begin.
```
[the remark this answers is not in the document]

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
                 "{1}.".format(type(node).__name__, str(possible_method_names)))
+
+    def loop_node(self, node):
+        '''This method is called when a Loop instance is found in the
+        PSyIR tree.
+
+        :param node: a Loop PSyIR node.
+        :type node: :py:class:`psyclone.psyGen.Loop`
+
+        :returns: the Fortran code as a string.
```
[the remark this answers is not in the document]
hiker: Yes, indeed.

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
+        '''Returns the format for a do/for loop. The format variables
+        will be replaced as follows:
+        0: indentation string
+        1: loop variable
+        2: first loop iteration
+        3: last loop iteration
+        4: step size
+        5: body of the loop
+        This is used by base.loop_node() to create a loop representation.
+        :return: the format of a loop.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
+        :rtype: str
+        '''
+
+    @abc.abstractproperty
+    def directive_start(self):
+        '''Returns the start of a directive, e.g. "#pragma ...\n{" in C,
+        or "!$" in Fortran. The string {0} in the result will be replaced
+        with the actual directive. It must be implemented by any visitor.
+        :return: the start of a directive.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
+        :rtype: str
+        '''
+
+    @abc.abstractproperty
+    def directive_end(self):
+        '''Returns the end of a directive, e.g. "}" in C (the start
+        directive contains the opening "}", or "!$" in Fortran. The
```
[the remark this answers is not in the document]
hiker: Done (and the second '}' should be a '{')

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,82 @@ def _visit(self, node):
...
+        string {0} will be replaced with the end directive content.
+        It must be implemented by any visitor.
+        :return: the end of a directive.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/base.py
hiker: Done.

src/psyclone/psyir/backend/c.py
hiker: Oops.

src/psyclone/psyir/backend/c.py
hiker: See discussion below.

src/psyclone/psyir/backend/fortran.py
hiker: Done.

src/psyclone/tests/psyclone_test_utils.py
hiker: Indeed.

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
hiker: Done.

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
hiker: For a full test I would need to create not only the directives, but also something in between (to test that children are converted). I just found it easier to just parse it :)

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
hiker: Oops

src/psyclone/psyGen.py

hiker:
I decided to add the 'subtract 1' t the code and adjusted the tests under psyir, but then had failures in main tests (gocean1p0_opencl test). Not knowing that code too well, I am not sure what to do - we can discuss in our next telco.

One thing I did notice when looking at examples/gocean/eg3 was a change of the variable used to get the the dimension: on CPU it would use cu_fld, with opencl it appears to be using p_fld - and also the loop boundaries are likely wrong - it is just using nx/ny from the field, not start/stop - so something is not quit right :)

As stated before, am happy to postpone this (using #523), I just need a decision if my latest commit (that subtracts '1' from array indices) should be reverted or not. Stand-alone it makes sense to have it in, but in case of OpenCL is is very likely wrong (opencl seems to use i=0; i<n as indices ... meaning we don't need to subtract 1 - but it likely also means that using the index variable outside an index (e.g. comparison or assignment) - I would strongly suspect that this would not work as expected.

hiker:
As discussed I've reverted the 'use -1 in C' commit (which is put on https://github.com/stfc/PSyclone/tree/subtract_one_from_c_arrays_indices for #523). 

Ready for next review.

src/psyclone/psyGen.py
arporter: Thanks but you've got "$!" instead of "!$".

src/psyclone/transformations.py
```diff
@@ -1634,7 +1634,8 @@ class OMPParallelTrans(ParallelRegionTrans):
     # The types of node that this transformation can enclose
     valid_node_types = (psyGen.Loop, psyGen.Kern, psyGen.BuiltIn,
                         psyGen.OMPDirective, psyGen.GlobalSum,
-                        psyGen.Literal, psyGen.Reference)
+                        psyGen.Literal, psyGen.Reference,
+                        psyGen.Assignment, psyGen.BinaryOperation)
```
arporter: Why have Assignment and BinaryOperation been added here but the rest of the file is unchanged?

src/psyclone/psyir/backend/c.py
hiker: Indeed if I add an abstract method to base.py, we don't get the pylint warnings, see e2282c6bdca27a4d4c97f7d319670c3d073328c8 where I removed a few #pragmas by adding abstract method (oh, and I moved one identical function from c.py and fortran.py to base). The only question is: do we then want to:

1. add abstract methods for every method that is implemented in a backend language class? 
2. Or only those that trigger warnings? 
3. Or for none, and we just accept a few directives to disable pylint warnings?

That might need a vote. First one is probably the cleanest one, but quite a bit of additional typing work ;) The last one is what we have so far. And 2. is ... an odd mix :)

src/psyclone/psyGen.py
hiker: Oops - we really need to support compilation tests.

src/psyclone/tests/psyir/backend/psyir_openmp_test.py
hiker: Done ("# visitor" was just a left-over when I reformatted the line breaks).

src/psyclone/psyGen.py
```diff
@@ -2478,6 +2470,38 @@ def gen_code(self, parent):
...
+        private_str = ",".join(private_list)
+
+        if private_str:
+            result = "{0} private({1})".format(result, private_str)
+        return result
+
+    def end_string(self):
+        '''Returns the end (or closing) statement of this directive, i.e.
+        "omp end parallel". The visitor is responsible for adding the
+        correct directive end (e.g. "!$").
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/psyGen.py
```diff
@@ -2707,6 +2731,28 @@ def gen_code(self, parent):
...
+        :returns: the beginning statement for this directive.
+        :rtype: str
+
+        '''
+        return "omp do schedule({0})".format(self._omp_schedule)
+
+    def end_string(self):
+        '''Returns the end (or closing) statement of this directive, i.e.
+        "omp end do". The visitor is responsible for adding the
+        correct directive beginning (e.g. "$!).
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +177,86 @@ def _visit(self, node):
...
+
+        :return: the start of a directive.
+        :rtype: str
+        '''
+
+    @abc.abstractproperty
+    def directive_end(self):
+        '''Returns the begining of the end of a directive, e.g. "}" in C (the
+        start directive also contains the opening "{"), or "!$" in Fortran (to
+        which e.g. "omp end do" will then be added). The string {0} will
```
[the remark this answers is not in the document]
hiker: No, it _can_. In Fortran we need it ("!$omp end do"), while in C we don't need it (it's just "}"). I've changed the description somewhat to make it easier to understand that.

src/psyclone/tests/psyir/backend/c_test.py
```diff
@@ -306,15 +309,21 @@ def test_cw_unaryoperator():
...
 
     # Test that an unsupported operator raises an error
     # pylint: disable=abstract-method, too-few-public-methods
     class Unsupported():
         '''Dummy class'''
-    # pylint: enable=abstract-method, too-few-public-methods
+        def __init__(self):
+            pass
+    # pylint: enable=too-few-public-methods
+    # pylint: disable=protected-access
```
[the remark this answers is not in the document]
hiker: Difficult question that, it had me puzzled for a while. Problem: I was sometimes running pylint directly from a subdirectory of tests, not tests itself. Then it would not pick up the .pylintrc file in tests, and give me those warnings again. But when we turned tests into a module (by adding some __init__.py files), then pylint will search the directory tree and find tests/.pylintrc .

So I've added some __init__.py files in tests subdirectories to make sure wherever we run pylint from, it will find tests/.pylintrc. I then also removed all unnecessary pylint directives in all tests.

You've added a .pylintrc file to nemo, which contained just "../.pylintrc". This seems to be invalid, both python2 and 3 pylints would just crash (when running in the nemo subdir). I've removed that file and also added __init__.py so that it will pick up tests/.pylintrc.

src/psyclone/transformations.py
hiker: I needed this for the test case to work - atm in psyir we can not handle a GOKern, so on Sergi's advice I've replaced this with an assignment (which psyir can already handle). There is a #TODO in place to track this.

hiker:
Ready for next review, or a vote about adding abstract methods in base.py (see https://github.com/stfc/PSyclone/pull/515#discussion_r334447827 for details).

src/psyclone/transformations.py
arporter: Oh I see - I was being dense :-)

src/psyclone/psyir/backend/c.py
arporter: We've just discussed this in the office and we disagree with your initial comment ;-) This is *not* the base class (it's the C backend) and we've provided this concrete method in order to produce a more tailored error message. We therefore all agree that using `_` for an un-used argument is the right thing to do in this case.

src/psyclone/psyir/backend/base.py
```diff
@@ -99,6 +102,52 @@ def __init__(self, skip_nodes=False, indent_string="  ",
         self._indent = indent_string
         self._depth = initial_indent_depth
 
+    @abc.abstractmethod
+    def codeblock_node(self, node):
```
arporter: I've chatted with @sergisiso and @rupertford about this and now understand things a bit better. By default the visitor will raise an Exception if a handler is not found. Therefore (contrary to what I said last time) we don't need these abstract methods as they aren't going to do anything for us.

src/psyclone/psyir/backend/base.py
```diff
@@ -174,3 +223,86 @@ def _visit(self, node):
...
+        '''Returns the start of a directive, e.g. "#pragma ...\n{" in C,
+        or "!$" in Fortran. The string {0} in the result will be replaced
+        with the actual directive. It must be implemented by any visitor.
+
+        :return: the start of a directive.
+        :rtype: str
+        '''
+
+    @abc.abstractproperty
+    def directive_end(self):
```
arporter: Again, I don't think a pure, abstract handler wins us anything (it will just change the error produced by the visitor pattern).

src/psyclone/psyir/backend/base.py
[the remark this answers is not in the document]
arporter: Thanks.

src/psyclone/tests/psyir/backend/c_test.py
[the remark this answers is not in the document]
arporter: Ah, that makes sense.

setup.py
```diff
@@ -45,7 +45,11 @@
 PACKAGES = find_packages(where=SRC_PATH,
                          exclude=["psyclone.tests",
                                   "psyclone.tests.test_files",
+                                  "psyclone.tests.core",
```
arporter: Can you do "psyclone.tests.*"?

src/psyclone/tests/psyir/backend/c_test.py
```diff
@@ -310,10 +313,9 @@ def test_cw_unaryoperator():
...
 
     # Test that an unsupported operator raises an error
-    # pylint: disable=abstract-method, too-few-public-methods
-    class Unsupported():
-        '''Dummy class'''
-    # pylint: enable=abstract-method, too-few-public-methods
+    # pylint: disable=too-few-public-methods,missing-docstring
+    class Unsupported(object):
+        pass
     unary_operation._operator = Unsupported
```
arporter: Do you need to re-enable 'missing-docstring' and 'too-few-public-methods' here? In fact, I think the latter one could be put in the `tests/.pylintrc` file.

src/psyclone/tests/psyir/backend/c_test.py
```diff
@@ -364,17 +366,49 @@ def test_cw_binaryoperator():
...
     binary_operation._operator = Unsupported
     with pytest.raises(VisitorError) as err:
         _ = cwriter(binary_operation)
     assert "The C backend does not support the '" in str(err)
     assert "' operator." in str(err)
 
 
+def test_cw_loop():
+    '''Tests writing out a Loop node in C. It parses Fortran code
+    and outputs it as C.'''
```
arporter: Please add a note that this is currently a literal translation and that, as a consequence, the resulting C code is not functionally equivalent to the Fortran. Please reference the new Issue you created for this.

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -49,8 +49,8 @@
 from psyclone.psyir.frontend.fparser2 import Fparser2Reader
 
 
-@pytest.fixture(scope="function")
-def fort_writer():
+@pytest.fixture(scope="function", name="fort_writer")
+def fixture_fort_writer():
```
arporter: That's nicer - thanks.

src/psyclone/psyir/backend/base.py
hiker: The directive_start/_end methods are not called using the visitor patterns, they are not handlers, but mainly helper functions, called from base. So I think there would be some benefit to of getting a pylint warning if those are not implemented. Though of course if we would envision a backend that does not support openmp directives, then it would indeed not make sense to enforce having this function everywhere. But I don't care one way or another.

src/psyclone/psyir/backend/c.py
hiker: OK, replaced with _, and then also remove the documentation of the parameter. I assume that was the intention?

src/psyclone/psyir/backend/base.py
hiker: I'm happy with this. I wasn't looking forward to add all those abstract methods 😁 So I've removed those (but left the duplicated code from c/Fortran moved into base)

setup.py
hiker: Ohh - great, yes, you can!!

src/psyclone/tests/psyir/backend/c_test.py
hiker: I did indeed put them in the wrong location. I moved them into the class, which then automatically re-enables them once the class-scope is left.

src/psyclone/tests/psyir/backend/c_test.py
hiker: Done.

src/psyclone/tests/psyir/backend/fortran_test.py
hiker: Yes, I was glad when I learned that this can be done - much better than dozens of pylint directives.

hiker:
> Looking good now apart from my change of heart (/greater understanding) about abstract handlers in the backend. As I've said inline, I don't think it makes sense to have abstract handlers as essentially that's just re-implementing the "this node is not supported" error that the backend already supplies.

Done - just let me know if you want to remove the abstract directive_start/end methods from base (I would argue they should be there, since they are called from base, not from c/Fortran or using the visitor pattern - but tbh, I don't care much about that).

> @rupertford did ask whether you've used the visitor in the `Directive.gen_code` method but I've had a look and don't think that that's possible - the visitor produces text while `gen_code` produces an f2pygen AST.

My understanding is that all that gen_code will be replaced with a visitor once we can? So I've on purpose copied code from `gen_code()` into `begin_string()`, instead of trying to refactor both functions to avoid code duplication (not sure if there would be much to refactor, it's mostly some if statements and call to functions that do the actual work).

> Thanks for tidying up all of the pylint stuff in the tests.
No worries ... I like tidy 👼

hiker:
Arg, mis-clicked 😡 

Ready for next review (or tell me if you want me to remove the abstract methods).

src/psyclone/psyir/backend/base.py
arporter: Oh yes, I missed that they weren't handlers - that's OK then.

PR #517, opened 2019-09-23
sir if statements 513

rupertford:


rupertford:
Subject to travis and codecov being happy, this is now ready for first review.

examples/nemo/eg4/if_example.f90
```diff
@@ -0,0 +1,64 @@
...
+
+  integer, parameter :: jpi=10, jpj=10, jpk=10
+  real, dimension(jpi,jpj,jpk) :: zwx, zslpx, tmpx
+  !real :: tmpx
+  integer :: ji, jj, jk
+  
+  DO jk = 1, jpk-1
+     DO jj = 2, jpj
+        DO ji = 2, jpi
+           !tmpx = zwx(ji,jj,jk) * zwx(ji-1,jj,jk)
```
arporter: Is there a reason for keeping the original form of the code commented out? If not, I think it can be removed.

src/psyclone/psyir/backend/sir.py
```diff
@@ -409,3 +418,52 @@ def unaryoperation_node(self, node):
...
                 "".format(self._nindent, oper, result))
+
+    def ifblock_node(self, node):
+        '''This method is called when an IfBlock instance is found in
+        the PSyIR tree.
+
+        :param node: an IfBlock PSyIR node.
+        :type node: :py:class:`psyclone.psyGen.IfBlock`
+
+        :returns: the SIR code as a string.
```
arporter: s/as a string// and is the SIR code Python?

src/psyclone/psyir/backend/sir.py
```diff
@@ -409,3 +418,52 @@ def unaryoperation_node(self, node):
...
+            else_part = ("make_block_stmt([{0}])"
+                         "".format(else_statements.lstrip().rstrip(",\n")))
+        else:
+            else_part = "None"
+
+        return ("{0}make_if_stmt({1}, {2}, {3})\n"
+                "".format(self._nindent, cond_part, then_part, else_part))
+
+    def schedule_node(self, node):
+        '''This method is called when an Schedule instance is found in the
```
arporter: s/an/a/

src/psyclone/psyir/backend/sir.py
```diff
@@ -409,3 +418,52 @@ def unaryoperation_node(self, node):
...
+        of PSyIR nodes and is therefore found in places such as the
+        contents of the 'then' part of an 'if' statement and the
+        contents of the 'else' part of an 'if' statement. The schedule
+        has no content so simply calls its children and returns the
+        agregated result.
+
+        :param node: a Schedule PSyIR node.
+        :type node: :py:class:`psyclone.psyGen.Schedule`
+
+        :returns: the SIR code as a string.
```
arporter: s/as a string/ and is it Python?

src/psyclone/psyir/backend/sir.py
```diff
@@ -409,3 +418,52 @@ def unaryoperation_node(self, node):
...
+        :rtype: str
+
+        '''
+        result = ""
```
arporter: This feels like something that ought to happen in a base class or default visitor - is that possible here?

src/psyclone/tests/psyir/backend/sir_test.py
```diff
@@ -493,27 +493,55 @@ def test_sirwriter_assignment_node(parser, sir_writer):
         in result)
 
 
-# (1/3) Method binaryoperation_node
-def test_sirwriter_binaryoperation_node_1(parser, sir_writer):
+# (1/4) Method binaryoperation_node
+@pytest.mark.parametrize("oper", ["+", "-", "*", "/", "**"])
```
arporter: Nice!

examples/nemo/README.md
```diff
@@ -33,30 +33,35 @@
 # -----------------------------------------------------------------------------
 # Author A. R. Porter, STFC Daresbury Lab
```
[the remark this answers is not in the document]
rupertford: Oops. I've gone for html comments <-- xxx --> as markdown is aimed for the web. There is the bigger question of whether the licence should be in this file (or any other documentation) as BSD is aimed at code, but I'm happy to ignore that in this PR :-)

examples/nemo/eg4/README.md
```diff
@@ -1,3 +1,38 @@
+# -----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
rupertford: Again, I've used html comments.

examples/nemo/eg4/if_example.f90
rupertford: Yes there is a reason, but possibly not one that you will support :-). They are to remind me of the original form of the example. The next thing I am working on is supporting scalars. Once scalars are supported then this example will be changed back to use a scalar. I mention that this current code is temporary and that it will use a scalar in the future and reference the associated issue in the comments at the start of the example. Is that enough or would you prefer me to remove them?

src/psyclone/psyir/backend/sir.py
rupertford: Changed to "SIR Python code."

src/psyclone/psyir/backend/sir.py
rupertford: Well spotted.

src/psyclone/psyir/backend/sir.py
rupertford: As above, I've changed this to "the SIR Python code.". I've also changed it in the rest of the class (all of the methods in the class used this form).

src/psyclone/psyir/backend/sir.py
rupertford: The only mechanism built in to the base class is to set skip_nodes to true, but that would ignore *all* nodes that were not defined in sir.py and I don't want to do that as I want sir.py to raise an exception if it finds something it does not recognise.

I guess we could extend the base class so that you could register certain node names to recurse through and ignore any content, but I think that needs discussion and should probably be another issue if so.

Alternatively the base class could provide a method which skipped the relevant nodes and called its children. This method could then call the base class method. This should be a small change so I would not mind doing that as part of this PR as long as the approach was agreed by others as well.

Personally I don't mind the current null implementation, but I take your point that we would expect this to be a commonly occurring use case so should support its implementation.

src/psyclone/tests/psyir/backend/sir_test.py
rupertford: :-)

rupertford:
Back to @arporter for 2nd review subject to travis being happy.

rupertford:
I'm taking this back as I forgot to update the documentation (not for the first time :-)).

rupertford:
OK, ready for review.

src/psyclone/psyir/backend/sir.py
arporter: OK. I understand now and am happy with the way it is. If we start getting lots of loops over children then we can have a re-think but it's not a big deal.

examples/nemo/eg4/if_example.f90
arporter: No, that's enough - I'm feeling kind :-)

Issue #500, opened 2019-09-09
Use SymbolTable in the NEMO API (and validate WHERE constructs)


arporter:
In converting WHERE constructs into the PSyIR the FParser2ASTProcessor has to identify which variable references within a construct are arrays. So long as the NEMO coding convention is followed this can be done by looking for refs of the form `a(:,:,:)`. However, if a developer omits the colon notation (or we are tackling a code where this convention is not used), then we need to be able to distinguish array references from scalars.

A first step towards this will be to start using the SymbolTable with the NEMO API and rejecting any constructs for symbols which we cannot identify as scalars/arrays. This will mean making the SymbolTable a bit more configurable and allowing for symbols without explicit declarations (since NEMO has lots of USEs without ONLY clauses).

arporter:
The existing NEMO frontend creates an Invoke for each routine/program unit found in a source file, passing a pointer to the appropriate part of the parse tree each time. In contrast, the frontend used to create kernel schedules (`generate_schedule()`) expects to find a source file containing a single module which itself contains the named subroutine. It then creates a Container to represent the module and a KernelSchedule for the subroutine.

arporter:
Every NEMO source file that we process contains a MODULE. @hiker is this true of ROMS too?

arporter:
If we were to follow the same approach as done for Kernels then for each NEMO source file we would have a `Container` with one or more `InvokeSchedule`s as children. This seems sensible but would (at least initially) limit us to source files containing a single MODULE. It will probably also break a lot of tests in the short term.

arporter:
I've noticed that `Fparser2Reader.generate_schedule()` actually creates a parent `Container` but returns the child `KernelSchedule`. It is also Kernel specific and therefore really ought to be `generate_kernel_schedule()`.

sergisiso:
> I've noticed that Fparser2Reader.generate_schedule() actually creates a parent Container but returns the child KernelSchedule. It is also Kernel specific and therefore really ought to be generate_kernel_schedule().

I agree the name should be `generate_kernel_schedule`, also note that in #545 I added a second entry point called `generate_container()` that returns the whole module (useful for finding `use` global vars), it is still limited to just one module but if needed it can be easily extended (or create a new function) to return a list of containers.

arporter:
[...]

Issue #393, opened 2019-05-27
[LFRic] Fortran Kernel generation should capture module information


rupertford:
At the moment, LFRic Fortran Kernel generation only captures the kernel subroutine information, not the surround module and any declarations contained within.

rupertford:
There is an issue with doing this. The Kernel Node is currently used to generate a Kernel call for the PSy-layer back-end and we would want to use it to also generate a module for the Kernel back-end.

some options are

a) give the back-end some boolean flags to decide what to do

b) try to use the context i.e. is it the first node encountered or not

c) use a different Fortran back-end for PSy-layer and Kernel generation (we could subclass the Kernel back-end to make the PSy-layer back-end and overwrite the Kernel method?)

d) split a Kernel into two types of Node - a KernelCall and a Kernel.

c) and d) sound the most promising to me but as I'm not sure where we're going to go with the PSy-layer Nodes and generation in the future, I might be missing something. Any thoughts @sergisiso or @arporter.

TeranIvy:
And I would also like the Kernel Node to know about its arguments and their attributes (e.g. data type, kind, array shape) for generating Psyke driver :) Kernel Call doesn't help as it doesn't have full information that need to be written/read. That information is scattered in various classes (mainly DynInvoke and DynCollection) and it would be really useful to have it one place.

It seems to me that common elements for Kernel Generation, Kernel Call, Kernel Driver and Kernel Stub are a) kernel arguments and their attributes and b) module information. Kernel Call/Driver/Stub also use kernel module information (metadata) to derive argument list and their attributes required for declarations (PSy layer, kernel driver and kernel stub, respectively) and kernel calls (PSy layer and kernel driver, respectively). I presume names and attributes for Kernel Generation are obtained from directly from the parsed kernel code.
The common method to all four classes is argument declaration. Kernel Call will add PSy layer call, Kernel Driver will add write/read statements, Kernel Generation will add other statements from subroutine body, etc. Another common method is use of module information although the specific implementation will differ for each subclass (e.g. generation of USE statements, kernel arguments).

To summarise, I think it would be good to have a parent class with common data (arguments and their attributes) and common methods (argument declaration, use of kernel module information). Subclasses will implemented specifics and add other things they need.

TeranIvy:
[...]

PR #518, opened 2019-09-25
393 container node

rupertford:


rupertford:
The addition of a Container Node and a symbol table has caused a couple of problems

1) the front end assumed that the root node was the Schedule Node when testing whether a symbol exists in the symbol table. I've fixed that by making the symbol check recurse up through ancestors checking any symbol table it finds - this supports the new hierarchy of symbol tables.

The new idea of a hierarchy of symbol tables has caused a failing test. The test has a module variable defined which is meant to cause an error in a transformation (as module variables are not supported). However the new test now allows this. I need to talk with @arporter about what is needed here.

rupertford:
I think this PR is ready for 1st review. I've given it to @sergisiso as he  will probably have views on the way I've linked references and the symbol table.

src/psyclone/psyGen.py
```diff
@@ -6566,4 +6628,73 @@ def __str__(self):
         return "Return[]\n"
 
 
+class Container(Node):
+    '''Node representing something that contains one or more
+    KernelSchedule nodes and/or Container nodes, as well as a name and
+    a SymbolTable. This construct can be used to scope variables,
```
sergisiso: I would delete "share variables between KernelSchedules nodes", maybe use something like: "This construct can be used to scope symbols of variables and KernelSchedule names"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1448,18 +1459,33 @@ def _name_handler(self, node, parent):
...
         return Reference(node.string, parent)
 
+    def _check_declared(self, name, node):
+        '''
+        :param str: name of the symbol.
+        :param node: parent node of the PSyIR node we are constructing.
+        :type node: :py:class:`psyclone.psyGen.Node`
+
+        raises SymbolError: if there is one or more ancestor symbol \
+        table(s) and the name is not found in one of them.
```
sergisiso: I think this docstring does not match the implementation, it should be:" if there is no symbol table or if the name is not found in any of the symbol tables in scope"

rupertford:
> @rupertford I just marked some docstring issues I found.
> 
> I have not marked as issues yet but I also have doubts about the `symbol` and `_check_declared` methods. Although the implementations are good, I would like to discuss with you tomorrow where each logic belongs (also because they repeat similar logic):
> 
> * I think the 'recurse up until you find a symbol_table' could be in the same 'symbol_table' property.
> * Then the recursive lookups (with a limited scope if we want) I am not sure if it should be done by the same symbol_table or is it OK for the Reference class to implement them.
> 
> (All this can be part of a future PR if we focus this one to just the new Container class, but I think it is worth discussing where this is going)

I expected you to have views on the relationship between the symbol tables and the reference elements which is why I asked you to review it :-). As you suspect I just implemented something that worked as I want to get the container working. My preference would be to push ahead with the implementation and create a new issue but I am happy to discuss and modify if there is a clear way forward.

src/psyclone/psyGen.py
```diff
@@ -6235,6 +6237,66 @@ def reference_accesses(self, var_accesses):
         '''
         var_accesses.add_access(self._reference, AccessType.READ, self)
 
+    def symbol(self, scope=None):
```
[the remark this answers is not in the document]
rupertford: Complete.

src/psyclone/psyGen.py
```diff
@@ -6235,6 +6237,66 @@ def reference_accesses(self, var_accesses):
...
+        '''Returns the symbol from a symbol table associated with this
+        reference or None is one is not found. The scope variable
```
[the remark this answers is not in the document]
rupertford: Modified.

src/psyclone/psyGen.py
```diff
@@ -6235,6 +6237,66 @@ def reference_accesses(self, var_accesses):
...
+        limits the symbol table search to nodes within the scope.
+
+        :param scope: optional Node which limits the search for a \
+        symbol in a symbol table to nodes within the scope or no \
```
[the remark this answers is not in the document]
rupertford: Agreed.

src/psyclone/psyGen.py
```diff
@@ -6566,4 +6628,73 @@ def __str__(self):
         return "Return[]\n"
 
 
+class Container(Node):
+    '''Node representing something that contains one or more
+    KernelSchedule nodes and/or Container nodes, as well as a name and
```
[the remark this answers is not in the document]
rupertford: Removed.

src/psyclone/psyGen.py
rupertford: Done.

src/psyclone/psyir/frontend/fparser2.py
rupertford: The logic is that an exception is not raised if there are no symbol tables found. This is to support the case of the NEMO api which currently does not have symbol tables. I've made a TODO to point this out and have referenced the relevant issue (#500). I've also updated the wording to try to make it clearer.

src/psyclone/tests/kernel_transformation_test.py
```diff
@@ -483,6 +483,37 @@ def test_no_inline_global_var():
...
-    assert ("'kernel_with_global_code' contains accesses to data that are not "
-            "captured in the PSyIR Symbol Table (Undeclared reference 'alpha'"
-            " found" in str(err))
+    assert ("'kernel_with_global_code' contains accesses to data (variable "
+            "'alpha') that are not captured in the PSyIR Symbol Table(s) "
+            "within KernelSchedule scope." in str(err))
+
+# Class KernelTrans
+
+# Method validate
```
sergisiso: If you really need these 2 comments (the test function already has this information in the name), move the 2 linebreaks separation before the comments instead of after them.

src/psyclone/tests/psyGen_test.py
```diff
@@ -3173,6 +3173,64 @@ def test_reference_can_be_printed():
...
+
+    # Symbol in Container SymbolTable with KernelSchedule scope
+    assert alpha.symbol(scope_limit=kernel_schedule) is None
+
+    # Symbol in Container SymbolTable with Container scope
+    assert isinstance(kernel_schedule.root, Container)
+    assert alpha.symbol(scope_limit=kernel_schedule.root).name == alpha.name
+
+    # Symbol method with invalid scope type
+    with pytest.raises(GenerationError) as excinfo:
```
sergisiso: It feels that this one should be a TypeError (but the test below is rightly a GenerationError) I am happy to overlook this issue if we consider the function as internal and the type checking is not needed.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -77,63 +78,88 @@ def f2008_parser():
 end module dummy_mod
 '''
 
+# Class Fparser2Reader
```
sergisiso: Same. The double line-break before the comments would look more natural.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -42,14 +42,15 @@
 from psyclone.psyGen import PSyFactory, Node, Directive, Schedule, \
     CodeBlock, Assignment, Return, UnaryOperation, BinaryOperation, \
     NaryOperation, Literal, IfBlock, Reference, Array, KernelSchedule, \
-    Symbol, SymbolTable, \
+    Symbol, SymbolTable, Container, \
     InternalError, GenerationError
 from psyclone.psyir.frontend.fparser2 import Fparser2Reader
 from psyclone.dynamo0p3 import DynKern, DynKernMetadata
-
+from fparser.common.readfortran import FortranStringReader
```
sergisiso: Can you solve this 3 pylint issues:

- [unused-import] Unused api imported from fparser as fpapi
- [unused-import] Unused DynKern imported from psyclone.dynamo0p3
- [wrong-import-order] third party import "from fparser.common.readfortran import FortranStringReader" should be placed before "from psyclone.psyGen ..."

sergisiso: And FortranStringReader is re-declared inside some tests.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -44,7 +44,7 @@
 from fparser.two.utils import walk_ast
 from psyclone.psyGen import UnaryOperation, BinaryOperation, NaryOperation, \
     Schedule, Directive, CodeBlock, IfBlock, Reference, Literal, Loop, \
-    Symbol, KernelSchedule, \
+    Symbol, KernelSchedule, Container, \
     Assignment, Return, Array, InternalError, GenerationError, SymbolError
```
sergisiso: Pylint says SymbolError is unused now.

src/psyclone/tests/kernel_transformation_test.py
rupertford: Changed.

src/psyclone/tests/psyir/frontend/fparser2_test.py
rupertford: Modified.

src/psyclone/tests/psyir/frontend/fparser2_test.py
rupertford: FIxed. There were a lot of re-declarations to remove!

src/psyclone/psyir/frontend/fparser2.py
rupertford: And it is correct!

src/psyclone/tests/psyGen_test.py
rupertford: Updated.

rupertford:
Ready for next review from @sergisiso as long as travis is happy.

src/psyclone/tests/psyGen_test.py
```diff
@@ -3173,6 +3173,64 @@ def test_reference_can_be_printed():
...
+    assert ("The scope node 'hello' provided to the symbol method, is not "
+            "an ancestor of this reference node 'Reference[name:'alpha']'."
+            in str(excinfo.value))
+
+    # Symbol method with invalid scope location
+    with pytest.raises(TypeError) as excinfo:
+        _ = alpha.symbol(scope_limit=alpha)
+    assert ("The scope node 'Reference[name:'alpha']' provided to the symbol "
+            "method, is not an ancestor of this reference node "
+            "'Reference[name:'alpha']'." in str(excinfo.value))
```
sergisiso: I am still not convinced of this, I think the first one is right now, a 'hello' string is a TypeError. But the second should be a ValueError or AttributeError. You probably need to modify the symbol method to handle the TypeError just at the begging with an `isinstance(scope_limit, Node)`

rupertford: Modified, although I must admit the original approach was to have a small number of PSyclone-specific exceptions which could then be picked up by anything external. The way we are going, it is going to be difficult to do anything other than catch a generic exception which will causes issues when linking packages together. Perhaps this is something to discuss at a future telco.

rupertford:
> @rupertford See the 1 remaining in-line comment. Also note there are TODOs referencing this issue:
> 
> ```
> $ grep -rn "393" src/
> src/psyclone/psyGen.py:4126:                # the subject of #393.
> src/psyclone/psyGen.py:4216:        of #393.
> src/psyclone/psyGen.py:4238:        # subject of #393 
> src/psyclone/tests/kernel_transformation_test.py:249:    back-end. This limitation is the subject of issue #393.
> src/psyclone/tests/psyGen_test.py:4198:    # #393.
> ```
> 
> Is this PR only a step to solve #393 and the issue will remain open? Or do you prefer to point the TODOs with a newer issue?
> 
> All the rest is ready to be merged.

Very good point @sergisiso. There is a new issue - #520 - which takes over the last part of #393. I've pointed the TODO's at this new issue. This means that #393 can be closed once this PR is on master.

rupertford:

Issue #557, opened 2019-10-24
[PSyIR] Refactor directive support to remove code from base class


hiker:
As discussed, the support for directives in PSyIR moved common code to the base class. Since nbackends like SIR does not need this, @rupertford prefers not to have any code creation happening in the base class. This would also avoid the issue that SIR triggers some pylint warnings since it does not implement some abstract function (which it does not need).

arporter:
PR has been merged.

PR #558, opened 2019-10-24
#557 Removed 'code creation' code from visitor into the C and Fortran…

hiker:
… backends ... as discussed in email thread. This should be a simple and straight forward minor patch.


arporter:
I've taken the liberty of fixing the one and only typo myself.

Issue #552, opened 2019-10-18
Kern refactoring


hiker:
ATM the design of the various 'kern' classes is based on the original design that each kernel is a call (note that I show the diagram including the change of adding an InlinedKern as a base class for NemoKern, in current master NemoKern inherits from CodedKern, same as DynKern and GOKern):
```
Kern  +-- Builtin
      |
      +-- CodedKern +-- DynKern
      |             |
      |             + -- GOKern
      |
      +-- InlinedKern --- NemoKern

```
Kern itself includes handling of arguments, code that is not at all used by InlinedKern (NemoKern, same in current master). NemoKern does not even call the constructor of Kern (if I am not mistaken).

I think this code could benefit from some refactoring, e.g. (a very first draft only, I don't like the name CodedKern ... all kerns are coded ;)  ):
```
Kern  +-- CalledKern +-- DynKern
      |              |
      |              +-- GOKern
      |
      |
      +-- InlinedKern +-- NemoKern
                      |
                      +-- Builtin

```
This is based o the idea that a Builtin kernel is inlined (which is what I think it is, of course, if a builtin is called, it should be moved to CalledKern). I am also not too familiar with the CodedKern and Kern differences. If it makes sense to have this distinction, then perhaps CalledKern could be split into CalledKern, and UserSuppliedKern  (from which then DynKern and GOKern is derived):
```
Kern  +-- CalledKern +-- UserSuppliedKern +-- DynKern
      |              |                    |
      |              |                    + --GOKern
      |              |
      |              + -- (LibraryKern???)
      |
      |
      +-- InlinedKern +-- NemoKern
                      |
                      + -- Builtin

```

This would also allow us to handle non-user-supplied kernels seamlessly (e.g. think of a math library or so ... not sure if we really need it, so I wouldn't do anything about it till we actually need it).

Back to @rupertford, @arporter, @sergisiso, @TeranIvy for discussion and extensions.

arporter:
Just a couple of notes:

We went for `CodedKern` as not all Kernels are provided to PSyclone as source code e.g. BuiltIns. We (well Rupert strictly) has always had the idea that BuiltIns could in fact be calls to some optimised library where appropriate and of course, although we do in-line the code when we generate the Fortran for the built-in, this is an implementation choice.

arporter:
Although NemoKern does not have explicit arguments, it is natural
to think of it having an interface with inputs and outputs -
essentially this is what we need when thinking about putting it
[...]

Issue #483, opened 2019-08-30
Broken OMP private variable detection or ... missing parent??


hiker:
Fixing the handling of loop boundaries (i.e. they are now available in PSyIR, not just as a string) broke my detection of OMP private variables:
```
    !$omp parallel default(shared), private(cff1,cff2,i,j)
    Jstr = BOUNDS(ng) % Jstr(tile)
...
    !$omp do schedule(static)
    DO j = Jstr, Jend
      DO i = IstrU, Iend
        cff1 = 0.25_r8 * (vbar(i, j, krhs) + vbar(i - 1, j, krhs) + vbar(i - 1, j + 1, krhs))
        cff2 = SQRT(ubar(i, j, krhs) * ubar(i, j, krhs) + cff1 * cff1)
...
    !$omp end parallel
...
    DO j = Jstr, Jend
```

Now this code needs #478 (to disable the node type check in the omp transform, which on master does not allow the assignment to be done in parallel) in order to work.

Issue is that in the past jstr was only detected with one WRITE access (because the READ in the loop was not detected).

My detection of omp private variables now sees a write statement, followed by a read statement, and decides to make jstr a private variable. Which then breaks the second loop, since now jstr is not initialised.

I totally agree that the handling of scalar variables need to be improved (ideally by renaming variables to distinguish shared and private use ... ROMS often uses the same variable sometimes shared, sometimes private in the same OMP region), but this is a complicated issue ;)

I decided to just improve my private variable detection  above by checking if the write (that triggers the declaration of jstr being a private one) happens in a loop. This way I can distinguish the handling WRITE followed by READ access of jstr as shared, and cff1 as private. So if I find a write followed by read access, I go up in the tree, till I either find the omp parallel (do) directive, or I find a loop. If it is a loop, the variable must be private. Unfortunately this does not work, since I get a node without a parent before the loop node is reached. Here the tree I see following the _children attribute (and I indicate the parent as well):
```
OMPParallelDirective
  Assignment
  ...
  OMPDoDirective: parent: OMPParallelDirective
    NemoLoop: parent: OMPDoDirective
      Start
      Stop
      Step
      Schedule: parent: NemoLoop
        NemoLoop: parent: Schedule
          Start
          Stop
          Step
          Schedule: parent: NemoLoop
            NemoKern (via _kern_schedule, attribute, not _children):  --> NO PARENT
              Assignment: parent: NemoKern
              Assignment
              Assignment

```

[...]

PR #559, opened 2019-10-25
483 create inlined kernel

hiker:
This pull request creates a new class InlinedKernel, which is the base class for all Nemo kernels.This allows that the KernelSchedule gets a pointer to NemoKern (which we don't have in any other API, since a kernel there is a separate subroutine).

This link to the parent is required to properly detect omp private/shared variables #483, which is fixed with this pull request.

Note that his is NOT a refactoring as discussed in #552, it basically does the same thing NemoKern currently does - inheriting from CodedKern, but not calling its constructor. #552 would clean this up.


hiker:
Oops - that was an embarrassing copy&paste error, thanks, travis :)

src/psyclone/tests/nemo/test_files/explicit_do_two_loops.f90
```diff
@@ -0,0 +1,61 @@
+! -----------------------------------------------------------------------------
+! BSD 3-Clause License
+!
+! Copyright (c) 2017-2018, Science and Technology Facilities Council.
```
arporter: If this is an entirely new file then just 2019 please.

src/psyclone/tests/nemo/transformations/openmp/openmp_test.py
```diff
@@ -89,6 +89,36 @@ def test_omp_explicit_gen():
...
+
+    In this case jpk should not be declared private, since then it
+    is not defined in the next loop.'''
+
+    _, invoke_info = parse(os.path.join(BASE_PATH,
+                                        "explicit_do_two_loops.f90"),
+                           api=API, line_length=False)
+    psy = PSyFactory(API, distributed_memory=False).create(invoke_info)
+    schedule = psy.invokes.get('explicit_do').schedule
+    schedule.view()
```
arporter: Please remove this line.

src/psyclone/nemo.py
```diff
@@ -377,14 +388,18 @@ class NemoKern(CodedKern):
...
         # The corresponding set of nodes in the fparser2 parse tree
         self._ast = parse_tree
-        # Create a kernel schedule
-        self._kern_schedule = KernelSchedule(self._name)
+        # Create a kernel schedule. Since in NEMO all functions are
+        # inlined, set this NemoKern as parent of the KernelSchedule.
+        # This will allow the dependency analysis going up from
+        # a kernel to the loops (which is required for declaring
+        # private variables).
+        self._kern_schedule = KernelSchedule(self._name, self)
```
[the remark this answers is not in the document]
hiker: I hope I have done what you wanted, but unfortunately this caused a bit of a problem. In transformations.py is the code to check if the node types are valid:
```
                flat_list = [item for item in child.walk(object)
                             if not isinstance(item, Schedule)]
```
Now this loop returns all PSyIR nodes (e.g. assignments, array references, ...) - and then declares that the transformation can not be applied. As an immediate fix I have just disabled this test (*grin* glad that the option-for-transforms pull request was merged). But that's obviously not  a good idea.

My suggested solution:
We add a new class 'Statement(Node)', which is otherwise empty. Then all statement nodes (IfBlock, Loop, ...) will be based on Statement (instead of node). Then the walk statement can collect only Statement nodes.

Alternative: we add a flag 'is_statement'?? Suggestions welcome :)

src/psyclone/nemo.py
```diff
@@ -460,6 +485,17 @@ def ast(self):
...
+    def gen_code(self, parent):
+        '''This method must not be called for NEMO, since the actual
+        kernels are inlied.
+
+        :param parent: The parent of this kernel call in the f2pygen AST.
+        :type parent: :py:calls:`psyclone.f2pygen.LoopGen`
+
+        :raises InternalError: if this function is called.
+        '''
+        raise InternalError("Do not call gen_code for a NEMO kernel.")
```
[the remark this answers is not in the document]
hiker: OK.

src/psyclone/psyGen.py
```diff
@@ -2568,7 +2569,28 @@ def _get_private_list(self):
...
+
+                # Go up the tree till we either find the InvokeSchedule,
+                # which is at the top, or a Loop statement (or no parent,
+                # which means we have reached the end of a called kernel).
+                parent = accesses[0].node
+                while parent and \
+                        not isinstance(parent, (Loop, InvokeSchedule)):
+                    parent = parent.parent
+
+                if parent and isinstance(parent, Loop):
```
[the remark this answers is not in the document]
hiker: Gee, that had me think for the better part of an hour. E.g.:
```
!$omp parallel ...
do i=1, n
    f=a(i)+b(i)
    c(i) = f+sqrt(f)
enddo
```
So even though the loop is not parallel, f should be private (not that the loop makes any sense ;) ). Reductions would be an exception:
```
!$omp parallel
   do i=1, n
       f = f + b(i)
```
Reductions are not supported, and it needs an initialisation first (which the existing ode would detect and then at least declare the variable shared ... though no synchronisation is added).

All in all, I think this way is actually good (for now).

src/psyclone/psyGen.py
```diff
@@ -4472,6 +4494,38 @@ def modified(self, value):
...
 
 
+class InlinedKern(Kern):
+    '''A class representing a kernel that is inlined. This is used by
+    the NEMO API, since the NEMO API has no function to call or parameters.
+    '''
+
+    def __init__(self):
+        # pylint: disable=super-init-not-called
+        self._kern_schedule = None
```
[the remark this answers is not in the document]
hiker: Changes, see above.

src/psyclone/psyGen.py
```diff
@@ -4472,6 +4494,38 @@ def modified(self, value):
...
+
+    def __str__(self):
+        return "inlined kern: " + self._name
+
+    @property
+    def coloured_text(self):
```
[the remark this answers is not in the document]
hiker: I think that got basically fixed when merging master, and I could just remove this.

src/psyclone/psyGen.py
```diff
@@ -4472,6 +4494,38 @@ def modified(self, value):
...
+        :returns: the name of this node type, possibly with control codes
+                  for colour
+        :rtype: string
+        '''
+        return colored("InlinedKern", SCHEDULE_COLOUR_MAP["InlinedKern"])
+
+    @abc.abstractmethod
+    def local_vars(self):
+        '''
+        :returns: list of the variable (names) that are local to this loop \
```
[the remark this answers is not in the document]
hiker: Done ... I think?

src/psyclone/psyGen.py
```diff
@@ -6182,14 +6236,17 @@ def __str__(self):
 
 class KernelSchedule(Schedule):
     '''
-    A kernelSchedule inherits the functionality from Schedule and adds a symbol
+    A KernelSchedule inherits the functionality from Schedule and adds a symbol
     table to keep a record of the declared variables and their attributes.
 
     :param str name: Kernel subroutine name
+    :param parent: Parent of the KernelSchedule, defaults to None.
```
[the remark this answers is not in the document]
hiker: ok.

src/psyclone/tests/nemo/test_files/explicit_do_two_loops.f90
hiker: ok.

src/psyclone/tests/nemo/transformations/openmp/openmp_test.py
hiker: Done (I removed all of them, there seems to be no sense in the others either).

src/psyclone/nemo.py
```diff
@@ -421,6 +436,16 @@ def match(node):
         # node in the result of the walk, this node can not be a kernel.
         return len(nodes) == 0
 
+    def get_kernel_schedule(self):
```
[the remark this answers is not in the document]
hiker: Done

hiker:
Back to Andy. It all seems to work, but as mentioned I had to disable the node type check. Let me know if you want a proper fix for this (and which one).

src/psyclone/psyGen.py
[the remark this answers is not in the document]

src/psyclone/nemo.py
```diff
@@ -439,7 +425,19 @@ def get_kernel_schedule(self):
         :returns: the kernel schedule representing the inlined kernel code.
         :rtype: :py:class:`psyclone.psyGen.KernelSchedule`
         '''
-        return self._kern_schedule
+        return self._children[0]
+
+    def node_str(self, colour=True):
```
arporter: I don't think you need this method as it's doing exactly the same as the one in the `Node` base class?

src/psyclone/nemo.py
```diff
@@ -458,7 +456,8 @@ def reference_accesses(self, var_accesses):
         :type var_accesses: \
             :py:class:`psyclone.core.access_info.VariablesAccessInfo`
         '''
-        self._kern_schedule.reference_accesses(var_accesses)
+        print("RA", len(self.children))
```
arporter: Not sure what this `print` is for?

src/psyclone/nemo.py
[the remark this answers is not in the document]
arporter: I don't quite understand. Are we talking about line 164 in transformations.py? It seems to me that it's doing what we want. Are you worried about the performance implications? Perhaps it would be better if we had `self.invalid_node_types` instead of `valid_node_types`? Maybe we need to have a chat about this?

src/psyclone/nemo.py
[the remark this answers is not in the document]
arporter: Great although I see no `kernel_body` method?

hiker:
OK, it looks like I forgot to push some changes, I'll fix this next week, I don't want to rush into pushing more out that might still be incomplete.

Re your question: yes, line 164 of transformations.py. This line finds now _all_ nodes in the InlinedKernel, including nodes like:
```
0: Assignment[]
   ArrayReference[name:'pta']
      Reference[name:'ji']
      Reference[name:'jj']
      Reference[name:'jk']
      Reference[name:'jn']
   BinaryOperation[operator:'ADD']
```
And they all will be tested to see if they are allowed. Before we added the InlinedKernel to have children, the walk would not go 'into' the kernel function, but now it does. And of course none of them are listed to be allowed in an OMP region. That is the problem.

But I also realise that my solution (to 'mark' statements using a special class) does not solve the issue. Yes, it would prevent things like "Reference" to be tested, but it would still find statements that are not marked to be allowed in OpenMP (Assignment, If, ...). So my suggested solution would only reduce the number of nodes down to statements, but not solve the main problem..

The only thing I can (atm) think of is to use a modified walk function, that will not recurse into nodes of a certain type - so we could tell the walk function to stop at an InlinedKernel. I _think_ that would work.

src/psyclone/nemo.py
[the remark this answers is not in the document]
hiker: OK, that push didn't come through (which likely means ... I pushed too early ;) ). But the problem is that this change breaks around 5 tests in sir_test.py. Now those tests all just use get_kernel_schedule for a NemoKern, so I could replace get_kernel_schedule with kernel_body - but the question is: do we really want to have different method names for basically the same functionality? It is called get_kernel_body in CodedKern.

src/psyclone/nemo.py
hiker: The function in the base class uses self._arguments, which is not defined for an Inlined (or Nemo) kern. While I could of course add a method to inlined class that returns an empty string as arguments, I think it would be better for future refactoring not to do this.

src/psyclone/nemo.py
hiker: Oops - removed.

hiker:
OK, not ready for review, but I need feedback (and am happy to wait for the telco):
1) do you want kernel_body for inlined kernel (and modify the tests), or revert this change and leave it as get_kernel_schedule?
2) What do we do with the failing OMP transformations, caused by walk() now returning assignment, Reference, ... nodes in the validate function? Modify walk to avoid it going into the kernel? Instead of testing for only-valid-nodes, test for no-invalid-nodes?

So I'll mark this as ready for review to get your attention, but at this stage all I need is an agreement on those two issues (which we can discuss during the telco).

hiker:
Summary: we agreed to:

- use get_kernel_schedule instead of kern_body to have consistency between InlinedKern and CodedKern.
- We implement a modified walk method that will stop at a specified Class (i.e. kernel).

hiker:
As agreed in the telco: reverted the change to use kernel_body, added an additional optional parameter to the walk function, that prevents from further recursions into the tree. While adding test case for this, I added support for the nemo api to tests.utilities.gen_invoke, and used this in the nemo tests. I also fixed the test  test_implicit_loop_sched2 (see #593 for additional details).

Ready for next review.

src/psyclone/nemo.py
arporter: Oh yes, my mistake.

src/psyclone/nemo.py
```diff
@@ -459,6 +471,19 @@ def ast(self):
         '''
         return self._ast
 
+    def gen_code(self, parent):
+        '''This method must not be called for NEMO, since the actual
+        kernels are inlied.
```
arporter: s/inlied/inlined/

src/psyclone/psyGen.py
```diff
@@ -1502,15 +1503,19 @@ def sameParent(self, node_2):
...
         an instance of 'my_type', which is either a single class or a tuple
         of classes. In the latter case all nodes are returned that are
         instances of any classes in the tuple.
 
         :param my_type: the class(es) for which the instances are collected.
-        :type my_type: either a single :py:class:`psyclone.Node` class\
-            or a tuple of such classes.
+        :type my_type: either a single :py:class:`psyclone.Node` class \
+            or a tuple of such classes
+        :param stop_type: the class(es) that will not be recursed into \
```
arporter: How about: "class(es) at which recursion is halted (optional)". Also, could you mention this parameter in the general description of the method in order to explain its purpose.

src/psyclone/psyGen.py
```diff
@@ -2659,7 +2669,28 @@ def _get_private_list(self):
...
             # assume the variable should be private:
             if accesses[0].access_type == AccessType.WRITE:
-                result.add(var_name.lower())
+                # Check if the write access is inside the parallel loop. If
+                # the write is outside of a loop, it is an assignment to
+                # a shared variable. Example where jpk is likely used
+                # outside of the parallel section later, so it must be
+                # declared as shared in order to have its value in other loops:
+                # !$omp parallel
+                # jpk = 100
```
arporter: Wouldn't this assignment cause us problems anyway as one thread could attempt to read `jpk` while another writes to it? We should really protect it with an `atomic` or some such?

arporter: Or suggest to the user that they move the assignment?

src/psyclone/psyGen.py
```diff
@@ -2659,7 +2669,28 @@ def _get_private_list(self):
...
+                # !omp do
+                # do ji = 1, jpk
+
+                # Go up the tree till we either find the InvokeSchedule,
+                # which is at the top, or a Loop statement (or no parent,
+                # which means we have reached the end of a called kernel).
+                parent = accesses[0].node
+                while parent and \
```
arporter: Looking at this afresh, I think you could use the `Node.ancestor()` method to achieve the same thing.

src/psyclone/tests/nemo/nemo_psy_test.py
```diff
@@ -52,6 +52,16 @@
...
 
 
+def test_no_gen_code():
+    '''Test that we raise an exception if gen_code is called
+    for a NemoKern.'''
+    kern = nemo.NemoKern([], None)
+    with pytest.raises(InternalError) as err:
+        kern.gen_code(None)
+    assert "NEMO kernels are assumed to be in-lined by default therefore " \
+           "the gen_code method should not have been called." in str(err)
```
arporter: Please use `str(err.value)` here.

src/psyclone/tests/utilities_test.py
```diff
@@ -143,26 +144,28 @@ def test_compile_str(monkeypatch, tmpdir):
...
-        _, _ = get_invoke("test11_different_iterates_over_one_invoke.f90",
-                          "invalid-api", name="invalid_name")
+        get_invoke("test11_different_iterates_over_one_invoke.f90",
+                   "invalid-api", name="invalid_name")
     assert "The API 'invalid-api' is not supported" in str(excinfo)
+
+    # Test that a non-existant file raises the right exception
+    with pytest.raises(ParseError) as excinfo:
+        get_invoke("does_not_exist", "nemo", idx=0)
+    assert "No such file or directory" in str(excinfo)
```
arporter: Please use `str(excinfo.value)` here.

src/psyclone/nemo.py
hiker: Done.

src/psyclone/psyGen.py
hiker: Done

src/psyclone/psyGen.py
hiker: Yes, the handling of scalars is a HUGE problem, e.g.:
```
!$omp parallel
   a=100
   b=100
   do i=1, b
     x(i) = a*i
   end do
!$omp end parallel
...
    do i=1, b
      a = x(i)*x(i)
      y(i) = sin(a)
```
In this example, b must be shared, but you can only know that when you see the read access in the 2nd loop. a could be private or shared, but if the second loop is included in the same parallel do directive, it must be private.

Since an assignment is only accepted if explicitly disabling the node-type check (and since this issue is not really related to the InlineKernel (except that the previous shared declaration of b as shared stopped working), I suggest to leave this for now as an issue the user has to handle (by explicitly disabling the node test). I am happy to open a separate issue.

Yes, technically I as a user would have to add an atomic, but since all threads store the same value (and each thread only reads after it has stored this value), the variable should at all times have the correct value (and yes, I have been bitten by a similar assignment, that the compiler decided to implement as 2 4-byte stores once, but in that case the values where not the same between all threads).

src/psyclone/psyGen.py
hiker: No, ancestor() does not return self (if applicable). I tried to support returning self in ancestor, but then the read_ancestor_test failed, so I assume it is important that ancestor does not return self. And I indeed need to be able to find self.

src/psyclone/tests/nemo/nemo_psy_test.py
hiker: Done.

src/psyclone/tests/utilities_test.py
hiker: Done.

hiker:
All issues addressed, except for the handling of shared variables, which (since it requires overwriting the node-type check) is not really a PSyclone issue. Ready for review.

arporter:
@hiker I see no new commits ;-) I'd say whether or not something can be safely parallelised *is* a PSyclone issue but am happy for a new issue to be created for that.

src/psyclone/psyGen.py
arporter: Would you mind creating a new issue for this and mark this bit of code with a TODO?

hiker:
> 
> 
> @hiker I see no new commits ;-) 

Arg - I looked, and saw a push ... just to the wrong repo (the disadvantage of juggling three repos *sigh*)

> I'd say whether or not something can be safely parallelised _is_ a PSyclone issue but am happy for a new issue to be created for that.

OK, I can do that - but PSyclone would not allow omp to be used in the first place - it only does so because I tell it to ignore the node-type check (the assignment statements outside of the loop would not be allowed)

src/psyclone/psyGen.py
arporter: As discussed, please could you open an Issue for this (or check that we don't already have one).

arporter:
Changes are all good now. Just needs the creation of two new issues (sigh) and (at least one) associated TODO and then this can be merged.

src/psyclone/psyGen.py
hiker: Done: #598.

src/psyclone/psyGen.py
hiker: Done: #597.

arporter:
New issues created and TODOs added.
All tests pass/xfail including with compilation.
Docs build OK.
Changes are pylint clean.

PR #562, opened 2019-10-30
473 fix dag

hiker:
👕 👕 I can't say no, right?

This implements @rupertford's short term solution: ignore the start, stop, and step value of loops, and add the schedules with a unique name to the graphs in all tests.


src/psyclone/psyGen.py
```diff
@@ -1136,8 +1136,14 @@ def dag_gen(self, graph):
             graph.edge(remote_name, local_name, color="blue")
         # now call any children so they can add their information to
         # the graph
-        for child in self.children:
-            child.dag_gen(graph)
+        if isinstance(self, Loop):
+            # Ignore the children that represents the start, stop
+            # and step value of the loop:
+            for child in self.children[3:]:
```
arporter: Please use `loop_body` here.

src/psyclone/tests/psyGen_test.py
```diff
@@ -2366,15 +2374,17 @@ def test_node_dag(tmpdir, have_graphviz):
     schedule.dag(file_name=my_file.strpath)
     result = my_file.read()
     print(result)
```
arporter: Please remove all three of these print statements.

src/psyclone/psyGen.py
hiker: Oops - how embarrassing.

src/psyclone/tests/psyGen_test.py
hiker: Done

hiker:
That should be ready for review - the failing coverage seems to be caused by travis not executing the dag tests.

arporter:
All requested changes have been made. All tests pass/xfail when I have the graphviz package installed. I was on the point of accepting when I remembered documentation :-( But then, I've realised that my PR #580 also requires that that same documentation be updated so I'm going to be kind and accept this as is :-)

Issue #375, opened 2019-05-06
[PSyIR] needs to capture Fortran kind declaration information


rupertford:
Fortran declarations can include the `kind` attribute, for example `integer(kind=i_def) :: idx`. This information needs to be captured in the PSyIR somehow so that the `kind` information can be reproduced if deparsing from PSyIR back to Fortran.

arporter:
Currently the Fortran back-end to the PSyIR is hardcoded to insert `use constants_mod, only: r_def, i_def` but this is LFRic specific. This means that kernel generation does not work for the GOcean API which does not have `constants_mod`.

arporter:
We've agreed that a Symbol needs a new size property that can be either a value, a reference (to a Symbol holding the KIND parameter) or None if it is a default type (e.g. real with no KIND specified).

arporter:
For e.g. `real(kind=wp) :: var` we get a `Fortran2003.Kind_Selector`:

    (Pdb) type_spec.items[1].items
    ('(', Name('wp'), ')')

arporter:
This means that the new `precision` argument that I've added to the `Symbol` constructor can be either a reference to another Symbol, an integer or a string (label).

rupertford:
Note that this PR #563 does not complete this issue as

1) the backends do not yet generate the appropriate code.
2) the following form of input is not yet captured rtime = real(istep,go_wp) * rdt.

Both of the above reference this issue in the source code. Note, the first should definitely be fixed before closing this issue, however the second is probably a separate issue as it is not to do with declaring data.

rupertford:
PR #563 is on master so closing this issue.

PR #563, opened 2019-10-30
Add support for processing KIND parameters (#375)

arporter:
Adds support to the fparser2 frontend for handling declarations with KIND parameters.


arporter:
Almost there. I have one failing test which relates to the OpenCL backend - we now get a declaration of an integer variable that is the KIND parameter found when parsing the Fortran. Although this doesn't really cause a problem it does highlight the fact that the KIND Parameter is something that exists only in Fortran and therefore should possibly not be treated as a completely standard Symbol?

arporter:
Ready for first review from @rupertford. Note that coverage of the diff is not quite 100% because CodeCov now includes the test files and I have an x-failing test so that not all of it is executed.

rupertford:
Note that this PR does not result in #375 being completed as the backends do not yet generate the appropriate code. As a result there are still references to #375 in the source code.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -650,6 +654,84 @@ def process_declarations(self, parent, nodes, arg_list):
...
+                kind_symbol = symbol_table.lookup(kind_name)
+                if not isinstance(kind_symbol, PrecisionSymbol):
+                    # The Symbol table contains an entry for the kind
+                    # parameter but it has not yet been identified as such. We
+                    # therefore replace the existing entry with a
+                    # PrecisionSymbol.
+                    new_symbol = PrecisionSymbol(kind_name)
+                    new_symbol.copy_properties(kind_symbol)
+                    # Old symbol may have had 'deferred' type
+                    new_symbol.datatype = "integer"
```
rupertford: @sergisiso  was quite keen to avoid potential inconsistencies in the symbol table so the addition of a remove method is something I think we should pass by him. I can't see how the current copy and swap methods would work (as we want to replace an instance of a class with an instance of a sub-class of the class). Perhaps a replace() method could be added?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -650,6 +654,84 @@ def process_declarations(self, parent, nodes, arg_list):
...
+            # We have kind=var
+            kind_name = str(kind_names[0])
+            try:
+                kind_symbol = symbol_table.lookup(kind_name)
+                if not isinstance(kind_symbol, PrecisionSymbol):
+                    # The Symbol table contains an entry for the kind
+                    # parameter but it has not yet been identified as such. We
+                    # therefore replace the existing entry with a
+                    # PrecisionSymbol.
+                    new_symbol = PrecisionSymbol(kind_name)
```
rupertford: OK, so I don't think we should have a PrecisionSymbol. I think we should just use a Symbol. I understand that the precision variable will be declared in languages other than Fortran but I think that is right. It is actually just an integer parameter which can be used in other contexts as well. For example:
```
subroutine test()
  integer, parameter :: r_def = 4
  real(kind=r_def) :: a
  a = r_def
end subroutine
```
If we end up agreeing on this point then it simplifies this PR quite a bit as we no longer need the remove() method in the SymbolTable and no longer need to distinguish between symbols.

rupertford:
Back to @arporter.

src/psyclone/psyir/frontend/fparser2.py
arporter: Oh yes, I hadn't thought about such a symbol being used in other contexts. I'll change it back to being a `Symbol`.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -650,6 +654,84 @@ def process_declarations(self, parent, nodes, arg_list):
...
+           isinstance(intrinsics[0].items[0],
+                      Fortran2003.Intrinsic_Name) and \
+                      str(intrinsics[0].items[0]).lower() == "kind":
+            # We have kind=KIND(literal). items[1] is an
+            # Actual_Arg_Spec_List with the first entry being the
+            # literal.
+            literal = intrinsics[0].items[1].items[0]
+            if isinstance(literal,
+                          Fortran2003.Real_Literal_Constant):
+                if "d0" in str(literal).lower():
```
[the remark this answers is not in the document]
arporter: Oh yes, that's a good point. I was writing a frontend for the code I know rather than allowing for the full spectrum :-) I've stared hard at the spec and I think pretty much anything goes :-( Obviously, `kind=KIND(1_idef)` == `kind=idef` so it's pretty pointless. I'll give it some thought.

src/psyclone/psyir/frontend/fparser2.py
arporter: I've now removed this change.

arporter:
While making the requested changes, I realised that I had a assumed that `KIND(some-int-literal)` always meant a default-precision integer. However, if `some-int-literal` is too large to be represented by 4 bytes then presumably we must get a KIND that indicates an 8-byte integer. I've therefore extended to cover this case but am unsure whether we should use `precision=4`/`8` or add e.g. `Symbol.Precision.LONG` - something for discussion.

rupertford:
> While making the requested changes, I realised that I had assumed that `KIND(some-int-literal)` always meant a default-precision integer. However, if `some-int-literal` is too large to be represented by 4 bytes then presumably we must get a KIND that indicates an 8-byte integer. I've therefore extended to cover this case but am unsure whether we should use `precision=4`/`8` or add e.g. `Symbol.Precision.LONG` - something for discussion.

In gfortran it seems that an integer literal picks up the default precision. This means that it is not possible to specify a number greater than the default precision. To get something larger you need to specify kind somewhere, which we could presumably pick up via the variable declaration? I'm not sure if this is "standard" or not, but if so, it would be right to assume a literal integer in a kind intrinsic implies default precision.

arporter:
> 
> 
> > While making the requested changes, I realised that I had assumed that `KIND(some-int-literal)` always meant a default-precision integer. However, if `some-int-literal` is too large to be represented by 4 bytes then presumably we must get a KIND that indicates an 8-byte integer. I've therefore extended to cover this case but am unsure whether we should use `precision=4`/`8` or add e.g. `Symbol.Precision.LONG` - something for discussion.
> 
> In gfortran it seems that an integer literal picks up the default precision. This means that it is not possible to specify a number greater than the default precision. To get something larger you need to specify kind somewhere, which we could presumably pick up via the variable declaration? I'm not sure if this is "standard" or not, but if so, it would be right to assume a literal integer in a kind intrinsic implies default precision.

That's a good point. I've done some googling and it seems that an integer without an explicit KIND is always of default precision. I've reverted my changes :-)

arporter:
OK, I think this is ready for next review, Travis permitting...

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -650,6 +654,113 @@ def process_declarations(self, parent, nodes, arg_list):
...
+
+        :param symbol_table: the SymbolTable associated with the code \
+                             being processed.
+        :type symbol_table: :py:class:`psyclone.psyGen.SymbolTable`
+        :param type_spec: the fparser2 parse tree of the type specification.
+        :type type_spec: :py:class:`fparser.two.Fortran2003.Intrinsic_Type_Spec
+
+        :returns: the precision associated with the type specification.
+        :rtype: int or :py:class:`psyclone.psyGen.Symbol.Precision` or \
+                :py:class:`psyclone.psyGen.PrecisionSymbol` or NoneType
```
rupertford: PrecisionSymbol should be changed to Symbol

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -650,6 +654,113 @@ def process_declarations(self, parent, nodes, arg_list):
...
+        parameter. If the supplied Symbol Table does not contain an appropriate
+        entry then one is created. If it does contain a matching entry then
+        its datatype must be 'integer' or 'deferred'. If the latter then the
+        fact that we now know that this Symbol represents a KIND parameter
+        means that we can change the datatype to be 'integer'.
+
+        :param str name: the name of the variable holding the KIND value.
+        :param symbol_table: the Symbol Table associated with the code being\
+                             processed.
+
```
rupertford: Type of symbol_table should be specified.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -650,6 +654,113 @@ def process_declarations(self, parent, nodes, arg_list):
...
+        :param symbol_table: the SymbolTable associated with the code \
+                             being processed.
+        :type symbol_table: :py:class:`psyclone.psyGen.SymbolTable`
+        :param type_spec: the fparser2 parse tree of the type specification.
+        :type type_spec: :py:class:`fparser.two.Fortran2003.Intrinsic_Type_Spec
+
+        :returns: the precision associated with the type specification.
+        :rtype: int or :py:class:`psyclone.psyGen.Symbol.Precision` or \
+                :py:class:`psyclone.psyGen.PrecisionSymbol` or NoneType
+
```
rupertford: NotImplementedErrors should be specified in the docstring.

src/psyclone/psyGen.py
```diff
@@ -5446,6 +5446,10 @@ class Symbol(object):
                            datatype of the constant value must be \
                            compatible with the datatype of the symbol.
     :type constant_value: int, str or bool
+    :param precision: The amount of storage required by the datatype (bytes) \
```
rupertford: 'the' not 'The' please

src/psyclone/psyGen.py
```diff
@@ -5446,6 +5446,10 @@ class Symbol(object):
...
+            or a reference to a Symbol holding the type information \
+            or a label identifying a default precision.
+    :type precision: int or :py:class:`psyclone.psyGen.Symbol` or str
```
rupertford: Are these the correct types? What about Symbol.Precision? Can it be str or int at the moment?

src/psyclone/psyGen.py
```diff
@@ -5546,15 +5558,33 @@ def module_name(self, value):
...
+                    " or Symbol but got '{0}'".format(
+                        type(precision).__name__))
+            if isinstance(precision, int) and precision <= 0:
+                raise ValueError("The precision of a Symbol when specified as "
+                                 "an integer number of bytes must be > 0 but "
+                                 "got {0}".format(precision))
+            if isinstance(precision, Symbol) and \
+               (precision.datatype not in ["integer", "deferred"]
+                or precision.is_array):
+                raise ValueError("A Symbol representing the precision of "
```
rupertford: You allow "deferred here too" but do not say so in the error condition.

src/psyclone/psyGen.py
```diff
@@ -5596,6 +5626,20 @@ def datatype(self):
...
+        ''' Setter for Symbol datatype.
+
+        :param str value: new value for datatype.
+
+        :raises NotImplementedError: if the specified data type is invalid.
+        '''
+        if value not in Symbol.valid_data_types:
+            raise NotImplementedError(
+                "Symbol can only be initialised with {0} datatypes but found "
+                "'{1}'.".format(str(Symbol.valid_data_types), value))
```
rupertford: Will this output unicode strings for Python2?

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -423,6 +427,125 @@ def test_process_declarations_intent(f2008_parser):
         Symbol.Access.READWRITE
 
 
+def test_process_declarations_kind_new_param(f2008_parser):
+    ''' Test that process_declarations handles variables declared with
+    an explicit KIND parameter that has not already been declared. '''
```
rupertford: Please put ''' on a newline with a blank line inbetween. Similarly for other new tests.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -423,6 +427,125 @@ def test_process_declarations_intent(f2008_parser):
...
+    from fparser.two.Fortran2003 import Specification_Part
```
rupertford: Move import to top of file and remove duplication of import from new tests.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -423,6 +427,125 @@ def test_process_declarations_intent(f2008_parser):
...
+    fake_parent = KernelSchedule("dummy_schedule")
```
rupertford: All new tests use the same processor, fake_parent, reader, fparser2spec but with different input Fortran. These should be made into a utility function.

rupertford:
Back to @arporter

src/psyclone/psyir/frontend/fparser2.py
arporter: Oops.

src/psyclone/psyir/frontend/fparser2.py
arporter: Oh yes. I'd forgotten to document the return value too.

src/psyclone/psyir/frontend/fparser2.py
arporter: Oops.

src/psyclone/psyGen.py
arporter: Oh yes.

src/psyclone/psyGen.py
arporter: Oh yes, apologies.

src/psyclone/psyGen.py
arporter: Oh yes.

src/psyclone/psyGen.py
arporter: As agreed, valid_data_types are strings we create internally and I've now added a check that `value` is a `str`.

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: OK. I don't like it but I can live with it :-)

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Done.

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Oh yes, that's nice.

arporter:
Travis permitting, this is ready for review.

Issue #567, opened 2019-10-30
[LFRic] Check the behaviour of the force-profile invokes option.


rupertford:
LFRic developers have reported that the force-profile invokes does not profile the whole of each invoke and does not provide names consistent with an invoke call. This needs to be checked and fixed.

rupertford:
The first problem I've found is that the profiling code raises an exception if the first kernel in the invoke is a builtin.

The second problem is that the names used to define a region are always based on the name of the first kernel in the invoke irrespective of where the profile calipers are placed. Therefore when `force-profile invokes` is added the names of the profile regions do not reflect the fact that they are for the whole invoke.

I think the names of the profile regions need to be better chosen (including fixing the error associated with builtins).

rupertford:
Created branch 567_profile_no_exception to fix the first problem. If/when this is merged to master this issue should not be closed as the second problem remains.

rupertford:
Created branch 567_profile_names.
This will allow users to set the profile names, which could be useful when manually adding profile nodes and also chooses better (unique and more appropriate) names automatically if names are not supplied.

rupertford:
The next (and possibly final) thing to do in this issue is to make the profile transformation choose better (unique and more appropriate) names automatically if names are not supplied. At the moment the names chosen for invokes are confusing and not unique.

rupertford:
Created branch 567_unique_profile_names

rupertford:
Once PR #618 (for issue #567) is on master this issue can be closed.

rupertford:
force-profile is not required and has been removed from psyclone. Closing this issue.

PR #574, opened 2019-11-07
fix for profiling with builtins (for #567)

rupertford:


rupertford:

Issue #509, opened 2019-09-13
Labeled section groups and home page in the User Guide


sergisiso:
Thinking about there to add #504 I thought it is already a bit difficult to find things in the User Guide. Maybe something like grouping the sections with common topics (see blue labels in
https://docs.readthedocs.io/en/stable/ ) would make it easier.

Also the empty Home page with the long TOC when first opening the RTD is something that should be improved.

This would be an easy change as it just requires updating the index.rst (all the other files stay the same). I propose something like:

Move Intro to the top of Home page and reduce the number of layers in the TOC

[Introduction]
- 	Getting Going
- 	Examples
- 	System-specific Set-up for Users
- 	Configuration
- 	Changelog (if we could generate this automatically)

[PSyKAl] (For the science domain user)
- 	Kernel Layer
- 	Algoritm Layer
- 	PSy Layer
- 	Built-ins

[PSyclone Scripts] (For the HPC user)
- 	PSyclone Scripts
- 	Distributed Memory
- 	The PSyIR
- 	Transformations

[APIs] (do we still call them APIs?)
- 	Dynamo 0.3
- 	GOcean 1.0
- 	Nemo

[Utilities]
- 	Stub Generation
- 	PSy Kernel Extractor
- 	Profiling

[Conventions/Gotchas ?]
- 	Line Length
- 	Fortran Naming Conventions

Bibliography

What do you think?

arporter:
I like that suggestion - it makes the documentation more user-friendly. Possibly we're replacing "API" with "front-end"?

sergisiso:
My worry with "front-end" is that can be misleading with the  compiler front-end terminology, which for the PSyIR would the fparser2 at the moment. But I don't have a better suggesting.

rupertford:
I like the suggestion too.

If we're going to continue to use back-ends to describe transforming from PSyIR to something else then I think we should have front-ends to describe transforming from something else to PSyIR.

The one change I would make to the proposed structure is to put APIs/Front-Ends before PSyclone scripts as this section is also for the science domain user so follows PSyKAl naturally and would need to be introduced before transformations/scripts in any case as some of these are API/front-end specific.

Oh, and isn't linelength a utility?

arporter:
PR now on master.

PR #579, opened 2019-11-11
Issue 509. Update User's Guide entry page

sergisiso:
As described in #509 I think adding labels and a small intro the the RTD webpage will improve the readability and navigation of the documentation. The changes I made in this PR just modify the `index.rst` and should not modify the PDF version of the documentation (other than reducing the toc depth to 2 but I can undo this if you don't like it)

The results look like this:
image


doc/user_guide/conf.py
```diff
@@ -139,7 +139,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+# html_static_path = ['_static']
```
arporter: Aha! Thanks for fixing.

doc/user_guide/index.rst
```diff
@@ -38,32 +33,85 @@
    -----------------------------------------------------------------------------
    Written by R. W. Ford and A. R. Porter, STFC Daresbury Lab
 
+.. only:: html
+
+    .. image:: psyclone_v1.0.png
```
arporter: Do we have to make a copy of this image file or is it possible to use the one in `../logo`?

doc/user_guide/index.rst
```diff
@@ -38,32 +33,85 @@
...
+        :alt: PSyclone
+
+PSyclone User's Guide
+=====================
+
+
+.. only:: html
+
+    PSyclone is a code generation system developed to produce domain-specific
+    languages (DSLs) for finite element, finite volume and finite difference
```
arporter: Sorry to be picky but PSyclone doesn't produce DSLs. Maybe "developed to support domain-specific..."

doc/user_guide/index.rst
```diff
@@ -38,32 +33,85 @@
...
+    codes. Notably, it is used in the
+    `LFRic Project <https://www.metoffice.gov.uk/research/modelling-systems/lfric/>`_,
+    but it is also being extended to support the
+    `GOcean <https://puma.nerc.ac.uk/trac/GOcean>`_ and
```
arporter: Picky again but GOcean isn't an ocean model, maybe it's a DSL?
How about:
"and it also supports the GOcean (2D, finite difference) DSL. In addition it is being extended to support the processing of existing finite difference codes such as the NEMO ocean model."

doc/user_guide/index.rst
sergisiso: Ups... you are right, updated reference and deleted copy

doc/user_guide/index.rst
sergisiso: Yeah 'support' is better

doc/user_guide/index.rst
sergisiso: Done

sergisiso:
@arporter Ready for next review. (Now builds with no warnings :))

doc/user_guide/index.rst
```diff
@@ -38,32 +33,86 @@
...
    Written by R. W. Ford and A. R. Porter, STFC Daresbury Lab
 
+.. only:: html
+
+    .. image:: ../logo/psyclone_v1.0.png
+        :width: 75%
+        :align: center
+        :alt: PSyclone
+
+PSyclone User's Guide
```
arporter: I've done some googling on "Users Guide", "User's Guide" etc. and think that "User Guide" is what we should call this document.

doc/user_guide/index.rst
```diff
@@ -38,32 +33,86 @@
...
-   line_length
-   fortran_naming_conventions
-   api
-   profiling
-   psyke
-   system_specific_setup
-   configuration
-   zz_bibliography
+    :maxdepth: 2
+    :caption: Introduction
```
arporter: As discussed, please could you alter this so that the PDF document has the same structure.

doc/user_guide/index.rst
sergisiso: Done

doc/user_guide/index.rst
sergisiso: Done

sergisiso:
@arporter Ready for next review.

Issue #578, opened 2019-11-10
[PSyIR] add support for kind to Fortran backend


rupertford:
#563 has added support for variable precision information derived from Fortran KIND information in the original code. This precision information needs to be output as KIND information in the Fortran backend.

rupertford:
Created branch 578_fortran_kind

arporter:
#581 has been merged and remaining problems are captured in #587.

Issue #11, opened 2017-03-22
Add support for logging


arporter:
PSyclone currently has no mechanism for issuing messages other than when an exception is raised. However, there are many cases where we might want to generate a warning to the user but allow execution of PSyclone to continue.

In this ticket we will add an interface to a logging mechanism that will allow us to generate warning/debug messages.

MatthewHambley:
This is probably related to issue stfc/fparser#65.

arporter:
Python provides the `logging` module which probably does what we want. (https://docs.python.org/3.8/library/logging.html) I'm experimenting with using it in my NEMO optimisation script and so far, so good.

rupertford:
in #581 some simple logging was added (and then commented out). However, when this is tested as part of the full set of tests the log information is not output. When the test is run in isolation or just the psyir directory tests are run then the log information is output. I've created a failing test for this (again which is now commented out), however the test does not fail (so an error is reported) if the full test suite is not run. I'm guessing it is to do with fparser and the way the logging is manipulated there.

arporter:
We need to decide what functionality we want from the logging system:

 1. Do we log to file or to the console?
 2. Do we want it to be configurable? (e.g. specify default level and the destination in the config file or via flags.)
 3. If logging to file, do we create a new one or append to an existing one?

Personally, I think it should be possible to control the logging from the command line (e.g. to make it more or less verbose by changing the default logging level). Probably it would make sense to also allow the control of the destination of the logging output from the command line too.

What do @rupertford, @hiker, @TeranIvy and @sergisiso think?

rupertford:
Yes, I think it should be configurable for level and where it is output. Perhaps it should be on the command line for overrides and in a config file for site-specific defaults.

hiker:
I am all in favour for KISS re where to output: output to stderr (or stdout if preferred), if someone wants it to a file, the build system can easily redirect. 

[...]

PR #581, opened 2019-11-11
Fortran backend precision support (#578)

rupertford:


rupertford:
Ready for review from either @arporter or @sergisiso.  Note, there is one question in the text from which I would appreciate feedback on what to do.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -109,6 +109,79 @@ def gen_dims(symbol):
...
+        datatype = "logical"
+    else:
+        datatype = symbol.datatype
+    if not symbol.precision:
+        # This symbol has no precision information so simply return
+        # the name of the datatype.
+        return datatype
+    if isinstance(symbol.precision, int):
+        if datatype not in ['real', 'integer', 'logical']:
+            raise VisitorError("Explicit precision not support for datatype "
```
arporter: "supported"

src/psyclone/psyir/backend/fortran.py
```diff
@@ -109,6 +109,79 @@ def gen_dims(symbol):
...
+
+    if isinstance(symbol.precision, Symbol.Precision):
+        # The precision information is not absolute so is either
+        # machine specific or is specified via the compiler. Fortran
+        # only distinguishes relative precision for single and double
+        # precision reals.
+        if datatype.lower() == "real" and \
+           symbol.precision == Symbol.Precision.DOUBLE:
+            return "double precision"
+        import logging
```
arporter: As discussed, although this is a clear use case for logging, we shouldn't just introduce it here without also adding top-level controls to the driver script. Therefore please comment it out for now and add a reference to the logging Issue (#11 ).

src/psyclone/psyir/backend/fortran.py
```diff
@@ -109,6 +109,79 @@ def gen_dims(symbol):
...
+        logging.warning(
+            ("Fortran does not support relative precision for the '%s' "
+             "datatype but '%s' was specified for variable '%s'.",
+             datatype, str(symbol.precision), symbol.name))
+        return datatype
+    if isinstance(symbol.precision, Symbol):
+        if datatype not in ["real", "integer", "logical"]:
+            raise VisitorError("kind not support for datatype "
```
arporter: "supported"

src/psyclone/psyir/backend/fortran.py
```diff
@@ -109,6 +109,79 @@ def gen_dims(symbol):
...
+                               "'{0}' in Fortran backend.".format(datatype))
+        # Question: Should we check whether the value of precision is
+        # known and, if so, what should we do if not?
+        # The precision information is provided by a parameter, so use KIND.
```
arporter: As discussed, I think we should opt for lazy evaluation here and not attempt to discover things that we don't actually need to know.

src/psyclone/psyir/backend/fortran.py
arporter: (Also, you have superfluous parentheses in the `logging.warning` call.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -109,6 +109,79 @@ def gen_dims(symbol):
     return dims
 
 
+def gen_datatype(symbol):
+    '''Given a Symbol instance as input, return the datatype of the
```
arporter: Please include the name of the Symbol in the error messages produced by this routine.

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -111,6 +112,115 @@ def test_gen_dims_error(monkeypatch):
     assert "unsupported gen_dims index 'invalid'" in str(excinfo)
 
 
+def test_gen_datatype():
+    '''Check the gen_datatype function produces the expected datatypes.'''
+    for datatype in ["real", "integer", "character"]:
```
arporter: `@pytest.mark.parametrize`?

arporter: Actually, you could fold this test into the following one?

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -111,6 +112,115 @@ def test_gen_dims_error(monkeypatch):
...
+    precision is specified.
+
+    '''
+    symbol = Symbol("dummy", datatype, precision=precision)
+    assert gen_datatype(symbol) == result
+
+
+# Note, in isolation and when only running tests from the psyir
+# directory, this test does not fail.
+@pytest.mark.xfail(reason="issue #582 backend logging output is affected by "
```
arporter: I don't think there's a need for a new issue - just refer to #11 here as it's something that will need to be addressed when we implement logging.

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -111,6 +112,115 @@ def test_gen_dims_error(monkeypatch):
...
+    # create logicals with a precision
+    monkeypatch.setattr(symbol, "precision", 32)
+    with pytest.raises(VisitorError) as excinfo:
+        _ = gen_datatype(symbol)
+    assert ("Datatype 'logical' supports fixed precision of [1, 2, 4, 8, 16] "
+            "but found '32'." in str(excinfo.value))
+
+    # Kind not supported for character
+    symbol = Symbol("dummy", "real", precision=Symbol("c_def", "integer"))
+    # This needs to be monkeypatched as the Symbols can not
```
arporter: "the Symbol constructor cannot"?

src/psyclone/psyir/backend/fortran.py
arporter: Also, this routine feels a bit long. Is it worth creating some internal routines that handle e.g. precision=int, precision=Symbol separately?

src/psyclone/psyir/backend/fortran.py
rupertford: Oops.

src/psyclone/psyir/backend/fortran.py
rupertford: Oh, alright then. As agreed I'll keep the code for future reference, but comment it out.

src/psyclone/psyir/backend/fortran.py
rupertford: Oops again, probably cut-n-paste.

src/psyclone/psyir/backend/fortran.py
rupertford: Symbol names added and I also forgot to document the exceptions so have done this too. I've not yet looked at splitting the function.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: OK. I've changed the reference and commented out the test.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: Updated.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: I've used parametrize. The test can't be folded into the next one as this test is without the optional precision argument and the next one is with it.

src/psyclone/psyir/backend/fortran.py
rupertford: I've looked at splitting the routine and don't think it would help to be honest.

rupertford:
The broken compile example now works which is good news. I needed to change `use constants_mod` to `use constants_mod, only : r_def` in the two relevant kernel files. The reason for this is that, at the moment, use statements without an only clause do not appear in the output. An additional issue is that an `integer : r_def` variable is declared unless the `use constants_mod, only : r_def` is specified in the subroutine. For some reason the compiler does not seem to mind this, although I'm not sure we would get the expected behaviour. In lfric kind constants are declared in the module so we will need to check ancestor symbol tables, not just the parent.

rupertford:
Ready for next review from @arporter subject to travis being happy.

Issue #584, opened 2019-11-12
DataSymbols can have unknown interfaces that may be resolved later


sergisiso:
For instance in this case:
```
real a(n)
int n
```
When we parse `real a(n)`  we want to have the symbol `n` in the symbol table but we don't know it is a local symbol until we parse the next statement.

While in the cas:
```
use mod
real(kind=a) :: b
```
`a` has an unknown interface but we probably don't need to resolve it during PSyclone execution

arporter:
@sergisiso are you happy if I take this one?

sergisiso:
Sure :)

arporter:
Rupert and I have just chatted about this and have agreed that this boils down to the concept of a Symbol being allowed to have a 'deferred' interface. As a first step we can then ensure that we don't generate declarations for any such Symbols (with the assumption being that they are covered by some use statement).

arporter:
I've branched from the 544_resolve_deferred branch for now as Sergi has been doing some heavy re-structuring in there and it's not quite on master yet.

rupertford:
#591 is now on master. Closing this issue.

Issue #544, opened 2019-10-17
Mechanism to resolve the deferred type symbols


sergisiso:

arporter:
PR merged, closing issue.

PR #590, opened 2019-11-15
Support kind parameters with a deferred interface (589)

arporter:
This is a short-term fix to ensure we don't declare kind parameters locally when they are in fact imported from a module. It will be superseded by the work in in #544 and #584.


arporter:
Assuming the merge with master hasn't broken anything, this is ready for review.

arporter:
#544 is ready earlier than expected so I'm aborting this PR and will continue with #584 instead.

Issue #597, opened 2019-11-21
Can ancestor function return self?


hiker:
There is some code duplication in psyGen's _get_private_list and the Node.ancestor function, see
https://github.com/stfc/PSyclone/pull/559/files/1e32f90fa9506eb38bdc82b95ad539a3d92d82c1#r347838523

At the moment ancestor() does never return self, while the loop in _get_private_list needs to be able to get self as result.

Check if ancestor can be changed to return self. The test_node_ancestor in psyGen_test fails atm, but it is possible that nothing else breaks when changing the ancestor functions (in which case we only need to adjust the test).

rupertford:
pr #602 is on master. Closing this issue.

PR #602, opened 2019-11-26
Add include_self arg to ancestor() (for #597)

arporter:


arporter:
I couldn't just make `ancestor` include self by default so I've added an optional argument to it and tidied up the implementation a bit. Ready for first review, either from @hiker or @rupertford.

Issue #566, opened 2019-10-30
Check whether force-profile is required


rupertford:
Once #565 is on master it may not be necessary to have a force-profile command line option to the psyclone script as profile nodes are added after the optimisation script has been applied and this should work in the general case.

rupertford:
Now that automatic profiling is added after any transformations, there is no need for the --force-profile option as it should always be possible to add profile nodes to an invoke or around kernels. Changes associated with #567 ensure that the generated names are valid and unique.

rupertford:
Created branch 566_remove_force_profile

arporter:
Associated PRs have all been merged.

PR #618, opened 2019-12-09
unique profile names (#567)

rupertford:
profile names in dynamo0.3 and gocean1.0 are currently based on kernel module and subroutine names. This is confusing when profiling an invoke or a region and also does not give unique profile names when the same kernel is used in different invokes. This PR associates the names with the psy-layer subroutine name, the invoke subroutine name and the kernel name (if there is only one kernel in the profile region). This PR also ensures that the names are unique when there are multiple profile regions in an invoke, by adding indices.

PR #624, opened 2019-12-15
remove force-profile (#566)

rupertford:


rupertford:
Ready for first review from @arporter or @hiker. Note, this will clash slightly with #618 so they must be reviewed one after the other.

rupertford:
Updated as requested and fixed a couple of small errors in the documentation as well.
Ready for next review from @arporter subject to travis being happy.

PR #631, opened 2019-12-20
Support compilation of additional dependencies when testing (for 628)

arporter:
Since I touched it, I've also re-named Dynamo0p3Build to LFRicBuild as part of our gradual migration to the new naming scheme.


arporter:
Ready for first review from either @hiker, @sergisiso or @rupertford. Note that coverage of the diff is not 100% because a lot of the changes are in code that only gets executed if compilation-testing is enabled (and we don't do that on Travis).

arporter:
Thanks for tweaking and that's a good point about documentation! I've added an appropriate sub-section. Ready for next review.

Issue #474, opened 2019-08-21
File/module restructuring


hiker:
I did a quick test: beginning of the year psyGen hat around ~3900 lines, now it has ~7300 lines - any prognosis where we will be at in 2 years (47k lines assuming 1.86 fold increase in 8 months :stuck_out_tongue_winking_eye: )?

The default for pylint is 1000 lines. Some bad side effects of those long files that code duplication might get unnoticed (e.g. compare psyGen.Symbol.Access and core/access_info - basically identical classes, but now we have: 
```
>>> from  psyclone.psyGen import Symbol
>>> from psyclone.core.access_type import AccessType
>>> Symbol.Access.READ == AccessType.READ
False
```
Could we come up with a new directory structure that will allow us to reduce file size? I am not sure if this might have a performance impact (reading multiple smaller files vs reading bigger files with a lot of code that might not be actually used) - probably tbh. 

I usually prefer one class per file (with file name derived from class name ... e.g. class Node --> node.py, VariableAccess-->variable_access.py etc), only exception would be helper classes that are 'closely related' (yes, vague). Example: VariablesAccessInfo contains VariableAccessInfo and AccessInfo  as helper class, which will very unlikely ever be used by itself (except for tests).

Directory layout ... not sure, but as a suggestion:
- core Contains core classes, that might be used in many different places (e.g. access type, symbol table, errors)
- psy_ir  (to avoid camel case ... or maybe psyir?) Anything related to psyir
- gocean0p1, gocean1.0, dynamo0.3, ... - one directory per API
- transformations
- tools

That already raises a few questions:
1) do we want to move the API-directories into a common sub directory (api/gocean0.1, api/gocean1.0, ...) ? I actually don't like directories to be nested too deep (which might make it harder to find things), but I don't mind one way or another

2) should the API-specific transforms go into the one transformations directory? Or should only be the base classes in transformations, and all API-specific transforms go into the API classes?

3) Would it make sense to have paradigm-specific directories? E.g. opencl, openacc, omp, mpi, ...

And I am sure I have forgotten quite a bit :) So I leave this open to start a discussion :)

arporter:
[...]

PR #669, opened 2020-01-20
Towards #474. Isolate errors in its own submodule

sergisiso:
- Only error classes that were in `psyGen.py`
- At the moment the new file is in `src/psyclone` directory. Is this the appropriate place?


arporter:
I would suggest:

    psyclone/errors/
        __init__.py
        InternalError.py
        GenerationError.py
        ...

but maybe that's overkill and psyclone/errors.py will suffice?

sergisiso:
@arporter Yeah, I thought splitting in multiple files was too much as each Exception is just ~10 lines (and very similar ones). We can easily expand this if we ever need as from an external point of view there is no difference (e.g. `from psyclone.errors import X,Y`)

arporter:
Yes, that's true.

sergisiso:
@arporter @rupertford This is ready for the first review. It is a relatively simple PR that move PSyclone custom exception classes to a separate submodule in order to avoid circular dependencies.

src/psyclone/psyir/nodes/node.py
```diff
@@ -120,7 +121,6 @@ class Node(object):
     valid_annotations = tuple()
 
     def __init__(self, ast=None, children=None, parent=None, annotations=None):
-        from psyclone.psyGen import InternalError
         if not children:
             self._children = []
         else:
```
arporter: pylint reports an error at line 342 (undefined `Loop`).

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -43,8 +43,9 @@
 import pytest
 from psyclone.configuration import Config
 from psyclone.parse.algorithm import parse
-from psyclone.psyGen import PSyFactory, InternalError
-from psyclone.generator import GenerationError, ParseError
+from psyclone.errors import InternalError, GenerationError
+from psyclone.psyGen import PSyFactory
+from psyclone.generator import ParseError
```
arporter: Might be clearer to change this to:

    from psyclone.parse.utils import ParseError

src/psyclone/psyir/nodes/node.py
hiker: I believe it actually needs the following patch:
```
diff --git a/src/psyclone/psyir/nodes/node.py b/src/psyclone/psyir/nodes/node.py
index ef944b45..55af5282 100644
--- a/src/psyclone/psyir/nodes/node.py
+++ b/src/psyclone/psyir/nodes/node.py
@@ -265,6 +265,7 @@ class Node(object):
         try:
             graph = gv.Digraph(format=file_format)
         except ValueError:
+            from psyclone.psyGen import GenerationError
             raise GenerationError(
                 "unsupported graphviz file format '{0}' provided".
                 format(file_format))
@@ -339,6 +340,7 @@ class Node(object):
             graph.edge(remote_name, local_name, color="blue")
         # now call any children so they can add their information to
         # the graph
+        from psyclone.psyir.nodes import Loop
         if isinstance(self, Loop):
             # In case of a loop only loop at the body (the other part
             # of the tree contain start, stop, step values):

```

hiker: OK, that was stupid - I just copied the patch I had to apply on my branch, without checking if this patch had already added the import of GenerationError ... which it had. So ignore the patch, but Loop is indeed needed :blush:

src/psyclone/errors.py
```diff
@@ -0,0 +1,76 @@
...
+#         J. Henrichs, Bureau of Meteorology
+# -----------------------------------------------------------------------------
+
+''' This module provides various error classes used in PSyclone'''
+
+
+class GenerationError(Exception):
+    ''' Provides a PSyclone specific error class for errors found during PSy
+        code generation. '''
+    def __init__(self, value):
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/errors.py
```diff
@@ -0,0 +1,76 @@
...
+        Exception.__init__(self, value)
+        self.value = "Generation Error: "+value
+
+    def __str__(self):
+        return str(self.value)
+
+
+class FieldNotFoundError(Exception):
+    ''' Provides a PSyclone-specific error class when a field with the
+    requested property/ies is not found '''
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/nodes/node.py
sergisiso: I added the import but the pylint still complain about next line as self(Node).loop_body may not exist as it does not take the condition into consideration, but I think it is OK for now.

src/psyclone/tests/gocean1p0_test.py
sergisiso: Done

sergisiso:
@arporter I addressed your comments and it is ready for review again.

src/psyclone/psyir/nodes/node.py
arporter: Thanks - I agree.

PR #672, opened 2020-01-20
psyir support for lbound/ubound for 651

arporter:


arporter:
This is ready for first review from either @sergisiso or @rupertford.

src/psyclone/tests/psyir/frontend/bound_intrinsic_test.py
```diff
@@ -0,0 +1,66 @@
...
+from fparser.common.readfortran import FortranStringReader
+from psyclone.psyir.frontend.fparser2 import Fparser2Reader
+
+
+@pytest.mark.usefixtures("parser")
+@pytest.mark.parametrize("bound", ["ubound", "lbound"])
+@pytest.mark.parametrize("expression", ["n = {0}(a, 3)",
+                                        "n = {0}(a(:,:,:), 3)",
+                                        "n = {0}(a, idx1 + 3)"])
+def test_size(bound, expression):
```
sergisiso: Change the name to `test_bound_intrinsics` to avoid confusion with the size intrinsic that is not here.

src/psyclone/tests/psyir/frontend/bound_intrinsic_test.py
```diff
@@ -0,0 +1,66 @@
...
+    from fparser.two.Fortran2003 import Execution_Part
+    from psyclone.psyir.nodes import Schedule, Assignment, BinaryOperation, \
+        Reference, Literal
+    fake_parent = Schedule()
+    processor = Fparser2Reader()
+    reader = FortranStringReader(expression.format(bound))
+    fp2intrinsic = Execution_Part(reader).content[0]
+    processor.process_nodes(fake_parent, [fp2intrinsic])
+    assert isinstance(fake_parent[0], Assignment)
+    assert isinstance(fake_parent[0].rhs, BinaryOperation)
```
sergisiso: We should also assert that the `fake_parent[0].rhs.operator` is the expected value for the given `bound` parameter.

src/psyclone/tests/psyir/frontend/bound_intrinsic_test.py
```diff
@@ -0,0 +1,66 @@
...
+    assert isinstance(fake_parent[0].rhs.children[0], Reference)
+    assert isinstance(fake_parent[0].rhs.children[1],
+                      (Literal, BinaryOperation))
```
sergisiso: I tried what would happen with `"n = ubound(3, 3)"` and it return a 'BinaryOperator(Literal, Literal)'. This probably should break at some point as the [Fortran intrinsic](https://gcc.gnu.org/onlinedocs/gcc-4.7.4/gfortran/UBOUND.html) is quite explicit about the types it accepts. However, it is not obvious to me who should check the types: fparser, fparser2reader, BinaryOperator constructor? I guess from a PSyIR prespective the last one would also help when building from scratch but the current enumerator operators doesn't make this trivial.

For this PR it is enough if you add a second test with this case expecting a generic `with pytest.raises(Exception) ...` and mark it as xfail and report the issue if it is not already in the issue tracker. What do you think?

src/psyclone/tests/psyir/frontend/bound_intrinsic_test.py
arporter: Oops.

src/psyclone/tests/psyir/frontend/bound_intrinsic_test.py
arporter: Done.

src/psyclone/tests/psyir/frontend/bound_intrinsic_test.py
arporter: Yes, this problem is becoming more pressing - I noticed it with one of Rupert's PRs too (we accept any node as the 'conditional' part of an IF). I'd say that both fparser and the BinaryOperator constructor should be checking. I'll follow your suggestion and add an x-failing test.

arporter:
Assuming Travis is happy then this is ready for next review.

Issue #685, opened 2020-01-31
[PSyIR] Make nodes have a datatype


rupertford:
At the moment it is not possible to determine the datatype of PSyIR expressions, assignments etc. without walking the tree. Therefore, it is not easy to perform validity checks on the datatypes as the PSyIR is composed (which would help #677).

rupertford:
Created branch 685_node_datatype

PR #688, opened 2020-02-01
[PSyIR] Add datatype() to relevant nodes and check validity of args where appropriate (issue #685)

rupertford:


rupertford:
There is currently a discussion about separating out datatype properties from `datasymbol`. This would allow the implementation in this PR to use this datatype property rather than `datasymbol` which makes more sense. Therefore I think this PR should be put on hold for the moment. I'll add the reference to the issue when it is created but the discussion is currently on issue #696.

rupertford:
This PR is well out of date and the approach used is not correct. Closing.

PR #691, opened 2020-02-05
Fix PSyIR frontend handling of allocatable array declarations (690)

arporter:


arporter:
I found this bug while doing #500 and thought it would be better fixed separately.

arporter:
This is just a small bug fix.  Ready for first review.

Issue #706, opened 2020-02-21
[PSyData] Add/fix compilation tests for extraction


hiker:
We need simple, stand-alone dummy libraries for gocean API and for LFRic, so we can add compilation tests (which need a mod file):
- [ ] Add dummy PSyData library for gocean API.
- [ ] Add compilations tests for gocean extract case.
- [ ] Add dummy PSyData library for LFRic API.
- [ ] Add compilation tests for LFRic extraxtion (re-enable compilation in test_extract_kernel_and_builtin_dynamo0p3 and test_extract_colouring_omp_dynamo0p3)

arporter:
Do we need a dummy lib for gocean? The existing code is straightforward to compile and I think we are already in a position where we could get Travis to do this. I was going to open a new issue but then found this one...

hiker:
This ticket was for extraction. While we do have a working code for gocean, but this is based on netcdf, and I thought we would better have a simpler library (e.g. one that only prints out the calls done and variable names). It would also be useful as a kind of template to develop other libraries for gocean. But if you think that the netcdf version is fine, then by all means use it ;)

arporter:
Simpler is always good but given that you have a working netcdf version I think we might as well use it :-)

hiker:
#1813 will provide the missing libraries, and allow compilation tests. I am keeping this ticket open to remember all the tests that need to be added.

PR #708, opened 2020-02-26
638 grid properties

hiker:
This adds limited support for GOcean grid properties. Due to the missing symbol table support atm, I added code that detects a grid property access by usage of "%" in an environment variable (fld%grid%dx). In this case correct stand-alone code is created by using only the last field name ('dx') as a local variable, and using a SymbolTable to avoid name clashes.

This does not solve the issue of symbol types (i.e. still all grid properties are considered arrays). While we have limited type information for grid properties ('array', or 'scalar'), that's not enough to add a proper Fortran declaration. But this patch allows grid properties to be used the same way user fields and variables.


hiker:
Besides the comments above this patch introduces a symboltable when creating the psydata driver.

Ready for first review.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
                      funcnames=["PSyDataType"])
         prog.add(use)
 
+        # Use a symbol table to make sure all variable names are unique
+        sym_table = GOSymbolTable()
+        sym = Symbol("PSyDataType")
```
arporter: You could use `sym_table.name_from_tag("PSyDataType")` here (and again below) as that adds the new symbol to the table too. However I note #720 that it is due to be deprecated. Do @sergisiso or @rupertford have any suggestions?

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
+
+        # Variables might need to be renamed in order to guarantee unique
+        # variable names in the driver: An example of this would be if the
+        # user code contains a variable 'dx', and the kernel takes a
+        # property 'dx' as well. In the original code that is no problem,
+        # since the property is used via field%grid%dx. But the stand-alone
+        # driver renames field%grid%dx to dx, which can cause a name clash.
+        # Similar problems can exist with any user defined type, since all
+        # user defined types are rewritten to just use the field name.
+        # We use a mapping to support renaming of variables: it takes as
```
arporter: This feels like something that the SymbolTable should be doing? Maybe using the concept of 'tag' that Sergi has recently introduced.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
+                # Strip off the derived type, and only leave the last
+                # field, which is used as the local variable name.
+                local_name = var_name[last_percent+1:]
+            else:
+                # No derived type, so we can just use the
+                # variable name directly in the driver
+                local_name = var_name
+            unique_local_name = sym_table.new_symbol_name(local_name)
+            rename_variable[local_name] = unique_local_name
+            sym_table.add(Symbol(unique_local_name))
```
arporter: Again, use `name_from_tag`?

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
+            decl = DeclGen(prog, "real", [local_name], kind="8",
                            dimension=":,:", allocatable=True)
             prog.add(decl)
             is_input = var_name in input_list
             is_output = var_name in output_list
 
             if is_input and not is_output:
                 # We only need the pre-variable, and we can read
-                # it from the file (and allocate its size)
+                # it from the file (which also allocate space for it)
```
arporter: Suggest: "... (this call also allocates space for it)." without the quotes.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
                 call = CallGen(prog,
-                               "psy_data%ReadVariable(\"{0}\", {0})"
-                               .format(var_name))
+                               "{0}%ReadVariable(\"{1}\", {2})"
+                               .format(psy_data, var_name, local_name))
                 prog.add(call)
             elif is_input:
                 # Now must be input and output:
-                # First read the pre-variable
+                # First read the pre-variable (which is also allocated):
```
arporter: s/is also allocated/also allocates it/

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
                 call = CallGen(prog,
-                               "psy_data%ReadVariable(\"{0}\", {0})"
-                               .format(var_name))
+                               "{0}%ReadVariable(\"{1}\", {2})"
+                               .format(psy_data, var_name, local_name))
                 prog.add(call)
                 # Then declare the post variable, and and read its values
                 # (ReadVariable will also allocate it)
-                decl = DeclGen(prog, "real", [var_name+post_suffix],
+                decl = DeclGen(prog, "real", [local_name+post_suffix],
```
arporter: Strictly speaking, we should use the SymbolTable for the name of the post variable too (in the unlikely case that `local_name+post_suffix` clashes with something).

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
+                                       post_suffix))
                 prog.add(call)
             else:
                 # Now the variable is output only. We need to read the
-                # post variable in, and allocate the pre variable with
-                # the same size as the post
-                decl = DeclGen(prog, "real", [var_name+post_suffix],
+                # post variable in, and create and allocate a pre variable
+                # with the same size as the post
+                decl = DeclGen(prog, "real", [local_name+post_suffix],
```
arporter: Again, we should use the SymbolTable for this name.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -141,66 +142,111 @@ def create_driver(self, input_list, output_list):
...
+                               .format(psy_data, var_name, local_name,
+                                       post_suffix))
                 prog.add(call)
-                decl = DeclGen(prog, "real", [var_name], kind="8",
+                decl = DeclGen(prog, "real", [local_name], kind="8",
                                dimension=":,:", allocatable=True)
                 prog.add(decl)
                 alloc = AllocateGen(prog, [var_name,
-                                           "mold={0}".format(var_name +
+                                           "mold={0}".format(local_name +
```
arporter: I failed to spot this last time but really, AllocateGen should be modified to take a `mold` argument. I did think about whether this should be replaced by PSyIR but, in all likelihood, it's going to be a while before we support memory allocation. Please could you therefore create an issue and reference it here.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -209,24 +255,48 @@ def create_driver(self, input_list, output_list):
...
+            last_percent = prop.fortran.rfind("%")
+            if last_percent > -1:
+                # Get the last field name, which will be the
+                # local variable name
+                local_name = prop.fortran[last_percent+1:]
+                unique_name = rename_variable.get(local_name, local_name)
+                all_props[name] = api_config.make_property(unique_name,
+                                                           prop.type)
+
+        # 2) grid_data is "{0}%data" --> this just becomes {0}
```
arporter: s/grid_data/field_data/ I think? After much head scratching, is this the thing we had a lot of debate over a few PRs ago? If so, I accept it is called `grid_data` but please make the comment clear that this actually refers to data on the grid, i.e. fields.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -235,12 +305,15 @@ def create_driver(self, input_list, output_list):
         for kernel in self.psy_data_body.walk(CodedKern):
             kernel.clear_cached_data()
 
-        # Recreate the instrumented region:
+        # Recreate the instrumented region. Due to the changes in the
+        # config files, fields and properties will now become local
+        # plain arrays and variables:
```
arporter: Nice!

src/psyclone/gocean1p0.py
```diff
@@ -1586,15 +1585,17 @@ def name(self):
 
     def dereference(self, fld_name):
         '''Returns a Fortran string to dereference a grid property of the
-        specified field. E.g."name%grid%dx". The stored value of
-        self._dereference_name is a format string, where {0} represents
-        the field name.
+        specified field. It queries the current config file settings for
+        getting the proper dereference string, which is a format string
+        where {0} represents the field name.
```
arporter: Please document the `fld_name` argument to this routine.

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -144,6 +144,10 @@ def add(self, new_symbol, tag=None):
         :raises KeyError: if the symbol name is already in use.
 
         '''
+        if not isinstance(new_symbol, Symbol):
+            raise InternalError("Symbol '{0}' is not a symbol, but '{1}'.'"
+                                .format(new_symbol, type(new_symbol)))
```
arporter: If you do `type(new_symbol).__name__` then the result is independent of whether you're using Python 2 or 3.

lib/extract/dl_esm_inf/netcdf/kernel_data_netcdf.f90
```diff
@@ -72,29 +72,32 @@ module psy_data_mod
...
         procedure :: PreStart, PreEndDeclaration, PreEnd
         procedure :: PostStart, PostEnd
         procedure :: OpenRead
 
         !> The generic interface for declaring a variable:
-        generic, public :: PreDeclareVariable => DeclareScalarInt,  &
-                                                 DeclareScalarReal, &
+        generic, public :: PreDeclareVariable => DeclareScalarInt,    &
+                                                 DeclareScalarReal,   &
+                                                 DeclareScalarDouble, &
```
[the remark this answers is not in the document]
hiker: Yes, it does work, but we need to compile in order to verify this. I've added this to #706 - with a dummy library this can be tested. I could also change a declaration in example/gocean/eg6 - should I do that as well?

src/psyclone/configuration.py
```diff
@@ -861,7 +861,8 @@ def __init__(self, config, section):
         # a property, and 'type' is a string.
         # These values are taken from the psyclone config file.
         self._grid_properties = {}
-        Property = namedtuple("Property", "fortran type")
+        # Store the namedtuple type to create all properties with
```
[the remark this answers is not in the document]
hiker: Yes, I wanted to avoid calling the tuple constructor several times (not so much for performance reasons, but ... it somehow felt wrong to create the type again all the time). Anyway, am as happy as doing this all in ``make_property`` - done.

src/psyclone/configuration.py
```diff
@@ -928,10 +930,33 @@ def __init__(self, config, section):
                                          "\"{1}\".".format(key,
                                                            config.filename))
     # ---------------------------------------------------------------------
+    def make_property(self, dereference_format, type_name):
+        '''Creates a property (based on namedtuple) for a given Fortran
+        code to access a grid property, and the type.
+
+        :param str dereference_format: The Fortran code to access a property \
+            given a field name (which will be used to replace a {0} in the \
+            string with. E.g. "0}%whole%xstop").
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/configuration.py
```diff
@@ -928,10 +930,33 @@ def __init__(self, config, section):
...
+                                "'{0}'.".format(type_name))
+
+        return self._property(dereference_format, type_name)
+
+    # ---------------------------------------------------------------------
     @property
     def grid_properties(self):
         ''':returns: the dictionary containing the grid properties.
-        :type: a dictionary with values of pairs (dereference-format, type)
+        :rtype: a dictionary with values of pairs (dereference-format, type)
```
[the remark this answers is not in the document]
hiker: I made it a bit longer (since after moving the named tuple into make_property we could not easily reference a name for it).

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: I'll wait for an answer here :)

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Done.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: That's indeed better.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: I think we agreed for now to call it all grid_property :) But I added comments to explain this here.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: I think we agreed for now to keep calling it grid_property, till we have time to split this into grid- and field-properties. But I've added comments to clarify this here.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Well, this is the part I am not really proud of - it feels too much like a hack tbh :( But ... it is kind of elegant to avoid coding too much.

src/psyclone/gocean1p0.py
hiker: Done.

src/psyclone/psyir/symbols/symboltable.py
hiker: I didn't know that, that's great.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Hmm - is this good enough? That does not modify fparser, that seems to be consistent with the way fparser alloc statement is handled: you parse 'allocate(dummy)', then replace the items ('dummy' in this case) with the value you actually want.

```git diff f2pygen.py
diff --git a/src/psyclone/f2pygen.py b/src/psyclone/f2pygen.py
index 99fb01fc..43d829a3 100644
--- a/src/psyclone/f2pygen.py
+++ b/src/psyclone/f2pygen.py
@@ -793,12 +793,14 @@ def adduse(name, parent, only=False, funcnames=None):
 
 class AllocateGen(BaseGen):
     ''' Generates a Fortran allocate statement '''
-    def __init__(self, parent, content):
+    def __init__(self, parent, content, mold=None):
         '''
         :param parent: node to which to add this ALLOCATE as a child
         :type parent: :py:class:`psyclone.f2pygen.BaseGen`
         :param content: string or list of variables to allocate
         :type content: list of strings or a single string
+        :param mold: A string to be used as the 'mold' parameter of ALLOCATE.
+        :type mold: str or None.
 
         :raises RuntimeError: if `content` is not of correct type
         '''
@@ -814,6 +816,8 @@ class AllocateGen(BaseGen):
             raise RuntimeError(
                 "AllocateGen expected the content argument to be a str or"
                 " a list, but found {0}".format(type(content)))
+        if mold:
+            self._decl.items.append("mold={0}".format(mold))
         BaseGen.__init__(self, parent, self._decl)
```

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: OK, I had another look: and name_from_tag will return an existing symbol, which is not useful - I have to make sure I have a unique symbol. So I am just leaving this the way it is.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: I had a look at that, but as far as I understand the 'tag' functionality it seems to be doing something else: I could get a unique name for ``psy_data`` (which might be ``psy_data_27``), and then just use the tag ``psy_data`` to get the actual name. I don't think that helps here, since I an just simply use the symbol directory.

But I would agree that renaming variables in the symbol table would be great. Each symbol when used would use a 'renamed name' or so. But that's a lot of unrelated work (e.g. if we rename 'a' to 'b', then we can't add a symbol 'b' anymore, but we could add another 'a', ...).  The renaming I need here is very limited, and I think an stay the way it is for now.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: As above, I need to make sure unique_local_name is not an already existing name, so can't use ``name_from_tag``.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Gee, how could I miss that. This actually showed me that there is a bit of trouble still around. I can easily add code to abort if the declared name already exists, that's easy.

But when creating a test case, I realised that I could actually create a duplicated key in the netcdf file:
```
  call invoke( compute_kernel(out_fld, in_out_fld, out_fld_post, dx))
```
If ``out_fld`` is written, it will create an entry in the netcdf file ``out_fld_post`` (to store the output value). But at the same time if the field ``out_fld_post`` is an input parameter, it is stored as ``out_fld_post``. I need to think about this - either we just don't allow a name that clashes with a "..._post" name, or we need to detect the name clash by renaming one of the keys of the netcdf file, then making the driver-creation function aware of this rename as well 😢

hiker: OK, in short the problem: we have an output variable 'a', and an input variable 'a_post' --> both will create a key in the netcdf file with the name 'a_post' (which then in turn also creates duplicated variable names in the driver).

I came up with the following (neat??) solution: I added a callback from the ExtractNode after all variables have been modified. This callback can be overwritten, so in GOCeanExtractNode I check if any output-variable when added "_post" is an input variable. If so, I modify the suffix ('post' --> 'postt'), then I repeat till I have a suffix that creates no duplicated keys (appending more and more 't's).

This solves two problems:
1) It makes sure all keys in the netcdf file are unique.
2) It also implies that the variable names created in the driver are unique (since the driver uses the suffix determined above, it will use "_postt" if required).

I call it neat, since during this call we could also remove loop-variables (which we know in the domain-specific node: gocean knows the name of the loop variables, to they can be removed from the list of variables to be written) - though I will do this as a separate PR later.

I am not touching LFRic at this stage, so I have left this all in GOceanExtractNode. If we need the same for LFRic, the code can be moved into ExtractNode.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: After the change to the suffix, it should not be possible that duplicates names are created, but I obviously have still added these new names to the symbol table. The symbol table will raise an exception if an existing name is reused.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: This is the corresponding change in the code creating the Allocate statement:
```
-                alloc = AllocateGen(prog, [var_name,
-                                           "mold={0}".format(local_name +
-                                                             post_suffix)])
+                alloc = AllocateGen(prog, [var_name],
+                                    mold="{0}".format(local_name +
+                                                      post_suffix))
```
I have not added this patch in yet, let me know if this is ok enough, or if you prefer a ticket with a more thorough solution (that will require changes to fparser as far as I can see)

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Oh, and yes I still add the new name to the symbol table, so we will get an exception if something should go wrong.

hiker:
There are quite a few questions left open for now, so I'll declare this ready-for-review, so I can get clarifications on those details.

lib/extract/dl_esm_inf/netcdf/kernel_data_netcdf.f90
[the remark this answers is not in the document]
arporter: Yes please.

src/psyclone/configuration.py
```diff
@@ -945,18 +946,20 @@ def make_property(self, dereference_format, type_name):
...
+        Property = namedtuple("Property", "fortran type")
+        return Property(dereference_format, type_name)
 
     # ---------------------------------------------------------------------
     @property
     def grid_properties(self):
-        ''':returns: the dictionary containing the grid properties.
-        :rtype: a dictionary with values of pairs (dereference-format, type)
+        ''':returns: the dict containing the grid properties.
+        :rtype: a dict with values of namedtuple("Property","fortran type")
```
arporter: You're missing a line-continuation char here.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -88,6 +88,26 @@ def dag_name(self):
         '''
         return "gocean_extract_" + str(self.position)
 
+    # -------------------------------------------------------------------------
+    def update_vars_and_postname(self):
+        '''
+        This function is called after the variables to be extracted
+        have been stored in self._input_list and self._output_list.
+        This functions makes sure that when adding the postfix to
```
arporter: Suggest:

This function prevents any name clashes that can occur when adding the postfix to output variable names (e.g. if there ...input variable.) We change the ...

I think it would also be nicer if we appended an integer value rather than appending 't's.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
arporter: That's fine - that's all I was thinking of.

src/psyclone/psyir/nodes/extract_node.py
```diff
@@ -125,6 +125,16 @@ def output_list(self):
         '''
         return self._output_list
 
+    def update_vars_and_postname(self):
+        '''
+        This function is called after the variables to be extracted
+        have been stored in self._input_list and self._output_list.
+        It can be used to e.g. remove unnecessary variables (e.g. loop
+        counter), or adjust the postfix to assure that no dupliated
```
arporter: s/dupliated/duplicated/

src/psyclone/tests/test_files/gocean1p0/driver_test.f90
```diff
@@ -43,11 +43,11 @@ PROGRAM extract_example_with_various_variable_access_patterns
   use kind_params_mod
   use grid_mod
   use field_mod
-  use kernel_driver_test,  only: compute_kernel
+  use kernel_driver_test,  only: compute_kernel, compute_kernel_clash
```
arporter: `compute_kernel_clash` is not used?

src/psyclone/psyir/nodes/extract_node.py
```diff
@@ -141,6 +151,9 @@ def gen_code(self, parent):
         from psyclone.psyir.tools.dependency_tools import DependencyTools
         dep = DependencyTools()
         self._input_list, self._output_list = dep.get_in_out_parameters(self)
+
+        self.update_vars_and_postname()
```
arporter: Please add a comment here explaining the purpose of this call.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
arporter: Great.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
arporter: I'm happy with that although, as I've commented somewhere else, I think it would be nicer to append an integer rather than 't' :-)

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
arporter: I think you've done what I meant rather than what I said - using `new_symbol_name()` is the way to go.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
arporter: Yes, I think my comment was a bit ambitious! That's something for the future.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
arporter: I think `new_symbol_name()` is the right way to go.

lib/extract/dl_esm_inf/netcdf/kernel_data_netcdf.f90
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Done, including integer :)

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: OK, added, including a test-case.

src/psyclone/configuration.py
hiker: Fixed.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
hiker: Yes, I wanted to avoid this, since the loop gets 'somewhat' messier, but you are right (and it's not as messy as I thought).

src/psyclone/psyir/nodes/extract_node.py
hiker: Done.

src/psyclone/tests/test_files/gocean1p0/driver_test.f90
hiker: Oops - I originally added a separate kernel to induce the name clash, before I realised that I can reuse the existing kernel. Time to get #706 (dummy library to allow compilation) done, that should have picked that up.

src/psyclone/psyir/nodes/extract_node.py
hiker: Done.

hiker:
Ready for review.

Issue #415, opened 2019-06-24
[PSyData] Data API - Distinguishing different classes of the PSyData API


hiker:
I am using this issue to keep track of ideas and suggestion for a future 'data api'.

We observed that we have transforms and are planning transforms that have a very similar structure and behaviour:
1) The profiling interface adds starts and end callbacks (to a profiling library)
2) The PSyKE interface adds this similarly, but will also access all parameters to a kernel.
3) Visualisation (in-situ) would similarly insert start and end calls (allowing to visualise a field before and after a kernel call.

The transforms to be inserted are very similar, just the number of fields vary (0 for profiling, 1-n for visualisation, n for PSyKE ... n being the number of arguments). This leads to the following suggested interface:

1) A first call that indicates how many data fields will be handled. It takes parameters to identify the kernel, and a saved data structure (which can be used to speed up accesses):
    ``subroutine PsyDataStart(module_name, region_name, PsyDataType, n)``
    Module_name and region_name are the subroutine and kernel name, PsyDataType is an opaque data structure provided by the library that want to use the PSyData API, and 'n' is the number of fields that will be sent.
2) This is followed by 'n' calls to provide data:
  ``subroutine PSyData(PSyDataType, field, dimension, s1, n1, s2, n2, ...)``
  This provides a pointer to the raw data, the number of dimensions, and start and end index for each dimension.
For the profiling library 'n' would be 0 (i.e. only a PSyDataStart call, no PSyData), for kernel extraction this would add 'n' calls to PSyData for each argument etc. The application library using this API needs to count to identify when all PSyData calls have been executed (the idea is to keep the overhead for profiling as low as possible, i.e. to avoid a 'end of data' call, which would be useless for profiling).

3) Then the kernel is executed.

4) This is then followed by a similar sequence:
  ``subroutine PsyDataEnd(PSyDataType, m)``
5) followed by 'm' more data fields:
  ``subroutine PSyData(PSyDataType, field, ...)``

Two quick examples:
1) Profiling:
```
type(ProfileData), save :: prof_data
call PSyDataStart("module", "kernel", prof_data, 0)   ! 0 indicating that no PSyData call will follow
do ... ! the actual kernel
call PSyDataEnd(prof_data, 0)   ! 0 indicating that no PSyData call will follow
```

2) Data Visualisation:
```
type(VisualiseData), save :: vis_data
call PSyDataStart("module", "kernel", vis_data, 2)   ! 2 indicating that 2 data fields will be provided
[...]

PR #727, opened 2020-03-17
415 psydata classes

hiker:
This implemented #415 - adding a prefix to the PSyData API so that several different PSyData applications can be used at the same time.


hiker:
Am closing this, since it is included in #731 (which avoids documentation conflicts).

Issue #740, opened 2020-04-07
Remove extra quadrature_mod in LFRic infrastructure


TeranIvy:
There are two quadrature infrastructure files in `src/psyclone/tests/test_files/dynamo0p3/infrastructure`, `quadrature_mod.F90` introduced in the recent PR #701 and `quadrature_mod.f90` from before. This confuses git repo on  Windows file system as it is case insensitive, but it is not visible on Linux (that is how I missed it when reviewing #701). The old  `quadrature_mod.f90` should be removed.

@arporter, would you mind taking this on? I am happy to review the related PR (more carefully than before).

TeranIvy:
Merged #743 and checked for the presence of old file - it is not there any more. Windows git repo is not confused any more, too. Closing the issue.

PR #743, opened 2020-04-08
#740 rm old, duplicate quadrature_mod file

arporter:


arporter:
Removes the old quadrature_mod file that I'd accidentally left behind. I've checked and all tests still pass with compilation enabled. Ready for review.

Issue #766, opened 2020-05-06
Operator ordering broken in PSyiR Fortran back-end


rupertford:
The Fortran back-end will write out `a*(b+c)` as a*b+c. As suggested by @sergisiso, the simplest solution would be to add brackets around all unary, binary and nary operators. This is what he already does in the C backend.

rupertford:
Created branch 766_operator_ordering

sergisiso:
Bug fixed by PR #770

PR #770, opened 2020-05-10
Fix PSyIR Fortran backend precedence bug (issue #766)

rupertford:
The PSyIR Fortran backend will currently output
`a=b*(c+d)` as `a=b*c+d`.
This PR fixes this bug.


rupertford:
Ready for review from @sergisiso or @arporter

src/psyclone/psyir/backend/fortran.py
```diff
@@ -540,18 +620,20 @@ def binaryoperation_node(self, node):
...
+            fort_oper = get_fortran_operator(node.operator)
+            if is_fortran_intrinsic(fort_oper):
                 # This is a binary intrinsic function.
-                return "{0}({1}, {2})".format(oper.upper(), lhs, rhs)
-            return "{0} {1} {2}".format(lhs, oper, rhs)
+                return "{0}({1}, {2})".format(fort_oper, lhs, rhs)
+            if isinstance(node.parent, Operation):
+                parent_fort_oper = get_fortran_operator(node.parent.operator)
+                if not is_fortran_intrinsic(parent_fort_oper) and \
+                   precedence(fort_oper) < precedence(parent_fort_oper):
```
sergisiso, line 633: There is still a subtle error here since floating point operation are not associative we still need to care about the order of operators with the same precedence level.
For instance, at the moment:
`a + (b + c)` would be rewritten as `a + b + c`, which for FP is not right.

You can solve it using `<=` instead of `<`, but this will put unnecessary parenthesis which you are trying to avoid. A more targeted solution is to find out if this node is the lhs or rhs of the parent operation and just put the paranthesis if it is the rhs.
This should work assuming all equivalent operations have left-to-right precedende in Fortran, which I think is the case.

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -327,6 +329,84 @@ def test_gen_datatype_exception_2():
 #             "variable 'dummy'." in caplog.text)
 
 
+def test_reverse_map():
```
sergisiso, line 332: Well done adding this 2 tests that should already have been here :)

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -763,6 +843,67 @@ def test_fw_binaryoperator_unknown(fort_writer, monkeypatch):
     assert "Unexpected binary op" in str(excinfo.value)
 
 
+def test_fw_binaryoperator_precedence(fort_writer):
```
sergisiso, line 846: You may want to add the 'same precedence level but no associativity' examples in this test as well.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -231,6 +231,86 @@ def _reverse_map(op_map):
...
 
+def get_fortran_operator(operator):
+    '''Determine the Fortran operator that is equivalent to the provided
+    PSyIR operator. This is achieved by reversing the Fparser2Reader
+    maps that are used to convert from Fortran operator names to PSyIR
+    operator names.
+
+    :param operator: a PSyIR operator.
+    :type operator: \
+        :py:class:`psyclone.psyir.nodes.[Unary,Binary,Nary]Operation.Operator`
```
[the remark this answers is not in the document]
rupertford: I don't mind either so I changed it.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -231,6 +231,86 @@ def _reverse_map(op_map):
...
+
+    :returns: the Fortran operator name.
```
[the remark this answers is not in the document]
rupertford: I went for the latter.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -231,6 +231,86 @@ def _reverse_map(op_map):
...
+    :rtype: str
+
+    :raises KeyError: if the supplied operator is not known by the \
+        PSyIR.
```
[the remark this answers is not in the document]
rupertford: Yes, good point. I just removed the "by the PSyIR." bit. Hopefully that is OK.

src/psyclone/psyir/backend/fortran.py
rupertford: Yes, well spotted. I've not seen code that does this before, but we should retain the original ordering. I went for the targeted solution.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: :-)

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: Done, thanks.

rupertford:
Ready for next review subject to travis being happy.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -231,6 +232,85 @@ def _reverse_map(op_map):
...
 
 
+def get_fortran_operator(operator):
+    '''Determine the Fortran operator that is equivalent to the provided
+    PSyIR operator. This is achieved by reversing the Fparser2Reader
+    maps that are used to convert from Fortran operator names to PSyIR
+    operator names.
+
+    :param operator: a PSyIR operator.
+    :type operator: \
```
sergisiso, line 242: Remove line break now?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -231,6 +232,85 @@ def _reverse_map(op_map):
...
+    '''Determine whether the supplied Fortran operator is an intrinsic
+    Fortran function or not.
+
+    :param str fortran_operator: the supplied Fortran operator.
+
+    :returns: true if the supplied Fortran operator is a Fortran \
+        intrinsic and false otherwise.
+
+    '''
+    if fortran_operator in FORTRAN_INTRINSICS:
```
sergisiso, line 275: it can also just be `return fortran_operator in FORTRAN_INTRINSICS`

src/psyclone/psyir/backend/fortran.py
rupertford: Oops. Thanks.

src/psyclone/psyir/backend/fortran.py
rupertford: Oh yes :-).

rupertford:
OK, I've fixed the +- unary/binary precedence bug.
Ready for next review.

Issue #645, opened 2020-01-09
Should Node (and its sub-classes) take children as argument to constructor


arporter:
`Node.__init__` takes a list of children as one of its arguments and stores them under `self._children`. However, it makes no attempt to set the parent of those child nodes.

In contrast, `Loop.__init__` and `IfBlock.__init__` do *not* have `children` as an argument to their constructors. They do have static `create` methods that, given suitable children, construct a fully-populated Loop or IfBlock object.

We therefore need to agree whether or not a Node (and its subclasses) should have `children` as an argument to the constructor and if so, whether or not the constructor should also set the parent of those nodes. Searching for 'children=' in psyGen.py, it seems that the only real use of the `children` constructor argument is when making `Schedule`s, i.e. the "PSy-layer" PSyIR.

One of the main ways we construct PSyIR is through a visitor that walks down the parse tree. We therefore are often creating the parent node before we proceed to create any children. Since the visitor updates the PSyIR tree in place (rather than returning e.g. a list of child nodes) this parent has to exist before we can proceed with the walk down the tree.
        
It therefore seems likely that we should drop the `children` argument from the constructor and provide other ways of setting things up (that can also help to ensure internal consistency in parent-child relationships).

arporter:
Note that some use of the `children` constructor argument is also made in transformations.py. However, this is not extensive (~5-6 instances) and is confined to the creation of Directives.

hiker:
I totally agree to have ``set_parent`` and ``add_child`` functions, which will also clean up existing parent/child relations. The ``add_child`` function should take either a single node, or a list of nodes I'd say (though ... should it then be named ``add_children``? Or two different functions, add_children just repeatedly calling add_child?)

I am neutral if we should have a parent/child optional parameter in the constructors of all node types. It is convenient (I am thinking of transformations, which often need to insert a node to an existing hierarchy - thought we might as well add a special ``insert_node`` function for that), but a lot of work to add these parameters to all nodes, so from my point of view I am happy to consistently remove them (or add them) - but whatever we do, it should set up/clean up all parent/child relations!

arporter:
[...]

PR #773, opened 2020-05-12
Manage symbols in the Fortran parameter initializations (issues 772 and 715)

### LINKED ISSUE #772: PSyclone fails to understand literal precision in parameter initialization expressions
I tried to psyclone-generate the OpenCL version of NemoLite2D with PSyclone master and I got the following error:
```psyclone.errors.InternalError: PSyclone internal error: Failed to find a symbol table to which to add the kind symbol 'go_wp'.```

This occurs while importing external global variables declarations from a module, specifically in the expression:
```REAL(go_wp), PARAMETER :: pi    = 3.1415926535897932_go_wp```
if I remove the "_go_wp" it fails in the following line:
```REAL(go_wp), PARAMETER :: d2r   = pi / 180```
with the error:
```psyclone.psyir.symbols.symbol.SymbolError: PSyclone SymbolTable error: No Symbol found for name 'pi'.```

If I replace `pi` with its value then it works fine again.

I think the problem is that `literal_precision` tries to find the symbol_table of its parent, but parameters use a dummy node to evaluate the initialisation expression.

### PR DESCRIPTION
**Pull request by sergisiso**
Fixes #772 and maybe all or part of #715

### CODE REVIEW

**Comment by sergisiso** [comment]
@arporter @rupertford This turned out to be easier than I anticipated, and as a bonus it allowed me to rejig the InvokeSchedule init to remove a couple of #645 TODOs. It also makes the NemoLite2D OpenCL work again.

If Travis is ok with it, this is ready for review.

File: src/psyclone/psyir/frontend/fparser2.py [hunk 1]
```diff
@@ -49,7 +49,7 @@
 from psyclone.psyGen import Directive, KernelSchedule
 from psyclone.psyir.symbols import SymbolError, DataSymbol, ContainerSymbol, \
     Symbol, GlobalInterface, ArgumentInterface, UnresolvedInterface, \
-    LocalInterface, ScalarType, ArrayType, DeferredType
+    LocalInterface, ScalarType, ArrayType, DeferredType, SymbolTable
```
rupertford, line 52: SymbolTable is not used.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1164,11 +1164,15 @@ def process_declarations(self, parent, nodes, arg_list):
 
                 if initialisation:
                     if has_constant_value:
-                        # If it is a parameter parse its initialization
-                        tmp = Assignment()
+                        # If it is a parameter parse its initialization into
+                        # a dummy Assignment inside a Schedule which temporally
+                        # hijacks the parent's node symbol table
+                        tmp_sch = Schedule(symbol_table=parent.symbol_table)
```
rupertford, line 1170: This is a nice solution but the general approach of creating a dummy parent still feels hacky somehow. This complaint is obviously not for this issue.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1164,11 +1164,15 @@ def process_declarations(self, parent, nodes, arg_list):
 
                 if initialisation:
                     if has_constant_value:
-                        # If it is a parameter parse its initialization
-                        tmp = Assignment()
+                        # If it is a parameter parse its initialization into
```
rupertford, line 1167: Could you please add a blank line before the final ''' in the docstring.

src/psyclone/psyir/nodes/schedule.py
```diff
@@ -67,12 +75,11 @@ def _validate_child(position, child):
         # pylint: disable=unused-argument
         return isinstance(child, Statement)
 
-    def __init__(self, children=None, parent=None):
+    def __init__(self, children=None, parent=None, symbol_table=None):
```
rupertford, line 78: Can `children` and `parent` now be removed as optional arguments?

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -668,6 +667,16 @@ def test_process_declarations(f2008_parser):
     assert isinstance(fake_parent.symbol_table.lookup("i5").constant_value,
                       BinaryOperation)
 
+    # Initialisation with constant expresions with symbols
```
rupertford, line 670: Initialise with a constant expression (`1`) and with a symbol (`val1`).

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -668,6 +667,16 @@ def test_process_declarations(f2008_parser):
...
+    reader = FortranStringReader("integer, parameter :: val1 = 1, val2 = val1")
+    fparser2spec = Specification_Part(reader).content[0]
+    processor.process_declarations(fake_parent, [fparser2spec], [])
+    assert fake_parent.symbol_table.lookup("val1").constant_value.value == "1"
+    assert isinstance(
+        fake_parent.symbol_table.lookup("val2").constant_value, Reference)
+    assert fake_parent.symbol_table.lookup("val2").constant_value.symbol == \
+        fake_parent.symbol_table.lookup("val1")
+
```
rupertford, line 679: Please add an example with a more complex expression e.g. `2*a-b` (to show that we can deal with this). Also add a literal with precision specified `=0.0_idef` (to show that is type of code now works).

src/psyclone/tests/psyir/nodes/schedule_test.py
```diff
@@ -42,6 +42,8 @@
 import os
 import pytest
 from psyclone.psyir.nodes import Schedule, Assignment, Range
+from psyclone.psyir.nodes.node import colored, SCHEDULE_COLOUR_MAP
```
rupertford, line 45: Thanks for moving these.

src/psyclone/psyGen.py
```diff
@@ -785,15 +786,13 @@ class InvokeSchedule(Schedule):
     def __init__(self, KernFactory, BuiltInFactory, alg_calls=None,
                  reserved_names=None):
 
-        # Initialise class attributes
-        self._parent = None
-        # TODO #645 once children are not passed to the base-class constructor
-        # we should call that constructor here and have *it* create the symbol
-        # table.
-        self._symbol_table = SymbolTable()
+        # Create the Symbol Table before initialising the super-class to
```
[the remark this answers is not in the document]
sergisiso: You are quite right, in fact we could have been done before adding the symbol_table argument. I looked at this implementation before and I couldn't see this seemingly easier solution.

src/psyclone/psyGen.py
```diff
@@ -785,15 +786,13 @@ class InvokeSchedule(Schedule):
...
+        # populate it with the reserved names.
+        symbol_table = SymbolTable(self)
         if reserved_names:
             for name in reserved_names:
-                self._symbol_table.add(Symbol(name))
+                symbol_table.add(Symbol(name))
+        Schedule.__init__(self, parent=None, symbol_table=symbol_table)
```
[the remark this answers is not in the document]
sergisiso: Done, and also used `super()` notation.

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Oops, removed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: I agree, process_node() should return the created node instead of having the side-effect of appending it to the given parent, and I suspect thats a difficult change with implications everywhere process_node method is used (the appending will be responsibility of the caller ...).

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/nodes/schedule.py
sergisiso: They can be, but I would rather not, I think parent and children are a common interface to almost all nodes. If any is removed should be part of a different discussion and applied to every node.

src/psyclone/tests/psyir/frontend/fparser2_test.py
sergisiso: Done

src/psyclone/tests/psyir/frontend/fparser2_test.py
sergisiso: Added a more complex example.

sergisiso:
@rupertford Ready for next review.

src/psyclone/psyir/nodes/schedule.py
arporter: This is (partially?) the subject of #645.

rupertford: Yes. I think #645 emerged from my request and Sergi not wanting to do it. I accept the argument and drop the request.

src/psyclone/tests/psyir/nodes/schedule_test.py
```diff
@@ -51,9 +53,23 @@
     os.path.abspath(__file__)))), "test_files", "dynamo0p3")
 
 
+def test_sched_init():
+    ''' Check the Schedule class is initialised as expected.'''
```
rupertford, line 57: Seeing as we're keeping the children and parent optional arguments could you please add a test for each which checks that any provided value on initialisation is stored as expected in self.children and self.parent and that providing nothing results in self.children and self.parent being None.

sergisiso: Done

sergisiso:
@rupertford Unless Travis disagrees, this is ready for review.

Issue #679, opened 2020-01-23
[PSyIR] Make loop node use a symbol rather than a string for its loop variable.


rupertford:

rupertford:
For historical reasons the PSyIR loop node keeps its loop variable name as a string. This should be changed to a reference as that is what the rest of the PSyIR uses.

rupertford:
Created branch 679_loop_node_var_use_symbol

sergisiso:
Loop node uses a symbol now. Closing this issue.

Issue #630, opened 2019-12-19
[PSyIR] make symbol table methods aware of other symbol tables


rupertford:
At the moment the methods in a symbol table act purely on a symbol table. For example, add() will reject a symbol if the name already exists in the current symbol table but does not reject it if the name exists in a symbol table associated with a parent node. what is needed is the ability to optionally check all parent symbol tables.

rupertford:
This sort of functionality is manually implemented in the check_declared() method of the Reference class in the psyGen.py file (currently line 6045).

rupertford:
After discussing with @sergisiso and @arporter earlier today, we agreed the following way forward to support hierarchical symbol tables

1) Support finding the nearest ancestor node with a symbol table and/or finding the symbol table attached to the nearest ancestor node.

2) Support finding a symbol in any ancestor symbol table by modifying the lookup and lookup_with_tag symboltable methods so that they also search any ancestor symbol tables.

3) Add support in the Fortran back end so that all symbols are declared in the expected place. It was agreed that the symbol table associated with and module and a subroutine would ensure all relevant symbols were created.

The above three bits of functionality will get us most of the way there. There will be other issues e.g. supporting the C backend and where to place new symbols (particularly if they are UnknownType symbols), but they can be sorted out later.

Each of the above 3 bits of functionality will be separate PRs but they will all be associated with this issue. Issue #723 is also related and will need to be updated when the new functionality is added.

rupertford:
Created branch 630_search_ancestor_symbol_tables

arporter:
Leaving this open as point 3 above (https://github.com/stfc/PSyclone/issues/630#issuecomment-643265732) still needs to be addressed.

arporter:
I think this one can be closed now @sergisiso?

sergisiso:
I agree

Issue #282, opened 2019-01-28
Remove the dynamo 0.1 API


arporter:
Dynamo 0.1 has been superseded by 0.3 and has not been developed for several years.

arporter:
Once #1100 is done, I shall remove all mention of the dynamo0.1 API from the docs.

rupertford:
I thought I might join the party and remove dynamo0.1 API from the docs. Created branch [282_rm_dynamo0p1_docs](https://github.com/stfc/PSyclone/tree/282_rm_dynamo0p1_docs).

PR #785, opened 2020-06-04
Make Loop Node variable a DataSymbol (issue #679)

rupertford:


rupertford:
This is ready for first review.

codecov rightly points out that the diffs are not 100% covered. This is because I've made changes to dynamo0p1.py to code that previously had no coverage. We have agreed to remove the dynamo0p1 api - see issue #282 - and are therefore not worrying about coverage issues in this file.

The only other area of contention that I am aware of is that the checking and adding of variables to the symbol table in the PSy-layer is flaky. Fixing this is the subject of #630 and I have referenced this issue in the code where this needs fixing.

rupertford:
It probably makes sense for @arporter or @sergisiso to review this.

src/psyclone/tests/nemo/nemo_test.py
```diff
@@ -0,0 +1,64 @@
...
+'''Module containing pytest tests for the nemo.py file'''
+
+from __future__ import absolute_import
+import pytest
+from fparser.common.readfortran import FortranStringReader
+from psyclone.nemo import NemoFparser2Reader
+from psyclone.errors import InternalError
+
+
+def test_loop_var_exception(parser):
```
sergisiso, line 45: This test may not be needed anymore if this is the same that fparser2 front-end does by default.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -59,8 +59,10 @@ class Loop(Statement):
 
     :param parent: parent of this node in the PSyIR.
     :type parent: sub-class of :py:class:`psyclone.psyir.nodes.Node`
-    :param str variable_name: optional name of the loop iterator \
-        variable. Defaults to an empty string.
+    :param variable: optional reference to the loop iterator \
+        variable. Defaults to None.
+    :type variable: :py:class:`psyclone.psyir.nodes.Reference` or \
```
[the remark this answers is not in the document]
sergisiso: If this changes, some other docstrings in a few places need to change accordingly.

src/psyclone/dynamo0p1.py
```diff
@@ -233,11 +233,28 @@ def __init__(self, parent=None, loop_type=""):
...
         elif self._loop_type == "colour":
-            self._variable_name = "cell"
+            tag = "colour_loop_idx"
+            root_name = "cell"
         else:
-            self._variable_name = "cell"
+            tag = "cell_loop_idx"
+            root_name = "cell"
+
+        # This will return the symbol table from the closest ancestor
```
sergisiso, line 245: This looks a bit concerning but I guess it's better than the hardcoded version that it is replacing. Maybe how to solve this issue could be a topic for one of the next design calls.

src/psyclone/dynamo0p1.py
```diff
@@ -233,11 +233,28 @@ def __init__(self, parent=None, loop_type=""):
...
+        # up declaring the variable twice. Issue #630 describes this
+        # problem.
+        symtab = self.find_symbol_table()
+        try:
+            data_symbol = symtab.lookup_with_tag(tag)
+        except KeyError:
+            name = symtab.new_symbol_name(suggested_name)
+            data_symbol = DataSymbol(name, INTEGER_TYPE)
+            symtab.add(data_symbol, tag=tag)
+        self.variable = Reference(data_symbol)
```
sergisiso, line 257: Remove Reference()

src/psyclone/dynamo0p3.py
```diff
@@ -6701,24 +6701,34 @@ def __init__(self, parent=None, loop_type=""):
...
+        # up declaring the variable twice. Issue #630 describes this
+        # problem.
+        symtab = self.find_symbol_table()
+        try:
+            data_symbol = symtab.lookup_with_tag(tag)
+        except KeyError:
+            name = symtab.new_symbol_name(suggested_name)
+            data_symbol = DataSymbol(name, INTEGER_TYPE)
+            symtab.add(data_symbol, tag=tag)
+        self.variable = Reference(data_symbol)
```
sergisiso, line 6731: Remove Reference from here and import

src/psyclone/gocean0p1.py
```diff
@@ -245,9 +245,25 @@ def __init__(self, parent=None,
...
+        # up declaring the variable twice. Issue #630 describes this
+        # problem.
+        symtab = self.find_symbol_table()
+        try:
+            data_symbol = symtab.lookup_with_tag(tag)
+        except KeyError:
+            name = symtab.new_symbol_name(suggested_name)
+            data_symbol = DataSymbol(name, INTEGER_TYPE)
+            symtab.add(data_symbol, tag=tag)
+        self.variable = Reference(data_symbol)
```
sergisiso, line 266: Remove Reference from here and import

src/psyclone/gocean1p0.py
```diff
@@ -430,14 +430,30 @@ def __init__(self, parent=None,
...
+        # up declaring the variable twice. Issue #630 describes this
+        # problem.
+        symtab = self.find_symbol_table()
+        try:
+            data_symbol = symtab.lookup_with_tag(tag)
+        except KeyError:
+            name = symtab.new_symbol_name(suggested_name)
+            data_symbol = DataSymbol(name, INTEGER_TYPE)
+            symtab.add(data_symbol, tag=tag)
+        self.variable = Reference(data_symbol)
```
sergisiso, line 455: Remove Reference from here and import

src/psyclone/nemo.py
```diff
@@ -48,7 +48,8 @@
 from psyclone.psyGen import PSy, Invokes, Invoke, InvokeSchedule, \
     InlinedKern
 from psyclone.errors import InternalError
-from psyclone.psyir.nodes import Loop, Schedule
+from psyclone.psyir.nodes import Loop, Schedule, Reference
```
sergisiso, line 51: Unused Reference and SymbolError imports

src/psyclone/psyir/nodes/loop.py
```diff
@@ -111,9 +113,42 @@ def __init__(self, parent=None, variable_name="", valid_loop_types=None,
...
+            scalar integer.
+
+        '''
+        if not variable:
+            # No variable has been supplied so there is nothing to
+            # check
+            return
+        if not isinstance(variable, DataSymbol):
+            raise GenerationError(
+                "variable argument in create method of Loop class should "
```
sergisiso, line 138: `_check_variable` is also called from the constructor and the variable setter (next comment), but the three error messages in this method point to just he create method. Can you make the 3 error message more generic? I wouldn't include the method name as it is already visible in the error stack trace.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -362,19 +395,30 @@ def kernel(self, kern):
...
+    def variable(self, var):
+        '''
+        Setter for the variable associated with this loop.
+
+        :param var: the control variable reference.
+        :type var: :py:class:`psyclone.psyir.symbols.DataSymbol`
+
         '''
-        return self._variable_name
+        self._variable = var
```
sergisiso, line 414: This should call `_check_variable` at least to avoid adding references, but also to check it is a scalar integer.

src/psyclone/tests/psyir/nodes/loop_test.py
```diff
@@ -193,39 +194,58 @@ def test_loop_create_invalid():
         Reference(DataSymbol("x", INTEGER_SINGLE_TYPE)),
         one)]
 
-    # var_name is not a string.
+    # variable is not a Reference.
```
sergisiso, line 197: s/Reference/DataSymbol/ ?

src/psyclone/psyir/nodes/loop.py
```diff
@@ -111,9 +113,42 @@ def __init__(self, parent=None, variable_name="", valid_loop_types=None,
...
+        self._variable = variable
         self._id = ""
 
+    @staticmethod
+    def _check_variable(variable):
+        '''The loop variable should be a scalar integer. Check that this is
+        the case and raise an exception if not.
+
+        :param variable: the loop iterator.
+        :type variable: :py:class:`psyclone.psyir.symbols.DataSymbol`
```
sergisiso, line 126: or NoneType, and adjust the method description docstring

src/psyclone/psyir/nodes/loop.py
```diff
@@ -59,8 +59,10 @@ class Loop(Statement):
 
     :param parent: parent of this node in the PSyIR.
     :type parent: sub-class of :py:class:`psyclone.psyir.nodes.Node`
-    :param str variable_name: optional name of the loop iterator \
-        variable. Defaults to an empty string.
+    :param variable: optional reference to the loop iterator \
+        variable. Defaults to None.
```
sergisiso, line 63: What does giving variable a None value mean here? That it is just not complete and at some point it will need to have the variable setter give a symbol before reaching the back-end, or that it is an unamed/independent variable and if the backend find a None they can create/reuse an independent variable here.

It is OK if it doesn't happen yet, I am just asking whats your view here.

src/psyclone/dynamo0p1.py
rupertford: Oops. The reason I missed these is that they are not picked up by the tests. The reason they are not picked up by the tests is that the `gen()` method is used to generate code not the Fortran back-end and this uses `f2pygen`'s `DoGen` which takes a string and does this by using `object.name`. It so happens that `Reference.name` returns Reference._symbol.name so it does not matter whether we provide a Reference to the Symbol or the Symbol when using the `gen()` method. Anyway, I've modified the code :-).

src/psyclone/dynamo0p3.py
rupertford: Done.

src/psyclone/gocean1p0.py
rupertford: Done.

src/psyclone/nemo.py
rupertford: Oops.

src/psyclone/psyir/nodes/loop.py
rupertford: Yes, quite right. I missed that.

src/psyclone/psyir/nodes/loop.py
rupertford: Good spot. I've modified the text.

src/psyclone/psyir/nodes/loop.py
rupertford: Actually I've changed the code so NoneType is no longer allowed. See my response to your next comment.

src/psyclone/psyir/nodes/loop.py
rupertford: It is there to mean that it is not complete. It allows a loop to be created before the name is known if that is required. To better reflect this I have made the `_check_variable()` method fail if the variable is `None` and no longer call `_check_variable()` in the constructor if its value is `None`. This means that the `create` and `setter` methods must now both supply a valid DataSymbol instance. Rightly or wrongly it does mean that the backends will need to check that a value has been provided.

src/psyclone/tests/psyir/nodes/loop_test.py
rupertford: Well spotted.

rupertford:
Ready for next review.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -362,19 +391,31 @@ def kernel(self, kern):
...
 
     @property
-    def variable_name(self):
+    def variable(self):
         '''
-        :returns: the name of the control variable for this loop.
-        :rtype: str
+        :returns: a reference to the control variable for this loop.
+        :rtype: :py:class:`psyclone.psyir.symbols.DataSymbol`
+        '''
```
sergisiso, line 398: I like that the create and setter methods don't accept None, but since the init still does this property can propagate a None value or usually fail with an error like name is not a valid method of NoneType(as it is often used as `loop.variable.name`). I think we need to check if it is None here and raise an error saying the Loop is still incomplete or something similar.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1677,7 +1678,14 @@ def _do_construct_handler(self, node, parent):
...
         variable_name = str(loop_var)
-        loop = self._create_loop(parent, variable_name)
+        try:
+            data_symbol = parent.find_or_create_symbol(variable_name)
+        except SymbolError:
+            raise InternalError(
+                "Loop-variable name '{0}' is not declared and there are no "
+                "unqualified use statements. This is currently unsupported."
+                "".format(variable_name))
+        loop = self._create_loop(parent, data_symbol)
```
sergisiso, line 1688: This is my fault but could you add the clarifying comment:
```
# The loop node is created with the _create_loop factory method as some
# APIs require a specialised loop node type.
```
before this line.

src/psyclone/psyir/nodes/loop.py
rupertford: Yes that's a great idea. As agreed I've added the `_check_variable` method and updated the tests. The only downside I can see is that testing whether the variable has been set will require exception handling, but we don't do that anywhere at the moment.

src/psyclone/psyir/frontend/fparser2.py
rupertford: Added.

rupertford:
Ready for next review.

PR #787, opened 2020-06-08
[PSyIR] Add a Call node (issue 768)

rupertford:


rupertford:
Ready for first review subject to travis being happy.

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -70,21 +69,12 @@ class DataSymbol(Symbol):
     '''
```
sergisiso, line 69: Fix the type interface docstring

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -70,21 +69,12 @@ class DataSymbol(Symbol):
...
                  constant_value=None, interface=None):
-        super(DataSymbol, self).__init__(name, visibility)
+        super(DataSymbol, self).__init__(name, visibility, interface)
 
         self._datatype = None
         self.datatype = datatype
-
-        # The following attributes have setter methods (with error checking)
-        self._interface = None
+        # The following attribute has a setter method (with error checking)
```
sergisiso, line 76: This is also true for datatype, so I would make this comment plural and move it to the top

src/psyclone/psyir/symbols/routinesymbol.py
```diff
@@ -0,0 +1,47 @@
...
+
+from __future__ import absolute_import
+from psyclone.psyir.symbols.symbol import Symbol
+
+
+class RoutineSymbol(Symbol):
+    '''Symbol identifying a callable routine.'''
+
+    def __str__(self):
+        return "routine(name='{0}')".format(self.name)
```
sergisiso, line 47: Can you make this consistent with the other Symbol `__str__`, which is `self.name [: <PROPERTIES>]`. I know this is not the best representation but when we do `symbol_table.view()` it gives a more tabular representation. In this case it has no properties yet.

Maybe at this point we should add the Symbol type as part of the textual representation to all symbols `self.name : Routine[<PROPERTIES>]` ? Although we can start with just this one as chaning all will need to touch a lot of tests.

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -67,6 +188,8 @@ class Symbol(object):
     :param str name: name of the symbol.
     :param visibility: the visibility of the symbol.
     :type visibility: :py:class:`psyclone.psyir.symbols.Symbol.Visibility`
+    :param interface: the interface of the symbol.
```
sergisiso, line 191: Please copy the same interface description as in the datasymbol docstring

sergisiso: Oh, and say that by default is set to Local

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -89,20 +212,35 @@ class Visibility(Enum):
...
                 "{0} 'name' attribute should be of type 'str'"
-                " but '{1}' found.".format(type(self).__name__, type(name)))
+                " but '{1}' found.".format(
+                    type(self).__name__, type(name).__name__))
         if not isinstance(visibility, Symbol.Visibility):
             raise TypeError(
                 "{0} 'visibility' attribute should be of type "
                 "psyir.symbols.Symbol.Visibility but '{1}' found.".format(
                     type(self).__name__, type(visibility).__name__))
+        if interface and not isinstance(interface, SymbolInterface):
```
sergisiso, line 227: This is already checked in the interface setter, maybe we don't need this condition here?

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -119,5 +257,68 @@ def visibility(self):
...
+    def is_argument(self):
+        '''
+        :returns: whether the Symbol has an Argument interface.
+        :rtype: bool
+
+        '''
+        return isinstance(self._interface, ArgumentInterface)
+
+    @property
+    def unresolved_interface(self):
```
sergisiso, line 315: Can we rename this property to 'is_unresolved' to be consistent with the other interface checking methods?

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -89,20 +212,35 @@ class Visibility(Enum):
     # given an explicit visibility.
     DEFAULT_VISIBILITY = Visibility.PUBLIC
 
-    def __init__(self, name, visibility=DEFAULT_VISIBILITY):
+    def __init__(self, name, visibility=DEFAULT_VISIBILITY, interface=None):
```
sergisiso, line 215: You also need to add the interface parameter to all sub-classes inits e.g. ContainerSymbol even if some combinations don't make sense for Fortran, this may need some extra checks in the backend gen_declarations.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1019,3 +1030,13 @@ def ompdirective_node(self, node):
         self._depth -= 1
         result_list.append("!${0}\n".format(node.end_string()))
         return "".join(result_list)
+
+    def call_node(self, node):
+        '''Translate the PSyIR call node to Fortran.'''
```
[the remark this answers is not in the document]
rupertford: Oops.

src/psyclone/psyir/nodes/call.py
```diff
@@ -0,0 +1,135 @@
...
+        :return: whether the given child and position are valid for this node.
+        :rtype: bool
+
+        '''
+        return isinstance(child, DataNode)
+
+    @property
+    def routine(self):
+        '''
+        :returns: the routine that this call calls.
```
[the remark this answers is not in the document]
rupertford: Yes, I agree it could be confusing. I guess we revisit it when we add in RoutineNode(Schedule) to the PSyIR (which will contain a RoutineSymbol).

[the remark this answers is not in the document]
rupertford: I've made the suggested modification to the text.

src/psyclone/psyir/symbols/datasymbol.py
rupertford: Oops again.

src/psyclone/psyir/symbols/datasymbol.py
rupertford: Well spotted.

src/psyclone/psyir/symbols/routinesymbol.py
rupertford: OK. I've used the class name i.e. `RoutineSymbol`, rather than just `Routine` and I've not changed the other symbol strings (thanks for not making me do this :-) ). If you don't like me using the class name then let me know and I'll change it. There are currently no properties so I've omitted the `<>`, however I guess we might want to add linked/not linked at some point (as we do with ContainerSymbol).

src/psyclone/psyir/symbols/symbol.py
rupertford: Updated.

src/psyclone/psyir/symbols/symbol.py
rupertford: Removed and call the setter from init.

src/psyclone/psyir/symbols/symbol.py
rupertford: I agree. I was just keeping it the same as before (as this is just a copy of the code that was in DataSymbol). Changed.

src/psyclone/psyir/symbols/symbol.py
rupertford: All symbols except ContainerSymbol already do this. For ContainerSymbol the class only allows one type of interface. I've therefore omitted to add interface as an option, but instead pass the fixed value to the Symbol base class. This seems sensible to me seeing as the class explicitly forbids other options at the moment. I suggest we allow other options when more than one is valid. However, I will change it if you insist :-)

rupertford:
I had to make some changes to containersymbol.py after merging with the latest version of master. This was to do with ContainerInterface not being a SymbolInterface (which it needs to be for validation) and the interfaces being passed as classes rather than instances (which I changed). So it's probably worth checking over.

rupertford:

src/psyclone/tests/psyir/symbols/routinesymbol_test.py
```diff
@@ -0,0 +1,70 @@
...
+# -----------------------------------------------------------------------------
+
+''' Perform py.test tests on the psygen.psyir.symbols.routinesymbol file '''
+
+from __future__ import absolute_import
+import pytest
+from psyclone.psyir.symbols import RoutineSymbol, Symbol, UnresolvedInterface
+
+
+def test_routinesymbol_init():
```
sergisiso, line 44: there will be no doubt about who wrote this test :)

src/psyclone/tests/psyir/symbols/symbol_test.py
```diff
@@ -65,9 +74,146 @@ def test_symbol_initialisation():
...
+
 
 def test_symbol_str():
     '''Test that a Symbol instance can be stringified'''
 
     sym = Symbol("my_symbol")
     assert str(sym) == "my_symbol"
+
+
+# SymbolError is declared but not used in symbol.py. There are many
```
sergisiso, line 127: Bring this comment to the top, or maybe even inside the module docstring?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1019,3 +1030,21 @@ def ompdirective_node(self, node):
         self._depth -= 1
         result_list.append("!${0}\n".format(node.end_string()))
         return "".join(result_list)
+
+    def call_node(self, node):
+        '''Translate the PSyIR call node to Fortran.
+
+        :param node: a Call PSyIR node.
+        :type node: :py:class:`psyclone.psyGen.Directive`
```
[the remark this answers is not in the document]
rupertford: Oops, cut-n-paste error

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -38,6 +38,7 @@
 
 ''' This module contains the generic Symbol and the SymbolError.'''
```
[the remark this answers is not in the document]
rupertford: Oh yes, thanks.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1019,3 +1030,21 @@ def ompdirective_node(self, node):
...
+
+        :returns: the Fortran code as a string.
+        :rtype: str
+
+        '''
+
+        result_list = []
+        for child in node.children:
+            result_list.append(self._visit(child))
+        args = ",".join(result_list)
```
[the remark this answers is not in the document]
rupertford: OK, I've done that, but in Fortran there is no accepted standard that I'm aware of..

src/psyclone/tests/psyir/symbols/routinesymbol_test.py
rupertford: :-)

src/psyclone/tests/psyir/symbols/symbol_test.py
rupertford: Yes, good idea.

rupertford:
Ready for next review.

arporter:
@sergisiso I'll wait for this to go on before I merge Iva's PR.

Issue #789, opened 2020-06-10
Return kernel argument objects instead of just names from "unique_declarations"


TeranIvy:
The functions `unique_declarations` and `unique_declns_by_intent` in the base `Invoke` class in `psyGen.py` currently return the lists and dictionaries of kernel argument names with the specific Fortran intent. This functionality is only used in LFRic (Dynamo0.3) API, however it may be used for other APIs that use kernel metadata descriptors for argument types such as scalars, fields and operators.

It would be more useful, however, for the above functions to return the lists and dictionaries of kernel argument objects instead of just their names. One such use would be in LFRic API to support arguments of different primitive datatypes where that information would be read from the kernel metadata and then used by kernel argument classes (see for example #774).

PR #792, opened 2020-06-11
Return kernel argument objects instead of names from unique declarations (for #789)

TeranIvy:
This pull requests makes `unique_declarations` and `unique_declns_by_intent` in the `Invoke` class in `psyGen.py` return the lists and dictionaries of kernel argument objects instead of just names. Also, the code is updated to use `argument_type` instead of `datatype` name for API argument types to avoid confusing them with Fortran datatypes. The latter was also done for `unique_proxy_declarations` code in `DynInvoke` class.

The tests are updated accordingly, however I have not added compilation to the modified tests as that is done in PR #784 currently in review.


src/psyclone/dynamo0p3.py
```diff
@@ -3271,17 +3275,23 @@ def _invoke_declarations(self, parent):
...
+                            self._real_scalars[intent]]
+                dtype = "real"
+                parent.add(DeclGen(parent, datatype=dtype,
+                                   kind=api_config.default_kind[dtype],
+                                   entity_decls=arg_list,
                                    intent=intent))
 
         for intent in FORTRAN_INTENT_NAMES:
             if self._int_scalars[intent]:
+                arg_list = [arg.declaration_name for arg in
```
arporter, line 3288: Since it's only the names we use in these local lists (`self._real_scalars` and `self._int_scalars`) I suggest we ammend the code that sets them up so that it still only stores the names and not the Arg objects. We could also name those lists appropriately by adding the `_names` suffix, e.g. `self._real_scalar_names`.

src/psyclone/dynamo0p3.py
```diff
@@ -5014,37 +5028,43 @@ def __init__(self, alg_invocation, idx, invokes):
...
+
+        :raises GenerationError: if the supplied argument type is invalid.
         :raises InternalError: if an invalid access is specified, i.e. \
-                not of type AccessType.
+                               not of type AccessType.
+
         '''
-        if datatype not in GH_VALID_ARG_TYPE_NAMES:
+        if argument_type not in GH_VALID_ARG_TYPE_NAMES:
             raise GenerationError(
```
arporter, line 5052: Not your fault but since we're here: I think this should be an `InternalError` as it 'should never happen'.

src/psyclone/psyGen.py
```diff
@@ -542,41 +543,49 @@ def schedule(self):
...
-                                required
-        :param access: Optional AccessType that the declaration should have.
-        :returns: List of all declared names.
-        :rtype: A list of strings.
-        :raises: GenerationError if an invalid datatype is given.
-        :raises: InternalError if an invalid access is specified.
-        '''
-        if datatype not in VALID_ARG_TYPE_NAMES:
+
+        :param string argument_type: the type of the kernel argument for \
```
arporter, line 552: Not your fault but s/string/str/.

src/psyclone/psyGen.py
```diff
@@ -542,41 +543,49 @@ def schedule(self):
...
+                                     the particular API for which the intent \
+                                     is required.
+        :param access: optional AccessType that the declaration should have.
```
arporter, line 555: Again, not your fault but please add a `:type access:` entry.

src/psyclone/psyGen.py
```diff
@@ -542,41 +543,49 @@ def schedule(self):
...
+
+        :returns: a list of all declared kernel arguments.
+        :rtype: list of :py:class:`psyclone.psyGen.KernelArgument`
+
+        :raises GenerationError: if an invalid argument type is given.
+        :raises InternalError: if an invalid access is specified.
+
+        '''
+        if argument_type not in VALID_ARG_TYPE_NAMES:
             raise GenerationError(
```
arporter, line 565: Please could you turn this into an Internal error (and update the doc string appropriately).

src/psyclone/psyGen.py
```diff
@@ -589,39 +598,43 @@ def first_access(self, arg_name):
...
-    def unique_declns_by_intent(self, datatype):
+    def unique_declns_by_intent(self, argument_type):
         '''
         Returns a dictionary listing all required declarations for each
         type of intent ('inout', 'out' and 'in').
 
-        :param string datatype: the type of the kernel argument for the \
-                                particular API for which the intent is \
-                                required
+        :param string argument_type: the type of the kernel argument for the \
```
arporter, line 606: s/string/str/

src/psyclone/psyGen.py
```diff
@@ -589,39 +598,43 @@ def first_access(self, arg_name):
...
+        :rtype: dict of :py:class:`psyclone.psyGen.KernelArgument`
+
         :raises GenerationError: if the kernel argument is not a valid \
-                                 datatype for the particular API.
+                                 argument type for the particular API.
 
         '''
-        if datatype not in VALID_ARG_TYPE_NAMES:
+        if argument_type not in VALID_ARG_TYPE_NAMES:
             raise GenerationError(
```
arporter, line 619: Again, please could you make this an `InternalError`.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1679,68 +1687,98 @@ def test_dyninvoke_first_access():
...
+    list of arguments for 'gh_field' argument type. '''
     _, invoke_info = parse(os.path.join(BASE_PATH,
                                         "1.7_single_invoke_2scalar.f90"),
                            api=TEST_API)
     psy = PSyFactory(TEST_API, distributed_memory=True).create(invoke_info)
     args = psy.invokes.invoke_list[0].unique_declns_by_intent("gh_field")
-    assert args['inout'] == []
-    assert args['out'] == ['f1']
-    assert args['in'] == ['f2', 'm1', 'm2']
+    args_inout = [arg.declaration_name for arg in args['inout']]
```
arporter, line 1711: I think you can leave this first one as it was? An empty list is an empty list :-)

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1679,68 +1687,98 @@ def test_dyninvoke_first_access():
...
     psy = PSyFactory(TEST_API, distributed_memory=True).create(invoke_info)
     args = psy.invokes.invoke_list[0].unique_declns_by_intent("gh_real")
-    assert args['inout'] == []
-    assert args['out'] == []
-    assert args['in'] == ['a']
+    args_inout = [arg.declaration_name for arg in args['inout']]
+    assert args_inout == []
+    args_out = [arg.declaration_name for arg in args['out']]
+    assert args_out == []
+    args_in = [arg.declaration_name for arg in args['in']]
```
arporter, line 1731: These first two don't need to be changed as args['in'] and args['inout'] should both be empty.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1679,68 +1687,98 @@ def test_dyninvoke_first_access():
...
                            api=TEST_API)
     psy = PSyFactory(TEST_API, distributed_memory=True).create(invoke_info)
     args = psy.invokes.invoke_list[0].unique_declns_by_intent("gh_integer")
-    assert args['inout'] == []
-    assert args['out'] == []
-    assert args['in'] == ['istep']
+    args_inout = [arg.declaration_name for arg in args['inout']]
+    assert args_inout == []
+    args_out = [arg.declaration_name for arg in args['out']]
+    assert args_out == []
```
arporter, line 1746: Again, no need to change the first two.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1679,68 +1687,98 @@ def test_dyninvoke_first_access():
...
     args = psy.invokes.invoke_list[0].unique_declns_by_intent("gh_operator")
-    assert args['inout'] == []
-    assert args['out'] == ['op']
-    assert args['in'] == []
+    args_inout = [arg.declaration_name for arg in args['inout']]
+    assert args_inout == []
+    args_out = [arg.declaration_name for arg in args['out']]
+    assert args_out == ['op']
+    args_in = [arg.declaration_name for arg in args['in']]
+    assert args_in == []
```
arporter, line 1764: No need to change the first and last.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1679,68 +1687,98 @@ def test_dyninvoke_first_access():
...
+                           api=TEST_API)
+    psy = PSyFactory(TEST_API, distributed_memory=True).create(invoke_info)
+    args = psy.invokes.invoke_list[0]\
+        .unique_declns_by_intent("gh_columnwise_operator")
+    args_inout = [arg.declaration_name for arg in args['inout']]
+    assert args_inout == ['cma_opc']
+    args_out = [arg.declaration_name for arg in args['out']]
+    assert args_out == ['cma_op1']
+    args_in = [arg.declaration_name for arg in args['in']]
+    assert args_in == ['cma_op1', 'cma_opb']
```
arporter, line 1781: Unless I'm missing something, we should not have `cma_op1` listed as having both intent 'out' and intent 'in'?

src/psyclone/dynamo0p3.py
```diff
@@ -3232,17 +3234,19 @@ class DynScalarArgs(DynCollection):
 
     :raises InternalError: if an unrecognised type of scalar argument is \
                            encountered.
+
     '''
     def __init__(self, node):
         super(DynScalarArgs, self).__init__(node)
 
         if self._invoke:
+            # Get a list of Invoke scalar arguments
```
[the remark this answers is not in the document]
TeranIvy: Thanks for the suggestion, done.

src/psyclone/dynamo0p3.py
TeranIvy: Ok, I amended the `__init__` method to store only names and not the objects and I also renamed those dictionaries as suggested. Please note that this required a  `for` loop over Fotran intents for an invoke scalar arguments (similar to the one for the kernels).

TeranIvy: Ok, I amended the `__init__` method to store only names and not the objects and I also renamed those dictionaries as suggested. Please note that this required a  `for` loop over Fotran intents for an invoke scalar arguments (similar to the one for the kernels).

src/psyclone/dynamo0p3.py
TeranIvy: Agreed. Changed.

src/psyclone/psyGen.py
TeranIvy: Yep, changed.

src/psyclone/psyGen.py
TeranIvy: Heh, I knew I forgot something when updating docstrings. Added.

src/psyclone/psyGen.py
TeranIvy: Turned and updated docstring.

src/psyclone/psyGen.py
TeranIvy: Corrected.

src/psyclone/psyGen.py
TeranIvy: Made and updated docstring.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: Yep. This is now the second one (`args['out']`) due to changes in #784.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: Yep. Reverted.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: Yep. Reverted.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: Yep. Reverted.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: In the test example used here, `20.5_multi_cma_invoke.f90`, `cma_op1` is "read" arg in `columnwise_op_app_kernel_type` call and "write" arg in `columnwise_op_mul_kernel_type` call. This was about the only example I found with multiple CMA operator calls. I added comments in the test to explain this.

TeranIvy:
First review comments addressed. Back to @arporter.

src/psyclone/tests/dynamo0p3_test.py
```diff
@@ -1803,13 +1823,13 @@ def test_dyninvoke_uniq_declns_intent_cma_ops():
...
     assert args_inout == ['cma_opc']
+    # 'cma_op1' is "write" arg in columnwise_op_mul_kernel_type call
     args_out = [arg.declaration_name for arg in args['out']]
     assert args_out == ['cma_op1']
+    # 'cma_op1' is "read" arg in columnwise_op_app_kernel_type call
     args_in = [arg.declaration_name for arg in args['in']]
     assert args_in == ['cma_op1', 'cma_opb']
 
-    assert LFRicBuild(tmpdir).code_compiles(psy)
-
```
arporter, line 1795: Does this code not compile? If not, please could you reinstate this but commented out with an explanation.

src/psyclone/dynamo0p3.py
```diff
@@ -3240,25 +3240,27 @@ class DynScalarArgs(DynCollection):
...
+            for intent in FORTRAN_INTENT_NAMES:
+                self._real_scalar_names[intent] = [arg.declaration_name for arg
+                                                   in real_scalars[intent]]
+                self._int_scalar_names[intent] = [arg.declaration_name for arg
+                                                  in int_scalars[intent]]
         else:
-            # We have a Kernel stub, so create a list of scalar arguments.
-            self._real_scalars = {}
-            self._int_scalars = {}
             for intent in FORTRAN_INTENT_NAMES:
```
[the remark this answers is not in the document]
TeranIvy: It was moved there. Are you talking about the `for` loop by Fortran intents where the values are initialised to empty lists? If so, I moved that as well, just below the dict initialisation. It seems to be needing both for the kernel stub.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: This compilation line is for `test_dyninvoke_uniq_declns_intent_ops(tmpdir):` test. `test_dyninvoke_uniq_declns_intent_cma_ops():` did not have compilation as there is already a compilation check for "20.5_multi_cma_invoke.f90" in the `dynamo0p3_cma_test.py`. I added it here, though, to be on the safe side.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: Yes, you're right and I think the problem is deeper than that. We have 
```
        # sum_args behave as if they are write_args from
        # the PSy-layer's perspective.
        write_args += sum_args
        # readwrite_args behave in the same way as inc_args
        # from the perspective of first access and intents
        inc_args += readwrite_args
```
so byt his logic checks for "write" args should reject anything else than `AccessType.WRITE, AccessType.SUM`, whereas checks for "inc" should reject anything else than `AccessType.INC, AccessType.READWRITE`. However, now we have
```
        for arg in inc_args:
            ...
            first_arg = self.first_access(arg.declaration_name)
            if first_arg.access != AccessType.WRITE:
                if arg not in declns["inout"]:
                    declns["inout"].append(arg)
            ...
```
and
```
        for arg in write_args:
            ...
            first_arg = self.first_access(arg.declaration_name)
            if first_arg.access == AccessType.READ:
                if arg not in declns["inout"]:
                    declns["inout"].append(arg)
            ...
```
No wonder Coverage shows red for these lines. I am happy to create an issue for this.

TeranIvy: Yes, you're right and I think we need to revisit the logic there. We have
```
        # sum_args behave as if they are write_args from
        # the PSy-layer's perspective.
        write_args += sum_args
        # readwrite_args behave in the same way as inc_args
        # from the perspective of first access and intents
        inc_args += readwrite_args
```
There are first checks for "inc" and "readwrite" args 
```
        for arg in inc_args:
            ...
            first_arg = self.first_access(arg.declaration_name)
            if first_arg.access != AccessType.WRITE:
                if arg not in declns["inout"]:
                    declns["inout"].append(arg)
            ...
```
and then for pure writers
```
        for arg in write_args:
            # For every argument that is written to by at least one kernel,
            # identify the type of the first access - if it is read
            # or inc'd before it is written then it must have intent(inout).
            # However, we deal with inc and readwrite args separately so we
            # do not consider those here.
            first_arg = self.first_access(arg.declaration_name)
            if first_arg.access == AccessType.READ:
                if arg not in declns["inout"]:
                    declns["inout"].append(arg)
```
However, I think "However, we deal with inc and readwrite args separately so we do not consider those here" is not fool-proof as there is a possibility that a "write" arg can have a  "read"-like first access access other than pure "READ"'. I tested this hypothesis with
```
        for arg in write_args:
            first_arg = self.first_access(arg.declaration_name)
            if first_arg.access in (AccessType.READ, AccessType.READWRITE, AccessType.INC):
                if arg not in declns["inout"]:
                    declns["inout"].append(arg)
```
Now 'cma_op1' is correctly placed in `inout` and `in` lists. 

I created an issue and referred it is as TODO in `psyGen` and this test.

TeranIvy:
Addressed the comments from 2nd review. I think the logic for declaring "write" args intents doesn't account for all possible "READ"-like accesses (see comment [above](https://github.com/stfc/PSyclone/pull/792#discussion_r442143446) and I am happy to create an issue for this.

Travis permitting, this can now go back to @arporter.

src/psyclone/tests/dynamo0p3_test.py
arporter: Yes please.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
arporter: Oh yes. I was thinking that we could get rid of the `self._real/int_scalars = {}` but of course we can't. Apologies but would you therefore mind undoing this last change and putting the loop that initialises the various dict keys back so that it's only done for the kernel stub.

src/psyclone/tests/dynamo0p3_test.py
TeranIvy: Created the issue and referred it accordingly.

src/psyclone/dynamo0p3.py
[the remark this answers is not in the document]
TeranIvy: Nope, we can't get rid of that, I tried :) Undone the last change and put the key initialisation back.

TeranIvy:
Addressed review comments and merged the latest master. Back to @arporter.

arporter:
All good now. (The two un-covered lines were not covered before and will be addressed as part of #801.)

PR #795, opened 2020-06-12
add support for finding the closest ancestor Schedule or Container Node (closes 793)

rupertford:


rupertford:
Ready for review

src/psyclone/psyir/nodes/node.py
```diff
@@ -1077,21 +1077,27 @@ def _insert_schedule(self, children=None, ast=None):
...
+        :raises InternalError: if there is no Schedule or Container \
+            ancestor.
 
         '''
+        from psyclone.psyir.nodes import Schedule, Container
         current = self
-        while current and not hasattr(current, "symbol_table"):
+        while current and not isinstance(current, (Container, Schedule)):
             current = current.parent
         if current:
```
sergisiso, line 1097: Since we agreed to be more explicit when checking for None, would this be `if current is not None:`?

src/psyclone/psyir/nodes/node.py
```diff
@@ -1077,21 +1077,27 @@ def _insert_schedule(self, children=None, ast=None):
...
-            return current.symbol_table
-        raise InternalError("Symbol table not found in any ancestor nodes.")
+            return current
+        raise InternalError(
+            "None of this node's ancestors are Container or Schedule nodes.")
```
sergisiso, line 1100: I imagine scope will usually be used together with other properties/methods in the same line. So maybe it will be helpful to me more explicit that is the scope property that failed? Something like `"Could't find the scope of '{0}' as none of its ancestors are Container or Schedule nodes.".format(self)`

src/psyclone/psyir/nodes/node.py
rupertford: I've restructured to remove the "if current" test. However the while loop is still `while current`. Is that OK?

src/psyclone/psyir/nodes/node.py
rupertford: Yes, that is nicer, thanks.

rupertford:
Ready for next review.

Issue #783, opened 2020-06-01
Duplicate orientation declarations for operators in PSy layer


TeranIvy:
Whilst adding compilation testing for modified test kerenels and examples in #764 I noticed that test example `4.3_multikernel_invokes.f90` (`test_multikern_invoke_orient` in `dynamo0p3_test.py`) fails to compile as it generates two identical declarations of for orientation in PSy layer when calling the same kernel twice:

'''
INTEGER(KIND=i_def), pointer :: orientation_w2(:) => null(), orientation_w2(:) => null()
'''

This happens because `_invoke_declarations` function in `class DynOrientations(DynCollection)` does not check for duplicates in the declaration list.

PR #811, opened 2020-06-24
Remove duplicate declarations of orientation pointers (for #783)

TeranIvy:
This PR fixes the duplicate declarations of orientation pointers in LFRic invokes and adds compilation to the relevant test.


src/psyclone/dynamo0p3.py
```diff
@@ -2708,8 +2708,9 @@ def _invoke_declarations(self, parent):
         '''
         api_config = Config.get().api_conf("dynamo0.3")
 
-        # TODO #783: Remove duplicates from the declaration list
         declns = [orient.name+"(:) => null()" for orient in self._orients]
+        # Remove duplicate orientation pointers by using OrderedDict
+        declns = list(OrderedDict.fromkeys(declns))
```
rupertford, line 2713: I appreciate it is probably subjective but I'm not sure I like the idea of creating duplicates then removing them afterwards. It seems better to create a unique list first then use that i.e. something like ...

```
        unique_orient_names = []
        for orient in self._orients:
            if orient.name not in unique_orient_names:
                unique_orient_names.append(orient.name)
        declns = [name+"(:) => null()" for name in unique_orient_names]
```

TeranIvy: I chose this way because we have one list iterator (relatively fast) and one `OrderedDict` plus Python built-in list conversion (fast). `OrderedDict`s are one of high-performance container datatypes (e.g. https://docs.python.org/2/library/collections.html) as they are implemented in C from Python 3.5 onwards and they also preserve order. This method was recommended on several Python discussion forums as most efficient without having to `pip` install other specialised packages (e.g. [StacOverflow discussion](https://stackoverflow.com/questions/480214/how-do-you-remove-duplicates-from-a-list-whilst-preserving-order) and a [blog post on list duplicates](https://www.pythonpool.com/python-remove-duplicates-from-list/)).

TeranIvy:
Responded to @rupertford's comment. Ready for another review.

src/psyclone/dynamo0p3.py
rupertford: If we are removing duplicates from a list then I am happy with the argument. However, the proposed solution avoids this but not creating duplicates in the first place. I might be missing something but there is only one list iterator in the suggested solution as well and it does not require a conversion to ordereddict and back again to list. I, of course, accept that it takes more lines due to the if test but it is not possible to do via a list comprehension (as far as I know).

rupertford: I've taken a second opinion from @arporter and lost :-) - mainly on the grounds of succinctness, so will accept as is.

TeranIvy:
Taking the branch back for a moment to refresh it.

TeranIvy:
Branch updated, so it can go back to @rupertford for review.

PR #822, opened 2020-07-02
Exclude Return nodes from profiling regions (closes #820)

### LINKED ISSUE #493: Make omp parallel check for invalid node types, not for valid ones
As discussed in #478 the omp parallel transform only allows a small set of node types:
```
                        psyGen.Literal, psyGen.Reference)
```

As discussed with @rupertford (and iirc @arporter suggested the same in the past), this should be replaced with a list of invalid node types, to allow more flexibility.

Suggested invalid node types:
- halo exchange
- subroutine call
- code block

Note that #478 will introduce a mechanism to disable/alter the tests if required.

### LINKED ISSUE #820: [PSyData] RETURN statement can be enclosed within PSyData region
In the NEMO API it is possible to enclose a RETURN within a PSyData region:

    CALL profile_psy_data0 % PreStart('sw_ptmp', 'r0', 0, 0)
    del_P = PR - P
    del_th = del_P * sw_adtg(S, T, P)
    ...
    sw_ptmp = th + (del_th - two * q) / (two * three)
    RETURN
    CALL profile_psy_data0 % PostEnd

which is likely to cause us problems!

arporter:
This PR fixes #820. It also changes the way transformation validation is done so that each transformation now has a list of node types that must be *excluded* instead of a list of nodes that are permitted. As such, it also fixes #493 although I think we have to permit Subroutine calls as that's precisely what both the GOcean and LFRic APIs are designed to do.

### DIFF
--- a/changelog
+++ b/changelog
@@ -40,6 +40,11 @@
 	16) #808 for #800. Adds support in the PSyIR for the case when a
 	variable used to dimension an array has deferred or unknown type.
 
+	17) #822 for #820. Excludes return statements from profile and
+	directive regions. Also reworks the relevant code to exclude nodes
+	rather than include them which makes more sense as exclusion is
+	the exception.
+
 release 1.9.0 20th May 2020
 
 	1) #602 for #597. Modify Node.ancestor() to optionally include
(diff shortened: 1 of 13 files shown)


arporter:
Despite what GitHub is saying, Travis has completed and is green. Therefore this is ready for first review, probably by either @hiker or @rupertford.

PR #837, opened 2020-07-10
Add start/stop support to nvidia profiling wrapper (closes #824)

### LINKED ISSUE #824: Add PSyData_Start and PSyData_Stop support.
It's often necessary to be able to reduce a profiling trace to a manageable size by only enabling profiling for e.g. a single time-step of a model run. For nvtx this requires calls to CUDA routines.

The Profiling API already has init() and shutdown() calls (which are currently unused in the nvtx wrapper). I propose to add the enable/disable calls into these wrapper routines although I'm slightly uneasy that I may be changing their expected semantics by doing that.

What do @hiker and @rupertford think?

arporter:


arporter:
This is ready for first review :-)

hiker:
No tests affected, no change to user guide. Dev guide builds well, new code compiles (I cannot fully test functionality as I don't have cuda access atm, but trust that @arporter has tested it). Changelog updated.

hiker:
OK, forcing a merge now (since no code that is tested was changed) - the tests are not starting (gee, you guys on the other side have to wait forever to get something to be tested, for me it typically starts immediately ;) ).

arporter:
> 
> OK, forcing a merge now (since no code that is tested was changed) - the tests are not starting (gee, you guys on the other side have to wait forever to get something to be tested, for me it typically starts immediately ;) ).

It's OK up until lunchtime and then I think America wakes up :-(

Issue #860, opened 2020-07-23
Remove the usage of links in documentation


hiker:
... since this doesn't work well with windows filesystems. The links are used to point to the references and bibliography files.

rupertford:
PR #862 has been merged to master. Closing this issue.

PR #862, opened 2020-07-23
#860 Replace links with sphinx-includes, and changed reference file

hiker:
to work from all doc directories. Problem with links happens when using a windows FS mounted on a linux VM (--> links do not work).


hiker:
I had two add two more changes here. I noticed that now and again creating the docs would fail with an error related to psyGen.PSy.inline (I never manged to reproduce this, even after ``make clean`` it would only show up now and again). Turns out PSy.inline had typos in the doc string, so I updated all the documentation for this class (``gen`` also documented an argument that does not exist anymore - neither here now in any derived class).

Then when I made sure I didn't break anything, I tested the inline related transformations, and one test would fail. Turns out that this test did not specify to use distributed memory, so depending on what test was run before the test would not use DM and then fail (and using ``-k test_module_inline`` caused the two tests to be run in the wrong order).

Seemed silly to open a new ticket and PR for these very small things, so I've just added this here.

doc/bibliography/zz_bibliography.rst
```diff
@@ -9,4 +9,4 @@
    contain citations then those citations aren't handled correctly.
    We therefore put a "zz" at the start of the name of the current file
    to ensure it is processed last.
-.. bibliography:: references.bib
+.. bibliography:: ../bibliography/references.bib
```
rupertford, line 12: Add a note saying that this file is included from other locations and the path is relative to those locations, not this one.

src/psyclone/psyGen.py
```diff
@@ -265,24 +265,35 @@ def __str__(self):
...
     @property
     def invokes(self):
+        ''':returns: the list of invokes.
+        :rtype: :py:class:`psyclone.psyGen.Invokes` or derived class
+        '''
         return self._invokes
 
     @property
     def name(self):
+        ''':returns: The name of the PSy object.
```
rupertford, line 275: the not The please.

src/psyclone/psyGen.py
```diff
@@ -265,24 +265,35 @@ def __str__(self):
...
         return "psy_"+self._name
 
     @property
     @abc.abstractmethod
     def gen(self):
         '''Abstract base class for code generation function.
-        :param parent: the parent of this Node in the PSyIR.
-        :type parent: :py:class:`psyclone.psyir.nodes.Node`.
+
+        :return: Root node of generated Fortran AST
```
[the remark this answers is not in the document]
rupertford: And also Root -> root

src/psyclone/psyGen.py
```diff
@@ -265,24 +265,35 @@ def __str__(self):
...
+        :rtype: :py:class:`psyir.nodes.Node`
```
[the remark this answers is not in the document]
hiker: Oops - one of my more embarrassing PRs all in all.

src/psyclone/psyGen.py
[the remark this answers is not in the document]
hiker: Ha, yes, consistency - it always gets me. Return or returns always depends what my editor finds first :)  OK, a count shows >500 uses of :returns:, and around 120:return:, but the latter is documented in the coding style. So I've updated the coding style to use :returns:, and consistently used :returns:.

doc/bibliography/zz_bibliography.rst
hiker: Done.

src/psyclone/psyGen.py
hiker: One day I will understand capitalisation after : ... I promise.

hiker:
ok, all done.

Issue #876, opened 2020-08-27
Node.find_or_create_symbol() creates a DataSymbol if no match is found. 


arporter:
`Node.find_or_create_symbol()` creates a new Symbol if no match is found (and there are one or more Containers from which it might be imported). However, it currently always creates a `DataSymbol`, even though we don't actually know whether it should be a `DataSymbol` or a `RoutineSymbol`. Possibly it should just create a `Symbol` in this case.

arporter:
I'm doing this before #880/#363 as that issue needs this functionality.

PR #880, opened 2020-09-03
Support for derived types (closes #363)

### LINKED ISSUE #363: Represent Fortran derived types in the PSyIR

arporter:


arporter:
I'm pausing this PR as it's getting very large. I'll do the creation of a `Symbol` (rather than a `DataSymbol`) for symbols of unknown type in #876 first (and the resulting changes to the SymbolTable.lookup() method).

arporter:
Closing this as I now have a better version in #911.

PR #882, opened 2020-09-07
Remove pylint required version (closes #881)

### LINKED ISSUE #881: pip fails when installing the [test] dependencies
On a new installation in Ubuntu 20.04, when doing a:
``` pip install -e .[test]```
I get the error `ERROR: pytest-pylint 0.17.0 has requirement pylint>=2.3.0, but you'll have pylint 1.6.5 which is incompatible.`

This is because `setup.py` requests a specific version of pylint `"pylint==1.6.5"`, if I remove the specific version or the whole pylint dependency and let it get it as part of pytest-pylint the problem is solved and everything seems to work fine.

I couldn't find any issue regarding why pylint is locked to a specific version. Does anybody know? Is it safe to unlock it now?

sergisiso:
Fixes #811

### DIFF
--- a/changelog
+++ b/changelog
@@ -93,6 +93,8 @@
 	code but not input code. The various options can be controlled by
 	command line switches.
 
+	32) #882 for #881 Removes specific version of pylint in setup.py.
+
 release 1.9.0 20th May 2020
 
 	1) #602 for #597. Modify Node.ancestor() to optionally include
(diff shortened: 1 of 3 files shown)


sergisiso:
@rupertford @arporter This is ready for review, is a simple change in the `setup.py` to remove the pylint version requirement.

rupertford (changes requested):
Simple change to remove version dependency of pylint in setup.py. Could you please update the documentation in developer_guide/system_specific_setup.rst. I suggest warning people about certain versions of pylint 1.7

sergisiso:
@rupertford I modified the warning phrasing, is this sufficient?

rupertford (approved):
Changes look good. Approving.

PR #887, opened 2020-09-08
Create Symbol for unknown symbols (closes #876)

### LINKED ISSUE #876: Node.find_or_create_symbol() creates a DataSymbol if no match is found.

arporter:


arporter:
Missed coverage showed that we had duplicated the functionality of generating a Symbol for a kind parameter. I've now removed this duplication. We also had `fparser2._get_symbol_table()` which has been superseded by Rupert's new `node.scope.symbol_table` functionality so I've removed that too.

arporter:
Functionality wise, this is a small change to use `Symbol` rather than `DataSymbol` when we encounter a symbol of unknown type. However, this then required generalisation of some functionality (esp. the `copy` method) and I've removed some code duplication.
Ready for first review from either @sergisiso or @rupertford.

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -244,6 +244,98 @@ def __init__(self, name, visibility=DEFAULT_VISIBILITY, interface=None):
...
+
+        :returns: A symbol object with the same properties as this \
+                  symbol object.
+        :rtype: :py:class:`psyclone.psyir.symbols.Symbol`
+
+        '''
+        return self._base_copy([], {}, interface=interface,
+                               visibility=visibility)
+
+    def _base_copy(self, args, kwargs, interface=None, visibility=None):
```
sergisiso, line 267: Again, I need a bit of context about why this is needed instead of the original implementation.

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -277,19 +278,28 @@ def __str__(self):
             ret += ", constant_value={0}".format(self.constant_value)
         return ret + ">"
 
-    def copy(self):
+    def copy(self, interface=None, visibility=None):
```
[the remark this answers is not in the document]
sergisiso: I do understand adding a `visibility=self.visibility` that is missing inside the copy.

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -323,5 +415,40 @@ def is_unresolved(self):
         '''
         return isinstance(self._interface, UnresolvedInterface)
 
+    def find_symbol_table(self, node):
```
sergisiso, line 418: The implementation looks good but looking where it is used I think it will benefit from a more descriptive name. Maybe something like
`find_[belonging|residing_in]_symbol_table` but I am not sure. What do you think?

src/psyclone/transformations.py
```diff
@@ -3682,49 +3682,54 @@ def apply(self, node, options=None):
...
 
             # Resolve the data type information if it is not available
-            if isinstance(globalvar.datatype, DeferredType):
-                globalvar.resolve_deferred()
+            if (type(globalvar) == Symbol or
+                    isinstance(globalvar.datatype, DeferredType)):
+                updated_sym = globalvar.resolve_deferred()
+                # If we have a new symbol then we must update the symbol table
+                if updated_sym is not globalvar:
+                    kernel.symbol_table.remove(globalvar)
```
sergisiso, line 3690: Use the proposed `swap` method here?

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -116,6 +116,7 @@ def resolve_deferred(self):
                     "Error trying to resolve symbol '{0}' properties, the lazy"
                     " evaluation of '{1}' interfaces is not supported."
                     "".format(self.name, self.interface))
+        return self
```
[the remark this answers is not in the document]
sergisiso: Is (part of) the interior of the `if isinstance(self.datatype, DeferredType):` condition the same than the generic symbol resolve_interface? can it call the super().resolve_deferred() at that point?

sergisiso (changes requested):
@arporter I made a first review of the high-level implementation aspects, I made some suggestions about the interface, but in general I like the simplifications of code that this PR brings (although I need more context regarding the changes of the symbol copy method).

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -39,6 +39,7 @@
     to transform each node into the equivalent PSyIR representation.'''
 
 from __future__ import absolute_import
+import six
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -315,6 +288,82 @@ def _is_range_full_extent(my_range):
...
 
 
+def _kind_symbol_from_name(name, symbol_table):
+    '''
+    Utility method that returns a Symbol representing the named KIND
+    parameter. If the supplied Symbol Table (or one of its ancestors)
+    does not contain an appropriate entry then one is created. If it does
+    contain a matching entry then it must be either a Symbol or a
+    DataSymbol. If it is a DataSymbol then it must have a datatype of
+    'integer' or 'deferred'. If the latter then the fact that we now know
```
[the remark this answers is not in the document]
arporter: Changed.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -315,6 +288,82 @@ def _is_range_full_extent(my_range):
...
+            # so we need to replace it with a DataSymbol of integer type.
+            # Since the lookup() above looks through *all* ancestor symbol
+            # tables, we have to find precisely which table the existing
+            # Symbol is in.
+            table = kind_symbol.find_symbol_table(symbol_table.node)
+            new_symbol = DataSymbol(lower_name,
+                                    default_integer_type(),
+                                    visibility=kind_symbol.visibility,
+                                    interface=kind_symbol.interface)
+            table.remove(kind_symbol)
```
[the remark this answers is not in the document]
arporter: Created #898

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1081,7 +1121,7 @@ def _process_use_stmts(parent, nodes):
                         "({1}). This is invalid Fortran.".format(
                             mod_name, str(container)))
 
-            # Create a 'deferred' symbol for each element in the ONLY clause.
+            # Create a basic Symbol for each element in the ONLY clause.
```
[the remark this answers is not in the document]
arporter: Oh yes, thanks.

src/psyclone/psyir/symbols/datasymbol.py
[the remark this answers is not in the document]
arporter: Oh yes, I vaguely remember thinking I needed to try and reduce duplication here but then forgot. I've added `Symbol.get_external_symbol()` which deals with this.

src/psyclone/psyir/symbols/datasymbol.py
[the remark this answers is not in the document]
arporter: I agree that it does seem over engineered. I ended up doing this because the `visibility` of a Symbol is currently immutable and therefore it needs to be supplied to the constructor (called from within `copy`). I then ended up with code duplication and therefore took this approach to remove it - the common code is now in `_base_copy()`. It was complicated because the constructors of DataSymbol and Symbol have different, required arguments. I think we have three choices:
1. allow the visibility to be mutable;
2. we live with the duplication of the 'do we override the visibility?' code in each of the two copy routines;
3. what I currently have.

(The duplication in 2. is less then I first through because we could just set the new interface after the call to copy(), as you suggest.) I think my preferred option is 2. but I'm open to persuasion :-)

src/psyclone/psyir/symbols/symbol.py
arporter: I thing `symbol.find_symbol_table()` is reasonably descriptive already. Maybe `symbol.containing_symbol_table()`? We could open the vote out to @rupertford?

src/psyclone/transformations.py
arporter: Yes, good idea.

arporter:
I've done most of the changes but there are still two things outstanding (how we deal with wanting to set the visibility of the new symbol created by the `copy()` method and what to call the `Symbol.find_symbol_table()` method). Hopefully Travis will be happy now but if not I'll fix it while we discuss these two things...

src/psyclone/psyir/symbols/datasymbol.py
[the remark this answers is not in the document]
sergisiso: That looks much better now (both here and the generic), well done.

src/psyclone/psyir/symbols/datasymbol.py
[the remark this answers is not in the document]
sergisiso: I think number 1) would be the easier. Is there any reason for Visibility to be immutable? If an external code links to a symbol and we change its visibility the compilation will fail at some point, but this seems an user error not necessarily something psyclone should prevent.

I think my problem is that then we introduced this method I was thinking on something akin to the OO copy constructor and we are now giving it specific functionalities that go beyond the simple copy semantics.

It also allows the implementation to be simpler:
generic copy be: `return type(self)(self.name, self.visibility, self.interface)`
DataSymbol copy be: `return DataSymbol(self.name, self.datatype, self.visibility, self.constant_value, self.interface)`
but this is a secondary consideration.

I am also open to be persuaded of the contrary.

src/psyclone/psyir/symbols/symbol.py
sergisiso: I was over-complicating things myself in this instance. After a second read I think the `symbol.find_symbol_table(node)` is probably fine.

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -81,41 +81,23 @@ def resolve_deferred(self):
...
-                # Copy all the symbol properties but the interface and
-                # visibility (the latter is determined by the current
-                # scoping unit)
-                tmp = self.interface
-                module = self.interface.container_symbol
-                try:
-                    extern_symbol = module.container.symbol_table.lookup(
-                        self.name, visibility=Symbol.Visibility.PUBLIC)
-                except KeyError:
-                    raise SymbolError(
```
sergisiso, line 102: SymbolError is now unused and the import needs to be deleted

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -244,6 +244,115 @@ def __init__(self, name, visibility=DEFAULT_VISIBILITY, interface=None):
...
+        except SymbolError as err:
+            six.raise_from(SymbolError(
+                "Error trying to resolve the properties of symbol "
+                "'{0}' in module '{1}': {2}".format(
+                    self.name, module.name, str(err.value))), err)
+
+    def resolve_deferred(self):
+        '''
+        Search for the Container in which this Symbol is defined and
+        create and return a symbol of the correct class and type. If this
```
sergisiso, line 338: As I said, I like this implementation now but since we slightly modify the functionality in a specialization of this method I would mention this in the generic docstring. For instance, by adding after the first sentence: "If the class and type are the same, some specializations of this method update the differing properties in-place."

sergisiso (changes requested):
@arporter I made some contributions to the current open discussions and added a couple more comments. I will do pylint and tests the next time.

Really the contentious thing for me was how to handle removal of symbols and this has been offloaded to a separate issue. I am not too worried about how the remaining issues are resolved.

src/psyclone/psyir/symbols/datasymbol.py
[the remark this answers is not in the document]
arporter: That's true, 1. is easiest. I was assuming that you wanted it to keep it immutable!

src/psyclone/psyir/symbols/datasymbol.py
arporter: Oh yes, thanks.

src/psyclone/psyir/symbols/symbol.py
arporter: Done.

arporter:
Assuming Travis is happy, this is ready for the next review.

arporter:
In tidying for pylint I had to turn off at least two warnings about "unidiomatic type check" in places where I had `type == Symbol`. On reflection over the weekend, I think this is telling me that I should have a new sub-class of Symbol, maybe `UnknownSymbol`, and then I can use `isinstance`. Is it OK with @sergisiso if I take this back to make this change?

sergisiso:
@arporter As you prefer, I saw the pylint warning but I was not going to flag it as I think in this case the type is well used. You may still want to do it if you think the other implementation is nicer but check this pylint issue (https://github.com/PyCQA/pylint/issues/3337) to see if an `in` or waiting for a future pylint version will solve this problem.

I will wait to review until this is done.

arporter:
Thanks for pointing to that issue Sergi. I've had a think and decided to take the path of least resistance and leave it as it is. We can always introduce an UnknownSymbol later if we feel the need.
Really ready for review now :-)

src/psyclone/tests/psyir/symbols/symbol_test.py
```diff
@@ -219,3 +223,81 @@ def test_argumentinterface_str():
...
+    asym = Symbol("a", visibility=Symbol.Visibility.PRIVATE,
+                  interface=GlobalInterface(csym))
+    new_sym = asym.copy()
+    assert new_sym is not asym
+    assert new_sym.name == asym.name
+    assert new_sym.interface == asym.interface
+    assert new_sym.visibility == asym.visibility
+
+
+def test_symbol_resolve_deferred():
```
sergisiso, line 268: Since we have now the `get_external_symbol` public method, could you split the functionality of the new method in a separate explicit test for it. Then maybe you can simplify this one with a monkeypatch for `get_external_symbol`?

Also, can you remove the duplicated parts in `datasymbol_test.py::test_symbol_resolve_deferred` and also monkeypatch that test?

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -323,5 +408,40 @@ def is_unresolved(self):
...
+        :param node: the PSyIR node from which to search.
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+
+        :returns: the SymbolTable containing this Symbol or None.
+        :rtype: :py:class:`psyclone.psyir.symbols.SymbolTable` or NoneType
+
+        :raises TypeError: if the supplied `node` argument is not a PSyIR Node.
+
+        '''
+        from psyclone.psyir.nodes import Node
```
sergisiso, line 425: Pylint prefers this import to be in the top-level.

There are similar ones in the node::scope and datasymbol::resolve_defered methods touched in this PR.

sergisiso (changes requested):
@arporter All comments regarding the implementation have been resolved. In addition I checked format, pylint and tests this time. I just found a couple of issues with these, when these are solved this should be ready to be merged.

src/psyclone/tests/psyir/symbols/symbol_test.py
arporter: Oh yes, that's much nicer now.

src/psyclone/psyir/symbols/symbol.py
arporter: Oh yes, I tried moving that one and got a circular dependence so left it where it was. In fact, that turns out to be the case for all three. I've added comments and disabled the pylint warning in each case.

arporter:
Assuming Travis is happy, this is ready for next review.

sergisiso:
@arporter I see that the issue mentioned in the title (issue 876) is still not solved with this PR, and in fact there are 2 TODOs still mentioning this issue int he code. So this is just some work needed to do towards solving the issue. Is that right?

arporter:
No, this *should* close #876. Let me take a look at the TODOs I've missed...

arporter:
Well, that was really weird - I appear to have simply not altered the routine that was supposed to be at the heart of this PR! I thought maybe I could blame a merge with master but I've not found any evidence for that. In fact 'fixing' find_or_create_symbol() actually required me to update one of the tests too so it seems unlikely that I ever did it.
Many thanks for spotting this!
Once Travis has reported this should be ready for review.

sergisiso (approved):
@arporter I removed myself one last unused symbol. This PR is now ready to be merged.

Issue #871, opened 2020-08-19
[GOcean] Provide device data synchronization methods in the GOcean PSy-layer


sergisiso:
The latest infrastructure modifications in dl_esm_inf/50 require that device data synchronization methods are provided by the PSy-layer. This affects the OpenCL and OpenACC generation in the GOcean API.

PR #895, opened 2020-09-16
Gocean OpenCL generates a read_from_device function (towards #871)

sergisiso:
To be consistent with the latest dl_esm_inf infrastructure, the GOcean OpenCL needs to generate a subroutine to specify how to read the data back from the device, and assign this subroutine to the read_from_device_f function pointer.

@hiker This PR also brings dl_esm_inf to master and adds FortCL as a dependency.


sergisiso:
@arporter This is ready for the first review, it is a small change to bring dl_esm_inf to master and use the new gocean infrastructure when generating OpenCL.

src/psyclone/gocean1p0.py
```diff
@@ -1359,16 +1359,73 @@ def gen_data_on_ocl_device(self, parent):
...
                                      rhs="clFinish(" + qlist + "(1))"))
 
+    def gen_ocl_read_from_device_function(self, f2pygen_module):
+        '''
+        Insert a f2pygen subroutine to retrieve the data back from an
+        OpenCL device using FortCL in the given f2pygen module and return
+        its name.
+
+        :param f2pygen_module: the module where the new function will be \
+                               inserted into.
```
arporter, line 1392: s/into//

arporter (changes requested):
It all looks good. It seems a shame to be adding more code that utilises f2pygen but I accept that that's the only practical solution at the moment.
All tests and examples run/compile OK.
As usual, I'm uncomfortable with the fact that I can't actually compile and run but then it struck me that perhaps we could now extend gocean/eg3 so that it has a compilation target. If that's practical then please could you put instructions in the README.md (i.e. which packages to install - it doesn't matter if these are linux-distribution specific as long as you say which distribution they are for) and add a target to the Makefile. Probably best not to use the `compile` target as this one won't work unless the necessary OpenCL packages have been installed. Perhaps `compile-ocl` instead? You'll also need to update the description of the example in the User Guide.
I get one pylint error (bad continuation) in gocean1p0_opencl_test.py and also at lines 1374 and 1419 in gocean1p0.py.

src/psyclone/gocean1p0.py
sergisiso: Done

sergisiso:
@arporter This is ready for next review

> As usual, I'm uncomfortable with the fact that I can't actually compile and run but then it struck me that perhaps we could now extend gocean/eg3 so that it has a compilation target. If that's practical then please could you put instructions in the README.md (i.e. which packages to install - it doesn't matter if these are linux-distribution specific as long as you say which distribution they are for) and add a target to the Makefile. Probably best not to use the compile target as this one won't work unless the necessary OpenCL packages have been installed. Perhaps compile-ocl instead? You'll also need to update the description of the example in the User Guide.

That's a good idea. I was testing this against PSycloneBench OpenCL generation but it is better to have some simple example in the same repository. I updated gocean/eg3 as suggested.

> I get one pylint error (bad continuation) in gocean1p0_opencl_test.py and also at lines 1374 and 1419 in gocean1p0.py.

I fixed some pycodestyle and pylint issues that I found in gocean but my pylint didn't complain about the one you mention (maybe different pylint versions?) so I am not actually sure I solved this.

arporter:
I can build and run. That feels good :-) All requested changes have been made and README and documentation have been updated. I've taken the liberty of fixing the things my pylint was complaining about.
Examples are still all OK.
Tests are all OK.
Updated documentation builds fine and looks fine.

arporter (approved):

PR #904, opened 2020-09-22
Fix bug when handling use statements (closes #903)

### LINKED ISSUE #903: [psyir FE] wildcard import followed by specific import causes failure
The following Fortran code causes the fparser2 frontend to fail:

        module test_mod
          use some_mod
        contains
          subroutine test_sub1()
            ! a_var here must be being brought into scope by the
            ! `use some_mod` in the module.
            a_var = 1.0
          end subroutine test_sub1
          subroutine test_sub2()
            use some_mod, only: a_var
            a_var = 2.0
          end subroutine test_sub2
        end module test_mod

with:

    E           KeyError: "Symbol table already contains a symbol with name 'a_var'."

This prevents the current master of PSyclone from processing (some of) the NEMO source.

arporter:


arporter:
This is a one-line fix plus as an associated new test. Ready for review.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1188,14 +1188,18 @@ def _process_use_stmts(parent, nodes):
                     # will replace a previous import with an empty only-list.
                     pass
                 for name in decl.items[4].items:
-                    # The DataSymbol adds itself to the list of symbols
-                    # imported by the Container referenced in the
-                    # GlobalInterface.
                     sym_name = str(name).lower()
                     if sym_name not in parent.symbol_table:
+                        # We're dealing with a symbol named in a use statement
+                        # in the *current* scope therefore we don't need to
```
rupertford, line 1194: s/don't need to check any/should not check/

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -2678,3 +2678,40 @@ def test_loop_var_exception(parser):
...
+          end subroutine test_sub2
+        end module test_mod
+        ''')
+    prog = f2008_parser(reader)
+    psy = PSyFactory(api="nemo").create(prog)
+    # We should have an entry for "a_var" in the Container symbol table
+    # due to the access in "test_sub1".
+    container = psy.invokes.container
+    avar1 = container.symbol_table.lookup("a_var")
+    assert type(avar1) == Symbol
```
rupertford, line 2707: pylint asks you to use isinstance instead
psyir/frontend/fparser2_test.py:2707:11: C0123: Using type() instead of isinstance() for a typecheck. (unidiomatic-typecheck)

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -2678,3 +2678,40 @@ def test_loop_var_exception(parser):
...
+    # There should be no entry for "a_var" in the symbol table for the
+    # "test_sub1" routine as it is not declared there.
+    schedule = psy.invokes.invoke_list[0].schedule
+    assert "a_var" not in schedule.symbol_table
+    # There should be another, distinct entry for "a_var" in the symbol table
+    # for "test_sub2" as it has a use statement that imports it.
+    schedule = psy.invokes.invoke_list[1].schedule
+    avar2 = schedule.symbol_table.lookup("a_var")
+    assert type(avar2) == Symbol
```
rupertford, line 2716: Here too.

rupertford (changes requested):
Nice simple change fixing a bug. I have a few very minor issues.

src/psyclone/psyir/frontend/fparser2.py
arporter: Done (although I went for "we do not check...").

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Yes, but pylint is wrong in this case as I want to check that it really is a Symbol and not a subclass. I've disabled the warning. (Sergi and I discussed this when I made the change to create Symbols instead of DataSymbols for unknown symbols. We could introduce a new UnknownSymbol class instead.)

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Disabled in this test now.

arporter:
Ready for next review.

src/psyclone/tests/psyir/frontend/fparser2_test.py
rupertford: OK, thanks for explaining. There should probably be a comment in the code to that effect, but seeing as it's test code it's not that important.

rupertford (approved):
Looks good. Minor comments have been addressed. Approving.

Issue #905, opened 2020-09-24
[PSyData] Add PSyData wrapper to check for invalid ieee numbers


hiker:
This should be similar to the extract-transformation (detect input/output parameter), but avoiding the 'post' suffix that is being added there. And the implementation of the wrapper library is similar to the read-only functionality.

PR #907, opened 2020-09-25
906 refactor psydata

### LINKED ISSUE #906: [PSyData] Some refactoring to make it easier to reuse code
E.g. error messages contain the hard-coded name of the transformation, which means when we want to inherit from an existing transformation, we have to catch all exceptions, change the error message, and raise it again.

hiker:
This closes #906, which makes implementing #905 easier (avoid code duplication). I've made it a separate PR to keep the PRs smaller.

This PR makes error messages more generic/depending on attributes so there is no need to catch and then re-raise exceptions to fix error messages, and removes some minor code duplication. Several pylint warnings were fixed (import not at top), which also required to add some more symbols to __init__.py (that code could only have worked because symbols had been imported previously, i.e. it was somehow dependent on order of imports somewhere).

Note that the somewhat clumsy "...that has a node of type ReadOnlyVerifyNode at its root." was picked to avoid the a/an issue when using the shorter "that has a ReadOnlyVerifyNode at its root" (e.g. when you replace ReadOnlyVerifyNode with ExtractNode). Alternatives would be to not use 'a/an' (which was done in some classes, i.e. "that has ReadOnlyVerifyNode at its root"), or ... something else that I didn't think of :) Overall I thought "a node of type ..." was best.

Note that the implementation of ``__str__()`` and ``name()`` could be moved up higher (into the Transformation base class even, it has no ``__str__()`` and ``name()`` is abstract), but wasn't sure if you would want that. Also the use of ``__class__.__name__`` appears to be the only portable (python 2/3) way of getting the name of a class as a string, and it has been used before (e.g. node.py).


src/psyclone/psyir/nodes/__init__.py
```diff
@@ -63,6 +63,8 @@
 # this package e.g. 'from psyclone.psyir.nodes import Literal'
 __all__ = [
         'Node',
+        'colored',
```
arporter, line 66: Not the fault of this PR (and probably not to be addressed here either) but looking at this now I don't think 'colored' should be under psyir.nodes.

src/psyclone/psyir/transformations/psy_data_trans.py
```diff
@@ -79,20 +82,24 @@ class PSyDataTrans(RegionTrans):
...
     def __str__(self):
-        return "Insert a PSyData node."
+        return ("Create a sub-tree of the PSyIR that has a node of type "
+                "{0} at its root.").format(self._node_class.__name__)
 
     @property
     def name(self):
-        ''' Returns the name of this transformation as a string '''
-        return "PSyDataTrans"
+        ''' Returns the name of this transformation as a string.
```
arporter, line 97: Sorry to be a pain but please could you reformat this so that it has a description plus :returns: and :rtype:.

src/psyclone/tests/psyir/transformations/profile_test.py
```diff
@@ -523,7 +521,8 @@ def test_transform(capsys):
     schedule._const_loop_bounds = True
 
     prt = ProfileTrans()
-    assert str(prt) == "Insert a profile start and end call."
+    assert str(prt) == "Create a sub-tree of the PSyIR that has " \
```
arporter, line 524: Now that I get to here, I'm concerned that this isn't a very user-friendly message although it's only the `__str__` method so perhaps that doesn't matter. It's not something your average user will see.

arporter (changes requested):
Almost perfect! Just the one docstring that needs formatting to use :returns: and :rtype:.
Tests all pass, including with compilation.
Examples are all OK.
Documentation all builds OK (modulo warnings for Ref. Guide but they were there before).
Code conformance is all clean and much better than it was before (thanks).

src/psyclone/psyir/nodes/__init__.py
hiker: I agree ... with everything. It shouldn't be there, and it's not to be addressed here :) Note that I only added this since in the past it worked because of a certain order of import statements (local imports, which I have moved to the top).

src/psyclone/tests/psyir/transformations/profile_test.py
hiker: I kind of agree, but also note that the previous text to a certain degree could be called wrong: the transformation does NOT insert start and end calls, that is just a 'side effect' of the node that is being inserted. So I think this is certainly an improvement.

hiker:
This should also be ready to go.

src/psyclone/psyir/nodes/__init__.py
arporter: I've created #913 for this.

arporter:
Requested change has been made. Tests and examples are all fine with compilation.
Reference guide still builds fine.

arporter (approved):

Issue #908, opened 2020-09-25
[LFRic] Make module-level USE statements flexible


TeranIvy:
Issue #817 introduces support for fields of different primitive types by specifying data type for the existing `real` fields and introducing `integer` fields. The LFRic infrastructure initially only had `field_type` as all fields had `real` data. Now it has `field_type` for `real`-valued fields (in [field_mod.F90](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/infrastructure/source/field/field_mod.F90)) and `integer_field_type` for `integer`-valued fields (in [integer_field_mod.F90](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/infrastructure/source/field/integer_field_mod.F90)). These classes are subclasses of the `field_parent_type` defined in [field_parent_mod.f90](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/infrastructure/source/field/field_parent_mod.f90).

From the PSyclone LFRic API point, this means that we need to generate `use integer_field_mod, only: integer_field_type, integer_field_proxy_type` to support integer fields.

The generation of `use` statements for LFRic fields and operator classes is currently fixed in [class DynamoPSy(PSy)](https://github.com/stfc/PSyclone/blob/master/src/psyclone/dynamo0p3.py#L1001) to always contain `use` statements for fields, LMA operators and CMA operators whether or not the generated PSy module uses them. We can go down this route for integer fields as well but I think this would not be wise as there will be more different field and potentially operator types with the addition of multi-data fields, 2D fields, different precisions etc.

I suggest making this generation dependent on what objects are required in the Invoke list generated for the PSy module. That means that the `use` statements for a specific object type (and its proxy type) would only be added if these objects are found in Invokes. From the technical point of view this means that the `class DynamoInvoke` that has access to the objects would inform `class DynamoInvokes` what objects are used and then `DynamoInvokes` could update the list of objects to add to `use` statements or generate the `use` statements for in `DynamoPSy`.

arporter:
I know I'm a bit late to the party here but unless this is a trivial amount of work then I wouldn't bother. The existing solution works OK, even though it's not ideal and, as Rupert said, this code-generation will all be replaced with the move to PSyIR. We have a lot of work on and don't need to give ourselves more :-)

TeranIvy:
[...]

Issue #817, opened 2020-06-26
Fields of different primitive types


TeranIvy:
This issue introduces support for fields of different primitive types (`integer` and `real`) in LFRic API and also sets the metadata type for operators (only `GH_REAL` for now until there is a need for more supported datatypes).

For the kernel metadata design, see this [LFRic wiki](https://code.metoffice.gov.uk/trac/lfric/wiki/LFRicInfrastructure/FieldEnhancements/FieldTypesMetadata). An example of kernel metadata is below

```
  type(arg_type) :: meta_args(4) = (/                                   &
       arg_type(GH_SCALAR,   GH_REAL,    GH_SUM),                       &
       arg_type(GH_FIELD,    GH_INTEGER, GH_READ, W3),                  &
       arg_type(GH_FIELD*3,  GH_REAL,    GH_INC,  W2),                  &
       arg_type(GH_OPERATOR, GH_REAL,    GH_READWRITE, ANY_SPACE_1, W3) &
       /)
```

The scalar arguments and the foundation work are done in #774.

Note that the only thing that is different between `integer` and `real` fields in LFRic API is the Fortran datatype of the field data. Everything else is the same.

Integer built-ins will be dealt with in another issue.

PR #915, opened 2020-09-30
Create "use" statements for data structures in LFRic PSy modules only when needed (for #908)

TeranIvy:
This PR creates `use field_mod, ...` and `use operator_mod, ...` statements that call relevant LFRic data structures when they are needed as opposed to always. This is a prerequisite to narrow down code changes for supporting `integer` fields (issue #817) only when those fields are used (as there we also need to add `use integer_field_mod, ...`).

**Update:** This turned out to be quite a small change in two PSyclone files. Tests with compilation pass. The branch was tested on LFRic code on Spice. The full test suite on Spice is green, the command-line compilation works normally and the generated code shows only the `use` statements that are required by the Invoke calls in a specific PSy-layer module. The branch is now ready for review from either @rupertford and @arporter.


src/psyclone/dynamo0p3.py
```diff
@@ -955,65 +955,94 @@ def is_intergrid(self):
...
         '''
-        :returns: the unmodified psy-layer name.
+        :returns: the unmodified PSy-layer name.
         :rtype: str
 
         '''
         return self._name
 
+    @property
+    def infmod_dict(self):
```
arporter, line 999: I think this needs a better name. Maybe `infrastructure_modules`?

src/psyclone/dynamo0p3.py
```diff
@@ -955,65 +955,94 @@ def is_intergrid(self):
...
+
+        # Add all invoke-specific information
+        self.invokes.gen_code(psy_module)
+        # Inline kernels where requested
+        self.inline(psy_module)
+
+        # Include required infrastructure modules. The sets of required
+        # LFRic data structures and their proxies are updated in the
+        # relevant field and operator subclasses of DynCollection.
+        # Here we sort the inputs in reverse order to have "_type" before
```
arporter, line 1031: Presumably this ordering is just a preference? If so that's OK. If not then I don't think we should be relying on alphabetical ordering in order to get things correct. Strictly speaking, there's no 'bubble up' (in the sense that it is used in f2pygen) here - just an order in which we generate use statements.

src/psyclone/dynamo0p3.py
```diff
@@ -955,65 +955,94 @@ def is_intergrid(self):
...
                         for code optimisation and generation.
     :type invoke_info: :py:class:`psyclone.parse.algorithm.FileInfo`
+
     '''
     def __init__(self, invoke_info):
         PSy.__init__(self, invoke_info)
         self._invokes = DynamoInvokes(invoke_info.calls, self)
+        # Initialise the dictionary that holds the names of required
+        # LFRic data structures and their proxies for the "use"
+        # statements in *_mod_psy.f90 modules
```
arporter, line 974: suggest: "...in modules that contain PSy-layer routines."

arporter (changes requested):
Good job.
Tests and examples are all fine.
Documentation still builds OK.
Conformance is fine.
I have just a few minor comments on naming and ordering.

src/psyclone/dynamo0p3.py
TeranIvy: Changed.

src/psyclone/dynamo0p3.py
TeranIvy: I tried to be short but I agree :smile: Changed.

src/psyclone/dynamo0p3.py
TeranIvy: It is not so much preference as it is simply wanting to keep the existing ordering of names of operator and field types (see explicitly written lists of types for `use` statements on the left) so I don't have to change the tests. Also, I would happily just use the original sets without the list conversions and sorting but that doesn't guarantee the same outcome in Python 2 and Python 3 tests.
Agreed about "bubble up" so I rephrased this sentence.

TeranIvy:
Addressed review comments and retested everything with compilation. Ready for another review from @arporter.

arporter:
All requested changes have been made. Conformance is still good.

arporter (approved):

PR #932, opened 2020-10-13
#931 Fix dependency testing by falling back to variable access info if no symbol table is found

### LINKED ISSUE #931: Crash in dependency test
The dependency tools use:
```
symbol = loop.find_or_create_symbol(var_name)
... symbol.is_array
```
Now, if ``var_name`` is not defined, symbol is apparently a Symbol, which does not have an ``is_array`` method (only Datasymbols have that). For the same reason, one dependency test is atm x-failing.

hiker:
As discussed on teams. Fixes #931.


arporter (approved):
Small change to workaround current PSyIR limitation. All tests are OK (including coverage). Examples are all OK. This is a bugfix so no documentation update required.
I've taken the liberty of doing a tiny bit of tidying.

PR #938, opened 2020-10-15
fix test_lookup_2 error

### LINKED ISSUE #937: test_lookup_2 test fails with AssertionError
The `test_lookup_2` unit test fails with the error:

`E       assert "'var2' exists in the Symbol Table but has visibility 'PRIVATE' which does not" in '<ExceptionInfo SymbolError tblen=2>'`
`E        +  where '<ExceptionInfo SymbolError tblen=2>' = str(<ExceptionInfo SymbolError tblen=2>)`

AdamVoysey:
Fixes #937 - trivial change to prevent an AssertionError in `test_lookup_2` when used with Python 3.6.8 (with pytest-5.2.1)


arporter (approved):
Change is purely to correct `str(err)` to `str(err.value)` in two places in symboltable_test.py. There are no other instances of `str(err)` in that file.
Documentation does not need updating.
Change is only to test code so coverage/examples not a problem.

arporter:
I've checked that tests/examples are all OK with compilation enabled after bringing branch up-to-date with master.
Will merge once Travis is green.

PR #940, opened 2020-10-15
add encoding for python2

### LINKED ISSUE #939: Some tests fail with Unicode errors under Python2 in non-unicode locales
When run under Python2 in a non-Unicode locale, some tests fail with Unicode errors.

This can be demonstrated by invoking `pytest` in the following way, which succeeds:

`LANG=en_GB.UTF-8 pytest`

However when invoked as:

`LANG=C pytest`

the tests `test_write_utf_file` & `test_utf_char` fail with errors, for example:

`E           UnicodeEncodeError: 'ascii' codec can't encode character u'\xb0' in position 84: ordinal not in range(128)`

AdamVoysey:
fixes #939 - adds an utf-8 encoding to python2 branch branch in `generator.py` such that the `test_write_utf_file` & `test_utf_char` tests pass in non-Unicode locales


src/psyclone/generator.py
```diff
@@ -445,7 +445,7 @@ def write_unicode_file(contents, filename):
         if not isinstance(contents, unicode):
             contents = unicode(contents, 'utf-8')
         # pylint: enable=undefined-variable
-        encoding = {}
+        encoding = {'encoding': 'utf-8'}
```
arporter, line 448: Since the `encoding = ` assignment is now duplicated, please could you move it to just before the call to `io.open`.

arporter (changes requested):
Thanks for tracking this one down. I was able to reproduce (once I'd installed Python 2).
Just a bit of code duplication to fix.
All tests now pass.
This is a bug fix so no need to update documentation or examples.
Please add yourself to the author list at the top of the modified file.

src/psyclone/generator.py
AdamVoysey: Requested changes made as 63476a2

arporter (approved):
Requested change has been made.
All tests pass/xfail (incl. with compilation).
Ditto examples.
Changes are pylint clean.

PR #942, opened 2020-10-19
Forbid parallel loop from containing CodeBlock (closes #941)

### LINKED ISSUE #941: ParallelLoopTrans should refuse to enclose a CodeBlock
Currently `ParallelLoopTrans` makes no attempt to check the body of the loop to which it is being applied. At the very least we need to ban `CodeBlocks`, ideally we need to check that the loop is safe to parallelise using @hiker's dependency analysis.

arporter:


arporter:
Assuming Travis is happy, this is ready for first review.

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -227,6 +227,25 @@ def test_regiontrans_wrong_options():
     assert ("Transformation apply method options argument must be a "
             "dictionary but found 'str'." in str(excinfo.value))
 
+
+def test_parallellooptrans_refuse_codeblock():
```
rupertford, line 231: As you have also updated ParallelRegionTrans, could you please add an associated test for this too.

rupertford (changes requested):
Looks great. Just one request for an additional test - see inline comment.

src/psyclone/tests/psyir/transformations/transformations_test.py
arporter: Oh yes, done.

arporter:
New test added as requested. I did try to be clever and do it by 'parameterising' the existing test but the text of the error messages is too different for that to be workable.
Ready for next review assuming Travis is happy.

rupertford:
Yes I noticed that the errors in the two cases were quite different and thought about suggesting they were aligned but it is not a big deal and I am fine with two tests in any case.

rupertford (approved):
Looks good. Approving.

Issue #955, opened 2020-10-27
[PSyData] Use Jinja for gocean netcdf extraction library


hiker:
It makes sense to share code using a jinja-template as base class for dl_esm_inf and LFRic.

First problem: at this stage the directory structure is not suited for this:
```
lib/extract/dl_esm_inf/netcdf
lib/extract/gocean/netcdf
```
That makes it inconvenient to have a common netcdf base class. The following structure would be better:
```
lib/extract/netcdf/dl_esm_inf
lib/extract/netcdf/gocean
```
This way we could potentially have a base class in ``extract``, a netcdf-base class in ``extract/netcdf``, and then the API-specific implementation.

PR #956, opened 2020-10-27
#955 Change directory structure to work better with inheritance.

hiker:
This implements the first part of #955, renaming the directory (separate PR to avoid renaming and adding/modifying file in one PR).


hiker:
Ready to go - first part of #955: only renaming the directories. I did also remove one paragraph from the docs that was in a completely wrong place (and a duplicate of a nearly identical previous paragraph).

arporter (approved):
All looks good.
All examples still OK (incl. with compile)
Tests all good.
Updated UG builds fine.
No src touched so no need to check Ref Guide.

arporter:
Updating to HEAD of master worked fine on my machine. Just need to wait for Travis now and then I'll merge.

PR #958, opened 2020-10-27
Ensure OMP/ACC Loop checks for parent region in NEMO API (closes #957)

### LINKED ISSUE #957: [nemo] ACCLoopDirective does not check for enclosing parallel region in NEMO API
As we are still relying on the fparser2 parse tree for the NEMO API, code generation is handled by the `update()` method rather than `gen_code`. `ACCLoopDirective.update()` is missing the validation check present in the `gen_code` method which means it fails to check that the directive is enclosed within some OpenACC parallel region.

arporter:


arporter:
Assuming Travis is happy, this is ready for review.

rupertford (approved):
In an ideal world I would have liked a set of tests for the two _pre_gen_validate functions but I appreciate these are covered via tests that call this function so I will let it slide. Otherwise, all looks good and I like the concept of a _pre_gen_validate.

PR #965, opened 2020-11-06
Create python-package.yml

arporter:
Begin activating GitHub action


arporter:
Self-hosted runners are documented here: https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners
Given that the runners just poll GitHub I think we'll be able to set them up on our desktops in the lab (for example).

arporter:
I think this now replicates the functionality of Travis with the exception that there's no attempt to deploy to pypi (I think we use a separate Action for that) and we don't run the tests with `termcolor` installed. We could, separately, move to having termcolor as a dependency and then always run the tests with it.

I've currently got it setup so that we test on Python 2.7, 3.5 and 3.8. I've chosen those versions somewhat arbitrarily (although I think @TeranIvy said that the MO have hit problems with 3.5?).

rupertford (approved):
I'm happy to put this on master and see what happens :-). Any problems etc we can fix in subsequent issues. I took the liberty of adding your name as the author of the new file.

PR #967, opened 2020-11-08
Enable testing using a self-hosted runner.

arporter:
Attempts to add an action that pushes to a private mirror repo


arporter:
Although this is quite fun, it needs more work and is not a priority therefore I'm closing the PR for now.

arporter:
We have renewed interest in this as we could really do with a semi-automatic way of testing the full pipeline with e.g. NEMO on GPU.

arporter:
Authentication is currently failing. I think an SSH key might be a good way to go, as described here: https://adventures.michaelfbryan.com/posts/configuring-cargo-auth-in-github-actions/

arporter:
I had to follow https://stackoverflow.com/questions/64953274/how-to-send-passphrase-for-ssh-add-with-github-actions to get the authentication working with an ssh key with a passphrase. Authentication does now work but I'm unsure how to configure the mirror repo to run on a self-hosted runner while not having that in PSyclone proper.

arporter:
Apparently I needed `fetch-depth: 0` in the `checkout` section. With that change the push is fine.

arporter:
The resulting push to PSyclone-mirror triggered the test suite on glados which did compilation testing (using gfortran). Probably I shouldn't use `-n auto` for this :-)

arporter:
I just need to update the Developer Guide now.

arporter:
The link-check failure is down to Binder is having another bad day.

This is ready for a first review now. Probably one for @sergisiso or @rupertford but @hiker might have comments too. Hopefully the new additions to the Developer Guide will explain how things work!

sergisiso:
@drewsilcock may also want to have a look that what we are doing is sensible, since I have been speaking about these changes with him before.

drewsilcock:
> @drewsilcock may also want to have a look that what we are doing is sensible, since I have been speaking about these changes with him before.

Looks eminently sensible to me 🙂 Interesting solution with the manual job instantiation - how come you didn't want to just make the actual on-prem test action require manual instantiation instead of going through the private mirror?

arporter:
> how come you didn't want to just make the actual on-prem test action require manual instantiation instead of going through the private mirror?

Thanks @drewsilcock. The reason I did this was that I read somewhere that it wasn't a good idea to have a self-hosted runner associated with a public repo. as any forks of that repo. will also have access to it. I may have misunderstood and/or the situation may have changed?

drewsilcock:
> > how come you didn't want to just make the actual on-prem test action require manual instantiation instead of going through the private mirror?
> 
> Thanks @drewsilcock. The reason I did this was that I read somewhere that it wasn't a good idea to have a self-hosted runner associated with a public repo. as any forks of that repo. will also have access to it. I may have misunderstood and/or the situation may have changed?
> 
> There's a big discussion here: https://github.com/orgs/community/discussions/26722. I think the summary is that it _is_ possible to tie things down to make it safe but it may be relatively easy to mistakenly open things up again.

Interesting, I'm sure you could restrict the usage sufficiently to make it secure but I don't know whether I'd trust myself to configure it properly so it sounds very sensible to use a private mirror!

The GitHub docs does still recommend only using self-hosted runner with private repos: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security

sergisiso (approved):
Everything looks good, thanks @arporter. I will approve and merge this PR and mark it as a "towards to" a new issue with the current limitation (#2132).

PR #970, opened 2020-11-13
955 jinja for extraction

hiker:
This introduces a jinja-created base class for netcdf extractions, and uses it in the existing dl_esm_inf extraction wrapper library.


lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+    !! file).
+    type, extends(PSyDataBaseType), public:: ExtractNetcdfBaseType
+        !> The NetCDF id used for this file.
+        integer                            :: ncid
+        !> Each variable id. This is required to associate data
+        !! with the declared variables: the variables are declared
+        !! in the same order in which their value is provided
+
+        integer, dimension(:), allocatable :: var_id
+        !> Number of variables that are written before the
```
arporter, line 113: Should this comment be here?

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+
+    ! -------------------------------------------------------------------------
+    !> This subroutine is called to open a NetCDF file for reading. The
+    !! filename is based on the module and kernel name. This is used by a
+    !! driver program that will read a NetCDF file previously created by the
+    !! PSyData API.
+    !! @param[inout] this The instance of the ExtractNetcdfBaseType.
+    !! @param[in] module_name The name of the module of the instrumented
+    !!            region.
+    !! @param[in] kernel_name The name of the instrumented region.
```
arporter, line 213: Would this be better as `region_name` to make it consistent with PreStart()?

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+        integer :: retval
+
+        call this%PSyDataBaseType%PreEndDeclaration()
+        retval = CheckError(nf90_enddef(this%ncid))
+    end subroutine PreEndDeclaration
+
+    ! -------------------------------------------------------------------------
+    !> This subroutine is called after the value of all variables has been
+    !! provided (and declared). After this call the instrumented region will
+    !! be executed. For this NetCDF interface this function is not required,
```
arporter, line 245: It isn't empty! :-)

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+        implicit none
+        class(ExtractNetcdfBaseType), intent(inout), target :: this
+        call this%PSyDataBaseType%PreEnd()
+    end subroutine PreEnd
+
+    ! -------------------------------------------------------------------------
+    !> This subroutine is called after the instrumented region has been
+    !! executed. After this call the value of variables after the instrumented
+    !! region will be provided. For this NetCDF interface this function is
+    !! not required, so it is empty.
```
arporter, line 258: Again, it's not strictly empty. It feels like here (and above) we're overriding a routine, only to call the base implementation. However, I can see it's not quite that simple as we have `this%PSyDataBaseType%PostStart()` rather than just `this%PostStart()`?

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+{# This defines a mapping of the 'userfriendly' names to the NetCDF data types. #}
+{% set NCDF_TYPE_MAPPING = { "Double": "NF90_DOUBLE",
+                             "Real":   "NF90_REAL",
+                             "Int":    "NF90_INT"}%}
+
+
+{% for name, type, bits in ALL_TYPES %}
+    {% set NETCDF_TYPE = NCDF_TYPE_MAPPING[name] %}
+    ! -------------------------------------------------------------------------
+    !> This subroutine declares a scalar integer value. A corresponding
```
arporter, line 289: s/integer/{{name}}/ I think?

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+  {% for dim in ALL_DIMS %}
+    {# Create the ':,:,:,:' string
+       We repeat the list [":"] DIM-times, which is then joined #}
+    {% set DIMENSION=([":"]*dim)|join(",") %}
+
+    {# Create list of variables: "dimid1,dimid2,dimid3,dimid4" #}
+    {% set vars = "dimid"~range(1,dim+1)|join(",dimid") %}
+
+    ! -------------------------------------------------------------------------
+    !> This subroutine declares a double precision field as defined in
```
arporter, line 358: s/double precision/{{name}}/ I think?

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+    !! to the NetCDF file, and the variable id is stored in the var_id field.
+    !! @param[inout] this The instance of the extract_PsyDataType.
+    !! @param[in] name The name of the variable (string).
+    !! @param[in] value The value of the variable.
+    subroutine DeclareArray{{dim}}d{{name}}(this, name, value)
+
+        use netcdf
+        implicit none
+
+        !class(PSyDataBaseType         ), intent(inout), target :: this
```
arporter, line 369: Please rm this line.

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+        {% endfor %}
+        dimids =  (/ {{vars}} /)
+        retval = CheckError(nf90_def_var(this%ncid, name, {{NETCDF_TYPE}},     &
+                                         dimids, this%var_id(this%next_var_index)))
+
+        call this%PSyDataBaseType%DeclareArray{{dim}}d{{name}}(name, value)
+    end subroutine DeclareArray{{dim}}d{{name}}
+
+    ! -------------------------------------------------------------------------
+    !> This subroutine writes a for a {{dim}}d-array of {{type}}
```
arporter, line 391: "writes a ??? for a ..."

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+
+        if (.not. is_enabled) return
+        retval = CheckError(nf90_put_var(this%ncid, this%var_id(this%next_var_index),  &
+                                         value({{DIMENSION}})))
+        call this%PSyDataBaseType%ProvideArray{{dim}}d{{name}}(name, value)
+
+    end subroutine WriteArray{{dim}}d{{name}}
+
+    ! -------------------------------------------------------------------------
+    !> This subroutine reads the values of a dl_esm_inf field (r2d_field).
```
arporter, line 413: This comment needs parameterising.

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+        {{type}}, dimension({{DIMENSION}}), allocatable, intent(out) :: value
+        integer :: retval, varid
+        integer :: dim_id
+        {# Create list of variables: "dimid1,dimid2,dimid3,dimid4" #}
+        {% set dim_sizes = "dim_size"~range(1,dim+1)|join(",dim_size") %}
+        integer :: {{dim_sizes}}
+        integer :: ierr
+        character(100) :: x
+
+        ! First query the dimensions of the original r2d_field from the
```
arporter, line 436: s/r2d_field/array/ I think?

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -0,0 +1,464 @@
...
+        ! Allocate enough space to store the values to be read:
+        allocate(value({{dim_sizes}}), Stat=ierr)
+        if (ierr /= 0) then
+            print *,"Cannot allocate array for ", name, &
+                    " of size ", {{dim_sizes}}, " in ReadArray{{dim}}d{{name}}."
+            stop
+        endif
+        ! Initialise it with 0, so that an array comparison will work
+        ! even though e.g. boundary areas or so might not be set at all.
+        value = 0.0d0
```
arporter, line 454: Is this hardwired double-precision constant OK here? Presumably the compiler will just do the Right Thing?

arporter (changes requested):
All builds OK.
All examples OK.
All tests OK.
It's possible that we should update documentation but that can be done in a future PR.
Just a few things to fix and then I'll merge.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Indeed not anymore.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Indeed.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: I have removed it. Now it is ... not empty, but gone.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: This should also be considered empty ... and it is gone. But: ``this%PostStart()`` would call itself recursively, you have to specify the base class here (afaik).

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Yep.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Yes.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Gone.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: I don't know either :grin:

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Done.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Yes.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: Yes, the compiler should automatically convert this to the appropriate type.

hiker:
> It's possible that we should update documentation but that can be done in a future PR.

I've created #988 for this. Technically the jinja implementation has been there, and nothing really new is here, but it is certainly worth checking.

I've also changed the documentation string to use ``{{type}}`` instead of ``{{name}}`` - so instead of '.. an array of Int', it is now ` 2d-array of integer(kind=int32)`, which I think is more precise.

> Just a few things to fix and then I'll merge.

arporter:
All requested changes have been made. All tests pass. All examples are OK.
Branch is up-to-date with master.
Will proceed to merge once Travis is green.

arporter (approved):

PR #974, opened 2020-11-15
Support skip ci with GitHub actions [closes #973]

### LINKED ISSUE #973: support [skip ci] with gihub actions
github actions do not respect the [skip ci] message in a commit i.e. will run irrespective. The suggestion workaround is to use an if statement ...

see e.g. https://github.com/marketplace/actions/skip-based-on-commit-message

which suggests adding ...

`if: "!contains(github.event.head_commit.message, 'ci skip')"`, or (skip ci in our case).

rupertford:


rupertford:
I can't seem to be able to get this to work :-(

arporter:
Is the issue to do with the fact that we don't have a `main:` section in our yml? Although, having said that, you'd have thought you'd either need toJson or you don't so it's weird that the solution at https://github.com/veggiemonk/skip-commit/issues/5 doesn't work.

arporter:
The yml now successfully gets the hash of the current commit. Next stage is to work out how to get the associated commit message.

rupertford:
Perhaps we could try this ... https://github.com/marketplace/actions/ci-skip-action

arporter:
The last commit did succeed in triggering the test suite - I just cancelled it manually for efficiency reasons. We therefore have a working solution!

arporter:
Since we have the codecov-reported coverage as a required check, we could choose to change fail-fast to `False` (so that the tests aren't reported as failing) and just ensure that no coverage was uploaded to codecov if skip-ci was supplied (as well as skipping the tests themselves, obviously). I'll let the reviewer decide :-)

arporter:
This is ready now but I realise that since @rupertford opened this PR, he isn't allowed to review it! I'll therefore request @hiker or @sergisiso to review but obviously @rupertford, feel free to informally review.

.github/workflows/python-package.yml
```diff
@@ -57,6 +56,12 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
+        fetch-depth: '0'
+    - name: Check for [skip ci] in commit message
+      uses: mstachniuk/ci-skip@v1
+      with:
+        fail-fast: true
+        commit-filter: 'skip ci'
```
sergisiso, line 64: I think the square brackets should be part of the filter to avoid false positives.

sergisiso (changes requested):
@arporter See the inline comment.

> we could choose to change fail-fast to False (so that the tests aren't reported as failing) and [...] I'll let the reviewer decide :-)

I prefer it fails the test as it is right now.

arporter:
Changed to include square brackets in filter, as suggested. Ready for another look.

sergisiso (approved):
@arporter PR approved. The changelog commit should bring the tests to pass again.

PR #975, opened 2020-11-15
972 - fix str(err) -> str(err.value)

### LINKED ISSUE #972: Fix asserts using str(err)
Similarly to #937, some tests fail due to `assert` statements which use `str(err)` rather than `str(err.value)`.

These appear to have been introduced recently in https://github.com/stfc/PSyclone/commit/83d82f123c3bdf77c64358701a859dbb7f8c1b4b & https://github.com/stfc/PSyclone/commit/b04d68478afe5f92d87aaebc96358a8ef2cdb847

AdamVoysey:
fixes #972 - change instances of `str(err)` to `str(err.value)`


arporter (approved):
Many thanks for fixing my mistake!
This is just a fix to two tests and therefore no need to update documentation etc.

PR #977, opened 2020-11-16
#976 Fix dimension parameter for jinja templates

### LINKED ISSUE #976: Jinja-code in PSyData libraries does create wrong dimensions for arrays
The ``ALL_DIMS`` parameter in the template is not set properly, which means the ``--dims`` parameter is ignored, and e.g. for gocean 1- to 4-dimensional arrays are being created, even though only 2d arrays are needed.

hiker:
Minor changes, had the wrong jinja name in process.py (it also fixes some other minor issues).

Fixes #976.


arporter (approved):
All changes are good.
Examples are all OK with compilation.
Tests all OK with compilation.
No need to update documentation.
Will proceed to merge once Travis is happy.

PR #981, opened 2020-11-16
(closes #980) mv jupyter tutorials and update various docs

### LINKED ISSUE #980: Restructure Jupyter-based tutorial directories
Now that we're adding other tutorial directories, we need to move the location of the notebooks.

arporter:


arporter:
Functionally this is a small change but various documentation etc. needed to be updated. We don't currently exercise these notebooks in our CI but that's something we can think about later :-)

rupertford (approved):
The user guide tutorial section could be a bit clearer about the different types of tutorial - perhaps we should have different sections at some point. However, seeing as none of the new tutorial stuff is on master, this does not matter and it can be updated later.

PR #983, opened 2020-11-16
Improve 'kernels' auto-profiling for nemo API (closes #982)

### LINKED ISSUE #982: [nemo] Kernels auto-profiling includes loop with CodeBlock in NEMO API
By definition, a loop body that contains a CodeBlock is not a kernel and therefore should not be profiled by the 'kernels' auto-profiling option.

arporter:
This is a bug fix for #982 - it generalises the Profiler.add_profile_nodes() so that it looks for valid kernels and moves up the PSyIR to find the parent loop.


arporter:
Assuming the CI is green, this is ready for review. Probably best done by @hiker as it touches his code but time is short so anyone can have a go!

hiker:
Note that I was thinking that this might need an update to the UG, but following your wording the definition of 'kernel' has not changed, so I considered it a bug fix only.

As a bug fix only, no documentation updated required, test cases for this have been added.

PR #1000, opened 2020-11-24
988 add dependencies

### LINKED ISSUE #998: Dependencies missing in makefile
The Makefiles for the gungho libraries in tutorial../3_time_evolution (and therefore also in 4_psydata) has no dependencies. It seems to compile files in the right order by coincidence. When I tried to compile with intel on a Cray, it repeatedly failed because it compiled in the wrong order. You can likely enforce the problem by compiling in parallel (-j 8 or so).

hiker:
Fixed #998. Verified back on the Cray to now compile. Hopefully not needed for tutorial (since it seems to work on most circumstances), but if someone should have a problem we can ask them to pull (or switch to branch).


TeranIvy (approved):
GitHub actions have completed successfully as well as Travis Python 3 jobs. Travis Python 2 jobs are still in progress but I'm confident everything is fine and the failures yesterday were glitches in the build system and not the branch. Proceeding to merge.

PR #951, opened 2020-10-23
Allow PSyIR blocks to be generated inside f2pygen (closes #950)

sergisiso:
Provides a temporary mechanism to start porting the PSy-layers to PSyIR without the need to port it all at once.

It already allows to do things like module-inlining modified kenrels into the PSy-layer.

Closes #950

I think #490 and #663 may also be fixed by this.

PR #1011, opened 2020-11-25
Fix auto kernel profiling in the presence of omp/acc directives (closes #1008)

### LINKED ISSUE #1008: [nemo] auto kernel profiling can fail after OpenMP transformations
Will Sawyer has reported that one of the steps in the OpenMP tutorial fails:

    3_nemo_openmp$ psyclone -api nemo -s solutions/omp_trans.py --profile kernels tra_adv_mod.F90

    Error, unexpected exception, please report to the authors:
    Description ...
    'Transformation Error: A PSyData node cannot be inserted between an OpenMP/ACC directive and the loop(s) to which it applies!'

arporter:


arporter:
This is ready for first review. Fixes a bug found during the tutorial (that I introduced when I hurriedly fixed the 'kernels' profiling for the NEMO API). While doing this I also discovered that the `psy_data_node` functionality in the NEMO API was a bit broken (wasn't setting pointers into the updated fparser2 AST) so I've fixed that too.

src/psyclone/profiler.py
```diff
@@ -129,6 +129,16 @@ def add_profile_nodes(schedule, loop_class):
                     target = parent_loop
                     parent_loop = parent_loop.ancestor(loop_class)
                 if target:
+                    # Have to take care that the target loop does not have
+                    # a directive applied to it. We distinguish this case
+                    # from that of a directive defining a region by checking
+                    # the number of children of the directive.
+                    if (isinstance(target.parent.parent, Directive) and
+                            len(target.parent.parent.dir_body.children) == 1):
```
hiker, line 137: I am not sure that this test is actually correct - what if the directive has just a single loop (I.e. one child). I modified the nemo test:
```
+++ b/src/psyclone/tests/nemo/transformations/profiling/nemo_profile_test.py
@@ -639,8 +639,10 @@ def test_profile_nemo_acc_kernels(parser):
                                       "  sto_tmp(ji,1) = 0.0d0\n"
                                       "end do\n"
                                       "end program do_loop\n")
-    acctrans.apply(schedule.children)
+    acctrans.apply(schedule.children[0])
+    schedule.view()
     Profiler.add_profile_nodes(schedule, Loop)
+    schedule.view()
     code = str(psy.gen).lower()
     assert ("  type(profile_psydatatype), target, save :: profile_psy_data1\n"
             "  !$acc kernels\n"
(END)
```
I.e. apply the acc trans only to the first loop, not both of them. Then I get:
```
    0: Profile[]
        Schedule[]
            0: Directive[ACC Kernels]
...
    1: Profile[]
        Schedule[]
            0: Loop[type='lon', field_space='None', it_space='None']
```
Now my understanding is that this means we would measure the time on the cpu side, which might not be what we want(?).
Would it be better to have a function in the directive class that indicates if 'code' (can't find the right term atm) can be inserted between it and its child?

hiker (changes requested):
Besides the one comment the patch looks all good. The tests did indeed trigger the bug originally reported (when used without the fix), pylint/pychodestyle are clean.

src/psyclone/profiler.py
arporter: I think this is OK as I was always imagining that the profiling would be run on the CPU. However, that has made me realise that, as written, this transformation will insert profiling calls *within* an ACC KERNELS region (if it contains >1 loop nest) and that won't compile unless the profiling routines are compiled for the GPU. It's a similar problem to having profiling within an OMP PARALLEL region in that in that case the profiling lib has to be thread safe. I'm not sure what the best solution is. Maybe forbid PSyData nodes from being inside an ACC region?

hiker: OK, that's the thing I didn't know - I thought GPU profiling would need to be inside OpenACC. It sounds somewhat similar to ``omp do`` - we need to make sure that we don't insert profiling nodes there. So from that point of view I'd say that forbidding profiling nodes after some directives (``omp do``, ``acc``) - though I just realise that this is slightly different: profiling inside an ``omp parallel`` **region** is fine (assuming we link with a thread-safe profiling library), inside an open acc region on the other hand we ... either can't or have to go through a lot of effort.

I'd agree with just disabling it for openacc. We can consider an extension as a separate ticket (e.g. each directive would have a method that specifies what to do with certain PSyData directives - or maybe even more general? ``get_psydata_info()``, which could return 'forbidden', 'up' (move to parent), 'allowed'. That's certainly out of scope for this simple bug fix.

arporter:
I think this is ready for another review now (Travis permitting). I've had to widen its scope a bit in light of the realisation that we can't put PSyData nodes within OpenACC regions.

arporter:
OK, this really is ready for review now :-)

src/psyclone/profiler.py
hiker: While I still don't like the test if there is only one element in a directive, I was not successful in creating a sensible corner case (an ``omp parallel`` containing a non-parallelised loop (e.g. an outer, iterative loop), which then has inside an ``omp do`` parallel loop). This is so hard to construct, that you would need to manually apply the profile transform, and then it's up to you anyway :)

hiker (approved):
It fixes the bug, test case are provided. Coverage of the new code is 100%. Pylint is happy except for a minor issue in ``kernels_directive_test.py``, which I have taken the liberty to fix. Am waiting for CI, then this can be merged in (I already updated changelog and pdf in this branch).

arporter:
@hiker, I'll wait for you to merge this before I do #951.

hiker:
> @hiker, I'll wait for you to merge this before I do #951.

Thanks - sorry, had a break, and then a conflict to resolve :)

Issue #1001, opened 2020-11-24
Scripts using `#!/usr/bin/env python` fail if only `python3` is available`


arporter:
Some of the examples scripts and the Jinja2 build script (lib/process.py) have `#!/usr/bin/env python` but this fails if only Python 3 is available. (e.g. a clean install of Ubuntu under WSL2 has this problem).

A simple workaround is to do `alias python=python3` but we need to find a robust solution.

arporter:
Ubuntu 20.0.4 has dropped Python 2 altogether and thus only ships with `python3`.

arporter:
Googling around, I found the relevant pep which essentially says, "yes, it's a problem and it's your problem...": https://www.python.org/dev/peps/pep-0394/#for-python-script-publishers

We should really move away from having (executable) python scripts in the examples - they are really a hangover from the old way of doing things and should be replaced with a suitable invocation of `psyclone`.

Not sure what to do about `process.py`. The aforementioned pip talks about re-writing shebangs in installed scripts as required - we could do that.

Alternatively, we could adopt the convention of appending a `3` to Python-3-specific scripts and maybe having these alongside the existing scripts. e.g. psyclone3 and genkernelstub3.

hiker:
One option could be to make process.py a shell script, which checks for various known python names, and uses this to execute process.py?

arporter:
Oh yes - that's a good idea @hiker. In fact, I see that our `psyclone` script also has the same shebang and yet that seemed to work for people yesterday. It's very confusing!

hiker:
I nearly wrote the same, but that's actually not true: the installation process created a script in .local/bin (or wherever), and IT has the right python version (guess they  query the system). This script then starts our script using this python version.

hiker:
I'll take this (we somehow are all assigned ... so I thought I'll let you know :grin: )

rupertford:
python-is-python3 seemed to fix this for me, or at least something similar :-)

sergisiso:
> python-is-python3 seemed to fix this for me, or at least something similar :-)

This is what I use in my WSL

rupertford:
I should probably point out that @sergisiso told me about it in the first place :-)

hiker:
[...]

PR #1012, opened 2020-11-26
#1001 Added script to find a python executable, and use it for all calls to process.py

hiker:
As discussed in #1001 . I decided not to try to install process.py (under whatever name), since I think it's better not to 'pollute' the user's PATH with unnecessary binaries. Once we install the psydata libraries (i.e. the lib directory), all makefiles use relative references to the scripts, so they all should work without any change.


lib/get_python.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/bash
```
arporter, line 1: Please add copyright/license info.

lib/get_python.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# This scripts find an executable python command. On
```
arporter, line 3: "This script finds an..."

lib/get_python.sh
```diff
@@ -0,0 +1,24 @@
...
+    # Necessary in case that 'python' is a directory
+    if [[ -f "$p" && -x $(realpath "$p") ]]; then
+        echo $p;
+        exit
+    fi
+done
+
+echo "Cannot find working python"
+
+exit -1
```
arporter, line 24: Missing new-line.

arporter (changes requested):
All Makefiles under lib have been changed.
Fix is straightforward and makes sense.
This branch works OK for me on a WSL2 installation (which only has python3).
I have no Mac to test this on so, for completeness, please could either @deardenchris or @christophermaynard checkout this branch on their Mac, `cd lib/nan_test` and do `make` and check that there are no errors.
Apart from that that there are just a few minor typos and a missing license/copyright.

deardenchris:
Hi Andy, I've just tried this on my Mac. I found that I had to change `#!/usr/bin/bash` in `get_python.sh` to `#!/bin/bash` otherwise I got the error `/bin/sh: ../get_python.sh: /usr/bin/bash: bad interpreter: No such file or directory`. Although regardless of whether I make the change or not to `get_python.sh`, the same output is produced in the nan_test directory (i.e. the same .mod files and object files) so I think the build is OK?

hiker:
At least on ubuntu there is also ``/bin/bash``, so that might work for everything? Alternatively, google indicates that ``#!/usr/bin/env bash`` might be a portable option (https://forums.fedoraforum.org/showthread.php?302429-BASH-vs-SHELL-usr-bin-bash-vs-bin-bash&s=f6d965a4a5d5cc63b68910950adbbbcd&p=1720154#post1720154) ... but it also mentions that 'nearly always' ``env`` is ``/usr/bin/env``

So, since on all platforms it seems that `/bin/bash`` works, I'll just use this.

lib/get_python.sh
hiker: Done.

lib/get_python.sh
hiker: Oops.

lib/get_python.sh
hiker: Done.

hiker:
I've changed it to ``/bin/bash``, which seems to work on nearly all systems (I tried various systems at BOM, and all had ``/bin/bash``). Ready for next review.

lib/get_python.sh
```diff
@@ -1,6 +1,42 @@
-#!/usr/bin/bash
+#!/usr/bash
```
arporter, line 1: Should this be `/bin/bash`?

arporter (changes requested):
All requested changes have been made but the script now refers to `/usr/bash` when I think it should be `/bin/bash`?

arporter (approved):
All good now.

PR #1014, opened 2020-11-27
Add backend support for derived types (towards #363)

arporter:


arporter:
This is not ready yet but please could @sergisiso and @rupertford take a look at the `examples/psyir/create_aggregate_types.py` script and tell me what they think. It turned out to be harder than I thought to implement this as we now have Symbols that are components of other Symbols but, additionally, we have references that are part of other references.

sergisiso:
@arporter I find this expression confusing:

```
DTYPE_SYMBOL = TypeSymbol("field_type", DeferredType(), ...)
...
# Create a local array of derived types
ARRAY_DTYPE = ArrayType(DTYPE_SYMBOL, [10])
FIELD_BUNDLE = DataSymbol("chi", ARRAY_DTYPE)
SYMBOL_TABLE.add(FIELD_BUNDLE)
```

Do I read correctly that 'chi' is and array of typesymbols (no data) which are also Deferred?

Could you write the Fortran equivalent of what you want 'chi' to be?

sergisiso:
Also probably I am missing details with my pen and paper thinking here but do we need the ComponentSymbols? (it seems to duplicate what the ScructureComponents do)

For `field%component` is enough to have 'field' as a DataSymbol with StructureType in the SymbolTable. Then as we know the field type we can check that 'component' is a valid component in the StructureType.

I think most of the work should be done in Reference. I would create a subclass of Reference (similar to what we did with ArrayReference) that has 'field' as the Symbol Reference and another parameter/logic/child to specify the component navigation, but this are validated with what the type allows, not the general symboltable.
(this may only allow for coarse grain data dependency in the first instance but we do the same for Arrays, treat them as one whole). Would this be feasible?

arporter:
> @arporter I find this expression confusing:
> 
> ```
> DTYPE_SYMBOL = TypeSymbol("field_type", DeferredType(), ...)
> ...
> # Create a local array of derived types
> ARRAY_DTYPE = ArrayType(DTYPE_SYMBOL, [10])
> FIELD_BUNDLE = DataSymbol("chi", ARRAY_DTYPE)
> SYMBOL_TABLE.add(FIELD_BUNDLE)
> ```
> 
> Do I read correctly that 'chi' is and array of typesymbols (no data) which are also Deferred?
> 
> Could you write the Fortran equivalent of what you want 'chi' to be?

Sure, `chi` is an array of `field_type` objects where `field_type` is imported from a module and is therefore of deferred type:

    type(field_type), dimension(10) :: chi

    wind%data(:)=2.0_RKIND
    chi(2_8)%data(1_8)=2.0_RKIND

arporter:
I was wondering about doing the 'coarse grained' approach but felt happier doing the full thing. Because I was doing the full thing and wanted a reference for each level, I obviously need a symbol for each level. This is distinct from the type information because it refers to a particular bit of storage.
For an Array reference (which we really should rename ArrayReference), aren't the index expressions also done as references?

sergisiso:
> and wanted a reference for each level, I obviously need a symbol for each level.

But doesn't the StructureType._components already act as a kind of symboltable scoped inside the Structure (note that we even proposed using a SymbolTable instead of the current OrderDict if the similarities kept increasing)

> This is distinct from the type information because it refers to a particular bit of storage.

But this can be implied from the first symbol already being a DataSybmol of the StructureType (a bit of storage for the whole structure) the other part of the expression are just accessors. (This is what I meant as similar to the Array)

The fine grain data dependencies can be done looking that this accessors expression (also for the Arrays).

But I have to admit I don't know how the AST part of the accessors should look like (specially when they are mixed with other structures or arrays). Maybe now that the tutorial is over we should make a design call to discuss this? (and also my psy-layer backends plan)

arporter:
Following our design discussion I've ditched this approach in favour of #1021 which is much nicer.

Issue #272, opened 2019-01-11
[nemo] replace eager conversion of implicit loops with transformation


arporter:
In #244 we found that the current implementation of eager conversion of implicit loops to explicit loops caused problems (because it is not yet robust). It is also likely to be unnecessary for OpenACC. Therefore we will move this functionality into a Transformation rather than have it happen by default.

arporter:
Cherry-picked commits 365eeb7b2b043ee9ba2e25c018a54746398c96a83 and 3ade5b672af5b6ecc661c72355e463c10716dd9 to bring in the code changes done under #244.

rupertford:
Closing this issue as PR #277 is now on master

PR #1034, opened 2020-12-15
1027 final locale fixes

### LINKED ISSUE #1027: Final locale issues for the tests
There are some final changes that are required to deal with PSyclone tests in non-unicode locales.

In particular, the fparser submodule should be updated to include recent fixes from there.

After these changes, we can run the tests at the Met Office without overriding the `LANG` environment variable.

AdamVoysey:
Fixes #1027. Updates the fparser submodule (to include [PR #272](/stfc/fparser/pull/272)), and fixes the setting of the final UTF-8 encoding.


arporter (changes requested):
Thanks for this Adam, the code changes look good. 
Presumably, since the fparser submodule has been updated we should be running the test suite against this version? If so, please edit `.github/workflows/python-package.yml` and uncomment the line that installs fparser from the submodule.
This also made me realise that the test suite is not actually exercising the bug(s) that you've  found because we don't run with the same LOCALE as the met-office. I therefore propose adding a step to the same yml file so that we run the test suite with the necessary locale for just one python version (I'll let you choose which one). I *think* (GitHub Actions are still a bit of a mystery to me) we can achieve that by adding the following under 'steps:'

    - name: Test with pytest and LOCALE=C
      if: ${{ matrix.python-version == 2.7 }}
      run: |
        LOCALE=C pytest --cov=psyclone --cov-report=xml src/psyclone/tests

AdamVoysey:
I've updated the fparser submodule as requested.

I've also added a new test; however I have deviated slightly from the way suggested as I think adding an `env` section to the yml file is probably more correct? I set it to use Python 3.8 as I think the fact that Python 3 defaults to Unicode strings makes this way round a more useful test.

arporter (approved):

Issue #1037, opened 2020-12-22
LFRic example do not compile with intel compiler


hiker:
The example in ``example/lfric/full_example`` results in:
```
ifort -g -I../../../src/psyclone/tests/test_files/dynamo0p3/infrastructure -c testkern_w0_mod.f90
testkern_w0_mod.f90(70): error #6415: This name cannot be assigned this data type because it conflicts with prior uses of the name.   [UNDF_W0]
    integer(kind=i_def)                                 :: ndf_w0, undf_w0
-------------------------------------------------------------------^
compilation aborted for testkern_w0_mod.f90 (code 1)
make: *** [Makefile:76: testkern_w0_mod.o] Error 1
```
and for tutorial 3:
```
ifort -g $(nf-config --fflags) -I../../../../../../src/psyclone/tests/test_files/dynamo0p3/infrastructure -I../gungho_lib -c ../time_evolution_driver.f90
../time_evolution_driver.f90(143): error #6053: Structure constructor may not have components with the PRIVATE attribute   [GLOBAL_MESH_PTR]
  partition = partition_type( global_mesh_ptr,   &
------------------------------^
../time_evolution_driver.f90(144): error #6053: Structure constructor may not have components with the PRIVATE attribute   [PARTITIONER_PTR]
                              partitioner_ptr,   &
------------------------------^
../time_evolution_driver.f90(145): error #6053: Structure constructor may not have components with the PRIVATE attribute   [PANEL_XPROC]
...
```

hiker:
ba3d90f49cdce80235de794237f976c373b999fc fixes the bugs for ``full_example`` and ``full_example_extract`` - thought not for ``full_example_netcdf``.

The fact that the order of declarations might matter could imply more changes to PSyclone required :cry:  While I admit that I like to declare variables before using them ... I thought Fortran does not require that 😕 
```
     integer                                             :: nlayers
+    integer(kind=i_def)                                 :: ndf_w0, undf_w0
     real(kind=r_def), dimension(undf_w0), intent(inout) :: fld1
     real(kind=r_def), dimension(undf_w0), intent(in)    :: fld2
-    integer(kind=i_def)                                 :: ndf_w0, undf_w0
     integer(kind=i_def), dimension(ndf_w0)              :: map_w0
```

hiker:
I also fixed full_example_netcdf, and made the three files in the full_example* directories as similar as possible.

TeranIvy:
#1062 is merged. Closing this issue.

Issue #999, opened 2020-11-24
PSyData with intel compiler


hiker:
At least the intel compiler 17 on Cray's doesn't like the generic interface used in PSyData. It needs to be investigated if this is a compiler issue, or if the code needs to be changed.

hiker:
Intel and old gfortran complain about the ``stop`` statement as well.

For intel: Apparently the generic interface used in psydata needs to be declared in the base classes as well (and the derived classes then use the same interface, but only add the new functions) ... under investigation.

hiker:
It turns out that I didn't use the generic interface properly (though gfortran accepted it). According to the [Fortran standard](https://j3-fortran.org/doc/year/18/18-007r1.pdf), rule 773:
   
>  A binding-name in a type-bound GENERIC statement shall not specify a specific binding that was inherited or specified previously for the same generic identifier in that derived type definition.

Or as [this stackoverflow](https://stackoverflow.com/questions/25410476/understanding-fortran-extends-types-and-override) questions states:

> You can [...] define a generic already in the base type and extend it in its derived types.

Important is that you can only extend it, not replace an existing binding to a set of argument types. This means that:
- ``psy_data_base.jinja`` needs to declare a generic interface for the variable **declaration** for (say) read-only-test (since neither the read-only base class nor the API-specific read-only class provide declaration). 
-  ``read_only_base.jinja`` needs to declare the generic interface for **providing** variables of standard Fortran types.
- the API-specific read-only implementations extend the generic interface for **providing+declaring** variable for the **API-specific types** (e.g. field types).

To make things more complicated, in e.g. the case of kernel extraction, the generic interface for variable declaration must be provided by the netcdf-base class, the ``psy_data_base`` class must NOT declare them in this case (since only one class in the inheritance tree can declare a generic interface for a certain set of argument types). The netcdf-base class will of course use the methods provided by the ``psy_data_base`` class using the inheritance functionality.

This requires a bit of refactoring - luckily the design that each API-specific library processes its own copy of all jinja files helps a lot, since e.g. for read-only we can process the ``psy_data_base`` class to include generic interface for variable declaration, while for extraction we process the ``psy_data_base`` class not to include this (I am really happy that this design decision pays off now 😄 ).

The additional advantage is that now that the API-specific classes do not need to declare the full generic interface for all Fortran base types (which is now done in one of the base classes), but only extends this generic interface by the API-specific data types - we don't need to use jinja in any API-specific library anymore.

[...]

PR #1038, opened 2020-12-22
999 fix compilation intel

hiker:
This fixes the issue with compiling the PSyData libraries with the Intel compiler. Some refactoring was required (see #999). It has been tested with Intel 20, Intel 17, and gfortran 9.3.0. I tested the gocean implementation with the gocean examples, and the LFRic ones with the tutorial. Due to #1037 , the tutorial does not compile, but I commented out the failing lines and at least linking then worked fine.

It should also fix compilation issues with older gfortran compiler (@TeranIvy , this fixes https://github.com/stfc/PSyclone/pull/1023#discussion_r543389657), though I couldn't test this.

- [x] Depending on order in which #1023 and this is merged, we need to update ``lib/README.me``, which mentions the compilation problem with intel compiler. I did not base this one on #1023.


lib/extract/netcdf/Makefile
```diff
@@ -67,11 +67,11 @@ extract_netcdf_base.o: psy_data_base.o
 # Jinja rules for creating the base class, and the
```
arporter, line 67: I'm unable to build the version of dl_esm_inf distributed with PSyclone with either Intel 17 or 19. I get:

    PSyclone/external/dl_esm_inf/finite_difference/src//field_mod.f90(1062): error #6638: An actual argument is an expression or constant; this is not valid since the associated dummy argument has the explicit INTENT(OUT) or INTENT(INOUT) attribute.   [B2]
    call exchange_generic(b2=self%get_data(), handle=exch, &
^

However, I can compile HEAD of master of dl_esm_inf OK. Does the submodule need updating?

lib/process.py
```diff
@@ -68,6 +68,12 @@
 parser.add_argument("-prefix", help="Prefix to add to the generated PSyData "
                                     "function names",
                     default="")
+parser.add_argument("-generic-declare", action="store_true",
```
arporter, line 71: Please could you add comments explaining why these arguments are needed.

lib/read_only/dl_esm_inf/read_only.f90
```diff
@@ -76,11 +72,7 @@ module read_only_verify_psy_data_mod
         !! to change the state from checksum computation to verification.
         !! And again the base class provides additional, unused methods
         !! like ProvideArray2dInt, which are not needed
-        generic, public :: ProvideVariable => ProvideScalarInt,    &
-                                              ProvideScalarReal,   &
-                                              ProvideScalarDouble, &
-                                              ProvideArray2dDouble,&
-                                              ProvideFieldDouble
+        generic, public :: ProvideVariable => ProvideFieldDouble
```
arporter, line 75: Please could you update the comment above to explain why only one routine is now listed here.

lib/read_only/lfric/Makefile
```diff
@@ -80,27 +80,21 @@ read_only.o: 	  $(LFRIC_LIB) read_only.f90 read_only_base.o
...
-
 %.o: %.f90
 	$(F90) $(F90FLAGS) -c $<
 
 # The LFRic ReadOnly library is built on the psy-data base library
 # and the read-only base library. So add the rules to process
 # these files to get the plain Fortran files:
 psy_data_base.f90: ../../psy_data_base.jinja Makefile
-	$(PROCESS) $(PROCESS_ARGS) $< > psy_data_base.f90
+	$(PROCESS) $(PROCESS_ARGS) -generic-declare $< > psy_data_base.f90
```
arporter, line 90: Please extend the comment to explain what the `-generic-xxx` arguments are for.

lib/read_only/read_only_base.jinja
```diff
@@ -109,15 +109,28 @@ module read_only_base_mod
...
+          {{- all_provides.append("ProvideScalar"~name) or "" }}
           {% for dim in ALL_DIMS %}
         procedure :: ProvideArray{{dim}}d{{name}}
+            {{- all_provides.append("ProvideArray"~dim~"d"~name) or "" }}
           {% endfor %}
         {% endfor %}
 
+        {% set indent="            " %}
+
+        {% if GENERIC_PROVIDE %}
```
arporter, line 126: Please add a comment explaining why this is only done if GENERIC_PROVIDE is set.

arporter (changes requested):
Looks good Joerg although you'll see I've requested a fair number of comments to explain what's going on - probably just a bit of your helpful explanation from #999. In fact, it might be worth putting some of that in the developer guide too.
I was unable to build the dl_esm_inf submodule with either Intel 17 or 19. Since master of dl_esm_inf is OK, I think this simply requires that the submodule be updated?
I *was* able to build the lfric extraction and nan-check wrappers without problems.

lib/extract/netcdf/Makefile
hiker: I've updated the submodule (I tested the intel compiler on a machine with no outside internet access ... so in short I forgot to copy a recent dl_esm version across). Done now, and it did now compile with Intel 19.

lib/process.py
hiker: Done.

lib/read_only/dl_esm_inf/read_only.f90
hiker: Done.

lib/read_only/lfric/Makefile
hiker: Done.

lib/read_only/read_only_base.jinja
hiker: Done.

hiker:
I've updated the documentation, including a (small) section in the dev guide about the generic interfaces etc., and one or additional files.

doc/developer_guide/psy_data.rst
```diff
@@ -652,6 +652,52 @@ takes the following parameters:
...
+functions. As an example, the ReadOnly verification library for
+LFRic in ``lib/read_only/lfric/`` uses ``-generic-declare`` when
+processing the PSyData base class (i.e. all declaration functions
+are implemented in the PSyData base class), and uses
+``-generic-provide`` when processing the ReadOnly base class (i.e.
+all provide functions are implemented in the ReadOnly base class).
+The LFRic-specific implementation extends the generic subroutine
+with two new subroutines: ``DeclareFieldDouble`` and
+``DeclareFieldVectorDouble`` (and the same for the corresponding
+'provide' functions).
```
arporter, line 695: That's really helpful, thanks Joerg.

arporter (approved):
I can now build the various dl_esm_inf wrapper libraries with Intel 17.
A very useful section has been added to the Developer Guide and this builds fine.
All tests are unaffected.
Examples with compilation are fine.
Not something for this PR but we do need to think about how we test this functionality.

Issue #874, opened 2020-08-25
Remove support for the "old style" arg_type metadata


TeranIvy:
Issues #774 and #817 change `arg_type` metadata by introducing the `GH_SCALAR` for scalar argument names and metadata for Fortran primitive types for all arguments. These issues also make sure that we keep supporting the "old style" metadata to support undisturbed PSyclone development. However, that support needs to be removed prior to move to new PSyclone release and that is the job of this issue.

PR #1044, opened 2021-01-08
Add data type metadata for fields and operators to test kernels (part of #817) 

TeranIvy:
Another PR towards fully resolving #817. It updates field and operator metadata in the LFRic `src/psyclone/tests/test_files/dynamo0p3` directory with the data type descriptors, except for 3 kernels:
- `columnwise_op_asm_kernel_mod.F90` (operator and column-wise operator),
- `testkern_w2_only_mod.f90` (continuous field writer),
- `testkern_w3_mod.f90`  (discontinuous field writer),

These kernels are left as a control that we can support both the old and the new style field and operator metadata and will be updated in #874 when the support for old-style metadata is removed.

This PR can be reviewed and merged independently of #1045 (updates examples) and #1051 (updates tutorials), but needs to be merged before #1052.


arporter (approved):
This is PR is well contained and only touches Fortran test kernels.
All examples still run OK, incl. with compilation.
Tests are all OK with compilation.
No need to update documentation as this is not a change in functionality.
Good job!

PR #1050, opened 2021-01-13
Remove duplicate pytest in GHA (closes #1048)

### LINKED ISSUE #1048: Can we have a git hook to update the copyright date on a file automatically?
Currently our working practise is to update the copyright date on a file when we touch it (if it is out of date). I'm pretty sure we could have git do this for us via a hook: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

rupertford:


rupertford:
Seems to work as expected (you can check the GHA build logs) and see the summary of times above. Ready for first review from @arporter

rupertford:
I'm a bit concerned that codecov seems to be hanging. It may be to do with this change although there is nothing obvious in the yml.

rupertford:
Ran a GHA a second time and codecov seemed happy so hopefully it was a one off.

arporter (approved):
Small change to avoid running pytest twice for python == 3.8.

Issue #1028, opened 2020-12-08
[psyir] Add support for dependency analysis with structures (derived types)


arporter:
#363 is adding support for Fortran derived types to the PSyIR. However, it's already a big piece of work so the support for dependency analysis has been postponed to this issue.

arporter:
I suspect this is going to be easiest to address as part of #845.

hiker:
@arporter I have started to support symbols for the dependency analysis, and while this initially worked great, I didn't realise that this of course does not work for derived types, since they are a PSyIR node (while symbols are ... symbols). And I am not sure if we should mix them (though we could of course) - eg. we need to be able to sort them (admittedly mostly for debugging - print all variables alphabetically) - for now I added a ``__lt__`` method to ``Symbol``.

So the question is - what should I use as 'key' (atm the variable name) in the dependency analysis? I see some obvious solutions:
1. in case of a derived type ``a%b...``, we use the symbol ``a`` to capture the access, and add a ``member_accessed`` (or ``psyir_node``) method, which would give us the rest. So an access like ``a%b(i,j)`` would be listed as access to ``a``, with index ``[i, j]``, and member one of ``a%b(i,j)``, ``a%b``, or ``%b``  - not sure which one is possible and would be best? Problem: it would be kind of hard to find all access to "a%b" (you would have to loop over all accesses to ``a``, then use the member access an analyse it further).
2. We change the key to be a PSyIR node. So ``a=b`` would be using a ``Reference`` for ``a`` (and ``b``) instead of a symbol etc, and we could then use the corresponding PSyIR class to reference ``a%b``. The huge problem with this: as far as I can tell if we have say several assignment to a scalar variable ``a``, each one is an instance of Reference - the references would be different, so instead of seeing one variable with several assignments, we would get several variables (with the same name, same symbol, but different psyir-instances).
3. Maybe keep on using strings as key ("a%b"), and add a method to get symbol and/or psyir node?
4. Maybe use a tuple of symbols? ``a%b%c`` would be stored as ``(a,b,c)`` - each a symbol. But then, there is no symbol ``b`` or ``c`` - we might be able to handle this by adding a symbol table to the ``VariableAccessInfo`` object ... but that smells so hacky 👿  

[...]

PR #1053, opened 2021-01-14
Add derived-type support to Fortran frontend (towards #363)

arporter:


arporter:
Much smaller PR this time. Just adds frontend support. Unfortunately, it does remove the little support we had for dependence analysis for derived-type accesses but there's #1028 for that.

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
```diff
@@ -176,11 +178,163 @@ def test_derived_type_accessibility():
...
     processor.process_declarations(fake_parent, fparser2spec.content, [])
     sym = symtab.lookup("my_type")
     assert isinstance(sym, TypeSymbol)
     flag = sym.datatype.lookup("flag")
     assert flag.visibility == Symbol.Visibility.PRIVATE
     scale = sym.datatype.lookup("scale")
     assert scale.visibility == Symbol.Visibility.PUBLIC
+
+
+def test_derived_type_ref(f2008_parser):
```
sergisiso, line 191: These are both nice tests. Would it be possible to add a "member" property in structure_referece.py that returns "children[0]" for the semantic navigations in this tests (e.g `assing.lhs.member.member`)

There will still be a semantic navigation property missing for the array nodes, but this is a bit more difficult because it can be 1 or more nodes (maybe always return a list) and they are represented by different children (will need to be implemented in multiple nodes). But is something to consider.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -2642,32 +2642,8 @@ def test_nodes_to_code_block_2(f2008_parser):
 
 
 # (3/4) fparser2reader::nodes_to_code_block
```
sergisiso, line 2644: Now this 3 expressions should have the denominator 3

sergisiso (changes requested):
@arporter This PR is almost ready, there is just a few in-line comments.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2683,6 +2685,68 @@ def _assignment_handler(self, node, parent):
...
+        # First we construct the full list of 'members' making up the
+        # derived-type reference. e.g. for "var%region(1)%start" this
+        # will be [("region", [Literal("1")]), "start"].
+        members = []
+        for child in node.children[1:]:
+            if isinstance(child, Fortran2003.Name):
+                # Members of a structure do not refer to symbols
+                members.append(child.string)
+            elif isinstance(child, Fortran2003.Part_Ref):
+                # In order to use process_nodes() we need a fake parent node
```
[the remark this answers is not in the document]
arporter: Good idea, done.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2683,6 +2685,68 @@ def _assignment_handler(self, node, parent):
...
+                # because array-index expressions must refer to symbols.
+                sched = parent.ancestor(Schedule, include_self=True)
+                fake_parent = ArrayReference(parent=sched,
+                                             symbol=Symbol("fake"))
+                self.process_nodes(parent=fake_parent,
+                                   nodes=child.children[1].children)
+                members.append((child.children[0].string,
+                                fake_parent.children))
+            else:
+                raise NotImplementedError(str(node))
```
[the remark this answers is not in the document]
arporter: Yes, done.

src/psyclone/psyir/nodes/array_of_structures_reference.py
```diff
@@ -118,12 +120,16 @@ def create(symbol, indices=None, members=None, parent=None):
                 "An ArrayOfStructuresReference must refer to a symbol of "
                 "ArrayType but symbol '{0}' has type '{1}".format(
                     symbol.name, symbol.datatype))
-        if not isinstance(indices, list) or not indices:
+        if indices is None:
```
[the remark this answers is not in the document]
arporter: Previously `indices` was mandatory but when I came to implement the frontend support in this PR I realised that I wanted to use `process_nodes()` to handle the array index expressions. In order for that to work, I needed to be able to create the parent `ArrayOfStructuresReference` first. Therefore, I had to alter `create()` so that `indices` is an optional argument. `members` is always mandatory. Does that make sense?

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
arporter: That's a good idea. As you suggest, I'm also attempting to add an `indices` property to the array nodes but this does lead to a lot of duplication - `ArrayMixin`, `ArrayOfStructuresReference` and `ArrayOfStructuresMember` all need implementations. I could avoid this by having the implementation in ArrayMixin check on `type(self)` and thus decide whether or not to include the first child in the list of returned nodes. What do you think?

arporter: Perhaps we should have an `ArrayOfStructures` abstract base class?

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Oh yes! This is why I don't like numbering things :-)

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
sergisiso: To be honest I think I would live with the duplication here, which is just twice for each case: ArrayReference and ArrayMember return children starting by number 0 and ArrayOfStructuresReference and ArrayOfStructuresMember return children starting by number 1.

The reason I am not convinced of implementing it on ArrayMixing checking type(self) is that this seems to convey the same information available in `_children_valid_format` and the `_validate_child` method (which child is what). And since it is the same information maybe better to encapsulate in the same place.

What do you think?

src/psyclone/psyir/nodes/array_of_structures_reference.py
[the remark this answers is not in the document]
sergisiso: For `members`  here by mandatory I was meaning removing the None in the `members=None` method signature. But I see this can't be done unless indices has the None removed first (as I still want to keep this ordering).

So then, about indices, the fparser2 can be done once again with a fakeparent ArrayReference node trick, right? I see this is not the nicest trick but then it's a question that if the create method should be nicer (always finish with a consistent/complete state) or the fparser2 frontend should have nicer code (avoid fakeparent tricks). One could argue that the fparser2 already uses this trick often but I don't know where the balance should be.  I will let you decide.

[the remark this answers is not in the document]
arporter: Oh yes - I only thought of the fake parent trick after I'd implemented this. I've changed the code now and made `indices` and `members` mandatory.

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
arporter: I've gone for a new `ArrayOfStructures` mixin class which provides appropriately specialised `_validate_child` and `indices` methods. That way we have pretty much no code duplication.

arporter:
This is ready for another look. I'm afraid it's grown quite a bit due to the addition of the `member` and `indices` properties.

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
sergisiso: Thanks for doing the extra work, I like how it looks!

src/psyclone/psyir/nodes/array_member.py
```diff
@@ -77,8 +77,12 @@ def create(member_name, indices, parent=None):
...
-            member_name, indices=indices, parent=parent)
+        obj = ArrayMember(member_name, parent=parent)
+        # Add any array-index expressions as children
+        for child in indices:
+            obj.addchild(child)
+            child.parent = obj
+        return obj
 
     @staticmethod
     def _validate_child(position, child):
```
sergisiso, line 88: Can we move this (and ArrayReferece._validate_children) inside ArrayMixin now that only these 2 classes use it directly?

src/psyclone/psyir/nodes/array_of_structures_member.py
```diff
@@ -87,25 +85,15 @@ def create(member_name, indices, inner_member, parent=None):
...
-        :rtype: bool
-
-        '''
-        if position == 0:
-            # The first child must be a Member
-            return isinstance(child, Member)
-        # All subsequent child must be array-index expressions
-        return isinstance(child, (DataNode, Range))
+        obj = ArrayOfStructuresMember(member_name, parent=parent)
+        # Add any child Member as the first child
```
sergisiso, line 89: Maybe "Add the provided inner_member node as the first child"?

src/psyclone/tests/psyir/nodes/structure_reference_test.py
```diff
@@ -180,3 +180,20 @@ def test_reference_accesses():
...
+        ("nx", symbols.INTEGER_TYPE, symbols.Symbol.Visibility.PUBLIC)])
+    grid_type_symbol = symbols.TypeSymbol("grid_type", grid_type)
+    ssym = symbols.DataSymbol("grid", grid_type_symbol)
+    # Reference to scalar member of structure
+    sref = nodes.StructureReference.create(ssym, ["nx"])
+    assert sref.member is sref.children[0]
+    # Break the first child to check that we get the expected error
+    sref._children = ["broken"]
+    with pytest.raises(InternalError) as err:
+        sref.member
```
sergisiso, line 197: You can prepend `_ = ` to avoid the pylint complain.

src/psyclone/tests/psyir/nodes/structure_member_test.py
```diff
@@ -124,3 +124,22 @@ def test_sm_child_validate():
...
+    grid_ref = nodes.StructureReference.create(grid_var, ['area', 'nx'],
+                                               parent=assignment)
+    smem_ref = grid_ref.member
+    assert isinstance(smem_ref, nodes.StructureMember)
+    assert isinstance(smem_ref.member, nodes.Member)
+    assert smem_ref.member.name == "nx"
+    # Break the node's children to check the exception
+    smem_ref._children = ["wrong"]
+    with pytest.raises(InternalError) as err:
+        smem_ref.member
```
sergisiso, line 143: You can prepend `_ =` to avoid the pylint complain.

src/psyclone/tests/psyir/nodes/structure_reference_test.py
```diff
@@ -180,3 +180,20 @@ def test_reference_accesses():
         dref.reference_accesses(var_access_info)
     assert ("Dependency analysis has not yet been implemented for "
             "Structures." in str(err.value))
+
+
+def test_struc_ref_semantic_nav():
+    ''' Test the 'member' method of the StructureReference. '''
```
sergisiso, line 186: s/method/property/

src/psyclone/psyir/nodes/structure_reference.py
```diff
@@ -210,6 +214,22 @@ def __str__(self):
...
+        '''
+        :returns: the member of the structure that this reference is to.
+        :rtype: :py:class:`psyclone.psyir.nodes.Member`
+
+        :raises InternalError: if the first child of this node is not an \
+                               instance of Member.
+        '''
+        if not isinstance(self.children[0], Member):
+            raise InternalError(
+                "StructureReference malformed or incomplete. The first child "
```
sergisiso, line 228: Since this can be inherited by AoSReference, can you replace "StructureReference" to `type(self).__name__` or something similar.

src/psyclone/psyir/nodes/structure_member.py
```diff
@@ -101,6 +102,22 @@ def _validate_child(position, child):
...
+        '''
+        :returns: the member of the structure that is being accessed.
+        :rtype: (sub-class of) :py:class:`psyclone.psyir.nodes.Member`
+
+        :raises InternalError: if the first child of this node is not an \
+                               instance of Member.
+        '''
+        if not isinstance(self.children[0], Member):
+            raise InternalError(
+                "StructureMember malformed or incomplete. The first child "
```
sergisiso, line 116: Since this can be inherited by AoSMember, can you replace "StructureMember" to `type(self).__name__` or something similar.

sergisiso (changes requested):
@arporter The ArrayOfStructureMixin turned out to work very well. Getting rid of the odd ArrayMixin._create_array_member and having a generic way to get the index children with `self.indices` improves the readability significantly imo. Well done.

There are only some minor comments left.

doc/user_guide/psyir.rst
```diff
@@ -159,8 +159,13 @@ To solve this issue some Nodes also provide methods for semantic navigation:
    .. automethod:: psyclone.psyir.nodes.IfBlock.if_body()
 
    .. automethod:: psyclone.psyir.nodes.IfBlock.else_body()
+- ``Array`` nodes (e.g. ``ArrayReference``, ``ArrayOfStructuresReference``):
+   .. automethod:: psyclone.psyir.nodes.ArrayReference.indices()
 - ``Directive``:
    .. automethod:: psyclone.psyGen.Directive.dir_body()
+- Nodes representing accesses of data within a structure (e.g.
+  ``StructureReference``, ``StructureMember``):
+   .. automethod:: psyclone.psyir.nodes.StructureReference.member()
```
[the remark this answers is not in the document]
arporter: Oh yes, done.

src/psyclone/psyir/nodes/array_member.py
arporter: Oh yes, that's nicer. I also realised I could move reference_accesses too. I had a bit of trouble with the MRO so had to alter the order with which the classes inherit from e.g. `Member` and `ArrayMixin`.

src/psyclone/psyir/nodes/array_of_structures_member.py
arporter: Yes, that's better.

src/psyclone/tests/psyir/nodes/structure_reference_test.py
arporter: Done.

src/psyclone/tests/psyir/nodes/structure_member_test.py
arporter: Done.

src/psyclone/tests/psyir/nodes/structure_reference_test.py
arporter: Done.

src/psyclone/psyir/nodes/structure_member.py
arporter: Oh yes, done.

src/psyclone/psyir/nodes/structure_reference.py
arporter: Oops, done.

arporter:
Ready for another look now. In bringing the branch up to date with master I gained a couple of test failures. As a result I've had to modify some other files. I also took the opportunity to replace one of the nemo test f90 files with in-line code since it was only used by two tests.

sergisiso (approved):
@arporter All done now. All new functionality is tested,  all tests pass (also with compilation), documentation builds, no pycodestyle (there was one on master that I fixed here) of pylint issues . This is ready for merging.

Issue #846, opened 2020-07-15
Add profile wrapper for LFRic timer


hiker:
This wrapper might become part of LFRic code (to simplify linking issues).

The one issue is how to make it easy to compile the wrapper (till #813 or #745, which might add enough infrastructure code). For now I suggest to add two LFRic infrastructure files (timer_mod, and constants_mod) locally (which I think is all that is required for compiling a wrapper library) .

hiker:
What about the following solution:
1) within lfric, the (existing) profile transformation is used like this (note we have #777 to be able to provide the prefix once in the constructor only, instead of in each apply):
```
    profile = ProfileTrans()
    _, _ = profile.apply(some_code, {"prefix": "lfric_profile"})
```
2) The Fortran wrapper code is written like this:
```
! In psyclone we compile with -DPREFIX=profile, but in LFRic PREFIX would be undefined:
#ifndef PREFIX
#  define PREFIX lfric_profile
#endif

! Note that `a##b` does not work in gfortran, and the official answer 'a/**/b' does not expand macros, 
! and needs more ifdefs depending on compiler ... but this seems to work just fine:

#define paste(a) a
#define CONCAT(a,b) paste(a)b

! This will then result in profile_psy_data_mod in PSyclone, and lfric_profile_psy_data_mod in LFRic:
module CONCAT(PREFIX,psy_data_mod)  
...
```
This way we can use the same Fortran wrapper file in LFRic (using `lfric_profile` as prefix) and PSyclone (using just `profile` as prefix). This way LFRic can use whatever other profiling library they want additionally. This needs to be tested with other compilers/pre-processors.

But two disadvantages:
1) when using the lfric profiling in LFRic, they must remember to set the prefix :(
2) you can't use the command line options for adding profiling :(

Alternative 1:
We add an lfric_profile_trans (which just changes the prefix). This fixes the first problem (remember to set the prefix), but might be more confusing since then they would have to use the normal profile transformation when using non-lfric profiling, but the lfric_profile_trans when using the lfric timer. But it would make it hard to describe this in the PSyclone user's guide (I would prefer to have PSyclone 'clean', i.e. everything only using profile as prefix, and hiding the special handling of lfric_profile as much as possible).

Alternative 2:
We provide an lfric_profile_trans in LFRic (in the optimisation directory)??

Alternative 3:
[...]

PR #1058, opened 2021-01-19
#917 Trigger compilation of PSyData profile wrapper libraries.

### LINKED ISSUE #917: [PSyData] Profiling example does not compile wrapper libraries if required
Other example trigger a build in the libraries included in psyclone (e.g. dl_esm_inf, or psydata wrapper libaries), but this example does not.

hiker:
Fixed #917. Note that there is some overlap with #846 (since it adds another profile wrapper, which can be used stand-alone as well). My suggestion would be to wait merging this till #846 is in, and once this is done, I'll update this PR to support LFRic.


arporter (changes requested):
Looks good. I tried building for dl_timer (which I have installed in a different location to you) and the Makefile wouldn't pick up the fact that I'd set DL_TIMER_ROOT. I got it to work by removing the `dl_timer:` before the conditional assignment. Would you mind investigating?

hiker:
> Looks good. I tried building for dl_timer (which I have installed in a different location to you) and the Makefile wouldn't pick up the fact that I'd set DL_TIMER_ROOT. I got it to work by removing the `dl_timer:` before the conditional assignment. Would you mind investigating?

OK, it looks like the 'target-specific settings' (whatever they are exactly called, can't remember now) are NOT passed on to a child makefile. I've worked around this issues.

hiker:
Ready, I hope :)

arporter (approved):
Build problem has been resolved.
Examples all OK, including with compilation.

PR #1063, opened 2021-01-20
Basic support for specialisation of symbols (towards #935)

### LINKED ISSUE #935: Replace generic PSyIR kernel argument data symbols with LFRic-specific PSyIR data symbols
In #901 we added LFRic-specific PSyIR datatypes and datasymbols and used them to check that kernel subroutine arguments are consistent with the expected kernel arguments, as specified by the kernel metadata. The next step is to take the generic PSyIR representation of kernel arguments and change them into LFRic-specific PSyIR. This will allow us to reason more about the kernel implementation (adding more domain knowledge) and is a stepping stone to higher level LFRic-specific PSyIR - at least that is what I currently think!

rupertford:


rupertford:
I have a basic working version now but there are a number of options

1) What needs to be done with precision? Precision should be set to i_def, r_def etc. in the psyir so perhaps nothing, other than kernel argument validation.
2) At the moment a symbol will update the datatype it references. An alternative would be for a datatype to have its own `specialise` and _`specialise_remote_symbol` method? I think I prefer the way it is at the moment.
3) What should the subclass method be called i.e. is `specialise_remote_symbol` OK?
4) Should the `_specialise_remote_symbol` method check the properties (and the values?) of the parent object are consistent with this subclass? At the moment we assume that validation has checked they are the same.
5) Should `_specialise_remote_symbol` check that the object passed in is the exact expected type? At the moment the checking is done by the `specialise` method.
6) Should I write something more generic that just copies over the differences in properties? I don't think this would work with updating datatype objects as well.
7) lfric psyir specifies the literal values that dimension basis functions and differential basis function as an LfricDImension. When should this be updated?. It seems simpler to update this separately to the `specialise` method.

rupertford:
@sergisiso, I think this is ready for you to take a look at what I've done. More tests would need to be added for the overall output and some more translation needs to be done but a specialisation structure is implemented.

sergisiso:
@rupertford I still would like to disscuss KernelInterface but I now see what I would do with 'specialise' regardless of the discussion.

I didn't expect the specialize argument to be another symbol and to have the `_specialise_remote_symbol`, I was expecting it to had a sub-class of the current symbol as argument.

e.g. `symbol.specialise(CellPositionDataSymbol)`

And I still think it should be this way because this method is needed in other parts where we have initially just created a Symbol and later with additional information we want to enrich the Symbol. 

I believe what you do in psyir_to_lfric() in the statement:
 - `kern_code_arg.specialise(interface_arg)`

 is really 2 steps:
 -  `kern_code_arg.specialise(type(interface_arg))`
 -  `kern_code_arg.copy_properties(interface_arg)`

Note that the 'copy_properties' already exist* for DataSymbol (it may need overriding in some subclasses with additional properties (for the fs_name?)). Also note the interface property should be deleted from copied properties, I think this is true for the other use cases of the method.

Then the specialise implementation looks like yours but with a symbolclass not an instance (and why not promoted to the generic Symbol?).

```
class Symbol:
	def specialise(symbol_class):
		# Check symbol_class is subclass of type(self) 
		self.__class__ = symbol_class
```
Would this work?

sergisiso:
The part that is not necessary for this PR (because properties are copied from another symbol) is what to do with additional required (and/or optional) attributes of the subclasses:

e.g :

```
a = Symbol("name") 
# And later we know that "name" is an INTEGER DatatSymbol (the datatype is mandatory in DataSymbols)
a.specialise(DataSymbol, datatype=INTEGER_TYPE)
```

and for future reference, I think this is possible with something like:

```
class Symbol:
	# Subclasses shouldn't override this constructor
	def __init__(name, args, Kawgs):
		_init_attributes(args, kwargs)
	
	def _init_attributes(visibility, interface):
		# All attributes are consumed here
		# Same code as current Symbol.__init__
	
	def specialise(symbol_class, kwargs):
		# Check symbol_class is subclass of type(self) 
		self.__class__ = symbol_class
		self._init_attributes(kwargs)

class DataSymbol():
	def _init_attributes(datatype, constant_value, kwargs):
		# Consumes datatype and constant_value, the rest are passed to parent
		super(DataSymbol)._init_attributes(kwargs)
		# Same code as current DataSymbol.__init__

```

rupertford:
Thanks @sergisiso. I like your suggestion.

As things are currently implemented I don't think your suggestion is possible as we don't want to blindly copy all properties from one instance to the other (as there are existing references e.g. to other symbols that we don't want to change) and we are not able to selectively copy any additional properties without knowing what the additional properties are (without either looking at the differences between the classes - which would break the separation, or deferring to the sub-class - which is what I did).

I think the way forward would be to use your constructor approach. To do this I would need to create a new sub-class of `arg_ordering` as that is where the additional properties of subclasses is known.

The additional problem is that, depending on the specialisation, we may want to modify some of the existing properties of the instance. Our separate discussion might remove the need for this which would make life easier, but assuming not, this would be another reason to subclass `arg_ordering`.

arporter:
I like Sergi's suggestion. I think it covers my use case (which I'm having to do quite a lot) e.g.:

                if type(sym) in [Symbol, TypeSymbol]:
                    # TODO #935
                    sym.__class__ = TypeSymbol
                    sym.__init__(sym.name, UnknownFortranType(str(decl)),
                                 visibility=dtype_symbol_vis)

rupertford:
OK, given the feedback and the suggested way forward I'm repurposing this PR to add in a simple specialise method that updates the class. This will solve the case where the specialisation requires no additional properties. So a much smaller step, hopefully in the right direction. I'm guessing it won't yet support what @arporter needs.

rupertford:
Ready for review from @sergisiso, subject to the CI being happy.

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -255,6 +255,32 @@ def copy(self):
         return type(self)(self.name, visibility=self.visibility,
                           interface=self.interface)
 
+    def specialise(self, subclass):
+        '''Specialise this class instance so that it becomes an instance of
```
sergisiso, line 259: Suggestions to improve readability (subjective, I won't mind if not taken):
s/class instance/symbol/
s/references to this instance/references to it/

Also show intent to the param, e.g.: the new subclass that will describe this symbol.
and type: \`type\` of a Symbol subclass

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -255,6 +255,32 @@ def copy(self):
...
+        except TypeError as info:
+            message = ("The specialise method in the Symbol class expects the "
+                       "subclass argument to be a class.")
+            six.raise_from(TypeError(message), info)
+        # pylint: disable = unidiomatic-typecheck
+        if not is_subclass or type(self) is subclass:
+            raise TypeError(
+                "The specialise method in the Symbol class expects the "
+                "subclass argument to be a class that is a sub-class of "
+                "Symbol.")
```
sergisiso, line 281: "is a sub-class of this Symbol"? Also in docstring

I am also wondering if `type(self) is subclass` is needed. It depends of how we consider a `.specialise(` in the code: as something that definitely happens, or just a guarantee that it is this subclass (if it already is it continues normally). What are your thoughts?

sergisiso (changes requested):
@rupertford See in-line suggestions. I am currently undecided if a specialise call to a class that it already is should raise an exception or not.

Also it will be useful a few lines in the developer guide explaining the motivation for this, something in the line of: the PSyIR symbols are highly interconnected with references, which makes symbol object changes complex to keep consistent, this solution allows for in-place class enrichment without breaking the references...

I think the User Guide can wait until we complete the interface with the parameters?

sergisiso:
I will merge your other PR #1093 that has a TODO about the specialisation. This will need to brought to master and fix the TODO here.

src/psyclone/psyir/symbols/symbol.py
rupertford: Updated. You probably won't like my definition of the type of subclass. The issue is that the type of all classes is `type` so I feel the need to point this out.

src/psyclone/psyir/symbols/symbol.py
rupertford: I think "of Symbol" is right here as I am talking about the Symbol class of which there is only one, so there is no "this".

It seemed like it would be an error if someone called specialise and did not change the class - as then there is no specialisation. I therefore thought it would be better to report this rather than allow it. Also, I was worried for the case where we allow constructor arguments. It would allow `specialise` to update some of the properties of a class without changing its type. That was my reasoning, but I appreciate there are situations where x is considered to be a subthingy of x and in the end it depends on what is most convenient for the user.

rupertford:
I've tried to address your comments but you may disagree and/or want to continue discussing some of the points. I've updated the TODO and added some use and developer documentation. I was in two minds about adding something to the modify.py example code seeing as I've added it to the user guide but in the end decided not to. Let me know if you think I should add something there - on reflection I think I probably should.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3268,12 +3268,9 @@ def _call_handler(self, node, parent):
         try:
             routine_symbol = symbol_table.lookup(call_name)
             if type(routine_symbol) is Symbol:
-                # TODO PR #1063: Symbol should be specialised to a
-                # RoutineSymbol here (if the symbol is part of a use
-                # statement). Without specialising, the Call class
-                # constructor will raise an exception. As a temporary
-                # fix, just change the class name.
-                routine_symbol.__class__ = RoutineSymbol
+                # Specialise routine_symbol from a Symbol to a
```
sergisiso, line 3271: In this case if the lookup wrongly returns a DataSymbol, it will not break here and I think it should. Also there is a pylint issue. Maybe change to:
```
if not isinstance(routine_symbol, RoutineSymbol):
    routine_symbol.specialise(RoutineSymbol)
```

sergisiso (changes requested):
@rupertford Thanks a lot for the documentation, it describes very well the design decision.

I have continued some open discussions.

I agree there should be an example in modify.py (it would be good to have an example for each subsection of the Modify PSyIR section of the UG and you added one in this PR)

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -255,6 +255,32 @@ def copy(self):
...
                           interface=self.interface)
 
+    def specialise(self, subclass):
+        '''Specialise this symbol so that it becomes an instance of the class
+        provided in the subclass argument. This allows this instance
+        to become a subclass without any references to it becoming
+        invalid.
+
+        :param subclass: the class that this symbol will become.
+        :type subclass: `type` (as `type` is the type of all classes)
```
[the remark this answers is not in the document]
rupertford: Unless I'm missing something I don't think type-hinting would help this particular problem as it would not help us describe and render the interface in Sphinx. I do like the idea of type hinting though, although I'm not sure how it fits in with interface documentation. Perhaps you could bring it up at the next telco?

What I've done is describe the class, rather than the type of the class. It is not strictly correct - I don't think - but is probably what people would expect to see - perhaps!

src/psyclone/psyir/symbols/symbol.py
rupertford: You're quite right. I've hopefully updated appropriately.

src/psyclone/psyir/frontend/fparser2.py
rupertford: Yes, good point. I've updated to check for invalid types and raise an exception if so.

rupertford: Also added a pylint disable-warning directive as I think that `type` is the right thing to use here.

rupertford:

src/psyclone/psyir/symbols/symbol.py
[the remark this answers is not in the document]
sergisiso: @rupertford I didn't explain it well. I was not suggesting type hinting yet (I don't think python2 would like it), I was suggesting using elements of its syntax for the ":type" fields of the docstring as it has a proper definition of each case and I think Sphinx is implemented to recognize this syntax. But it doesn't really matter for this issue, maybe a future discussion.

Anyway, I do agree with you that the plain class is not strictly correct (as it not the instance we are passing here).
Since this is the only issue remaining I will change it before merging from the current:
``` :type subclass: sub-class of :py:class:`psyclone.psyir.symbols.Symbol```
to 
``` :type subclass: type of sub-class of :py:class:`psyclone.psyir.symbols.Symbol```
unless you disagree.

sergisiso (changes requested):
@rupertford The code is good now (beside the last comment remaining - could you add the "type of"), pycodestyle and pylint don't show new issues, new code is covered and new features are documented.

The only part missing is adding an example of specialising a symbol in ` examples/psyir/modify.py` the this PR will be ready.

src/psyclone/psyir/symbols/symbol.py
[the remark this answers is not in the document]
rupertford: Updated.

rupertford:

sergisiso (approved):
@rupertford All requested changes done, this is ready for merging.

Issue #1076, opened 2021-01-25
Update .gitignore file


hiker:
My editor stores project- and workspace-files in the working directory, which should be ignored.
I also noticed a ``.idea`` folder - no idea where this comes from, but I've added that, too.

rupertford:
PR #1078 is now on master. Closing this issue.

PR #1078, opened 2021-01-25
#1076 Ignore sublime project files and .idea.

hiker:
Trying again - this time properly based on master. Sorry, missed that I was on 1022.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ dist
 src/*.egg-info
 .mypy_cache
 .ipynb_checkpoints
+*.sublime-project
+*.sublime-workspace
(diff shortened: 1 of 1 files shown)


hiker:
Ah, while in general the .idea is some kind of java ide thing, it is apparently used by pycharm. That might explain why I have this folder. Still, I think it's the right thing to ignore (strangely it has a .gitignore file inside)

rupertford (approved):
Simple additions to .gitignore for pycharm (apparently!).

rupertford:
As this is just a change to a file that tells git which files to ignore I'll not update the changelog.

PR #1092, opened 2021-01-29
Bug fix for auto-invoke profiling and return statements (closes #1091)

### LINKED ISSUE #1091: Auto-invoke profiling includes Return nodes.
While testing PSyclone master with NEMO I've discovered that the auto-profiling code does not check for Return nodes in the region to be enclosed. Since the profiling transformation itself *does* check this causes failures. I have a simple fix that does two things:
 1. For the common case where a routine finishes (with a superfluous) 'return', this is removed from the list of nodes passed to the profiling transformation;
 2. For anything more complicated (more than one 'return' or 'return' not the last statement) the code quietly skips adding profiling.

Having written that, I can see we need to log situation 2.

arporter:


arporter:
Probably one for @hiker if you don't mind - I could do with your feedback on whether there's some logging mechanism I could use to flag the fact that we're refusing to add profiling.

hiker:
> Probably one for @hiker if you don't mind - I could do with your feedback on whether there's some logging mechanism I could use to flag the fact that we're refusing to add profiling.

I'll try to do the review tomorrow, but re logging: I am not aware of anything. The tool class used for automatic parallelisation has some minimal logging functionality, but it's basically just some functions add_warning, add_error, and a getAllMessages() - around 20 lines of code, and the printing is done by the caller. For now the only option I can see is printing a warning to stderr :(

src/psyclone/tests/psyir/transformations/profile_test.py
```diff
@@ -814,3 +816,56 @@ def test_omp_transform():
...
+    a return statement within the profiling region if it is the last statement
+    in the routine. '''
+    symbol_table = SymbolTable()
+    arg1 = symbol_table.new_symbol(
+        symbol_type=DataSymbol, datatype=REAL_TYPE)
+    zero = Literal("0.0", REAL_TYPE)
+    assign1 = Assignment.create(Reference(arg1), zero)
+    kschedule = KernelSchedule.create(
+        "work", symbol_table, [assign1, Return()])
+
```
hiker, line 832: I have I don't like tests that assemble their own PSyIR tree like this - I find it very hard to understand, I usually put a simple subroutine in a string and parse it. This way I find the test easier to understand, and it feels to me the test is less dependent on implementation details (e.g. you have dependencies to SymbolTable, REAL_TYPE, Literal, ... - while parsing code would only have the normal parsing interface).

Having said that, this test is fine and does what it is supposed to test, and I don't see a need to change it.

src/psyclone/tests/psyir/transformations/profile_test.py
```diff
@@ -814,3 +816,56 @@ def test_omp_transform():
...
+    Profiler.set_options([Profiler.INVOKES])
+    Profiler.add_profile_nodes(kschedule, Loop)
```
hiker, line 834: Maybe add an assert here that ``kschedule[-1]`` is indeed a return statement. Yes, obvious - but I am always a bit scared when constructing my own tree that something else might change.

src/psyclone/tests/psyir/transformations/profile_test.py
```diff
@@ -814,3 +816,56 @@ def test_omp_transform():
...
+    # The Return should be a sibling of the ProfileNode rather than a child
+    assert isinstance(kschedule[0], ProfileNode)
```
hiker, line 836: One additional assert that theProfileNode contains the assignment? Again, extra careful admittedly - I wouldn't know what else it could contain :)

src/psyclone/tests/psyir/transformations/profile_test.py
```diff
@@ -814,3 +816,56 @@ def test_omp_transform():
...
+    Profiler.add_profile_nodes(kschedule, Loop)
+    # No profiling should have been added
+    assert not kschedule.walk(ProfileNode)
+
+    # Create Schedule with a Return at the end as well as in the middle.
+    kschedule = KernelSchedule.create(
+        "work", symbol_table, [assign1, Return(), assign2, Return()])
+    Profiler.add_profile_nodes(kschedule, Loop)
+    # No profiling should have been added
+    assert not kschedule.walk(ProfileNode)
```
hiker, line 871: *grin* After what I wrote above for the previous test ... I have to admit that this is rather compact and elegant 👍

hiker (changes requested):
I believe two files need their (C) updated - src/psyclone/psyir/transformations/psy_data_trans.py  and src/psyclone/tests/psyir/transformations/profile_test.py , the rest is fine.

I think it would be good to print a message to stderr if no profiling is applied when requesting it (which then also should be tested). A proper logging facility would be great, but can be separate ticket. I seem to remember that we had some contribution that used logging ... ahh -  fparser only, e.g.:
```
        import logging
...
        logging.getLogger(__name__).debug(message)
```
nothing in PSyclone.

I am not sure what state to set this to, since you actually didn't mark it 'ready for review' ... I"ll still set it to 'request change`.

src/psyclone/tests/psyir/transformations/profile_test.py
arporter: I did it this way to be API agnostic (since the profiling routine I'm testing is API agnostic). I admit it's not as simple as just parsing some Fortran though.

src/psyclone/tests/psyir/transformations/profile_test.py
arporter: Yes, done.

src/psyclone/tests/psyir/transformations/profile_test.py
arporter: Done.

arporter:
Assuming the CI is happy, this is ready for another look. I'll set the label properly this time!

hiker (approved):

hiker:
All issues addressed (and bug indeed fixed), tests and compilation successful. No need to update manual.

PR #1105, opened 2021-02-04
Add support for program, module and subroutine to Fparser2Reader (closes #1104)

### LINKED ISSUE #1104: parse modules containing subroutines in one function/method
There is no `fparser2 to psyir` function/method that takes Fortran code and turns it into PSyIR. The actual functionality needed at the moment is to support a module containing subroutines so any implementation could just support this for the moment.

rupertford:


rupertford:
This has not been tidied yet and the tests are not complete but would be worth a first look so the reviewer can decide whether they are happy with the general approach. This could be either @arporter or @sergisiso.

This implementation has highlighted to me that we don't support reading multiple modules/subroutines within a file. I think this could be simply supported with a FileContainer(Container), which would be useful for C/C++ as well, as variables can have file/protected scope. I'm guessing we would also want a ProgramContainer(Container) too to extend this to Programs.

arporter:
I think what you've done is good but I do think we should update `generate_schedule` so that it *doesn't* attempt to create a Container. It should also be renamed `generate_routine` or similar.

rupertford:
> I think what you've done is good but I do think we should update `generate_schedule` so that it _doesn't_ attempt to create a Container. It should also be renamed `generate_routine` or similar.

So, do you think I should do this as part of this PR? Happy to do so if that would be useful.

arporter:
> > I think what you've done is good but I do think we should update `generate_schedule` so that it _doesn't_ attempt to create a Container. It should also be renamed `generate_routine` or similar.
> 
> So, do you think I should do this as part of this PR @arporter? Happy to do so if that would be useful.

Yes, I think this would be a sensible PR to do it in.

sergisiso:
Seeing this routine is comparing and acting on the fparser node type (Fortran2003.Program, Fortran2003.Module, Fortran2003.Subroutine_Subprogram) I am wondering if each condition should be a handler for that node type instead (at least the last 2).

I agree with renaming to "generate_routine" but I would also make sure that the whole generate_container and generate_routine still make sense or (part of them) can also be handlers.

Finally, as a matter of taste, I think I would have this as a method instead of an independent function, it feels more OO. But no problem is you prefer the outside function.

rupertford:
OK, I hear you and agree that handlers are the way forward and I'll push the function into the class. I didn't really want to get into adding all the extra mechanics now, which is why I wrote something that did the required job with the functionality we have now.

rupertford:
After discussion with @arporter, we have decided to leave `generate_schedule` as it is. The functionality being added here is expected to replace that method (and `generate_container`) at some point so there seems to be no reason to make any modifications beyond the new functionality added here.

rupertford:
Ready for first review from @sergisiso. There are a couple of possible open questions that I am aware of (1: should `generate_psyir` allow fparser2 input that does not have program as a parent and 2: is there a better way to deal with the top PSyIR node i.e. not creating and then deleting a dummy Container) but I'll let you pick those up and see what you think.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -903,6 +926,31 @@ def _create_schedule(name):
...
+        :raises GenerationError: if the root of the supplied fparser2 \
+            parse tree is not a Program
+
+        '''
+        if not isinstance(parse_tree, Fortran2003.Program):
+            raise GenerationError(
+                "The Fparser2Reader generate_psyir method expects root the "
+                "supplied fparser2 tree to be a Program, but found '{0}'"
+                "".format(type(parse_tree).__name__))
+        node = Container("dummy")
```
sergisiso, line 948: Since we have guaranteed that the node is a Forrtan2003.Program can we just simply do:
```retrun self._program_handler(parse_tree, None)```

if this is be problematic, the dummy parent node is something we already use a few times in this front-end, so it is fine as a solution.

File: src/psyclone/psyir/frontend/fparser2.py [hunk 2]
```diff
@@ -3290,3 +3326,126 @@ def _call_handler(self, node, parent):
...
+            decl_list = []
+            arg_list = []
+        finally:
+            self.process_declarations(routine, decl_list, arg_list)
+
+        try:
+            sub_exec = first_type_match(node.content,
+                                        Fortran2003.Execution_Part)
+        except ValueError:
+            pass
```
sergisiso, line 3372: Add a comment saying that: Routines with no Execution part are still valid.

src/psyclone/tests/psyir/frontend/fparser2_generate_psyir_test.py
```diff
@@ -0,0 +1,100 @@
...
+from __future__ import absolute_import
+import pytest
+
+from fparser.common.readfortran import FortranStringReader
+from psyclone.errors import GenerationError
+from psyclone.psyir.nodes import Container, Routine, CodeBlock
+from psyclone.psyir.frontend.fparser2 import Fparser2Reader
+from psyclone.psyir.backend.fortran import FortranWriter
+
+MODULE_IN = (
```
sergisiso, line 50: Would it be better to split the python string line after the \n as well for readability in all these code variables?

sergisiso (changes requested):
@rupertford Thanks for doing the extra work of putting everything in handlers even if some code is repeated of the generate_schedule/container. I think its the right way to do it. I made some inline comments. Regarding your open questions: 

> 1: should generate_psyir allow fparser2 input that does not have program as a parent

I think it's OK to leave it as it is for now if it covers the use cases where this is needed. In the future if we replace the generate_schedule we may need to think if its better to extend generate_psyir or create a second entry point to substitute generate_schedule. But I would leave this decision to the future PR. Maybe if an issue about removing or simplifying generate_shcedule/container doesn't already exist it is worth creating it, referencing it and writing a comment about these possibilities. 

>2: is there a better way to deal with the top PSyIR node i.e. not creating and then deleting a dummy Container

Answered inline

src/psyclone/tests/psyir/frontend/fparser2_generate_psyir_test.py
sergisiso: Some in the other test files.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -80,6 +80,26 @@
                   "inout": ArgumentInterface.Access.READWRITE}
 
 
+def first_type_match(nodelist, typekind):
```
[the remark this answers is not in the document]
rupertford: Good point.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -903,6 +926,31 @@ def _create_schedule(name):
         '''
         return KernelSchedule(name)
 
+    def generate_psyir(self, parse_tree):
+        '''Translate the supplied fparser2 parse_tree into PSyIR.
+
+        :param parse_tree: the supplied fparser2 parse tree
```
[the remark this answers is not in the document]
rupertford: Oops.

src/psyclone/psyir/frontend/fparser2.py
rupertford: I originally did `self.process_nodes(None, [parse_tree])` which does the same thing as you suggest. However, in either case the problem is that `process_nodes` always assumes that a PSyIR parent exists and tries to add children to it. Rather than changing the logic, I went with a dummy container. I think something like the following would work though if you would prefer that ...

```
                ...
                if psy_child:
                    if not parent:
                        if not len(nodes) == 1:
                            raise InternalError("xxx")
                        parent = nodes[0]
                    else:
                        self.nodes_to_code_block(parent, code_block_nodes)
                        parent.addchild(psy_child)
                ...
```

src/psyclone/psyir/frontend/fparser2.py
rupertford: Good idea.

src/psyclone/tests/psyir/frontend/fparser2_generate_psyir_test.py
rupertford: I'm guessing that is a rhetorical question and you would not accept the answer no, because more compact code in tests is more important. OK done. Well that was lots of fun!

rupertford:
I've created issue #1115 which suggests the removal of `generate_container` and `generate_schedule`.

rupertford:

src/psyclone/psyir/frontend/fparser2.py
sergisiso: I don't understand this `parent = nodes[0]`, parent is a PSyIR but nodes is fparser2 ast. Also process_nodes currently doesn't return anything, it just gives back the new children by appending them as the parent's children (which is why it always need some parent).

I was recommending the handler because handlers should always return a node (or give an exception). But now I see `_program_hanlder` doesn't. I believe program_handler should have  `return self._create_child(node.items[0], parent)` instead of the process_nodes call (similar to `_parenthesis_handler`). Would it work then?

src/psyclone/tests/psyir/frontend/fparser2_generate_psyir_test.py
sergisiso: I was more on the fence about this than what you thought. But now that it is done I do like the change. Thanks for doing it. 😄

src/psyclone/psyir/frontend/fparser2.py
sergisiso: I know what I am proposing won't work for top-level CodeBlocks (?). So if it requires changing the code too much remember that I am OK with the dummy parent as well.

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
```diff
@@ -0,0 +1,91 @@
...
+        "end module\n")
+    expected = (
+        "module a\n"
+        "  implicit none\n\n"
+        "  contains\n\n"
+        "end module a\n")
+    processor = Fparser2Reader()
+    reader = FortranStringReader(code)
+    parse_tree = parser(reader)
+    container = Container("dummy")
```
sergisiso, line 66: I wouldn't expect this tests to need the "dummy" node after the program_handler is updated to return a node as output. It should look like the module handler test I believe.

sergisiso (changes requested):
@rupertford Almost ready, just pushing back in one remaining open issue.

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
rupertford: Removed thanks to your explanation in a separate discussion.

src/psyclone/psyir/frontend/fparser2.py
rupertford: After discussion, we have agreed to keep the use of a dummy node in this PR and to defer the potential restructuring of `process_nodes` as it would be a separate issue.

rupertford:

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -917,6 +940,31 @@ def _create_schedule(name):
...
+        :rtype: :py:class:`psyclone.psyir.nodes.Container` or \
+            :py:class:`psyclone.psyir.nodes.Routine`
+
+        :raises GenerationError: if the root of the supplied fparser2 \
+            parse tree is not a Program.
+
+        '''
+        if not isinstance(parse_tree, Fortran2003.Program):
+            raise GenerationError(
+                "The Fparser2Reader generate_psyir method expects root the "
```
sergisiso, line 959: Is this error message incomplete?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3341,6 +3377,131 @@ def _call_handler(self, node, parent):
...
+            decl_list = []
+            arg_list = []
+        finally:
+            self.process_declarations(routine, decl_list, arg_list)
+
+        try:
+            sub_exec = _first_type_match(node.content,
+                                         Fortran2003.Execution_Part)
+        except ValueError:
+            # Routines without any execution statemens are still
```
sergisiso, line 3422: "statements"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3341,6 +3377,131 @@ def _call_handler(self, node, parent):
...
+        node of a complete fparser2 tree and may contain one or more
+        program-units. At the moment PSyIR does not support the
+        concept of multiple program-units so can only support
+        one. Therefore this handler simply checks that this constraint
+        is observed in the supplied tree.
+
+        :param node: top level node in fparser2 parse tree.
+        :type node: :py:class:`fparser.two.Fortran2003.Program`
+        :param parent: parent node of the PSyIR node we are constructing.
+        :type parent: :py:class:`psyclone.psyir.nodes.Node`
```
sergisiso, line 3492: return and rtype docstrings missing

sergisiso (changes requested):
@rupertford There are some final docstring/comments modifications needed. Pylint and pycodestyle don't find new issues.

The PR additions are fully tested, although it doesn't feel too unit-testy to call the Writer I do see that the important aspect to the tested methods is that they create the new node and iterate though the relevant process_nodes/declarations/access_statements... and this is seen conveniently with the Writer.

src/psyclone/psyir/frontend/fparser2.py
rupertford: Oops, bad English. It should be "expects the root of the". Modified this and the associated test.

src/psyclone/psyir/frontend/fparser2.py
rupertford: Oops, well spotted.

src/psyclone/psyir/frontend/fparser2.py
rupertford: Oops again.

rupertford:
Good point about using the writer. The problem is that there is not simple way to check the contents of the PSyIR and this way is quite nice as the input to the test is Fortran as is the output so it is simple to see whether it is working and to add new test code if required. We could require the PSyIR to output a concise (repr like) form in the future (or perhaps serialise with json) and then test against this. However, we would probably want to implement the tree traversal as a visitor which would essentially have the same non-unit test issue as the Fortran writer. Something for future discussion?

Assuming @sergisiso is happy that how to check the PSyIR in pytest tests is a separate problem then this PR is ready for next review.

sergisiso (approved):
@rupertford All requested changes made. This PR is ready for merging.

PR #1111, opened 2021-02-07
Support use of Call create method in a subclass of Call (closes #1110)

### LINKED ISSUE #1110: Call create method should be a classmethod not a staticmethod
In the future we may want to subclass a `Call` node and use the `create` method in the `Call` node to create an instance of the subclass. An example of this could be in the generation of a PSyIR version of the algorithm layer for LFRic where invoke calls will contain `builtin` or `coded` calls which are subclasses of `Call` and could be simply created using the `create` method.

rupertford:
Changes the implementation of the `create` method in the `Call` class so that subclasses can use the method to create instances of themselves.


sergisiso (approved):
Besides one small error in the docstring which I corrected, I like this change and I think it should be applicable to most nodes, so in addition to approve this PR I created #1119

PR #1121, opened 2021-02-10
(closes #1120) add bibtex_bibfiles entry to conf.py for UG and DG

### LINKED ISSUE #1120: Dev Guide fails to build on Read the Docs
RTD uses a more recent version of sphinxcontrib-bibtex and it causes the error:

    sphinx.errors.ExtensionError: You must configure the bibtex_bibfiles setting

I think we simply need to add `bibtex_bibfiles = ['../bibliography/references.bib']` to the relevant `conf.py` (assuming that the relative path is OK).

arporter:
Tiny PR that just fixes build errors on Read The Docs. I could reproduce those errors locally by doing `pip install --upgrade sphinxcontrib-bibtex` and then attempting to build the docs.

### DIFF
--- a/changelog
+++ b/changelog
@@ -354,11 +354,16 @@
         114) PR #1054 for #1047. Adds check that the intrinsic types of
         scalar arguments in the LFRic API are consistent. Extends argument-
         filtering routines to additionally filter on intrinsic type.
-	
+
 	115) PR #1102 towards #1031. Improves PSyIR fparser2 frontend support for
 	structures types.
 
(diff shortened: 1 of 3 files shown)


rupertford (approved):
The version of sphinxcontrib-bibtex that was previously working, still works with this change for the user a developer guides. I can  also reproduce the error reported by upgrading to the latest version of sphinxcontrib-bibtex and this change fixes those errors. Changes conform to pycodestyle and pylint - although I note pylint complains quite a bit about other content of the conf.py files. Approving.

PR #1122, opened 2021-02-10
Restructure node colour implementation (closes #1112)

### LINKED ISSUE #1112: Specifying PSyIR node colours
PSyIR nodes can be given a colour so that the output of the `view()` method is more readable. At the moment a node specifies a keyword name in a `_colour_key` variable and a mapping from keywords to colours is provided in the `Node` base class.

This solution does not seem particularly nice as the base class needs to have information from all of its child classes (that want to specify a colour). It is a particular problem when the PSyIR is specialised for a domain as it makes no sense for specialised classes to add keyword names to the Node base class.

One simple solution would be to remove the idea of a keyword and simply specify the colour in the class itself (modifying `_colour_key` to `_colour`. This would then defer the choice of colour to the class itself which seems to make sense.

A (more hacky) solution would be to support generic mappings from a colour to a colour as well as the existing generic PSyIR ones in the `Node` baseclass. For example, the base class could support `"red": "red"` which would allow any domain-specific classes to override their colour explicitly, whilst keeping the same implementation as we have now.

If it is decided that the choice of colour should continue to be separated from the class (to allow the colours to be changed independently) then a more radical solution might be preferred. A mapping would need to be kept, but I think this would be better kept in configuration files (with separate mappings for domain-specific nodes). I'm not sure how that would be implemented though as I think we would need to know which domain a domain-specific node was from (so we could look in the appropriate mapping). Perhaps all domain-specific nodes would need to identify themselves as such via another variable, or they could all inherit from a particular domain-specific class as well as `Node` (i.e. support multiple inheritance)?

Any thoughts, suggestions or preferences @sergisiso and @arporter?

rupertford:


rupertford:
Ready for first review from either @arporter or @sergisiso, subject to GHA being happy.

src/psyclone/tests/psyGen_test.py
```diff
@@ -50,8 +50,9 @@
 import pytest
 from fparser import api as fpapi
 from psyclone.core.access_type import AccessType
+from psyclone.psyGen import GlobalSum, InvokeSchedule
```
arporter, line 53: We now have two separate `from psyclone.psyGen import ...` lines.

src/psyclone/tests/psyGen_test.py
```diff
@@ -386,16 +386,17 @@ def test_kern_get_kernel_schedule():
 
 
 def test_codedkern_node_str():
-    ''' Tests the node_str method in the CodedKern class. The simplest way to
-    do this is via the dynamo0.3 subclass '''
-    from psyclone.psyir.nodes.node import colored, SCHEDULE_COLOUR_MAP
+    '''Tests the node_str method in the CodedKern class. The simplest way
+    to do this is via the dynamo0.3 subclass
```
arporter, line 390: Full stop please.

src/psyclone/tests/psyir/nodes/ifblock_test.py
```diff
@@ -75,10 +75,10 @@ def test_ifblock_node_str():
 def test_ifblock_view_indices(capsys):
     ''' Check that the view method only displays indices on the nodes
     in the body (and else body) of an IfBlock. '''
-    from psyclone.psyir.nodes.node import colored, SCHEDULE_COLOUR_MAP
-    colouredif = colored("If", SCHEDULE_COLOUR_MAP["If"])
-    colouredreturn = colored("Return", SCHEDULE_COLOUR_MAP["Return"])
-    colouredref = colored("Reference", SCHEDULE_COLOUR_MAP["Reference"])
+    from psyclone.psyir.nodes.node import colored
```
arporter, line 78: You now have this import at the top of the file.

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -84,33 +84,27 @@ def test_node_coloured_name():
...
     tnode._text_name = "ATest"
-    tnode._colour_key = "Schedule"
+    tnode._colour = "white"
     assert tnode.coloured_name(False) == "ATest"
-    assert tnode.coloured_name(True) == colored(
-        "ATest", SCHEDULE_COLOUR_MAP["Schedule"])
-    # Check that an unrecognised colour-map entry gives us un-coloured text
-    tnode._colour_key = "not-recognised"
-    assert tnode.coloured_name(True) == "ATest"
+    assert tnode.coloured_name(True) == colored("ATest", "white")
```
arporter, line 89: Probably need to put back/modify the old part of the test for _colour = None. What happens if a developer accidentally mis-spells the name of the colour? I assume we get a reasonable stack-trace in that case but possibly we should catch it. In fact, this relates to a slight concern that I have about this change in that we've gone from having `termcolor`-specific settings (i.e. the names of the colours) in a single location to having them scattered through the source.

src/psyclone/tests/psyGen_test.py
arporter: pylint also complains that `InvokeSchedule` is imported in other places in this file too.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -338,9 +338,9 @@ def node_str(self, colour=True):
         :returns: description of this node, possibly coloured.
         :rtype: str
         '''
-        from psyclone.psyir.nodes.node import colored, SCHEDULE_COLOUR_MAP
+        from psyclone.psyir.nodes.node import colored
```
arporter, line 341: Please move this import up.

arporter (changes requested):
Looks good, I can't remember why I didn't do it this way in the first place! I do have one concern though and that is that we now have termcolor-specific settings (the actual names of the colours) scattered through all of the classes. Realistically, I suspect this is unlikely to ever be a problem.
I do think we should have a defensive check that `self._colour` has been overridden by the sub-class - see in-line.
Apart from that there are just a few minor pylint things.
All tests pass, incl. with compilation. Ditto the examples.
Updated documentation builds fine.
Thanks to the removal of the colour-map dict. the ref. guide now builds with only 4 warnings :-)
I've grepped all source and tests and there are no references to the old way of doing things.
I've checked all modified files for pylint/pycodestyle.

src/psyclone/psyir/nodes/loop.py
rupertford: Done.

src/psyclone/tests/psyir/nodes/ifblock_test.py
rupertford: Oops.

src/psyclone/tests/psyGen_test.py
rupertford: Oops.

src/psyclone/tests/psyGen_test.py
rupertford: Updated.

src/psyclone/tests/psyir/nodes/node_test.py
rupertford: Good point. I've added a check for an invalid colour and an associated test.

src/psyclone/psyir/nodes/node.py
```diff
@@ -361,11 +329,7 @@ def coloured_name(self, colour=True):
                 "given a string value in the concrete class '{0}'."
                 "".format(type(self).__name__))
         if colour:
-            try:
-                return colored(self._text_name,
-                               SCHEDULE_COLOUR_MAP[self._colour_key])
-            except KeyError:
-                pass
+            return colored(self._text_name, self._colour)
```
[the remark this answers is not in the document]
rupertford: Yes good point. I've gone for raising an exception as I think that makes more sense.

rupertford:

src/psyclone/psyir/nodes/node.py
```diff
@@ -329,7 +330,19 @@ def coloured_name(self, colour=True):
...
+            if self._colour is None:
+                raise NotImplementedError(
+                    "The _colour attribute is abstract so needs to be given "
+                    "a string value in the concrete class '{0}'."
+                    "".format(type(self).__name__))
+            try:
+                return colored(self._text_name, self._colour)
+            except KeyError as info:
+                message = (
+                    "The _colour attribute in class '{0}' has been set to an "
```
arporter, line 342: "...set to a colour ('{0}') that is not supported by the termcolor package."

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -80,16 +81,42 @@ def test_node_coloured_name():
...
+            "value in the concrete class 'Node'." in str(err.value))
+    # Valid values
     tnode._colour = "white"
     assert tnode.coloured_name(False) == "ATest"
     assert tnode.coloured_name(True) == colored("ATest", "white")
 
 
+def test_node_coloured_name_exception(monkeypatch):
+    '''Test that the expected exception is raised if the colour provided
+    to the colored function is invalid. Note, an exception if the
```
arporter, line 98: "...an exception is only raised if..."?

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -80,16 +81,42 @@ def test_node_coloured_name():
...
+    tnode._text_name = "ATest"
+    tnode._colour = "invalid"
+    with pytest.raises(InternalError) as err:
+        _ = tnode.coloured_name()
+    assert ("The _colour attribute in class 'Node' has been set to an "
+            "unsupported colour 'invalid'." in str(err.value))
+
+
 def test_node_str():
+
```
arporter, line 119: Please rm this blank line.

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -80,16 +81,42 @@ def test_node_coloured_name():
...
+    termcolor package is installed. Therefore we monkeypatch the
+    function to force the exception whether termcolor is installed or
+    not.
+
+    '''
+    def dummy(error):
+        '''Utility used to raise the required error from a lambda function.'''
+        raise error()
+
+    monkeypatch.setattr(node, "colored", lambda _1, _2: dummy(KeyError))
```
arporter, line 108: I don't think you need to pass in the error class or use a `lambda`:

    def dummy(_1, _2):
        raise KeyError()

    monkeypatch.setattr(node, "colored", dummy)

arporter (changes requested):
Thanks for adding the new checks and extra tests. I think it's good now.
Just a few minor things to fix and then this can be merged.
Tests with compilation are still OK. Ditto examples.

src/psyclone/psyir/nodes/node.py
rupertford: Yes, that's better, thanks.

arporter (approved):
All good now - all requested changes have been made.
Examples and tests all OK with compilation.
Ref guide builds cleanly.

PR #1135, opened 2021-02-27
Add replace_with method to node (closes #1124)

### LINKED ISSUE #1124: Add a Node.replace method or equivalent that replaces one node with another
In PSyclone transformations we might want to remove a node and replace it with another. It would be nice to have a simple way to do this.

rupertford:


rupertford:
Requested first review from @sergisiso

doc/user_guide/psyir.rst
```diff
@@ -579,3 +579,17 @@ would need to be set explicitly.
...
+
+In certain cases one might want to replace a node in a PSyIR tree with
+another node. All nodes provide the `replace_with()` method to support
+this.
+
+.. code-block:: python
+
+    node.replace_with(new_node)
+
+Note, the node and all of its children is replaced with new_node and
```
sergisiso, line 594: I think this is quite significant and should be part of the description above instead of just a note. Maybe:
"All nodes provide the `replace_with()` method to replace any specific node and all its children with another given node and all its children."

src/psyclone/psyir/nodes/node.py
```diff
@@ -1149,6 +1149,40 @@ def scope(self):
             "Unable to find the scope of node '{0}' as none of its ancestors "
             "are Container or Schedule nodes.".format(self))
 
+    def replace_with(self, node):
+        '''Replaces self with the supplied node in the PSyIR tree.
```
sergisiso, line 1153: This sentence seems ambiguous. Maybe: "Replaces self, and all its children, from the PSyIR tree where it belongs with with supplied node."

src/psyclone/psyir/nodes/node.py
```diff
@@ -1149,6 +1149,40 @@ def scope(self):
...
+
+        :param node: the node that will replace self in the PSyIR \
+            tree.
+        :type node: subclass of :py:class:`psyclone.psyir.nodes.node`
+
+        :raises TypeError: if the argument 'node' is not a Node.
+        :raises SymbolError: if this node does not have a parent.
```
sergisiso, line 1160: You meant GenerationError here and line below

src/psyclone/psyir/nodes/node.py
```diff
@@ -1149,6 +1149,40 @@ def scope(self):
...
+        if node.parent is not None:
+            raise GenerationError(
+                "The parent of argument node in method replace_with in the "
+                "Node class should be None but found '{0}'."
+                "".format(type(node.parent).__name__))
+
+        parent = self.parent
+        position = self.position
+        node.parent = parent
+        self.parent.children.remove(self)
```
sergisiso, line 1182: I think this remove operation is dangerous when removing a children and the index position is significant (e.g. children 0 of IfBlock which should be an expression - a Schedule can not be displaced into it). Could you add a test for this?
Anyway, this two lines can be converted to `self.parent.children[position] = node` to avoid the issue.

sergisiso (changes requested):
@rupertford The replace_with() functionality looks good but there is an issue regarding children displacement that you need to consider, also some docstring suggestions.

doc/user_guide/psyir.rst
rupertford: Good point. Updated.

src/psyclone/psyir/nodes/node.py
rupertford: Updaed.

src/psyclone/psyir/nodes/node.py
rupertford: Oops, I was trying different options, one of which was SymbolError, and forgot to update the docstring.

src/psyclone/psyir/nodes/node.py
rupertford: That's much simpler, thanks. If I understand correctly, there is no additional test required now that I've changed to your proposed much simpler solution.

rupertford:
Ready for next review from @sergisiso subject to CI being happy.

src/psyclone/psyir/nodes/node.py
```diff
@@ -1149,6 +1149,40 @@ def scope(self):
             "Unable to find the scope of node '{0}' as none of its ancestors "
             "are Container or Schedule nodes.".format(self))
 
+    def replace_with(self, node):
+        '''Replaces self with the supplied node in the PSyIR tree.
+
+        :param node: the node that will replace self in the PSyIR \
+            tree.
+        :type node: subclass of :py:class:`psyclone.psyir.nodes.node`
```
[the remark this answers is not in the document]
sergisiso: I don't know if there is a good reason. Its probably a docstring formatting decision, but in most of the code when we specify a generic concept (a least for psyir.nodes.node this is used a lot) we usually don't have the "subclass of" prepended as it is implicitly understood.

src/psyclone/psyir/nodes/node.py
sergisiso: I would still prefer the test in case somebody changed this implementation the issue should be evident, and at the moment because replace_with is still not used across the code it wouldn't fail anywhere else.

Just replacing the "Schedule<-Statement, Statement, ..." construct in test_replace_with() by a construct where order is important would already do.

examples/psyir/create.py
```diff
@@ -90,52 +90,72 @@ def create_psyir_tree():
 
     # Nodes which do not have Nodes as children and (some) predefined
     # scalar datatypes
-    zero = Literal("0.0", REAL_TYPE)
-    one = Literal("1.0", REAL4_TYPE)
+    # Issue #1136 looks at how to avoid all of the _x versions
```
sergisiso, line 93: The issue reference should have the TODO tag

src/psyclone/psyir/nodes/node.py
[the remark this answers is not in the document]
rupertford: Recently we seem to have moved towards using "subclass of" if we are returning an unknown subclass or a base class, although there is no explicit rule. @arporter, do you have any thoughts on this - in general, not for this particular PR.

sergisiso (changes requested):
@rupertford There is one new suggestion and one response to previous comment that still needs addressing.

Documentation looks good and I also checked formatting and there are no TODOS to the issue. When the remaining comments are addressed this PR should be ready.

src/psyclone/psyir/nodes/node.py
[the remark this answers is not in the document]
arporter: I think Rupert and I in a recent PR agreed to use "subclass of" but, TBH, since a subclass of a class *is* an instance of that class we could drop it for the sake of brevity. I could be persuaded either way although less typing is always good :-)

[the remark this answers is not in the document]
rupertford: I've removed "subclass of" in this case as requested.

src/psyclone/psyir/nodes/node.py
rupertford: OK, I've added an additional test.

examples/psyir/create.py
rupertford: Added.

rupertford:

sergisiso (approved):
@rupertford This PR is now ready to be merged

PR #1141, opened 2021-03-03
#1132 Make depth() return correct values for a pointwise access.

### LINKED ISSUE #1132: GO Stencil inconsistent value
If a GOStencil is set as ``go_pointwise``, you can still query its depth, but you will get ``[000,00,000]``. Either getting the depth of a stencil with ``go_pointwise`` should raise an exception, or (my preferred solution) this should return ``[000,010,000]``. The latter would  imho express the meaning of ``go_pointwise`` adequately (and remove if-tests for some of my dependency code). 

Am happy to fix this as part of #1130, but in case that we don't agree on the best way we have this ticket to discuss this independently. @rupertford , @arporter, @sergisiso

hiker:
Simple PR that fixes #1132.


hiker:
I did not touch anything else - especially I left 'has_stencil' untouched, so has_stencil can still be used to see if a halo exchange is required.

arporter (approved):
All changes are good.
Ref guide still builds fine.
No code conformance problems.

PR #1150, opened 2021-03-09
(Closes #1146) Ensure swap method in symbol table compares normalised symbol names.

### LINKED ISSUE #1146: Swapping of symbols is case sensitive (and shouldn't be).
As reported by Simon:

     $ psyclone -api nemo -l output src/TOP/trc.F90
      Error, unexpected exception, please report to the authors:
      Description ...
      PSyclone SymbolTable error: Error when generating Container for module 'trc': Cannot swap symbols that have different names, got: 'obc_data' and 'OBC_DATA'
      Type ...
      <class 'psyclone.psyir.symbols.symbol.SymbolError'>
      Stacktrace ...
      [Stacktrace omitted]

When I rename 'OBC_DATA' to 'obc_data' in trc.F90, PSyclone seems to process the file well.

arporter:
Small bug fix plus associated test extensions.


arporter:
Simon has kindly confirmed that this works for him.
Ready for review from anyone :-)

rupertford (changes requested):
I'm not sure what this is fixing I'm afraid. I don't think the name needs to be lowered as the PSyIR is case-insensitive. I removed the change to `fparser2.py` and updated the change to `test_name_clash_derived_type` to what it was and that test was then OK. I then found an error in `SymbolTable.swap()` where the name comparisons should be case insensitive (I just used `lower()` with both) - on line 606. When I fixed this the other tests then worked. This either means we are not testing the actual problem you originally observed properly, or that the root of the problem was an error in the `copy` method in the SymbolTable. Could you please check.

arporter:
Ah, my mistake. I assumed that the symbol table must be case sensitive because otherwise getting it to work for other languages is going to be a pain. I never checked the assumption. If it's supposed to be case insensitive then clearly the fix needs to be inside the PSyIR rather than outside so I'll take this back.

arporter:
If the PSyIR is not case sensitive, then should we be converting all symbol names to e.g. lower case? At the minute the Symbol constructor just stores the name it is given, unchanged. What do @sergisiso and @rupertford think?

sergisiso:
@arporter The Symbol Table is case insensitive because it normalises the symbol table keys, but the Symbols stay with the original case. The idea was to provide the Fortran functionality to lookups without necessarily having to output code in a different case than the input code (Of course there is only one Symbol object in the PSyIR so if the case is inconsistent in the input code, all references will appear with the case that created the symbol in the output)

This is the reason for the current behavior, but if you think a different alternative will make more sense we can change it.

It is true that to with other languages it may be easier with case sensitive (although there is alternatives) and the Fortran frontend could lowercase everything, but then we lose the original case in generated Fortran. I don't know if that's important.

rupertford:
I would go with keeping the functionality as it is for the moment i.e. fix the error in SymbolTable.copy() (if that is where the problem comes from) and raise a separate issue that there will be a potential problem if and when someone wants to parse a case sensitive language to the PSyIR.

arporter:
OK, I think this is now ready for another look.

rupertford (approved):
Looks good. All changes conform to the styling rules. No need to update the documentation as this is a bug fix.

PR #1155, opened 2021-03-11
Add support for real and int binary operations (closes #529)

### LINKED ISSUE #529: [nemo] fparser2 frontend does not handle REAL intrinsic with two arguments
Currently, processing code such as `REAL(jk, wp)` results in a `CodeBlock` whereas it should give a `BinaryOperation`.

arporter:


arporter:
In the end the change I've gone for is minimal but it does highlight (again) the lack of type checking in the PSyIR. I'll let the reviewer decide whether this is sufficient for now or whether we need to refactor things somewhat. Should be ready to go once the CI completes.

src/psyclone/tests/psyir/nodes/type_convert_operation_test.py
```diff
@@ -0,0 +1,97 @@
...
+                          (BinaryOperation.Operator.INT, "int")])
+def test_type_convert_binaryop_create(operation, op_str):
+    '''Test that the create method in the BinaryOperation class correctly
+    creates a BinaryOperation instance for the REAL and INT type-conversion
+    operations..
+
+    '''
+    sym = DataSymbol("tmp1", REAL_SINGLE_TYPE)
+    lhs = Reference(sym)
+    wp_sym = DataSymbol("wp", REAL_SINGLE_TYPE)
```
sergisiso, line 60: Shouldn't the "correct" version of the test be with wp_sym a DataSymbol with INTEGER_SINGLE_TYPE and a constant_value?

Maybe you can add something like this (a non-constant datasymbol) in the xfailing test below to be aware of non checked constrain.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -220,6 +223,18 @@ class BinaryOperation(Operation):
        :returns: the value of the upper bound of the `index` dimension of \
                  `array`.
 
+    Casting Operators:
+
+    .. function:: REAL(arg0, precision)
+
+       :returns: `arg0` converted to a floating point number in the \
```
sergisiso, line 230: Would "of" instead of "in" be better? Please ignore if this is just my non-native English and both work well here.

sergisiso (changes requested):
@arporter I think if this works for the CodeBlocks cases you try to resolve it is already fine. I agree PSyIR type-checking would be helpful in this situation but it has to be part of a larger generic type-checking functionality imo.

The PR looks good I just have a couple of minor comments.

src/psyclone/tests/psyir/nodes/type_convert_operation_test.py
arporter: Oops!

arporter: I've extended the x-failing test as you suggest.

src/psyclone/psyir/nodes/operation.py
arporter: Yes, I think you're right. Done (and for INT too).

arporter:
Ready for another look once the CI has completed.

sergisiso (approved):
@arporter This PR is ready to merge

PR #1157, opened 2021-03-12
478 same space option

### LINKED ISSUE #478: Providing Options to Transformations
For my ROMS work I need to disable the test RegionTrans._validate to check for valid node types: if I want to apply an OMP parallel directive to statements that contains assignments (required since I sometimes have assignment between loop that should be in one parallel region), the _validate check prohibits this (since assignments are not allowed):
```
                        psyGen.Literal, psyGen.Reference)
```
While I could easily add assignment to this list, I think it would be good to have a more flexible system of handling this.

Obviously we could provide a parameter to OMPParallelTrans constructor or apply() function (or a separate setter for valid node types) where we can modify the valid node types.

But I am wondering if it might be worth adding a more flexible 'transformation option' system for all transformations: each apply() function would take a dictionary, where additional options can be specified. So we could call for example:
```
omptrans.apply({"disable node type check":1}), 
or:
omptrans.apply({"additional_node_types":[psyGen.Assignment, psyGen.If] }), 
```
The advantage is that all apply() function would have the same signature (one dict, instead of: this transform takes the optional flags a, b, or c, and that transforms might take a and d ... you get the idea)  making it easy to call other transformations, since we just pass the option-dictionary from one apply() function to the next. While there is no real immediate benefit of doing it this way, it _might_ make it more easier long term to have much more complex transformations. A (made up) example:

assume we have a transform ''SuperTrans'' that first renames scalar variables (in case of different usage of the same variable, once thread private, then global) and moves them to the beginning, to allow loop fusion. Then we fuse loops, and apply openmp parallel and omp do as much as possible. Obviously ''SuperTrans'' would consist of several simpler transformations.

Then it might be the case that the omp transform can not decide if a loop can be parallelised (as a real example: assignments to a user defined type a%b(i,j)=... are currently not handled in nemo/PSyir, and I need to overwrite the handling of this special case for one loop). Since the user script only calls ''SuperTrans'', we can't immediately set any flags in the omp transformation.But if we would have one generic option dictionary, we could call ''SuperTrans'' with {'omp-parallelise-user-defined-type':1}. This would not have any affect to any other transform, but the omp transform can test for this setting, and if so enforce parallelisation of the loop.

Downside of course - if there should be a typo in a parameter name, it will not be detected (unless we add some complicated code to keep track of valid parameters ...).

It might be an overkill (or have other disadvantages that I am not thinking of atm). And there might be other solutions (provide ''SuperTrans'' with instances of the transformation it needs, so we can create customise the behaviour this way??, ...)

Any comments? And as a more direct question: if you don't like this approach, what is the best way for me to disable the validate check? I like the idea of the check, so I wouldn't suggest to remove it, or just add more in (since basically nearly everything might be allowed in a parallel region, so if we add all, the test becomes useless). Would you prefer a simple solution (just a flag for the omp transform to ignore the test, or a way to optionally add more allowed classes?

hiker:
Fixes #478: removes the last option 'same_space' that was used for LFRicLoopFuse as constructor parameter instead of using it via an option dictionary. This need #1154 to be merged first.


hiker:
Absolutely no idea why codecov failes. Ready for review.

src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
```diff
@@ -68,60 +68,16 @@ class LFRicLoopFuseTrans(LoopFuseTrans):
 
     The optional argument `same_space` can be set as
 
-    >>> ftrans.same_space = True
+    >>> ftrans.apply(schedule[0], schedule[1], {"same-space": True})
```
arporter, line 71: s/same-space/same_space/

src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
```diff
@@ -132,6 +88,11 @@ def validate(self, node1, node2, options=None):
         :type node2: :py:class:`psyclone.dynamo0p3.DynLoop`
         :param options: a dictionary with options for transformations.
         :type options: dictionary of string:values or None
+        :param bool options["same_space"]: this optional flag, set to `True`, \
+            asserts that an unknown iteration space (i.e. `ANY_SPACE`)
```
arporter, line 92: Missing line-continuation symbol.

src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
```diff
@@ -158,9 +119,18 @@ def validate(self, node1, node2, options=None):
...
         # pylint: disable=too-many-locals,too-many-branches
         # Call the parent class validation first
+
+        if not options:
+            options = {}
+        same_space = options.get("same_space", False)
+        if same_space and not isinstance(same_space, bool):
+            raise TransformationError(
+                "Error in {0} transformation: The value of the 'same_space' "
+                "flag must be either Boolean or None type, but the type of "
```
arporter, line 129: I think in Python it is `bool` so please s/Boolean/bool/

arporter (changes requested):
All changes are pylint/pycodestyle clean.
UG has been updated.
All docs build fine.
All tests and examples work with compilation.
Just a few tiny things to fix (I almost did them myself but decided that was a slippery slope).
Issue #478 is no longer mentioned anywhere in code or in the docs.

src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
hiker: Done.

src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
hiker: Done.

src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
hiker: Fixed (including the corresponding test).

hiker:
All done, but I can't help asking one question: why is ``LFRicLoopFuseTrans`` not documented in ``transformations.rst`` - none of the specific ones are ... so I guess it's ok for now (note that I have problems building the reference guide, which I guess would contain that and would have triggered a warning ... might open a ticket for my problem :) ).

Anyway, this PR is ready to go.

rupertford:
> All done, but I can't help asking one question: why is `LFRicLoopFuseTrans` not documented in `transformations.rst` - none of the specific ones are ... so I guess it's ok for now (note that I have problems building the reference guide, which I guess would contain that and would have triggered a warning ... might open a ticket for my problem :) ).
> 
> Anyway, this PR is ready to go.

The user guide is structured so that APIs have their own sections and it was decided that API's-specific transformations should go there, with generic ones being left in the generic section.

hiker:
> The user guide is structured so that APIs have their own sections and it was decided that API's-specific transformations should go there, with generic ones being left in the generic section.

Ah ... I searched for the term, but got only one hit - in dynamo3, and I assumed it would be the note there - but it only lists chapters, independent on number of hits *sigh* Should have looked better.

arporter (approved):
All docs build OK.
I've already checked that 478 isn't mentioned anywhere.
Proceeding to merge...

PR #1162, opened 2021-03-17
(closes #1160) Add check for wildcard import in any outer scope to Fortran backend

### LINKED ISSUE #1160: [psyir] Fortran backend rejects unresolved symbols
After the prompt by Rupert at the telco this morning I tried processing standard NEMO code with the branch I'm working on in #1147. Unfortunately doing this immediately showed me that the PSyIR backend won't allow any unresolved symbols (other than kind parameters). Since NEMO very rarely uses the `ONLY` clause on `USE` statements, pretty much any symbol that is accessed via a `USE` will immediately prevent the backend from generating code. Given that we don't want to be parsing code unnecessarily, I think this constraint should be weakened so that the backend only checks that there is at least one import that could be bringing an unresolved symbol into scope?

arporter:


arporter:
This is ready for a first look from either @rupertford or @sergisiso. It's a small PR that just adds a new method to the SymbolTable to search for wildcard imports and then uses this in `gen_decls()` in the Fortran backend.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,54 @@ def gen_decls(self, symbol_table, args_allowed=True):
             which does not have a GlobalInterface or LocalInterface as this \
             is not supported by this backend.
         :raises VisitorError: if args_allowed is False and one or more \
-                              argument declarations exist in symbol_table.
-        :raises VisitorError: if any symbols representing variables (i.e. \
-            not kind parameters) without an explicit declaration or 'use' \
-            are encountered.
+            argument declarations exist in symbol_table.
+        :raises VisitorError: if any symbols representing variables without \
+            a (local) explicit declaration or 'use' are encountered and there \
```
sergisiso, line 606: I think the "(local)" should be removed as get_unresolved_datasymbols checks at any level.

Also, can the sentence be simplyfied? Maybe "if any symbols without an explicit declaration is found and there are no wildcard imports."

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,54 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
+                           if isinstance(symbol, RoutineSymbol)]
+        for sym in routine_symbols:
+            if (not isinstance(sym.interface, (LocalInterface,
+                                               GlobalInterface)) and
+                    sym.name.upper() not in FORTRAN_INTRINSICS):
+                if not wildcard_imports_checked:
+                    first_wildcard_import = \
+                        symbol_table.get_first_wildcard_import()
+                    wildcard_imports_checked = True
+                if not first_wildcard_import:
```
sergisiso, line 626: Unless it returns a boolean as suggested it should be "if first_wilcard_import is not None"

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,54 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
-            raise VisitorError(
-                "The following symbols are not explicitly declared or imported"
-                " from a module (in the local scope) and are not KIND "
-                "parameters: {0}".format(symbols_txt))
+            # We do have unresolved symbols. Is there at least one wildcard
+            # import which could be bringing them into scope?
+            if not wildcard_imports_checked:
+                first_wildcard_import = \
+                    symbol_table.get_first_wildcard_import()
+            if not first_wildcard_import:
```
sergisiso, line 644: Unless it returns a boolean as suggested it should be "if first_wilcard_import is not None"

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,54 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
-                ["'" + sym + "'" for sym in unresolved_datasymbols])
-            raise VisitorError(
-                "The following symbols are not explicitly declared or imported"
-                " from a module (in the local scope) and are not KIND "
-                "parameters: {0}".format(symbols_txt))
+            # We do have unresolved symbols. Is there at least one wildcard
+            # import which could be bringing them into scope?
+            if not wildcard_imports_checked:
+                first_wildcard_import = \
+                    symbol_table.get_first_wildcard_import()
```
sergisiso, line 643: It may be worth adding a `wildcard_imports_checked = True` here, even if now it's the last time we check it, it will be consistent if this method gets extended in the future.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,54 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
+            if not first_wildcard_import:
+                symbols_txt = ", ".join(
+                    ["'" + sym + "'" for sym in unresolved_datasymbols])
+                raise VisitorError(
+                    "The following symbols are not explicitly declared or "
+                    "imported from a module (in the local scope) and there "
```
sergisiso, line 649: Maybe I am missing something but why the "(in the local scope)"? Isn't get_first/has_wildcard_import checks for wildcard imports at any level?

sergisiso (changes requested):
@arporter I made a few in-line suggestions. Also, it seems that now there is no special case needed for the precision symbols (they can be undeclared if there is at least 1 wildcard import). Can we remove the ignore_precision bit of logic from the get_unresolved_datasymbols SymbolTable method?

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1011,6 +1011,23 @@ def rename_symbol(self, symbol, name):
         # Re-insert modified symbol
         self.add(symbol)
 
+    def get_first_wildcard_import(self):
```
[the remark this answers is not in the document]
arporter: I was thinking about this after I put it in for review. In the longer term we may want the full list of containers with wildcard imports if we're trying to chase down the definition of a particular symbol. However, probably I should simplify the implementation so that it just does what is currently required.

src/psyclone/psyir/backend/fortran.py
arporter: I was getting a bit confused. I've sorted myself out and reworded :-)

src/psyclone/psyir/backend/fortran.py
arporter: It's a boolean now.

src/psyclone/psyir/backend/fortran.py
arporter: Done.

src/psyclone/psyir/backend/fortran.py
arporter: Good idea, done.

src/psyclone/psyir/backend/fortran.py
arporter: No, it was me getting a bit confused. Fixed now.

arporter:
I've had a look at the `ignore_precision` argument to `get_unresolved_datasymbols` and it's still used in the OpenCL backend:
https://github.com/stfc/PSyclone/blob/9eab45c7d4fbbf53373885b69dfa936fd0d49184/src/psyclone/psyir/backend/opencl.py#L202-L207
 so I have left it alone.

Assuming the CI is happy then this is ready for another look.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,53 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
-                "supported by the Fortran back-end.")
+        # Keep a record of whether we've already checked for any wildcard
+        # imports to save doing so repeatedly
+        wildcard_imports_checked = False
+        has_wildcard_import = False
+
+        routine_symbols = [symbol for symbol in symbol_table.symbols
+                           if isinstance(symbol, RoutineSymbol)]
+        for sym in routine_symbols:
+            if (not isinstance(sym.interface, (LocalInterface,
```
sergisiso, line 619: The test that checks this uses:
```
# Now add a user-defined routine symbol
symbol_table.add(RoutineSymbol("arg_sub", interface=ArgumentInterface()))
```
This surprised me as Fortran would need to declare an interface to pass a routine as argument, and thebackend doesn't do anything like this and will probably produce incorrect code. Also this isinstance could just check for UnresolvedInterface I think. 

So would it make sense to update to the following code?
```
for sym in routine_symbols:
    if (isinstance(sym.interface, UnresolvedInterface) and sym.name.upper() not in FORTRAN_INTRINSICS):
        # Current loop body
    else if isinstance(sym.interface, ArgumentInterface):
        # Fails, not supported
```

src/psyclone/psyir/backend/fortran.py
```diff
@@ -601,34 +601,53 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
+                                               GlobalInterface)) and
+                    sym.name.upper() not in FORTRAN_INTRINSICS):
+                if not wildcard_imports_checked:
+                    has_wildcard_import = symbol_table.has_wildcard_imports()
+                    wildcard_imports_checked = True
+                if not has_wildcard_import:
+                    raise VisitorError(
+                        "Routine symbol '{0}' does not have a global or local"
```
sergisiso, line 627: I think it would be a bit clearer to use the same full name and capitalisation as the classes it refers in this error message: RoutineSymbol, GlobalInterface and LocalInterface

sergisiso (changes requested):
@arporter Almost there, I just found one of the test examples a bit unexpected and I think we can add and extra check to make the backend safer.

I checked TODOs, tests, pycodestyle and pylint and there are no new issues in this PR. Once the last comments are addressed this will be ready.

src/psyclone/psyir/backend/fortran.py
arporter: Oh yes, that's a good point. I've done as you suggest and extended the test appropriately.

src/psyclone/psyir/backend/fortran.py
arporter: Yes, done.

arporter:
Assuming the CI is happy, this is ready for another look.

sergisiso (approved):
@arporter All looks good now. This PR is ready to be merged.

Issue #658, opened 2020-01-16
[PSyIR] determine the type of nodes


rupertford:
At the moment it is difficult to determine the type of nodes - unless they have an explicit type associated with them i.e. Literal or Reference. This would be very useful for checking correctness when creating or modifying PSyIR and possibly for backends as well. After discussion with @arporter and @sergisiso it is suggested that all nodes have a get_type() method (or similar) which will work out what type the node is, or returns.

arporter:
This is needed when attempting to determine whether an array access is to a range or not. e.g.

    my_array(my_func(a(:))) = 1.0

could be an assignment to a single element of `my_array` or it could be an assignment to a slice. To know this we need to know what `my_func` returns. We have the same problem with Operators but that is temporarily solved by keeping a list of those Operators that perform a reduction (currently just the unary, binary and nary forms of SUM).

arporter:
All nodes (apart from Calls) now have a `datatype` method so this issue can be closed.

PR #1165, opened 2021-03-18
(closes #1164) generalise assignment.is_array_range and add tests

### LINKED ISSUE #1164: [psyir] Assignment.is_array_range does not work for Structures
Currently, `Assignment.is_array_range` only checks that the lhs is an `ArrayReference`. Now that we can have e.g.

    some_struc(ji)%data(:,:) = 0.0

this is clearly insufficient and results in missed parallelisation opportunities when processing NEMO, as reported by @deardenchris .

arporter:


arporter:
Another small PR that fixes a small bug in the code used to determine whether we have an assignment to an array range.

src/psyclone/psyir/nodes/assignment.py
```diff
@@ -188,16 +190,16 @@ def reference_accesses(self, var_accesses):
...
+        returns: True if the lhs of the assignment is an array access with at \
+            least one of its dimensions being a range and False otherwise.
         rtype: bool
 
         '''
-        from psyclone.psyir.nodes import ArrayReference, Range
-        if not isinstance(self.lhs, ArrayReference):
-            return False
-        return any(dim for dim in self.lhs.children if isinstance(dim, Range))
+        if isinstance(self.lhs, (ArrayReference, StructureReference)):
```
rupertford, line 198: I expect I am worrying unduly but this test is a little looser than the previous one, in that it considers that any `Range` node within an `ArrayReference` or `StructureReference` implies that this access is an array range. The particular example I am thinking about is an array slice within an expression. For example `a(min(b(:))) =`

src/psyclone/tests/psyir/nodes/assignment_test.py
```diff
@@ -158,3 +160,46 @@ def test_is_array_range():
     array_ref = ArrayReference.create(symbol, [my_range, int_one.copy()])
     assignment = Assignment.create(array_ref, one.copy())
     assert assignment.is_array_range
+
+    # Check when lhs consists of various forms of structure access
+    grid_type = StructureType.create([
```
rupertford, line 165: Could you please add an example that is a structure, but does not have an array range.

src/psyclone/tests/psyir/nodes/assignment_test.py
```diff
@@ -158,3 +160,46 @@ def test_is_array_range():
...
+        BinaryOperation.Operator.LBOUND,
+        StructureReference.create(field_symbol, ["data"]), int_one.copy())
+    ubound = BinaryOperation.create(
+        BinaryOperation.Operator.UBOUND,
+        StructureReference.create(field_symbol, ["data"]), int_one.copy())
+    my_range = Range.create(lbound, ubound)
+
+    data_ref = StructureReference.create(field_symbol, [("data", [my_range])])
+    assign = Assignment.create(data_ref, one.copy())
+    assert assign.is_array_range is True
```
rupertford, line 188: Is the use of `is True` to check that a boolean is returned? If so, that's fine but it is different to a pre-existing assert earlier in the test.

rupertford (changes requested):
Good self-contained PR. I have a concern about the modified test in its current form and have requested a small update to the tests.

src/psyclone/psyir/nodes/assignment.py
arporter: No, that's a fair point. I had a quick think about it (when I did it) and couldn't think of an example where it would break but you've found one! I'll improve it.

src/psyclone/tests/psyir/nodes/assignment_test.py
arporter: Yes, that's right. I've updated the other ones now.

src/psyclone/psyir/nodes/assignment.py
arporter: In this situation (and maybe others) I think it would be helpful if we had the concept of operations that perform reductions, or alternatively, whether they return scalars or arrays. Maybe that comes back to the issue of type-checking PSyIR expressions?

arporter: For completeness, your example should actually use `minval` which we don't yet support in the PSyIR. However, we do support `SUM` so that's what I've used in the test.

src/psyclone/tests/psyir/nodes/assignment_test.py
arporter: Done.

arporter:
CI permitting, this should be ready for another look now.

src/psyclone/tests/psyir/nodes/assignment_test.py
rupertford: Thanks, there was one more but I've taken the liberty of updating that

src/psyclone/tests/psyir/nodes/assignment_test.py
rupertford: Great.

src/psyclone/psyir/nodes/assignment.py
rupertford: Thanks.

src/psyclone/psyir/nodes/assignment.py
```diff
@@ -188,16 +191,31 @@ def reference_accesses(self, var_accesses):
...
+        # reduction and thus returns a scalar result, e.g. a(SUM(b(:))) = 1.0
+        # TODO #658 this check for reductions needs extending to also support
+        # user-implemented functions.
+        if isinstance(self.lhs, (ArrayReference, StructureReference)):
+            ranges = self.lhs.walk(Range)
+            for array_range in ranges:
+                opn = array_range.ancestor(Operation)
+                while opn:
+                    if opn.operator in REDUCTION_OPERATORS:
+                        # The current array range is in an argument to a
```
rupertford, line 210: I think this is an over-zealous check as a reduction operation may still return an array. For example, for the sum operation, if the second argument is a dimension value and the array has more than one dimension. If this is not an issue for nemo then I'm happy to leave it as it is as we can argue that it is an issue for #658, but I thought it best to raise it. This highlights two things to me. 1) the problem of just accepting Fortran intrinsics with all their options in the PSyIR; I think it would be better for us to distinguish between the different types of `sum` with different names in the PSyIR to make life simpler and 2) the issue of what to do if we don't know this information at code generation time (as I don't think we can always guarantee that we do).

src/psyclone/tests/psyir/nodes/assignment_test.py
```diff
@@ -133,28 +135,133 @@ def test_assignment_children_validation():
...
     one = Literal("1.0", REAL_TYPE)
     int_one = Literal("1", INTEGER_TYPE)
+
+    # lhs is an array reference with a range
+    array_type = ArrayType(REAL_TYPE, [10, 10])
+    symbol = DataSymbol("x", array_type)
+    my_range = Range.create(int_one, int_one.copy(), int_one.copy())
+    array_ref = ArrayReference.create(symbol, [my_range, int_one.copy()])
+    assignment = Assignment.create(array_ref, one.copy())
+    assert assignment.is_array_range
```
rupertford, line 150: Is this purposefully different to the rest i.e. no `is True`?

rupertford (changes requested):
There is a potential issue with the logic that assumes a reduction returns a scalar which I think is worth checking as NEMO might make use of this.

src/psyclone/psyir/nodes/assignment.py
arporter: Ooh, that's difficult. I've see-sawed back and forth and come down in favour of delaying this to #658. I've added an x-failing test to illustrate the problem. I've also checked with the NEMO code base and this situation never occurs (for the particular configuration I'm building).

src/psyclone/tests/psyir/nodes/assignment_test.py
arporter: Oops.

arporter:
Once the CI has done its thang, this is ready for another look.

src/psyclone/psyir/nodes/assignment.py
rupertford: I'm happy with that, thanks for adding the test.

rupertford (approved):
All issues have been addressed. Changes are conformant. No need to update docs or examples.

PR #1166, opened 2021-03-18
(Closes #1152) add Unary SUM operation

### LINKED ISSUE #1152: [nemo] fparser2 frontend does not handle SUM intrinsic with a single argument
For example, the following code (from `sbcisf.f90`)

     IF ((SUM(e3t_n(ji, jj, ikt : jk - 1)) < rhisf_tbl(ji, jj)) .AND. (tmask(ji, jj, jk) == 1)) ikb = jk

currently gets put within a `CodeBlock` whereas it should give a `UnaryOperation`.

arporter:


arporter:
A really small PR that just adds Unary SUM() support. Ready for review.

rupertford (approved):
All looks good. There is a constraint here that the argument should be an array, but we don't capture that constraint. However, that is a general issue for sum and is not particular to this PR so is not relevant here.

PR #1176, opened 2021-03-24
(closes #1174) add check that schedule is not empty plus associated test.

### LINKED ISSUE #1174: Profiler.add_profile_nodes attempts to add profiling around empty Schedule
When using the 'profile invokes' option, we need to check that a `Schedule` does have at least one child before attempting to enclose it in a profiling region (line 164):

https://github.com/stfc/PSyclone/blob/6ab2fc0fee5a5d38560ab89310339a8d3c8b685e/src/psyclone/profiler.py#L150-L164

arporter:
Small bug fix to the Profiler class that just avoids an error when we attempt to add profiling to an empty routine (invoke).


arporter:
Ready for review, probably from @hiker :-)

arporter:
Just a gentle nudge for @hiker to take a look at this when he has time :-) (It's not big.)

hiker (approved):
All new lines covered, tests works, no changes to documentation.

PR #1177, opened 2021-03-25
(closes 1168) add `lower_to_language_level()` method to `ProfileNode`

arporter:
This is part of the push towards being able to generate code for the NEMO API from the PSyIR Fortran backend. In order to maintain our existing profiling functionality, we have to lower any ProfileNodes to language-level PSyIR nodes so that Fortran can be generated.


arporter:
This is ready for a first look now from either @sergisiso or @rupertford. It primarily adds just one new `lower_to_language_level()` method to the `ProfileNode` class. However, that required a slight tweak to `CodeBlock` (supporting an annotation) and modifications to the `PSyDataNode` base class. I realised we were missing a `profile_node_test` file so I've created one and moved a few of the tests out of the transformation section into it, as well as new tests for the lower... method.
Note that I don't test the code generated from the 'lowered' PSyIR as that's still a bit of a work in progress (and was what spawned this issue in the first place). Now that I've written that, I realise that I probably could add such a test for the GOcean API as I think Sergi has the backend working there? I'll let the reviewer comment...

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -133,12 +135,38 @@ def __init__(self, ast=None, children=None, parent=None, options=None):
             # FIXME: This may not be a good solution
             symtab = SymbolTable()
 
+        # Ensure that we have a container symbol for the API access
```
sergisiso, line 138: Generally the node constructors do not modify other parts of the tree or the symbol tables. But I see this psy_data constructor (not just what you added but what already was here) tries to do lots of things (insert children, parts of gen_code, ...).

I am not too familiar with how this is used but from a high level view I would expect the transformation that inserts the PSyDataNode to perform these manipulations and the constructor, if anything, just check if given arguments exist/are valid. Is there any reason all these need to be inside the constructor in this case?

That said, if moving these steps to the transformation is not a trivial modification, I am fine to overlooking this for the purpose of this PR.

src/psyclone/tests/psyir/nodes/profile_node_test.py
```diff
@@ -0,0 +1,216 @@
...
+    assert isinstance(ptree[0], Fortran2003.Call_Stmt)
+    assert kschedule[1] is assign1
+    assert kschedule[2] is assign2
+    assert isinstance(kschedule[-2], CodeBlock)
+    assert kschedule[-2].annotations == []
+    ptree = kschedule[-2].get_ast_nodes
+    assert len(ptree) == 1
+    assert isinstance(ptree[0], Fortran2003.Call_Stmt)
+    assert isinstance(kschedule[-1], Return)
+    # Check that the symbol table contains the appropriate symbols:
```
sergisiso, line 146: Are the lines from here to the end of the test something that the constructor is currently in charge of . Would it make more sense to check this is the constructor test*?

*or somewhere else, see previous comment

sergisiso (changes requested):
Looks good, I have some general questions about the PSyData class but the lower_to_language_level looks fine.

>Now that I've written that, I realise that I probably could add such a test for the GOcean API as I think Sergi has the backend working there? I'll let the reviewer comment...

Not fully, I just have it working for gocean/eg6 which is just a single kernel and it doesn't include PSyData nodes. I have a branch which extends support for most of gocean, I think the gocean tests can wait until that one is pushed.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -56,12 +56,20 @@ class CodeBlock(Statement, DataNode):
...
     :param parent: the parent node of this code block in the PSyIR.
     :type parent: :py:class:`psyclone.psyir.nodes.Node`
+    :param annotations: tags that provide additional information about \
+        the node. The node should still be functionally correct when \
+        ignoring these tags.
+    :type annotations: list of str or NoneType
 
     '''
-    # Textual description of the node.
+    #: Textual description of the node.
```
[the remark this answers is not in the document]
arporter: Yes - it lets Sphinx (or maybe Doxygen?) treat the comments as documentation for the following property.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -75,8 +83,8 @@ class Structure(Enum):
         # The Code Block comprises one or more Fortran expressions.
         EXPRESSION = 2
 
-    def __init__(self, fp2_nodes, structure, parent=None):
-        super(CodeBlock, self).__init__(parent=parent)
+    def __init__(self, fp2_nodes, structure, parent=None, annotations=None):
```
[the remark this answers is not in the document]
arporter: That would indeed be nicer. Might be quite a lot of work though?

src/psyclone/psyir/nodes/profile_node.py
```diff
@@ -136,3 +140,70 @@ def gen_c_code(self, indent=0):
         '''
         raise NotImplementedError("Generation of C code is not supported "
                                   "for profiling")
+
+    # -------------------------------------------------------------------------
+    def lower_to_language_level(self):
+        '''
+        Lowers this node (and all children) to language-level PSyIR. The
+        PSyIR tree is modified in-place. This ProfileNode is replaced by
+        a pair of CodeBlocks (representing the calls to the start and stop
```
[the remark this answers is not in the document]
arporter: Good idea - done.

src/psyclone/tests/psyir/nodes/profile_node_test.py
```diff
@@ -0,0 +1,216 @@
...
+    pnode = ProfileNode()
+    assert pnode._class_string == "profile_"
+    assert str(pnode) == ("ProfileStart[var=profile_psy_data]\n"
+                          "ProfileEnd")
+    pnode2 = ProfileNode(options={"region_name": ("my_mod", "first")})
+    assert pnode2._module_name == "my_mod"
+    assert pnode2._region_name == "first"
+
+
+def test_c_code_creation():
```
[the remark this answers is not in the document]
arporter: Done. I didn't spot that!

src/psyclone/tests/psyir/nodes/profile_node_test.py
```diff
@@ -0,0 +1,216 @@
...
+    name is provided to a ProfileNode.
+
+    '''
+    with pytest.raises(InternalError) as excinfo:
+        _ = ProfileNode(options={"region_name": value})
+    assert ("Error in PSyDataNode. The name must be a tuple containing "
+            "two non-empty strings." in str(excinfo.value))
+
+
+def test_lower_to_lang_level_single_node(parser):
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/nodes/psy_data_node.py
arporter: I think most of what I've added was necessary in order to update the way the name of the symbol associated with the profiling region is obtained. However, whether that really needs to be done in the constructor is debatable. As the comments say, it means it can be displayed in the `str()` but, since this is a variable in generated code, I'm not sure we need this. In an ideal world, would you say that most of this symbol creation would be perfomed in the `lower_to_language_leve()` method? I can certainly see that the creation of the Schedule and association of the child nodes should be done in the Transformation. In fact, we probably need a `create()` method to perform this, as we do elsewhere.
As four different classes inherit from `PSyDataNode`, I think most of these modifications should be done in a separate issue. If @hiker agrees, I'm happy to remove the setting of the variable name from the constructor (provided that it doesn't then result in lots of other changes to this PR).

arporter:
Marking this as ready for review as I need some feedback from @sergisiso and @hiker before deciding what to do (and what not to do) in this PR.

src/psyclone/psyir/nodes/psy_data_node.py
hiker: That's rather a difficult question, since I think a lot of code here comes from various sources (e.g. region name is from @rupertford , I get only blamed because I moved the code from ProfileNode to PsyData node when using psydata for profile).

1) Why there is so much done ... as far as I can see: I think mostly that's a different philosophy - I don't like to repeat code that inserts children in the transformations, and I would prefer to have the nodes take care of that. Isn't @sergisiso adding part of the consistency checking in a recent (or current) PR? This way it is guaranteed that the node sits at the right place. While I don't mind that much (I think we agreed to disagree :) ) - I want to point out the fact that we already ran into problems because some constructors do NOT fully initialise a node: gocean loop nodes don't have bounds properly defined, meaning we can't do any bounds checking in the validation function of a transformation. IIRC it was pointed out that this was difficult to fix now (and I checked it certainly requires more than just calling the functions that define the bounds, there are other things missing preventing this from working). So, by moving parts of the initialisation out of the node, we risk (am not saying that we indeed are) that we create the same situation again.

2) I think the handling of the profiling name is originally from @rupertford, git is just blaming me because I moved Rupert's code from Profile into PsyData when I made the ProfileNode based on PSyData. But Rupert might have a better idea of why the name is set there. 

In general I prefer to have nodes to be 'ready' to be used anywhere once they are constructed.

sergisiso: No worries for now, I just mentioned it partly to understand the reason it was done this way, and partly to see if the change was easier for somebody more familiar with this code, but I agree this is not part of the PR and can be discussed separately.

@hiker :

>  Isn't @sergisiso adding part of the consistency checking in a recent (or current) PR? 

The parent link consistency PR indeed removes some part of this code and from lots of transformations :)

> In general I prefer to have nodes to be 'ready' to be used anywhere once they are constructed.

> we already ran into problems because some constructors do NOT fully initialise a node

I would in general agree with you in these points, but we needed a constructor that initiases incomplete Node (w/o its connections) for the way the parsing and some PSyIR construction is done which is essentially a top-down approach. To have a more safe constructor we added the Node.create() static method with includes the children and symbol checking.

It's true that the PSyData family of nodes won't appear when the parsing from source and maybe we won't use the thin constructor of this nodes at all, but I think its still worth having the same expectations of what they do like in the other nodes.

arporter: OK, I think we all have the same goals in mind, perhaps just different routes to get there :-) I've opened #1185 for this.

src/psyclone/tests/psyir/nodes/profile_node_test.py
arporter: I've moved them into the constructor test and added a TODO.

sergisiso (approved):
@arporter All requested changes done or created appropriate issues/TODOs.

PR #1182, opened 2021-03-26
(closes #1181) OpenACC support in PSyIR Fortran backend

### LINKED ISSUE #1181: [psyir] Add Fortran backend support for OpenACC directives
Currently the Fortran backend only supports OpenMP directives. In this issue this will be extended to OpenACC.

arporter:


arporter:
This is ready for first review now. It adds OpenACC support to the PSyIR Fortran backend. In doing this I realised that we had different behaviour for `acc parallel` depending upon the API - in some we always put 'default(present)' and in others we didn't. I've changed this so that we do now always put 'default(present)'.
I also realised that we don't do any validation of the PSyIR before generating code so have created #1184.
Really one for @sergisiso or @rupertford although if @hiker wouldn't mind checking my use of the dependence analysis (`begin_string` at line 4566 in psyGen.py) that would be great.

src/psyclone/psyGen.py
```diff
@@ -4486,6 +4563,78 @@ def update(self):
...
+
+        # Identify the inputs and outputs to the region (variables that
+        # are read and written).
+        from psyclone.core.access_info import VariablesAccessInfo
+        var_accesses = VariablesAccessInfo(self)
+        readers = set()
+        writers = set()
+        for var in var_accesses.all_vars:
+            accesses = var_accesses[var]
+            if not accesses.all_accesses[0].indices:
```
hiker, line 4601: I don't know enough about the overall code structure, but in general I try to deprecate the ``indices`` method, and instead use the symbol-table, e.g. (from dependency_tools):
```
            symbol_table = loop.scope.symbol_table
            if var_name not in symbol_table:
                # TODO #845: Once we have symbol tables, any variable should
                # be in a symbol table, so we have to raise an exception here.
                # We need to fall-back to the old-style test, since we do not
                # have information in a symbol table. So check if the access
                # information stored an index:
                is_array = var_info[0].indices is not None
            else:
                # Find the symbol for this variable
                symbol = symbol_table.lookup(var_name)
                is_array = symbol.is_array
```
This should much better handle the case if there is an array expression without ``:`` (eg. not ``a(:) = b(:) * 2``, but just ``a=b*2`` - the latter case would return no indices since at that time iirc we had no array representation that I could use).

sergisiso:
@rupertford I see you added the Under Review label on this PR but I am not sure if you started reviewing it. If you haven't, I wouldn't mind reviewing it myself as I am waiting for this (and something equivalent for OpenMP directives) to switch Gocean to use the PSyIR backends.

rupertford:
@sergisiso - please do, that would be great. You are right, I've not started reviewing it yet.

src/psyclone/psyGen.py
```diff
@@ -1708,6 +1744,30 @@ def update(self):
                 text += " COLLAPSE({0})".format(self._collapse)
         self._add_region(start_text=text)
 
+    def begin_string(self):
```
sergisiso, line 1747: Could the gen_code() and update() methods use this string instead of duplicating the logic that generates the string?
I see the DirectiveGen expect separate parameters but maybe some unwrapped list like: `*self.begin_string().split()` would work? If it turns out not to be that easy that won't mind the duplication.

Can you also consider this in the other Directives that this PR created begin_ and end_ strings for?

sergisiso (changes requested):
@arporter The PR looks good, I made just a few comments on the implementation.

(Offtopic: I see OpenMP directives are already supported by the backends, I don't know why I saw some failing when using them for Gocean, but maybe is just some small syntax change)

src/psyclone/psyGen.py
arporter: Oh yes, good call. I've done that now.

src/psyclone/psyGen.py
```diff
@@ -1486,35 +1488,48 @@ def dag_name(self):
...
         return "ACC_parallel_" + str(self.abs_position)
 
-    def gen_code(self, parent):
+    def _pre_gen_validate(self):
         '''
-        Generate the elements of the f2pygen AST for this Node in the Schedule.
+        Check that the PSyIR tree containing this node is valid. Since we
+        use 'default(present)', this node must either be the child of an
+        ACCDataDirective or the parent Schedule must contain an
+        ACCEnterDataDirective.
```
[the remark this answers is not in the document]
arporter: Yes, done.

src/psyclone/psyGen.py
```diff
@@ -1486,35 +1488,48 @@ def dag_name(self):
...
 
-        :param parent: node in the f2pygen AST to which to add node(s).
-        :type parent: :py:class:`psyclone.f2pygen.BaseGen`
+        :raises GenerationError: if this ACCParallel node is not preceeded by \
```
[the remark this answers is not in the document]
arporter: Oops, at least I was consistent...

src/psyclone/psyGen.py
```diff
@@ -1528,6 +1543,26 @@ def gen_code(self, parent):
 
         parent.add(DirectiveGen(parent, "acc", "end", "parallel", ""))
 
+    def begin_string(self):
+        '''
+        Returns the beginning statement of this directive, i.e.
+        "acc begin parallel" plus any qualifiers. The visitor is responsible
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyGen.py
```diff
@@ -4426,6 +4486,23 @@ def ref_list(self):
                     variables.append(arg)
         return variables
 
+    def begin_string(self):
+        '''Returns the beginning statement of this directive, i.e.
+        "acc kernels ...". The visitor is responsible for adding the
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyGen.py
```diff
@@ -4426,6 +4486,23 @@ def ref_list(self):
...
+        :returns: the beginning statement for this directive.
+        :rtype: str
+
+        '''
+        result = "acc kernels"
+        if self._default_present:
+            result += " default(present)"
+        return result
+
+    def end_string(self):
```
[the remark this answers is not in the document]
arporter: Oops.

src/psyclone/psyGen.py
```diff
@@ -4486,6 +4563,78 @@ def update(self):
         self._add_region(start_text="DATA", end_text="END DATA",
                          data_movement="analyse")
 
+    def begin_string(self):
+        '''Returns the beginning statement of this directive, i.e.
+        "acc data". The visitor is responsible for adding the
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyGen.py
```diff
@@ -4486,6 +4563,78 @@ def update(self):
...
+            from psyclone.psyir.backend.fortran import FortranWriter
+            fwriter = FortranWriter()
+            ref_list = [fwriter(ref) for ref in struct_accesses]
+            raise NotImplementedError(
+                "Structure (derived-type) references are not yet supported "
+                "within OpenACC data regions but found: {0}".format(ref_list))
+
+        # Identify the inputs and outputs to the region (variables that
+        # are read and written).
+        from psyclone.core.access_info import VariablesAccessInfo
```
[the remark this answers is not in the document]
arporter: Yes, don't know why I'd put that there. Gone now.

src/psyclone/tests/nemo/transformations/openacc/parallel_directive_test.py
```diff
@@ -152,9 +165,11 @@ def test_parallel_repeat_update(parser):
     code = parser(reader)
     psy = PSyFactory(API, distributed_memory=False).create(code)
     schedule = psy.invokes.invoke_list[0].schedule
+    data_trans = TransInfo().get_trans_name('ACCDataTrans')
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyGen.py
arporter: Oh yes, I was trying to avoid touching `update()` since it will be removed soon but actually it was easy to tidy both it and `gen_code`.

arporter: I couldn't do this in `ACCParallelDirective` because, as you say, DirectiveGen expects separate parameters and I can't pass it a list without modifying `DirectiveGen` itself. Similarly for `ACCKernelsDirective` and `ACCDataDirective`.

arporter:

src/psyclone/psyGen.py
```diff
@@ -1488,6 +1490,39 @@ def dag_name(self):
...
+        '''
+        # We can't use Node.ancestor() because the enter data directive does
+        # not have children. Instead, we go back up to the Schedule and
+        # walk down from there.
+        routine = self.ancestor(Routine)
+        nodes = routine.walk(ACCEnterDataDirective)
+        # Check that any enter-data directive comes before this parallel
+        # directive
+        if nodes and nodes[0].abs_position > self.abs_position:
+            raise GenerationError(
```
sergisiso, line 1513: Codecov points to this line as not covered, can you check this exception.

src/psyclone/psyGen.py
sergisiso: I was going to resolve this issue because as you mention this will be removed soon and is not a big deal, but since there is still another comment in the review, what I meant here was:
```parent.add(DirectiveGen(parent, *self.begin_string().split()))```
and below:
```parent.add(DirectiveGen(parent, *self.end_string().split(), ""))```

These should works without modifying the DirectiveGen

File: src/psyclone/psyGen.py [hunk 12]
```diff
@@ -1465,7 +1465,9 @@ def data_on_device(self, parent):
 class ACCParallelDirective(ACCDirective):
     '''
     Class representing the !$ACC PARALLEL directive of OpenACC
-    in the PSyIR.
+    in the PSyIR. By default it includes the 'DEFAULT(PRESENT)' clause which
```
sergisiso, line 1468: Several AccDirective classes have the pylint issue  "Method '__str__' is abstract in class 'Node' but is not overridden", but the `node.py:node.__str__` should not be abstract, could you remove the `@abc.abstractmethod` in that method.

src/psyclone/psyGen.py
```diff
@@ -1530,6 +1545,26 @@ def gen_code(self, parent):
 
         parent.add(DirectiveGen(parent, "acc", "end", "parallel", ""))
 
+    def begin_string(self):
```
sergisiso, line 1548: Several of the `begin_string` and `end_string` methods have the pylint issue "no-self-use: Method could be a function"

I think its right for them to be methods as other versions of begin/end_string use self and its better to have a consistent interface. Could could you add the pylint disable statement for these warnings.

src/psyclone/tests/domain/lfric/transformations/dynamo0p3_transformations_test.py
```diff
@@ -6352,7 +6352,7 @@ def test_accenterdata_builtin(tmpdir):
         _, _ = acc_loop_trans.apply(loop)
     _, _ = parallel_trans.apply(sched.children)
     _, _ = acc_enter_trans.apply(sched)
-    output = str(psy.gen)
+    output = str(psy.gen).lower()
```
sergisiso, line 6355: Was there any reason for lowering the output in this test?

sergisiso (changes requested):
@arporter There is a line missing coverage and a couple of pylint warnings that we can avoid, otherwise it looks good.

src/psyclone/psyGen.py
arporter: Oh yes, I live and learn! That's nice.

src/psyclone/psyGen.py
arporter: Oh yes, sorry I missed that. I wonder if it has always been uncovered? Took a bit of work but I've added a couple of new tests.

src/psyclone/psyGen.py
arporter: Yep, done.

src/psyclone/psyGen.py
arporter: Good idea, done.

src/psyclone/tests/domain/lfric/transformations/dynamo0p3_transformations_test.py
arporter: Yes - in increasing the code-reuse the new code uses a different case to the old code (for directives) so I just opted to convert everything to lowercase in the test.

arporter:
OK, I think we're all good now. Ready for another look.

sergisiso (approved):
@arporter All looks good now. Ready to merge.

Issue #1178, opened 2021-03-26
Cannot create Reference Guide


hiker:
When I try to ``make html`` in ``doc/reference_guide``, I get:
```
Generating example index...
finalizing index lists...
writing tag file...
Running plantuml with JAVA...
Running dot...
lookup cache used 10203/65536 hits=44693 misses=10674
finished...

Extension error:
Could not import extension autoapi.sphinx (exception: No module named 'autoapi.sphinx')
make: *** [Makefile:19: html] Error 2
```

I couldn't find much about this when googling, https://github.com/readthedocs/sphinx-autoapi/issues/69 indicates to test the following command:
```
~/work/psyclone/doc/reference_guide$ python3 -mautoapi.extension
/usr/lib/python3.8/runpy.py:127: RuntimeWarning: 'autoapi.extension' found in sys.modules after import of package 'autoapi', but prior to execution of 'autoapi.extension'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
~/work/psyclone/doc/reference_guide$ which python3
/usr/bin/python3

```
So while there is a warning, I don't think that's important.

Any idea what might be wrong? I thought it was because of my two python3 installations (system and OneAPI), but I still have the same issue when I disable oneapi.

arporter:
In a word, "no." I have the following packages installed:
    (py3) psyir$ pip list | grep autoapi
    autoapi                       2.0.1     
    (py3) psyir$ pip list | grep sphinx
    sphinx-rtd-theme              0.5.0     
    sphinxcontrib-applehelp       1.0.2     
    sphinxcontrib-bibtex          2.1.4     
    sphinxcontrib-devhelp         1.0.2     
    sphinxcontrib-htmlhelp        1.0.3     
    sphinxcontrib-jsmath          1.0.1     
    sphinxcontrib-qthelp          1.0.3     
    sphinxcontrib-serializinghtml 1.1.4

arporter:
If you make a clean virtualenv and then do a full pip install of PSyclone (from the source) does it work then?

hiker:
That was extremely useful, I didn't have autoapi installed, only sphinx-autoapi. Not sure if this makes any sense this way, but it's working now. THANKS!!

hiker:
I'll reopen this to document this somewhere.

Issue #693, opened 2020-02-08
proposed unit-testing rule


rupertford:
Each PSyclone Python file has an associated test file.
I think that the tests in the test file should provide 100% coverage for the Python file.
This promotes tests to be unit tests and allows simpler and quicker checking of coverage when developing.
At the moment this is not the case so we would need to enforce a new rule that states that any modified files must also ensure the associated test file provided 100% coverage (and/or someone could just go ahead and do it for some/all files in a PR).

Thoughts ...

hiker:
While I know I am going to hate me for this, I agree that this would indeed be nice (especially since I can't usually run all tests, so being certain that all lines are covered takes a bit trial and error).

I just wonder if we somehow can 'chain' tests: We have tests in different files (for good reasons, e.g. to have all gocean tests in one place). But with PSyData there is a base class psy_data_node, which by itself is not used (it might be abstract or not), only the derived classes will be instantiated. So in order to test all of psy_data_node, I need to run tests from (say) gocean_extract_node. So ideally I would like to run just tests/psyir/nodes/psy_data_node.py, but that should trigger to run (certain/all) tests from (say) domain/gocean/nodes/gocean_node_test.

I admit I am not sure if this is a valid example ... I could obviously just move the gocean tests into the psy_data_node test. But then ... how do I test the gocean implementation?

rupertford:
I'm not sure of the specifics but I would say that we should be able to test the base class in a base class test file by either creating some fake subclass or by choosing an existing subclass example. These tests should be designed to test all the lines in the base class code. If it so happens the test also tests some code in another file then we just accept/ignore this as it is not relevant. This is likely  to mean that we will have similar tests in different files but I think we pay that price.

hiker:
I still can't make my mind up about this :) So as part of #684 I have started to refactor some tests to follow this proposed rules. Some experiences:
1. One transform was a base class with some virtual functions. I had to add a dummy class to implement the abstract functions in order to create an instance (or get a derived class from somewhere, but it feels odd that e.g. in RegionTrans I then use ProfileTrans for testing).
[...]

PR #1183, opened 2021-03-26
1075 validate loop fusion

hiker:
Creates a separate NEMO Loop Fusion transformation that tests:
- for identical loop boundaries
- that accesses are in the right dimension (e.g. not ``a(i,j)``, and ``a(j,i)``)


hiker:
Hooray, tests passed now. Ready for first review. There is one obvious question that I need feedback for:
I have added the tests to the new NemoLoopFuse transformations. In general, these tests could be moved to the base class LoopFuse - BUT then we might have a problem with e.g. gocean (undefined loop boundaries, ...). So for now I left them only in Nemo, and all tests are therefore in Nemo. I am not sure if we should move these tests into a nemo-specific file? While I like the idea of #693, how do we do this here? We likely need some additional tests to get (say) nemo fully covered.Do we import the tests to avoid code duplication? *Sigh* Anyway, let me know if you prefer to have the (ATM only nemo) tests into a specific NEMO file.

hiker:
Oh, I also added some limited handling of scalar variables - it's by far not perfect (conditionals are an issue), but should at least add some useful protection from bad loop fusing..

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+
+
+# ----------------------------------------------------------------------------
+def test_fusetrans_error_not_same_parent():
+    ''' Check that we reject attempts to fuse loops which don't share the
+    same parent '''
+
+    sch1 = Schedule()
+    sch2 = Schedule()
+    loop1 = Loop(variable=DataSymbol("i", INTEGER_TYPE), parent=sch1)
```
arporter, line 101: I appreciate that you've just moved this test over but it could be shortened a lot by using `loop1 = Loop.create(DataSymbol("i", INTEGER_TYPE), Literal("1", INTEGER_TYPE), Literal("10", INTEGER_TYPE), Literal("1", INTEGER_TYPE), [Return()])`

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+    assert ("Error in LoopFuseTrans transformation. Loops do not have the "
+            "same parent" in str(err.value))
+
+
+# ----------------------------------------------------------------------------
+def fuse_loops(parser=None, fortran_code=None):
+    '''Helper function that fuses the first two nodes in the given
+    Fortran code, and returns the fused Fortran code as string.
+
+    :param string fortran_code: the Fortran code to loop fuse.
```
arporter, line 130: s/string/str/

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+
+    :returns: String representation of fused code.
+    :rtype: str
+
+    :raises TransformationError: if the loop fuse transformation cannot
```
arporter, line 135: This routine doesn't raise any exceptions.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+
+
+# ----------------------------------------------------------------------------
+def fuse_loops(parser=None, fortran_code=None):
+    '''Helper function that fuses the first two nodes in the given
+    Fortran code, and returns the fused Fortran code as string.
+
+    :param string fortran_code: the Fortran code to loop fuse.
+
+    :returns: String representation of fused code.
```
arporter, line 132: It appears to return a tuple rather than just a str.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+    loop2 = schedule.children[1]
+    fuse = NemoLoopFuseTrans()
+    fuse.apply(loop1, loop2)
+    writer = FortranWriter()
+    return writer(schedule), schedule
+
+
+# ----------------------------------------------------------------------------
+def test_fuse_ok(parser):
+    '''This tests verifies that the valid loop fuse statements work
```
arporter, line 153: "...that loop fusion can be successfully applied to conformant loops."

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+
+    # Try to fuse loops with different parents
+    with pytest.raises(TransformationError) as err:
+        fuse.validate(loop1, loop2)
+    assert ("Error in LoopFuseTrans transformation. Loops do not have the "
+            "same parent" in str(err.value))
+
+
+# ----------------------------------------------------------------------------
+def fuse_loops(parser=None, fortran_code=None):
```
arporter, line 126: The `parser` argument is not documented and probably neither of these arguments should be optional?

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+                 do ji=1, 10
+                    s(ji, jj)=t(ji, jj)+1
+                 enddo
+              enddo
+              end subroutine sub'''
+    with pytest.raises(TransformationError) as err:
+        fuse_loops(parser, code)
+    assert "Step size in loops must be identical, but are" in str(err.value)
+
+    # Test step size - make sure it defaults to 1
```
arporter, line 290: I'm not quite sure what this last bit is doing since the step size is explicitly set to '1' for both loops?

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+                 enddo
+              enddo
+              end subroutine sub'''
+    fuse_loops(parser, code)
+
+
+# ----------------------------------------------------------------------------
+def test_fuse_different_loop_vars(parser):
+    '''
+    Test that loop variables are identicla
```
arporter, line 311: ??

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+def test_fuse_correct_bounds(parser):
+    '''
+    Test that loop boundaries must be identical.
+    '''
+    # TODO: This test needs evaluation
+    # of constant expressions in PSyclone
+    code = '''subroutine sub()
+              integer :: ji, jj, n
+              integer, dimension(10,10) :: s, t
+              do jj=2-1, n
```
arporter, line 344: The lower bounds of these two loops differ so the fusion should always fail, irrespective of evaluation of constant expressions. Perhaps the lower bounds should be the same?

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+                 enddo
+              enddo
+              do jj=1, n+1
+                 do ji=1, 10
+                    s(ji, jj)=s(ji, jj) + t(jj, jj) + t(ji, ji)
+                 enddo
+              enddo
+              end subroutine sub'''
+
+    fuse_loops(parser, code)
```
arporter, line 383: Please add a quick check that the resulting code is OK.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+
+# ----------------------------------------------------------------------------
+def test_fuse_independent_array(parser):
+    '''Test that using arrays which are not dependent on the loop variable
+    are handled correctly. Example:
+    do j  ... a(1) = b(j) * c(j)
+    do j ...  d(j) = a(1)
+    '''
+
+    # The first example can be merged, since 't' is read-only,
```
arporter, line 439: It's 's' that's independent of loop variable, oh, and it's written to in the first loop :-)

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -0,0 +1,585 @@
...
+
+    with pytest.raises(TransformationError) as err:
+        fuse_loops(parser, code)
+    assert "Variable 's' does not depend on loop variable 'jj', but is " \
+           "read and written" in str(err.value)
+
+
+# ----------------------------------------------------------------------------
+def test_fuse_scalars(parser):
+    '''Test that using arrays which are not dependent on the loop variable
```
arporter, line 464: Please correct this docstring.

arporter (changes requested):
It seems loop fusion is harder than I thought! Generally all looks good and I'm happy with where you've put the tests - although they use the NEMO API I think it is generic PSyIR functionality.
I think you might need to allow for variables declared in an outer scope (see comments) but other than that it's just minor things.
I'll fully check tests etc. next time. You can probably add compilation testing for some of the new tests.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Indeed - fixed.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+    def validate(self, node1, node2, options=None):
+        ''' Perform NEMO API specific validation checks before applying
+        the transformation.
+
+        :param node1: the first Node that is being checked.
+        :type node1: :py:class:`psyclone.psyir.nodes.Node`
+        :param node2: the second Node that is being checked.
+        :type node2: :py:class:`psyclone.psyir.nodes.Node`
+        :param options: a dictionary with options for transformations.
+        :type options: dictionary of string:values or None
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+
+        :raises TransformationError: if the lower or upper loop boundaries \
+            are not the same.
+        :raises TransformationError: if the loop step size is not the same.
+        :raises TransformationError: if the loop variables are not the same.
+        '''
```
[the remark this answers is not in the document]
hiker: *sigh* I just don't ... one day I'll get used to it :(

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # First check constraints on Nodes in the node_list inherited from
```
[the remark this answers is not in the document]
hiker: Indeed.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+            if var_name == loop_var1:
+                continue
+            var_info1 = vars1[var_name]
+            var_info2 = vars2[var_name]
+
+            # Variables that are only read in both loops can always be fused
+            if var_info1.is_read_only() and var_info2.is_read_only():
+                continue
+
+            if var_name in symbol_table:
```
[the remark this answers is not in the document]
hiker: I have re-implemented this, I believe the 'in' operation would have caused problems (since I believe that would only look at the specified symbol table, not the hierarchy of them). Please check if I did that correctly.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+            if not is_array:
+                NemoLoopFuseTrans.validate_scalar(var_info1, var_info2)
+            else:
+                NemoLoopFuseTrans.validate_array(var_info1, var_info2,
+                                                 loop_var1)
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def validate_scalar(var_info1, var_info2):
+        '''Validates if the accesses to the scalar ``var_name`` can be fused.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+        # about a conditional access, so the test below could result in
+        # incorrectly allowing fusion. But the test is essential for many
+        # much more typical use cases (especially inner loops).
+        if var_info1[0].access_type == AccessType.WRITE and \
+                var_info2[0].access_type == AccessType.WRITE:
+            return
+
+        raise TransformationError(
+            "Scalar variable '{0}' is written in one loop, but only read "
+            "in other loop.".format(var_info1.var_name))
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+        '''Validates if the accesses to the scalar ``var_name`` can be fused.
+
+        :param var_info1: access information for variable in the first loop.
+        :type var_info1: :py:class:`psyclone.core.var_info.VariableInfo`
+        :param var_info2: access information for variable in the second loop.
+        :type var_info2: :py:class:`psyclone.core.var_info.VariableInfo`
+
+        :raises TransformationError: a scalar variable is written in one \
+            loop, but only read in the other.
+        '''
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+            return
+
+        raise TransformationError(
+            "Scalar variable '{0}' is written in one loop, but only read "
+            "in other loop.".format(var_info1.var_name))
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def validate_array(var_info1, var_info2, loop_variable):
+        '''Validates if the accesses to the array ``var_name`` can be fused.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+        :param var_info1: access information for variable in the first loop.
+        :type var_info1: :py:class:`psyclone.core.var_info.VariableAccessInfo`
+        :param var_info2: access information for variable in the second loop.
+        :type var_info2: :py:class:`psyclone.core.var_info.VariableAccessInfo`
+        :param str loop_variable: name of the loop variable that will \
+            be fused.
+
+        :raises TransformationError: an array that is written to uses \
+            inconsistent indices, e.g. a(i,j) and a(j,i).
+        '''
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # First check if all accesses to the array have the same dimension
+        # based on the loop variable
+        # Now detect which dimension(s) is/are parallelised, i.e.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def validate_array(var_info1, var_info2, loop_variable):
+        '''Validates if the accesses to the array ``var_name`` can be fused.
+        :param var_info1: access information for variable in the first loop.
+        :type var_info1: :py:class:`psyclone.core.var_info.VariableAccessInfo`
+        :param var_info2: access information for variable in the second loop.
+        :type var_info2: :py:class:`psyclone.core.var_info.VariableAccessInfo`
+        :param str loop_variable: name of the loop variable that will \
```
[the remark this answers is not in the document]
hiker: I've changed this to be a DataSymbol now - I assume that's the correct type? I am still mostly using ``loop_variable.name`` in the subroutine though for now.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+        all_accesses = var_info1.all_accesses + var_info2.all_accesses
+
+        found_dimension_index = -1
+        all_indices = []
+
+        # Loop over all the accesses of this variable
+        for access in all_accesses:
+            list_of_indices = access.indices
+            # Now determine all dimensions that depend
+            # on the parallel variable:
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+
+            if not is_array:
+                NemoLoopFuseTrans.validate_scalar(var_info1, var_info2)
+            else:
+                NemoLoopFuseTrans.validate_array(var_info1, var_info2,
+                                                 loop_var1)
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def validate_scalar(var_info1, var_info2):
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+                var_info2[0].access_type == AccessType.WRITE:
+            return
+
+        raise TransformationError(
+            "Scalar variable '{0}' is written in one loop, but only read "
+            "in other loop.".format(var_info1.var_name))
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def validate_array(var_info1, var_info2, loop_variable):
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -0,0 +1,237 @@
...
+                if loop_variable not in accesses:
+                    continue
+
+                # If a previously identified index location does not match
+                # the current index location (e.g. a(i,j), and a(j,i) ), then
+                # the loop in general cannot be fused.
+                if found_dimension_index > -1 and \
+                        found_dimension_index != dimension_index:
+                    raise TransformationError(
+                        "Variable '{0}' is using loop variable {1} in index "
```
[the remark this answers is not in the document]
hiker: OK.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Indeed, but I think it is important to mention it since the tests rely on an exception to check that indeed incorrect fusion is detected. I've removed the ``:raises...`` part, and added a sentence to the description of the function.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Arg, indeed - I had to change that later :(

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Sound so nice.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Indeed.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Fixed.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Yep

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done.

hiker:
I believe all issues are addressed now. As suggested I also added support for compilation tests, but I had to rewrite the ``fuse_loops`` function to use ``generate_psyir``, since in my version the resulting Fortran code contained only the loops, not the whole subroutine. No idea if there is a better way of doing this, but it seems to work. Incidentally that got rid of the NEMO API parsing in that function, but I still had to use the ``NemoLoopFuseTrans`` (since only that function contains the new tests ... see below for what I tried). 

Note that in order to help me debug things I added a change to the ``string_compiles`` function so that this function uses a hash when generating the filenames - this way if you have more than one compile test in one test function, you still can see all the various generated files (assuming that you always create a new ``Compile`` instance).

I also tried to move the tests from ``NemoLoopFuse`` into the ``LoopFuseTrans`` base class. It mostly worked, but I ran into a few problems:
1) in gocean loops the 'same loop boundary' test got triggered, since e.g. the loop lower bounds could be ``a%x_start`` and ``b%x_start``. Indeed, the base transformation cannot know that these are the same, but the ``GOceanLoopFuseTrans`` knows and already tests for this. So we need to add an option to disable this test  (so the GOcean specific fuse transformation will test for same iteration space, but then disable the loop boundary test in the base class - that is actually quite nice, it shows the use of domain-specific knowledge).
2) I found some and fixed one issue where we had a test that fuses two loops that cannot be fused - in one case that was incorrect meta-data declaration, but in other cases (including lfric) it was not that obvious, and might need more work.

So I decided not to add more to this PR, and leave moving most (all?) of the Nemo loop fusion validation test to the base class for later.

hiker:
I had to work a bit for full coverage, and during this also discovered that a variable imported anonymously form a module was not handled as good as possible (I believe we all agree that in general we cannot handle this case anyway, e.g. 
```
use mymod
a=b+c
```
Since we don't know anything about the variables, we don't know if these are scalars, vectors, ...  At least by using the observed access pattern (i.e. if the next access uses array indices) we get a slightly better outcome in case that array indices are used.

Accepting this restriction, I think the code is now as good as it gets. One comment: For now the tests are in nemo only, and in nemo we always have a SymbolTable, so the code in ``nemo_loop_fuse``:
```
            try:
                # Find the symbol for this variable. We only need to check
                # one symboltable.
                ...
            except KeyError:
                # TODO #845: Once we have symbol tables, any variable should
                # be in a symbol table, so we have to raise an exception here.
```
does not make any sense (and I had to monkeypatch in order to create a test case), but the idea is that soon many (if not all) of these tests will be moved from NemoLoopFuse to the base class LoopFuse, and I believe we are still not at a stage that all APIs have all symbols (or???), so I prefer to leave this code for now in Nemo (it has a #TODO).

The missing code coverage line is only triggered when compiling, so that should be fine - ready for next review.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -108,12 +109,20 @@ def validate(self, node1, node2, options=None):
...
                 # Find the symbol for this variable. We only need to check
                 # one symboltable.
                 symbol = symbol_table.lookup(var_name)
-                is_array = symbol.is_array
-            else:
+                if isinstance(symbol, DataSymbol):
+                    is_array = symbol.is_array
+                else:
+                    # This typically indicates a symbol is used that we do
+                    # not have detailled information for, e.g. based on a
```
arporter, line 120: "detailed"

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -108,12 +109,20 @@ def validate(self, node1, node2, options=None):
...
+                    # generic 'use some_mod' statement. In thise case use
+                    # the information based on the access pattern, which
+                    # is at least better than having no information at all.
```
arporter, line 123: As discussed on Teams, I think this is good enough for this PR but please could you add a note that we could chase down the definition of the symbol. In contrast to what I said in Teams, we don't quite have *all* the necessary functionality for this since we would need to get a list of *all* Container symbols with 'wildcard' imports and then search each of them in turn, probably recursively. I expect we will eventually need this functionality in other places too so please could you create an Issue for it.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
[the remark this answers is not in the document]
arporter: Looks good to me. I would recommend having a test with a suitable variable defined in a module and then accessed in a subroutine - I haven't got as far as looking at the tests yet so you may have already done this.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -155,32 +166,36 @@ def validate_scalar(var_info1, var_info2):
...
     # -------------------------------------------------------------------------
     @staticmethod
-    def validate_array(var_info1, var_info2, loop_variable):
-        '''Validates if the accesses to the array ``var_name`` can be fused.
+    def validate_written_array(var_info1, var_info2, loop_variable):
+        '''Validates if the accesses to an array, which is at least written
+        once, allows loop fusion. The access pattern to this array is
+        specified in the two parameters (which includes the name of the
+        variable).
         :param var_info1: access information for variable in the first loop.
```
arporter, line 178: Blank line before `:param:` section please.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
arporter: Good idea.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -123,37 +122,43 @@ def test_fusetrans_error_not_same_parent():
...
-    psy = PSyFactory("nemo").create(ast)
-    schedule = psy.invokes.get("sub").schedule
-
-    loop1 = schedule.children[0]
-    loop2 = schedule.children[1]
+    fp2_ast = parser(reader)
+    fp2_reader = Fparser2Reader()
+    psyir = fp2_reader.generate_psyir(fp2_ast)
+    loop1 = psyir.children[0]
+    loop2 = psyir.children[1]
```
arporter, line 148: This touches on what I'm doing in #1210 - eventually you'll apply a Transformation to the generic PSyIR and it will give you NEMO-Specific PSyIR. In fact, please could you add a TODO here and reference that issue.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -436,8 +461,8 @@ def test_fuse_independent_array(parser):
     do j ...  d(j) = a(1)
     '''
 
-    # The first example can be merged, since 't' is read-only,
-    # so it doesn't matter that it is accessed differently
+    # The first example can be merged, since 's does not
```
arporter, line 464: Missing closing single quote after "s".

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -583,3 +608,63 @@ def test_fuse_scalars_incorrect(parser):
...
+      t(ji,jj) = s(ji,jj) + t(ji,jj)
+    enddo
+  enddo""" in out
+
+    # Case 2: remove the symbol table, which means that not
+    # even a Symbol instance is returned. This should then look
+    # up the information in the variable access information
+    # to determine which variables are an array
+    loop = psyir.children[0]
+    new_symbol_table = SymbolTable()
```
arporter, line 656: I can imagine that this might cause problems in the future (as the SymbolTable becomes more integrated). It might be better to be more 'surgical' and just remove the `ContainerSymbol` "mymod" from the SymbolTable - if that works?

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
```diff
@@ -583,3 +608,63 @@ def test_fuse_scalars_incorrect(parser):
...
+    fuse.apply(loop.loop_body.children[0], loop.loop_body.children[1])
+
+    writer = FortranWriter()
+    out = writer(psyir)
+    assert """  do jj = 1, n, 1
+    do ji = 1, 10, 1
+      s(ji,jj) = t(ji,jj) + 1
+      t(ji,jj) = s(ji,jj) + t(ji,jj)
+    enddo
+  enddo""" in out
```
arporter, line 670: Please could you also add a new test where e.g. `t` is defined in an outer module that then contains `subroutine sub`.

src/psyclone/tests/test_files/gocean1p0/boundary_conditions_ne_offset_mod.f90
```diff
@@ -16,10 +16,10 @@ module boundary_conditions_ne_offset_mod
   !=======================================
 
   type, extends(kernel_type) :: bc_ssh
-     type(go_arg), dimension(3) :: meta_args =                 &
-          (/ go_arg(GO_READ,      GO_I_SCALAR, GO_POINTWISE),  &
-             go_arg(GO_READWRITE, GO_CT,       GO_POINTWISE),  &
-             go_arg(GO_READ,                   GO_GRID_MASK_T) &
+     type(go_arg), dimension(3) :: meta_args =             &
+          (/ go_arg(GO_READ,  GO_I_SCALAR, GO_POINTWISE),  &
+             go_arg(GO_WRITE, GO_CT,       GO_POINTWISE),  &
```
arporter, line 21: Not sure where these changes have come from but I don't think they're right - a boundary-condition kernel doesn't write to all elements of a field and therefore we must preserve those input values that are not modified.

arporter (changes requested):
Looking good now. Thanks for adding all the compilation checks.
Building the reference guide reveals one new warning:

    PSyclone/src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py:docstring of psyclone.domain.nemo.transformations.nemo_loop_fuse.NemoLoopFuseTrans.validate_written_array:1: WARNING: Inline interpreted text or phrase reference start-string without end-string.

Just a bit of tidying to do and one test to add.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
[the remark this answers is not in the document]
hiker: I would actually say that this needs to be tested in the symbol table code (or as part of #1213), imho it doesn't make any sense to add this when evaluating loop fusion.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
hiker: Done.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
hiker: Done, #1213.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Indeed.

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: I thought I had tried that, and got an exception that this type of symbol could not be removed ... I tried again, and it works now (oh, and I need to remove the  symbol 't', not 'mymod' - this way 't' is not found in the symbol table, which then triggers the use of the 'symbol not in symbol table' branch in the code).

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
hiker: Done (though, as above, I would actually say that this should be done in the symbol table test, I am just using existing functionality here).

src/psyclone/tests/test_files/gocean1p0/boundary_conditions_ne_offset_mod.f90
hiker: The code in question is:
```
    if(tmask(ji,jj) <= 0) return
    IF     (tmask(ji,jj-1) < 0) THEN
       ssha(ji,jj) = amp_tide * sin(omega_tide * rtime)
    ... similar else if tests with assignmne tto ssha()
```
The variable ``ssha`` is NOT read, so as far as I can see the declaration as 'readwrite' was wrong (and I just included this what I thought was a simple typo here). It led to problems when I tried to use these tests for gocean (lowering them from ``NemoLoopFuse`` to ``LoopFuse``). I think a completely unrelated test failed (since it tried to fuse something which then failed). If this is an issue, I am happy to revert this and handle this later.

hiker:
Ready for next review (coverage failure is caused by a change to a line that's used in compilation tests only).

hiker:
> Building the reference guide reveals one new warning:
> 
> ```
> PSyclone/src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py:docstring of psyclone.domain.nemo.transformations.nemo_loop_fuse.NemoLoopFuseTrans.validate_written_array:1: WARNING: Inline interpreted text or phrase reference start-string without end-string.
> ```

I think I've fixed that, but it would be good if someone could help me with #1178, since I can't test this myself ;(

src/psyclone/tests/test_files/gocean1p0/boundary_conditions_ne_offset_mod.f90
arporter: The problem here I think is the mapping between the access described in the metadata and the INTENT used in Fortran. We should probably give arguments that are `GH_WRITE` intent `INOUT`. Otherwise, if we just give it intent `OUT` then Fortran is well within its rights to wipe the array before passing it into the kernel. Obviously we don't want this to happen given that we will typically be only writing to part of the data array (even in a 'standard' kernel we don't write to the halos for instance). This is a change (I think) that @TeranIvy recently had to make for the LFRic API. If I've got all that right then I suggest creating a new issue for this and putting the meta-data and intent back to how it was along with a TODO.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -109,16 +109,17 @@ def validate(self, node1, node2, options=None):
             if var_info1.is_read_only() and var_info2.is_read_only():
                 continue
 
+            # TODO #1213 - try to find symbol in case of 'wildcard' imports.
             try:
                 # Find the symbol for this variable. We only need to check
```
arporter, line 114: I didn't notice this before but what does "We only need to check one symbol table" mean?

src/psyclone/tests/psyir/transformations/loop_fusion_test.py
arporter: I can see your point but this does check that you're using the existing functionality correctly :-) Thanks for adding it.

src/psyclone/tests/test_files/gocean1p0/boundary_conditions_ne_offset_mod.f90
hiker: OK, I'll do that - but I believe the issue you are referring to is that if you pass an object to a subroutine, Fortran needs to read the object information (in order to get access to the pointer to the data it is writing). So I think the issue is that the meta-data should indicate "READ", but the intent must be "INOUT".

hiker: #1215, but I believe the original issue (passing in objects, so read access is required) does not apply to gocean, which passes in Fortran arrays. But then I found several issues (in the same file) where a ``GO_READ`` is matched with ``intent(inout)``.

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
hiker: That you don't need to look manually at outer scopes - but agrees, that was bad wording, I have changed that.

hiker:
@arporter  Oops - I only now noticed that it still had the 'under review' state ... sorry, I just pushed with two changes (undoing the kernel change, updating the comment).

src/psyclone/tests/test_files/gocean1p0/boundary_conditions_ne_offset_mod.f90
arporter: Having slept and looked at this again, I think we should just change the kernel definition (I'd got confused and forgotten that we don't generate this) so that the field that is GO_WRITE has INTENT(inout). That should then keep everyone happy?
If you agree then please make this change and then we're good to go.

arporter (changes requested):
All good now apart from my metadata confusion.
Tests and examples all OK.
Reference guide builds OK.

Once we've sorted out the boundary_conditions_ne_offset kernel this can be merged.

src/psyclone/tests/test_files/gocean1p0/boundary_conditions_ne_offset_mod.f90
arporter: `ssha` is only written and has `intent(inout)` so I'm happy.

arporter (approved):
All changes have been made.
All docs build with no new warnings.
Once 2.0 is released, this can be merged.

PR #1191, opened 2021-04-01
Rationalise config file list processing (closes #1163)

### LINKED ISSUE #1163: lists in the config file
ConfigParser does not support lists natively, it just returns a string, so the reader has to handle it manually. At the moment there are two different formats for list in the config file, one just with spaces and one with commas. The user must keep to these different formats for the entries otherwise the code will break. I don't like this and I don't like the fact that the code developer has to handle the formatting each time.

What I suggest is that we add the following to ConfigParser when we create it ...

```
ConfigParser(converters={'list': lambda x: [i.strip() for i in x.split(',')]})
```

and then access all lists like this ...

```
section.getlist(name)
```

This would mean that all lists would have to be comma separated.

rupertford:


rupertford:
Note, this is a non-backwards-compatible change. The config file previously supported a list of things as space separated in one case and comma separated in others. Now, all lists must be comma separated.

rupertford:
Ready for first review. Probably from @hiker seeing as he is the config king, but very happy for anyone else to pick it up if they have the time and/or inclination.

src/psyclone/configuration.py
```diff
@@ -734,34 +736,32 @@ def __init__(self, section):
...
+    def create_dict_from_list(input_list):
+        '''Takes a list of strings each with the format: key:value and creates
+        a dictionary with the key,value pairs. Any leading or trailing
+        white space is removed.
+
+        :param input_list: the input list.
+        :type input_list: list of str
+
+        :returns: a dictionary with the key,value pairs from the input \
+            string.
```
hiker, line 748: s/string/list/

src/psyclone/configuration.py
```diff
@@ -734,34 +736,32 @@ def __init__(self, section):
...
-        for entry in entries:
+        for entry in input_list:
             try:
                 key, value = entry.split(":", 1)
             except ValueError:
                 # Raised when split does not return two elements:
                 raise ConfigurationError("Invalid format for mapping: {0}".
                                          format(entry.strip()))
-            return_dict[key.strip()] = value.strip()
+            # Remove spaces and convert unicode to normal strings in Python2
```
hiker, line 763: Good comment that ... I was wondering about the additional str() :)

hiker (approved):
All tests, examples and tutorial work, no need to update manual. Only one minor typo, which I take the liberty to fix myself.

Issue #753, opened 2020-04-24
[PSyIR] translate the algorithm layer to PSyIR


rupertford:
The algorithm layer is currently parsed and transformed with fparser2. We should use the PSyIR instead for flexibility and to support potential future optimisations. The idea would be to parse to low level PSyIR then to recognise invoke regions, Kernels and arguments with particular types (fields, scalars, ...) and have domain-specific nodes for these. In the first instance we would need to be able to re-create the existing algorithm transformation performed by fparser2.

Once the above is working then additional functionality to try to determine the datatypes, function spaces etc. of algorithm arguments (within invokes) that passed into the PSy-layer. This could then be used to perform static correctness checks between algorithm fields and kernel arguments (specified with metadata).

This would complement issue #79.

arporter:
This issue will also remove the limitations we currently have on the form of arguments that can be passed in an Invoke. (e.g. specifying a stencil depth as an integer literal with explicit kind fails.)

rupertford:
PSyIR has made significant progress and is nearly at the stage where PSyclone could use the PSyIR to translate and analyse the algorithm layer. The following code parses simple examples:

```
from fparser.two.parser import ParserFactory
from fparser.common.readfortran import FortranFileReader
from psyclone.psyir.frontend.fparser2 import Fparser2Reader
from psyclone.psyir.backend.fortran import FortranWriter

# Parse code with fparser2
reader = FortranFileReader("alg.x90",
                           ignore_comments=True)
f2003_parser = ParserFactory().create(std="f2003")
parse_tree = f2003_parser(reader)

# Transform code from fparser2 to psyir
processor = Fparser2Reader()
name = "nonsense_alg"
container = processor.generate_container(parse_tree)
schedule = processor.generate_schedule(name, parse_tree, container=container)

# TODO transform from psyir to lfric algorithm psyir

# Take a look
container.view()

# TODO transform alg to processed alg

# print the transformed output
writer = FortranWriter()
code = writer(container)
print (code)
```

rupertford:
Here is a list of known issues that need to be addressed/understood

1) calls are treated as codeblocks as the front-end does not yet parse them
2) Can we have a generic Fparser2Reader.generate() that just reads Fortran code, without the separate container and subroutine steps?
3) Seems to be a problem when leaving in comments
4) Test that code generated is correct for all LFRic algorithm files (run through this then parse again in the original way?)
5) Write psyir to lfric_algorithm_psyir transform
[...]

PR #1192, opened 2021-04-05
Translate Alg PSyIR to processed PSyIR (towards issue #753)

rupertford:


rupertford:
@sergisiso has pointed out that it would make sense to use the `lower_to_language_level` method rather than a transformation. The one known issue with this approach is that we need to know the index of an `invoke`. After discussion, the suggestion is to store the index of the invoke when going from generic (language-level) to domain-specific PSyIR.

rupertford:
Ready for first review. Probably best for @sergisiso to take this seeing as he has already commented on it and suggested modifications. Note, there are no examples or documentation yet as it only makes sense to add these once this new approach replaces the current one.

Next steps will be to check this approach works with the `gocean` and then `lfric` APIs before making it the the new default approach for generating processed algorithm PSyIR. After that, the algorithm PSyIR should be passed into the PSy-layer to replace the current solution.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -61,6 +76,95 @@ def _validate_child(position, child):
...
+        :rtype: str
+
+        '''
+        routine_root_name = "invoke_{0}".format(index)
+        if len(self.children) == 1:
+            # Add the name of the kernel if there is only one call
+            routine_root_name += "_" + self.children[0].name
+        return routine_root_name
+
+    def create_language_level_symbols(self, index):
```
sergisiso, line 96: I may be missing some use case but I don't understand why this is a public method and takes an external `index`. 

This make the interface on all the tests be:
```
    psyir = create_alg_psyir(code)
    invoke = psyir.children[0]
    invoke.create_language_level_symbols(0)
    invoke.lower_to_language_level()
```
Where you must know the index that this invoke has in the algorithm code, disconnected after parsing the algorithm code. Also the lowering method fails if this hasn't been called first, this doesn't look very safe.

I think the indexing should be assigned inside `alg_trans.apply(psyir)` (as I remember we said that the number is given from the initial position of the invoke - when raising(before any Psyclone script transformation)) and should possibly be stored as a string (e.g passed as an string index parameter in `__init__` or in `create()`) and this be fixed for the live of the AlgorithmInvokeCall (*with some transformations like fusing invokes able to update it to "_index1_index2"). Although I don't mind if it is an integer for now until we do fuses.

Then the lowering method should directly call `self._create_language_level_root_name(self._index)` instead of possibly failing. The symbols shouldn't exist while they are not used (as they may not exist at all depending on the fusing operations).

Does that make sense?

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -36,18 +36,33 @@
...
+        or NoneType
+
+    '''
     _children_valid_format = "[KernelFunctor]*"
     _text_name = "AlgorithmInvokeCall"
     _colour = "green"
 
+    def __init__(self, routine, parent=None):
+        super(AlgorithmInvokeCall, self).__init__(routine, parent=parent)
+        self._routine_symbol = None
```
sergisiso, line 63: I think these 2 attributes should probably not exist and have a index attribute instead. See previous comment.

sergisiso (changes requested):
@rupertford I am afraid I am coming with another interface level comment here. However the rest of the code and tests looks fine, there is just the issue of how the index should be assigned and when the symbols created.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: The problem is that the PSyclone Algorithm-layer is actually transformed into a processed algorithm-layer *and used to generate the PSy-layer*. When transforming (lowering) the PSyclone Algorithm-layer to processed Algorithm-layer we destroy the algorithm layer information that is required to generate the PSy-layer. This means we must generate the PSy-layer first (or copy everything). In order to generate the PSy-layer we need the `container` and `routine` symbols that are going to connect the PSy-layer and the processed algorithm layer. It is possible we could defer these but my working assumption is that they are needed at this point. These symbols are required for *both* the processed algorithm layer and the PSy-layer (as the former calls the latter). I would be very happy to discuss how and when we should create these symbols and where they should be stored.

The next question is when to create an index for an invoke (which is needed when generating the container and routine symbols). I was concerned about creating the indices as we raise the generic PSyIR to Algorithm-layer PSyIR as we would be relying on the processing happening in a particular order. I therefore deferred providing this value as well (and merged it with the creation of the required symbols). However, you are right that the order is determined in `alg_trans.apply()` so I should and will provide the index here instead.

The steps that I think are needed are below and I guess we need to agree on how to support step 3 - or agree on a different approach.

1. Create Algorithm-layer PSyIR, including invoke index
2. Transform/optimise Algorithm-layer PSyIR (merge invokes etc and update index as appropriate)
3. Generate Container and Routine symbols
4. Create PSy-layer PSyIR using Algorithm-layer PSyIR and Container and Routine symbols (+ kernel metadata)
5. Create Processed (lowered) Algorithm PSyIR from Algorithm-layer PSyIR using Container and Routine symbols
6. Transform/optimise Processed Algorithm PSyIR, PSy-layer PSyIR and Kernel PSyIR

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -36,18 +36,33 @@
...
+        self._container_symbol = None
```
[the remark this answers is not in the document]
rupertford: For the processed algorithm layer yes, but not for the PSy-layer. This is storing the information before processing - please see the next converstation for context.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: I agree that it makes sense to store the index separately. However, I think the routine and container symbols need to be stored somewhere - please see previous comment for context.

rupertford:
Thanks for the review @sergisiso. I've tried to explain the problem with the container and routine symbols as I see it. If you could take a look at my response then we can discuss the best way forward. I've made this ready for review again to indicate a request for feedback.

src/psyclone/domain/common/algorithm/psyir.py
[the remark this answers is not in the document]
sergisiso: We agreed in the call that the container reference is probably not needed as it is given by the routine_symbol. Also to differentiate clearly the routine and routine_symbol purpose.

src/psyclone/domain/common/algorithm/psyir.py
sergisiso: We agreed in the call that providing the index must be done in the constructor but there are alternatives about when exactly to create the symbols. This depends on the ordering of steps the PSyclone does. The main alternatives are: create in the Transformation and provide them in the constructor, or create them by a getter and store them (subsequent calls to the get return the already created), in both cases they need to be stored in a attribute inside by the class.

src/psyclone/domain/common/algorithm/psyir.py
sergisiso: I agree with this

sergisiso:
@rupertford I returned the Reviewed with actions label and added comments with what we discussed.

src/psyclone/domain/common/algorithm/psyir.py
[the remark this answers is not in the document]
rupertford: Hopefully done the above.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: Moved setting of the index to the constructor. Note, this could be changed to a list of integers in the future which would allow the capturing of the history of any invoke fusion.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: Updated.

rupertford:

src/psyclone/domain/common/transformations/alg_trans.py
```diff
@@ -88,9 +88,11 @@ def apply(self, psyir, options=None):
 
         '''
         self.validate(psyir, options=options)
+        idx = 0
         for call in psyir.walk(Call):
             if call.routine.name.lower() == "invoke":
-                self._invoke_trans.apply(call, options=options)
+                self._invoke_trans.apply(call, idx, options=options)
```
sergisiso, line 94: I did not forsee that the index would be computed in the meta-transformation that calls the InvokeCallTrans, but this is right, it should be done here.

I am a bit uncomfortable breaking the transformation abstract apply() API with the additional argument but a grep shows that it in not the only transformation that does this. Maybe when we have the different raising->psyclone->lowering steps decided and how it works in the other APIs we can give this another look.

sergisiso (changes requested):
@rupertford Thanks for addressing my comment. I still have a some more docstring/naming suggestion but this should be easier now.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -36,18 +36,80 @@
...
+            raise ValueError(
+                "AlgorithmInvokeCall index argument should be a non-negative "
+                "integer but found {0}.".format(index))
+
+        self._index = index
+        self._language_level_routine_symbol = None
+
+    @classmethod
+    def create(cls, routine, arguments, index):
+        '''Create an instance of class cls given valid instances of a routine
```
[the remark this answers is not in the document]
rupertford: Updated to "the calling class" as agreed via telco.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -87,7 +221,7 @@ def __init__(self, symbol, parent=None):
     @classmethod
     def create(cls, symbol, arguments):
         '''Create an instance of the calling class given valid instances of a
-        TypeSymbol, and a list of child nodes for its arguments.
+        TypeSymbol, a list of child nodes for its arguments an an index.
```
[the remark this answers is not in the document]
rupertford: Oops.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -36,18 +36,80 @@
...
+            raise TypeError(
+                "AlgorithmInvokeCall index argument should be an int but "
+                "found '{0}'.".format(type(index).__name__))
+        if index < 0:
+            raise ValueError(
+                "AlgorithmInvokeCall index argument should be a non-negative "
+                "integer but found {0}.".format(index))
+
+        self._index = index
+        self._language_level_routine_symbol = None
```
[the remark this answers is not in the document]
rupertford: I was going to argue with you about this but actually you are completely right - the symbols are PSy-layer symbols. The only reason we are storing them in the algorithm layer is to avoid clashes. However, languages usually allow renaming when importing, so we could avoid clashes that way instead. Therefore, a more logical long term solution would be to generate the symbols in the PSy-layer and pass them to the algorithm lower method. Anyway, given that renaming is not something we support at the moment, I will plough ahead with the current approach regardless, and rename as you suggest.

src/psyclone/domain/common/transformations/alg_trans.py
rupertford: I'm happy with that.

rupertford:

sergisiso (approved):
@rupertford This looks fine now, thanks for addressing the comments. I checked all code, pylint, pycodestyle and tests this time. I think user/developer guide documentation can wait until the #753 is complete and we have the full picture. This PR is ready to be merged

PR #1195, opened 2021-04-07
#1194 - fix str(err) -> str(err.value)

### LINKED ISSUE #1194: Fix asserts using str(err)
Similarly to #972, some tests fail due to `assert` statements which use `str(err)` rather than `str(err.value)`.

AdamVoysey:
fixes #1194 - change instances of `str(err)` to `str(err.value)`


src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1400,7 +1401,7 @@ def test_symbol_from_tag():
     with pytest.raises(SymbolError) as err:
         symtab.symbol_from_tag("tag3", symbol_type=RoutineSymbol)
     assert ("Expected symbol with tag 'tag3' to be of type 'RoutineSymbol' "
-            "but found type 'DataSymbol'." in str(err))
+            "but found type 'DataSymbol'." in str(err.value))
```
arporter, line 1404: Thanks for fixing. Is there some way we can reproduce the failure in the test suite on GitHub Actions so that we don't keep making this mistake?

arporter (changes requested):
Thanks for fixing these Adam. I think it raises two things though:

 1. The tests on GitHub Actions are not picking up the incorrect usage, even in tests that (presumably) are being run. Presumably we can change something so that they do?
 2. Some of the test utility code is not covered by the tests. That's not in the scope of this PR though so I'll create a new issue.

AdamVoysey:
I have managed to reproduce the difference in behaviour for the two systems, and it is dependant on the version of pytest used. I have further tracked it down to the particular change that cased this - https://github.com/pytest-dev/pytest/pull/5934.

What appears to be happening is on the older versions of pytest, the name of the exception class is used within `str(err)`. However in the newer versions, the `__repr__` method of the underlying exception is used. As none of the Psyclone exceptions define `__repr__`, we instead fall back to the inherited parent `Exception.__repr__` - which includes the error message within it. The assertions therefore work by accident in this case, as they are always able to match the error message for the exception.

To combat this, I have added a `__repr__` to each of the Psyclone exception classes which is the same as that which pytest would previously have used. I have additionally added a new unit test which walks the Psyclone code to find all the exceptions, then ensures they define an appropriate `__repr__`.

With these changes, the assertions pass when `str(err.value)` is used, across all pytest versions. Furthermore, there are failures if only `str(err)` is used across all pytest versions.

Hopefully this is an acceptable solution to the problem.

arporter (changes requested):
Thanks for taking the time to dig into this and find the root cause of the problem, it's much appreciated.
Although the changes fix the problem, I think it might be better to refactor things a bit and introduce an abstract 'base' class `PSycloneException` which all of our other exceptions should subclass. I *think* we could just then implement `repr` in this base class as:

    def __repr__(self):
        return type(self).__name__ + "()"

and therefore not need it anywhere else. Would you be happy to do this or would you prefer to create a new Issue for someone else to tackle?

I like the idea of the new test but found it hard to follow and codecov (https://app.codecov.io/gh/stfc/PSyclone/compare/1195/diff) points out that there are a few lines not covered. I had a play and came up with something a bit simpler (still using your `all_sub_exceptions()` routine):

    def import_submodules(package, recursive=True):
        """ Import all submodules of a module, recursively, including subpackages
    
        :param package: package (name or actual module)
        :type package: str | module
    
        :rtype: dict[str, types.ModuleType]
        """
        if isinstance(package, str):
            package = importlib.import_module(package)
        results = {}
        for loader, name, is_pkg in pkgutil.walk_packages(package.__path__):
            full_name = package.__name__ + '.' + name
            if "test" not in full_name:
                results[full_name] = importlib.import_module(full_name)
                if recursive and is_pkg:
                    results.update(import_submodules(full_name))
        return results
    
    if __name__ == "__main__":
    
        mods = import_submodules("psyclone")
        for mod in mods:
            package = importlib.import_module(mod)
        exceptions = all_sub_exceptions(Exception)
        for exc in exceptions:
            if "psyclone." in str(exc):
                print(exc)

AdamVoysey:
Sorry this took so long to respond to - I was called away to work on other things. I have now simplified the test and added a new abstract base class for the exceptions as suggested.

src/psyclone/tests/exceptions_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Recursively walk through the psyclone module, importing sub-modules.
+    # Store any class definitions we come across.
+
+    modules = import_submodules("psyclone")
+    for mod in modules:
+        package = importlib.import_module(mod)
+
+    all_excpetions = all_sub_exceptions(Exception)
+    psy_excepts = [exc for exc in all_excpetions if "psyclone." in str(exc)]
```
arporter, line 99: Could we now have `PSycloneError` instead of `Exception` on the line above and then remove this line entirely?

src/psyclone/tests/exceptions_test.py
```diff
@@ -0,0 +1,138 @@
...
+    psy_excepts.append(DummyPsycloneError)
+
+    # Different vertions of pytest behave differently with repect to their
+    # handeling of an exception's representation. This can lead to some tests
```
arporter, line 103: "handling"

src/psyclone/tests/exceptions_test.py
```diff
@@ -0,0 +1,138 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2017-2021, Science and Technology Facilities Council.
```
arporter, line 4: Just 2021 please.

src/psyclone/psyir/transformations/transformation_error.py
```diff
@@ -36,14 +36,12 @@
...
+class TransformationError(PsycloneError):
     ''' Provides a PSyclone-specific error class for errors found during
         code transformation operations. '''
 
     def __init__(self, value):
-        Exception.__init__(self, value)
-        self.value = "Transformation Error: "+value
-
-    def __str__(self):
-        return repr(self.value)
```
arporter, line 49: Please could you add:

    # For AutoAPI documentation generation
    __all__ = ["TransformationError"]

to the bottom of this file.

src/psyclone/errors.py
```diff
@@ -32,51 +32,56 @@
...
     :param str value: the message associated with the error.
     '''
     def __init__(self, value):
-        Exception.__init__(self, value)
-        self.value = "PSyclone internal error: "+value
-
-    def __str__(self):
-        return str(self.value)
+        PsycloneError.__init__(self, value)
+        self.value = "PSyclone internal error: "+str(value)
```
arporter, line 87: Not your fault but, since we're here, please could you add an `__all__` at the bottom of this file and list the classes it contains:

    # For Sphinx AutoAPI documentation generation
    __all__ = ["PSycloneError", etc. etc.]

src/psyclone/alg_gen.py
```diff
@@ -40,21 +40,18 @@
 '''
 
 from __future__ import absolute_import
+from psyclone.errors import PsycloneError
```
arporter, line 43: For consistency with other class names, please could you make this "PSycloneError" rather than "PsycloneError".

arporter (changes requested):
Thanks very much for re-factoring Adam.
Just some minor things to tweak and I think we should call it `PSycloneError` for consistency with other classes (apologies if I suggested `PsycloneError` in the first place). Once that's done this can be merged.

src/psyclone/tests/exceptions_test.py
AdamVoysey: In principle yes - but I think I would argue this should remain as it is to make the test more robust. If a future change added an exception that (accidentally) did not inherit `PSycloneError`, it would be still be picked up by the current formulation; however it would be missed if suggested change was made.

AdamVoysey:
I think I've made all the requested changes except the `all_sub_exceptions(Exception)` -> ` all_sub_exceptions(PSycloneError)` as I think this will make the test more fragile. Happy to change it if you disagree with this reasoning.

src/psyclone/tests/exceptions_test.py
arporter: Actually, could we therefore go one better and check that every exception we find is an instance of `PSycloneError`. Should just be one or two extra lines?

src/psyclone/psyir/transformations/transformation_error.py
arporter: Thanks for doing this. Unfortunately I've discovered that we now get a truckload of warnings when building the reference guide I've asked a question on the sphinx user group (https://groups.google.com/g/sphinx-users/c/vuW6OOb96Yo) so we'll see what they say.

arporter (changes requested):
Very, very nearly there now.
Please could you extend the exception testing to check that all exception classes defined by PSyclone sub-class PSycloneError.
There's also a problem with warnings generated when building the documentation. This is not your problem though.

src/psyclone/psyir/transformations/transformation_error.py
arporter: Please could you just comment-out this `__all__` line for now with a comment to say that it causes 'more than one target for cross-reference' warnings when building the reference guide and add a TODO referring to #1280.

arporter:
Please could you also bring this branch up-to-date with master.

AdamVoysey:
Changes now made

arporter (approved):
All tests and examples OK with compilation.
Will proceed to merge.

PR #1205, opened 2021-04-13
(closes #870) rm support for iterates_over in lfric metadata

### LINKED ISSUE #870: [i-first] remove support for `iterates_over` kernel metadata in LFRic
#866 introduces the `operates_on` kernel metadata entry. However, in order to avoid disrupting PSyclone development it did not remove support for `iterates_over`. That will be done in this issue.

arporter:


arporter:
Pretty straightforward change. The only wrinkle was that I found an `if..elif..else` block that didn't catch an unexpected value so I've extended it and added a test. (The value of this is that it makes developing code much easier as it will catch the situation where you've updated one thing but overlooked another.)

arporter:
Assuming the CI is happy, I think this is almost ready to go. The only question is (for @TeranIvy), should I update `argument_mod.F90` in this PR or are you doing that in #874?

TeranIvy:
> 
> 
> Assuming the CI is happy, I think this is almost ready to go. The only question is (for @TeranIvy), should I update `argument_mod.F90` in this PR or are you doing that in #874?

I am happy to do this in #874 to keep it in line with LFRic so please pass me this PR for review when it is ready, @arporter.

arporter:
It's all green so over to @TeranIvy for review... :-)

src/psyclone/dynamo0p3.py
```diff
@@ -6549,12 +6548,17 @@ def __init__(self, parent=None, loop_type=""):
...
                 tag = "dof_loop_idx"
                 suggested_name = "df"
-            else:
+            elif self.loop_type == "":
                 tag = "cell_loop_idx"
                 suggested_name = "cell"
+            else:
+                raise InternalError(
+                    "Unsupported loop type '{0}' supplied. Supported values "
+                    "are 'colours', 'colour', 'dof' or '' (for cell-columns).".
```
TeranIvy, line 6560: What about "null" in the `VALID_LOOP_TYPES` above? If this is supposed to be in the list as well then I would suggest using the list for the error message here.

src/psyclone/dynamo0p3.py
```diff
@@ -6528,14 +6519,22 @@ def _compute_from_field(self, field):
 
 
 class DynLoop(Loop):
-    ''' The Dynamo specific Loop class. This passes the Dynamo
+    '''
+    The Dynamo specific Loop class. This passes the Dynamo
```
TeranIvy, line 6523: I would suggest using "LFRic" in this docstring. Also, "LFRic-specific Loop class" (compound adjective).

src/psyclone/tests/domain/lfric/lfric_loop_test.py
```diff
@@ -59,6 +60,23 @@
...
 
 
+def test_constructor_invalid_loop_type(monkeypatch):
+    ''' Check that the constructor raises the expected errors when an invalid
+    loop type is specified. '''
+    # An invalid type should be caught by the setter in the base Loop class.
+    with pytest.raises(GenerationError) as err:
+        DynLoop(loop_type="wrong")
+    assert ("Error, loop_type value (wrong) is invalid. Must be one of "
+            "['dof', 'colours', 'colour', '', 'null']." in str(err.value))
```
TeranIvy, line 70: How about using `VALID_LOOP_TYPES` from `dynamo0p3.py` here?

src/psyclone/parse/kernel.py
```diff
@@ -814,6 +814,9 @@ def __init__(self, ast, name=None):
         self._name = name
         self._ast = ast
         self._ktype = get_kernel_metadata(name, ast)
+        # TODO #1204 since the valid form of the metadata beyond this point
```
TeranIvy, line 817: Thanks for adding the "TODO" here. I think the Copyright year should be changed to 2021.

TeranIvy (changes requested):
Looks good, I only have some small remarks (see below).

The changes are `pycodestyle` and `pylint` clean. The tests with compilation run, examples run and documentation builds fine.

All changes are covered. There are no references to #870 except in the `argument_mod.F90` as we agreed.

src/psyclone/dynamo0p3.py
arporter: No, this is just for the loops that have a variable. I've updated the error message to make this clearer.

src/psyclone/dynamo0p3.py
arporter: Oh yes, done.

src/psyclone/tests/domain/lfric/lfric_loop_test.py
arporter: Oh yes, done.

src/psyclone/parse/kernel.py
arporter: Done.

arporter:
Ready for another look now.

src/psyclone/dynamo0p3.py
TeranIvy: Thanks for the explanation!

TeranIvy (approved):
All fine now. To be on the safe side I retested the branch with compilation and checked again for `pycodestyle`, `pylint` and running of examples. `pylint` reported some wrong indentations in `transformations.py` not related to this PR so I took the liberty of fixing them.

All changes are covered. Updated the changelog and User Guide. Assuming that Travis CI is fine we are ready to go.

PR #1207, opened 2021-04-15
(closes #1184) extends the Visitor base class to check global constraints for each node visited

### LINKED ISSUE #1184: [psyir] Perform tree validation prior to code generation
In the 'old' way of doing things, the `gen_code` routines perform validation (where necessary) before generating code. This is necessary because we do not mandate the order in which Transformations are applied and therefore we can only check that the tree is correct at the point of code generation. Some (but not all) nodes have the `_pre_gen_validate()` method which performs these checks.
With the move to the use of a Visitor to generate code from the PSyIR, these checks are no longer being performed. The best way of implementing them is not yet clear. Some options:

 1. Have an entirely separate visitor that performs validation;
 2. Update the Visitor base class to call something like `_pre_gen_validate()` on each node;

Currently, I favour 2. as the checks are language independent and are specific to a particular node type.

arporter:


arporter:
In doing this PR I noticed that the 'skip-nodes' functionality in the Visitor base class wasn't working properly - no text would be returned if an unsupported node was skipped. However, @sergisiso has asked whether we in fact need this functionality anyway? @rupertford I see you have tests for it in the SIR visitor tests. Do you need it for SIR support? 

I was thinking it may ease the implementation (or rapid prototyping) of new visitors but have realised that all that is required in such cases is that the new visitor implement `node_node` which simply returns the text representation of all its children.

Does @hiker have any thoughts as well?

arporter:
Nudging @rupertford for his views :-)

rupertford:
> In doing this PR I noticed that the 'skip-nodes' functionality in the Visitor base class wasn't working properly - no text would be returned if an unsupported node was skipped. However, @sergisiso has asked whether we in fact need this functionality anyway? @rupertford I see you have tests for it in the SIR visitor tests. Do you need it for SIR support?
> 
> I was thinking it may ease the implementation (or rapid prototyping) of new visitors but have realised that all that is required in such cases is that the new visitor implement `node_node` which simply returns the text representation of all its children.

I agree it is not required but would argue it could still be a useful/convenient feature and I don't think it is a bad thing. The fix to the implementation of Visitor is also pretty simple (concatenate and return a list of strings). However, if people disagree, it is not required for the SIR backend, although it would be more messy to implement what I've used it for. 

I use it in the SIR visitor to allow the option of SIR being generated even if an unsupported node is found (but to fail by default). As @arporter suggests, this is mainly useful when developing, however, I could imagine someone implementing a visitor for a particular subset of nodes and not caring about the others - in this case it would avoid the need for any `node_node` code.

arporter:
Thanks @rupertford. Since you care and you use it I shall fix it rather than remove it :-)

arporter:
Ready for a first look from either @rupertford or @sergisiso.

doc/developer_guide/psyir_backends.rst
```diff
@@ -149,9 +149,8 @@ One example of the power of this approach makes use of the fact that
 all PSyIR nodes have `Node` as a parent class. Therefore, some base
 functionality can be added there and all nodes that do not have a
 specific method implemented will call this. To see the
-class hierarchy, the following code can be written:
+class hierarchy, the following code can be written::
 
-::
 
    from __future__ import print_function
```
sergisiso, line 155: Please fix indentation in this example

src/psyclone/psyir/backend/visitor.py
```diff
@@ -154,21 +169,27 @@ def _visit(self, node):
         :param node: A PSyIR node.
         :type node: :py:class:`psyclone.psyir.nodes.Node`
 
+        :returns: representation of the sub-tree with `node` at its root.
```
sergisiso, line 172: Maybe "text representation of the given PSyIR node sub-tree"

sergisiso (changes requested):
@arporter The code looks good (thanks for cleaning up the Visitor and documentation), pytest passes and all new code is covered. New functionality is documented and developer and reference guide build without new issues.

I was about to approve this PR but I found a TODO mentioning this issue in src/psyclone/psyir/backend/fortran.py:1296 . Can this TODO be deleted now? Maybe adding just one test triggering a global constraint GenerationError in src/psyclone/tests/psyir/backend/fortran_test.py for a directive will be good/clarifying (even though it is not strictly unit testing as it is inherited functionality)?

doc/developer_guide/psyir_backends.rst
arporter: Oops, thanks for spotting.

src/psyclone/psyir/backend/visitor.py
arporter: Done.

arporter:
Thanks for finding that TODO. I've removed it now and extended one of the PSyIR OpenMP tests as that seemed a sensible place to put it.
Assuming the CI is happy this is ready for another look.

sergisiso (approved):
@arporter It all looks good now. PR ready to merge.

Issue #737, opened 2020-04-06
Going from generic PSyIR to Domain-Specific PSyIR


arporter:
In #596 the process of creating PSyIR for the NEMO API has been brought into line with the way Kernel PSyIR is created. While discussing that PR @rupertford suggested that the PSyIR frontend be purely responsible for creating the 'lowest level' PSyIR of the source code being processed. This PSyIR could then be processed to 'raise' the level of abstraction. e.g. in the NEMO API we could convert recognised Fortran Do Loops into instances of `NemoLoop`, the bodies of such loops into instances of `NemoCodedKernel` and ultimately, subroutine calls that are halo swaps into instances of `NemoHaloExchage` or similar.
This is in contrast to what we do at the minute which is to have NEMO-specific handlers in the PSyIR frontend that create Nemo-specific objects in response to certain elements in the fparser2 parse tree.
As an aside: a domain-specific PSyIR could be DPSyir, "d-sire" :-)

sergisiso:
I will copy here part of what I said at #596.

I think the bit that I wrongly placed into the generic front-end reader is that `generate_schedule` generates a KernelSchedule. I think there should be a more generic concept of Subroutine (or similar) for and when this generate_subroutine is called we know what are we expecting (we don't need a complex pattern matching pass) and we can to the necessary preparations to generate the kernelSchedule, InvokeSchedule or Algorithm.

There is probably some minor renaming of a few classes/methods needed.

@rupertford @arporter Does it make sense?

arporter:
I think what you said above @sergisiso does make sense - I'm looking at this because of #430. Shall I introduce a new `Routine` node? @rupertford, how is a 'standard' Kernel represented in the PSyIR? Obviously we get the subroutine statement and associated declarations when we use the backend there...

arporter:
OK, so I've answered my own question: we create a `Container` for a module and this has `KernelSchedule` nodes as children, each of which represents a routine. I agree with Sergi that this is a bit misleading and it would be better if we had generic `Routine` nodes instead.

rupertford:
Yes, I agree with you both. When we introduced the concept of a call into the PSyIR, the plan was to introduce the concept of a Routine that a call calls and this was discussed in #786 but no issue was added (and we hadn't agreed on a name at the time). PR #787 added a RoutineSymbol which should be used when adding a Routine Node.

arporter:
[...]

PR #1211, opened 2021-04-16
(Towards 1210) Add a NemoKernelTrans to convert generic PSyIR into a Kernel object

arporter:
Since doing this also allowed me to simplify/move code from the outdated locations holding NEMO-related code, I thought it would be best done as a separate PR.


arporter:
This is a small step away from the use of a NEMO-specific processor towards raising generic PSyIR using Transformations. I've used the new functionality in the 'old' location in order to minimise changes. #737 will re-structure the NEMO frontend properly once the various raising Transformations are implemented.

Probably one for @sergisiso or @rupertford assuming that the CI is happy.

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
```diff
@@ -0,0 +1,127 @@
...
+        :param options: a dictionary with options for \
+            transformations. No options are used in this \
+            transformation. This is an optional argument that defaults \
+            to None.
+        :type options: dict of string:values or None
+
+        '''
+        self.validate(sched)
+
+        nemokern = NemoKern(sched.pop_all_children(),
```
sergisiso, line 125: If I am not wrong the second parameter is never used for NemoKern (is always `None`), can it be deleted?

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
```diff
@@ -0,0 +1,127 @@
...
+NEMO Kernel.
+'''
+
+from psyclone.transformations import Transformation, TransformationError
+from psyclone.psyir.nodes import Schedule, Loop, Call, CodeBlock, Assignment
+from psyclone.psyir.backend.fortran import FortranWriter
+from psyclone.nemo import NemoKern
+
+
+class NemoKernelTrans(Transformation):
```
sergisiso, line 47: NemoKernTrans seems a very broad name and it sounds as it is for transforming an existing NemoKern, only when reading the docstring it becomes clear what it does. I would suggest something like RaiseNemoKernTrans or CreateNemoKernTrans.

(there is another issue about the location of this transformation as this probably won't be part of the transformations that the user can do in the script, but I am not yet sure about this and probably will become clearer later on so I don't suggest changing it for now)

sergisiso (changes requested):
@arporter The PR looks good, I have some comments mainly about a more descriptive name and if some maybe-unnecessary code can be deleted.

src/psyclone/nemo.py
```diff
@@ -95,34 +97,6 @@ def _create_loop(self, parent, variable):
 
         return loop
 
-    def _process_loopbody(self, loop_body, node):
```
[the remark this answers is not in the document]
arporter: Oh yes, I vaguely remember thinking about that but then forgot. It can be removed so I have done so.

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
```diff
@@ -0,0 +1,127 @@
...
+
+    '''
+    @property
+    def name(self):
+        '''
+        :returns: the name of the transformation.
+        :rtype: str
+
+        '''
+        return type(self).__name__
```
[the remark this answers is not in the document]
arporter: Yes, it could be but I've resisted doing that as it ought to be its own PR as it will mean fairly widespread changes. I thought we already had a Issue but couldn't find it so have created #1214.

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
arporter: That was a mistake - 'options' should have been passed into `validate()` so that it can potentially be picked up by a base class.

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
arporter: I've gone for `CreateNemoKernelTrans`. I've also added a new section to the Developer guide and referenced this transformation there. I've done the same for a couple of the LFRic transformations but please could you check whether I've got them all?

arporter:
CI permitting, this is ready for another look...

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
sergisiso: Sorry for not being clear here, I meant the second attribute on the NemoKern constructor. Which is declared at nemo.py:357 Using grep I see this is always None, but maybe I am missing some use? I guess the _ast if assigned by the Fortran frontend instead than the constructor?

Also looking then at the NemoKern class it seems the "pylint disable" is no longer needed?

sergisiso:
@arporter I replied to one of the previous comments, can you have a look at that. Other than that the code looks good, there are no pycodestyle or pylint issues that need to be solved, test and examples pass and compile,  and new documentation for the raising transformations has been added. After the remaining issue is addressed this PR should be ready.

src/psyclone/domain/nemo/transformations/nemo_kernel_trans.py
arporter: Ah, no, you were clear - I just misread it! You are right, I'm able to remove the fparser2 AST argument now and have gone through the code and tidied up. Felt good :-)

arporter:
Provided the CI is happy, this is ready for another look now.

sergisiso (approved):
@arporter All looks good now. PR ready to be merged.

Issue #595, opened 2019-11-20
Transformations should not return anything.


arporter:
Currently all Transformations return a tuple of `(Schedule, Memento)`. This has always been a bit odd because we pass a reference to the relevant bit of a Schedule to the transformation and it *transforms* it in place. There is therefore no need to return the Schedule and indeeed, in many tests we ignore the returned object. Worse than that, a transformation always returns the 'root' of the tree to which the nodes it has been passed belong. In the NEMO API this now means that we get back the root `Container` node describing the whole Fortran module. Although I could probably fix this by returning the `InvokeSchedule` instead, it seems to me that it would be better not to return a Schedule object at all. What do @rupertford, @sergisiso, @hiker and @TeranIvy think?

hiker:
I was always wondering why we are getting a new schedule - so I am good with removing this.

arporter:
@rupertford suggests (in reviewing #615) that we could remove the memento object that is also returned since we never use it. What do @sergisiso and @TeranIvy think? Once we have an agreed strategy this issue will morph in order to implement it (if necessary).

TeranIvy:
If we are not using Memento then I vote for removing it. As for Schedule, as long as the transformed schedule is able to be transformed by another transformation in row (I presume that this is given) and  visible via `view` or a similar method I am fine with transformations not returning it as well.

sergisiso:
I see no problem in removing `Memento`, the user can always make a copy in the script if needed (also the memento `deepcopy` is commented out anyways (?)). I am also ok with removing the returned `Schedule` as I agree it would be more clear that the transformations are in-place.

arporter:
Now that that's agreed, I've changed this into a 'proper' issue.

sergisiso:
In #1229 I remove most of the return values in front of the transformation callers: `schedule, _ = transformation.apply(...` -> `transformation.apply(...` , but I have not modified the Transformation interface (still returns something that gets ignored).

After #1229 is merged this issue will be much simpler to solve.

PR #1217, opened 2021-04-21
Add remaining raising transformations for NEMO (towards #1210)

### LINKED ISSUE #1210: Transformation to convert generic PSyIR into NEMO PSy layer
In #1147 I just have a `raise_psyir` function but I know that @sergisiso and @rupertford have agreed that this type of functionality should be a transformation.

arporter:
This PR only adds the transformations and associated tests. Use of the transformations in anger is postponed to #1147. Only once that is done and the two TODOs that refer to #1210 removed can #1210 be closed.


arporter:
Ready for a first look from either @sergisiso or @rupertford, if they'd be so kind.

src/psyclone/domain/nemo/transformations/create_nemo_invoke_schedule_trans.py
```diff
@@ -0,0 +1,159 @@
...
+                                             routine.pop_all_children(),
+                                             is_program=routine.is_program)
+
+        # We need to replace the top node in the (possibly sub-) PSyIR
+        # tree that we've been passsed.
+        if routine.parent:
+            routine.replace_with(new_node)
+        del routine
+
+        return new_node
```
sergisiso, line 155: This is different that the Transformation generic method that returns `root, memento`. I know that we are not consistent on the apply interface (which we should fix at some point), but in this case couldn't we always return the root node as in the generic interface `return new_node.root, None`?

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
```diff
@@ -0,0 +1,148 @@
...
+        self.validate(loop, options=options)
+
+        # Convert a generic loop into a NEMO Loop by creating a new
+        # NemoLoop object and inserting it into the PSyIR.
+        nodes = loop.pop_all_children()
+        new_loop = NemoLoop.create(loop.variable,
+                                   nodes[0], nodes[1], nodes[2],
+                                   nodes[3].pop_all_children())
+        loop.replace_with(new_loop)
+        del loop
```
sergisiso, line 143: I have the same question here about if the `del` is necessary and what is the expectation if there is an already existing reference to the original node?

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
```diff
@@ -0,0 +1,148 @@
...
+        return new_loop
```
sergisiso, line 144: I see at the moment the `return new_loop` is necessary for the CreateNemoPSyTrans apply:
```
        for loop in root.walk(Loop):
            new_loop = loop_trans.apply(loop)

            try:
                kern_trans.apply(new_loop.loop_body)
            except TransformationError:
                # Not all loop bodies are valid kernels (e.g. if they do IO)
                pass
```
but again this is different that the typical generic Transformation interface.

Would maybe this be more natural if we brought the try/except block inside here. So CreateNemoPSyTrans calls CreateNemoLoopTrans which callscalls CreateNemoKernelTrans?

Then this transformation can return `new_loop.root, None` or simply nothing, as the other transformations usually do.

src/psyclone/domain/nemo/transformations/create_nemo_psy_trans.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Deal with Routines first. Have to take care of the case where the
+        # supplied top-level node is itself a Routine and must therefore be
+        # replaced.
+        root = psyir
+
+        for routine in psyir.walk(Routine):
+            new_node = invoke_trans.apply(routine)
+            if routine is root:
+                root = new_node
```
sergisiso, line 157: I think with the proposed modification this may not be needed anymore as the transformation always would return root? This could be:
```
for routine in psyir.walk(Routine):
    root = invoke_trans.apply(routine)
```

src/psyclone/nemo.py
```diff
@@ -58,15 +58,15 @@ class NemoFparser2Reader(Fparser2Reader):
...
         Create an empty InvokeSchedule. The un-named argument would be 'name'
         but this isn't used in the NEMO API.
 
         :returns: New InvokeSchedule empty object.
         :rtype: py:class:`psyclone.nemo.NemoInvokeSchedule`
         '''
-        return NemoInvokeSchedule()
+        return NemoInvokeSchedule(name)
 
     def _create_loop(self, parent, variable):
```
sergisiso, line 71: I understand this can't be removed yet because the raising transformations are created but not used in the nemo path yet, but eventually it will replace this code. Is this right?

Looking at this method, is the assignment of a loop.loop_type given the variable name loop_type_mapping something that the CreateNemoLoopTrans should be doing? Or is it left out on purpose?

sergisiso (changes requested):
@arporter A made some comments about the new Transformations, mainly regarding a clarification about the `del` use and the transformations return value which is different than the generic Transformation.

doc/developer_guide/transformations.rst
```diff
@@ -90,12 +90,20 @@ domain-specific knowledge, this language level PSyIR must be 'raised'
 to a domain-specific PSyIR. The resulting PSyIR will then contain
 nodes representing higher-level concepts such as kernels or halo
 exchanges. This raising is performed by means of the transformations
-listed in this section.
+listed in the sections below.
```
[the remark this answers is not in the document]
arporter: Yes, fair point.

src/psyclone/domain/nemo/transformations/create_nemo_invoke_schedule_trans.py
```diff
@@ -0,0 +1,159 @@
...
+    NemoInvokeSchedule.
+
+    """
+    @property
+    def name(self):
+        '''
+        :returns: the name of the transformation.
+        :rtype: str
+
+        '''
```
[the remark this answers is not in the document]
arporter: Yes, done.

src/psyclone/domain/nemo/transformations/create_nemo_invoke_schedule_trans.py
```diff
@@ -0,0 +1,159 @@
...
+        new_node = NemoInvokeSchedule.create(routine.name,
+                                             routine.symbol_table,
+                                             routine.pop_all_children(),
+                                             is_program=routine.is_program)
+
+        # We need to replace the top node in the (possibly sub-) PSyIR
+        # tree that we've been passsed.
+        if routine.parent:
+            routine.replace_with(new_node)
+        del routine
```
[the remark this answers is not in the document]
arporter: I added this as an afterthought thinking that maybe we should clean up thoroughly but I think your concerns are justified. I've removed it.

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
arporter: Yes (or no), removed.

src/psyclone/domain/nemo/transformations/create_nemo_invoke_schedule_trans.py
arporter: We have issue #595 for this. There I argued that transformations shouldn't return anything but given that this particular transformation might change the root node of the whole tree I'm not sure we can do that anymore. Having said that, I've seen mention of a `FileContainer` node so maybe we can argue that a `Routine` will never be the root of the whole tree? If we do that though then a lot of the NEMO tests will need updating as they often work from a string that just contains a program or subroutine definition.

Anyway, for now I've followed you suggestion and it now returns the suggested tuple.

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
arporter: I was uncomfortable about making the loop trans into a meta-transformation that also did the kernels. Therefore, I've re-ordered the transformations in this method so that kernels are done first, then loops, then routines. That way it doesn't matter if nodes in the tree are replaced.

src/psyclone/domain/nemo/transformations/create_nemo_psy_trans.py
arporter: Yes, it's simpler now.

src/psyclone/nemo.py
arporter: Yes, that's right - it will be removed. Your second question made me think but in fact this assignment is now done in the `NemoLoop.create()` method. I can't use that here though because it requires that all the children of the loop be supplied.

arporter:

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
```diff
@@ -0,0 +1,153 @@
...
+        if not isinstance(node, Loop):
+            raise TransformationError(
+                "Error in CreateNemoLoopTrans transformation. The supplied "
+                "node should be a PSyIR Loop but found '{0}'".format(
+                    type(node).__name__))
+
+    def apply(self, loop, options=None):
+        '''
+        Takes a generic PSyIR Loop node and replaces it with a NemoLoop.
+        The original loop object is deleted.
```
sergisiso, line 122: Delete this line now?

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
```diff
@@ -0,0 +1,153 @@
...
+
+        :param sched: the Loop node to be transformed.
```
sergisiso, line 124: s/sched/loop/ also line below

src/psyclone/tests/domain/nemo/transformations/create_nemo_psy_trans_test.py
```diff
@@ -0,0 +1,141 @@
...
+from psyclone.transformations import Transformation, TransformationError
+from psyclone.domain.nemo.transformations import CreateNemoPSyTrans
+from psyclone.nemo import NemoInvokeSchedule, NemoKern, NemoLoop
+
+
+@pytest.fixture(scope="session")
+def psy_trans():
+    ''' pytest fixture that creates an instance of the Transformation that
+    this module is testing. '''
+    return CreateNemoPSyTrans()
```
sergisiso, line 52: Is there any benefit of having this one-line constructor as a fixture? If you still prefer a fixture can you add a top level pylint disable redefined-outer-name statement as it is giving this warning in each test method.

sergisiso (changes requested):
@arporter My main concerns are about the transformation interface have been addressed. 

There is still some open questions about what happens when replacing the root which are more general than this PR and don't need to be solved now. An alternative would be to use a "specialise" method similar to what we do with symbols that would avoid creating a different object while still raising the class to a higher level concept. But since we don't have this feature now, I am fine with the current implementation.

I checked all code and tests, built developers and reference documentations without issues and check pycodestyle and pylint. The PR is almost there, I just commented inline in some last typos and pylint issues I found.

Also I understand this is "towards #1210" for now, since there are still TODOs for this issue.

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
arporter: Oops.

src/psyclone/domain/nemo/transformations/create_nemo_loop_trans.py
arporter: Oops.

src/psyclone/tests/domain/nemo/transformations/create_nemo_psy_trans_test.py
arporter: Well, it does save one line per test. I've renamed it to avoid the pylint warnings.

arporter:
I see what you mean about the TODOs and I think it's right to leave them alone for this PR. I'll therefore change the title of this PR and then this is ready for review again (subject to the CI).

sergisiso (approved):
@arporter All comments have been addressed, the PR is ready.

I notice that the tests introduced in this PR could be updated to use the new Fortran Frontend fixture that is queued to be merged into master. Since I am responsible of this merge conflict I will wait until the new fixture is on master and fix it myself while updating the changelog.

arporter:
#1147 is removing the one remaining TODO for this issue so closing it now.

PR #1219, opened 2021-04-21
remove gen_code abstract method in Node base class (closes #1218)

### LINKED ISSUE #1218: Remove abstract gen_code method from Node base class
All new PSyIR nodes that are being created rely on the PSyIR backend writers to output code so do not implement the gen_code method. As this is declared as abstract in the Node baseclass this results in pylint warnings. This abstract method can now be removed as it is no longer a requirement for new nodes.

rupertford:


rupertford:
Very small PR that reduces the number of pylint warnings in PSyIR nodes.

rupertford:
I've suggested either @arporter or @sergisiso as they are both working on PSyIR.

arporter (approved):
Changes are good. Docs all still build OK.
Examples and tests are OK with compilation.
pylint complained that abc was imported but unused in node.py but I've fixed that.
Will update changelog and UG and proceed to merge.

Issue #777, opened 2020-05-15
[PSyData] Improved design and testing for class-prefix


hiker:
Atm you have to specify a non-default class-prefix in the apply call of a PSyData-based transformation, e.g.:
```
    data_trans = PSyDataTrans()
...
    data_trans.apply(schedule1, {"prefix": "non-default"})
...
    data_trans.apply(schedule2, {"prefix": "non-default"})
```

I think it would be more convenient if the prefix could be specified in the constructor (i.e. the same instance of a transformation would automatically create the same prefix when applied):
```
    my_data_trans = PSyDataTrans({"prefix": "non-default"})
...
    my_data_trans.apply(schedule1)
...
    my_data_trans.apply(schedule2)
```
Additionally, atm a node based on `PSyDataNode` can specify a default class-prefix that is invalid (i.e. not defined in the psyclone.cfg file) - since the test for validity is done in the `validate()` function when applying a transformation. I will add a second test for a valid prefix during `gen_code()` to catch that problem.

@arporter (since you reviewed the corresponding pull request, but anyone else feel free to comment as well) 
1. do you agree with specifying the class-prefix in the constructor?
2. Should we additionally (still) allow to overwrite the prefix in the apply function? If not, I would suggest to raise an exception if 'prefix' is specified in apply (as a way for #613 - while this would not really be a typo, but incorrect use and as such should be flagged).

PS: The reason I suggest to add an option dictionary argument to the transformation constructor instead of a normal python function argument is that long term we might want to support more default values in the constructor (e.g. the module-name could be the same for all `apply()`'s since one transformation typically will be instanced for one module, so we would only have to specify the region-name in the apply. But this is a bigger restructuring that I won't do as part of this ticket).

arporter:
Yes, I agree. It makes sense to only have to supply the information once and I think removing it from the `apply` method is also sensible. 

The use of a dictionary sounds very like the suggestion @sergisiso made on Teams to use `**kwargs` in sub-classes. We could change the constructor to use this instead perhaps?

hiker:
Thanks, Andy.

Well, the dictionary suggestion is what we are using for the apply/validate functions already, so to a certain degree that would break the consistency there. And I like the fact that using `options={'bla':...}` kind of documents that these options are used in the 'bowels of the transformation' somewhere (and ideally the option dictionary could become an object, which would allow us to track which options are used, so we could flag any unused option, see #613, but that would be quite a bit of change).

[...]

Issue #435, opened 2019-07-12
Use the PSyIR Fortran backend for the NEMO API


arporter:
Currently the NEMO API works by manipulating the fparser2 parse tree and then using the de-parsing functionality of that tree to produce the final Fortran code. Although this has enabled us to get up and running fairly quickly it has limitations. Once the Fortran backend to the PSyIR is sufficiently complete we will migrate the NEMO API to use that.

arporter:
This Issue will also fix #396.

arporter:
In #500 I've done some experimentation by adding the following to `NemoPSy.gen()`:

        from psyclone.psyir.backend.fortran import FortranWriter
        fvisitor = FortranWriter()
        result = fvisitor(self._invokes._container)
        return result

arporter:
In sbcblk.f90 we get a CodeBlock within a WHERE/ELSEWHERE block which results in the script attempting to put KERNELS around the individual statements within the WHERE. This is wrong because they are *not* implicit loops.

`CodeBlock[[<class 'fparser.two.Fortran2003.Masked_Elsewhere_Stmt'>]]` comes from:

        ELSEWHERE(phs(:, :, :) <= 0._wp .AND. phi(:, :, :) >= 0.1_wp)

arporter:
If I do what I said earlier (and allow for cases with a missing Container) then `test_api_no_alg` shows that we get the body of the routine OK but we have no surrounding Subroutine or Module declaration. Probably the problem is that we have no `Container`.

arporter:
We have no Container because this particular test is for a Program which then of course has no enclosing module. We generate an Invoke for the Program and this Invoke has a Schedule which is what enables us to generate the body of the routine. (That Schedule also has an associated symbol table.) However, we have no way of generating the Program declaration.

arporter:
I now realise this is precisely the problem that #430 is addressing!

PR #1224, opened 2021-04-26
(closes #1185) refactor PSyDataNode constructor

### LINKED ISSUE #1185: Re-factor PSyDataNode constructor
In #1177 it was noted that the constructor for the PSyDataNode no longer follows the pattern that has been adopted for other PSyIR nodes (where we have a 'thin' constructor and a static 'create' method that does most of the work). It's also possible that some of what it does should be performed by the associated transformation.

arporter:


arporter:
Assuming the CI is happy this is ready for review. It brings the PSyDataNode into line with the way other Nodes are implemented by slimming-down the constructor and moving functionality to a `create` method. This then permits checks for any name clashes with existing symbols to be done in the Transformation rather than being delayed until we `update` the parse tree (in the NEMO API). That change means that those checks that are done in the `update()` method are no longer covered by the test suite but I argue that that's OK as the `update()` method is very shortly going to be removed (#435).

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -216,7 +162,111 @@ def __init__(self, ast=None, children=None, parent=None, options=None):
...
+            # The tag doesn't exist which means that we haven't already added
+            # this Container as part of a PSyData transformation.
+            csym = ContainerSymbol(fortran_module)
+            symbol_table.add(csym, tag=fortran_module)
+
+        # A PSyData node always contains a Schedule
+        # TODO 435 we can probably get rid of _insert_schedule() once we
+        # do away with the ast argument?
+        # pylint: disable=protected-access
+        sched = data_node._insert_schedule(children=children, ast=ast)
```
sergisiso, line 237: I am not sure if this is what 435 already refers to but I am wondering if instead of "inserting" a Schedule, we could make PSyData sub-class Schedule. But since this PR already change enough things it is just an idea for now.

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -216,7 +162,111 @@ def __init__(self, ast=None, children=None, parent=None, options=None):
...
+        data_node.addchild(sched)
+
+        # The use statement that will be inserted by the update() method. Any
+        # use of a module of the same name that doesn't match this will result
+        # in a NotImplementedError at code-generation time.
+        # TODO #435 remove this when removing the update() method
+        names = [data_node.add_psydata_class_prefix(symbol.name) for symbol in
+                 PSyDataNode.symbols]
```
sergisiso, line 245: s/PSyDataNode/self/ ? Or we explicitly don't want to allow subclasses to redefine this?

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -216,7 +162,111 @@ def __init__(self, ast=None, children=None, parent=None, options=None):
...
+        data_node.use_stmt = "use {0}, only: "\
+            .format(fortran_module) + ", ".join(names)
+        # Add the symbols that will be imported from the module. Use the
+        # PSyData names as tags to ensure we don't attempt to add them more
+        # than once if multiple transformations are applied.
+        for sym in PSyDataNode.symbols:
```
sergisiso, line 251: s/PSyDataNode/self/ ?

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -91,95 +91,42 @@ class PSyDataNode(Statement):
     # PSyData interface Fortran module
     fortran_module = "psy_data_mod"
     # The symbols we import from the PSyData Fortran module
-    symbols = ["PSyDataType"]
+    _PSyDataSymbol = namedtuple("_PSyDataSymbol", "name symbol_type")
+    symbols = [_PSyDataSymbol("PSyDataType", TypeSymbol)]
```
sergisiso, line 95: Do we need this to be class attributes instead of instance attributes? For instance if we have 2 profiling regions with 2 different ProfileNode classes?

Also why do we need the nametuple, can we just have: `self._symbols = [TypeSymbol("PSyDataType")]`
and use `self._symbols[0].name` in the gen_code()?

Finally, we only ever use this symbol, is this going to change in the future?, otherwise it could be:
`self._typesymbol = TypeSymbol("PSyDataType")`
or even better `self._typesymbol = TypeSymbol(self._text_name + "Type")` but this will require changes in the tests.

To be honest I can't see clearly where this is needed, so I don't know exactly what is the best option.

src/psyclone/psyir/nodes/read_only_verify_node.py
```diff
@@ -74,9 +70,7 @@ def __init__(self, ast=None, children=None, parent=None, options=None):
         # If there is no value passed to the constructor, default
         # to the "read_only_verify" prefix.
         my_options["prefix"] = my_options.get("prefix", "read_only_verify")
-        super(ReadOnlyVerifyNode, self).__init__(ast=ast, children=children,
-                                                 parent=parent,
-                                                 options=my_options)
+        super(ReadOnlyVerifyNode, self).__init__(ast=ast, options=my_options)
         self._text_name = "ReadOnlyVerify"
         self._colour = "green"
```
sergisiso, line 75: Shouldn't this be class attributes? Also see other comments on others constructors as the rest of this method could be deleted if the default_prefix is set on the generic constructor.

sergisiso (changes requested):
@arporter Thanks for tacking this issue, I think is a step on the right direction.  I added a few comments regarding the constructors. Hopefully when 435 is done and the TODOs are resolved this will be further simplified.

src/psyclone/domain/gocean/nodes/gocean_extract_node.py
```diff
@@ -55,10 +55,6 @@ class GOceanExtractNode(ExtractNode):
     :param ast: reference into the fparser2 parse tree corresponding to \
         this node.
     :type ast: sub-class of :py:class:`fparser.two.Fortran2003.Base`
-    :param children: the PSyIR nodes that are children of this node.
```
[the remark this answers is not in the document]
arporter: Oops. I hadn't realised that they are args to the `Node` base class. I've reverted this change (and those in the other classes).

src/psyclone/psyir/nodes/nan_test_node.py
```diff
@@ -73,16 +68,15 @@ class NanTestNode(PSyDataNode):
...
 
-    def __init__(self, ast=None, children=None, parent=None, options=None):
+    def __init__(self, ast=None, options=None):
         if options:
             my_options = options.copy()
         else:
             my_options = {}
         # If there is no value specified in the constructor, default
         # to the "nan_test" class.
         my_options["prefix"] = my_options.get("prefix", "nan_test")
```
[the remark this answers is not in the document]
arporter: You're right, things are tidying up nicely now.

src/psyclone/psyir/nodes/extract_node.py
```diff
@@ -79,16 +75,15 @@ class ExtractNode(PSyDataNode):
...
-            my_options = {}
+    def __init__(self, ast=None, options=None):
+        if not options:
+            options = {}
         # If there is no value specified by in the constructor, default
         # to the "extract" class.
-        my_options["prefix"] = my_options.get("prefix", "extract")
-        super(ExtractNode, self).__init__(ast=ast, children=children,
-                                          parent=parent, options=my_options)
+        my_options = options.copy()
```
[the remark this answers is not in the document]
arporter: Gone now.

src/psyclone/psyir/nodes/profile_node.py
```diff
@@ -73,7 +67,7 @@ class ProfileNode(PSyDataNode):
     _text_name = "Profile"
     _colour = "green"
 
-    def __init__(self, ast=None, children=None, parent=None, options=None):
```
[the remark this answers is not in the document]
arporter: Constructor removed.

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -68,15 +74,9 @@ class PSyDataNode(Statement):
...
-    :param children: a list of child nodes for this node. These will be made \
-                     children of the child Schedule of this PSyDataNode.
-    :type children: list of :py:class:`psyclone.psyir.nodes.Node` \
-                    or derived classes
-    :param parent: the parent of this node in the PSyIR.
-    :type parent: :py:class:`psyclone.psyir.nodes.Node`
     :param options: a dictionary with options for transformations.
     :type options: dictionary of string:values or None
-    :param str options[prefix"]: a prefix to use for the PSyData module name \
+    :param str options["prefix"]: a prefix to use for the PSyData module name \
```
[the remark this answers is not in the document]
arporter: I am inclined to agree - I would prefer to keep the `options` dict for the `apply()` method of transformations but pass specific, named arguments to the constructor. However, this is already the subject of #777 so I've put comments there and won't modify it in this PR.

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -41,8 +41,14 @@
 or profiling.'''
 
 from __future__ import absolute_import, print_function
+from collections import namedtuple
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -216,7 +162,111 @@ def __init__(self, ast=None, children=None, parent=None, options=None):
...
+                "must be an instance of psyir.symbols.SymbolTable but got "
+                "'{0}'.".format(type(symbol_table).__name__))
+
+        data_node = cls(ast=ast, options=options)
+
+        # Ensure that we have a container symbol for the API access
+        fortran_module = data_node.add_psydata_class_prefix(
+            data_node.fortran_module)
+        try:
+            csym = symbol_table.lookup_with_tag(fortran_module)
```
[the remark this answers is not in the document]
arporter: It is intentional and I *think* it's right. The reasoning is that we need to know whether we've already inserted symbols related to the PSyData API (e.g. if we add more than one profiling region to a routine) or whether we have a clash with symbols that pre-existed in the source. By giving tags to those symbols inserted via the PSyData API this becomes possible.

src/psyclone/psyir/nodes/psy_data_node.py
arporter: No, #435 is really about getting rid of the need to carry around the fparser2 AST (i.e. we use the PSyIR everywhere). I like the idea of making PSyData sub-class Schedule although does that fit with the fact that, when it is lowered, it corresponds to several 'opening' and 'closing' statements around its children? Also, if the entire contents of e.g. a routine are enclosed in a PSyData region then we will have a Schedule with another (sub-class of) Schedule as its child.

src/psyclone/psyir/nodes/psy_data_node.py
arporter: I wasn't imagining that they would but of course they could. Changed.

src/psyclone/psyir/nodes/psy_data_node.py
arporter: Done.

src/psyclone/psyir/nodes/read_only_verify_node.py
arporter: Yes, good spot. The constructor is gone now.

src/psyclone/psyir/nodes/psy_data_node.py
arporter: I've re-worked this a fair bit. I think you were right about needing instance attributes so I've added those and stored the full names (ie with prefixes added) in them. I've also added a test that we can have two different prefixes in the same invoke.
We can't currently work with `Symbols` because this construction of the names happens in the PSyDataNode constructor which doesn't have any access to a symbol table. (This means we can't solve any name clashes for symbols imported from the PSyData Fortran module (which we could do in principle). If there's a clash with the name of the PSyData Fortran module itself then I don't think there's anything we can do anyway.) I went for the namedtuple because I always think that trying to understand code that just does e.g. `symbol[0]` is harder than `symbol.name`. It is slightly overkill in this situation but I don't think it hurts.

arporter:
For some reason, under Python 3.5 the ordering of the USE statements can vary from run to run. I've simply relaxed the affected test to allow for this. Hopefully the CI will be happy now and then this is ready for another look.

src/psyclone/psyir/nodes/psy_data_node.py
sergisiso: > We can't currently work with Symbols because this construction of the names happens in the PSyDataNode constructor which doesn't have any access to a symbol table.

I see, that makes things complicated but I think its fine for now, al least while we have the gen_code and update.

>  I went for the namedtuple because I always think that trying to understand code that just does e.g. symbol[0] is harder than symbol.name. It is slightly overkill in this situation but I don't think it hurts.

I think that is a good reason to use namedtuple, makes for better code.

One issue remaining is that we don't use its contents as much as possible. Just in the same file there is the "PSyDataType" hardcoded in 4 different places, which would make changing the attribute break things I believe. Why is this a list? Do we expect more things to be in this list? And why is _psydata_fortran_module a special case and not in the list?

Alternatively, since the type is also mandatory for all PSyData classes, why don't we put the type in a _psydata_fortran_type equivalent to the module so then when we need its string it can be unequivocally referenced? Then we can get rid of the _psydata_symbols unless there is a use case for it. Also then we don't need to say is a TypeSymbol here (as we don't say the fortran_module is a ContainerSymbol) as this is known at the point of use.

Also depending on my other comment on the prefix. If this turns out to be static why not also construct these names statically, e.g. _psydata_fortran_module = _prefix + "psy_data_mod"

sergisiso (changes requested):
@arporter The clean up of the sub-classes looks very nice. I have a few more comments about how we construct the generic PSyData class.

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -87,99 +93,72 @@ class PSyDataNode(Statement):
...
+        if prefix and prefix not in Config.get().valid_psy_data_prefixes:
+            raise InternalError(
+                "Invalid 'prefix' parameter: found '{0}', expected one of {1} "
+                "as defined in {2}".format(
+                    prefix, Config.get().valid_psy_data_prefixes,
+                    Config.get().filename))
+        if not prefix:
+            self._class_string = ""
+        else:
+            self._class_string = prefix + "_"
```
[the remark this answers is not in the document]
arporter: I realise your question also means that we could better explain what is going on in the code so I've added a comment with a link to the relevant part of the UG.

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -87,99 +93,72 @@ class PSyDataNode(Statement):
...
-            self._psy_data_symbol_with_prefix, symbol_type=DataSymbol,
-            datatype=psydata_type).name
-
-        # Name of the region. In general at constructor time we might
-        # not have a parent subroutine or any child nodes, so
-        # the name is left empty, unless explicitly provided by the
-        # user. If names are not provided here then the region and
-        # module names are set the first time gen() is called (and
-        # then remain unchanged).
+        self._region_identifier = ("", "")
```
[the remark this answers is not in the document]
arporter: At first I thought you were right so I removed _region_identifier. However, that broke a test so then I read the comment properly and realised we cannot remove _region_identifier so I've put it back. My understanding is that, if no explicit region name is provided by the user, we generate a region name. We need access to this while generating the code and potentially while generating driver code (for the case of kernel extraction). However, we *don't* want this generated name to be used if a `gen` is performed a second time as the tree might have changed and there may now be other PSyData nodes, meaning the generated name must be changed. Possibly it would be better to cope with this situation by having a separate attribute that records whether the name of the node was originally specified by the user or auto-generated. However, I think (hope!) that's something for another PR and possibly a discussion with @hiker as there may be subtleties I've missed.

arporter:
Thanks for that review Sergi, the code gets better each time :-) CI permitting, this is ready for another look now.

sergisiso (approved):
@arporter All review suggestion have been addressed. There are no hanging TODOs of this PR. No pycodestyle or significant pylint issues introduced. New code is tested and it passes with compilation. Since this is a refactoring but the API/parameters are essentially the same, the documentation for PSyData still looks appropriate and no modifications are needed. A few classes have been added to the reference guide, these build without issues. This PR is ready to be merged.

Issue #1151, opened 2021-03-09
PSyclone release 2.0.0


TeranIvy:
This issue gathers the requirements for the new PSyclone release 2.0.0 and tracks the progress towards creating the release. 

This PSyclone relase is tied to the [fparser release 0.0.12](https://github.com/stfc/fparser/issues/296).

The required issues/pull requests are listed below.

**LFRic**

- PR #1145 for issue #1107 (add required built-ins) **Done**
- PR #1131 for issue #1116 (make the number of `ANY_*_SPACE` function space IDs and the supported Fortran datatypes configurable) **Done**
- PR #1126 for issue #1098 (update kernel stub generator for CROSS2D stencils) **Done**
- PR #1082 towards #925 (add code-generation for `domain` kernels) **Done**
- PR #1205 for issue #870 (remove support for `iterates_over` kernel metadata) **Done**
- PR #1156 for issue #874 (remove support for the "old style" `arg_type` metadata) **Done**
- PR #1072 for issue #1067 (restructure LFRic examples) - this would mean all examples are consistent, and it is a really small change otherwise. **Done**
- PR #1186 for issue #1173 (reshaping of LFRic intergrid mesh maps). **Done**
- PR #1167 for issue #1097 (install wrapper libraries). **Done**

**GOcean**
- ??

**NEMO**
- In theory improved loop fusion would be useful, but depending on schedule for 2.0 I don't think I can finish this in time. Adding 'some' tests (same loop iterations and loop variable) to NEMO would be ready, but likely only of limited use, am happy to wait.

**PSyIR**
- ??

arporter:
The bug fix for fparser has been merged by Rupert Ford and I have `1151_update_fparser` where I've brought the fparser submodule up to master so that we can test. I'll test with NEMO. @TeranIvy would you mind testing this branch with LFRic?

arporter:
Once that testing is complete I think we are OK to go ahead and release fparser @rupertford? Once that's done we need to update the PSyclone setup.py to refer to the new version and also alter the GHA config so that the test suite uses the released version of fparser.

arporter:
I can confirm that I can process the NEMO code base with this branch without any parsing problems.

TeranIvy:
I can confirm that the full LFRic test suite works fine on Spice and Crays with the PSyclone environment built from `1151_update_fparser`.

TeranIvy:
#1226 is merged.

arporter:
[...]

PR #1226, opened 2021-04-27
(towards #1151) update fparser to 0.0.12

arporter:


arporter:
This is ready for first review from either @TeranIvy or @rupertford.

arporter:
I've just noticed a typo in our README so I've taken the opportunity to fix it.

.github/workflows/python-package.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         # If you wish to install the version of fparser pointed to by the
         # submodule instead of the released version (from pypi) then
```
TeranIvy, line 79: I would suggest "PyPI".

.github/workflows/python-package.yml
```diff
@@ -78,7 +78,7 @@ jobs:
...
         # uncomment the following line:
-        pip install external/fparser
+        # pip install external/fparser
```
TeranIvy, line 81: I assume this is commented as it points to a wrong path?

TeranIvy (changes requested):
Looks good, I only have a suggestion and a question (see inline), after that we are ready to go. Could you also bring the branch up to date with the master?

- Tests with compilation pass.
- Changes are `pycodestyle` and `pylint` clean and covered.
- Documentation builds fine.
- Examples run fine.
- LFRic full test suite on Spice and Cray XC40 is all fine with PSyclone built from this branch.

.github/workflows/python-package.yml
arporter: Done.

.github/workflows/python-package.yml
arporter: Nope - this line is needed only when the fparser submodule included with PSyclone needs to be installed. When we're working with a released version of fparser (as we must do when creating a release of PSyclone) then it will be installed automatically as part of the pip installation of PSyclone itself.

arporter:
I've brought the branch up-to-date. Ready for another look once the CI has run.

.github/workflows/python-package.yml
TeranIvy: Thanks for explaining!

TeranIvy (approved):
All well now. Retested everything except for the LFRic runs (there are no changes that would trip them up).

Updated changelog and user guide. CI permitting, we are ready to go.

PR #1228, opened 2021-04-27
Make gocean api work with PSyIR version of the algorithm layer (towards #753)

rupertford:


rupertford:
The code in this PR can't become the default until #1230 is addressed as the nemolite2d example can't be parsed without restructuring the code.

rupertford:
This PR is now ready for review. However I am holding it back until issue #1230 has been addressed as without that the nemolite2d benchmark would no longer work if this PR went onto master.

rupertford:
The nemolite_2d serial PSyclone version builds OK with this version of PSyclone i.e. the PSyIR algorithm generation works OK. The shallow serial PSyclone version currently fails due to extra quotes being added. I think #1265 will fix this problem so I will hold off requesting a review on this PR until this is on master.

rupertford:
Shallow seems to be OK now, so I'm going to put this out for review. I think it needs to be @sergisiso given that it affects nemolite2d and shallow. There is a risk that we might be breaking other codes e.g. MOST but I think we are at the stage where we can fix any other issues that arise as new PRs (as long as the reviewer agrees of course).

rupertford:
Most of the changes here are to do with adding support for an invoke description back into the base class as the gocean1.0 api makes use of it and I incorrectly thought only the lfric api made use of it. Doing this results in quite a bit of movement of functionality between parent and child classes. Apart from that it is mostly just a case of changing `generate` to use PSyIR for gocean1.0.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -132,43 +171,48 @@ def _def_routine_root_name(self):
...
-        routine_root_name = "invoke_{0}".format(self._index)
-        if len(self.children) == 1:
-            # Add the name of the kernel if there is only one call
-            routine_root_name += "_" + self.children[0].name
+        if self._description:
+            routine_root_name = self._description.lower().strip()
+            if routine_root_name[0] == '"' and routine_root_name[-1] == '"' \
+               or \
+               routine_root_name[0] == "'" and routine_root_name[-1] == "'":
+                # fparser2 (issue #295) currently includes quotes as
```
sergisiso, line 179: Thinking about this it may be a more general issues than the fparser double quotes. The character set of a Python string (UTF-8) can contain many characters that are not valid in a variable name. Maybe we need something like: `routine_root_name = routine_root_name.encode("ascii", "ignore").decode("utf-8")` (although this may be python 3 specific)

Also this may be an issue that needs to be solved in the Symbol or SymbolTable classes for any possible case.

sergisiso (changes requested):
@rupertford I made a first review of the implementation, its another good step towards 753, I added some in-line comments.

I also tried to build NemoLite2D (specifically the OpenMP version) and I am getting the following syntax error in alg.90 in the the line `call timer_start(itimer0, label = 'Time-stepping'num_repeats = nrepeat)`. Could you have a look at this issue.

arporter:
I've got #1311 that will fix the problem with the missing commas in the argument list.

src/psyclone/generator.py
```diff
@@ -229,12 +236,34 @@ def generate(filename, api="", kernel_path="", script_name=None,
...
+            if api == "gocean1.0":
+                # Create language-level PSyIR from fparser2 ast
+                psyir_reader = Fparser2Reader()
+                psyir = psyir_reader.generate_psyir(ast)
+
+                # Raise to Algorithm PSyIR
+                alg_trans = AlgTrans()
+                alg_trans.apply(psyir)
+
+                # issue #753
```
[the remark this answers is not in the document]
rupertford: Added

src/psyclone/generator.py
```diff
@@ -229,12 +236,34 @@ def generate(filename, api="", kernel_path="", script_name=None,
...
+                # 1) call to algorithm optimisation script will go here
+                # 2) psygen creation and symbol generation will go here
+                #    (+ algorithm processing could go here, which would
+                #    include the removal of the invoke symbol).
+                # 3) psygen optimisation script will go here
+
+                # Process and lower to language-level PSyIR
```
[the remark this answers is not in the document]
rupertford: It could and has.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -132,43 +171,48 @@ def _def_routine_root_name(self):
         :rtype: str
 
         '''
-        routine_root_name = "invoke_{0}".format(self._index)
-        if len(self.children) == 1:
-            # Add the name of the kernel if there is only one call
-            routine_root_name += "_" + self.children[0].name
+        if self._description:
```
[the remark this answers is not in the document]
rupertford: I changed it to `name`, as that is what it is called in the invoke call. Hopefully that is OK.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -189,10 +233,21 @@ def lower_to_language_level(self):
...
+            container_symbol = symbol_table.new_symbol(
+                root_name=container_tag, tag=container_tag,
+                symbol_type=ContainerSymbol)
+
+        routine_tag = self.psylayer_routine_root_name
+        interface = GlobalInterface(container_symbol)
+        routine_symbol = symbol_table.new_symbol(
+            root_name=routine_tag, tag=routine_tag, symbol_type=RoutineSymbol,
+            interface=interface)
+
```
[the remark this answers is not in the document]
rupertford: You are correct, but this was the case before this work as well (or at least we would produce incorrect code previously). The issue is that we are not fully there at the moment. The PSy-layer is still receiving information in the old way (a hierarchy called `invoke_info` that is passed into `GOPSy()`) but we have updated the way the algorithm layer is generated (using PSyIR). The step after this one will be to pass the algorithm psyir into the psy-layer instead of `invoke_info` (or probably as well as, in the first instance) and then the psy-layer will be able to pick up the generated names from the algorithm psyir, rather than using `invoke_info`. This is mentioned in #753 but probably not prominently enough.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -77,14 +84,26 @@ def __init__(self, invoke_routine_symbol, index, parent=None):
...
+            raise TypeError(
+                "AlgorithmInvokeCall description argument should be a str but "
+                "found '{0}'.".format(type(description).__name__))
+
+        if description:
+            # Avoid unicode issues with Python2
+            description = str(description)
         self._index = index
-        self.psylayer_routine_symbol = None
+        self.psylayer_routine_root_name = None
```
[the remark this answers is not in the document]
rupertford: I've added a comment. It should become clearer in the next associated PR, assuming it all works (but some restructuring may still be required). The idea is that the PSy-layer will get access to the root names (tags) before the algorithm-layer psyir is lowered, then once the algorithm-layer psyir is lowered the psy-layer psyir can use these tags to extract the actual names from the lowered algorithm-layer psyir. A (or perhaps the) problem is that the lowered algorithm-layer psyir loses knowledge about the root (tag) names. I guess we could potentially lower to a specialised class (sub-class of call), which kept this additional information, but I'd need to think about that.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: That is a good point. The old version does a check so I've replicated that.

rupertford:
> I also tried to build NemoLite2D (specifically the OpenMP version) and I am getting the following syntax error in alg.90 in the the line `call timer_start(itimer0, label = 'Time-stepping'num_repeats = nrepeat)`. Could you have a look at this issue.

This seems to be working. I'm guessing that a separate PR has fixed the problem.

rupertford:

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -128,54 +146,86 @@ def _validate_child(position, child):
...
+        an additional name argument.
+
+        :param bool colour: whether or not to include colour control \
+            codes. Optional argument that defaults to True.
+
+        :returns: description of this PSyIR node.
+        :rtype: str
+
+        '''
+        return f"{self.coloured_name(colour)}[name=\"{self._name}\"]"
```
sergisiso, line 161: That's the same as the super implementation, can we delete it?

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -279,8 +338,7 @@ def node_str(self, colour=True):
         :rtype: str
 
         '''
-        return "{0}[name='{1}']".format(
-            self.coloured_name(colour), self.symbol.name)
+        return f"{self.coloured_name(colour)}[name='{self.symbol.name}']"
```
sergisiso, line 341: This is almost the same (= to :) from the super implementation, can we delete this?

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -68,27 +73,35 @@ class AlgorithmInvokeCall(Call):
...
+            raise TypeError(
+                f"AlgorithmInvokeCall name argument should be a str but "
+                f"found '{type(name).__name__}'.")
         self._index = index
-        self.psylayer_routine_symbol = None
+        # Keep the root names as these will also be needed by the
+        # PSy-layer to use as tags to pull out the actual names from
+        # the algorithm symbol table, once issue #753 is complete.
+        self.psylayer_routine_root_name = None
+        self.psylayer_container_root_name = None
```
sergisiso, line 96: Usually we have attributes with a _ prefix and a property getter to access them (do we have a rule that it always must be this way?). But in any case, it may be a good idea here regardless if the names should always be created following the rules of "create_psylayer_symbol_root_names" and not by a direct assignment.

src/psyclone/domain/common/algorithm/psyir.py
```diff
@@ -128,54 +146,86 @@ def _validate_child(position, child):
...
-            root_name=routine_root_name)
-
-        container_root_name = "{0}_mod".format(routine_name)
-        container_name = symbol_table.next_available_name(
-            root_name=container_root_name)
-
-        interface = ImportInterface(ContainerSymbol(container_name))
-        self.psylayer_routine_symbol = RoutineSymbol(
-            routine_name, interface=interface)
+        self.psylayer_routine_root_name = self._def_routine_root_name()
```
sergisiso, line 209: I think it would be a good idea to only allow the name creation the first time:
```
if self.psylayer_routine_root_name is None:
    # Current implementation
```
Since there may already exist a matching PSy-layer routine if the create has been called before, and this could leave it orphan.

src/psyclone/domain/nemo/transformations/create_nemo_kernel_trans.py
```diff
@@ -79,7 +80,7 @@ class CreateNemoKernelTrans(Transformation):
                                 ArrayReference[name:'tmp']
                                     Reference[name:'ji']
                                 BinaryOperation[operator:'MUL']
-                                    Literal[value:'2', Scalar<INTEGER, UNDEFINED>]
+                                    Literal[value:'2.0', Scalar<REAL, UNDEFINED>]
```
sergisiso, line 83: Line > 80 chars

src/psyclone/generator.py
```diff
@@ -232,24 +235,46 @@ def generate(filename, api="", kernel_paths=None, script_name=None,
...
+
+                # Create Fortran from Algorithm PSyIR
+                writer = FortranWriter()
+                alg_gen = writer(psyir)
+            else:
+                alg_gen = Alg(ast, psy).gen
+        else:
+            alg_gen = None
+    except Exception:
+        raise
```
sergisiso, line 277: Why do we need this try/except, if we just raise an exception again?

src/psyclone/domain/common/algorithm/psyir.py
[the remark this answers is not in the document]
sergisiso: Ok, so can you add a TODO to #753 here saying that this can currently produce a routine_symbol that doesn't match the PSy-layer routine name.

sergisiso (changes requested):
@rupertford Thanks for re-taking this, it's been a while from the last review :) I do appreciate that this is a towards PR and there are some open questions about how to bridge old and new way of doing things. I am happy with how most of my previous comment have been addressed and there are just minor things to fix now.

I also checked that it works with the NemoLite2D algorithm layer.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: The name used here is different to the parent class. It is mentioned in the docstring.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: Yes, well spotted. I guess there is a more general issue of whether we should use ':' or '=' for consistency but that is to do with the parent classes so is separate to this PR.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: Yes good point. At the moment they are only used internally so there is no need to make accessors.

src/psyclone/domain/common/algorithm/psyir.py
rupertford: I've protected the assignments and added tests. Note, I've done separate first time tests for both assignments as it feels more robust, even though it is not necessary.

src/psyclone/domain/nemo/transformations/create_nemo_kernel_trans.py
rupertford: There is an issue for this (and for `create_nemo_psy_trans.py`) - see #1528 (where the indentation of the `view()` output will be made controllable (so it can be reduced))

src/psyclone/generator.py
rupertford: It was like this so I kept it. However, there are no excuses as I wrote the offending code in the first place. Now removed.

src/psyclone/domain/common/algorithm/psyir.py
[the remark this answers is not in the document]
rupertford: Added, and also put a comment in #753.

rupertford:

sergisiso (approved):
@rupertford All comments have been addressed and I double check it still works with NemoLite2D. This is ready to merge.

PR #1233, opened 2021-05-01
Create PSyclone TL2AD wrapper script (closes #1231)

### LINKED ISSUE #1231: Outline script that will translate TL kernels to AD kernels
An initial outline script is required for the adjoint project. This should take a TL kernel file as input, parse it with fparser2, translate it to language-level PSyIR, provide hooks/comments for translating to LFRic-specific PSyIR, provide hooks/comments for performing transformations to convert to an AD kernel and then write out the resultant transformed PSyIR to file.

rupertford:


rupertford:
@arporter, before I start adding tests, would you mind taking a quick look to see if you are happy with the general structure.

src/psyclone/psyad/tl.f90
```diff
@@ -0,0 +1,12 @@
+module x
```
arporter, line 1: I'd put this in a new examples directory.

arporter:
The structure looks as I imagined it would. I would expect us to pass PSyIR around rather than strings though. Sergi's new `FortranReader` frontend should help with that (and is on master now).

Are we going to get the name of the kernel routine from the metadata or just do every routine we can find (assuming that everything is active)? I would guess the former?

src/psyclone/psyad/tl.f90
rupertford: Yes, this will be removed. I've only added it for quick testing.

rupertford:
> The structure looks as I imagined it would. I would expect us to pass PSyIR around rather than strings though. Sergi's new `FortranReader` frontend should help with that (and is on master now).

I've kept a strings interface as I think this will be a useful API for integrating into other code (and also for testing!). We don't want to have to read from a file every time and (I would argue) might want to start from a string representation of code. However, I have also changed the `psyad()` function to take PSyIR as input and produce PSyIR as output which I think is what you were after.

I have also updated the code to use Sergi's new frontend. This makes the code simpler but didn't really affect the file vs string vs PSyIR issue.

> Are we going to get the name of the kernel routine from the metadata or just do every routine we can find (assuming that everything is active)? I would guess the former?

I'm not sure, but yes I think it makes sense to start from a recognised kernel and ignore any other code, unless the kernel itself calls it. We should really do some validation that the code is what we expect as well. Having said that, I think we should also be considering using this functionality on more general code as well (presumably with a different front end as it would have different constraints) so we could theoretically take any tangent linear code and provide its adjoint.

rupertford:
This is finally ready for first review from @arporter subject to github actions being happy. It's been complicated by problems with Python2/3 strings, logging output with github actions and with pytest affecting argparse output.

rupertford:
Note, I've not gone for any documentation or examples yet as the script doesn't actually do anything. I appreciate the reviewer may have a different attitude.

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+                      been invoked with.
+
+    '''
+    parser = argparse.ArgumentParser(
+        description="Run the PSyclone adjoint code generator on an LFRic "
+        "tangent-linear kernel file")
+    parser.add_argument(
+        '-v', '--verbose', help='increase the verbosity of the output',
+        action='store_true')
+    parser.add_argument('-oad', help='filename for the transformed code')
```
arporter, line 68: I know this is consistent with how vanilla PSyclone does things but would it make more sense just to support, e.g.:

    psyad my_tl_kernel_mod.f90 my_adj_tl_kernel_mod.f90

i.e. just have the input and output filenames as mandatory arguments.

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+        logging.basicConfig(level=logging.DEBUG)
+
+    # TL Fortran code
+    filename = args.filename
+    logging.info("Reading file %s", filename)
+    with open(filename) as my_file:
+        tl_fortran_str = my_file.read()
+        tl_fortran_str = six.text_type(tl_fortran_str)
+
+    ad_fortran_str = main_str(tl_fortran_str)
```
arporter, line 83: I think this name could be clearer (my first thought was "what on earth does that do?"). Maybe `adjoint_of_fortran_str()`?

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+    parser.add_argument('filename', help='LFRic tangent-linear source code')
+
+    args = parser.parse_args(args)
+
+    if args.verbose:
+        logging.basicConfig(level=logging.DEBUG)
+
+    # TL Fortran code
+    filename = args.filename
+    logging.info("Reading file %s", filename)
```
arporter, line 78: Perhaps "Reading kernel file"

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+    with open(filename) as my_file:
+        tl_fortran_str = my_file.read()
+        tl_fortran_str = six.text_type(tl_fortran_str)
+
+    ad_fortran_str = main_str(tl_fortran_str)
+
+    # AD Fortran code
+    if args.oad:
+        logging.info("Writing file %s", args.oad)
```
arporter, line 87: Perhaps "Writing adjoint of kernel to %s"?

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+        write_unicode_file(ad_fortran_str, args.oad)
+    else:
+        print(ad_fortran_str)
```
arporter, line 90: Please add a `file=sys.stdout` on this to be explicit.

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def main_str(tl_fortran_str):
+    '''Takes an LFRic tangent-linear kernel encoded as a string as input
+    and returns its adjoint encoded as a string.
+
+    :param str tl_fortran_str: a string containing the LFRic \
```
arporter, line 97: s/a string containing the/Fortran implementation of an/

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+        tangent-linear kernel.
+
+    :returns: a string containing the adjoint of the supplied \
```
arporter, line 100: s/a string containing the/Fortran implementation of the/

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+        tangent-linear kernel.
+    :rtype: str
+
+    '''
+    logging.debug(tl_fortran_str)
+
+    # TL Language-level PSyIR
+    reader = FortranReader()
+    tl_psyir = reader.psyir_from_source(tl_fortran_str)
+    with Capturing() as output:
```
arporter, line 110: Presumably this is the bit you would replace by the proposed alteration to `view()` so that it returns a `str`?

src/psyclone/psyad/main.py
```diff
@@ -0,0 +1,143 @@
...
+        tl_psyir.view()
+    logging.debug("\n".join(output))
+
+    # TL to AD translation
+    ad_psyir = generate(tl_psyir)
```
arporter, line 115: Perhaps `generate_adjoint`?

src/psyclone/psyad/tl2ad.py
```diff
@@ -0,0 +1,68 @@
...
+
+'''The implementation of PSyAD : the PSyclone Adjoint
+support. Transforms an LFRic tangent linear kernel to its adjoint.
+
+'''
+import logging
+
+
+def generate(tl_psyir):
+    '''Takes an LFRic tangent-linear kernel represented in language-level PSyIR
```
arporter, line 44: If this is LFRic specific then perhaps we need to take care with naming. However, I'm happy to leave that for now given the time constraints. We can generalise later.

src/psyclone/psyir/nodes/node.py
```diff
@@ -807,11 +809,13 @@ def view(self, indent=0, index=None):
...
-                                  self.node_str(colour=True)))
+            result = "{0}{1}".format(self.indent(indent),
+                                     self.node_str(colour=True))
+            print(six.text_type(result))
         else:
-            print("{0}{1}: {2}".format(self.indent(indent), index,
-                                       self.node_str(colour=True)))
+            result = "{0}{1}: {2}".format(self.indent(indent), index,
+                                          self.node_str(colour=True))
+            print(six.text_type(result))
```
arporter, line 818: Please bring this `print` out of the `if` block so that you can remove the one at line 814.

src/psyclone/tests/psyad/main_test.py
```diff
@@ -0,0 +1,260 @@
...
+    assert error == ""
+    # The name of the executable is replaced with either pytest or -c
+    # when using pytest, therefore we split this test into sections.
+    expected1 = "usage: "
+    expected2 = (
+        "[-h] [-v] [-oad OAD] filename\n\n"
+        "Run the PSyclone adjoint code generator on an LFRic tangent-linear "
+        "kernel file\n\n"
+        "positional arguments:\n"
+        "  filename       LFRic tangent-linear source code\n\n"
```
arporter, line 67: s/source code/kernel source/

src/psyclone/tests/psyad/main_test.py
```diff
@@ -0,0 +1,260 @@
...
+    filename_out = str(tmpdir.join("ad.f90"))
+    with open(filename_in, "a") as my_file:
+        my_file.write(tl_code)
+    with caplog.at_level(logging.DEBUG):
+        main([filename_in, "-v", "-oad", filename_out])
+
+    output, error = capsys.readouterr()
+    assert error == ""
+    assert output == ""
+    assert "INFO     root:main.py:75 Reading file /" in caplog.text
```
arporter, line 197: Just a note that it would be nicer if we could have s/root/psyclone.psyad/

src/psyclone/tests/psyad/tl2ad_test.py
```diff
@@ -0,0 +1,108 @@
...
+        "a = 0.0\n"
+        "end program test\n")
+    expected_ad_fortran_str = (
+        "program test\n"
+        "  integer :: a\n\n"
+        "  a = 0.0\n\n"
+        "end program test\n")
+    reader = FortranReader()
+    tl_psyir = reader.psyir_from_source(tl_fortran_str)
+
```
arporter, line 65: Unless I'm missing something, this test doesn't actually call `generate()`?

src/psyclone/tests/psyad/tl2ad_test.py
```diff
@@ -0,0 +1,108 @@
...
+    with caplog.at_level(logging.INFO):
+        ad_psyir = generate(tl_psyir)
+        assert caplog.text == ""
+
```
arporter, line 93: Should have a check here that the expected logging output has been created?

arporter (changes requested):
Looks good. My only real quibble is with some of the names but (agreement permitting) that is easily fixed.
I think we should get into the habit of doing documentation now for two reasons:

  1. Writing down what it is we do is very helpful in checking the design
  2. We don't want to have to write it all from scratch at some point in the future
 
That said, I don't think we should go overboard. A quick cut-n-paste of the 'psyclone script' description and then some editing should do. Rather than add to the already massive PSyclone UG, perhaps we should have a new RTD site, PSyclone-adj perhaps?

src/psyclone/psyad/main.py
rupertford: I'm not convinced. I think it might be useful for the user to be able to see the output quickly if they are developing/testing, rather than having to look in an output file. How about we ask the users what they think?

src/psyclone/psyad/main.py
rupertford: I agree. The name is awful. I just couldn't think of something appropriate and was banking on you suggesting something :-). However, I'm not sure about what you are suggesting I'm afraid. How about `psyad_str()` and we also rename the `psyad()` function to `psyad_psyir()`?

src/psyclone/psyad/main.py
rupertford: OK.

src/psyclone/psyad/main.py
rupertford: Done.

src/psyclone/psyad/main.py
rupertford: OK

src/psyclone/psyad/main.py
rupertford: Updated.

src/psyclone/psyad/main.py
rupertford: Yes, I've added a comment referencing the issue.

src/psyclone/psyad/main.py
rupertford: Changed.

src/psyclone/psyad/tl2ad.py
rupertford: That's a good point. I would like to keep the functionality general and then specialise for LFRic kernels, i.e. in general we can take in arbitrary Fortran TL code and translate it.

src/psyclone/psyir/nodes/node.py
rupertford: Oops.

src/psyclone/tests/psyad/main_test.py
rupertford: updated.

src/psyclone/tests/psyad/main_test.py
rupertford: Updated to provide the package information instead of just using root.

src/psyclone/psyad/main.py
rupertford: I've actually ended up removing the Capturing class as it was not allowing me to format the output as I would have liked (as everything including escape characters are turned into strings).

src/psyclone/tests/psyad/tl2ad_test.py
rupertford: You are right. It also wasn't being called (and would have failed if so) as it had the same name as the next test!

src/psyclone/tests/psyad/tl2ad_test.py
rupertford: I'm not sure I understand the comment I'm afraid. This does test the logging output with the `assert caplog.text == ""` line. It just happens that there is no logging output when the logging level is set to INFO.

src/psyclone/psyad/main.py
rupertford: modified.

rupertford:
OK, ready for next review from @arporter

src/psyclone/psyad/main.py
arporter: I didn't say, but I meant that if no output file was provided, then the script would still output to stdout. However, asking the users sounds like a good plan.

arporter: TBH, having just read the new documentation (thanks) I don't think it's worth holding this PR up for this.

src/psyclone/tests/psyad/tl2ad_test.py
arporter: No, I don't understand the comment either!

arporter (changes requested):
Almost there now. Please could you just check whether we do need the `make.bat` file and get rid if not.
New documentation (thanks) builds OK but warns that the 'static' path doesn't exist - please could you comment out the setting in conf.py.
Tests all OK.
No changes to examples.

arporter:
Please could you also add a `doc/psyad/user_guide/pip_requirements.txt` file for RTD.

doc/psyad/user_guide/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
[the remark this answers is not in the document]
rupertford: Removed. I just kept the files that the sphinx setup generated. I've removed it now. Note, we have one for the psyclone user guide (but not developer or reference) :-)

doc/psyad/user_guide/conf.py
```diff
@@ -0,0 +1,55 @@
...
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
```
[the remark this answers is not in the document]
rupertford: Ah, yes good point.

rupertford:
> Please could you also add a `doc/psyad/user_guide/pip_requirements.txt` file for RTD.

I've added this. For futureproofing I kept the bibtex requirement even though I presume we don't need it at the moment.

PR #1252, opened 2021-05-19
(closes #1251) bug fix for ranges in derived-type accesses

### LINKED ISSUE #1251: PSyIR for ranges in derived-type references is wrong
Attempting to compile the Fortran generated from the PSyIR for NEMO has revealed problems with the PSyIR constructed for derived type references, e.g.:

    vars(1)%region%subgrid(3)%data(:) = 1.0
    vars(:)%region%subgrid(3)%xstop

The `Range` nodes introduced still contain references to the 'fake parent' nodes introduced while building the reference. These should in fact be either `StructureReference` or `ArrayOfStructuresReference` nodes.

While working on this, I've also noticed that the code in the Fortran backend that reduces `my_array(LBOUND(my_array, 1):UBOUND(my_array, 1))` to `my_array(:)` doesn't work when the reference is to a derived type.

arporter:


arporter:
Assuming the CI is now happy this is ready for review.

src/psyclone/psyir/nodes/array_member.py
```diff
@@ -88,6 +91,128 @@ def create(member_name, indices):
...
+            return False
+        if not isinstance(lower.children[0], Reference):
+            return False
+
+        # Check that the full structure access being passed to the LBOUND
+        # operator matches the one for this node.
+        return self._matching_access(lower.children[0])
+
+    def is_upper_bound(self, index):
+        '''Returns True if the specified array index contains a Range node
```
sergisiso, line 130: Same as previous method: contains -> is, consider removing example (I am less sure about this), use semantic navigation when possible to make code clearer.

src/psyclone/psyir/nodes/array_member.py
```diff
@@ -88,6 +91,128 @@ def create(member_name, indices):
...
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.backend.fortran import FortranWriter
+        fwriter = FortranWriter()
+
+        # This node is at the 'bottom' of the structure access so we
+        # need to get the parent Reference.
+        parent_member = self.ancestor(Reference)
+        current_member = node
+
+        if type(current_member) != type(parent_member):
```
sergisiso, line 185: There is an unidiomatic typecheck pylint warning here. Also can you add a comment about why this check is needed?

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
```diff
@@ -278,14 +330,28 @@ def test_derived_type_ref(f2008_parser):
     assert isinstance(amem, ArrayOfStructuresMember)
     assert isinstance(amem.member, ArrayMember)
     assert isinstance(amem.member.indices[0], Range)
-    assign = assignments[4]
+    assert isinstance(amem.member.indices[0].children[0], BinaryOperation)
+    assert amem.member.indices[0].children[0].children[0].symbol.name == "var"
+    gen = fortran_writer(amem)
+    assert (gen == "subgrid(3)%data(:)")
```
sergisiso, line 336: pylint says the parenthesis are not needed

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
```diff
@@ -278,14 +330,28 @@ def test_derived_type_ref(f2008_parser):
...
     assert isinstance(amem, ArrayMember)
     assert isinstance(amem.indices[0], Range)
     assert isinstance(amem.indices[0].children[0], StructureReference)
     assert isinstance(amem.indices[0].children[0].member, Member)
     assert amem.indices[0].children[0].member.name == "start"
+    # vars(1)%region%subgrid(3)%data(:) = 1.0
+    assign = assignments[5]
+    amem = assign.lhs
+    gen = fortran_writer(amem)
+    assert (gen == "vars(1)%region%subgrid(3)%data(:)")
```
sergisiso, line 349: pylint says the parenthesis are not needed

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
```diff
@@ -278,14 +330,28 @@ def test_derived_type_ref(f2008_parser):
...
+    # vars(:)%region%subgrid(3)%xstop
+    assign = assignments[6]
+    amem = assign.lhs
+    gen = fortran_writer(amem)
+    assert (gen == "vars(:)%region%subgrid(3)%xstop")
```
sergisiso, line 354: pylint says the parenthesis are not needed

sergisiso (changes requested):
@arporter Thanks for improving this, some of these methods are getting quite complex but  should be ok as they are relatively self-contained.

One aspect that I am wondering is that if the ArrayMember is_lower/upper_bound are needed since they look the same as their generic ArrayMixin versions? The ArrayMixin versions end with an extra:

```
                 upper.children[0].name == self.name):
             return False
         if not (isinstance(upper.children[1], Literal) and
                 upper.children[1].datatype.intrinsic ==
                 ScalarType.Intrinsic.INTEGER
                 and upper.children[1].value == str(index+1)):
             return False
         return True
```
Isn't the first part (`upper.children[0].name == self.name`) essentially their "_matching_access"? Could ArrayMixin have an abstract _matching_accesses and this bit of code be moved to the a ArrayReference "_matching_access implementation?

Then the second condition shouldn't is also be part of the ArrayMember is_lower/upper_bound?
Since now something like `a%b(:,1:ubound(a%b,1))` would return True while it should be false because the boundary is `a%b(:,1:ubound(a%b,2))`. Is that right? Probably we need some tests about this.

If both these things are true maybe you can get rid of the specialized is_upper/lower_bound and just let the _matching_accesses deal with the specialisation.

arporter:
I thought that perhaps I could use the new Signature functionality added by @hiker to make it easier to decide whether a UBOUND/LBOUND in an array index expression is referring to the parent array. (We use this in the backend to get back to ':' notation from PSyIR.) However, I have a problem. Consider `a%b(:)%c`. In the PSyIR this becomes `a%b(LBOUND(a%b):UBOUND(a%b))%c` and so my task is to spot that the arg to LBOUND (`a%b`) matches the array expression it occurs in. If I want to use `get_signature_and_indices` then I have to go 'up' from `b` to `a` in order to call it but the resulting signature will then be for `a%b%c` which does not match with the signature of the argument to LBOUND. I'm thinking that it might make sense to change the implementation of `get_signature_and_indices` so that, for a *member* of a structure, it itself goes 'up' to the parent reference and not down. For a *Reference*, it would still go 'down'. So the signature of StructureMember `b` in `a%b%c` would be `a%b` but, for the StructureReference `a`, it would still give `a%b%c`. 

An alternative would be to extend `get_signature_and_indices` to allow a `stop` argument to be supplied but I don't think this would be as nice.

The final alternative would be to go back to what I had before but I don't want to reinvent the wheel.

arporter:
Having written all that, I think that what I suggest as my preferred option is not possible as it would require the `StructureMember.get_signature_and_indices()` method to know whether it is being called directly (and therefore needs to go 'up') or whether it is being called in the context of a call to e.g. `StructureReference.get_signature_and_indices()` in which case it needs to go 'down'. I suppose I could modify `StructureReference.get_signature_and_indices()` so that it finds the 'bottom' access and then calls `get_signature_and_indices` on that which would then work its way back up.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1009,52 +1009,15 @@ def range_node(self, node):
...
+        if node.parent.is_lower_bound(node.parent.indices.index(node)):
             # The range starts for the first element in this
             # dimension. This is the default in Fortran so no need to
             # output anything.
             start = ""
         else:
             start = self._visit(node.start)
 
-        if _full_extent(node.stop, BinaryOperation.Operator.UBOUND):
+        if node.parent.is_upper_bound(node.parent.indices.index(node)):
```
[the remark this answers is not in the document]
arporter: Oops, done.

src/psyclone/psyir/nodes/array_member.py
```diff
@@ -88,6 +91,128 @@ def create(member_name, indices):
             obj.addchild(child)
         return obj
 
+    def is_lower_bound(self, index):
+        '''Returns True if the specified array index contains a Range node
```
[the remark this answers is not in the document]
arporter: This method has been removed.

src/psyclone/psyir/nodes/array_member.py
```diff
@@ -88,6 +91,128 @@ def create(member_name, indices):
...
+        which has a starting value given by the 'LBOUND(access-expr,index)'
+        intrinsic where 'access-expr' is the full access to the current
+        ArrayMember and 'index' matches the specified array index. Otherwise
+        False is returned.
+
+        For example, if a Fortran array A%B was allocated with extent 10
```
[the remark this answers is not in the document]
arporter: Gone now.

src/psyclone/psyir/nodes/array_member.py
```diff
@@ -88,6 +91,128 @@ def create(member_name, indices):
...
+        :rtype: bool
+
+        '''
+        self._validate_index(index)
+
+        array_dimension = self.indices[index]
+        if not isinstance(array_dimension, Range):
+            return False
+
+        lower = array_dimension.children[0]
```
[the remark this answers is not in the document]
arporter: Oh yes, done.

src/psyclone/psyir/nodes/array_member.py
arporter: Done (in array_mixin now).

src/psyclone/psyir/nodes/array_member.py
arporter: That too is gone.

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
arporter: Done.

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
arporter: Done/removed (because this was really only testing the backend and actually there were errors in the PSyIR constructed by the frontend).

src/psyclone/tests/psyir/frontend/fparser2_derived_type_test.py
arporter: Done.

arporter:
My last idea also didn't work as starting the recursion from e.g. StructureReference also ended up with the problem where, ultimately, we needed to recurse back up to that class. Instead, I've added a `max_depth` arg to the get_signature_and_indices() methods which simply limits how deeply into a structure access it should recurse.

arporter:
The test failures under python 2 are due to #1283. That will need looking at before I decide how to complete this PR.

arporter:
@hiker has pointed out that #1278 will add a `getitem` to the Signature which I think will completely do away with my need to alter `get_signature_and_indices` in order to limit the depth. Therefore this PR now depends on #1278 and #1283.

arporter:
Assuming the CI is now happy, this is ready for another look. I'm afraid that quite a lot has changed (albeit for the better) so you may need to do it from scratch - sorry.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -451,6 +451,62 @@ def _is_range_full_extent(my_range):
     return is_lower and is_upper and is_step
 
 
+def _create_array_bound_arg(node):
```
sergisiso, line 454: It looks like the name explains why is it called by the current caller but not what it does. Could this function be named `_copy_full_base_reference`?

(Another possibility that I am not going to ask in this PR is for this to be methods in Member and Reference, there is nothing frontend specific and this could be useful in other places)

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -451,6 +451,62 @@ def _is_range_full_extent(my_range):
...
+              UBOUND applied to the supplied `node`.
+    :rtype: :py:class:`psyclone.psyir.nodes.Node`
+
+    :raises InternalError: if the supplied node is not an instance of \
+                           either Reference or Member.
+    '''
+    if isinstance(node, Reference):
+        return Reference(node.symbol)
+
+    elif isinstance(node, Member):
```
sergisiso, line 479: pylint says the "else" is not necessary

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -451,6 +451,62 @@ def _is_range_full_extent(my_range):
...
+        # to find the copy of 'node'.
+        inner = arg
+        for step in range(depth-1):
+            inner = inner.member
+        # Change the innermost access to be a Member.
+        inner.children[0] = Member(node.name, inner)
+        return arg
+
+    raise InternalError(
+        "The supplied node must be an instance of either ArrayReference "
```
sergisiso, line 506: Looking at the conditions I guess this should be "either a Reference or a Member"

src/psyclone/tests/psyir/frontend/fparser2_subscript_triplet_handler_test.py
```diff
@@ -0,0 +1,117 @@
...
+    reader = FortranStringReader("x(:) = a")
+    fp2node = Execution_Part.match(reader)
+    triplet = fp2node[0][0].children[0].children[1].children[0]
+    with pytest.raises(InternalError) as err:
+        processor._subscript_triplet_handler(triplet, fake_parent)
+    assert ("Expected parent PSyIR node to be either a Reference or a Member "
+            "but got 'Schedule' when processing ':'" in str(err.value))
+
+
+def test_ubound_lbound_arg_error(fortran_reader):
```
sergisiso, line 67: Unused argument

sergisiso (changes requested):
@arporter This is almost ready now. The way the array mixin is implemented now is much simpler, well done.

I checked implementation, pycodestyle and pylint, tests with compilation and examples, diff coverage, no remaining TODOs and the reference guide builds without errors. Since this is a bugfix no additional documentation is needed.

src/psyclone/psyir/frontend/fparser2.py
arporter: Good point, renamed.

src/psyclone/psyir/frontend/fparser2.py
arporter: Oops again.

src/psyclone/tests/psyir/frontend/fparser2_subscript_triplet_handler_test.py
arporter: Oops.

arporter:
Assuming the CI is happy, this is ready for another look now.

sergisiso (approved):
@arporter All comments addressed, this is ready to be merged.

PR #1255, opened 2021-05-20
(Closes #1253) Adds support for interface blocks and checks *all* statements in specification section

### LINKED ISSUE #1253: PSyIR loses interface blocks
In Fortran, a module can define an interface to a routine, e.g.:

    module my_mod

      INTERFACE eos
         MODULE PROCEDURE eos_insitu, eos_insitu_pot, eos_insitu_2d
      END INTERFACE

    contains

    ...
    end module my_mod

Currently, in parsing the Fortran into PSyIR and then using the backend to recreate it, such interface blocks are lost. I'm wondering whether we could have an `UnknownFortranType` for `eos` in this case?

arporter:
This PR primarily improves the fparser2 frontend such that:

 * Named interface blocks are captured as UnknownFortranType declarations;
 * Access (visibility) information is captured for UnknownFortranType declarations;
 * Backend extended to produce access specifiers for symbols of UnknownFortranType;
 * the frontend now raises NotImplementedError if it encounters an unsupported node in the parse tree of a specification-part;


arporter:
Assuming the CI is happy, this is ready for review.

arporter:
Taking this back as, on testing with NEMO, I've realised that there's a problem: we now get a `DataSymbol` for a named interface block but this causes an error in the frontend if a call to this routine is encountered as Call is only permitted to refer to a `Symbol` or a `RoutineSymbol`. I think the solution to this is to give `RoutineSymbol` a datatype as @sergisiso has previously suggested. This then fits with describing functions that have a return value as well as allowing us to solve this problem.

arporter:
In fact, the latter problem is a consequence of my workaround for the former - we should really only generate an access statement for Routine symbols but I've currently changed that to do it for symbols with UnknownFortranType too. So, I need to add a datatype to Routine symbols (#1259).

arporter:
An alternative would be to add a temporary workaround that examines the content of any UnknownFortranType to see whether it is in fact an interface block. This isn't nice but it would work in the short term.

arporter:
I've undone the changes in the backend so that access statements are now only generated for RoutineSymbols. I've anticipated #1294 going on master and altered the frontend to use a RoutineSymbol of UnknownFortranType when it encounters a named interface. Obviously, until #1294 is on, this won't work!

arporter:
This is ready for a first look now. Probably one for either @rupertford or @sergisiso.

arporter:
Note that there is one line reported as not covered but it is within a class defined within a test.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2026,40 +2049,83 @@ def process_declarations(self, parent, nodes, arg_list,
...
+        for node in nodes:
+
+            if isinstance(node, Fortran2003.Interface_Block):
+
+                # We only support named interface blocks, and then only
+                # partially.
+                names = walk(node.children[0], Fortran2003.Name)
+                if not names:
+                    raise NotImplementedError()
+                name = names[0].string
```
sergisiso, line 2061: Is there any situation where an unnamed interface block contains something which can have a Fortran2003.Name construct and would trick this check?

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
```diff
@@ -0,0 +1,154 @@
...
+      INTEGER P (1000)
+      LOGICAL Q (1000)
+      END FUNCTION EXT3
+      END INTERFACE
+    end module dummy_mod
+    '''
+    container = fortran_reader.psyir_from_source(dummy_module)
+    assert isinstance(container, FileContainer)
+    assert len(container.children) == 1
+    assert isinstance(container.children[0], CodeBlock)
```
sergisiso, line 107: This could still be problematic as although the output will contain the Interface, the symbol table will not contain EXT1, EXT2 and EXT3, and it may refer to them, is that right? No need to be solved it in this PR but maybe worth mentioning it.

sergisiso (changes requested):
@arporter The initial support for interface blocks looks good, I also like the processing of all declaration nodes one by one instead of the walk. I made a few questions about the interfaces and if we can avoid some code duplication.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -632,6 +633,22 @@ def gen_decls(self, symbol_table, args_allowed=True):
...
+                        "Routine symbol '{0}' is of UnknownType rather than "
+                        "UnknownFortranType. This is not supported by the "
+                        "Fortran back-end.".format(sym.name))
+                if not isinstance(sym.interface, LocalInterface):
+                    raise VisitorError(
+                        "Routine symbol '{0}' is of UnknownFortranType but has"
+                        " interface '{1}' instead of LocalInterface. This is "
+                        "not supported by the Fortran back-end.".format(
+                            sym.name, sym.interface))
+                declarations += self.gen_vardecl(sym)
```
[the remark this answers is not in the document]
arporter: That's a good point. I've moved/removed both and improved the error messages a bit.

src/psyclone/tests/psyir/backend/fortran_unknown_declns_test.py
```diff
@@ -0,0 +1,139 @@
...
+    class OtherType(UnknownType):
+        ''' UnknownType is abstract so sub-class it for this test '''
+        def __str__(self):
+            return "OtherType"
+
+    container = Container("my_mod")
+    container.symbol_table.add(RoutineSymbol("eos", OtherType("some code")))
+    with pytest.raises(VisitorError) as err:
+        fortran_writer.gen_decls(container.symbol_table)
+    assert ("Routine symbol 'eos' is of UnknownType rather than "
```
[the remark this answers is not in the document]
arporter: As it happens, I did that as part of the earlier change :-)

src/psyclone/psyir/frontend/fparser2.py
arporter: I don't think so. I've stared at the Fortran spec some more and am pretty confident. However, I've added a comment and hardened the check a bit so that it no longer just does a walk.

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
arporter: You had me worried there but actually, it's OK because we now get a CodeBlock for the whole module containing such an interface definition. I've added a comment.

sergisiso: I know it creates a CodeBlock and kind of solve the problem is they aren't touched, my worry was for EXT1, EXT2 and EXT3 symbols not being declared and not accessible from inside the module "contains".

arporter: I don't quite follow - the CodeBlock will contain the whole module - both any variable/interface declarations and any routines within a CONTAINS section? Do you mean if some other code `use`s the module then we won't be able to find (any) variables that it defines?

arporter:
I think this is ready for review again - looks like I forgot to change the tag.

sergisiso (approved):
@arporter All comments have been addressed. Compilation tests and examples run successfully. There is no pycodestyle issues, there is 1 pylint complain of an unused import that I will delete myself, no TODOs remaining. No additional documentation is needed.

I approve the PR but I will wait for #1327 to be merged first.

arporter:
Thanks Sergi. I'd forgotten about #1327 but it's merged now :-)

Issue #1258, opened 2021-05-26
[LFRic] add PSy-layer symbols to symbol table


arporter:
In #1239 we added 'lower-to-language-level' support for some LFRic BuiltIns and that meant we needed Symbols for the various arguments that they referred to. However, in general we should be populating the SymbolTable for the PSy layer with the symbols for the various quantities that are both passed in and created. This can be done alongside the existing f2pygen code-generation functionality.

This is also related to #696 which will require symbols for the various kind parameters needed to specify the precision of constants.

arporter:
Currently the names of LFRic infrastructure modules (e.g. "field_mod") are hardwired in the code. These should be read from the config file.

TeranIvy:
@arporter, is this going to be one PR or a few of them? I'd like to get involved in doing this work so I could do a PR towards adding symbols for all classes in LFRic.

TeranIvy:
We agreed that the names of the LFRic infrastructure modules should actually be in the `LFRicConstants` class, as this is mandated by the LFRic infrastructureis and it is not something that the user can change. It is extremely unlikely that LFRic will change the name of modules that store data structures as  the entire code would need to change. Moreover, any big LFRic infrastructure change will necessarily have to be paired with the corresponding PSyclone change.

I will create a PR for this work. The changes should be confined to `LFRicConstants` and  `dynamo0p3.py`.

As part of this work I will also introduce the mapping `<infrastructure_module_name>:<data_structure_type>:<fortran_datatype>:<default_kind>`.  The name of a data structure in LFRic (e.g. `field_type`) is tied to the infrastructure module name where this structure is defined (e.g. `field_mod`). The data structure type has its Fortran datatype and kind defined in this module (e.g. `real(kind=r_def)` for `field_type`). Potentially, a data structure type can have allowed kinds in addition to the default kind but they would also be infrastructure-defined and this is not something we need to worry about now.

We currently support the following mappings for the LFRic data structures and their proxies:
- Field and its proxy: `field_mod`:`field_type`:`real`:`r_def`,
- Integer field and its proxy: `integer_field_mod`:`integer_field_type`:`integer`:`i_def`,
- Operator and its proxy: `operator_mod`:`operator_type`:`real`:`r_def`,
- Columwise operator and its proxy: `operator_mod`:`columnwise_operator_type`:`real`:`r_def`.

[...]

PR #1264, opened 2021-05-27
[LFRic] Move names of infrastructure modules to LFRicConstants (towards #1258)

TeranIvy:
This PR adds mappings of LFRic infrastructure types/data to their LFRic infrastructure modules to LFRic constants, so they are not hard-coded any more.

I followed the existing example of `LFRicConstants.QUADRATURE_TYPE_MAP` that uses internal PSyclone names for quadrature types and adds information about module, type and proxy type names (the latter two where available).


src/psyclone/domain/lfric/lfric_constants.py
```diff
@@ -62,13 +63,13 @@ def __init__(self):
         LFRicConstants.HAS_BEEN_INITIALISED = True
         api_config = Config.get().api_conf("dynamo0.3")
```
arporter, line 64: Please could you disable the pylint 'too-many-statements' warning for this constructor.

arporter (changes requested):
A nice PR - thanks for keeping it short.
We've gained one new pylint warning because of the additional statements but that can be disabled.
Documentation does not require updating.
I did wonder whether it would be better to introduce a namedtuple with `module`, `type` and `proxy_type` properties but decided it wouldn't buy us much. We could always do this in the future if we think it would be better.
There's just one other minor thing to fix.

src/psyclone/domain/lfric/lfric_constants.py
```diff
@@ -283,6 +269,86 @@ def __init__(self):
...
+        # ---------- Infrastructure module maps -------------------------------
+
+        # Dictionary allowing us to look-up the name of the Fortran module,
+        # type and proxy-type associated with each LFRic data type.
+        LFRicConstants.DATA_TYPE_MAP = {
+            # 'real'-valued scalar of kind 'r_def' (used for global
+            # reductions of "field_type" data)
+            "scalar": {"module": "scalar_mod",
+                       "type": "scalar_type",
+                       "proxy_type": "field_proxy_type"},
```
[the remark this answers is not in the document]
TeranIvy: Oops, it doesn't - copy and paset gone wrong! Removed.

src/psyclone/domain/lfric/lfric_constants.py
TeranIvy: Disabled.

TeranIvy:
Addressed review comments, back to @arporter.

arporter (approved):
Tests and examples are OK, even after merging latest master.
Will proceed to merge.

Issue #1272, opened 2021-06-01
Test failure in gocean1p0_config_test.py


hiker:
If I run only this one test, I get parsing errors:
```
gocean1p0_config_test.py:359: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../configuration.py:380: in load
    self._api_conf[api] = GOceanConfig(self, self._config[api])
../configuration.py:1092: in __init__
    GOLoop.add_bounds(it_space)
../gocean1p0.py:712: in add_bounds
    six.raise_from(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value = None, from_value = NoMatchError("Nonlabel_Do_Stmt: 'do i=1, 1'")

>   ???
E   psyclone.configuration.ConfigurationError: "PSyclone configuration error: Expression '1' is not a valid do loop boundary. Error message: 'Nonlabel_Do_Stmt: 'do i=1, 1''."

<string>:3: ConfigurationError
```
I can avoid this if I run certain other tests before the config test, e.g. gocean0p1_test, dependency_test, alggen_test, but not all (e.g. running f2pygen_test first does not avoid the problem).

At this stage it appears that the failure is triggered by the class ``Base`` in fparser.two.utils.py. Around line 370:
```
       elif result is None:
            # Loop over the possible sub-classes of this class and
            # check for matches
            print("Looping", Base, cls.__name__)
            for subcls in Base.subclasses.get(cls.__name__, []):
                print("Looping", subcls)
                if subcls in parent_cls:  # avoid recursion 2.
                    continue
                try:
                    obj = subcls(string, parent_cls=parent_cls)
                except NoMatchError:
                    obj = None
                if obj is not None:
                    return obj
```
[...]

PR #1273, opened 2021-06-01
1272 fix failure running single test

hiker:
Fix for #1272 - as far as I can see there was an ``autouse=True`` missing. As a result the fparser initialisation was not executed on every test run, and I guess it was always dependent on something else initialising the parser.

I've also fixed some pylint issues.


arporter (changes requested):
Thanks for fixing and tidying. However, your good idea of adding an autouse means that all of the `@pytest.mark.usefixtures("f2008_parser")` (or "parser") are now redundant. Please could you create an issue to remove them.

hiker:
@arporter Is there a 'better' solution? I.e. can I just add this ``usefixtures`` to that file (I had no idea what this was :) ) ? I have no idea if the same problem exists for other tests, or what makes this one file 'special' (note that I often run one particular test of a file as well ... so would this mean we need to add this for every test that would show this problem?).

If you don't have a better idea, then I'll remove the ``usefixtures`` as part of this PR, that can likely be scripted.

hiker:
I had another look, as far as I can tell the problem is that in case of the gocean_configuration test, psyclone verifies that  valid loop boundaries are specified in the config file by creating a dummy loop with the loop boundaries, replacing `{start}` and `{stop}` with integers, and then parsing this dummy loop as `Nonlabel_Do_Stmt`. So at this stage no parser has been involved.

If we assume that this is the only place where we do something like this, we could just call the parser factory once before that line, that might be an alternative solution.

Just let me know what you prefer - autouse, and removing the other `usefixtures`, or trying to fix this in goloop (or even in config, before calling the goloop.add_bounds  function - since this will definitely be used less frequently, meaning less overhead).

hiker:
As discussed in #1272: @arporter commented on teams:

> To answer your question on the PR Joerg Henrichs I would just add a use fixtures to the offending test. We could keep the Issue itself and look to introduce the autouse setting to replace all the individual use fixtures.

After having yet another look, I have to disagree with both my original solution, and this suggestion. Reason is that if you use a config file with a custom gocean iteration space, both psyclone and genkernstub crash, because the config file in each case is read before the parser is initialised. While I agree that custom iteration spaces are rare, we need to support it. So I replaced this fix by making sure fparser is properly initialised before the `Nonlabel_Do_Stmt` is used.

src/psyclone/gocean1p0.py
```diff
@@ -703,6 +706,10 @@ def add_bounds(bound_info):
             # Now replace any {start}/{stop} expression in the loop
             # with a valid integer value:
             do_string = do_string.format(start='15', stop='25')
+            # We need to make sure the fparser is properly initialised, which
```
arporter, line 709: I think this fix is the right one. Please could you move the parser creation up out of this loop?

arporter (changes requested):
The fix looks good.
Please could you add a test that explicitly exercises the old bug (i.e. ensures that the parser has not been initialised and then attempts to generate loop bounds).
Ensuring that the parser has not been initialised is going to be tricky - some judicious use of monkeypatch will be required to unset some of the things created by fparser.two.parser.ParserFactory.create().
Please could you also bring the branch up-to-date with master.

src/psyclone/gocean1p0.py
hiker: Oops, I never looked further up 😊

hiker:
> The fix looks good.
> Please could you add a test that explicitly exercises the old bug (i.e. ensures that the parser has not been initialised and then attempts to generate loop bounds).
> Ensuring that the parser has not been initialised is going to be tricky - some judicious use of monkeypatch will be required to unset some of the things created by fparser.two.parser.ParserFactory.create().

I added a simple framework for this kind of test (fixture to 'remove' fparser, and then re-initialise it at the end). A bit of an overkill for a single test, but I wouldn't be surprised that we will have similar issues elsewhere, which is then pretty easy to add.

I got away by calling `...create([])`, which as a side effect removes the classes previously created. And it triggered the bug (if I removed the bug fix of course).

> Please could you also bring the branch up-to-date with master.

Also done.

src/psyclone/tests/psyir/frontend/fparser2_not_init_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Now execute all tests
+    yield
+
+    # We need to properly initialise fparser,
+    # otherwise followup tests will fail (if this test should fail)
+    ParserFactory().create(std="f2008")
+
+
+def test_loop_bound_when_fparser_not_initialised():
```
arporter, line 67: Thanks for adding a new file. However, since this is a test of GOLoop functionality, I think it should live in a GOcean-appropriate location (maybe `tests/domain/gocean/goloop_test.py`?) rather than in the PSyIR frontend. You'll need to move the fixture too and turn off its 'autouse'. As you say, the fixture could be useful in general so please could you put it in `conftest.py` and add some words about it to the section on fixtures in the dev guide.

arporter (changes requested):
Thanks for adding the new fixture Joerg, I think it will be useful. However, as I say in the comments, I don't think it should setup the parser once it is complete - that is the responsibility of each individual test. (On reflection I realised this would also have been a problem with your initial 'autouse' solution for the parser fixture - it would mask any problems caused by us failing to initialise the parser within PSyclone itself.)
Also, although I appreciate that you've gone to the trouble of adding a whole new test file, I think that actually the test belongs with those for goloop - again see the comment.
Once those things are done then this is good to go.

hiker:
> Thanks for adding the new fixture Joerg, I think it will be useful. However, as I say in the comments, I don't think it should setup the parser once it is complete - that is the responsibility of each individual test. (On reflection I realised this would also have been a problem with your initial 'autouse' solution for the parser fixture - it would mask any problems caused by us failing to initialise the parser within PSyclone itself.)

That was my original thought. But when I then verified (again after creating the fixture etc) that the test will correctly trigger the bug if the fix in gocean is removed, **all** follow-up tests failed. I admit I didn't investigate this any further - we need to do something after this test to avoid triggering false failures in other tests if this test fails, and just re-initialising seems the easiest option.


> Also, although I appreciate that you've gone to the trouble of adding a whole new test file, I think that actually the test belongs with those for goloop - again see the comment.
> Once those things are done then this is good to go.

Yes ... and no. I thought as well originally to leave this in gocean somewhere. But there is other code in psyclone that imports similar 'sub-parser' classes from fparser, e.g. `f2pygen` imports `abs_name` (I didn't check any further what it exactly does, and given the name it will likely initialise fparser :) ). Especially with more and more code moving into the config file I would expect that similar code might be used to check for validity.

Having written all of this ... I don't actually mind, I am happy to move it, and if other tests will be required at some time in the future, we can dig out the original patch :)

src/psyclone/tests/psyir/frontend/fparser2_not_init_test.py
```diff
@@ -0,0 +1,73 @@
...
+    At the end of the tests re-initialse parser.
+    '''
+
+    # Remove all fparser classes and functions
+    ParserFactory()._setup([])
+
+    # Now execute all tests
+    yield
+
+    # We need to properly initialise fparser,
```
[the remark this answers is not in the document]
hiker: Moved ... more comments re the other failing tests see below.

hiker:
I had a closer look at the test failures, especially the final re-initialisation of fparser in the fixture. Here is what happens:

If a test is run that uses the parser fixture before the goloop test, this first test will initialise fparser, then the goloop-test removes all classes of fparsers, and all further tests using fparser fail (even when they correctly request the parser fixture). This is because pytest initialises the object of a session fixture like parser the first time it is **used**. 

If you don't run a test that uses parser before this goloop test, all these tests are successful: the goloop-test will initialise an 'empty' fparser, but the bug fix in goloop will re-initialise fparser with the right classes. Then everything works (with or without re-initialising fparser in the goloop fixture).

Calling re-init will also result in the expected behaviour if the bug should re-occur in goloop: the goloop test will remove all parser classes from fparser, triggering the error, raising an exception and causing the test to fail. But then the fixture will re-initiliase fparser with all required classes, and all work as expected, i.e. it only triggers a failure in this one test, not in any further tests.

I admit it would be nicer if we could monkey-patch this, since we then would have a clear per-test behaviour, but to be very honest, I have no idea how to do this. You need to dig into the python meta-data-structures to figure out what to patch, and tbh - I consider this whole adding classes on the fly stuff to be highly confusing and it should be removed. I really don't want to spend any more time on this.

src/psyclone/tests/psyir/frontend/fparser2_not_init_test.py
hiker: Done, but I put the fixture in goloop_test. I think for a fixture which atm is only used once, it should be very close to the test where it is used (adding something to conftest imho indicates it is used more than once). This way the whole program flow is easily visible and well documented on one screen.

When we get additional tests that need this fixture, it should be moved.

arporter: OK, thanks.

arporter:
Oh yes, I was forgetting that the fparser fixture has 'session' scope. We could change it to be 'function' but I think we'll stick with your solution.

arporter (approved):
No documentation update needed.

PR #1284, opened 2021-06-11
(closes #1266) add an adjoint loop transformation.

### LINKED ISSUE #1266: Add transformation to take adjoint of a loop.
At its simplest, the adjoint of a loop is that same loop done in reverse. In fact, if each loop iteration is independent then reversing it is not necessary but that is a future optimisation. The plan is to have separate transformations for the loop itself and the statements within a loop.

arporter:


arporter:
Ready for a first look now.

arporter:
Shall I just close this one @rupertford ?

rupertford:
> Shall I just close this one @rupertford ?

It's up to you. Might be worth keeping it open until we look at adding loops to the visitor in case there is anything to re-use?

rupertford:
Now that PR #1492 is in progress I think we can remove this PR, branch and associated issue if you agree @arporter

arporter:
Yep, I agree.

Issue #1268, opened 2021-05-28
Create separate class to store list of lists of indices in AccessInfo


hiker:
In #1028 handling of structured types is added, and this uses a list of lists of indices. It would be clearer if this is stored in a separate class, which allows:
- [x] easier access to more meaning full error messages (#1028 only prints the two indices, which is not that useful for users)
- [x] would potentially allow iterators to be used to remove some of the double nested loops (outer loop over components, inner loop over all indices).
- [x] Provide more intuitive constructor for scalars (--> [[]]) and non-structure arrays ([i, j] --> [ [I,j] ])

hiker:
Note that I've decided to leave the interface for ``add_location`` the way it is (providing a ComponentIndices class can be done, but it still accepts None, a list, or list of lists). The reason for this is that given a signature like ``field%whole%something(i,j)`` would require a not exactly user-friendly set of parameters(``[i,j]`` would not work, since there is no indication to which component it should be applied).

By leaving the 'list' argument to the ``add_location`` call, this function knows the signature, and can automatically add the right number of [] for the components - e.g. in the example above providing the indices ``[i, j]`` will create a ComponentIndices objects with `[[], [], [I,j]]`.

Issue #1220, opened 2021-04-23
Use doctest to validate example code in docstrings (esp. for Transformations)?


arporter:
I've been vaguely aware of doctest (https://docs.python.org/3/library/doctest.html) for a while but have just had a play with it and think it could be useful for us to make sure that our docstrings that contain example code (particularly in the various Transformation classes) are kept up-to-date. Usage is:

    $ python -m doctest -v domain/nemo/transformations/create_nemo_invoke_schedule_trans.py

or Sphinx provides the `doctest` target (but I'm not entirely sure which docstrings are in scope when you do that).

I've been playing with it in #1217 and it seems good. (The only problem I have is that the `Node.view()` method always produces coloured output and that causes the text comparison done by doctest to fail. We could add an optional `colour` argument to that method though.)

What do @rupertford, @sergisiso, @TeranIvy and @hiker think?

rupertford:
I've just had a play and like it.

I was worried about any code snippets that are not executable but we can presumably not add `>>>` to those - but if so will they be picked up by Sphinx?

Also, presumably you were thinking of adding it as part of the CI?

For info: it took me a while to work out that continuation lines should have "... "!

arporter:
Yes, I forgot to mention the continuation lines - it took me quite a while to get it working too!

For Sphinx, we can always tell it that code snippets are Python (but I think that's its default anyway) - it doesn't need the `>>>`.

I think it makes sense to do it as part of the CI - it shouldn't add much to the run time.

The only snag is that it makes the docstrings very verbose but then, they are almost certainly a lot more useful if the code fragments they show are actually correct :-)

hiker:
My only concern would be that the examples need to become a lot more elaborate since we then would have to have fully working code, not just a few lines to show the important lines. And if we would require working examples, it would also mean more work for us ... but:

I am not exactly sure if I ever added a 'real' example, I add them mostly to the manual, and I admit that's not ideal, since that code is not tested :( 

Would there be any chance to combine this? I vaguely remember that we can refer to py files in the documentation, including line numbers to be included. That might be nice.

arporter:
Yes, that is the main drawback although, as you say, we only need to go this route in a few places. I believe doctest can do its thang on pretty much any file - it just looks for '>>>' - so we could use it to test the manual. I'm not sure what you mean by 'combin this'? It might be that that's what Sphinx's `doctest` build target does. I shall have a google.

arporter:
[...]

PR #1318, opened 2021-06-29
#1291 Created new chapter for dependency analysis.

### LINKED ISSUE #1291: Finding dependence analysis in the dev guide
I keep getting confused when it comes to looking for the documentation on Joerg's dependence analysis functionality.
Even this morning (admittedly, pre-coffee), I came to the docs, searched for 'dependence analysis' and found the 'old', loop-based functionality. For some reason my brain does not spot that 'Variable Accesses' is what I want. I suggest that we perhaps merge the docs for the old and new way of doing things into a single Dependence Analysis section with appropriate sub sections.

hiker:
Fixes #1291.


doc/developer_guide/dependency.rst
```diff
@@ -0,0 +1,458 @@
...
+.. code-block:: python
+
+  for index_group in access_info.component_indices:
+      for index_expression in index_group:
+        # Create an access info object to collect the accesses
+        # in the index expression
+          accesses = VariablesAccessInfo()
+          index_expression.reference_accesses(accesses)
+          # Then test if the index variable is used. Note that
+          # the key of `access` is a signature
```
arporter, line 305: s/access/accesses/

doc/developer_guide/dependency.rst
```diff
@@ -0,0 +1,458 @@
...
+about one reference. The outer loop is over the indices used for each component,
+i.e. in case of `a(i)%b(j,k)%c(l)` it would loop over the three
+members of the list `[ [i], [j,k], [l] ]`. The inner loop then
+tests each index used, e.g. it would be `[i]` for the first index group,
+then `[j, k]` etc. Then the function `reference_accesses` is used to
+analyse the index expression.
+
+.. code-block:: python
+
+  for index_group in access_info.component_indices:
```
arporter, line 298: Thinking about #1220, how much more code would be required here to make this example executable? e.g.

    psyir = fortran_reader.psyir_from_source("program my_prog\n....end program my_prog\n")
    ...

doc/developer_guide/dependency.rst
```diff
@@ -0,0 +1,458 @@
...
+
+Access Examples
+---------------
+
+Below we show a simple example of how to use this API. This is from the
+`psyclone.psyGen.OMPParallelDirective` (so `self` is an instance of this
+node), and this code is used to determine a list of all the scalar
+variables that must be declared as thread-private::
+
+  var_accesses = VariablesAccessInfo()
```
arporter, line 365: Can this be made executable (i.e. testable)?

doc/developer_guide/dependency.rst
```diff
@@ -0,0 +1,458 @@
...
+      if len(accesses) == 1:
+          continue
+
+      # We have at least two accesses. If the first one is a write,
+      # assume the variable should be private:
+      if accesses[0].access_type == AccessType.WRITE:
+          result.add(var_name.lower())
+
+The next, hypothetical example shows how the `VariablesAccessInfo` class
+can be used iteratively. Assume that you have a function that determines
```
arporter, line 394: I think we do actually have such a function now?

doc/developer_guide/dependency.rst
```diff
@@ -0,0 +1,458 @@
...
+if the given variable accesses can be parallelised, and the aim is to
+determine the largest consecutive block of statements that can be
+executed in parallel. The accesses of one statement at a time are added
+until we find accesses that would prevent parallelisation::
+
+   # Create an empty instance to store accesses
```
arporter, line 400: Can this be made testable?

doc/developer_guide/dependency.rst
```diff
@@ -0,0 +1,458 @@
...
+    are part of the same PSyIR node. An expression like `i+k+1` is stored as
+    `(i+k)+1`, so if it is compared with `i+1+k` they are not considered to
+    be equal, because `i+1` and `i+k` are not the same.
+
+
+An example of how to use this class is shown below. It takes a list of statements
+(i.e. nodes in the PSyIR), and adds 'OMP DO' directives around loops that
+can be parallelised::
+
+  parallel_loop = OMPLoopTrans()
```
arporter, line 437: Can this be made testable?

arporter (changes requested):
Looks good Joerg, thanks very much for doing this.
Just a few minor comments. Main thing I'm wondering about is whether any of the code fragments in the documentation you've moved can be made testable (#1220).

hiker:
@arporter I was actually thinking of going the other direction, including code from PSyclone/examples/tests - though as far as I can see that would be vulnerable to line number changes, so not a good idea.

Just before I study how doctests work:
What is the goal of doctest rst files: just to make sure the examples in there are syntactically valid, or should they actually do something useful (the latter would be a lot more work - for the first case we could somehow create empty lists or so).

arporter:
> What is the goal of doctest rst files: just to make sure the examples in there are syntactically valid, or should they actually do something useful (the latter would be a lot more work - for the first case we could somehow create empty lists or so).

It's purely to make sure that they are syntactically valid (and remain that way ;-)). They don't have to do anything useful. As I say, if it's too much work to achieve this for a given code fragment then don't worry. In that case, your suggestion of linking to the code sounds good - so long as we use Sphinx's automethod/function or something then it should be robust. At least we'll get a warning/error if the function referred to disappears :-)

hiker:
@arporter I had a look, and my understanding is that there are two ways: one using the sphinx target `make doctest`, the other using `python -m doctest`. The first one supports 'testsetup', the second one doesn't - but it could be make to to work if we add the setup to the actual test (though that's a bit ugly imho).

Which of the two ways do you want to go? I assume (based on other tests) using `make doctest` (and use 'testsetup'?)

The one question I have with this: can we avoid adding the ">>>" before each line? I think the most useful part of my code examples is that they can be copied and pasted to get started with scripts, but the '>>>' makes this more difficult.

The alternative of this would be to add a sample script to the `code_snipets` directory (which we already have), and inline the code from that example (I can't really inline code from our generic examples or code since I need to specify the exact line numbers :( ). Then we could add some tests to the `code_snipets` directory.

arporter:
I *think* I want to go the Sphinx route. It appears to have support for both `>>>` and a `.. testcode::` markup. In the latter you don't need the ">>>" so you could still cut-n-paste the code out. (See https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html)
We could always do the `python -m doctest` thing on our source files (as opposed to the documentation) separately.

hiker:
Ah, `testcode` - that seems to work, and now I even have syntax highlighting (which I had not before). I can initialise the objects I need in the testsetup section, and only show the part required, that works just fine.
The only somewhat minor issue is I can't use `return True` - otherwise python complains that this is a return outside a function. I'll just put this in a comment, but then need a `break` (or `pass`)`:
```
    for ...
            if Signature(index_variable) in accesses:
                 # The index variable is used as an index somewhere
                 break
    else:
        print("Not used")
```
And `testoutput` then even checks that nothing is printed for my empty example - looks good!

Thanks! I'll try to convert all the examples appropriately.

hiker:
I will postpone this PR till #1313 (#1268) is on master, since this will create significant conflicts if I start modifying the examples here now, which are mostly completely rewritten in that PR.

doc/developer_guide/dependency.rst
hiker: Done.

doc/developer_guide/dependency.rst
hiker: Well, yes, and we always had. This is just a coding example that can be used as a kind of template on how to use the access information.

PR #1323, opened 2021-07-02
(closes #1305) Capture default visibility of symbols in a Container (Fortran module)

### LINKED ISSUE #1305: [psyir] Scoping of symbols in imported containers not preserved
NEMO uses the following idiom fairly frequently:

    module my_mod
      use some_other_mod
      private
      integer :: some_internal_flag
      integer, public :: my_useful_flag
    contains
      ..
    end module my_mod

Unfortunately, the code that deals with symbol visibility (which I wrote) thinks it can do away with the `private` statement and just declares the accessibility of each symbol explicitly. This means that we end up generating:

    module my_mod
      use some_other_mod
      integer, private :: some_internal_flag
      integer, public :: my_useful_flag
    contains
      ..
    end module my_mod

This then means that any code that does a wildcard import from `my_mod` will now also import all the public symbols from `some_other_mod`. This is incorrect.

arporter:


arporter:
This is quite a small PR that just affects a few test files. I took the opportunity of improving the coverage given by the tests in symboltable_test while I was there.
One for @rupertford or @sergisiso again :-)

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2014,7 +2014,7 @@ def process_declarations(self, parent, nodes, arg_list,
                 default_visibility = Symbol.Visibility.PUBLIC
```
sergisiso, line 2014: Can the `parent.symbol_table.default_visibility` be updated with this new info now?

If it can, maybe some more functions (that are below this point) can look at `parent.symbol_table.default_visibility` instead of passing it by argument.

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -65,7 +65,7 @@ class SymbolTable(object):
     :raises TypeError: if node argument is not a Schedule or a Container.
 
     '''
-    def __init__(self, node=None):
+    def __init__(self, node=None, default_visibility=None):
```
sergisiso, line 68: Missing docstring

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -83,6 +83,39 @@ def __init__(self, node=None):
                 "Schedule or a Container but found '{0}'."
                 "".format(type(node).__name__))
         self._node = node
+        # The default visibility of symbols in this symbol table. (Required
+        # when we have symbols of UnknownType since their associated
+        # declaration may or may not contain visibility information.)
+        self._default_visibility = None
```
sergisiso, line 89: I would prefer that we gave a default value to it (the same that Fortran does when nothing is specified). Instead of having the possibility for it to be None, because then we technically check that is "not None" every time we use it.

Document the default value in the docstring

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -83,6 +83,39 @@ def __init__(self, node=None):
...
+        :returns: the default visibility of symbols in this table.
+        :rtype: :py:class:`psyclone.psyir.symbols.Symbol.Visibility`
+        '''
+        return self._default_visibility
+
+    @default_visibility.setter
+    def default_visibility(self, vis):
+        '''
+        Sets the default visibility of symbols in this table. (Only used for
+        Symbols of UnknownType since their visibility may or may not be
```
sergisiso, line 105: I would leave "Sets the default visibility parameter of this table." and don't mention how this is used externally to the class.

sergisiso (changes requested):
@arporter I made a first review of this PR. If I understand this right the symbol_table default visibility is just a parameter that we happen to store in the Symbol Table (because it is where is makes more sense) but the symbol table itself doesn't use this information. It's the user that has to fetch and provide the information to the symbol on its creation. I was thinking if the ".add" or ".new_symbol" method could do this automatically, but this comes with another set of problems.

I am fine with this implementation in this PR but if the visibility keeps getting more complicated we can think about other ways to implement it. One thought was if the symbols themselfs don't need to have the visibility information at all and its the symbol_table that could manage the visibility of all its contained symbols? But as I said, this is for future discussion, not this PR.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -755,6 +783,9 @@ def container_node(self, node):
         # not allow argument declarations.
         declarations = self.gen_decls(node.symbol_table, args_allowed=False)
 
+        # Generate the access statement (PRIVATE or PUBLIC)
+        declarations += self._nindent + self.gen_access_stmt(node.symbol_table)
```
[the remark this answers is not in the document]
arporter: Oh yes, fixed and tested.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -522,6 +522,34 @@ def gen_typedecl(self, symbol):
...
+    def gen_access_stmt(self, symbol_table):
+        '''
+        Generates the access statement for a module - either "private" or
+        "public". Although the PSyIR captures the visibility of every Symbol
+        explicitly, this information is required in order
+        to ensure the correct visibility of symbols that have been imported
+        into the current module from another one using a wildcard import
+        (i.e. a `use` without an `only` clause) and also for those Symbols
+        that are of UnknownFortranType (because their declaration may or may
+        not include visibility information).
```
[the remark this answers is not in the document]
arporter: No, this remains unchanged. I did start to look at capturing visibility information for symbols of UnknownFortranType but then realised that I still needed the default visibility to cope with wildcard imports anyway.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -522,6 +522,34 @@ def gen_typedecl(self, symbol):
         result += "{0}end type {1}\n".format(self._nindent, symbol.name)
         return result
 
+    def gen_access_stmt(self, symbol_table):
```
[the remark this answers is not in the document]
arporter: Yep, fixed now :-)

src/psyclone/psyir/frontend/fparser2.py
arporter: Oh yes, I've got rid of a few of those arguments now.

src/psyclone/psyir/symbols/symboltable.py
arporter: Fixed.

src/psyclone/psyir/symbols/symboltable.py
arporter: Yes, that is better. However, that made me revisit what I'd done in `process_declarations()` and I decided that the current implementation (which had side effects for the symbol table of the supplied PSyIR node) wasn't very nice. I've therefore removed the call to `_parse_access_statements` from inside it and added it, where necessary, before every location in the code where `process_declarations()` is called. Due to this 'promotion' I've also renamed `_parse_access_statements` to `process_access_statements`.

src/psyclone/psyir/symbols/symboltable.py
arporter: Yes, done.

arporter:
CI permitting, this is ready for another look now.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -700,14 +700,15 @@ def test_process_declarations_accessibility():
     reader = FortranStringReader("private :: x\n"
                                  "real :: x\n")
     fparser2spec = Specification_Part(reader).content
-    processor.process_declarations(sched, fparser2spec, [])
+    _, vis_map = processor.process_access_statements(fparser2spec)
+    processor.process_declarations(sched, fparser2spec, [], vis_map)
```
sergisiso, line 704: It seems the test docstring is outdated as mappings are provided in 2 out of the 3 process_declaration calls. Can you update the docstring.

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -731,14 +732,14 @@ def test_process_unsupported_declarations(fortran_reader):
     # with unknown type.
     reader = FortranStringReader("real:: a = 1.1")
     fparser2spec = Specification_Part(reader).content[0]
-    processor.process_declarations(fake_parent, [fparser2spec], [])
+    processor.process_declarations(fake_parent, [fparser2spec], [], {})
```
sergisiso, line 735: You have added a `{}` argument in all the `process_declarations` from here to the end of this file? Is this on purpose or is it the remains of an intermediate change, as it seems unnecessary because it is already the optional default value.

src/psyclone/psyir/symbols/symboltable.py
sergisiso: I like how this turned out, having the `process_access_statements` as a previous step to `process_declarations` makes things more clear in my view.

For the symbol table constructor, since this is not setting a mutable container, you could encode the default value directly into the function definition: ` def __init__(self, node=None, default_visibility=Symbol.Visibility.PUBLIC):`. I think it makes for a bit better documentation.
And for the docstring of this parameter I suggest something like: "optional default visibility value for this symbol table, if not provided it defaults to PUBLIC visibility."

sergisiso (changes requested):
@arporter Almost there now, there is a reply to a previous conversation and two more comments.

src/psyclone/psyir/symbols/symboltable.py
arporter: Oh yes, that's nicer - thanks.

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Oops, I got a bit gung-ho when making all the tests work.

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Argh. It is the remains of an intermediate change. I thought I'd gone through and fixed all of them, sorry.

arporter:

sergisiso (approved):
@arporter All comments have been addressed, tests and examples pass with compilation, no pycodestyle and pylint issues introduced, no documentation is needed, this is approved to be merged.

PR #1365, opened 2021-08-02
tl to adj visitor (closes 1357)

rupertford:
Creates a visitor that works with filecontainer, schedule, loop and assignment, and generates adjoint code. Also integrates this into the psyad script.


rupertford:
Now that #1457 is in progress, all of the functionality added in this PR has been added, or is being added in separate PRs. This PR can now be closed.

Issue #1367, opened 2021-08-04
Crash in lower_to_language when a real constant is used as argument in gocean


hiker:
The following code crashes when calling lower_to_language:
```
    call invoke (init_field(a_fld, 1.0))
```
It does work when the `1.0` is replaced with `1` (but then the compiler complains about passing an integer where a real is expected).
The reason seems to be:
```
        if six.text_type(self.name).isnumeric():
            return Literal(self.name, INTEGER_TYPE)
```
`isnumeric`  (and the rest of this code ;) ) only accepts integer.

I will likely fix this as part of #1288.

PR #1369, opened 2021-08-06
1367 support real constants arguments in gocean

hiker:
This is a small fix for #1367.

Note that it would be quite easy to support double precision values (`float` does not support 'd', so we would have to replace all 'd's with 'e' in the string before calling `float`, then extend the `Literal` re to support 'dD')  - but I wasn't sure of any other implications (would double need a new intrinsic type, ...).

Assigning to @sergisiso , since I assume the changes (and future plans for) `Literal` are in his court :)

### DIFF
--- a/src/psyclone/gocean1p0.py
+++ b/src/psyclone/gocean1p0.py
@@ -61,9 +61,9 @@ from psyclone.errors import GenerationError, InternalError
 import psyclone.expression as expr
 from psyclone.f2pygen import CallGen, DeclGen, AssignGen, CommentGen, \
     IfThenGen, UseGen, ModuleGen, SubroutineGen, TypeDeclGen, PSyIRGen
+from psyclone.parse.algorithm import Arg
 from psyclone.parse.kernel import Descriptor, KernelType
 from psyclone.parse.utils import ParseError
-from psyclone.parse.algorithm import Arg
 from psyclone.psyGen import PSy, Invokes, Invoke, InvokeSchedule, \
     CodedKern, Arguments, Argument, KernelArgument, args_filter, \
     AccessType, HaloExchange
@@ -76,7 +76,7 @@ from psyclone.psyir.nodes import Loop, Literal, Schedule, Node, \
 from psyclone.psyir.symbols import SymbolTable, ScalarType, ArrayType, \
     INTEGER_TYPE, DataSymbol, ArgumentInterface, RoutineSymbol, \
     ContainerSymbol, DeferredType, DataTypeSymbol, UnresolvedInterface, \
-    UnknownFortranType, LocalInterface, BOOLEAN_TYPE
+    UnknownFortranType, LocalInterface, BOOLEAN_TYPE, REAL_TYPE
 
 
 # Specify which OpenCL command queue to use for management operations like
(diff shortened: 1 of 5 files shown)


src/psyclone/gocean1p0.py
```diff
@@ -2535,6 +2535,13 @@ def psyir_expression(self):
         if six.text_type(self.name).isnumeric():
             return Literal(self.name, INTEGER_TYPE)
 
+        # Now try for a real value:
+        try:
+            float(self.name)
```
sergisiso, line 2540: This is a good idea but since I am not sure we can guarantee that python floats syntax always match with PSyIR syntax, could we maybe use `if re.match(Literal._real_value, self.name):`?

src/psyclone/psyir/nodes/literal.py
```diff
@@ -74,7 +74,7 @@ class Literal(DataNode):
     _children_valid_format = "<LeafNode>"
     _text_name = "Literal"
     _colour = "yellow"
-    _real_value = r'^[+-]?[0-9]+(\.[0-9]*)?(e[+-]?[0-9]+)?$'
+    _real_value = r'^[+-]?[0-9]+(\.[0-9]*)?([eE][+-]?[0-9]+)?$'
```
sergisiso, line 77: Now that I notice it, I believe line 105 should also be a `re.match` instead of a `re.search`

sergisiso: Also, can you double check if in C/C++ an upper case E is valid for scientific notation. If it does no problem, otherwise we have to account for this in the backend, or revert the change and do a self.name.lower() in the psyir_expression check.

sergisiso (changes requested):
@hiker The fix looks good, I just added some comments to make sure we convert between the Python-Fortran-C syntaxes without issues.

src/psyclone/gocean1p0.py
hiker: Even better - I just realised that we can remove the use of `float`, since the constructor for `Literal` will raise an exception. I've documented this, it is indeed nicer.

src/psyclone/psyir/nodes/literal.py
hiker: Yes, confirmed with gcc and icc, and https://docs.microsoft.com/en-us/cpp/c-language/c-floating-point-constants?view=msvc-160, which specifies lower and upper case for `e` 😄

hiker:
All fixed, I won't touch the 'd' issue, I don't want to deal with C/C++ ('f' as suffix for single precision, introducing a new intrinsic type ... I guess, ...). This is good enough for my examples (Compiler doesn't accept integer constants used in a subroutine call when a double precision is expected, but it accepts single precision).

sergisiso (approved):
@hiker All comments addressed. Test, examples, pycodestyle, pylint and TODOs all look good. This fixes a bug and it doesn't need additional documentation.

sergisiso (approved):
@hiker All comments addressed. Test, examples, pycodestyle, pylint and TODOs all look good. This fixes a bug and it doesn't need additional documentation.

PR #1403, opened 2021-09-21
Add private attribute to DeclGen

arporter:
This is a small PR that just updates f2pygen to support adding the 'private' attribute to a declaration. This is required in #1147 so that we can ensure that symbols declared as part of profiling instrumentation remain private to the module they are in.


arporter:
Ready for review from anyone!

src/psyclone/f2pygen.py
```diff
@@ -1042,6 +1045,7 @@ class DeclGen(BaseDeclGen):
     :param initial_values: Initial value to give each variable.
     :type initial_values: list of str with same no. of elements as \
                           entity_decls.
+    :param bool private: whether this declaration has the PRIVATE attribute.
```
rupertford, line 1048: Seeing as you're on a declaration tidying spree, could you please get rid of the capital in "Initial" on line 1045 and  remove the full stop at the end of the type declarations.

rupertford: Could you also say, defaults to False as you do for TypeDecl.

src/psyclone/f2pygen.py
```diff
@@ -906,10 +906,11 @@ class BaseDeclGen(BaseGen):
                           DIMENSION(xx))
     :param bool allocatable: whether this declaration is for an \
                              ALLOCATABLE quantity
-    :param bool save: whether this declaration has the SAVE attribute
-    :param bool target: whether this declaration has the TARGET attribute
+    :param bool save: whether this declaration has the SAVE attribute.
+    :param bool target: whether this declaration has the TARGET attribute.
     :param initial_values: Initial value to give each variable.
     :type initial_values: list of str with same no. of elements as entity_decls
+    :param bool private: whether this declaration has the PRIVATE attribute.
```
rupertford, line 913: Could you add 'defaults to False'. I realise the other descriptions don't have this (but they should).

rupertford (changes requested):
Just a few minor docstring changes

src/psyclone/f2pygen.py
arporter: Done and done.

src/psyclone/f2pygen.py
arporter: Done.

arporter:
Ready for another look, subject to CI happiness.

PR #1412, opened 2021-09-24
Add visitor to psyad script and call the assignment transformation (closes #1411)

### LINKED ISSUE #1411: Add visitor to psyad script and call the assignment transformation
Once #1401 is on master we can add a visitor for the tangent linear code. We need to wait as the visitor will need to know which variables are active. In the first instance this visitor should simply call the assignment transformation when it finds an assignment with active variables. This is the second issue that is split from pr #1365 (the first being #1400).

rupertford:
Makes use of code first developed in pr #1365


rupertford:

Issue #1338, opened 2021-07-12
[Transformations] Add OpenMP tasking transformations.


LonelyCat124:
As part of Excalibur we need to add the ability to generate OpenMP task code.

I think there are a few likely requirements to have this working, some of which I'm happy to at least look at.

I discussed with Rupert as a first idea to create a transformation for the `OMP TASKLOOP` directive. This is probably almost a drop in replacement for omp parallel do, except I expect would need the addition of a parallel node and a taskloop node.

The one to use explicit tasks is more complex. The nemolite2d code blocks the outer loops:
```
#define TASK_SIZE some_integer_value

for(jj = internal_start; jj <= internal_stop; j+= TASK_SIZE){
  int end = min(jj+TASK_SIZE, internal_stop)
  --TASK PRAGMA HERE, ff is firstprivate
  for(j=jj; j <= end; j++){
  --Inner loop(s) and kernel(s).
  }
}
```

I will hopefully have some time to look at the taskloop directive soon.

LonelyCat124:
As a question - The OMPLoopTrans is separate from the code that generates the OpenMP parallel code?

If so, I think I should be able to create OMPTaskloopTrans as a drop-in replacement for OMPLoopTrans as a starting point to see how these things fit together.

arporter:
I think the answer to your question is "yes" - it is separate. However, there's a check in the code-generation that the resulting OMP directive is within a parallel region.

LonelyCat124:
Ok - I'm going to work through the documentation to work out how things fit together tomorrow and go from there.

LonelyCat124:
Ok - I realised I need another thing that doesn't yet exist - which is OpenMP's single/master directives, as well as `taskwait`.

From what I can work out/understand it looks like I could create as many transformations as needed to be applied inside a code, and then just do them in innermost first order? So for example
``` 
    from psyclone.parse.algorithm import parse
    from psyclone.parse.utils import ParseError
    from psyclone.psyGen import PSyFactory
    from psyclone.errors import GenerationError
    api = "gocean1.0"
    filename = "nemolite2d_alg.f90"
    ast, invokeInfo = parse(filename, api=api, invoke_name="invoke")
    psy = PSyFactory(api).create(invokeInfo)
    print psy.invokes.names
    
    from psyclone.psyGen import TransInfo
    t = TransInfo()
    ltrans = t.get_trans_name('OMPTaskloopTrans')
[...]

PR #1415, opened 2021-09-28
(Towards #1338) Implementation of an OMPTaskwaitTrans

LonelyCat124:
Initial implementation of the OMPTaskwaitTrans is implemented here.

I've fixed all the pycodestyle and pylint issues I've found other than these, which I'm not clear if I can resolve or not:
```
transformations.py:677:4: R0914: Too many local variables (19/15) (too-many-locals)
transformations.py:677:4: R0912: Too many branches (15/12) (too-many-branches)
```

I've not yet implemented any tests, nor the `{'fail_on_no_taskloops':True/False}` option that Sergi suggested.

Some of the tests are easy to implement, but others I could do with suggestions of a starting case (from a code perspective).
I need to know how to write a piece of general code with (2+, not sure exactly) loops that I can use as the basis for tests, what do I need to put together for PSyclone to handle this? @sergisiso @arporter @rupertford any ideas on what I need to put together for that?

### DIFF
--- a/src/psyclone/transformations.py
+++ b/src/psyclone/transformations.py
@@ -651,6 +651,11 @@ class OMPTaskwaitTrans(Transformation):
             forward_dep = taskloop.forward_dependence()
             if forward_dep is None:
                 continue
+            # If the forward dependence is not in the same parallel region
+            # then we don't need to worry
+            if (taskloop.ancestor(nodes.OMPParallelDirective) is not
+                    forward_dep.ancestor(nodes.OMPParallelDirective)):
+                continue
             # Check if the taskloop and its forward dependence are in the
             # same serial region
             if (taskloop.ancestor(OMPSerialDirective) is not
--- a/src/psyclone/transformations.py
+++ b/src/psyclone/transformations.py
@@ -63,7 +63,8 @@ from psyclone.psyir import nodes
 from psyclone.psyir.nodes import CodeBlock, Loop, Assignment, Schedule, \
     Directive, ACCLoopDirective, OMPDoDirective, OMPParallelDoDirective, \
     ACCDataDirective, ACCEnterDataDirective, OMPDirective, \
-    ACCKernelsDirective, OMPTaskloopDirective
+    ACCKernelsDirective, OMPTaskloopDirective, OMPSerialDirective, \
+    OMPTaskwaitDirective
 from psyclone.psyir.symbols import SymbolError, ScalarType, DeferredType, \
     INTEGER_TYPE, DataSymbol, Symbol
 from psyclone.psyir.transformations import RegionTrans, LoopTrans, \
@@ -561,6 +562,235 @@ class OMPTaskloopTrans(ParallelLoopTrans):
             # Reset the nogroup value to the original value
             self.omp_nogroup = current_nogroup
         return rval
+
+
+class OMPTaskwaitTrans(Transformation):
+
+    '''
+    Adds zero or more OpenMP Taskwait directives to an OMP parallel region.
+    This transformation will add directives to satisfy dependencies between
+    Taskloop directives without an associated taskgroup (i.e. no nogroup
+    clause). It also tries to minimise the number added to maximise available
+    parallelism.
+
+    For example:
+
+    >>> from pysclone.parse.algorithm import parse
+    >>> from psyclone.psyGen import PSyFactory
+    >>> api = "gocean1.0"
+    >>> filename = "nemolite2d_alg.f90"
+    >>> ast, invokeInfo = parse(filename, api=api, invoke_name="invoke")
+    >>> psy = PSyFactory(api).create(invokeInfo)
+    >>>
+    >>> from psyclone.transformations import OMPParallelTrans, OMPSingleTrans
+    >>> from psyclone.transformations import OMPTaskloopTrans, OMPTaskwaitTrans
+    >>> singletrans = OMPSingleTrans()
+    >>> paralleltrans = OMPParallelTrans()
+    >>> tasklooptrans = OMPTaskloopTrans()
+    >>> taskwaittrans = OMPTaskwaitTrans()
+    >>>
+    >>> schedule = psy.invokes.get('invoke_0').schedule
+    >>> schedule.view()
+    >>>
+    >>> # Apply the OpenMP Taskloop transformation to *every* loop
+    >>> # in the schedule.
+    >>> # This ignores loop dependencies. These must be manually handled
+    >>> # either through end of regions.
+    >>> # TODO: #1368 These can also be handled through the taskwait
+    >>> # directive
+    >>> for child in schedule.children:
+    >>>     tasklooptrans.apply(child, nogroup = true)
+    >>> # Enclose all of these loops within a single OpenMP
+    >>> # SINGLE region
+    >>> singletrans.apply(schedule.children)
+    >>> # Enclose all of these loops within a single OpenMP
+    >>> # PARALLEL region
+    >>> paralleltrans.apply(schedule.children)
+    >>> taskwaittrans.apply(schedule.children)
+    >>> schedule.view()
+    '''
+    def __str__(self):
+        rval = ("Adds 'OpenMP TASKWAIT' directives to a loop to satisfy "
+                "'OpenMP TASKLOOP' dependencies")
+        return rval
+
+    def validate(self, node, options=None):
+        '''
+        Validity checks for input arguments.
+
+        :param node: the PSyIR node to validate.
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+        :param options: a dictionary with options for transformations.
+        :type options: dictionary of string:values or None
+
+        :raises TransformationError: If the supplied node is not an \
+                                     OMPParallelDirective
+        :raises TransformationError: If there are no OMPTaskloopTrans nodes \
+                                     in this tree.
+        :raises TransformationError: If taskloop dependencies can't be \
+                                     satisfied due to dependencies across \
+                                     barrierless OpenMP Serial Regions.
+        '''
+        # Check the supplied node is an OMPParallelDirective
+        if not isinstance(node, nodes.OMPParallelDirective):
+            raise TransformationError("OMPTaskwaitTrans was supplied a {0} "
+                                      "node, but expected an "
+                                      "OMPParallelDirective".format(
+                                        node.__class__.__name__))
+
+        # Walk the tree to find any OMPTaskloopTrans
+        if node.walk(OMPTaskloopDirective) == []:
+            raise TransformationError("OMPTaskwaitTrans was supplied a "
+                                      "schedule containing no "
+                                      "OMPTaskloopDirectives")
+
+        # Check that all of the dependencies are satisfiable
+        taskloops = node.walk(OMPTaskloopDirective)
+
+        for taskloop in taskloops:
+            forward_dep = taskloop.forward_dependence()
+            if forward_dep is None:
+                continue
+            # Check if the taskloop and its forward dependence are in the
+            # same serial region
+            if (taskloop.ancestor(OMPSerialDirective) is not
+                    forward_dep.ancestor(OMPSerialDirective)):
+                # They're not in the same serial region. Check if our
+                # taskloop is in a waiting Single Directive
+                valid = True
+                ancestor = taskloop.ancestor(OMPSerialDirective)
+                valid = valid and isinstance(ancestor,
+                                             nodes.OMPSingleDirective)
+                if valid:
+                    valid = valid and (not ancestor.nowait())
+                # If not valid, then we're in a Master Directive or a
+                # Single Directive with nowait. In this case we can't
+                # safely guarantee our forward dependency so throw an
+                # error
+                if not valid:
+                    raise TransformationError("Couldn't satisfy the "
+                                              "dependencies due to "
+                                              "taskloop dependencies "
+                                              "across barrierless OMP "
+                                              "serial regions.")
+
+    def apply(self, node, options=None):
+        '''
+        Apply an OMPTaskwait Transformation to the supplied node
+        (which must be a OMPParallelDirective). In the generated code this
+        corresponds to adding zero or more OMPTaskwaitDirectives as
+        appropriate:
+
+        .. code-block:: fortran
+
+          !$OMP PARALLEL
+            ...
+            !$OMP TASKWAIT
+            ...
+            !$OMP TASKWAIT
+            ...
+          !$OMP END PARALLEL
+
+        :param node: the node to which to apply the transformation.
+        :type node: :py:class:`psyclone.f2pygen.DoGen`
+        :param options: a dictionary with options for transformations\
+                        and validation.
+        :type options: dictionary of string:values or None
+
+        :returns: two-tuple of transformed schedule and a record of the \
+                  transformation.
+        :rtype: (:py:class:`psyclone.psyir.nodes.Schedule, \
+                 :py:class:`psyclone.undoredo.Memento`)
+        '''
+
+        self.validate(node, options=options)
+
+        schedule = node.root
+        # create a memento of the schedule and the proposed transformation
+        keep = Memento(schedule, self, [node])
+
+        # Find all the OpenMP Single & Master regions
+        task_regions = node.walk(OMPSerialDirective)
+
+        # Loop over the task regions
+        for task_region in task_regions:
+            # Find all of the taskloops
+            taskloops = task_region.walk(OMPTaskloopDirective)
+            # Get the positions of all of the taskloops
+            taskloop_positions = [-1] * len(taskloops)
+            # Get the forward_dependence position of all of the taskloops
+            dependence_position = [None] * len(taskloops)
+            for taskloop, i in enumerate(taskloops):
+                taskloop_positions[i] = taskloop.abs_position()
+                forward_dep = taskloop.forward_dependence()
+                # Check if the taskloop and its forward dependence are in the
+                # same serial region
+                if (taskloops[i].ancestor(OMPSerialDirective) is not
+                        forward_dep.ancestor(OMPSerialDirective)):
+                    # The barrier at the end of our parent
+                    # Single Directive will handle this dependency.
+                    # The validate function already checked that this happens
+                    dependence_position[i] = None
+                else:
+                    # We're in the same OMPSerialDirective so store the
+                    # position of the forward dependency
+                    dependence_position[i] = forward_dep.abs_position()
+            # Forward dependency positions are now computed for this region.
+            # Next we eliminate unneccessary dependencies. We loop over
+            # each dependence. For each dependence we loop over all other
+            # taskloops whose position is before that dependence. If that
+            # dependence is fulfilled by this dependence (i.e.
+            # dependence_position[this] < dependence_position[that]) then
+            # we remove that dependence. If this dependence is fulfilled
+            # by that dependence (i.e. dependence_position[that] <
+            # dependence_position[this]) then we remove this dependence.
+            # This assumes that taskloop_positions is in ascending order,
+            # which I believe to be true by construction.
+            for dep_pos, i in enumerate(dependence_position):
+                if dep_pos is not None:
+                    # Grab the position of the next dependence
+                    next_dependence = dep_pos
+                    # Loop over the other taskloops
+                    for j in range(i+1, len(dependence_position)):
+                        # If this taskloop happens after the next_dependence
+                        # then we can just stop.
+                        if taskloop_positions[j] >= next_dependence:
+                            break
+                        # If the jth taskloop has no dependency then continue
+                        if dependence_position[j] is None:
+                            continue
+                        # Check if next_dependence will satisfy the jth
+                        # taskloops dependency
+                        if next_dependence <= dependence_position[j]:
+                            dependence_position[j] = None
+                        # Check if the jth taskloops dependence will satisfy
+                        # the next_dependence
+                        if dependence_position[j] < next_dependence:
+                            dependence_position[i] = None
+                            # If it does then we can move to the next taskloop
+                            break
+            # dependence_position now contains only the required dependencies
+            # to satisfy the full superset of dependencies. We can loop over
+            # these by index, and if dependence_position[i] is not None then
+            # go to its forward dependency, and insert a TaskwaitDirective
+            # immediately before.
+            for dep_pos, i in enumerate(dependence_position):
+                if dep_pos is not None:
+                    forward_dep = taskloops[i].forward_dependence()
+                    fdep_parent = forward_dep.parent
+                    # Find the position of the forward_dep in its parent's
+                    # children list
+                    loc = -1
+                    for child, j in enumerate(fdep_parent.children):
+                        if child is forward_dep:
+                            loc = j
+                            break
+                    # We've found the position, so we now insert an
+                    # OMPTaskwaitDirective in that location instead
+                    fdep_parent.addchild(loc, OMPTaskwaitDirective())
+
+        # All done, return
+        return schedule, keep
 
 
 class OMPLoopTrans(ParallelLoopTrans):
--- a/src/psyclone/psyir/nodes/omp_directives.py
+++ b/src/psyclone/psyir/nodes/omp_directives.py
@@ -254,6 +254,16 @@ class OMPSingleDirective(OMPSerialDirective):
         _, position = self._find_position(self.ancestor(Routine))
         return "OMP_single_" + str(position)
 
+    @property
+    def nowait(self):
+        '''
+        Returns whether the nowait clause is specified for this directive.
+
+        :returns: whether the nowait clause is specified for this directive.
+        :rtype: bool
+        '''
+        return self._nowait
+
     def gen_code(self, parent):
         '''Generate the fortran OMP Single Directive and any associated
         code


arporter:
The easiest thing to do is write what you want in Fortran (as a string) and then use the `fortran_reader` test fixture to give you the PSyIR:

    psyir = fortran_reader.psyir_from_source(my_string)

arporter:
The Fortran will need to be a valid Subroutine (or Program/Function).

LonelyCat124:
Added some first few tests, added the `{'fail_on_no_taskloops':True/False}` option, and deleted a line from `forward_dependence` in `nodes.py` which will definitely cause no issues at all...

LonelyCat124:
Concerns me a little that tests pass xD. Will finish a draft of this off tomorrow am

LonelyCat124:
I'm going to tentatively mark this as ready to review, though I will also leave it as draft for now.

The main thing that might change is the use of `forward_dependence` may change to use @hiker 's new dependence analysis system if I can work out an appropriate way to implement a "forward_dependence" style function for it.

I think I've tested simplified versions of all possible dependency structures for taskloops that I can think of for now (except for non-parallelised loops I guess), so if anyone wants to try to break it as well it would be good to have more tests there.

LonelyCat124:
Removing ready to review tag until the new version is pushed!

LonelyCat124:
New version is completed. If the review wants I can try to resolve the linter issues:
```
src/psyclone/transformations.py:692:4: R0912: Too many branches (17/12) (too-many-branches)
src/psyclone/transformations.py:797:4: R0914: Too many local variables (19/15) (too-many-locals)
src/psyclone/transformations.py:797:4: R0912: Too many branches (18/12) (too-many-branches)
```
by refactoring after review but I'm not sure if it would make the code clearer or not.

LonelyCat124:
From discussion with Tobias it seems like:
```
!$omp parallel
!$omp single
!$omp taskloop nogroup
do i=1, n
  a(i) = 2 * i
end do
!$omp end taskloop
!$omp end single
!$omp do
do i=1,n
  b(i) = a(i) * 3
end do
!$omp end do
!$omp end parallel
```

Is not valid unless I add a `taskwait` at the end of the single. I will implement this now if I can

LonelyCat124:
OpenMP does define: 
> task synchronization construct: A taskwait, taskgroup, or a barrier construct.

However I'm going to assume that implicit barriers do not count for this.

LonelyCat124:
I think there are still some over sensitivities in my `get_forward_dependence`, I will continue to improve it.

LonelyCat124:
I have removed the over sensitivities - i cannot work out why the examples are suddenly failing to build though (they work fine locally)

LonelyCat124:
I think coverage is 100% on this diff (from checking locally with 3.6), so this is ready to review.

LonelyCat124:
Not sure where to ask this, but can I check:
Does the `LoopTrans` class's validate (which i can't find) or the `Loop` class require the Loop to meet OpenMP's canonical loop form (https://www.openmp.org/spec-html/5.0/openmpsu40.html#x63-1260002.9.1)? The main thing I can't think of being covered when I looked last week is that I couldn't see is that I don't think any of them check `incr-expr` (C/C++) or `incr` (Fortran).
I think in theory PSyclone could create invalid code here? @sergisiso what do you think, should we make an issue for this?

sergisiso:
LoopTrans is in `src/psyclone/psyir/transformations/loop_trans.py`

I wasn't aware of this restrictions and you are right that this could be validated. Maybe subclassing LoopTrans with OMPLoopTrans and adding this additional restrictions? Which if I read correctly for Fortran are:
- Only BinaryOperation.[ADD, SUB, MUL] + Literals + References are valid in in the incr
- Loop invariant with respect to the outermost associated loop. As discussed the other day I think for a single loop in Fortran, the incr is loop invariant by definition, but I don't think this is true for nested loops then and could be checked in the validate.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        # Find our parent serial region if it has a barrier
+        parent_single = taskloop.ancestor(OMPSingleDirective)
+        if parent_single is not None and parent_single.nowait:
+            parent_single = None
+        # Find our parent parallel region
+        parent_parallel = taskloop.ancestor(OMPParallelDirective)
+        for node in node_list:
+            if node.abs_position <= taskloop.abs_position:
+                continue
+            # Ignore our children
```
arporter, line 733: s/our children/any descendents of the taskloop directive/

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+            if anc is taskloop:
+                continue
+            node_vars = None
+            # Special case:
+            # 1. If we have a different parent parallel node, then our parent
+            # parallel node is our dependency
+            if node.ancestor(OMPParallelDirective) is not parent_parallel:
+                return parent_parallel
+
+            if (isinstance(node, OMPTaskwaitDirective) and
```
arporter, line 744: Bearing in mind that going up and down the tree can get expensive, please could you cache some of these results (e.g. `taskloop.ancestor(OMPSingleDirective)`) so that we find it just once and then re-use it. It might also help make this `if` a bit shorter :-)

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                    taskloop.ancestor(OMPSingleDirective) is not None and
+                    taskloop.ancestor(OMPSingleDirective).nowait is False):
+                # If we find a taskwait barrier inside the same
+                # OMPSingleDirective and that OMPSingleDirective has a no
+                # nowait clause, then it acts as a barrier for dependencies
+                # as well
+                node_vars = taskloop_vars
+            elif not isinstance(node, (OMPSerialDirective,
+                                       OMPParallelDirective)):
+                # For all our other node types that cannot be our parents,
```
arporter, line 756: I don't understand this comment. Perhaps this relates to what the expected relationship between `taskloop` and `root` is? Please could you clarify and maybe add a validation step to check that things are as you expect upon entry to this routine?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                # they use their own signatures
+                node_vars = VariablesAccessInfo()
+                for child in node.walk(nodes.Node):
```
arporter, line 759: We do this walk earlier so could we re-use the results from there?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                    if child is not node and not isinstance(child,
+                                                            (Schedule, Loop)):
+                        node_vars.merge(VariablesAccessInfo(child))
+            else:
+                # If we find node types that could be our parents (are are
```
arporter, line 764: s/are are/and are/ I think? Please could you also clarify what "known to not be valid options" means?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+            node_signatures = node_vars.all_signatures
+            # Once we have the node's variable accesses, check for collisions
+            for sig1 in taskloop_signatures:
+                # If this signature is not in the node signatures then continue
+                if sig1 not in node_signatures:
+                    continue
+                access1 = taskloop_vars[sig1]
+                access2 = node_vars[sig1]
+                # If both read only, we can ignore this signature
+                # Otherwise, one of themwrites so return this node as
```
arporter, line 777: s/themwrites/them writes/

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                # we have a WaR or RaW dependency
+                if access1.is_written() or access2.is_written():
+                    # If we have a different parent serial node, and
+                    # parent_single is not None then our parent_single is our
+                    # dependency, otherwise this node is the dependency
+                    if (taskloop.ancestor(OMPSerialDirective) is not
```
arporter, line 783: Use cached result for `taskloop.ancestor(OMPSerialDirective)`?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                            parent_single is not None):
+                        return parent_single
+                    return node
+        # If we found no dependencies, then return that!
+        return None
+
+    def apply(self, node, options=None):
+        '''
+        Apply an OMPTaskwait Transformation to the supplied node
+        (which must be a OMPParallelDirective). In the generated code this
```
arporter, line 794: s/be a/be an/

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+          !$OMP PARALLEL
+            ...
+            !$OMP TASKWAIT
+            ...
+            !$OMP TASKWAIT
+            ...
+          !$OMP END PARALLEL
+
+        :param node: the node to which to apply the transformation.
+        :type node: :py:class:`psyclone.f2pygen.DoGen`
```
arporter, line 809: Must be a PSyIR node, i.e. `psyclone.psyir.nodes.OMPParallelDirective`

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        :param options: a dictionary with options for transformations\
+                        and validation.
+        :type options: dictionary of string:values or None
```
arporter, line 812: "dict of str:values or None"

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        :param bool options["fail_on_no_taskloop"]:
+                indicating whether this should throw an error if no \
+                OMPTaskloop nodes are found in this tree.
```
arporter, line 815: Please add what the default is.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                 :py:class:`psyclone.undoredo.Memento`)
+        '''
+
+        self.validate(node, options=options)
+
+        schedule = node.root
+        # create a memento of the schedule and the proposed transformation
+        keep = Memento(schedule, self, [node])
+
+        # Find all the OpenMP Single & Master regions
```
arporter, line 829: As this is new code and we've decided to drop Memento, please remove this bit and just have the routine return (None, Node) and add a TODO for #595.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+
+        :returns: two-tuple of transformed schedule and a record of the \
+                  transformation.
+        :rtype: (:py:class:`psyclone.psyir.nodes.Schedule, \
+                 :py:class:`psyclone.undoredo.Memento`)
+        '''
+
+        self.validate(node, options=options)
+
+        schedule = node.root
```
arporter, line 825: I'm not sure you want to do this or is it just here for the Memento?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                        # good way to detect that now (since there is no
+                        # "end of single region" abs_position to map to).
+                        # I could store a list of all the taskloops that
+                        # require this final taskwait, and walk from each one
+                        # to see if a taskwait has been placed to satisfy their
+                        # dependency, but that does not seem elegant somehow.
+                        if forward_dep is task_region:
+                            endwaits.append(taskloop)
+                            create_endtaskwait = True
+                        forward_dep = None
```
arporter, line 871: Could just do `continue` here to make it a bit more obvious?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+            # each dependence. For each dependence we loop over all other
+            # taskloops whose position is before that dependence. If that
+            # dependence is fulfilled by this dependence (i.e.
+            # dependence_position[this] < dependence_position[that]) then
+            # we remove that dependence. If this dependence is fulfilled
+            # by that dependence (i.e. dependence_position[that] <
+            # dependence_position[this]) then we remove this dependence.
+            # This assumes that taskloop_positions is in ascending order,
+            # which I believe to be true by construction.
+            for i, dep_pos in enumerate(dependence_position):
```
arporter, line 892: Please could you break out this elimination into a separate routine (could be a private one).

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                if dep_pos is not None:
```
arporter, line 893: Perhaps:

    if dep_pos is None:
        continue

so as to reduce the indentation of the rest of the code.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                            # If it does then we can move to the next taskloop
+                            break
+            # dependence_position now contains only the required dependencies
+            # to satisfy the full superset of dependencies. We can loop over
+            # these by index, and if dependence_position[i] is not None then
+            # go to its forward dependency, and insert a TaskwaitDirective
+            # immediately before.
+            for i, dep_pos in enumerate(dependence_position):
+                if dep_pos is not None:
+                    forward_dep = OMPTaskwaitTrans.get_forward_dependence(
```
arporter, line 923: I'm a bit confused here, don't we already have the forward dependence from earlier?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                            taskloops[i], node)
+                    fdep_parent = forward_dep.parent
+                    # Find the position of the forward_dep in its parent's
+                    # children list
```
arporter, line 927: I think you can just do `fdep_parent.children.index(forward_dep)`

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                    # Find all the taskwaits in the task_region
+                    nodes = task_region.walk(OMPTaskwaitDirective)
+                    for node1 in nodes:
+                        # If the taskwait appears after this taskloop then
+                        # its dependency is satisfied
+                        if node1.abs_position > taskloop_pos:
+                            unsatisfied = False
+                    create_endtaskwait = create_endtaskwait or unsatisfied
+                if create_endtaskwait:
+                    task_region.children[0].addchild(OMPTaskwaitDirective())
```
arporter, line 951: Should `children[0]` be `directive_body` or similar?

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                    taskloop_pos = taskloop.abs_position
+                    # Find all the taskwaits in the task_region
+                    nodes = task_region.walk(OMPTaskwaitDirective)
+                    for node1 in nodes:
+                        # If the taskwait appears after this taskloop then
+                        # its dependency is satisfied
+                        if node1.abs_position > taskloop_pos:
+                            unsatisfied = False
+                    create_endtaskwait = create_endtaskwait or unsatisfied
+                if create_endtaskwait:
```
arporter, line 950: If I've understood this, as soon as we have `unsatisfied = True` then we know we need to create an OMPTaskWaitDirective at the end of this region? That being so I think we could simplify a bit:

    if any(node1.abs_position <= taskloop_pos for node1 in nodes):
        task_region.directive_body.addchild(....)
        break

(but I may be wrong)

src/psyclone/tests/test_files/gocean1p0/reordered_dependent_invoke.f90
```diff
@@ -0,0 +1,52 @@
+!-------------------------------------------------------------------------------
```
arporter, line 1: Please could you use the current license header here.

src/psyclone/tests/test_files/gocean1p0/reordered_dependent_invoke.f90
```diff
@@ -0,0 +1,52 @@
...
+! Met Office and NERC 2014.
+! However, it has been created with the help of the GungHo Consortium,
+! whose members are identified at https://puma.nerc.ac.uk/trac/GungHo/wiki
+!-------------------------------------------------------------------------------
+! Author A. Porter STFC Daresbury Lab
+! Modified A. B. G. Chalk STFC Daresbury Lab
+
+PROGRAM dependent_invoke_test
+
+  ! Fake Fortran program for testing aspects of
```
arporter, line 12: Please extend this to say what aspect this file is testing. (Also, are you sure there isn't one that already does this?)

src/psyclone/tests/test_files/gocean1p0/reordered_compute_cu_mod.f90
```diff
@@ -0,0 +1,114 @@
+!> \brief Compute the mass flux in the x direction, cu
```
arporter, line 1: Please add the license/copyright header and extend the comment to say what this file is testing. You'll also need to remove the comment that says what the code originally looked like as that no longer matches the implementation (presumably because this is now a different kernel that doesn't exist in the original NemoLite2D code?).

.github/workflows/python-package.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.8]
```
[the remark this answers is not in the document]
LonelyCat124: Switched it back

src/psyclone/transformations.py
```diff
@@ -63,7 +64,9 @@
 from psyclone.psyir.nodes import CodeBlock, Loop, Assignment, Schedule, \
     Directive, ACCLoopDirective, OMPDoDirective, OMPParallelDoDirective, \
     ACCDataDirective, ACCEnterDataDirective, OMPDirective, \
-    ACCKernelsDirective, OMPTaskloopDirective
+    ACCKernelsDirective, OMPTaskloopDirective, OMPSerialDirective, \
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+    >>> # SINGLE region
+    >>> singletrans.apply(schedule.children)
+    >>> # Enclose all of these loops within a single OpenMP
+    >>> # PARALLEL region
+    >>> paralleltrans.apply(schedule.children)
+    >>> taskwaittrans.apply(schedule.children)
+    >>> schedule.view()
+    '''
+    def __str__(self):
+        rval = ("Adds 'OpenMP TASKWAIT' directives to a loop to satisfy "
```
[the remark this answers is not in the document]
LonelyCat124: I changed this to read correctly. I'm not sure they do, but I couldn't think of a short way to describe this transformation (Adding a single directive to a loop still fully explains what those transformations do)

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+    >>> paralleltrans = OMPParallelTrans()
+    >>> tasklooptrans = OMPTaskloopTrans()
+    >>> taskwaittrans = OMPTaskwaitTrans()
+    >>>
+    >>> schedule = psy.invokes.get('invoke_0').schedule
+    >>> schedule.view()
+    >>>
+    >>> # Apply the OpenMP Taskloop transformation to *every* loop
+    >>> # in the schedule.
+    >>> # This ignores loop dependencies. These must be manually handled
```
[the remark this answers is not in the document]
LonelyCat124: Fixed this.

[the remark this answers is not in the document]
LonelyCat124: Also updated this for tasklooptrans

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        return rval
+
+    def validate(self, node, options=None):
+        '''
+        Validity checks for input arguments.
+
+        :param node: the PSyIR node to validate.
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+        :param options: a dictionary with options for transformations.
+        :type options: dictionary of string:values or None
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+    def __str__(self):
+        rval = ("Adds 'OpenMP TASKWAIT' directives to a loop to satisfy "
+                "'OpenMP TASKLOOP' dependencies")
+        return rval
+
+    def validate(self, node, options=None):
+        '''
+        Validity checks for input arguments.
+
+        :param node: the PSyIR node to validate.
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+        :param options: a dictionary with options for transformations.
+        :type options: dictionary of string:values or None
+        :param bool options["fail_on_no_taskloop"]:
+                indicating whether this should throw an error if no \
+                OMPTaskloop nodes are found in this tree.
```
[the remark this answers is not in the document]
LonelyCat124: I've improved this.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        :raises TransformationError: If taskloop dependencies can't be \
+                                     satisfied due to dependencies across \
+                                     barrierless OpenMP Serial Regions.
+        '''
+        fail_on_no_taskloop = True
+        if options is not None:
+            fail_on_no_taskloop = options.get("fail_on_no_taskloop", True)
+        # Check the supplied node is an OMPParallelDirective
+        if not isinstance(node, nodes.OMPParallelDirective):
+            raise TransformationError("OMPTaskwaitTrans was supplied a \"{0}\""
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                                      " node, but expected an "
+                                      "OMPParallelDirective".format(
+                                        node.__class__.__name__))
+
+        # Walk the tree to find any OMPTaskloopTrans
+        if fail_on_no_taskloop and node.walk(OMPTaskloopDirective) == []:
+            raise TransformationError("OMPTaskwaitTrans was supplied a "
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                                      "schedule containing no "
+                                      "OMPTaskloopDirectives")
+
+        # Check that all of the dependencies are satisfiable
+        taskloops = node.walk(OMPTaskloopDirective)
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+
+        for taskloop in taskloops:
+            forward_dep = OMPTaskwaitTrans.get_forward_dependence(taskloop,
```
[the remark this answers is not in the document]
LonelyCat124: I've added one, I could add more detail but I'd just have to describe the get_forward_dependence function if so.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                                                                  node)
+            if forward_dep is None:
+                continue
+            # If the forward dependence is always in the same Parallel
```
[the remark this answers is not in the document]
LonelyCat124: First two lines of this comment block are for old code that I think removed, so I've removed those lines

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+            # Region or the parent Parallel region
+            # Check if the taskloop and its forward dependence are in the
+            # same serial region
+            if (taskloop.ancestor(OMPSerialDirective) is not
+                    forward_dep.ancestor(OMPSerialDirective)):
+                # They're not in the same serial region. Check if our
+                # taskloop is in a waiting Single Directive
+                valid = True
```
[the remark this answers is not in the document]
LonelyCat124: Yup, fixed!

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                ancestor = taskloop.ancestor(OMPSerialDirective)
+                valid = valid and isinstance(ancestor,
+                                             nodes.OMPSingleDirective)
+                if valid:
+                    valid = valid and (not ancestor.nowait)
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                # If not valid, then we're in a Master Directive or a
+                # Single Directive with nowait. In this case we can't
+                # safely guarantee our forward dependency so throw an
+                # error
+                if not valid:
+                    raise TransformationError("Couldn't satisfy the "
```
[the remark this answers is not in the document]
LonelyCat124: I think I've done this now.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                                              "dependencies due to "
+                                              "taskloop dependencies "
+                                              "across barrierless OMP "
+                                              "serial regions.")
+
+    @staticmethod
+    def get_forward_dependence(taskloop, root):
+        '''
+        Returns the next forward dependence for a taskloop using the new
```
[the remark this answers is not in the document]
LonelyCat124: Changed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        OMPTaskloopDirective
+        OMPTaskwaitDirective (If in same OMPSingle and that single has
+        nowait=False)
+        OMPSingleDirective (If ancestor OMPSingle has nowait=False)
+        OMPParallelDirective
+
+        The forward dependency is never a child of taskloop, and must have
+        abs_position > taskloop.abs_position
+
+        :param taskloop: the taskloop node to search for the
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+                         forward_dependence.
+        :type taskloop: :py:class:`psyclone.psyir.nodes.OMPTaskloopDirective
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        :param root: the tree in which to search for the forward_dependence.
+        :type root: :py:class:`psyclone.f2pygen.DoGen`
```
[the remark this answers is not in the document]
LonelyCat124: I don't know - I copied this from the main `get_forward_dependence` function I think. I'l lchange it for `Node` for now.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+
+        :returns: the forward_dependence of taskloop.
+        :rtype: :py:class:`psyclone.f2pygen.DoGen`
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+        '''
+        # We only look for specific types
+        node_list = root.walk((Loop, OMPDoDirective, OMPTaskloopDirective,
+                               OMPTaskwaitDirective, OMPSingleDirective,
+                               OMPParallelDirective))
+        taskloop_vars = VariablesAccessInfo()
+        for child in taskloop.walk(nodes.Node):
```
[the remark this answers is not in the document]
LonelyCat124: I will add a comment explaining this. Essentially the functionality that I need is not quite provided in `dependency_tools`. The `VariablesAccessInfo` that results from using the dependency_tools directly is the read/write accesses to the loop variable (by the loops), which means that since loop variables are probably reused by each loop, all loops point dependencies to each other. The easiest way I could think of to do this is to implement it myself, where I skip over all Loops explicitly.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+            if child is not taskloop and not isinstance(child,
+                                                        (Schedule, Loop)):
+                taskloop_vars.merge(VariablesAccessInfo(child))
+        taskloop_signatures = taskloop_vars.all_signatures
+        # Find our parent serial region if it has a barrier
+        parent_single = taskloop.ancestor(OMPSingleDirective)
+        if parent_single is not None and parent_single.nowait:
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/transformations.py
```diff
@@ -563,6 +566,394 @@ def apply(self, node, options=None):
...
+            parent_single = None
+        # Find our parent parallel region
+        parent_parallel = taskloop.ancestor(OMPParallelDirective)
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/transformations.py
LonelyCat124: Done

src/psyclone/transformations.py
LonelyCat124: I think I've cached the ones I can, unfortunately the code checks for different types of ancestor a lot :(

src/psyclone/transformations.py
LonelyCat124: Actually it is never a parent of OMPTaskloop by this check:
```
if node.abs_position <= taskloop.abs_position:
                  continue
```

LonelyCat124: This has been rewritten but hopefully clears up a lot of the confusing behaviour.

src/psyclone/transformations.py
LonelyCat124: I think we only do this for taskloop previously, not for node.

src/psyclone/transformations.py
LonelyCat124: Removed code.

src/psyclone/transformations.py
LonelyCat124: Fixed

src/psyclone/transformations.py
LonelyCat124: I previously only computed `taskloop.ancestor(OMPSingleDirective)` - the difference here is there is another subclass of `OMPSerialDirective` which is legal to contain these nodes (`OMPMasterDirective`), but does not allow some dependency patterns between regions to be satisfiable.

src/psyclone/transformations.py
LonelyCat124: Fixed

src/psyclone/transformations.py
LonelyCat124: Fixed

src/psyclone/transformations.py
LonelyCat124: Fixed

src/psyclone/transformations.py
LonelyCat124: Using the better rewritten docstring from `validate` now for this

src/psyclone/transformations.py
LonelyCat124: Fixed

src/psyclone/transformations.py
LonelyCat124: It was for momento, removed.

src/psyclone/transformations.py
LonelyCat124: Fixed

src/psyclone/transformations.py
LonelyCat124: I didn't cache it before, I will change that now.

src/psyclone/transformations.py
LonelyCat124: Changed

src/psyclone/transformations.py
LonelyCat124: Yes, I think thats almost correct, I think its:
```
if all(node1.abs_position <= taskloop_pos for node1 in nodes):
    # do stuff and break
```
as its only if there is no taskwait between this taskloop and the end of the single that we need the extra barrier.

src/psyclone/tests/test_files/gocean1p0/reordered_dependent_invoke.f90
LonelyCat124: I think this reordered one is a test I made to try some coverage improvements, but actually I can remove.

src/psyclone/tests/test_files/gocean1p0/reordered_dependent_invoke.f90
LonelyCat124: Removed

src/psyclone/tests/test_files/gocean1p0/reordered_compute_cu_mod.f90
LonelyCat124: Done

LonelyCat124:
Ok, I have resolved most comments (one refactor to do) and fixed the tests. I expect that some of the coverage will have dropped so I will look into that next.

I have some linter issues I'm not clear how to resolve:
```
src/psyclone/transformations.py:691:60: W0640: Cell variable fwriter defined in loop (cell-var-from-loop)
src/psyclone/transformations.py:691:68: W0640: Cell variable taskloop defined in loop (cell-var-from-loop)
src/psyclone/transformations.py:693:60: W0640: Cell variable fwriter defined in loop (cell-var-from-loop)
src/psyclone/transformations.py:694:64: W0640: Cell variable forward_dep defined in loop (cell-var-from-loop)
```
where I'm using some variables in a lambda (which is being passed to `LazyString`. Are these save to ignore or is there some way to fix?

arporter:
re. the linting errors: I haven't actually used LazyString yet so I defer to @sergisiso and @rupertford  who have...

sergisiso:
> I have some linter issues I'm not clear how to resolve

I think what it says is to be careful with the lambda capturing variables since these are references and are going to reference to somewhere else in the next iteration of the loop. But in this case this is not a problem since this is a "raise" so we know there won't be a next iteration. So I would say it is safe to disable the warning with this explanation or alternatively break the loop at this point and raising the exception after the loop if it has not succeeded may also get rid of the pylint warning.

src/psyclone/transformations.py
```diff
@@ -563,6 +566,422 @@ def apply(self, node, options=None):
...
+                # as well, so we return it
+                return node
+            if not isinstance(node, OMPTaskwaitDirective):
+                # For all our other node types we calculate their own
+                # variable accesses
+                node_vars = VariablesAccessInfo()
+                for child in node.walk(nodes.Node):
+                    if child is not node and not isinstance(child,
+                                                            (Schedule, Loop)):
+                        node_vars.merge(VariablesAccessInfo(child))
```
hiker, line 801: Just a quick comment: creating a new instance of VariablesAccessInfo is not necessary, you can simply do:
```
   child.reference_accesses(node_vars)
````
This will add the new accesses to the existing variable (and saves the overhead of creating a new instance, and then merging them together).

src/psyclone/transformations.py
LonelyCat124: I checked `ancestor` docstring and it says it only returns the first instance found. I didn't check the code explicitly so its possible thats wrong.

src/psyclone/transformations.py
LonelyCat124: Fixed, thanks!

src/psyclone/transformations.py
LonelyCat124: Done as a private static method

LonelyCat124:
Should be ready for review again now :)

src/psyclone/transformations.py
hiker: No idea what I was thinking here :(  Please ignore ;)

LonelyCat124:
Hmm, not sure why I can't comment on the new review comments but I'll try to go through them today. Probably I'll have to put my responses in this thread @arporter 

> Please add a comment, e.g. "Ensure loop dependencies are satisfied." More generally, how can we prevent a user from generating incorrect code? We could check dependencies at code-generation time but that won't work if there are places where we need to manually override such checks. Perhaps we should have a meta-transformation that performs these various steps and that's the one we publicise to users?

I don't want to have any restrictions on what is applied to loops in a region (as it may well be only some loops want to use tasks). I think I could add a `_dependency_checked` member to taskloops, and require that to be set to True during its validation. I can't really make the access to it private though, so there's nothing to stop a user just setting it to be True themselves, but the idea would be that a region containing a taskloop must have the taskwaitTrans applied before code generation.

PR #1427, opened 2021-10-01
(closes #1405) Add PSyIR nodes for OMPTarget and OMPLoop directives

### LINKED ISSUE #1405: Add OpenMP Target and OpenMP Loop pragmas
Using unified_memory it seems that OpenMP target and OpenMP loop (with collapse clause) are the only pragmas needed to use GPU acceleration using OpenMP.

There is a problem to add OpenMP Loop pragma that is that the OpenMP DO/FOR insertion transformations already takes the name of OMPLoopTrans and OMPParallelLoopTrans (and also versions of it with the API in front). This leave us with 3 options:
- Don't insert new transformations and just let this ones produce a omp loop instead of a do/for. This could be valid long term but we would need to be sure the current compiler support for this pragma is ready in most/all compilers and don't break current scripts performance.
- Rename this transformation to OMPDoTrans to leave space for the new OMPLoopTrans. This would require changing all the LFRic and Gocean scripts that already use this transformation.
- Keep the OMPLoopTrans as it is and create a transformation with a different name for the  loop directive. This could be confusing.

@arporter @rupertford what do you think? I prefer the first option as with OMPLoopTrans we already mean to make the loop parallel without any deeper meaning. Maybe if some compilers don't support the loop pragma for the time being we can add a configuration option like "enable_openmp5" and produce a "loop" if true and a "do/for" if false, so we don't even need to add new directive nodes.

sergisiso:


sergisiso:
@arporter @rupertford This is ready for first review.

PR #1432, opened 2021-10-05
(closes #1168) PSyIR lowering support for generic PSyData nodes

### LINKED ISSUE #1168: [PSyData] Add support for PSyData nodes to Fortran visitor
While working on #737 I've realised that there's no Visitor support for PSyData profiling nodes. This issue will add it.

sergisiso:


sergisiso:
@arporter This is ready for another look.

> gocean/eg5 fails with compilation - it looks as though the generated code does not contain the declarations of the various psy-data objects.

I checked that this works as expected now. There were 2 possible reasons for now working before:
-  There was an inconsistency as the PSyDataNode transformation was inserting the symbols in the Container but the gen_code was generating them on the Routine. (now they are always inserted in the Routine)
- The symbols were created in the .create() method so initializing only with the constructor could miss the symbols.

I fixed both. Also I realized many PSyDataNode logic was tested indirectly by the transformations, so I added more direct PSyDataNode unit tests to improve this aspect.

Issue #1387, opened 2021-09-08
Dependence analysis for hoist transformation


rupertford:
Dependence analysis needs to be used for the Hoist transformation so that it only hoists code that is safe to hoist. Some initial code to try to do this is given below but it is unclear (to me) what is the best approach to take.

```
# 1: All accesses in the assignment should be independent of
        # the loop iterator.
        iterator = parent_loop.variable
        for reference in node.walk(Reference):
            if reference.symbol is iterator:
                raise TransformationError(
                    "The supplied assignment node '{0}' depends on the parent "
                    "loop iterator '{1}'.".format(self._writer(node), iterator.name))

        # Known limitations for this dependence analysis
        # 1: Assumes all accesses are dependent (so does not
        # take into account a(1) and a(2) being different)
        # 2: Ignores potential side-effects within calls.

        # 2: The modified reference symbol should not be read or
        # written within the loop before this assigment.
        lhs_reference = node.lhs
        prev_reference = lhs_reference.previous_reference
        if prev_reference.abs_position > parent_loop.abs_position:
            raise TransformationError(
                "Writer has a previous read or write within the loop")

        # 3: The reference symbols that are read should not be written
        # within the loop before this assigment.
        for reference in self.rhs.walk(Reference):
            prev_reference = reference.previous_reference(access=WRITE)
            if prev_reference.abs_position > parent_loop.abs_position:
                raise TransformationError(
                    "Reader has a previous write within the loop")

        # 4: The writer can't be read and then written in the loop afterwards
        lhs_reference = assignment.lhs
        next_reference = lhs_reference.following_reference
        if next_reference.READ and next_reference.abs_position <= after_parent_loop.abs_position:
            next_write_reference = lhs_reference.following_reference(access=WRITE)
            if next_reference.abs_position <= after_parent_loop.abs_position:
                raise TransformationError(
                    "Writer has a following read then write within the loop")
            
        # 5: The readers can't be written in the loop afterwards
        for reference in self.rhs.walk(Reference):
            next_reference = reference.following_reference(access=WRITE)
            if next_reference.abs_position <= after_parent_loop.abs_position:
                raise TransformationError(
                    "Reader has a subsequent write within the loop")
```

rupertford:
And some more stuff cut from the code ...

```
# TODO: proper dependence analysis
# Use arrays but dep an tested for scalars and structures
# w only ok.
# do i=1,n
#   a = 0.0
# end do
#
# w -> w OK
[...]

PR #1433, opened 2021-10-07
1387 validate dependency for hoist

hiker:
This PR adds dependency analysis for the hoist transformation. Note that I added three convenience functions (``is_read_before()``, ``is_written_before()``, ``is_accessed_before()``), but in the end I only needed one (``is_accessed_before()``). I decided to leave them in, since I think they could be useful (even if just as an example for other functions). Tests have been added.

The ``validate_dependency()`` function is written to not assume an assignment statement, it should be able to handle many other cases. I've added tests for hoisting if statements out of a loop.

I think most of the cases listed in #1387 should be covered (though I simplified the code somewhat, e.g. instead of testing separately if there is a write and then a read, I just test if there is any kind of access.

The one case that is not supported (on purpose) is a double write, e.g.:
```
   do i=1, n
        a = 2
        a = 3
        b(i) = a
```
Hoisting ``a=2`` out will not be accepted (it will complain about at least one more write statement in the rest of the loop). Besides being rather useless code in the first place, it's quite difficult to handle similar looking cases (``if something then a=2 else a=3 endif`` would look the same from a dependency analysis point of view for this variable, and we open a can of difficult corner cases). If we have a real test case we improve this. 

I also just noticed: the above example works if we hoist both assignments to ``a`` out at the same time - I just tried. It turns out that (while not being document), the ``validate_dependencies`` function actually handles a list of nodes just fine, not only a single node. So it would actually be trivial to change the hoist transform to be a ``RegionTrans``, which makes it easy to then allow a list of nodes to be hoisted. But that can b a different PR :smile:


hiker:
No idea why it passed the coverage tests now, but not before. Ready for first review, @rupertford  or @arporter I'd guess.

Issue #1074, opened 2021-01-25
Add example for NAN tests and documentation


hiker:
After the restructuring of makefiles and examples has been merged (#1022,  #1061, #1067), add the gocean example for NAN testing.

PR #1439, opened 2021-10-08
#1074 Added NAN test example.

hiker:
Completing the set of PSyData examples for gocean.


examples/gocean/eg5/nan/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+The source code computes divisions by 0 on the diagonals, resulting in
+a invalid numbers (Infinity).
+
+## Running
+```
+$ ./nan_test
+...
+Allocating C-T field with bounds: (1:   6,1:   6), internal region is (2:   4,2:   4)
+ PSyData: Variable a_fld has the invalid value
```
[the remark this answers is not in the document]
arporter: Great, thanks.

examples/gocean/eg5/nan/README.md
```diff
@@ -23,24 +23,20 @@ This script uses PSyclone's ``NanTestTrans`` to instrument the two
...
-                  Infinity  at index/indices            2           2
- mainupdate
+PSyData: Variable a_fld has the invalid value Inf at index/indices 1 1 in module 'main' region 'update'.
+PSyData: Variable a_fld has the invalid value Inf at index/indices 2 2 in module 'main' region 'update'.
 
 ```
-After calling the kernel ``mainupdate``, five warnings are being printed,
-indicating that the field ``a_fld`` has infinity as value on the diagonals
-(i.e. indices 1 1, ..., 5 5)
+After calling the kernel ``mainupdate``, five warnings are printed,
```
arporter, line 37: Unfortunately I can't reproduce this! I can build and run the example but no warnings are produced. If I examine the generated PSy code then I see no variables are actually provided for checking:

      TYPE(nan_test_PSyDataType), save, target :: nan_test_psy_data

      CALL nan_test_psy_data % PreStart("main", "init", 0, 0)
      DO j = a_fld%whole%ystart, a_fld%whole%ystop, 1
        DO i = a_fld%whole%xstart, a_fld%whole%xstop, 1
          CALL init_field_code(i, j, a_fld%data, 1.0)
        END DO
      END DO
      DO j = b_fld%whole%ystart, b_fld%whole%ystop, 1
        DO i = b_fld%whole%xstart, b_fld%whole%xstop, 1
          CALL init_field_code(i, j, b_fld%data, 2.0)
        END DO
      END DO
      CALL nan_test_psy_data % PostEnd

Please could you investigate?

arporter (changes requested):
All looks good now. However, when I try running the NaN-test example it seems that the PSy layer is no longer being generated correctly and thus no warnings are generated. Please could you investigate. Weirdly though (given that) all of the tests run OK. Perhaps this means we're missing a test too?
Please could you also bring up-to-date with master.

examples/gocean/eg5/nan/README.md
hiker: The NanNode did not have a `lower_to_language_level` function. As such it worked when using the old `gen` method, but not when using the psyir FortranWriter (which is done under the hood in the example).

I've added that missing function, created a new test (very similar to an existing test, but using the FortranWriter for output), and verified that the test would have triggered (without the missing function), and works now.

There is some concern that other codes might have the same issue (extract has the lower function, read_only has, profile doesn't need it (it falls back to the base function, which works since profile does not use any variables)). I didn't check any further (e.g. if they have tests that uses a visitor - I doubt that somehow :) ). But I assume that soonish we will disable the old-style gen approach for gocean, and then we can update all these tests just once (instead of now duplicating them to use old style and visitor style, then later removing them). Let me know if you want a ticket to keep track of that.

hiker:
I've merged in latest master - this PR is ready for next review.

examples/gocean/eg5/nan/README.md
arporter: It all works for me now thanks. As discussed on Teams, once we've done away with gocean0.1 and Sergi has finished refactoring the OpenCL support then we can remove the `gen` functionality for GOcean.

arporter (approved):
All looks good now and the example works correctly.
Thanks for adding the new test.
All docs build OK.
Examples and tests are OK with compilation.
1074 is not mentioned anywhere.

PR #1450, opened 2021-10-14
1291 new dependency docs

hiker:
This moved the two dependency sections (old and new) into a separate chapter. It also adds doctests for the new dependency handling.

There is one example from the old style involving halo exchanges, I have to admit I am not certain how to test for that - the existing tests seems to involve parsing an external file, I am not sure if we want to do that for doctests.


hiker:
Read for a first review. Note that I've opened #1451 to start fixing failing tests in the developer's guide.

hiker:
Ready for next review,

PR #1475, opened 2021-10-27
(closes #595) Clean up Memento and Transformation return values

### LINKED ISSUE #595: Transformations should not return anything.

sergisiso:
This was agreed some time ago in #595 , the Memento wasn't working and the return values were already inconsistent. So in this PR I delete them.


src/psyclone/transformations.py
```diff
@@ -1692,32 +1614,6 @@ def name(self):
...
-        :returns: 2-tuple of new schedule and memento of transform.
-        :rtype: (:py:class:`psyclone.psyir.nodes.Schedule`,
-                 :py:class:`psyclone.undoredo.Memento`)
-
-        '''
-        if not options:
-            options = {}
-
-        return super(OMPMasterTrans, self).apply(node_list, options)
-
```
hiker, line 1720: While technically this is of course correct, I think we are losing some documentation here - as far as I can see the fact that this transformation must be within an OpenMP parallel region is not anywhere else. Either add this to the class documentation or bring this 'wrapper' function back (and maybe add a comment that this function is only for documentation purposes - Andy suggested something similar for me in the past as well).

hiker (changes requested):
Impressive work, and I really like to get rid of this old stuff :)  There is one small change, but I also noticed that there are still a few assignments from an ``apply`` left:
./src/psyclone/domain/nemo/transformations/create_nemo_psy_trans.py
./src/psyclone/tests/domain/gocean/transformations/gocean_opencl_trans_test.py
./src/psyclone/tests/dependency_test.py
./src/psyclone/tests/psyir/backend/psyir_openacc_test.py
./tutorial/practicals/nemo/4_nemo_openacc/solutions/kernels_trans.py
./tutorial/practicals/nemo/3_nemo_openmp/omp_trans.py
./tutorial/practicals/nemo/3_nemo_openmp/README.md
./tutorial/practicals/nemo/3_nemo_openmp/solutions/parallel_region_omp_trans.py
./tutorial/practicals/nemo/3_nemo_openmp/solutions/omp_trans.py

I used ``grep -i "=.*apply("  -r .   | grep -v pyc `` (and you still need to exclude documentation).

src/psyclone/transformations.py
sergisiso: I documented this is the class docstring, I think it makes sense because there is also the example of applying both transformations there.

sergisiso:
@hiker Thanks for the suggested grep command, I clean up those cases now. This is ready for next review.

sergisiso:
@hiker Brought this up to date to master again because there was new transformations to deal with.

src/psyclone/transformations.py
hiker: Thanks, perfect solution.

hiker:
I took the liberty of fixing one missed return statement/#todo 595.

All raised issues have been addressed, examples compile and all tests work. User's guide has been updated. The failed coverage line is raised from DummyTransformation (apparently used for pyreverse), and wasn't covered previously either.

hiker (approved):

PR #1482, opened 2021-11-03
(closes #1481) Rename symbol_from_tag to find_or_create_tag and add find_or_create (by name)

### LINKED ISSUE #1481: SymbolTable find_or_create methods
There is a symbol table method called "symbol_from_tag" that is used quite often but we have agreed previously that the name of what it does is confusing/unclear.

Also it will be beneficial to have a similar method that did the same but for names instead of tags (that is a lookup by name and if nothing is found create a symbol with this name and any additional parameters provided, this hides the try/except necessary with the lookup). In fact a few "_find_or_create_x" utility functions have been created in several places to do just that.

I propose to:
- rename SymbolTable method "symbol_from_tag" -> "find_or_create_tag"
- create new SymbolTable "find_or_create" to standardise ulility functions and simplify some try/except code.

both will have matching interfaces and functionality the only difference is that the lookup is by name or by tag.

sergisiso:
In contrast to what I said in the issue, I didn't end up replacing the utility functions because these are complex enough that it would hinder the readability if I inline these functions.

I did replace a few try/catches to find_or_create, but there are still some uses of the try/except lookup pattern and I think it is fine. In some situations I like it being more explicit. The use case I find the new method useful when there is multiple symbols (this is coming in my next OpenCL PR), in these cases a series of one-line statements seem clearer than a series of ty/except blocks.


sergisiso:
If CI is happy this is ready for review, from @rupertford @arporter @TeranIvy or @hiker

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1486,54 +1486,95 @@ def test_new_symbol():
...
+    ''' Tests the SymbolTable find_or_create method find existing symbols or
+    otherwise creates a new symbol with the given properties. '''
+    symtab = SymbolTable()
+    existing_symbol = Symbol("existing")
+    symtab.add(existing_symbol, tag="tag1")
+
+    # If the given name exists, return the symbol
+    assert symtab.find_or_create("existing") is existing_symbol
+
+    # If the given name does not exist, crate and return new symbol
```
arporter, line 1499: "create"

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1486,54 +1486,95 @@ def test_new_symbol():
...
+                                 visibility=Symbol.Visibility.PRIVATE,
+                                 constant_value=3)
+    assert new2.name == "new2"
+    assert isinstance(new2, DataSymbol)
+    assert new2.datatype is INTEGER_TYPE
+    assert new2.visibility is Symbol.Visibility.PRIVATE
+    assert new2.constant_value.value == "3"
+    assert symtab.lookup_with_tag("mytag") is new2
+
+    # Check that it fails if the Symbol type is different than expected by
```
arporter, line 1519: "...if the named Symbol exists but is not of the specified type."

arporter (changes requested):
Almost perfect. I would have let this through but there are test failures after merging with master.
Besides that, there are just a couple of typos that need correcting.
The issue number isn't referenced anywhere.
All examples are OK with compilation.
Ref. guide still builds fine.

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Fixed

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Fixed

sergisiso:
@arporter I fixed the typos, brought to master and fix the issues and remove the six uses from the symbol table. This is ready for another look.

arporter (approved):
Branch is up-to-date.
Latest changes are pylint clean.

PR #1485, opened 2021-11-05
(Closes #1429) Adds functionality to update argument intents to psyad

### LINKED ISSUE #1429: psyad does not changes intents of variables
The iintents of active variables may differ from the TL and its adjoint. For example, something that is only read (intent in) in the tangent linear might be read and written in its adjoint (intent inout). psyad does not currently deal with this. The example has had to be changed as a result of this problem in order to allow it to compile.

arporter:


arporter:
This is almost ready but I need to update the PSyAD documentation to mention that argument intents are updated by the visitor.

arporter:
Assuming CI happiness, this is ready for a first review from @rupertford  now.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -1784,7 +1784,7 @@ def _process_decln(self, parent, symbol_table, decl, visibility_map=None):
                 symbol_table.add(sym)
 
             # The Symbol must have the interface given by the declaration
-            sym.interface = interface
+            sym.interface = interface.copy()
```
rupertford, line 1787: As an alternative we could presumably allow interfaces to be shared in the PSyIR and then replace with a modified copy when we want to change one. That feels safer to me. Are there some gotchas with that approach? I'm wondering whether that should be done here in any case, just to make sure. With this implementation I'm guessing you are thinking that shared interfaces should be disallowed? - but if so how do we enforce this? Again I feel a Sergi discussion coming along at some point.

rupertford (changes requested):
Good job as usual. The main question I have is whether you should be copying the interfaces before updating them to be safe as I think you are relying on interfaces no longer being shared in the PSyIR.

doc/psyad/user_guide/implementation.rst
```diff
@@ -50,6 +50,13 @@ node in the PSyIR tree and transforms each node into its adjoint form. Once
 this is complete, the PSyIR representation is then written back out as
 code.
 
+If the supplied tangent-linear code takes the form of a subroutine
+with active and inactive variables passed as arguments then PSyAD must
+also update the ``intent`` of each argument. This is implemented as a
+final step in the visitor for a Schedule node: dependence analysis is
```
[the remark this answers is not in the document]
arporter: Good point, done.

doc/psyad/user_guide/psyad_command.rst
```diff
@@ -151,6 +151,9 @@ the previous command
 
    > psyad -oad ad_kernel.f90 -a var1 var2 -otest harness.f90 tl_kern.f90
 
+.. note:: A test harness can only be generated if the supplied tangent-linear
```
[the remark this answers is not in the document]
arporter: True although I don't think we've done anything LFRic-specific yet?

examples/psyad/testkern_mod.f90
```diff
@@ -44,11 +45,10 @@ module testkern_mod
...
     real, intent(in) :: ascalar
     integer, intent(in) :: npts
-    ! issue #1429. Active variables need to be declared as inout as
-    ! the intents can change in the adjoint version and PSyclone does
-    ! not currently deal with this.
     real, intent(inout), dimension(npts) :: field2
-    real, intent(inout), dimension(npts) :: field1
+    ! Deliberately leave off the intent for 'field1' to demonstrate that
+    ! it is correctly set in the generated adjoint code.
+    real, dimension(npts) :: field1
```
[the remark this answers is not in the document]
arporter: Good idea. I've added one and updated the Makefile.

src/psyclone/psyad/adjoint_visitor.py
```diff
@@ -148,6 +150,36 @@ def schedule_node(self, node):
...
 
+        # Creating the adjoint may have altered the way variables are
+        # accessed within the code. If any of the active variables are
+        # subroutine arguments then we must update the access property of
+        # the associated ArgumentInterface.
+        # Since a piece of code could contain many Schedules, ensure we are
+        # currently handling the outermost one.
+        if not node.ancestor(Schedule):
+            dtools = DependencyTools()
+            # 'input' parameters are those whose first access is a read
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyad/adjoint_visitor.py
```diff
@@ -148,6 +150,36 @@ def schedule_node(self, node):
             # else:
             #     node_copy.children.append(result)
 
+        # Creating the adjoint may have altered the way variables are
+        # accessed within the code. If any of the active variables are
+        # subroutine arguments then we must update the access property of
+        # the associated ArgumentInterface.
+        # Since a piece of code could contain many Schedules, ensure we are
+        # currently handling the outermost one.
+        if not node.ancestor(Schedule):
```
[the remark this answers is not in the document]
arporter: I take your point (at least, what I understand to be your point). I've changed this so that it checks whether node is an instance of Routine instead. An alternative would be to add a Routine handler to the visitor but since a Routine is a Schedule I thought that might get complicated.

src/psyclone/psyir/frontend/fparser2.py
arporter: In my mind an interface belongs to a Symbol and should not be shared. However, I think you're right about the conversation with @sergisiso as I raised similar concerns in his PR to resolve symbols in external modules and he disagreed with me :-)

arporter:
Ready for another look now @rupertford.

doc/psyad/user_guide/psyad_command.rst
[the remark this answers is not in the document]
rupertford: In the note it points out that the harness assumes that subroutines are contained in modules which is true in LFRic but that might not be the case in general.

doc/psyad/user_guide/implementation.rst
```diff
@@ -50,6 +50,14 @@ node in the PSyIR tree and transforms each node into its adjoint form. Once
 this is complete, the PSyIR representation is then written back out as
 code.
 
+If the supplied tangent-linear code contains active variables that are
+passed by argument the the intent of those arguments may change when
```
rupertford, line 54: "the the" -> "then the"

examples/psyad/eg1/testkern_mod.f90
```diff
@@ -34,31 +34,37 @@
 ! Author: A. R. Porter, STFC Daresbury Lab
 ! Modified by: R. W. Ford, STFC Daresbury Lab
 
-! Example module containing a very simple, one-line kernel subroutine.
+! Example module containing a very simple kernel subroutine that
+! contains only assignments.
```
rupertford, line 38: Please update as it now also contains a loop.

src/psyclone/psyad/adjoint_visitor.py
```diff
@@ -156,6 +158,41 @@ def schedule_node(self, node):
...
+            # We must update the symbols in the table of the new tree
+            adj_table = node_copy.symbol_table
+
+            for vname in self._active_variable_names:
+                sym = adj_table.lookup(vname)
+                if not sym.is_argument:
+                    continue
+                if vname in in_names:
+                    if vname in out_names:
+                        sym.interface.access = \
```
rupertford, line 189: It is here that I am concerned about potentially updating an interface that is share between symbols. I think we should copy the interface first to be safe.

src/psyclone/psyir/frontend/fparser2.py
rupertford: I've added another inline comment to make it clear where I think we should be copying the interface (in the visitor) to make sure we get no side effects. I think this should be done irrespective of this discussion.

rupertford (changes requested):
A couple of very minor things and one outstanding point about copying interfaces that I should have made more clear in the previous review.

examples/psyad/eg1/testkern_mod.f90
arporter: Oh yes, done.

src/psyclone/psyad/adjoint_visitor.py
arporter: Ah, I see. Copy added along with a TODO.

arporter:

rupertford (approved):
All issues have been addressed. docs build. example build. all changes are pycodestyle and pylint clean. Approving.

PR #1510, opened 2021-11-23
(closes #1499) Method to resolve properties of symbols imported from wildcard imports

### LINKED ISSUE #1499: Resolve datatype/attributes of symbols imported with wildcard imports (currently UnresolvedInterface)
For example, in this code
```
 use a, only: a_1
 use b
 use c
 
 a_1 = b_1 + 2
```

-  a, b, c are symbols with FortranModuleInterface with an import_container() method
-  a_1 is a symbol with ImportInterface with a container_symbol pointing to a
-  b_1 is a symbol with UnresolvedInterface

But we don't have a way to get more information about UnresolvedInterfaces, e.g. their datatypes, even though with can parse/access the b and c container tree and symbol tables.
 
Currently all symbols have:
 - get_external_symbol: if its a ImportInterface returns the equivalent symbol in the referred container (calls import_container the first time)
 - resolve_deferred: create a new symbol equivalent to the returned by get_external_symbol)
 
 Options to fix this issue:
 
 1) The get_external_symbol in UnresolvedInterface iterates all wildcard imports until it finds the symbol.
 2) A method (maybe in the symbol table) calls the import_container for a given wildcard import and updates the UnresolvedInterface to ImportInterface for the found symbols.

The benefit of 1 is that it uses the already existing methods but can be more expensive and expose more bugs (parsing unneeded files) although it may be fine to just skip and end up with the possibility to still be UnresolvedInterface after the iteration, this may be unavaidable anyway * **. The benefit of 2 is that we can be specific with API information. For instance in Nemo raising we can do a symboltable.resolve_symbol_interfaces_from(module_with_the_global_arrays) and similar for lfric/gocean infrastructure that we know definitions are, and prevent attempting parsing other files.

\* there is things like imports that are also part of a wildcard import in the module that I won't attempt for now.
** also do Fortran imports can have name clashes so if we miss one or don't respect the order it end up with a wrong definition?

sergisiso:


sergisiso:
@arporter @rupertford This is ready for review. Codecov marks a missing coverage line that has me quite confused, if I replace the line with pdb.set_trace() one of the tests stops on it, and if I replace it with a "raise" and send it to the CI the test does fail as expected. So I am not sure why the CodeCov is not happy with it. Otherwise the PR is ready for a first look.

src/psyclone/tests/psyir/symbols/data_type_symbol_test.py
```diff
@@ -72,3 +72,22 @@ def test_datatypesymbol_copy():
...
+def test_data_type_symbol_copy_properties():
+    ''' Check that the copy_properties() method works as expected. '''
+    symbol = DataTypeSymbol("origin", ArrayType(REAL_SINGLE_TYPE, [1, 2]))
+    new_sym = DataTypeSymbol("new_name", DeferredType())
+
+    new_sym.copy_properties(symbol)
+
+    # new_sym name should be unchanged, but its datatype should be updated
+    assert new_sym.name == "new_name"
+    assert new_sym.datatype == symbol.datatype
```
arporter, line 86: As commented in the implementation, I'm concerned that the `datatype` of `new_sym` is now a reference to the `datatype` of `symbol`?

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+            end subroutine test
+        end module test_mod
+    ''')
+    subroutine = psyir.walk(Routine)[0]
+    # Add Generic unresolved reference to "not_used"
+    subroutine.symbol_table.add(
+            Symbol("not_used1", interface=ImportInterface(
+                subroutine.symbol_table.lookup("b_mod"))))
+
+    # After parsing the a_1, a_2, b_1, b_2, not_used1 and not_used2 will have#
```
arporter, line 1689: s/have#/have/

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+    assert isinstance(a_2.interface, ImportInterface)
+    assert not isinstance(a_2, DataSymbol)
+    assert a_2.visibility == Symbol.Visibility.PRIVATE
+    assert isinstance(b_1.interface, UnresolvedInterface)
+    assert not isinstance(b_1, DataSymbol)
+    assert isinstance(b_2.interface, UnresolvedInterface)
+    assert not isinstance(b_2, DataSymbol)
+
+    # Set up include_path to import the proper modules
+    Config.get()._include_paths = [str(tmpdir)]
```
arporter, line 1709: Probably should use monkeypatch here to avoid polluting other tests.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+            integer :: a_1, a_2
+            integer :: b_1  ! Name clash but it is not imported
+        end module a_mod
+        ''')
+    with open(os.path.join(str(tmpdir), "b_mod.f90"), "w") as module:
+        module.write('''
+        module b_mod
+            integer, parameter :: b_1 = 10
+            integer, save, pointer :: b_2
+            integer :: not_used1
```
arporter, line 1665: Please could you add a private variable to this module and extend the test accordingly.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+
+    # a_2 is not yet resolved because if is from another symbol table,
+    # resolve that symbol table too
+    assert not isinstance(a_2, DataSymbol)
+    subroutine.parent.symbol_table.resolve_imports()
+    # In this case check that the visibility stays PRIVATE
+    assert isinstance(a_2, DataSymbol)
+    assert a_2.visibility == Symbol.Visibility.PRIVATE
+
+    # Clean up the config instance
```
arporter, line 1759: monkeypatch will remove the need for this (and may also be better if the test should fail).

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+    Config._instance = None
+
+
+def test_resolve_imports_name_clashes(fortran_reader, tmpdir):
+    ''' Tests the SymbolTable resolve_imports method raises the appropriate
+    errors when it find name clashes. '''
```
arporter, line 1765: "finds"!

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+
+    with open(os.path.join(str(tmpdir), "a_mod.f90"), "w") as module:
```
arporter, line 1767: pylint complains about these various `open` statements because they don't specify an encoding. Please could you add one.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+
+                name_clash1 = name_clash2 + not_a_name_clash
+            end subroutine test
+        end module test_mod
+    ''')
+    subroutine = psyir.walk(Routine)[0]
+    symtab = subroutine.symbol_table
+
+    # Set up include_path to import the proper modules
+    Config.get()._include_paths = [str(tmpdir)]
```
arporter, line 1799: monkeypatch please.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+
+    with pytest.raises(SymbolError) as err:
+        symtab.resolve_imports([symtab.lookup('b_mod')])
+    assert ("Found a name clash with symbol 'name_clash2' when importing "
+            "symbols from container 'b_mod'." in str(err.value))
+
+    with pytest.raises(SymbolError) as err:
+        symtab.resolve_imports([symtab.lookup('a_mod')])
+    assert ("Found a name clash with symbol 'name_clash1' when importing "
+            "symbols from container 'a_mod', this symbol was already defined "
```
arporter, line 1809: NIce.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+            "in 'b_mod'." in str(err.value))
+
+    # Clean up the config instance
+    Config._instance = None
+
+
+def test_resolve_imports_with_datatypes(fortran_reader, tmpdir):
+    ''' Tests the SymbolTable resolve_imports method. '''
```
arporter, line 1817: Please extend this docstring to distinguish this test from the first new one you added above.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+    with open(os.path.join(str(tmpdir), "my_mod.f90"), "w") as module:
+        module.write('''
+        module my_mod
+            type my_type
+                integer :: field
+                integer, dimension(:,:) :: array
```
arporter, line 1823: I think `array` must either be allocatable or have fixed dimensions.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+    assert "global1" not in symtab
+    # Some symbols types / datatype are inferred
+    assert isinstance(symtab.lookup("my_type"), DataTypeSymbol)
+    assert symtab.lookup("local1").datatype == symtab.lookup("my_type")
+    # but we don't know anything about the imported type
+    assert isinstance(symtab.lookup("my_type").datatype, DeferredType)
+    assert not isinstance(symtab.lookup("other_type"), DataTypeSymbol)
+
+    # Set up include_path to import the proper modules and resolve symbols
+    Config.get()._include_paths = [str(tmpdir)]
```
arporter, line 1856: monkeypatch please.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1643,3 +1646,232 @@ def test_rename_symbol():
...
+    symtab.resolve_imports()
+
+    # The global1 exist and is DataSymbols now
```
arporter, line 1859: "...is a DataSymbol now"

arporter (changes requested):
New functionality looks OK and is well tested. However, I'm a bit worried that we might be biting off more than we can chew by automatically looking into all modules that we can find. Have you tried this functionality on e.g. the NEMO code? I'd be happier if we only did the resolving one module at a time until we found what we were looking for.
I think we may also need to deepcopy a symbol's datatype rather than the shallow copy that I think is happening at the moment.
I've checked all modifications for pylint - see inline.
Probably the Developer Guide would benefit from a description of this method (once we've agreed upon what it should do!).

src/psyclone/psyir/symbols/data_type_symbol.py
```diff
@@ -107,6 +107,22 @@ def datatype(self, value):
...
+        :type symbol_in: :py:class:`psyclone.psyir.symbols.DataSymbol`
+
+        :raises TypeError: if the argument is not the expected type.
+
+        '''
+        if not isinstance(symbol_in, DataTypeSymbol):
+            raise TypeError("Argument should be of type 'DataTypeSymbol' but "
+                            "found '{0}'.".format(type(symbol_in).__name__))
+        super(DataTypeSymbol, self).copy_properties(symbol_in)
+        self._datatype = symbol_in.datatype
```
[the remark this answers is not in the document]
sergisiso: I am not totally sure on this one. If we made some modification inside a deep_copied type, then when generating the code it will not be reflected anywhere because the type is only described in the symbol definition (which in this case is in the external file) and they should match.

Also the INTEGER_TYPE, REAL_TYPE, ... that we have use a single instance of the specific scalar type and re-use the same instance everywhere.. which I think is more akind of a shallow reference for datatype.

Moreover, I see the types do not have a deepcopy method yet. So I would say it shouldn't be deepcopied for now, but probably at some point we need to design/think properly how types should behave as it has been done ad-hoc for now.

src/psyclone/psyir/symbols/data_type_symbol.py
```diff
@@ -107,6 +107,22 @@ def datatype(self, value):
...
+        symbol_in, apart from the name (which is immutable) and visibility.
+
+        :param symbol_in: the symbol from which the properties are copied.
+        :type symbol_in: :py:class:`psyclone.psyir.symbols.DataSymbol`
+
+        :raises TypeError: if the argument is not the expected type.
+
+        '''
+        if not isinstance(symbol_in, DataTypeSymbol):
+            raise TypeError("Argument should be of type 'DataTypeSymbol' but "
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
                         " tag '{1}' is already used by another symbol."
                         "".format(imported_var, tag))
 
+    def resolve_imports(self, container_symbols=None):
+        ''' Try to resolve deferred and unknown information from imported
+        symbols in this symbol table by searching for its definitions in
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+        for c_symbol in container_symbols:
+            try:
+                external_container = c_symbol.container
+            except SymbolError:
+                # Ignore this container if the associated module file has not
+                # been found in the given include_path.
+                # TODO #11: It would be useful to log this.
+                continue
+
+            for symbol in external_container.symbol_table.symbols:
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                if symbol.visibility == Symbol.Visibility.PRIVATE:
+                    continue  # We must ignore this symbol
+
+                if symbol.name in self:
+                    symbol_match = self.lookup(symbol.name)
```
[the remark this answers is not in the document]
sergisiso: Done (a slightly different comment)

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                    # If the import statement is not a wildcard import, the
+                    # matching symbol must have the appropriate interface
+                    # referring to this c_symbol
+                    if not c_symbol.wildcard_import:
+                        if not isinstance(interface, ImportInterface) or \
+                                interface.container_symbol is not c_symbol:
+                            continue  # It doesn't come from this import
+
+                    # Found a match, update the interface if necessary or raise
+                    # and error if it is an ambiguous match
```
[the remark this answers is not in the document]
sergisiso: Fixed

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                    if isinstance(interface, UnresolvedInterface):
+                        # Now we know where the symbol is coming from
+                        interface = ImportInterface(c_symbol)
+                    elif isinstance(interface, ImportInterface):
+                        if interface.container_symbol is not c_symbol:
+                            raise SymbolError(
+                                "Found a name clash with symbol '{0}' when "
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                                "importing symbols from container '{1}', "
+                                "this symbol was already defined in '{2}'."
+                                "".format(symbol.name, c_symbol.name,
+                                          interface.container_symbol.name))
+                    else:
+                        raise SymbolError(
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                            "Found a name clash with symbol '{0}' "
+                            "when importing symbols from container "
+                            "'{1}'.".format(symbol.name, c_symbol.name))
+
+                    # Now copy the external symbol properties, but keep the
+                    # interface and visibility as this are local properties
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                    if not type(symbol) == type(symbol_match):
+                        if isinstance(symbol, TypedSymbol):
+                            # All TypedSymbols have a mandatory datatype
+                            # argument
+                            symbol_match.specialise(type(symbol),
+                                                    datatype=symbol.datatype)
+                        else:
+                            symbol_match.specialise(type(symbol))
+                    symbol_match.copy_properties(symbol)
+                    symbol_match.interface = interface
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -1578,6 +1578,41 @@ def test_use_stmt():
...
+    assert symtab.lookup("this_mod").visibility == Symbol.Visibility.PUBLIC
+    assert symtab.lookup("var1").visibility == Symbol.Visibility.PUBLIC
+    assert symtab.lookup("other_mod").visibility == Symbol.Visibility.PUBLIC
+    assert symtab.lookup("var2").visibility == Symbol.Visibility.PUBLIC
+    assert symtab.lookup("my_mod").visibility == Symbol.Visibility.PUBLIC
+    assert symtab.lookup("some_var").visibility == Symbol.Visibility.PUBLIC
+
+
+def test_process_use_stmts_with_accessibility_statements(parser):
+    ''' Same than the previous test, but now from a module with a Fortran
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/tests/psyir/symbols/data_type_symbol_test.py
sergisiso: Yes, but this is similar to what we do with `== INTEGER_TYPE`, so if it is not what we want/need this needs a redesign expanding outside the scope of this PR.

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Deleted #

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Oh yes, done!

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Done

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Fixed

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
...
+                    symbol_match.visibility = visibility
+                else:
+                    if c_symbol.wildcard_import:
+                        # This symbol is PUBLIC and inside a wildcard import,
+                        # so it needs to be declared in the symbol table.
+                        new_symbol = symbol.copy()
+                        new_symbol.interface = ImportInterface(c_symbol)
+                        new_symbol.visibility = self.default_visibility
```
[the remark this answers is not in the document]
sergisiso: Yes the "private var" statement creates a generic var symbol with the appropriate visibility, so importing doesn't have to create a new symbol, just update the properties of the generic one. I added a new test to check this specific case.

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Instead, I added a new test to better check the accessibility statements are respected, does it cover what you had in mind?

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: This is a new one :) Done

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Done

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Done

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Done

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Added fixed dimensions, and since I was at it, I now assert that this dimension value is seen in the shape of the local symbol after resolving the import.

src/psyclone/tests/psyir/symbols/symboltable_test.py
sergisiso: Done, also for other tests in this PR

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1067,6 +1069,93 @@ def copy_external_import(self, imported_var, tag=None):
                         " tag '{1}' is already used by another symbol."
                         "".format(imported_var, tag))
 
+    def resolve_imports(self, container_symbols=None):
```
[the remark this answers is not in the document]
sergisiso: I added an optional `target_symbol` argument to allow to call this method with a more constrained scope (but still allow to do a wildcard search if no target_symbol is specified as I think is the most similar to what the compiler would to).

I case of not finding the specified `target_symbol` I raise and error. This seemed appropriate if we are searching for a particular symbol, but let me know if you think differently.

sergisiso:
@arporter This is ready for the next review. I addressed all comments but the one on the datatype deepcopy which probably requires its own separate discussion.

sergisiso:
@arporter I took this out of review because when processing Nemo there is one issue remaining with the "wp" symbol getting the wrong visibility. This doesn't come from the resolve method because the visibility is already wrong in the external container. I think it is that the frontend has a separate handler to process precisions that may not be respecting the symbol table default visibility, but I will explore this as I would like this to work well for the whole Nemo code.

sergisiso:
@arporter Now I checked it works with Nemo (*at least the files not excluded by OpenMP/OpenACC) for a couple of dependencies I am interested, I had to make sure the symbol table default visibility are respected when parsing the external containers. Now I can do something like:
```
if "oce" in mod_sym_tab:
	oce_symbol = mod_sym_tab.lookup("oce")
	mod_sym_tab.resolve_imports(container_symbols=[oce_symbol])

if "dom_oce" in mod_sym_tab:
	oce_symbol = mod_sym_tab.lookup("dom_oce")
	mod_sym_tab.resolve_imports(container_symbols=[oce_symbol])
```
With this a 8-way parallel psyclone of all nemo source code takes: 56s
Without resolving wildcard imports I took: 48s

sergisiso:
@arporter This is ready for review again

src/psyclone/psyir/symbols/data_type_symbol.py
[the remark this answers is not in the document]
arporter: OK, I think we need a discussion. My feeling was that a type should belong to a Symbol and not be shared but am happy to discuss. Look's like we'll be back to doing so on Teams though :-(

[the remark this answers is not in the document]
arporter: (I'm happy to mark this as resolved for now.)

arporter (changes requested):
I had a bit of trouble viewing the right diff of this PR this time around and also I got a bit delayed mid-review so I hope my comments still make sense.
There are just three comments left now.
I've taken the liberty of bringing this up-to-date with master.
All of the docs build OK.

src/psyclone/domain/lfric/psyir.py
```diff
@@ -101,10 +101,10 @@
     # Create the specific symbol
     exec(
         "class {0}DataSymbol(DataSymbol):\n"
```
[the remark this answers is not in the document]
sergisiso: This was modified because these dynamically created LFRic symbol classes didn't have 'visibility' as a constructor argument. This argument should have been here anyway but this became an issue now because the new_symbol method assigns an explicit visibility to new symbols depending on the symbol table default visibility.

Since visibility (and the interface argument which has here before) are not responsibilities of this class, I opted for just a kwargs argument which is passed to the super constructor.

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1077,6 +1084,150 @@ def copy_external_import(self, imported_var, tag=None):
...
+        if symbol_target and not isinstance(symbol_target, Symbol):
+            raise TypeError(
+                f"The resolve_imports symbol_target argument must be a Symbol "
+                f"but found '{type(symbol_target).__name__}' instead.")
+
+        for c_symbol in container_symbols:
+            try:
+                external_container = c_symbol.container
+            # pylint: disable=broad-except
+            except Exception:
```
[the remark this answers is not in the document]
sergisiso: When I wrote this I wanted to capture any exception here, my take is:
The IOErrors happen often if the file is not found, but can also be ValueError is a file is found but inside doesn't contain the expected module. This are alright because we don't aim to find all dependencies (and could be named exceptions).

But the reason for a broad except is that then fparser can also (and does in Nemo) fail for some files and this can be NoMatchError, ParseError, FortranSyntaxError, ... We could chose to fail here if something of this happens if you prefer? A compiler would fail for any of these in a dependency. And we can target just specific modules to search for the wildcard import to circumvent these errors. But I wrote this more as a "do whatever it can" and if we find an error in some dependency just ignore this particular dependency and continue. I wouldn't list by name fparser errors because then it means updating/adding fparser errors means we need to update this (and we probably forget until it fails), plus the list of errors is becoming quite long.

I don't think this will contribute to the generic transformation scripts errors messages because this Exception is dropped here and not raised upwards, but of course logging it would be useful to know what had happen in each step of the dependency tree (well we still just do depth 1 for now).

src/psyclone/domain/lfric/psyir.py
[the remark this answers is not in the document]
sergisiso: I added a visibility argument on some of the tests for this classes

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1077,6 +1084,150 @@ def copy_external_import(self, imported_var, tag=None):
...
+                                interface.container_symbol is not c_symbol:
+                            continue  # It doesn't come from this import
+
+                    # Found a match, update the interface if necessary or raise
+                    # an error if it is an ambiguous match
+                    if isinstance(interface, UnresolvedInterface):
+                        # Now we know where the symbol is coming from
+                        interface = ImportInterface(c_symbol)
+                    elif isinstance(interface, ImportInterface):
+                        if interface.container_symbol.name != c_symbol.name:
```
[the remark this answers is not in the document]
sergisiso: That's true. I updated the method and added a new tests for this specific case.

src/psyclone/domain/lfric/psyir.py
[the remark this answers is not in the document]
arporter: Aha, thanks for explaining.

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -343,6 +345,11 @@ def new_symbol(self, root_name=None, tag=None, shadowing=False,
         else:
             symbol_type = Symbol
 
+        # If no visibility parameter has been provided use this symbol's table
```
arporter, line 348: "symbol table's"

src/psyclone/psyir/symbols/symboltable.py
```diff
@@ -1077,6 +1084,149 @@ def copy_external_import(self, imported_var, tag=None):
                         " tag '{1}' is already used by another symbol."
                         "".format(imported_var, tag))
 
+    def resolve_imports(self, container_symbols=None, symbol_target=None):
+        ''' Try to resolve deferred and unknown information from imported
+        symbols in this symbol table by searching for their definitions in
+        referred external container. A single symbol to resolve can be
+        specified for a more targeted import.
+
+        :param container_symbols: list of container symbols to search in
```
arporter, line 1093: Missing line continuation.

src/psyclone/psyir/symbols/symboltable.py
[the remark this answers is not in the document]
arporter: OK, and I see you have a TODO to add logging so that's fine.

src/psyclone/tests/psyir/symbols/symboltable_test.py
```diff
@@ -1504,6 +1507,24 @@ def test_new_symbol():
             " one of its sub-classes but found" in str(err.value))
 
 
+def test_new_symbol_with_private_default_visibility():
+    '''Test that the new_symbol method creates method with the appropriate
```
arporter, line 1511: "...creates a symbol with the appropriate..."

src/psyclone/psyir/symbols/symboltable.py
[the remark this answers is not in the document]
arporter: Great

src/psyclone/psyir/symbols/symboltable.py
arporter: As this is just a docstring I've fixed it myself.

src/psyclone/psyir/symbols/symboltable.py
arporter: As this is just a comment I've fixed it myself.

src/psyclone/tests/psyir/symbols/symboltable_test.py
arporter: Just a docstring so fixed.

arporter (approved):
I've taken the liberty of fixing a few typos in comments/docstrings
All docs build OK.

PR #1512, opened 2021-11-23
(closes #1511 and #1516) Test utility compile_file moves to tmpdir before compiling

### LINKED ISSUE #1511: Test utility method compile_file() leaves a temporal file in the working directory
For a long time there has been a continuity_mod.mod and continuity_mod.o files created after running pytests with the compilation flag.

This can be solved by moving to tmpdir and coming back after the compilation has complete as it is already done by other tests in the utilities Compile class.

sergisiso:


hiker:
I tried to review this PR, but ran into #1516 - most of the time compilation would not happen for me at all. Considering that #1516 is the same file, and also pretty small as far as I can see  - would it make sense to combine these two bug fixes? Otherwise from what I have seen this patch is good - I just could not run all compilation, since most of the time nothing gets compiled :(, so I feel my testing is insufficient.
Alternatively, we merge this, and add a comment to #1516 to check that really no files are created elsewhere.

sergisiso:
@hiker Well spotted, I was not aware of this issue. I think I fixed it now by separating the compilation of dependencies from the final one and bringing the try/except construct to the dependencies loop body, so if one fails it continues with the next one.

I also added some prints and converted escape sequences from the compiler output so that if it fails it prints a much nicer text.

I had to update a a failing compilation assert and remove another one (but doesn't seem essential).

It CI passes this is ready for next review.

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -1317,7 +1317,6 @@ def test05p1_kernel_add_iteration_spaces(tmpdir):
         "Schedule:\n"
         "kern call: compute_cu_code\n")
     assert expected_sched in str(schedule)
-    assert GOcean1p0Build(tmpdir).code_compiles(psy)
```
hiker, line 1320: I've opened #1524 for this.

src/psyclone/tests/utilities.py
```diff
@@ -344,26 +350,33 @@ def _code_compiles(self, psy_ast, dependencies=None):
...
+            try:
                 name = self.find_fortran_file([self.base_path,
                                                str(self._tmpdir)], fort_file)
                 self.compile_file(name)
-            # Finally, we can build the psy file we have generated
+            except IOError:
+                # Not all modules need to be found, for example API
+                # infrastructure modules will be provided already built.
+                print("File {0} not found for compilation.".format(fort_file))
+                print("It was search in: {0}".format([self.base_path,
```
hiker: s/search/searched/

hiker:
The compilation tests now work reliably (59MB of it), failing coverage tests are caused by the fact that GHA does not run any compilation tests. I've brought it up to current master, no need to build UG. Will merge once CI is done.

hiker (approved):
Minor typo, which I will fix myself. I've opened #1524 for one issue, which I will fix.

PR #1515, opened 2021-11-24
1514 move gocean swap trans

hiker:
Closes #1514 by moving GOLoopSwapTrans into ``domain`` tree, and removes usage of six.

PR #1522, opened 2021-11-26
1519 generic loop swap

hiker:
After a discussion with @sergisiso I've decided to redo #1515 and base it all on master.
Note that this involves some changes to GOLoopSwap to make it usable as generic LoopSwap - I've removed the tests that all loops are GOLoop (there are already tests that it is indeed a nested loop). I can't think of a good reason why we would need a gocean-specific implementation that only allows GOLoops (but obviously it would be easy to re-implement this).


hiker:
Ready for review. I removed some 'six' usages and pylint warning - and I think the failed coverage is based on me fixing up a pylint warning that was not covered before.

src/psyclone/psyir/transformations/loop_swap_trans.py
```diff
@@ -0,0 +1,159 @@
...
+     >>> swap.apply(schedule.children[0])
+     >>> # Uncomment the following line to see a text view of the schedule
+     >>> # schedule.view()
+
+    '''
+    def __str__(self):
+        return "Exchange the order of two nested loops: inner becomes " + \
+               "outer and vice versa"
+
+    def validate(self, node_outer, options=None):
```
sergisiso, line 81: We should also check for triangular loops. These can not be swapped:

```
do i = 1, 10
    do j = 1, i
       ! body ...
```

I think it will be enough to walk for References in the inner_loop.start_expr/stop_expr.step_expr and make sure that any of them has outer_loop.variable as a symbol.

sergisiso: And maybe also check the opposite (that inner.variable is not in the outer start/stop/step - although it is probably bad code I think is still technically valid code to find) e.g:
```
do i = j + 3,  N
    do j = 1, 10
        ...
```

src/psyclone/psyir/transformations/loop_swap_trans.py
```diff
@@ -0,0 +1,159 @@
...
+        # pylint: disable=arguments-differ
```
sergisiso, line 82: One more check, could you also prevent to swap loops with Calls and CodeBloks inside (anywhere start/stop/step/body), this is because this may have side-effects and we are changing the order. e.g:

```
do i = 1, 10
  do j = 1, 10
     write_to_file(i, j)
     write(*,*) i , j  ! Codeblock
  enddo
enddo
```
Swapping would lead to a different outputs. (or worse deadlocks if it were MPI send/recvs )

I think is just a matter of adding them in the `excluded_node_types` class attribute and the super will deal with the check.

(Hopefully this is not a problem for any existing test, in the APIs kernel calls don't inherit from Call and we must guarantee they are side-effect free, so this shouldn't impact PSy-layer code)

src/psyclone/psyir/transformations/loop_swap_trans.py
```diff
@@ -0,0 +1,159 @@
...
+                                      "inner statement is not a valid loop:\n"
+                                      "{1}.".format(node_outer,
+                                                    str(err.value))) from err
+
+        if len(node_outer.loop_body.children) > 1:
+            raise TransformationError(
+                "Error in LoopSwap transformation. Supplied node '{0}' must"
+                " be the outer loop of a loop nest and must have exactly one "
+                "inner loop, but this node has {1} inner statements, the "
+                "first two being '{2}' and '{3}'"
```
sergisiso, line 124: Add message final punctuation mark.

src/psyclone/psyir/transformations/loop_swap_trans.py
```diff
@@ -0,0 +1,159 @@
...
+        self.validate(outer, options=options)
+
+        inner = outer.loop_body[0]
+
+        # Detach the inner code
+        inner_loop_body = inner.loop_body.detach()
+
+        # Swap the loops
+        outer.replace_with(inner.detach())
+        inner.addchild(Schedule())
```
sergisiso, line 151: Could this be `inner.addchild(outer.loop_body.detach())`?

outer.loop_body should be an empty Schedule after the previous inner.detach() but it could have an attached symbol_table with a defined symbol (not form Fortran code since there are only subroutine scope symbols but we may have added it manually).  It could also have a node annotation (although there are no schedule annotations for now). So I think its safer to re-use the existing Schedule than try to create a new one.

Also can you create a test where you manually add a symbol to each loop_body symbol table and check that this still exist after the transformation.

src/psyclone/tests/psyir/transformations/loop_swap_trans_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+import re
+
+import pytest
+
+from psyclone.domain.gocean.transformations import GOceanLoopFuseTrans
+from psyclone.psyir.transformations import TransformationError
+from psyclone.tests.gocean1p0_build import GOcean1p0Build
+from psyclone.tests.utilities import get_invoke
+from psyclone.psyir.transformations import LoopSwapTrans
```
sergisiso, line 50: Can you merge this statement with the previous `psyclone.psyir.transformations` import

src/psyclone/tests/psyir/transformations/loop_swap_trans_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def test_loop_swap_correct(tmpdir):
```
sergisiso, line 53: Please rename this test to `test_loop_swap_apply` and in the next ones replace `error` and `wrong` with `validate`. This is to keep a closer track to what method each test covers (although it is fairly obvious in this case).

sergisiso (changes requested):
@hiker The modification looks great, I agree there was nothing API specific in the original and this now has been moved to the intended location.

I picked up a few things that needed to be improved also in the original version to make the transformation more robust. Hopefully this can be solved in this PR but feel free to push back if this have some consequences in existing code.

src/psyclone/psyir/transformations/loop_swap_trans.py
hiker: I've opened #1536 for this. I would hope that proper dependency analysis picks up these dependencies automatically anyway (and yes, it is correct code - unfortunately ;) ). I've listed these two cases explicitly in the ticket, in case that they are not picked up automatically, so they will then be explicitly added.

src/psyclone/psyir/transformations/loop_swap_trans.py
hiker: Done - and all other tests seem to be fine with that.

src/psyclone/psyir/transformations/loop_swap_trans.py
hiker: It doesn't work out of the box (one of the loop is incomplete, to some validation is triggered).

My main question is: should I also swap the schedules, or leave them? If I swap them, I am not sure if we might not get different behaviour (if both loops declare the same variable?? But maybe this would then be detected when we have dependency analysis)? Let's discuss this in the Telco today.

src/psyclone/psyir/transformations/loop_swap_trans.py
hiker: And funnily enough, I immediately thought of your second case when I read about your first case - I have spent too much of my time programming Fortran :)

doc/user_guide/transformations.rst
```diff
@@ -220,6 +220,12 @@ can be found in the API-specific sections).
 
 ####
 
+.. autoclass:: psyclone.psyir.transformations.LoopSwapTrans
+   :members:
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/transformations/loop_swap_trans.py
hiker: Done.

src/psyclone/tests/psyir/transformations/loop_swap_trans_test.py
hiker: Embarrassing, considering that I was the advocate of sorting the imports.

src/psyclone/tests/psyir/transformations/loop_swap_trans_test.py
hiker: Done.

src/psyclone/psyir/transformations/loop_swap_trans.py
hiker: Implemented as discussed in the telco:
1. Now the schedules remain in place, no new schedule is created.
2. A TransformationError is raised if one of the two schedules involved contains a non-empty symbol table (every Schedule apparently has an empty symbol table as far as I can tell).

For the test I did not add a symbol, I tested that each old schedule ``is`` the same as new_schedule, i.e. same instance.

hiker:
Ready for next review!

sergisiso (approved):
@hiker All comments addressed. There is a new issue created for the remaining dependency analysis that we need for this transformation's validate. All test pass with compilation, there are no remaining TODOs, pylint and pycodestyle are clean and a few python2 compatibility constructs have been clean up. This is ready to merge.

PR #1526, opened 2021-11-26
(closes #1430) Fix array ranges in adjoint transformation

### LINKED ISSUE #1430: psyad assignment transformation does not work with array ranges
I'm sure it used to work but assignments with array ranges now causes psyad to complain that the assignment is not in tangent-linear form. I have had to change the example to avoid this problem.

arporter:


arporter:
This is ready for a first look when you're ready @rupertford.

src/psyclone/tests/psyad/transformations/test_assignment_trans.py
```diff
@@ -49,6 +49,13 @@
 from psyclone.tests.utilities import Compile
 
 
+@pytest.fixture(name="index_str", params=["1", "i", ":", "2:3"])
```
rupertford, line 52: When I wrote the tests I made it so that the first two tests included different datatypes and then the remaining tests I left as using directly addressed arrays. In keeping with this I think it would be nice to add basic array notation examples to `test_zero()` and `test_single_assign()`. I certainly would not remove what you have done as it adds rigour, but I do have a question, which is why did you choose these particular tests? Is it just a randomly selected subset?

rupertford (changes requested):
Looks great. Just the one comment inline and also, could you please update the example as it references this issue. This change might stress the harness generation - apologies if so.

src/psyclone/tests/psyad/transformations/test_assignment_trans.py
arporter: I think it was just the tests where I thought it was appropriate. I've extended the tests as you suggest and added some more for edge cases where we have the same array but with different elements being accessed.

arporter:
I've updated psyad/eg2 - thanks for pointing this out. It doesn't stress the harness generation as we can't generate a harness for it at the moment. I've checked through all the code and doc files and there are no other references to #1430.

arporter:
This is ready for review again now.

src/psyclone/psyad/transformations/assignment_trans.py
```diff
@@ -286,6 +320,35 @@ def validate(self, node, options=None):
...
+                        f"reference to the LHS variable "
+                        f"'{node.lhs.symbol.name}' without array notation"
+                        f" on the RHS: '{self._writer(node)}'")
+
+                sym_maths = SymbolicMaths.get()
+
+                for pos, idx in enumerate(active_variable.indices):
+                    lhs_idx = node.lhs.indices[pos]
+                    if not (type(idx) == type(lhs_idx) and
+                            sym_maths.equal(idx.start,
```
rupertford, line 340: Should there also be a TODO here referencing #1537?

src/psyclone/tests/psyad/transformations/test_assignment_trans.py
```diff
@@ -317,7 +370,59 @@ def test_multi_add(tmpdir):
...
+    we follow the rule:
+
+    A=B+xC -> B*=B*+A*; C*=C*+xA*; A*=0.0
+
+    where A=a(2:5), B=a(1:4) and C=b(1:4).
+
+    '''
+    tl_fortran = (
+        "  real a(10), b(10)\n  real :: x\n"
+        "  a(2:5) = A(1:4)+x*B(1:4)\n")
```
rupertford, line 407: I thought the transformation failed if the lhs ranges were not the same as the rhs?

rupertford (changes requested):
I still have a couple of issues/questions I'm afraid.

arporter:
Ready for another look now.

rupertford (approved):
All issues have been addressed. I've fixed a few minor issues. Approving.

Issue #1525, opened 2021-11-26
Apply lower_to_language in all LFRic built-ins


TeranIvy:
The `gen_code` method in some of the LFRic built-ins has already been replaced with `lower_to_langauge` method instead. This issue replaces the methods in all the remaining LFRic built-ins.

TeranIvy:
Related PR to reorganise global reductions is #2045. The support for symbol representation of precision has been added.

PR #1527, opened 2021-11-26
lower_to_language LFRic built-ins (towards #1525)

TeranIvy:
This PR is work towards #1525. It lowers most of the built-ins except for the following:
- `LFRicXInnerproductYKern`,
- `LFRicXInnerproductXKern`,
- `LFRicSumXKern`,
- `LFRicIntXKern`,
- `LFRicRealXKern`.

Sum and inner product are global sums, which requires changes in `psyGen.py` (change from generation of code as string to using symbols).
Conversions from `real` to `int` and vice-versa require that the `kind`s for the target precisions are also available as symbols. This is being added in the mixed precision PR #1349, so it makes sense to wait until that PR is merged.

Ready for the first review from either @arporter or @rupertford.


src/psyclone/tests/domain/lfric/lfric_integer_builtins_test.py
```diff
@@ -125,6 +125,15 @@ def test_int_X_plus_Y(tmpdir, monkeypatch, annexed, dist_mem):
             "f2_proxy%data(df)\n"
             "      END DO")
         assert output in code
+
+        # Test the lower_to_language_level() method
```
arporter, line 129: This isn't needed because the Integer builtins don't have their own `lower_to_language_level()` methods so this is effectively just testing the non-integer version which of course we already do elsewhere.

I think the same is true throughout this file?

arporter (changes requested):
Good job Iva, almost perfect.
Changes are pylint clean.
The 'LFRic->lowering->Builtins' section of the Dev Guide needs updating.
I think you've been a bit over-zealous with your tests of the integer built ins :-)
Apart from that it's all fine (although I haven't been able to check the builds of the documentation because of a separate problem.)

src/psyclone/tests/domain/lfric/lfric_integer_builtins_test.py
TeranIvy: @arporter, I followed your example of testing for `int_a_minus_X` where there is also a test for `lower_to_language` method. Besides, isn't all the code generation ultimately going to be done trhough PSyIR? If so, then we will be expanding the testing of lowering and removing the current testing of `gen_code`. Therefore I would like to leave those tests in there.

arporter: Oops, that's my mistake then - there's no need to test a routine that doesn't exist!

TeranIvy: OK. Would you like me to remove all `lower_to_language_level()` tests for integer built-ins, including yours? I think leaving the tests for the code generation that covers declarations is essential as this is where the real difference is between the `real` and `integer`-valued built-ins (different field types).

arporter: Yes please and yes, I agree - we definitely need the existing code-generation tests.

TeranIvy: Done.

TeranIvy:
Addressed review comments (`integer` built-in tests and the developer guide) and brought the branch up to date with master. Ready for another look from @arporter.

doc/developer_guide/APIs.rst
```diff
@@ -1041,9 +1041,19 @@ layer routine. A lot of this work is currently performed in the
...
+To date, all the LFRic BuiltIns have had ``lower_to_language_level()``
+methods implemented except for the following:
+
+* ``LFRicXInnerproductYKern``,
+* ``LFRicXInnerproductXKern``,
+* ``LFRicSumXKern``,
+* ``LFRicIntXKern``,
+* ``LFRicRealXKern``.
+
+The sum and inner product BuiltIns require implementing PSyIR representation
```
arporter, line 1053: The 'R' in 'PSyIR' stands for 'Representation' so you don't need "representation". Also, please could you say what needs to be implemented or, alternatively, simplify and just say that the PSyIR needs to be extended to handle reductions.

arporter (changes requested):
Docs all build fine.
Thanks for updating the dev guide - it just needs a bit of tidying and then this is good to go.

TeranIvy:
Rephrased the user guide text as suggested and refreshed the branch with the master. Back to @arporter for further review (note that the markdown check is against randomly failing).

PR #1531, opened 2021-11-29
#1338 Found an error in documentation of taskwait (wrong module import)

LonelyCat124:
The documentation of how to import taskwait used the old `from psyclone.transformations` which is not how this is imported.


src/psyclone/psyir/transformations/omp_taskwait_trans.py
```diff
@@ -67,7 +67,8 @@ class OMPTaskwaitTrans(Transformation):
     >>> psy = PSyFactory(api).create(invokeInfo)
     >>>
     >>> from psyclone.transformations import OMPParallelTrans, OMPSingleTrans
-    >>> from psyclone.transformations import OMPTaskloopTrans, OMPTaskwaitTrans
+    >>> from psyclone.transformations import OMPTaskloopTrans
+    >>> from psyclone.psyir.transformations import OMPTaskwaitTrans
     >>> singletrans = OMPSingleTrans()
```
arporter, line 72: This same mistake is also made in the docstring of OMPTaskLoopTrans. Please could you fix it there too.

arporter (changes requested):
Thanks for fixing. In checking the rebuilt documentation I noticed that we have the same mistake in one other location so it would be good to fix that too.
Apart from that, I've brought the branch up-to-date with master.
Docs all build fine. Tests and examples are all OK with compilation.

src/psyclone/psyir/transformations/omp_taskwait_trans.py
LonelyCat124: Done :)

arporter (approved):
Thanks for fixing the second one.

PR #1533, opened 2021-11-30
1523 Add SymPy symbols for Fortran names

### LINKED ISSUE #1523: SymPy and pre-defined functions
While trying to fix #753 with
```
$ git diff domain/common/algorithm/psyir.py
diff --git a/src/psyclone/domain/common/algorithm/psyir.py b/src/psyclone/domain/common/algorithm/psyir.py
index 395b45196..38197c287 100644
--- a/src/psyclone/domain/common/algorithm/psyir.py
+++ b/src/psyclone/domain/common/algorithm/psyir.py
@@ -39,6 +39,7 @@
 '''
 
+from psyclone.core import SymbolicMaths
 from psyclone.psyir.nodes import Call, Reference, DataNode, Literal, \
     ArrayReference
 from psyclone.psyir.symbols import DataTypeSymbol, ContainerSymbol, \
@@ -175,6 +176,7 @@ class AlgorithmInvokeCall(Call):
 
         arguments = []
         arguments_str = []
+        sym_maths = SymbolicMaths.get()
         for kern in self.children:
             for arg in kern.children:
                 if isinstance(arg, Literal):
@@ -183,7 +185,10 @@ class AlgorithmInvokeCall(Call):
                 elif isinstance(arg, (Reference, ArrayReference)):
                     # TODO #753 use a better check for equivalence
                     # (symbolic maths)
-                    if str(arg).lower() not in arguments_str:
+                    for existing_arg in arguments:
+                        if sym_maths.equal(arg, existing_arg):
+                    else:
                         arguments_str.append(str(arg).lower())
                         arguments.append(arg.copy())
                 else:
```
I realised that the corresponding test in 'tests/domain/common/algorithm/alg_psyir_test.py` still fails. Reason appears to be that this test uses:
```
        "  call invoke(kern(field(i+1), field(1+i)))\n"
```
and `field` is a SymPy python functions, so parsing (which uses `eval`) fails. Changing the name from `field` to something else fixes the issue.

We need to find a way to disable that using a variable that clashes with a SymPy Python function causes this kind of failures :(

hiker:
Fixes #1523. As discussed in the telco, I've added some tests that use a member variable and stand-alone variable with the same name, and that seems to work just fine (I guess SymPy doesn't really care what the symbols are). 

Note that when we do more advanced maths (comparing indices to see if one is larger than the other), and solving equations, which would involve assigning values to symbols, that might change - but some quick tests have shown that cases where this could become a problem, are too hard for SymPy to find a solution that's useful for us anyway. Anything we do will either be extremely conservative (i.e. if sympy can't solve, we assume a dependency/... and don't allow the transformation), or needs explicitly overwriting using an option anyway. If we should detect some common cases that would be nice to solve automatically, we can revisit this (e.g. anything involving indirect addressing `a(b(i))` is hard/impossible for SymPy to handle, since it knows nothing about `b`. We could then look at providing additional information  ... but things are getting quite complicated quickly here).


hiker:
SymPy 1.9.1 on GHA does not handle it, so we need to properly declare arrays as functions (and scalar looking accesses as symbols).

src/psyclone/tests/core/symbolic_maths_test.py
```diff
@@ -240,3 +238,34 @@ def test_symbolic_math_functions_with_constants(fortran_reader, expressions):
...
+    # A dummy program to easily create the PSyIR for the
+    # expressions we need. We just take the RHS of the assignments
+    source = '''program test_prog
+                 use some_mod
+                 integer :: field(10), i
+                 type(my_mod_type) :: a, b
+                 x = {0}
+                 x = {1}
+                 end program test_prog
+             '''.format(expressions[0], expressions[1])
```
arporter, line 267: Please use an f-string.

src/psyclone/tests/core/symbolic_maths_test.py
```diff
@@ -240,3 +238,34 @@ def test_symbolic_math_functions_with_constants(fortran_reader, expressions):
...
+                                         ("b+a%b(a%c,a%c,a%c)",
+                                          "b+a%b(a%c,a%c,a%c)")
+                                         ])
+def test_symbolic_math_use_reserved_names(fortran_reader, expressions):
+    '''Test that reserved names are handled as expected. The SymPy parser
+    uses 'eval' internally, so if a Fortran variable name should be the
+    same as a SymPy function (e.g. 'field'), parsing will fail.
+
+    '''
+    print("--------------------------")
```
arporter, line 257: Please rm this line.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -136,6 +138,74 @@ def test_sym_writer_functions(fortran_reader, expressions):
...
+
+    '''
+    # A dummy program to easily create the PSyIR for the
+    # expressions we need. We just take the RHS of the assignments
+    source = '''program test_prog
+                use my_mod
+                type(my_type) :: a, b(10)
+                integer :: i, j, x, a_c, b_c, a_b_c, a_b_c_1
+                x = {0}
+                end program test_prog '''.format(expressions[0])
```
arporter, line 169: Please use an f-string.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -52,11 +55,37 @@ class SymPyWriter(FortranWriter):
...
     attached, e.g. 2_4) and some intrinsic functions (e.g. MAX, which SymPy
     expects to be Max).
+    It additionally supports accesses to structure types. A full description
+    can be found in the manual:
+    https://psyclone-dev.readthedocs.io/en/latest/sympy.html#internal-details
 
     '''
 
-    def __init__(self):
+    def __init__(self, list_of_expressions=None):
```
arporter, line 64: Please document the `list_of_expressions` argument. I'm wondering whether it would be nicer to have it as a `list_of_references` as, in use, it looks a bit odd to construct the SymPyWriter and give it a list of expressions, only to then pass those same expressions to the writer to get the generated string.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -136,6 +138,74 @@ def test_sym_writer_functions(fortran_reader, expressions):
...
+
+    psyir = fortran_reader.psyir_from_source(source)
+    expr = psyir.children[0].children[0].rhs
+    sympy_writer = SymPyWriter(list_of_expressions=[expr])
```
arporter, line 173: This is what makes me uneasy about the fact we specify the same things to the constructor and then when calling the writer instance. I can understand the reason but it feels like we should be supplying the constructor with a list of reserved names or even a type map?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -136,6 +138,74 @@ def test_sym_writer_functions(fortran_reader, expressions):
...
+    _ = sympy_writer(expr)
+    assert sympy_writer.get_sympy_types() == expressions[1]
```
arporter, line 211: I'm beginning to think that this construction of a type map should be done by a separate, (static?) method of the Writer?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -136,6 +138,74 @@ def test_sym_writer_functions(fortran_reader, expressions):
...
+                use my_mod
+                type(my_type) :: a, b(10)
+                integer :: i, j, x, a_c, b_c, a_b_c, a_b_c_1
+                x = {0}
+                end program test_prog '''.format(expressions[0])
+
+    psyir = fortran_reader.psyir_from_source(source)
+    expr = psyir.children[0].children[0].rhs
+    sympy_writer = SymPyWriter(list_of_expressions=[expr])
+    _ = sympy_writer(expr)
```
arporter, line 210: Is this call required? If it is then that's a bit worrying as it implies that state is generated by this call which doesn't seem right to me. I think the other writers don't create state - they just give back a string.

src/psyclone/tests/psyir/nodes/member_test.py
```diff
@@ -71,3 +72,17 @@ def test_member_can_be_copied():
...
+
+
+def test_member_is_array():
+    ''' Test that we can check if a member is an array. '''
+    mem = nodes.Member("fred")
+    assert mem.is_array() is False
+
+
+def test_member_get_signature():
+    ''' Test that we can check if a member is an array. '''
```
arporter, line 84: Please update this docstring :-)

src/psyclone/psyir/nodes/reference.py
```diff
@@ -64,6 +64,14 @@ def __init__(self, symbol, parent=None):
         super(Reference, self).__init__(parent=parent)
         self.symbol = symbol
 
+    def is_array(self):
```
arporter, line 67: For consistency with the Signature.is_structure method, I suggest making this a property (and also in ArrayMixin).

arporter (changes requested):
Thanks for getting to grips with this Joerg. It's a tricky issue but I think your solution is good.
I do have some queries about the design of the SymPyWriter - particularly how we should best supply a list of reserved names/type maps. I think that a separate, static utility method that creates such a type map might be better than having it in the constructor of the writer itself.
Updated docs build fine but are a bit Fortran centric - I've made suggestions to try and fix this.
Updated source files are all pylint clean (apart from f-string suggestions).

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: @arporter Yes, I agree, that's not nice. It's hard to avoid - the SymPy writer needs to have all references (to avoid renaming something that will later when handling the second expression require to rename a symbol).
The only idea I have: I add a static function `convert_to_sympy` that takes a list of PSyIR nodes, and returns a list of strings containing each converted PSyIRnode. So this would mean we would do:
```
  exp1_str, exp2_str = SympyWriter.convert_to_sympy([exp1, exp2]).
```

The static method `convert_to_sympy` would then first collect all references and provide them to the SymPy constructor. I think it's 'nicer' to have this all in one class than in an additional utility function elsewhere (unless we need this functionality, but I don't think so atm).

Is that acceptable?

arporter: I think it's better but it then changes the semantics with which the SympyWriter is called and thus it is no longer consistent with the other writers. How about:

    map = SympyWriter.create_type_map(expr)
    writer = SymPyWriter(map)
    lhs = writer(expr1)
    rhs = writer(expr2)
    ...
    str_exp1 = parse_expr(str_exp1, local_dict=map)

Since you have to separately supply a mapping to the `parse_expr` call anyway?

hiker: Yes, but `map` is (potentially) modified when writing the two expressions (with the potential renaming of member names). While this would technically work (SympyWriter keeps a pointer to `map` and modifies it), but I think it's very unexpected that `map` is (potentially) modified after each `writer(exp1/2)` call :(

Plus, you are still supplying the arguments twice (argument of `create_type_map` must actually have both expressions to find all references :(  ).

What about:
```
map = SympyWriter.create_type_map(expr1, expr2)
writer = SymPyWriter()
lhs = writer(expr1, map)
rhs = writer(expr2, map)
```
This way there is at least some indication that map might be modified, though it still feels like a bad side effect (plus we break the normal interface of a writer). And we still supply each argument twice. 

Or:
```
writer = SympyWriter(exp1, exp2).
lhs = writer.get_expression(1)
rhs = writer.get_expression(2)
```
Not really good ... I'll leave this here in the spirit of wild brainstorming :)

I'll ponder about this tomorrow, but I am getting the feeling that we have to accept at least the fact that the arguments are specified twice :(

hiker: I am going to implement your way, but also provide a convenience function that hides the detail from the users as a static method. This way I can well document the fact that the mapping is modified by the other calls to writer.
Won't be able to finish that today though :(

arporter: I hadn't appreciated that the map would be modified by the calls to writer. Is there any way of avoiding this (by doing all the setup earlier)?

arporter: Alternatively, would it help if we had a transformation that did something appropriate to the PSyIR before giving it to the writer? I've also just remembered that one of the problems I had when comparing index expressions was that my use of the Fortran writer got massively expensive because it copies the (whole?) tree each time. Is that a consideration here as well?

hiker: > Is there any way of avoiding this (by doing all the setup earlier)?

Not really: we need to get all references (from all expressions) into the symbol table first (so that we don't rename any member into an existing variable that might only be used in the second expression). Then we need to handle all members. 
The only way to avoid this would be to do the whole conversion at the constructor (and then in the writer either do it again, or use cached results ... though caching results would also mean that we have to compare the trees to make sure we get the right cached result).

OK, now that I have written this: technically, after collecting all references, we could just do a `walk` in the constructor to find all members, add them to the symbol table and get new unique names for each member, and then store the new names for later. But that would just mean yet another walk across the whole tree.

> Alternatively, would it help if we had a transformation that did something appropriate to the PSyIR before giving it to the writer?

That's what I did originally (kind of ... it wasn't a stand-alone transformation) , and we all agreed that this way (copying the tree and changing the name of the members in the copied tree) is not the right way to g., and that it would be much easier to understand to use a writer.

I also think if we hide this from the user by providing a static function that takes a list of PSyIR expressions, and that returns a list of _sympified_ strings, that should be fine (if it is properly documented). This way the user does not even see that a SymPy writer is used.

We could provide a slightly different interface to the writer: instead of taking one expression, we let the writer convert a list (or tuple) of expressions all at the same time. Then we don't need this information at the constructor anymore, don't need to deal with keeping a state , ... - but it changes the interface for this writer (which I think is much worse than having a writer that keeps a state), and ... from a user's point of view it's pretty much the same as the aforementioned static function :)

TBH, I also don't see it as too much of a problem that there is one writer that keeps a state - that's why we have an OO approach. And I don't think a 'normal' user will ever use the SymPy writer: the SymbolicMaths class will provide all functions required for symbolic maths, and will hide the usage of the writer.

> I've also just remembered that one of the problems I had when comparing index expressions was that my use of the Fortran writer got massively expensive

Hmm - my initial thought was that for my use cases (dependency analysis) I will mostly be comparing indices, and while in theory they can be arbitrarily complex, in reality we will mostly see `i, j, k +- some constant`.  But if this had already been a bottleneck, this is a bit of a concern (I only remembered that converting expressions for error messages which were de-facto caught and ignored was expensive). Do you have any details (example code)? One thought I had: we might try to first throw the expression as is to sympy (without replacing members, kind constants with _) - if the most frequent use cases do not use structure, ... - that might be a bit cheaper. But I don't like that idea too much :(  One question would be: why do we actually copy the tree first? The writer is not supposed to modify anything.

I am happy to discuss this in the next telco, and on Monday I can just focus on creating more training material.

arporter: It takes a copy (IIRC) because we have to flatten all of the symbol tables. The code where I hit problems was: https://github.com/stfc/PSyclone/blob/2fc39510a3bca5d07849dff5151e90600fc6b8aa/src/psyclone/psyir/nodes/array_mixin.py#L227-L252. Since SympyWriter is not attempting to produce Fortran it doesn't need to collapse the symbol tables and therefore we can probably avoid doing the tree copy. (Unless @sergisiso can remember that there are other reasons for doing the copy?).

arporter: I also agree with you that we're probably best hiding everything in a static method - as you say, it's not really for public consumption.

arporter:
Hi @hiker, this is the file that once took 10 minutes to process:

[sbccpl.txt](https://github.com/stfc/PSyclone/files/7838664/sbccpl.txt)

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+   ``a%b`` will create two SymPy symbols: ``a`` and ``a_b`` (or a similar
+   name if a name clash was detected).
+3. No precision or kind information is added to a constant (e.g. a Fortran
+   value like ``2_4`` will be written just as ``2``).
+4. The intrinsic functions ``Max``, ``Min``, ``Mod`` are returned with a
+   capitalised first letter only. The Fortran writer would write them
+   as ``MAX`` etc., which SymPy does not recognise - it would handle them
+   as unknown functions.
+
+.. autoclass:: psyclone.psyir.backend.sympy_writer.SymPyWriter
```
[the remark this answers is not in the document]
hiker: Done (assuming that this is the setting `epub_copyright = u'2017-2022, STFC'`).

src/psyclone/psyir/nodes/reference.py
hiker: Done.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: Yes, it is, since the writer will add member variables to the type mapping (but we had this state discussion elsewhere, so more details ... elsewhere :) )

hiker (commented):
I wonder why I only get to see very few comments in the 'review' pan. I will address the rest individually.

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+Converting Fortran to Sympy - SymPyWriter
+-----------------------------------------
+The method ``equal`` of the SymbolicMaths class expects two PSyIR
+nodes. It converts these expression first into strings before parsing
+them as SymPy expressions. The conversion is done with the SymPyWriter
+class. As described in the previous section, a member of a structure
+in Fortran becomes a stand alone symbol or function in sympy. The SymPy
+writer will rename members to better indicate that they are members:
+an expression like ``a%b%c`` will be written as ``a%a_b%a_b_c``, which
+SymPy then parses as ``MOD(a, MOD(a_b, a_b_c))``. These convention
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+makes it easier to identify what the various expression in SymPy are.
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+
+This handling of member variables can result in name clashes. Consider
+the expression ``a%b + a_b + b``. The structure access will be using
+two symbols ``a`` and ``a_b`` - but now there are two different symbols
+with the same name. Note that the renaming of the member from ``b`` to
+``a_b`` is not the reason for this - without renaming the same clash would
+happen with the symbol ``b``.
+
+SymPy uses a symbol table to make sure it creates unique symbols. It
```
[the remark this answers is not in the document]
hiker: Indeed I do :)

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+first adds all references in the expression to the symbol table, which
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+guarantees that no Fortran reference is renamed. The writer then renames
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
 6. Since all arguments of the ``Mod`` function are identical,
    SymPy will report these two functions to be the same, which
    is the expected outcome.
+
+Converting Fortran to Sympy - SymPyWriter
```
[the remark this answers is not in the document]
hiker: Indeed.

doc/developer_guide/sympy.rst
```diff
@@ -102,46 +102,27 @@ be fused have the same loop boundaries using code like this:
...
-2. The intrinsic functions ``Max``, ``Min``, ``Mod`` are returned with a
-   capitalised first letter only. The Fortran writer would write them
-   as ``MAX`` etc., which SymPy does not recognise - it would handle them
-   as unknown functions.
-
-SymPy also has no concept of Fortran structure references or array syntax
+Handling of Fortran Structures and Arrays
+-----------------------------------------
+SymPy has no concept of Fortran structure references or array syntax
 like ``a(i)%b``. But this case is not handled especially, the Fortran
```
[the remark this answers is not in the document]
hiker: I've improved that.

doc/developer_guide/sympy.rst
```diff
@@ -102,46 +102,27 @@ be fused have the same loop boundaries using code like this:
...
-
-1. No precision or kind information is added to a constant (e.g. a Fortran
-   value like ``2_4`` will be written just as ``2``).
-2. The intrinsic functions ``Max``, ``Min``, ``Mod`` are returned with a
-   capitalised first letter only. The Fortran writer would write them
-   as ``MAX`` etc., which SymPy does not recognise - it would handle them
-   as unknown functions.
-
-SymPy also has no concept of Fortran structure references or array syntax
+Handling of Fortran Structures and Arrays
```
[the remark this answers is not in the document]
hiker: Frantically googling for a thesaurus to find synonyms for 'done' :)

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+``a_b`` is not the reason for this - without renaming the same clash would
+happen with the symbol ``b``.
+
+SymPy uses a symbol table to make sure it creates unique symbols. It
+first adds all references in the expression to the symbol table, which
+guarantees that no Fortran reference is renamed. The writer then renames
+all members and makes sure it uses a unique name. In the case of
+``a%b + a_b + b``, it would create ``a%a_b_1 + a_b + b``, using the name
+``a_b_1`` for the member to avoid the name clash with the reference
+``a_b`` - so an existing Fortran reference will not be renamed, only
```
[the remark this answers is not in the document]
hiker: Done.

doc/developer_guide/sympy.rst
```diff
@@ -152,3 +133,54 @@ considered to be equal:
...
+following, SymPy specific features:
+
+1. It will declare any array access as a SymPy unknown function, and any
+   scalar access as a SymPy symbol. These declarations are stored in a
+   dictionary, which can be queried. This dictionary is parsed into the
+   SymPy writer to ensure the correct interpretation of any names found
+   in the expression. Declaring arrays as functions results in the correct
+   behaviour of SymPy: in case of an unknown function SymPy will compare
+   all arguments, which are the array indices.
+2. It renames members as described above. So a Fortran expression like
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/core/symbolic_maths.py
```diff
@@ -108,8 +101,22 @@ def equal(self, exp1, exp2):
...
             return exp1 == exp2
 
-        str_exp1 = parse_expr(self._writer(exp1))
-        str_exp2 = parse_expr(self._writer(exp2))
+        # Avoid circular import
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.backend.sympy_writer import SymPyWriter
+
+        # Create a new writer, and pre-fill its internal symbol table
+        # with all references (so we do not rename any symbols, only
```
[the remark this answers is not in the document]
hiker: Indeed.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -52,11 +55,37 @@ class SymPyWriter(FortranWriter):
...
         super().__init__()
+
+        # The symbol table is used to create unique names for structure
+        # members that are being accesses (these need to be defined as
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
             self._intrinsic.add(op_str)
             self._op_to_str[operator] = op_str
 
+    def get_sympy_types(self):
+        ''':returns: returns the mapping of symbols in the written
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -52,11 +55,37 @@ class SymPyWriter(FortranWriter):
...
+        # references in the expression).
+        self._symbol_table = SymbolTable()
+
+        if list_of_expressions is None:
+            list_of_expressions = []
+        # First add all references. This way we can be sure that the writer
+        # will never rename a reference. This directory keeps track of which
+        # names are arrays (--> must be declared as a SymPy function) or
+        # non-array (--> must be declared as a SymPy symbol).
+        self._sympy_type = {}
```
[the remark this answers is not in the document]
hiker: OK.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
             self._intrinsic.add(op_str)
             self._op_to_str[operator] = op_str
 
+    def get_sympy_types(self):
```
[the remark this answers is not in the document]
hiker: Yes ... and no ... I did it, then at the very end realised that this function is not really used at all, so I have removed it completely.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        ''':returns: returns the mapping of symbols in the written
+            PSyIR expressions to SymPy types (Function or Symbol).
+        :rtype: dict of string to SymPy.Symbol or SymPy.Function
+
+        '''
+        return self._sympy_type
+
+    def member_node(self, node):
+        '''In SymPy a structure access `a%b` is handled as the 'MOD'
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        function `MOD(a, b)`. We have therefore make sure that a member
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        access is unique (e.g. `b` could already be a scalar variable).
+        This is done by creating a new name, which replaces the `%`
+        with an `_`. So `a%b` becomes `MOD(a, a_b)`. This makes it easier
+        to see where the function names come from.
+        Additionally, we still need to avoid a name clash, e.g. there
+        could already be a variable `a_b`. This is done by using a symbol
+        table, which was prefilled with all references (`a` in the example
+        above) in the constructor. We use the string with '%' as unique
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        tag and get a new, unqiue symbol from the symbol table based on the
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        new name using `_`. . For example, `a(i)%b` would get the tag
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        `a%b` and might get a symbol name like `a_b`, `a_b_1`, ...
+        '''
+
+        # We need to find the parent reference in order to make a new
+        # name (a%b%c --> a_b_c)
+        parent = node
+        tag = [node.name]
+        while not isinstance(parent, Reference):
+            parent = parent.parent
+            tag.insert(0, parent.name)
```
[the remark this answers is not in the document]
hiker: Done, but I agree, won't make much difference, I expect this list to be at most two or three elements.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        above) in the constructor. We use the string with '%' as unique
+        tag and get a new, unqiue symbol from the symbol table based on the
+        new name using `_`. . For example, `a(i)%b` would get the tag
+        `a%b` and might get a symbol name like `a_b`, `a_b_1`, ...
+        '''
+
+        # We need to find the parent reference in order to make a new
+        # name (a%b%c --> a_b_c)
+        parent = node
+        tag = [node.name]
```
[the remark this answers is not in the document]
hiker: I called it `name_list`. I would have preferred `member_list`, but the last/first element is not a member. Better suggestions welcome.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -71,6 +100,57 @@ def __init__(self):
...
+        while not isinstance(parent, Reference):
+            parent = parent.parent
+            tag.insert(0, parent.name)
+        # The root name uses _, the tag uses % (which are guaranteed
+        # to be unique, the root_name might clash with a user defined
+        # variable otherwise.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/core/symbolic_maths_test.py
hiker: Done. I think I've fixed up all modified files in this PR with f-strings.

src/psyclone/tests/core/symbolic_maths_test.py
hiker: Ooops.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
hiker: Done (though that argument is now elsewhere :) ).

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: OK, I've implement your original suggestion:
```
map = SympyWriter.create_type_map(expr)
writer = SymPyWriter(map)
lhs = writer(expr1)
rhs = writer(expr2)
...
str_exp1 = parse_expr(str_exp1, local_dict=map)
```

But then also added as discussed a static convenience function. And while I originally let this function return strings (i.e. sympy-formatted representation of the PSyIR), I realised that I can take it a step further and let this convenience function return sympy expressions. The sympy-writer already imported sympy, so this change did not add a new dependency, and it also means that the `SymbolicMaths` class is (slightly) more portable. Also, if at some stage we should decide (e.g. for performance reasons) to avoid using a writer based on Fortran completely, we have one location where we could try to create the sympy tree based on PSyIR tree, without changing anything else (IIRC sympy has this functionality).

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: Done as well - though tbh, now that everything is hidden behind this convenience function, I am not sure if it helps that much :)  But since it was suggested in the review, I left it in place :)

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: I've added a comment describing that this call can modify the type_map.

src/psyclone/tests/psyir/nodes/member_test.py
hiker: Done.

hiker:
I think I have addressed the issues, though I have taken the static convenience function a step further (by returning sympy instead of just strings). I have also decided to remove the `get_type_map` function since it's not needed at this stage.

hiker:
Ready for next review.

hiker:
@arporter - I did get your 'slow' file, I will try in the next day or so to see how it performs.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -94,7 +95,7 @@ def symbol(self, symbol):
         if not isinstance(symbol, Symbol):
             raise TypeError(
                 "The Reference symbol setter expects a PSyIR Symbol object "
```
arporter, line 97: `f` please.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -49,6 +49,9 @@
 def test_sym_writer_boolean():
     '''Test that booleans are written in the way that SymPy accepts.
     '''
+    sympy_writer = SymPyWriter({})
+    lit = Literal("true", BOOLEAN_TYPE)
+    # Also test that not specifying a type map as argument works:
```
arporter, line 54: I think you're missing a line that tests `sympy_writer(lit)` here?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -132,17 +137,45 @@ def test_sym_writer_functions(fortran_reader, expressions):
...
 
 
+@pytest.mark.parametrize("expressions", [("i", {'i': Symbol('i')}),
+                                         ("f(1)", {'f': Function('f')}),
+                                         ("f(:)", {'f': Function('f')}),
+                                         ("a%b", {'a': Symbol('a')}),
+                                         ("a%b(1)", {'a': Symbol('a')})
+                                         ])
+def test_sympy_writer_create_type_map(expressions, fortran_reader):
+    '''Tests that the static create_symbol_table creates a dictionary
```
arporter, line 159: s/create_symbol_table/create_type_map/ I think?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -197,15 +231,42 @@ def test_sym_writer_symbol_types(fortran_reader, expressions):
...
+    sympy_writer = SymPyWriter(type_map)
+    # Note that this call can extend the type_map with type information
+    # about member names.
     _ = sympy_writer(expr)
-    assert sympy_writer.get_sympy_types() == expressions[1]
+    assert sympy_writer._sympy_type_map == expressions[1]
+
+
+def test_sym_writer_convert_to_sympy_expressions(fortran_reader):
+    '''Tests that convenience function `write_as_sympy_strings works
```
arporter, line 252: Should be `convert_to_sympy_expressions` ?

arporter (changes requested):
Looks good now - just a few minor things.
Docs build fine.

arporter:
Forgot to say: please could you also resolve the conflict with master?

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -100,49 +103,122 @@ def __init__(self, list_of_expressions=None):
...
+        # Create a SymPy writer that uses this type_map
+        writer = SymPyWriter(type_map)
+        expression_str_list = []
+        for expr in list_of_expressions:
+            # Convert each expression. Note that this call might add
+            # additional entries to type_map if it finds member names
+            # that clash with a symbol (e.g. a%b --> it will try to
+            # create a SymPy symbol `a_b`, but if `a_b` is the same as
+            # symbol, `a_b_1`, ... will be used instead).
+            expression_str_list.append(writer(expr))
```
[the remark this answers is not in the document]
hiker: I just tried, but at this stage sympy does not handle array expressions (basically anything with a : in it - #1537). I tried some rough replacing of : expressions with constants, but ended up with various Fortran problems and/or psyclone crashes (invalid Fortran ... using where or elsewhere without an actual array expression:
```
../psyir/frontend/fparser2.py:3852: in _module_handler
    self.process_declarations(container, spec_part.children,
../psyir/frontend/fparser2.py:1975: in process_declarations
    self._process_decln(parent, parent.symbol_table, node,
>       return root_loop
E       UnboundLocalError: local variable 'root_loop' referenced before assignment

../psyir/frontend/fparser2.py:3070: UnboundLocalError
```
Well, I did have some limited success: it takes a bit over 3 seconds to parse this file without doing anything. I then added:
```
    for i in psyir.walk(ArrayReference):
        print("I", i)
        assert sym_maths.equal(i, i) is True
```
So basically comparing any ArrayReference with itself, and it took 31 seconds to process 58 expressions before it crashed (due to an array expression). When I used `writer._visit(...)` instead of just `writer(...)` ... the time went down to 3.7 seconds. Gee, HUGE difference.

Since you have used the _visit work around, I will use the same here then! Very good catch, thanks !!

doc/developer_guide/conf.py
```diff
@@ -285,7 +285,7 @@
     'Sergi Siso'
 epub_publisher = u'Rupert Ford, Joerg Henrichs, Iva Kavcic, Andrew Porter ' \
     'and Sergi Siso'
-epub_copyright = u'2017-2020, STFC'
+epub_copyright = u'2017-2022, STFC'
```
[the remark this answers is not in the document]
hiker: I've done that, and updates all three conf.py files. FWIW, epub_copyright defaults to copyright anyway (according to the doc), and I did not manage to create a document that actually uses epub_copyright (I did create an epub file, and my reader did not find a the `epub_copyright` string anywhere). But since it can't hurt to have this explicit, I left it all in.

doc/developer_guide/sympy.rst
```diff
@@ -172,15 +173,22 @@ following, SymPy specific features:
...
+   as ``MAX`` etc., which SymPy does not recognise and would then handle
    as unknown functions.
 
 .. autoclass:: psyclone.psyir.backend.sympy_writer.SymPyWriter
     :members:
+
+.. note::
+    The SymPyWriter class provides the static function
+    ``convert_to_sympy_expressions`` which hides the complexities of the
+    conversion from PSyIR expressions to a SymPy expressions. It is
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -59,32 +60,34 @@ class SymPyWriter(FortranWriter):
     can be found in the manual:
     https://psyclone-dev.readthedocs.io/en/latest/sympy.html#internal-details
 
+    :param type_map: Optional initial mapping that contains the SymPy data \
+        type of each reference in the expressions. This is the result of the \
+        static function \
+        :py:meth:`psyclone.core.sympy_writer.create_type_map`.
+    :type type_map: dictionary of string:Sympy-data-type values
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -100,49 +103,122 @@ def __init__(self, list_of_expressions=None):
...
         '''
-        return self._sympy_type
+        This function takes a list of PSyIR expressions, and converts
+        them all into Sympy expressions using the SymPy parser.
+        It takes care of all Fortran specific conversion required (e.g.
+        constants with kind specification, ...), including the renaming of
+        member accesses, as described in
+        https://psyclone-dev.readthedocs.io/en/latest/sympy.html#internal-details
+
+        It returns the converted PSyIR expressions as a list of SymPy
```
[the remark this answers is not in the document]
hiker: OK.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -100,49 +103,122 @@ def __init__(self, list_of_expressions=None):
...
+        expressions.
+
+        :param list_of_expressions: the list of expressions which are to be \
+            converted into SymPy-parsable strings.
+        :type list_of_expressions: list of \
+            :py:class:`psyclone.psyir.nodes.Node`
+        :returns: a list of SymPy expressions that are the conversion from
```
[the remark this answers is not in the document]
hiker: I tried ... very hard :)

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -100,49 +103,122 @@ def __init__(self, list_of_expressions=None):
...
+
+        # Create a SymPy writer that uses this type_map
+        writer = SymPyWriter(type_map)
+        expression_str_list = []
+        for expr in list_of_expressions:
+            # Convert each expression. Note that this call might add
+            # additional entries to type_map if it finds member names
+            # that clash with a symbol (e.g. a%b --> it will try to
+            # create a SymPy symbol `a_b`, but if `a_b` is the same as
+            # symbol, `a_b_1`, ... will be used instead).
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -100,49 +103,122 @@ def __init__(self, list_of_expressions=None):
...
+        above) in the constructor. We use the string containing the '%' as
+        a unique tag and get a new, unique symbol from the symbol table
+        based on the new name using `_`. For example, the access to member
+        `b` in `a(i)%b` would result in a new symbol with tag `a%b` and a
+        name like `a_b`, `a_b_1`, ...
 
+        :param node: a Member PSyIR node.
+        :type node: :py:class:`psyclone.psyir.nodes.Member`
+
+        :returns: the code as string
```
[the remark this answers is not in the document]
hiker: OK.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -173,8 +249,8 @@ def literal_node(self, node):
             return node.value.capitalize()
 
         if node.datatype.intrinsic == ScalarType.Intrinsic.CHARACTER:
-            raise TypeError("SymPy cannot handle strings like '{0}'."
-                            .format(node.value))
+            raise TypeError("SymPy cannot handle strings "
```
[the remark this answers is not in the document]
hiker: Yes, I noticed. All done.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -110,13 +111,13 @@ def _validate_index(self, index):
         '''
         if not isinstance(index, int):
             raise TypeError(
-                "The index argument should be an integer but found '{0}'."
-                "".format(type(index).__name__))
+                "The index argument should be an integer but found "
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -301,16 +302,16 @@ def indices(self):
         '''
         if not self._children:
             raise InternalError(
-                "{0} malformed or incomplete: must have one or more "
-                "children representing array-index expressions but found "
-                "none.".format(type(self).__name__))
+                f"{type(self).__name__} malformed or incomplete: must have "
+                "one or more children representing array-index expressions "
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -301,16 +302,16 @@ def indices(self):
...
         for idx, child in enumerate(self._children):
             if not self._validate_child(idx, child):
                 raise InternalError(
-                    "{0} malformed or incomplete: child {1} must by a psyir."
-                    "nodes.DataNode or Range representing an array-index "
-                    "expression but found '{2}'".format(
-                        type(self).__name__, idx, type(child).__name__))
+                    f"{type(self).__name__} malformed or incomplete: child "
+                    f"{idx} must by a psyir.nodes.DataNode or Range "
+                    "representing an array-index expression but "
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/nodes/member.py
```diff
@@ -75,10 +75,9 @@ def __init__(self, member_name, parent=None):
         if (parent and
                 not isinstance(parent, (StructureReference, StructureMember))):
             raise TypeError(
-                "The parent of a {0} must be either a "
+                f"The parent of a {type(self).__name__} must be either a "
                 "(ArrayOf)Structure(s)Reference or (ArrayOf)Structure(s)Member"
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/nodes/reference.py
hiker: Done.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: I think I actually have a duplicated line - the two required tests are just following in the next two lines:
```
    lit = Literal("true", BOOLEAN_TYPE)
    assert sympy_writer(lit) == "True"
    lit = Literal("false", BOOLEAN_TYPE)
    assert sympy_writer(lit) == "False"
````
I've removed the duplicated line.

File: src/psyclone/tests/psyir/backend/sympy_writer_test.py (hunk 36 above)
Comment (hiker) [reply src/psyclone/tests/psyir/backend/sympy_writer_test.py]: Indeed.

File: src/psyclone/tests/psyir/backend/sympy_writer_test.py (hunk 37 above)
Comment (hiker) [reply src/psyclone/tests/psyir/backend/sympy_writer_test.py]: Yes.

**Comment by hiker** [comment]
Updated to latest master, ready for next review.

File: doc/developer_guide/conf.py (hunk 39 above)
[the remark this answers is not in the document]
Comment (arporter) [reply doc/developer_guide/conf.py]: That's great, thanks.

File: src/psyclone/tests/domain/common/algorithm/alg_psyir_test.py [hunk 51]
```diff
@@ -296,12 +301,7 @@ def test_aic_createpsylayersymbolrootnames():
     assert invoke._psylayer_routine_root_name is None
     assert invoke._psylayer_container_root_name is None
 
-    routine_name = "invoke_0_kern"
-    routine_symbol = invoke.psylayer_routine_symbol
-    assert isinstance(routine_symbol, RoutineSymbol)
-    assert routine_symbol.name == routine_name
-    container_symbol = routine_symbol.interface.container_symbol
-    assert container_symbol.name == f"{routine_name}_mod"
+    invoke.create_psylayer_symbol_root_names()
```
arporter, line 304: What's the reason for this change?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
arporter: I can't see where you've done that (maybe due to a merge confusion?)

src/psyclone/psyir/backend/sympy_writer.py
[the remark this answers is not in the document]
arporter: Great. Please could you add a TODO for #1587 here too.

arporter (changes requested):
Almost, almost there. I just want to check why some new changes have appeared and, while we're here, I've requested that you add a TODO to the newly-created #1587.
All docs build fine and have updated copyright dates :-)

src/psyclone/tests/domain/common/algorithm/alg_psyir_test.py
hiker: According to `git blame` that was Rupert doing this, not me. I'd guess a side effect of merging in the latest master??

hiker:
I somehow can't answer the comment about the sympy writer test. So here I go:

>     # Also test that not specifying a type map as argument works:
> 
> I can't see where you've done that (maybe due to a merge confusion?)

Well, if there would be an error when not specifying any arguments, the code would crash and the test would fail. I could add an 'assert sympy_writer`   (or `sympy_writer is not None``) ?? But `sympy_writer` is used just below, so it would fail if `sympy_writer` would find a way to be None.

src/psyclone/psyir/backend/sympy_writer.py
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: I have no idea what had happened here, I am pretty certain that I removed the duplicated line. I have just redone this, making one separate test for the constructor, and removing the duplicated line. I hope it should be all ok now.

hiker:
OK, not sure about the previous comment about the thing you cannot see (removed duplicate line I think you meant). As I've said before, I've redone this, and split it into two tests (one for constructor, one for boolean - that was pretty hacky previously, sorry).

src/psyclone/tests/psyir/backend/sympy_writer_test.py
arporter: Ah, I see. I'd misunderstood!

src/psyclone/tests/domain/common/algorithm/alg_psyir_test.py
arporter: OK.

arporter (approved):
All good now.
Tests and examples all still OK with compilation.
All docs build fine.

Issue #1524, opened 2021-11-26
gocean1p0_test: test05p1_kernel_add_iteration_spaces does not compile


hiker:
```
Compiling: gfortran psy.f90 -I /tmp/pytest-of-joerg/pytest-5832/dl_esm_inf0 -c
psy.f90:7:10:

    7 |       USE kernel_invalid_iterates_over, ONLY: compute_cu_code
      |          1
Fatal Error: Cannot open module file ‘kernel_invalid_iterates_over.mod’ for reading at (1): No such file or directory
compilation terminated.
```
Renaming the module name of kernel_invalid_iterates_over.f90 from 'compute_cu_mod' to 'kernel_invalid_iterates_over' should fix this.

PR #1539, opened 2021-12-07
#1524 Fixed compilation error (that caused this compilation test to be removed recently).

hiker:
Small three liner.


sergisiso (approved):
All tests pass with compilation now. No TODOs of this issue are remaining, pylint is clean and no new documentation is needed. Thanks for fixing this @hiker

Issue #1396, opened 2021-09-14
Managing data movement: add information to data directive nodes and mv clause generation to visitor backend


arporter:
In #1385 Sergi pointed out that the current approach (of generating the code for the clauses of an ACC DATA directive in a `begin_string` method inside the node) isn't really consistent with the visitor/backend code-generation step. He also suggested that we need to capture the variables that will be moved by a DATA directive as part of the Directive node (at the moment, we only generate this list at code-generation time). This will then allow us to examine neighbouring data Directives and optimise between them - i.e. removing unnecessary data movement.

arporter:
Sergi also mentioned that in C++, giving a bare Reference to a DATA directive isn't sufficient because it's just a pointer without associated extent information. Therefore it has to be of the form `my_array[0:N]`.

arporter:
Since LFRic currently still relies on the `gen_code` code-generation path, any solution here needs to support both it and the visitor/backend path.

arporter:
I've now added some ACC clause classes and am experimenting with using them with ACCDataDirective. It almost worked out of the box which is very nice. However, now that I look at the code we currently have, it won't cope with complex structures where more than one Member is an array. From Dave Case's report on the NEMOVAR work, they had to do things like:

    DO j_var = 1, pjb%fbge_vars%n3d 

      i_var = pjb%fbge_vars%getVar3d( j_var ) 

      !$ACC ENTER DATA IF(ll_transfer) & 

      !$ACC COPYIN(pjb%fbge%fbgemod_3d(i_var)) & 

      !$ACC COPYIN(pjb%fbge%fbgemod_3d(i_var)%fcor) 

      DO j_cov = 1, pjb%fbge%fbgemod_3d(i_var)%ncov 

      !$ACC ENTER DATA IF(ll_transfer)  COPYIN(pjb%fbge%fbgemod_3d(i_var)%fcor(j_cov)%fdif%cofsym_ipjk)

arporter:
i.e. it can't be done using a single DATA directive. For now I could simply refuse to deal with such cases but that's just delaying the inevitable. Perhaps the right answer is to let clauses contain array-range expressions and then expand these into explicit loops in a bespoke `lower_to_language_level` method? However, for the example above that would require e.g. `pjb%fbge%fbgemod_3d(:)%fcore(:)%fdif%cofsym_ipjk` and we've recently decided to forbid such things in the PSyIR because they aren't valid in Fortran.

arporter:
[...]

PR #1554, opened 2022-01-04
(closes #310) add support for acc enter data and acc update to NEMO API

### LINKED ISSUE #310: [nemo] add support for ACCEnterDataTrans
Currently ACCEnterDataTrans is only supported for the GOcean API. We will need it for the NEMO API if we are to generate efficient OpenACC code.

arporter:


nmnobre:
I've quite radically reworked `ACCUpdateDirective`. The constructor no longer takes a list of signatures, but rather a list of nodes from which a new method in the class, `lower_to_language_level()`, is responsible to extract the required signatures. I first tried to assign this list of nodes as the children of the directive, but as it inherits from `StandaloneDirective()`, which can't have any children, it didn't quite work... thus the new argument `nodes`. The tests are currently broken.
Do we have to keep `children` as a constructor argument even for classes which disallow them?

arporter:
Hi @nmnobre, yes, all node constructors have to take `children`. In this case the children will need to be `Clause` nodes, in a similar way to what I am doing in #1396. This may be better done in a `create()` method rather than through the constructor.

nmnobre:
> Hi @nmnobre, yes, all node constructors have to take `children`. In this case the children will need to be `Clause` nodes, in a similar way to what I am doing in #1396. This may be better done in a `create()` method rather than through the constructor.

Ah yes, that's right. We are changing the way we think about these. Even standalone directives will have children now!...

nmnobre:
In hindsight, I think I made the _wrong_ decision...

I tried to make the `update` directive implementation mimic that of `enter data`, but I should've done the opposite...
We shouldn't assume `enter data` will always take all the data in `parallel` and `kernels` regions within the routine. The decision of what data the directive takes should belong to the transformation class which is context aware.... Back to the drawing board.

arporter:
> We shouldn't assume enter data will always take all the data in parallel and kernels regions within the routine.

What's your thinking here? My assumption was that we should assume that as we know it's going to be needed.

nmnobre:
> What's your thinking here? My assumption was that we should assume that as we know it's going to be needed.

I agree, it's eventually going to be needed, but might we not be too eager? In the sense a CPU region might precede the first GPU usage.

nmnobre:
@arporter OpenACC enter data directives are now placed just before the first compute construct, causing some tests to fail. What's the right way to address this? Should we change the test or return the directive to its original position for some of the APIs?

arporter:
> OpenACC enter data directives are now placed just before the first compute construct, causing some tests to fail. What's the right way to address this? Should we change the test or return the directive to its original position for some of the APIs?

I don't mind. If the new location is valid and it's easiest to do it that way then I'm happy for the tests to be modified.

nmnobre:
I've tried tidying the code a bit but:
- If there's an `acc_directives` module, shouldn't we switch to use an `acc_transformations` module too?
- If the variables for the `enter data` directive are collected in the directive's code, why are the variables for the `update` directive collected in the transformation's code?
- I'd rather make the NEMO API, "THE OTHER GENERIC" API, dropping the NEMO moniker and making Dynamo and GOcean the exceptions. I've had to add if conditionals specific for NEMO which don't really depend on NEMO, but rather on not being Dynamo or GOcean... (for some reason for some GOcean tests using `ACCRegionDirective.kernel_references`, checking if the schedule is a `GOceanInvokeSchedule` doesn't work so I resorted to checking if it is a `NemoInvokeSchedule` instead)....

arporter:
Thanks @nmnobre. re. your first point, we're still in a state of flux with regards to where transformations live. We used to have everything in `transformations.py`, then we had stuff under `nemo/transformations` and now, with the move to PSyIR, we have them under either `domain/xxx/transformations` or `psyir/transformations` depending on whether or not they are API specific.
If we're introducing a new transformation here (I can't remember!) then we should put that in the appropriate location. If not then our practise is to only move them in PRs that are restricted to just that change (otherwise it's hard to review).

arporter:
re. your second point, that sounds like an oversight on my part. In order to be able to reason about the tree we're trying to ensure that Nodes hold all necessary information rather than delaying to code-generation time (as we used to). Since the list of variables may need to be updated in light of subsequent transformations we probably want the functionality that generates it to be a part of the Node itself. Does @sergisiso agree?

arporter:
re. your third point, that sounds like something that needs further discussion. If we've got a lot of conditionals then probably we need to restructure some classes as things that are domain specific should be in domain specific classes (ideally).
Finally, your fourth point also needs more discussion for me to understand. I think we've got sufficient dependence analysis (with the exception of CodeBlocks).

hiker:
> re. your third point,
...
I can't comment on the if-statements, but I think it would be great to rename NEMO to ... something else. We might appear to people interested in PSyclone  to not really support generic codes, only NEMO. We could provide different config files. e.g. psyclone-nemo.cfg, psyclone-roms.cfg, ... and a generic one (which might just be the nemo one :) ).

We could call it .... 'evolution'  :) 

Time for a separate ticket for that?

nmnobre:
Hi @arporter, as discussed, this should now be ready for a first look 😊

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+            elif idx == OUT:  # outputs
+                node_index = -1
+                node_offset = 1
+                direction = "device"
+                tentative_update_pos = "end"
+
+            # Since the supplied nodes may be the children of an IfBlock or
+            # Loop, we have to search up to find the ancestor that *is* a
+            # child of a Schedule.
+            child = node_list[node_index]
```
arporter, line 219: You can do `sched = node_list[node_index].ancestor(Schedule)`

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+                    self._place_update(sched, update_pos, loop_sync, direction)
+                    cpu_sig.difference_update(loop_sync)
+
+                # This schedule is the body of a routine and, at least until we
+                # can do interprocedural analysis, this is the end of the road.
+                if isinstance(sched, InvokeSchedule):
+                    self._place_update(sched, update_pos, cpu_sig, direction)
+                    break
+
+                while True:
```
arporter, line 283: Use `ancestor` again here?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+            # child of a Schedule.
+            child = node_list[node_index]
+            sched = child.parent
+            while not isinstance(sched, Schedule):
+                child, sched = child.parent, sched.parent
+
+            while cpu_sig:
+                update_pos = sched.children.index(child) + node_offset
+
+                if idx == IN:
```
arporter, line 227: Could you set these up at lines 205-214 instead?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+            # Since the supplied nodes may be the children of an IfBlock or
+            # Loop, we have to search up to find the ancestor that *is* a
+            # child of a Schedule.
+            child = node_list[node_index]
+            sched = child.parent
+            while not isinstance(sched, Schedule):
+                child, sched = child.parent, sched.parent
+
+            while cpu_sig:
+                update_pos = sched.children.index(child) + node_offset
```
arporter, line 225: Please add a comment describing what `update_pos` is for.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+
+                if idx == IN:
+                    beg, end = None, update_pos
+                elif idx == OUT:
+                    beg, end = update_pos, None
+
+                text_dep_stmts = sched.children[beg:end]
```
arporter, line 232: Comments please!

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+
+                loop_dep_stmts = sched.children[end:beg]
+                loop_sig, loop_sync = set(), set()
+
+                for dep_stmts, sig, sync in [
+                     (text_dep_stmts, text_sig, text_sync),
+                     (loop_dep_stmts, loop_sig, loop_sync)]:
+                    for stmt in dep_stmts:
+                        for acc in stmt.walk(self._acc_regions):
+                            # Workaround for lack of precise access
```
arporter, line 243: This sounds like it needs a TODO plus Issue number.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+        for node in node_list:
+            if isinstance(node, CodeBlock):
+                for symbol_name in node.get_symbol_names():
+                    inputs.add(Signature(symbol_name))
+                    outputs.add(Signature(symbol_name))
+
+        # Copy any data that is read by this region to the host (resp. device)
+        # if it is on the device (resp. host).
+        IN, OUT = 0, 1
+        for idx, cpu_sig in enumerate((inputs, outputs)):
```
arporter, line 202: This is a complex and deeply-nested looping structure which, if I understand it correctly, mainly exists to avoid code duplication for the 'IN' and 'OUT' cases. I suspect it can be made clearer by refactoring and adding some utility routines that are just called with appropriate arguments instead.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+                elif idx == OUT:
+                    beg, end = update_pos, None
+
+                text_dep_stmts = sched.children[beg:end]
+                text_sig, text_sync = set(), set()
+
+                loop_dep_stmts = sched.children[end:beg]
+                loop_sig, loop_sync = set(), set()
+
+                for dep_stmts, sig, sync in [
```
arporter, line 238: As per my earlier comment, please can you refactor this to simplify the code structure?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+                ACCUpdateDirective
+
+        cpu_sig = cpu_sig.copy()
+
+        if update_pos == "beg":
+            update_pos = 0
+        elif update_pos == "end":
+            update_pos = len(sched.children)
+
+        for update_offset in (-1, 0):
```
arporter, line 316: Comment please :-)

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+                    if neighbour_node.direction == direction:
+                        neighbour_node.sig_set.update(cpu_sig)
+                        cpu_sig.clear()
+                    # Avoid updating the host just after updating the device.
+                    else:
+                        cpu_sig.difference_update(neighbour_node.sig_set)
+            except IndexError:
+                pass
+
+        if cpu_sig:
```
arporter, line 330: Comments please :-)

src/psyclone/transformations.py
```diff
@@ -2812,7 +2812,7 @@ def apply(self, sched, options=None):
         :param sched: schedule to which to add an "enter data" directive.
         :type sched: sub-class of :py:class:`psyclone.psyir.nodes.Schedule`
         :param options: a dictionary with options for transformations.
-        :type options: dictionary of string:values or None
+        :type options: dict of str:values or None
```
arporter, line 2815: `Optional[Dict[str, str]]` please.

src/psyclone/transformations.py
```diff
@@ -2825,15 +2825,28 @@ def apply(self, sched, options=None):
...
+        elif isinstance(sched, NemoInvokeSchedule):
+            from psyclone.nemo import NemoACCEnterDataDirective as \
+                AccEnterDataDir
         else:
             # Should not get here provided that validate() has done its job
             raise InternalError(
                 f"ACCEnterDataTrans.validate() has not rejected an "
                 f"(unsupported) schedule of type {type(sched)}")
 
+        # Determine where to place the directive
```
arporter, line 2837: Please extend this comment to explain a bit about the algorithm used - i.e. what is driving the choice of location for the directive?

src/psyclone/transformations.py
```diff
@@ -2844,29 +2857,22 @@ def validate(self, sched, options=None):
         :param sched: Schedule to which to add an "enter data" directive.
         :type sched: sub-class of :py:class:`psyclone.psyir.nodes.Schedule`
         :param options: a dictionary with options for transformations.
-        :type options: dictionary of string:values or None
+        :type options: dict of str:values or None
```
arporter, line 2860: `Optional[Dict[str, str]]` please.

src/psyclone/transformations.py
```diff
@@ -2844,29 +2857,22 @@ def validate(self, sched, options=None):
...
-                f" a schedule of type {type(sched)}")
+            raise TransformationError("Cannot apply an OpenACC enter data "
+                                      "directive to something that is not a "
+                                      "Schedule")
 
         # Check that we don't already have a data region of any sort
-        directives = sched.walk(Directive)
-        if any(isinstance(ddir, (ACCDataDirective, ACCEnterDataDirective))
-               for ddir in directives):
+        directive_cls = (ACCDataDirective, ACCEnterDataDirective)
```
arporter, line 2874: Thanks for improving.

arporter (commented):
Many thanks for this.
I've gone through all of the code changes so far but not the tests.
I have to admit that I found some of the code hard to understand so you'll see I've requested some refactoring and comments :-) Next time around I will hopefully be able to get to grips with what it's doing and thus check that it's OK.
Having written that, I realise that we also need to document what's being done as it is significant functionality. I think the place to do this is the OpenACC section of the Developer Guide (https://psyclone-dev.readthedocs.io/en/latest/transformations.html#openacc). Please could you describe the strategy used - i.e. the combination of Enter Data and Update and the algorithm used to place them. That will then be a big help to me in reviewing the code too.

I've also just realised that I'm the original author of this PR so GitHub isn't going to let me approve it. I suggest responding to all of my current comments and once that's done we can create a new PR.

sergisiso:
> I've also just realised that I'm the original author of this PR so GitHub isn't going to let me approve it. I suggest responding to all of my current comments and once that's done we can create a new PR.

Alternatively, @arporter when you finish your review I can do a quick read and finish the approval myself. I would like to be aware of what is being done regarding data directives anyway so it would be useful for me.

src/psyclone/nemo.py
```diff
@@ -327,3 +329,39 @@ def kernel(self):
                     f"{len(kernels)}")
             return kernels[0]
         return None
+
+
+class NemoACCEnterDataDirective(ACCEnterDataDirective):
```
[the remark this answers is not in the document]
nmnobre: Where exactly should they be placed? Since `src/psyclone/domain/nemo` includes a `transformations` directory, should I create a `directives` directory too?
The fact they are the same was a very conscious design decision: I like the code of things that behave the same to look the same. How should I go about implementing `NemoDataMovementMixin`? You mean by keeping the `lower_to_language_level` specialisations for both classes and calling a static method in `NemoDataMovementMixin` to remove the loop variables?

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -104,13 +104,37 @@ def kernel_references(self):
...
+            for call in self.kernels():
+                for arg in call.arguments.acc_args:
+                    sig_set.add(arg)
+        else:
+            sig_set = set().union(self.in_kernel_references,
+                                  self.out_kernel_references)
+        return sig_set
+
+    @property
+    def in_kernel_references(self):
```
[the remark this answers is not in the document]
nmnobre: I used them in `ACCUpdateTrans._add_update_directives`, should I still remove them?

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -206,20 +225,18 @@ def lower_to_language_level(self):
...
-            self._acc_dirs = self.ancestor(InvokeSchedule).walk(
-                    (ACCParallelDirective, ACCKernelsDirective))
-            # 2. For each directive, add the fields used by the kernels it
-            # contains (as given by kernel_references) and add it to our set.
-            # TODO GOcean grid properties are duplicated in this set under
-            # different names (the OpenACC deep copy support should spot this).
-            for pdir in self._acc_dirs:
-                self._sig_list.update(pdir.kernel_references)
-            self._node_lowered = True
+        # We must generate a list of all of the fields accessed by OpenACC
```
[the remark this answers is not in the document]
nmnobre: Ah, so that's what the conditional was for! 😝
Since `_sig_set ` is a set, it should just stay the same.
Besides the performance penalty of reinterpreting and rerunning the code, there shouldn't be a problem?

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -751,45 +772,53 @@ class ACCUpdateDirective(ACCStandaloneDirective):
...
+    :param signatures: the access signature(s) that need to be synchronised \
+                       with the accelerator.
+    :type signatures: Set[:py:class:`psyclone.core.Signature`]
     :param str direction: the direction of the synchronisation.
     :param children: list of nodes which the directive should have as children.
     :type children: List[:py:class:`psyclone.psyir.nodes.Node`]
     :param parent: the node in the InvokeSchedule to which to add this \
                    directive as a child.
     :type parent: :py:class:`psyclone.psyir.nodes.Node`
+    :param bool conditional: whether or not to include the 'if_present' clause
```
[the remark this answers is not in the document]
nmnobre: There is an example of such a type specification in `Node.reductions`. Is that written correctly, i.e. should I write `:type param: Optional[bool]`, or should I instead write `:type conditional: Optional[bool]`?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+            if isinstance(child, ACCEnterDataDirective):
+                continue
+            elif not child.walk(self._acc_regions + self._brk_nodes):
+                node_list.append(child)
+            else:
+                self._add_update_directives(node_list)
+                node_list.clear()
+                if isinstance(child, self._brk_nodes):
+                    # Conservatively add an update host statement just before
+                    # the Call node for fear of temporary operands.
```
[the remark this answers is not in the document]
nmnobre: I could for pass-by-reference languages, since even though there is the risk of passing a reference to outdated data, an `acc update host` directive in the callee would always be able to update the host since the data location is preserved.  However, since `Call` nodes may represent function calls in pass-by-value languages where an `acc update host` directive in the callee would do nothing, I must guarantee the host has already been updated in the caller.
I must admit I wrote the comment thinking about Fortran alone, but I can update it to reflect what I just said?

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: How would I then get the child of `sched` which is an ancestor of `child`?

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Ditto.

src/psyclone/psyir/nodes/acc_directives.py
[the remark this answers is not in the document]
nmnobre: Done?

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -813,14 +842,14 @@ def direction(self):
         return self._direction
 
     @property
-    def symbol(self):
+    def sig_set(self):
         '''
-        Returns the symbol to synchronise with the accelerator.
+        Returns the set of symbols to synchronise with the accelerator.
```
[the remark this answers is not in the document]
nmnobre: Done

src/psyclone/psyir/transformations/__init__.py
```diff
@@ -82,7 +83,8 @@
 # this package e.g.:
 # from psyclone.psyir.transformations import ExtractTrans
 
-__all__ = ['ArrayRange2LoopTrans',
+__all__ = ['ACCUpdateTrans',
```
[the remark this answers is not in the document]
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2021, Science and Technology Facilities Council.
```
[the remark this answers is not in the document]
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: A. R. Porter and N. Nobre, STFC Daresbury Lab
+
+'''
+This module provides the ACCUpdateTrans transformation that ensures that
+data is kept up-to-date on the host.
+'''
+
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def validate(self, node, options=None):
+        '''
+        Checks that it is valid to apply this transformation to the supplied
+        schedule.
+
+        :param node: the Schedule that is to be transformed.
+        :type node: :py:class:`psyclone.psyir.nodes.Schedule`
+        :param options: any options to this transformation.
+        :type options: dict
```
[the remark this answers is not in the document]
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+        '''
+        Applies this transformation to the supplied Schedule. Identifies any
+        regions of code outside of ACC regions and adds the necessary ACC
+        update directives to ensure that the host and device copies of any
+        variables are kept up-to-date.
+
+        :param node: the Schedule that is to be transformed.
+        :type node: :py:class:`psyclone.psyir.nodes.Schedule`
+        :param options: any options to this transformation.
+        :type options: dict
```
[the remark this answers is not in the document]
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+            if isinstance(node, CodeBlock):
+                for symbol_name in node.get_symbol_names():
+                    inputs.add(Signature(symbol_name))
+                    outputs.add(Signature(symbol_name))
+
+        # Copy any data that is read by this region to the host (resp. device)
+        # if it is on the device (resp. host).
+        IN, OUT = 0, 1
+        for idx, cpu_sig in enumerate((inputs, outputs)):
+            if not cpu_sig:
```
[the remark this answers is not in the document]
nmnobre: `Call` nodes have no outputs and `CodeBlock` nodes might have neither inputs nor outputs if writing a string literal for instance.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: `update_pos` is being updated at every loop iteration though

src/psyclone/transformations.py
nmnobre: Done

src/psyclone/transformations.py
nmnobre: Done

src/psyclone/transformations.py
nmnobre: Done

src/psyclone/transformations.py
nmnobre: No worries.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: I've improved the comment instead, it wasn't accurate.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done

src/psyclone/psyir/transformations/acc_update_trans.py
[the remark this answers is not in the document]
nmnobre: Even then, the check is useless because of the loop some twenty lines later which is checking the exact same. Removed.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done?

arporter:
Hi @nmnobre, is this ready for another look now?

nmnobre:
> Hi @nmnobre, is this ready for another look now?

Hmmm.... I haven't really addressed all of your comments yet...
I don't know if you'd rather see it all at once or if you're happy to close the conversations I've addressed....

arporter:
Unless there are things with wide-ranging effects, I'd suggest doing the whole thing so we (I) don't get confused :-)

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -832,7 +861,29 @@ def begin_string(self):
...
+
+
+def _sig_set_to_string(sig_set):
+    # TODO this is Fortran specific
+    names = set()
+    for sig in sig_set:
+        if isinstance(sig, str):
+            names.add(sig)
+        elif isinstance(sig, Signature):
+            names.update({"%".join(sig[:idx+1]) for idx in range(len(sig))})
```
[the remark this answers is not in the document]
nmnobre: Please see #1847.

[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -832,7 +861,29 @@ def begin_string(self):
...
+                "variables this directive was created to update.")
+
+        condition = "if_present " if self._conditional else ""
+        sym_list = _sig_set_to_string(self._sig_set)
+
+        return f"acc update {condition}{self._direction}({sym_list})"
+
+
+def _sig_set_to_string(sig_set):
+    # TODO this is Fortran specific
```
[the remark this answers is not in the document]
nmnobre: I _think_ changing to `Signature.to_language` might fix the issue as far as this piece of code is concerned?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+                                  ACCEnterDataDirective, ACCUpdateDirective,
+                                  ACCKernelsDirective, ACCParallelDirective)
+from psyclone.psyir.tools import DependencyTools
+
+
+class ACCUpdateTrans(Transformation):
+    '''
+    Examines the supplied Schedule and adds "acc update" directives
+    for any data accessed outside of a kernels or parallel region.
+
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -832,7 +861,29 @@ def begin_string(self):
         :rtype: str
 
         '''
-        return "acc update " + self._direction + "(" + self._symbol.name + ")"
+        if not self._sig_set:
+            # There should be at least one variable to update.
+            raise GenerationError(
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+        :raises TransformationError: if the supplied Schedule contains any \
+            CodeBlocks and the 'allow-codeblocks' option is False.
+        '''
+        # transformations module file needs moving into the psyir
+        # hierarchy.
+        # pylint: disable=import-outside-toplevel
+        from psyclone.transformations import TransformationError
+
+        if not isinstance(node, Schedule):
+            raise TransformationError(f"Expected a Schedule but got a node of "
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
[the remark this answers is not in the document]
nmnobre: Perhaps an even better implementation would change `kernel_references` to output a 2-tuple whose members are the outputs of `in_kernel_references` and `out_kernel_references` a bit like `DependencyTools.get_in_out_parameters`?

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -832,7 +861,29 @@ def begin_string(self):
...
+                "ACCUpdate directive did not find any data to update. "
```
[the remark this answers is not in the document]
nmnobre: Yes, I suppose so. It's the exact same we're already doing for `ACCEnterDataDirective`. I'll take the liberty of adding the same note there.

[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,332 @@
...
+
+        inputs, outputs = self._dep_tools.get_in_out_parameters(node_list)
+        inputs, outputs = set(inputs), set(outputs)
+
+        # Workaround for lack of precise access descriptions, e.g. in the
+        # presence of array slices we currently overapproximate dependencies.
+        inputs.update(outputs)
+
+        # Workaround for CodeBlocks
+        # TODO We are probably relying on undefined behaviour here.
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: I did add a related TODO earlier in the function though. Done.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Hopefully the new comments and documentation should clarify the algorithm and alleviate the need for this. If that's not the case I'll have a second look.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Ditto.

src/psyclone/transformations.py
nmnobre: It does have the advantage of potentially being placed after an update directive, potentially avoiding the copies incurred by the latter.

src/psyclone/psyir/transformations/acc_update_trans.py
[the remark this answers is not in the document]
nmnobre: Done.

nmnobre:
> Having written that, I realise that we also need to document what's being done as it is significant functionality. I think the place to do this is the OpenACC section of the Developer Guide (https://psyclone-dev.readthedocs.io/en/latest/transformations.html#openacc). Please could you describe the strategy used - i.e. the combination of Enter Data and Update and the algorithm used to place them. That will then be a big help to me in reviewing the code too.

Done?

nmnobre:
I feel this is now ready for a second look @arporter !

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,403 @@
...
+        for node in node_list:
+            if isinstance(node, CodeBlock):
+                for symbol_name in node.get_symbol_names():
+                    inputs.add(Signature(symbol_name))
+                    outputs.add(Signature(symbol_name))
+
+        # Copy any data that is read by this region to the host (resp. device)
+        # if it is on the device (resp. host).
+        IN, OUT = 0, 1
+        for idx, host_sig in enumerate((inputs, outputs)):
```
arporter, line 233: The comments have helped a lot but I still find this code pretty hard to understand and for future maintenance that's an important consideration. Please could you therefore try replacing this loop with, e.g. two successive calls to a new, internal method, once for inputs and once for outputs. That way, for instance, `idx` won't get carried through the code and used to make decisions here and there (hopefully).
I'm not totally certain of the best way to refactor - you'll have to try and it and see what works best and/or we can discuss it together.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,403 @@
...
+            # While there are variables from this host region which have not
+            # been placed in an update directive, we check if they need to be
+            # placed at the current schedule, in which case they are, or
+            # if we can bring them up to the next deepest ancestor schedule.
+            # The main goal is to avoid placing update directives inside loop
+            # statements, but the algorithm also tries to minimise the total
+            # number of update directives by promoting the coalescence of
+            # neighbouring directives.
+            while host_sig:
+                # Most conservative position for a new update directive at this
```
arporter, line 261: It could be that the body of this loop might also be best replaced with a method - it aids readability and testing.

arporter (commented):
Thanks Nuno. I've been through everything now.
Although the comments really help, I still find the core part of the code hard to understand and that worries me for future maintainability/extensibility. You'll see that I've therefore requested you try breaking it up into routines. These can be 'private' and thus not have to do type checking etc. but I think they will improve readability.
Apart from that, there are a few small things that need doing but we're getting there now... :-)

Documentation has been updated and is good (thanks).
pylint has spotted a few things that need tidying.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2021, Science and Technology Facilities Council.
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: R. W. Ford, A. R. Porter, S. Siso and N. Nobre, STFC Daresbury Lab
+
+'''Module containing py.test tests for the transformation of the PSy
+   representation of NEMO code using the OpenACC enterdata directive.
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+
+'''
+
+from __future__ import print_function, absolute_import
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+from psyclone.psyir.nodes import ACCKernelsDirective, CodeBlock, Schedule
+from psyclone.psyir.transformations import TransformationError, ACCUpdateTrans
+from psyclone.transformations import ACCEnterDataTrans, ACCKernelsTrans
+from psyclone.tests.utilities import Compile
+
+# Constants
+API = "nemo"
+
+
+def test_validate(parser):
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+SUBROUTINE tra_ldf_iso()
+  USE some_mod, only: dia_ptr_hst
+  INTEGER, PARAMETER :: jpi=2, jpj=2, jpk=2
+  INTEGER :: jn=2
+  REAL :: checksum
+  REAL, DIMENSION(jpi,jpj,jpk) :: zdit, zdjt, zftv, zftw
+  zftv(:,:,:) = 0.0d0
+  CALL dia_ptr_hst( jn, 'ldf', zftv(:,:,:)  )
+  checksum = SUM(zftv)
+  !write(*,*) checksum
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+'''
+    reader = FortranStringReader(code)
+    ast = parser(reader)
+    psy = PSyFactory(API, distributed_memory=False).create(ast)
+    schedule = psy.invokes.invoke_list[0].schedule
+    acc_update = ACCUpdateTrans()
+    acc_update.apply(schedule)
+    assert isinstance(schedule[1], CodeBlock)
+    gen_code = str(psy.gen).lower()
+    # TODO We are probably relying on undefined behaviour here
```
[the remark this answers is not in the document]
nmnobre: No, it was because, when I wrote that, I thought `Codeblock.get_symbol_names()` would also retrieve other symbols beyond variable names, like Fortran statements or called intrinsics, functions, subroutines. At the moment, we do incur the risk of collecting a function name mistaken by an array name, but that's already documented in the code, i.e. in `acc_update_trans.py`, including a reference to #1872. I've removed it as this codeblock calls no user-defined functions, rendering the comment irrelevant.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -267,25 +328,28 @@ def _add_update_directives(self, node_list):
                 # always exists as all outputs are inputs - has already been
                 # moved out as well. This happens iff there's no textually
```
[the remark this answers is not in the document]
nmnobre: Done, "if and only if".

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -213,52 +242,84 @@ def _add_update_directives(self, node_list):
...
+                # to be updated straight away. Indeed, for update device
+                # directives, but not quite for update host directives, where
+                # placing the directive earlier and adding the async clause
+                # might perform better.
                 self._place_update(sched, update_pos, text_sync, direction)
-                cpu_sig.difference_update(text_sync)
+                host_sig.difference_update(text_sync)
 
+                # All the remaining variables that need synchronisation at this
+                # schedule can placed in an update directive at the edge
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -213,52 +242,84 @@ def _add_update_directives(self, node_list):
...
                 elif idx == OUT:
                     beg, end = update_pos, None
 
-                text_dep_stmts = sched.children[beg:end]
-                text_sig, text_sync = set(), set()
-
-                loop_dep_stmts = sched.children[end:beg]
-                loop_sig, loop_sync = set(), set()
+                # The (in)ability to place variables in update directives in
+                # outer schedules is determined by the (in)existence of textual
```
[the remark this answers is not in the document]
nmnobre: Done, rephrased.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: I have refactored the code by creating a new internal method that finds the variables requiring synchronisation against a set of statements including compute regions. This allowed me to reduce the nesting depth to no more than three... Let me know what you think.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -0,0 +1,403 @@
...
+    def _add_update_directives(self, node_list):
+        '''
+        Adds the required update directives before and after the nodes in
+        the supplied list.
+
+        :param sched: the schedule which contains the nodes in node_list.
+        :type sched: :py:class:`psyclone.psyir.nodes.Schedule`
+        :param node_list: the PSyIR nodes representing code executed on the \
+                          host rather than the device.
+        :type node_list: list of :py:class:`psyclone.psyir.nodes.Node`
```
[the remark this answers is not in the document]
nmnobre: Done, and removed `sched` as well since it's no longer a parameter.

src/psyclone/tests/domain/nemo/transformations/enter_data_test.py
```diff
@@ -0,0 +1,100 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2021, Science and Technology Facilities Council.
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/enter_data_test.py
```diff
@@ -0,0 +1,100 @@
...
+# COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: R. W. Ford, A. R. Porter and S. Siso, STFC Daresbury Lab
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/enter_data_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+'''Module containing py.test tests for the transformation of the PSy
+   representation of NEMO code using the OpenACC enterdata directive.
+
+'''
+
+from __future__ import print_function, absolute_import
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/enter_data_test.py
```diff
@@ -0,0 +1,100 @@
...
+from psyclone.transformations import ACCEnterDataTrans, ACCKernelsTrans
+from psyclone.psyir.transformations import ACCUpdateTrans
+from psyclone.tests.utilities import Compile, get_invoke
+
+
+# Constants
+API = "nemo"
+# Location of the Fortran files associated with these tests
+BASE_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                         "../../test_files")
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+    reader = FortranStringReader(code)
+    ast = parser(reader)
+    psy = PSyFactory(API, distributed_memory=False).create(ast)
+    schedule = psy.invokes.invoke_list[0].schedule
+    acc_update = ACCUpdateTrans()
+    acc_kernels = ACCKernelsTrans()
+    acc_kernels.apply(schedule[0].if_body[0])
+    acc_update.apply(schedule)
+    gen_code = str(psy.gen).lower()
+    print(gen_code)
```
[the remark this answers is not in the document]
nmnobre: Done.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+            "    zftw(:,:,:) = -1.0d0\n"
+            "    !$acc end kernels\n"
+            "    zftv(1,:,:) = 1.0d0\n"
+            "  end if\n"
+            "  !$acc update if_present device(zftv)") in gen_code
+
+
+def test_if_update_device(parser):
+    ''' Test the placement of update device directives when an IfBlock contains
+     a host statement writing the same variable as a subsequent kernel. '''
```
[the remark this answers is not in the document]
nmnobre: It was definitely wrong, thank you.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+    acc_enter.apply(schedule)
+    acc_update.apply(schedule)
+    gen_code = str(psy.gen).lower()
+    # Loop variable should not be copied to device
+    assert "device(ji)" not in gen_code
+    assert ("  !$acc update if_present host(jpi,start,step,zftv)\n"
+            "  zftv(:,:,:) = 0.0d0\n"
+            "  !$acc enter data copyin(zftw)\n"
+            "  do ji = start, jpi, step\n"
+            "    !$acc update if_present host(zftw)\n"
```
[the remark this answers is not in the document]
nmnobre: That's a problem alright, but one we've already reported in `acc_update_trans.py` and #1872.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+    reader = FortranStringReader(code)
+    ast = parser(reader)
+    psy = PSyFactory(API, distributed_memory=False).create(ast)
+    schedule = psy.invokes.invoke_list[0].schedule
+    acc_update = ACCUpdateTrans()
+    acc_update.apply(schedule)
+    gen_code = str(psy.gen).lower()
+    # Loop variable should not be copied to device
+    assert "device(ji)" not in gen_code
+    # TODO Currently jpi is copied to and from. We need a list of variables
```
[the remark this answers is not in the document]
nmnobre: Done, added to #1872.

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+    acc_kernels.apply(schedule[0])
+    acc_kernels.apply(schedule[1].if_body[2])
+    acc_kernels.apply(schedule[3])
+    acc_trans.apply(schedule)
+    acc_update.apply(schedule)
+    gen_code = str(psy.gen).lower()
+    assert ("  !$acc update if_present host(jn,l_ptr)\n"
+            "  !$acc enter data"
+            ) in gen_code
+    assert ("  !$acc update if_present host(zftv)\n"
```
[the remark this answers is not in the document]
nmnobre: Quite.
I currently have no machinery in place to determine if I'm using `zftv` as a temporary operand either, so we'd need to rule that's not the case as well.

src/psyclone/nemo.py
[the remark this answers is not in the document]
nmnobre: Done, `src/psyclone/domain/nemo/nodes` is a better place in my opinion though. Should I move them in this PR?

src/psyclone/psyir/nodes/acc_directives.py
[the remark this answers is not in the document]
nmnobre: Done.

nmnobre:
Ready for another look @arporter !

src/psyclone/tests/domain/nemo/transformations/acc_update_test.py
```diff
@@ -0,0 +1,463 @@
...
+            "  zftv(:,:,:) = 0.0d0\n"
+            "  !$acc end kernels\n"
+            "  !$acc update if_present host(zftv)\n"
+            "  if (l_ptr) then\n" in gen_code)
+    assert ("    call dia_ptr_hst(jn, 'ldf', zftv(:,:,:))\n"
+            "    !$acc update if_present host(zftv)\n"
+            "    zftv(:,:,:) = 1.0d0\n"
+            "    !$acc update if_present device(zftv)\n" in gen_code)
+    assert ("  !$acc update if_present host(jn,zftv)\n"
+            "  call" in gen_code)
```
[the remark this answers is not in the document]
arporter: On closer examination, I see this code fragment uses a routine from another module so we can't test compilation here anyway.

src/psyclone/psyir/transformations/acc_update_trans.py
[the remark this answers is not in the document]
arporter: Oh yes, thanks.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -254,7 +225,10 @@ def lower_to_language_level(self):
         # TODO GOcean grid properties are duplicated in this set under
         # different names (the OpenACC deep copy support should spot this).
         for pdir in self._acc_dirs:
-            self._sig_set.update(pdir.kernel_references)
+            try:
+                self._sig_set.update(pdir.kernel_references)
+            except TypeError:
+                self._sig_set.update(*pdir.kernel_references)
```
arporter, line 231: Please add a comment here explaining why we have this try-catch.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -889,6 +863,17 @@ def begin_string(self):
 
 
 def _sig_set_to_string(sig_set):
+    '''
+    Converts the provided set of signatures in a lexically sorted
```
arporter, line 867: s/in a/into a/ perhaps?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -269,40 +260,22 @@ def _add_update_directives(self, node_list):
...
-                    sync.update(host_sig.intersection(kern_sig))
-                    # If there is a statement (e.g. a call) in the current
-                    # schedule that may launch device kernels touching data
-                    # which is also touched in this host region, we
-                    # conservatively insert update directives in this schedule
-                    # for all the variables involved regardless of direction.
-                    if any(stmt.walk(self._brk_nodes) for stmt in dep_stmts):
-                        sync.update(host_sig)
+                # The inability to place variables in update directives in
+                # outer schedules is determined by their involvement in textual
```
arporter, line 269: I think this is looking better now - thank you :-) Please could you clarify/re-word what is meant by "textual dependence" (both here and in earlier and later comments). I've googled for it and am non the wiser.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -227,20 +223,15 @@ def _add_update_directives(self, node_list):
                     inputs.add(Signature(symbol_name))
                     outputs.add(Signature(symbol_name))
 
-        # Copy any data that is read by this region to the host (resp. device)
-        # if it is on the device (resp. host).
+        # Copy any data that is accessed by this host region to the host
+        # (resp. device) if it is on the device (resp. host).
+        global IN, OUT
```
arporter, line 228: Could you rename these - perhaps `HOST_INPUT` and `HOST_OUTPUT`?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -314,54 +287,93 @@ def _add_update_directives(self, node_list):
...
                 # Move up the code tree to the next deepest ancestor schedule.
-                while True:
+                child, sched = child.parent, sched.parent
+                while not isinstance(sched, Schedule):
                     child, sched = child.parent, sched.parent
-                    if isinstance(sched, Schedule):
-                        break
 
-    def _place_update(self, sched, update_pos, host_sig, direction):
+    def _sync_sig(self, dep_stmts, host_sig, acss_type):
```
arporter, line 328: Please could you make this name more readable, maybe `_get_signatures_to_sync`?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -314,54 +287,93 @@ def _add_update_directives(self, node_list):
...
+        requires synchronisation because of accesses in the compute regions
+        within the statements in dep_stmts.
+
+        :param dep_stmts: list of statements which may include compute regions.
+        :type dep_stmts: List[:py:class:`psyclone.psyir.nodes.Statement`]
+        :param host_sig: access signature(s) that need to be synchronised \
+                         with the accelerator device.
+        :type host_sig: Set[:py:class:`psyclone.core.Signature`]
+        :param int acss_type: the access type from the point of view of the
+                              host, either IN or OUT.
```
arporter, line 340: So perhaps `HOST_INPUT` and `HOST_OUTPUT`?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -314,54 +287,93 @@ def _add_update_directives(self, node_list):
...
+
+        '''
+        # If there is a statement (e.g. a call) among the dependent statements
+        # that may launch device kernels, we conservatively assume a dependency
+        # for all variables in the host region regardless of access type.
+        if any(stmt.walk(self._brk_nodes) for stmt in dep_stmts):
```
arporter, line 346: I think `self._brk_nodes` could also do with a more readable name. Perhaps `_synchronisation_node_types` or something?

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -314,54 +287,93 @@ def _add_update_directives(self, node_list):
...
+            return host_sig.copy()
+
+        kern_sig = set()
```
arporter, line 349: Please add a comment saying what this will collect, e.g. "Signatures that must be synchronised for compute kernels" or something.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -373,19 +385,26 @@ def _place_update(self, sched, update_pos, host_sig, direction):
...
+                update_pos = 0
+        elif acss_type == OUT:
+            direction = "device"
+            if update_pos is None:
+                update_pos = len(sched.children)
 
         # Check preceding and succeeding schedule positions for existing update
-        # directives. If at the beginning or at the end of the schedule, we may
-        # go out of bounds, thus the try statement.
+        # directives, coalescencing with existing directives if possible.
```
arporter, line 402: "coalescing"

arporter (commented):
Thanks Nuno, I think the refactoring has helped a lot.
Just a few remaining requests - mostly to do with naming things (which is always hard).
Once those are done this will be ready for @sergisiso to take a look.
Tests with compilation are fine.
Ref. guide builds fine.

Please could you fix/disable (as appropriate) the new pylint warnings in acc_update_trans:

    psyir/transformations/acc_update_trans.py:382:12: W0621: Redefining name 'ACCUpdateDirective' from outer scope (line 45) (redefined-outer-name)
    psyir/transformations/acc_update_trans.py:45:0: W0611: Unused ACCUpdateDirective imported from psyclone.psyir.nodes (unused-import)

src/psyclone/psyir/nodes/acc_directives.py
nmnobre: I was kinda hoping you'd tell me not to do this and instead change the return type of `kernel_references` to either a 1-tuple or a 2-tuple of sets rather than the current set or 2-tuple of sets... I didn't do that at the first instance because it always looks a bit weird to return a 1-tuple but...

src/psyclone/psyir/nodes/acc_directives.py
nmnobre: Yes, definitely. It's one of those where your brain reads what it wants, not what's there. :)

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
arporter: If you were already thinking it would be better done a different way then I change my mind and request that you do that :-)

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
nmnobre: Done. 😝

nmnobre:
> Please could you fix/disable (as appropriate) the new pylint warnings in acc_update_trans:
> 
> ```
> psyir/transformations/acc_update_trans.py:382:12: W0621: Redefining name 'ACCUpdateDirective' from outer scope (line 45) (redefined-outer-name)
> psyir/transformations/acc_update_trans.py:45:0: W0611: Unused ACCUpdateDirective imported from psyclone.psyir.nodes (unused-import)
> ```

Done.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Such a loooong name though, it'll force me to break lines.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Oh! That's crucial to understanding what's happening, thank you for letting me know, does it make more sense now?

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done. Kinda.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -98,15 +98,16 @@ def validate_global_constraints(self):
...
+        children of this directive. If a 2-tuple, the first entry is the set of
+        input references and the second entry is the set of output references.
+        The set(s) describe the quantities that must be available on the remote
         device (probably a GPU) before the parallel region can be begun.
 
-        :returns: set or 2-tuple of input and output sets of variable names
-        :rtype: Set[str] or Tuple[Set[:py:class:`psyclone.core.Signature`],
-                                  Set[:py:class:`psyclone.core.Signature`]]
+        :returns: 1-tuple or 2-tuple of input and output sets of variable names
+        :rtype: Tuple[Set[:py:class:`psyclone.core.Signature`]]
```
arporter, line 110: `...], Optional[Set[:py:class:`...`]]]`

src/psyclone/psyir/nodes/acc_directives.py
arporter: Thanks - I think you were right; it's nicer now.

src/psyclone/psyir/transformations/acc_update_trans.py
arporter: It's not as good as my name but it's good enough ;-)

src/psyclone/psyir/transformations/acc_update_trans.py
arporter: Yes, thanks, it does. It's a relief that it means what I was assuming it meant. "within no loop" sounds a bit strange though, do you mean "not within a loop"?

src/psyclone/psyir/transformations/acc_update_trans.py
arporter: Failing that, please extend the comment along the lines of what you've put in L343-344.

arporter (commented):
Nice. I'm happy now. Just a couple of comments to improve and then please request a review from Sergi.
(Tests and examples are all fine.
Ref. guide builds fine.)
I've taken the liberty of bring the branch up to master and also fixing a reference to #310 that I found in one of the tests. Therefore you'll need to pull before you make any changes.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done.

src/psyclone/psyir/transformations/acc_update_trans.py
nmnobre: Done.

src/psyclone/psyir/nodes/acc_directives.py
nmnobre: I thought a bit about it, what's the rule? Of course, if this were a 20-tuple or an unknown-dim-tuple of `str` say, you would rather type `Tuple[str]`.  In this case, it is either a one- or a two-tuple, but it feels odd to say it's an option...

nmnobre:
> Nice. I'm happy now. Just a couple of comments to improve and then please request a review from Sergi.
> (Tests and examples are all fine.
> Ref. guide builds fine.)
> I've taken the liberty of bring the branch up to master and also fixing a reference to #310 that I found in one of the tests. Therefore you'll need to pull before you make any changes.

Ready for yet another look. :) Could you please also make sure all resolved conversations are marked as such?

src/psyclone/nemo.py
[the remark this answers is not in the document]
arporter: No, let's leave them be.

arporter (commented):
One pylint issue to fix and an outstanding discussion on a docstring to be resolved.
After that, it's really going to Sergi :-)
I *think* I've resolved all other conversations but it's possible I may have missed some.
Please also merge with the latest master.

src/psyclone/psyir/transformations/acc_update_trans.py
```diff
@@ -72,261 +94,337 @@ def validate(self, node, options=None):
...
                     outputs.add(Signature(symbol_name))
 
-        # Copy any data that is read by this region to the host (resp. device)
-        # if it is on the device (resp. host).
+        # Copy any data that is accessed by this host region to the host
+        # (resp. device) if it is on the device (resp. host).
+        # We distinguish between two access modes, IN (read) and OUT (write).
+        # In this function, these two parameters are used to determine the
+        # access mode of data references on the host, as per the value of mode.
+        global IN, OUT
```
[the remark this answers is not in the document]
nmnobre: Is there a way to do that and avoid `self.IN` or similar? I brought them to module scope instead, I don't particularly like it, but would agree it is probably better than the current implementation where the definition is hidden in a method...

src/psyclone/psyir/nodes/acc_directives.py
nmnobre: Done, but used `Union` instead to the same effect.
And, for completeness, I was wrong. An tuple of strings of arbitrary length is denoted `Tuple[str, ...]`.
I was confused because, for instance, lists or sets of strings of arbitrary length are typed `List[str]` or `Set[str]` instead. I can't say I fully understand that PEP, I find it hard to read with many examples of usage before definition.

src/psyclone/psyir/transformations/acc_update_trans.py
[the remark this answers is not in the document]
arporter: Normally we would have an instance of an Enum class but I don't think it's worth it in this case.

arporter:
Link check is failing for the Yaxt repo. but that's not the fault of this PR. Examples and tests are all good with compilation. This is now ready for @sergisiso to take a look.

Issue #1347, opened 2021-07-17
Declaration ordering error


rupertford:
If a new Fortran type is declared and then used to define a variable then the ordering of these declarations is not honoured when translating to PSyIR and then back to Fortran. This means that the resulting code will not compile. For example ...
``` 
        type :: field_type
          real :: data(10)
        end type
        type(field_type) :: a
```
is output as ...
```
        type(field_type) :: a
        type :: field_type
          real :: data(10)
        end type
```
which causes a compile error (at least for gfortran).

arporter:
This is a problem when processing NEMO's obs_fbm.f90. In that case the declaration in question ends up as an UnknownFortranType because it uses a KIND intrinsic to assign an initial value to the kind paremeter:

      INTEGER, PARAMETER :: fbdp = SELECTED_REAL_KIND(12, 307)

arporter:
Martin Schrieber has also reported:

The problematic code seems to be related to

    INTEGER,PARAMETER     :: rlg=8
    REAL(kind=rlg), INTENT(OUT)  :: datetosec

since the backend of PSyIR first prints all IN, then all OUT fields.

arporter:
I have the basics working, however my check for dependencies does not yet include any symbols used to define the precision of a symbol. (Note that my previous comment on `obs_fbm.f90` in NEMO is also already fixed because we now support having a CodeBlock on the RHS of a declaration. However, it wouldn't work if that CodeBlock itself contained a reference to another parameter.)

arporter:
Can now include two files in NEMO processing that we previously had to exclude (obs_fbm.f90 and par_kind.f90). Resulting code builds fine.

Issue #717, opened 2020-03-07
Where statement array syntax is limited


rupertford:
Array syntax in Fortran where statements must be for the full extent i.e. the dimension must only contain a `:`. If it is for a subset of the data in the dimension e.g. `2:4` then the PSyIR front-end raises an exception.
We also do not currently support use of 'bare' arrays, e.g. `WHERE(some_array < 0)` because we rely on array notation to identify array accesses. Once #500 is complete we will have the necessary functionality to look-up the shape of such arrays.

arporter:
While tackling reviewer comments on #1518, it was suggested that the `_array_notation_rank` method in the frontend ought really to be associated with a PSyIR node. I started to do this but it got quite big and then started to touch code that is related to this issue so I've created `717_lift_where_restrictions` (branched off `1506_where_dtyptes`) and put my changes onto that.

arporter:
Have now added `rank_of_subsection` to most nodes that need it. However, I need to check that I've not lost fixes that were added to the old _array_notation_rank() method in fparser2.py. I also need to think about whether this routine needs extending to cope with subsections that are not for the full extent of a given dimension.

arporter:
It would also be good to handle the case where a WHERE contains references to arrays without any subscripting.

sergisiso:
I believe tacking #1858 first will allow to solve this issue without having to duplicate logic from Reference2Array and ArrayAssignment2Loop, as they all need the same.

PR #1556, opened 2022-01-09
(closes #1490) psyad dot product support

### LINKED ISSUE #1490: TL to AD translation does not support the dot_product intrinsic

rupertford:


rupertford:
A feature of the current implementation is that a dot product intrinsic will be transformed whether it is active or not. This could be changed relatively easily if requested.

rupertford:

src/psyclone/psyir/nodes/operation.py
```diff
@@ -211,7 +211,7 @@ class BinaryOperation(Operation):
         # Array Query Operators
         'SIZE', 'LBOUND', 'UBOUND',
         # Matrix and Vector Operators
-        'MATMUL'
+        'MATMUL', 'DOT_PRODUCT'
         ])
     _non_elemental_ops = [Operator.SUM, Operator.MATMUL, Operator.SIZE,
```
arporter, line 216: I think DOT_PRODUCT needs to be included in the list of _non_elemental_ops as it returns a single scalar given an array input.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+        (BinaryOperation, BinaryOperation, Literal)
+
+    :raises TransformationError: if both vector1 and vector2 are not \
+        references to array symbols.
+
+    '''
+    if not ((isinstance(vector1, Reference) and vector1.symbol.is_array) or
+            (isinstance(vector2, Reference) and vector2.symbol.is_array)):
+        raise TransformationError(
+            f"DotProduct2CodeTrans._get_array_bound requires at least one of "
```
arporter, line 76: This error message is a bit strange because obviously the arguments to a dotproduct have to be arrays. Perhaps be more precise and break this check down into checking that they are both references and then that at least one of them is an ArrayReference. Also, you need to allow for e.g. `DOT_PRODUCT(grid%var(:), grid%var(:))` or at least catch this with an informative NotImplementedError.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+            f"'{writer(vector1)}' and '{writer(vector2)}'.")
+
+    # Look for explicit bounds in one of the array declarations
+    for vector in [vector1, vector2]:
+        symbol = vector.symbol
+        my_dim = symbol.shape[0]
+        if isinstance(my_dim, ArrayType.ArrayBounds):
+            lower_bound = my_dim.lower
+            upper_bound = my_dim.upper
+            step = Literal("1", INTEGER_TYPE)
```
arporter, line 87: I don't think we can assume that the step is 1. We could have: `dot_product(a(1:4:2), b(1:4:2))`

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+            (isinstance(vector2, Reference) and vector2.symbol.is_array)):
+        raise TransformationError(
+            f"DotProduct2CodeTrans._get_array_bound requires at least one of "
+            f"the dotproduct arguments to be an array but found "
+            f"'{writer(vector1)}' and '{writer(vector2)}'.")
+
+    # Look for explicit bounds in one of the array declarations
+    for vector in [vector1, vector2]:
+        symbol = vector.symbol
+        my_dim = symbol.shape[0]
```
arporter, line 83: What about `dot_product(a(1,:),b(2,:))`?

arporter: Again, you could raise a NotImplementedError for the time being given the time pressures.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+            step = Literal("1", INTEGER_TYPE)
+            return (lower_bound, upper_bound, step)
+
+    # No explicit array bound information could be found for either
+    # array so use the LBOUND and UBOUND intrinsics.
+    symbol = vector1.symbol
+    my_dim = symbol.shape[0]
+    lower_bound = BinaryOperation.create(
+        BinaryOperation.Operator.LBOUND, Reference(symbol),
+        Literal("1", INTEGER_TYPE))
```
arporter, line 96: See earlier note - we can't assume that the dimension is number 1.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+from psyclone.psyir.nodes import BinaryOperation, Assignment, Reference, \
+    Loop, Literal, ArrayReference, Range, Routine
+from psyclone.psyir.symbols import DataSymbol, INTEGER_TYPE, REAL_TYPE, \
+    ArrayType, ScalarType
+from psyclone.psyir.transformations import TransformationError
+from psyclone.psyir.transformations.intrinsics.operator2code_trans import \
+    Operator2CodeTrans
+
+
+def _get_array_bound(vector1, vector2, writer):
```
arporter, line 52: This is useful functionality that perhaps should be in `ArrayMixin` rather than here? Possibly it should be called something like `subsection_shape`? We have similar functionality when processing WHERE statements which all needs rationalising. (e.g. #717)

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+        self._operators = (BinaryOperation.Operator.DOT_PRODUCT,)
+
+    def validate(self, node, options=None):
+        '''Perform checks to ensure that it is valid to apply the
+        DotProduct2CodeTran transformation to the supplied node.
+
+        :param node: the node that is being checked.
+        :type node: :py:class:`psyclone.psyir.nodes.BinaryOperation`
+        :param options: a dictionary with options for transformations.
+        :type options: dictionary of string:values or None
```
arporter, line 170: dict of str:str

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+                        f"intrinsic with an argument containing an array "
+                        f"slice if the array slice is for the 1st dimension "
+                        f"of the array, but found {self._writer(arg)} in "
+                        f"{self._writer(node)}.")
+
+                if not arg.is_full_range(0):
+                    raise TransformationError(
+                        f"The DotProduct2CodeTrans transformation only "
+                        f"supports the transformation of a dotproduct "
+                        f"intrinsic with an argument not an array "
```
arporter, line 229: "...an argument containing an array slice if the..."

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+        '''Apply the DOT_PRODUCT intrinsic conversion transformation to the
+        specified node. This node must be a DOT_PRODUCT
+        BinaryOperation.  If the transformation is successful then an
+        assignment which includes a DOT_PRODUCT BinaryOperation node
+        is converted to equivalent inline code.
+
+        :param node: a DOT_PRODUCT Binary-Operation node.
+        :type node: :py:class:`psyclone.psyir.nodes.BinaryOperation`
+        :param options: a dictionary with options for transformations.
+        :type options: dictionary of string:values or None
```
arporter, line 258: dict of str:str

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+            "i", symbol_type=DataSymbol, datatype=INTEGER_TYPE)
+
+        # Create temporary result variable. There is an assumption
+        # here that the DOT_PRODUCT Operator returns a PSyIR real
+        # type. This might not be what is wanted (e.g. the args might
+        # PSyIR integers), or there may be errors (arguments are of
+        # different types) but this can't be checked as we don't have
+        # the appropriate methods to query nodes (see #658).
+        symbol_res_var = symbol_table.new_symbol(
+            "res_dot_product", symbol_type=DataSymbol, datatype=REAL_TYPE)
```
arporter, line 279: Can we not do `datatype=vector1.datatype` and only fall back to REAL_TYPE if we don't have that information? I ask because we will need to ensure that the precision (kind) matches.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -0,0 +1,321 @@
...
+
+        # Replace operation with the temporary result variable.
+        result_ref = Reference(symbol_res_var)
+        node.replace_with(result_ref)
+
+        # Create "vector1(i)"
+        vector1_dims = [Reference(i_loop_symbol)]
+        if len(vector2.children) > 1:
```
arporter, line 287: Should be `vector1.children` - implies a missing test?

src/psyclone/tests/psyad/main_test.py
```diff
@@ -319,6 +319,23 @@ def test_main_not_implemented_error(tmpdir, capsys):
...
+    part of the internal support for transformations.
+
+    '''
+    code = (
+        "program test\n"
+        "real :: a, b(10), c(10)\n"
+        "a = dot_product(b(:), c(:))\n"
+        "end program test\n")
+    filename = str(tmpdir.join("tl.f90"))
+    with open(filename, "w") as my_file:
```
arporter, line 334: pylint complains about a missing `encoding='utf-8'` here.

src/psyclone/tests/psyad/main_test.py
```diff
@@ -319,6 +319,23 @@ def test_main_not_implemented_error(tmpdir, capsys):
     assert "This is not currently supported" in output
 
 
+def test_main_new_var(tmpdir):
```
arporter, line 322: Copyright date if necessary please.

src/psyclone/tests/psyad/tl2ad_test.py
```diff
@@ -94,6 +94,36 @@ def test_generate_adjoint_str(caplog):
     assert test_harness is None
 
 
+def test_generate_adjoint_str_trans():
```
arporter, line 97: Copyright date if necessary please.

src/psyclone/tests/psyad/transformations/test_script.py
```diff
@@ -0,0 +1,88 @@
...
+        "    res_dot_product = res_dot_product + b(i) * c(i)\n"
+        "  enddo\n"
+        "  a = res_dot_product\n\n"
+        "end program test\n")
+    reader = FortranReader()
+    psyir = reader.psyir_from_source(code)
+    kern_trans(psyir)
+    writer = FortranWriter()
+    result = writer(psyir)
+    assert result == expected
```
arporter, line 88: Please could you add a compilation check if possible.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+
+
+# Utilities
+
+def check_validate(code, expected, fortran_reader, index=0):
+    '''Utility function that takes Fortran code that is unsupported by the
+    DotProduct2CodeTrans transformation and checks that the expected
+    exception is raised when the validate function is applied.
+
+    :param str code: the tangent linear code stored in a string.
```
arporter, line 58: s/code stored in a string/code./

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+
+
+def check_trans(code, expected, fortran_reader, fortran_writer, tmpdir,
+                index=0):
+    '''Utility function that takes fortran code and checks that the
+    expected code is produced when the DotProduct2CodeTrans
+    transformation is applied. The code is also checked to see if it
+    will compile.
+
+    :param str code: the tangent linear code stored in a string.
```
arporter, line 82: s/code stored in a string/code./

arporter: and again below.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+
+def test_bound_error(fortran_writer):
+    '''Test that the expected exception is raised if both arguments are
+    not references to arrays.
+
+    '''
+    with pytest.raises(TransformationError) as info:
+        _get_array_bound(Literal("1.0", REAL_TYPE), Literal("2.0", REAL_TYPE),
+                         fortran_writer)
+    assert ("DotProduct2CodeTrans._get_array_bound requires at least one of "
```
arporter, line 115: In fact, we should throw an error if any argument is *not* a Reference. I think this relates to the comments I made in validate() earlier.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+            "the dotproduct arguments to be an array but found '1.0' and "
+            "'2.0'." in str(info.value))
+
+
+@pytest.mark.parametrize("dim1,dim2", [("2:10", "2:10"), (":", "2:10"),
```
arporter, line 120: What about a non-unit stride?

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+        f"subroutine dot_product_test(a1,a2)\n"
+        f"real,intent(in) :: a1(:,:), a2(:,:)\n"
+        f"real :: result\n"
+        f"result = dot_product(a1,a2)\n"
+        f"end subroutine\n")
+    expected = (
+        "The DotProduct2CodeTrans transformation only supports the "
+        "transformation of a dotproduct intrinsic with an argument not "
+        "containing an array slice if the argument is a 1D array, but "
+        "found a1 in DOT_PRODUCT(a1, a2).")
```
arporter, line 230: "found 'a1' which has shape xxxx in DOT_PRODUCT..."

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+
+    '''
+    trans = DotProduct2CodeTrans()
+    with pytest.raises(TransformationError) as info:
+        trans.apply(None)
+    assert ("The supplied node argument is not a DOTPRODUCT operator, found "
+            "'NoneType'." in str(info.value))
+
+
+@pytest.mark.parametrize("dim1,dim2", [("10", "10"), (":", "10"), ("10", ":")])
```
arporter, line 331: Could you extend this so that the bounds are given by References too?

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+        "  res_dot_product = 0.0\n"
+        "  do i = LBOUND(v1, 1), UBOUND(v1, 1), 1\n"
+        "    res_dot_product = res_dot_product + v1(i) * v2(i)\n"
+        "  enddo\n"
+        "  result = res_dot_product\n\n")
+    check_trans(code, expected, fortran_reader, fortran_writer, tmpdir)
+
+
+@pytest.mark.parametrize("arg1,arg2,res1,res2",
+                         [("(:,n)", "(:,1)", "(i,n)", "(i,1)")])
```
arporter, line 428: As commented earlier, please could you extend this test to cases where the two arrays have different numbers of dimensions.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -0,0 +1,451 @@
...
+        f"end subroutine\n")
+    expected = (
+        f"  integer :: i\n"
+        f"  real :: res_dot_product\n\n"
+        f"  res_dot_product = 0.0\n"
+        f"  do i = LBOUND(v1, 1), UBOUND(v1, 1), 1\n"
+        f"    res_dot_product = res_dot_product + v1{res1} * v2{res2}\n"
+        f"  enddo\n"
+        f"  result = res_dot_product\n\n")
+    check_trans(code, expected, fortran_reader, fortran_writer, tmpdir)
```
arporter, line 451: Please could you add a test for the case where the RHS of an assignment has more than one dot_product. It should all be fine but worth checking.

arporter (changes requested):
Looks good, just a few minor things to fix. You'll see I got a bit anxious about the utility method, only to find that you did all the checks I was worried about in the `validate` method. Therefore, I think all that is required is more comments in the utility method to say that the assumptions it makes are checked in validate.
More generally, I think the utility method or something like it would be generally useful. This may be part of #717 which will involve some rationalising of functionality that is duplicated between the fparser frontend and the PSyIR itself.
Updated docs all build (and look) fine.
I've checked all modifications with pylint (see inline).

doc/psyad/user_guide/implementation.rst
```diff
@@ -364,6 +364,33 @@ active then the loop statement is considered to be active. In this case:
...
+it must be transformed to its associated adjoint form.
+
+If an unsupported intrinsic function is found then PSyAD will raise an
+exception.
+
+The only supported intrinsic at this time is ``dot_product``.
+
+If a ``dot_product`` intrinsic is found in the tangent-linear code it
+is first transformed into equivalent inline code before the code is
+transformed to its adjoint form. The PSyIR ``DotProduct2CodeTrans``
```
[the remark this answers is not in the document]
rupertford: Added.

doc/psyad/user_guide/implementation.rst
```diff
@@ -364,6 +364,33 @@ active then the loop statement is considered to be active. In this case:
...
+transformation is used by PSyAD to perform this transformation.
+
+.. note:: At the moment all ``dot_product`` instrinsics are transformed
+	  irrespective of whether they act on active variables or not.
+
+.. note:: Note, the transformed tangent-linear code can contain new
+          variables, some of which might be active. Any such active
+          variables will need to be specified as active on the PSyAD
+          command-line using the ``-a`` flag even though they do not
+          (yet) exist in the tangent linear code.
```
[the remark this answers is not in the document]
rupertford: I must admit I thought we already had one (but I was wrong). We now have issue #1595 and reference it here.

examples/psyad/eg2/Makefile
```diff
@@ -42,7 +42,7 @@ GENERATED_FILES +=
 # second, if supported, would write both it and the corresponding test
 # harness to separate files.
 transform:
```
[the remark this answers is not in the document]
rupertford: I've added some words and referenced the issue. Hopefully that is OK.

examples/psyad/eg2/tl_hydrostatic_kernel_mod.F90
```diff
@@ -215,9 +215,9 @@ subroutine tl_hydrostatic_code(nlayers,          &
       exner_e(df) = exner( map_w3(df) + k )
     end do
     do df = 1, ndf_wt
-      ! Issue #1491 Unsupported form of rhs expression (multiple
-      ! active terms multiplied by an inactive term, as PSyclone sees
-      ! this as a single term).
+      !*** Issue #1491 Unsupported form of rhs expression (multiple
```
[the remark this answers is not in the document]
rupertford: Updated.

src/psyclone/psyad/tl2ad.py
```diff
@@ -94,6 +95,13 @@ def generate_adjoint_str(tl_fortran_str, active_variables, create_test=False):
     # to the logger.
     # logger.debug(tl_psyir.view())
 
+    # Apply any required transformations to the TL PSyIR
+    kern_trans(tl_psyir)
```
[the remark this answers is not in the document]
rupertford: Originally I was thinking that this could be a PSyclone kernel script interface (similar to `trans()` for the PSy layer), that we might add in PSyclone scripts as well - hence the generic name. It would then allow us to simply change to providing a generic user script interface in PSyAD rather than the current hidden one. However, I'm not sure if this is the way we would want to go and it is not difficult to change in the future so I don't really mind. As for an appropriate name, I don't like either of yours :-). I'll go with `preprocess_trans` and see what you think.

src/psyclone/psyad/transformations/script.py
```diff
@@ -0,0 +1,60 @@
...
+'''Module containing a PSyAD kernel transformation script that applies
+any required tranformations to the tangent linear PSyIR before it is
+translated to adjoint PSyIR.
+
+'''
+from psyclone.psyir.nodes import BinaryOperation
+from psyclone.psyir.transformations import DotProduct2CodeTrans
+
+
+def kern_trans(kernel_psyir):
```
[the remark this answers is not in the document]
rupertford: Changed.

src/psyclone/psyir/nodes/operation.py
rupertford: Good spot.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: There are checks in the validation method that ensures that both arguments are either a `Reference` to an array or an `ArrayReference`. This means that `_get_array_bound` should not need to perform any checks on its arguments. I decided to remove this test altogether as it is not possible for this error to happen (due to the validate method). As suggested I've added some words to that effect.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Again the validate method ensures this as there is a constraint that array references must be for the full range i.e. `(:)`.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: The validate method ensures this as there is a constraint that array slices must be for the first dimension of an array.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: As mentioned earlier, the validate method ensures this and I've added some information to the `_get_array_bound()` function to explain this.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: In its current form it is specialised for this case (2 args and no validation). I've added this information to the docstring and referenced #717. That may or may not be enough. As for changing the name, at the moment it only works for full array dimension accesses (e.g. using (:)) so I don't think subsection shape is any better. I've left the name as is for the moment but please request it again if you disagree.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Oops.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Oops.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Oops.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Well spotted. I've added a test that would pick this up (`test_apply_extra_dims_sizes`)

src/psyclone/tests/psyad/main_test.py
rupertford: I changed all examples of this in the file

src/psyclone/tests/psyad/main_test.py
rupertford: Updated.

src/psyclone/tests/psyad/tl2ad_test.py
rupertford: Updated.

src/psyclone/tests/psyad/transformations/test_script.py
rupertford: Good point. Added.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: OK

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Done*2

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: This test has been removed as the offending exception has also been removed. This is because the exception was causing confusion and was not needed as the validate method checks for any errors.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: The input is assumed to be unit stride and this is a constraint in the validate method so non-unit stride will never be passed to this function. I've added some words to the docstring of the function. However, I admit that the constraint in the validate method is also that only full ranges are supported so the functionality here is beyond what is required (but that does not break anything).

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Better, thanks.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Extended.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: I've created an additional test `test_apply_extra_dims_sizes` that does this.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Good point. I know it works as we have two dot_producta in our eg2 example, but it should be tested. Added two tests.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Yes, good point. At the moment we can as we only allow plain arrays as arguments. I've updated this. I've also added a validate check that both arguments have the same precision and associated tests.

rupertford:

src/psyclone/psyad/tl2ad.py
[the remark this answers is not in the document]
arporter: I like it.

src/psyclone/psyad/transformations/script.py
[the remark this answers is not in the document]
arporter: Please could you rename the file too? `script` doesn't convey any information.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
arporter: Great.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
arporter: I can't see a check for the precision in `validate` - there's a check that both arguments have REAL intrinsic type but no more. Am I missing something?

src/psyclone/tests/psyad/tl2ad_test.py
```diff
@@ -37,6 +37,9 @@
 within the psyad directory.
 
 '''
+
+# pylint: disable=protected-access
```
arporter, line 41: You don't need this - it's set in tests/.pylintrc.

src/psyclone/tests/psyir/nodes/operation_test.py
```diff
@@ -35,9 +35,10 @@
 #         I. Kavcic, Met Office
 #         J. Henrichs, Bureau of Meteorology
 # -----------------------------------------------------------------------------
-
 ''' Performs pytest tests on the Operation PSyIR node and its sub-classes. '''
 
+# pylint: disable=protected-access
```
arporter, line 40: Please remove this.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -37,10 +37,12 @@
 transformation.
 
 '''
+# pylint: disable=protected-access
```
arporter, line 40: Please remove this two lines.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -328,44 +316,52 @@ def test_apply_calls_validate():
...
+        f"  enddo\n"
+        f"  result = res_dot_product\n\n")
     check_trans(code, expected, fortran_reader, fortran_writer, tmpdir)
 
 
 def test_apply_unknown_dims(tmpdir, fortran_reader, fortran_writer):
     '''Test that the DotProduct2CodeTrans apply method produces the
-    expected PSyIR when neither of the vectors have a known size.
+    expected PSyIR when neither of the vectors have a known size. Also
+    check that the correct precision is used.
```
arporter, line 353: Please add a new test that the validate() rejects a case where the two DOT_PRODUCT arguments have different precision.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
arporter: Thanks.

arporter (changes requested):
Getting there now. I think there's one test (and check) missing and I've asked that you rename the `script` file to something more meaningful.
Apart from that, you've started adding unnecessary pylint 'ignore' statements to the test files. These aren't needed because we have `tests/.pylintrc` that sets these things.
Examples are all OK with compilation.
I get one test failure:

    FAILED psyir/transformations/intrinsics/dotproduct2code_trans_test.py::test_apply_unknown_dims
(because `r_def` is not defined).
Updated documentation builds fine and new link works (thanks).

src/psyclone/tests/psyad/tl2ad_test.py
rupertford: OK, I'll remove them. The reason I've started adding them is because the `./pylintrc` defaults are only picked up if I run tests from the `test` directory, at least for my version of pylint, and I usually go to where the tests are. I guess I'll just have to remember to specify specific tests or directories from the `tests` directory in the future.

src/psyclone/tests/psyir/nodes/operation_test.py
rupertford: Removed.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Removed. I had to add too-many-arguments to the .pylintrc file. I also found a couple of other pylint errors which I fixed.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: I did not add a check for this in the validation so there is no associated test. I did start but then decided that this is really a job for the node itself, not the transformation. Similarly for mixed datatypes. My argument is that if the code is already written in this form in the PSyIR then it should already have been checked. I could add a reference to the appropriate issue in the validate method and/or test, so it is not forgotten?

rupertford:
I fixed the compilation error by skipping and referencing an existing issue which is the cause of the problem (#1347).

rupertford:
OK, that was weird. I was getting a failure in unrelated code after merging with master, but only on Python 3.10. However, in the second merge with master it is OK again.

Anyway, ready for next review now from @arporter

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
arporter: Actually, I looked at the Fortran standard and it says nothing about precision. Anyway, if you're not checking for it please can you update the docstring :-)

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
```diff
@@ -352,9 +348,12 @@ def test_apply_unknown_dims(tmpdir, fortran_reader, fortran_writer):
     check that the correct precision is used.
 
     '''
+    if Compile.TEST_COMPILE:
+        pytest.skip("issue #1347, type declaration is being written in the "
```
arporter, line 352: That's weird. I thought the backend always generated parameter declarations first. Oh well.

arporter (changes requested):
There's one outstanding comment and you haven't renamed `script.py` (and its associated test).
Once those things are done, I think this is good to go.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Not yet!

src/psyclone/psyad/transformations/script.py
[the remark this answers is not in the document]
rupertford: Sorry I missed this before. Updated to `preprocess.py`

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: As discussed later I think this should be deferred to the PSyIR and we should assume it is correct in this transformation.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
rupertford: Hopefully this is done.

src/psyclone/tests/psyir/transformations/intrinsics/dotproduct2code_trans_test.py
rupertford: Updated the docstrings in the transformation and transformation test.

rupertford:
Hopefully I've not missed anything this time. Ready for next review from @arporter subject to CI.

arporter (approved):
All docs build OK.

PR #1592, opened 2022-01-26
(Closes #1585) First implementation of Clause class

### LINKED ISSUE #1585: [PSyIR] [Suggestion] Clause class
Looking again at #1551 and the future validation I need to do, and Andy's comments about #310 I think it might be helpful to have a clause class. Particularly for the future validation, being able to easily compare clauses would be important, as the alternative is simply to have to look through generated strings which sounds like a bad plan.

For straightforward clauses (e.g. `nowait` etc.) these would be relatively simple. For other clauses, they might need children, as the private_clause example below may show.

```
@six.add_metaclass(abc.ABCMeta)
def clause(Node):
  _children_valid_format = None
  _clause_text = ""
  def __init__(self, ast=None, children=None, parent=None, annotations=None):
    pass

  #function to convert to string - couldn't find the real name.
  def to_code_str(self)
    children_strings = []
    for child in children:
        children_strings.append(child.to_code_str()) #Use correct function name
   child_strings = ""
    if children_strings != []:
        child_strings = "(" + ", ".join(children_strings) + ")"
    return self._clause_text + child_strings

# Maybe also equality/compare functions which can be overriden.


def nowait_clause(clause):
  _children_valid_format = None
  _clause_text = "nowait"
   def __init__(self, children=None):
      super().__init__(children=children)

def private_clause(clause):
  _children_valid_format = "[Reference+]" (not sure of this syntax)
  _clause_text = "private"
  def __init__(self, children=None):
    # Children can't be None, not sure if I need to validate that here or if super does it.
    super().__init__(children=children)
```

I think we'd need another type (similar) to `schedule` for lists of clause's (e.g. `clause_list` maybe) to contain clauses for those classes who should have clauses (e.g. OpenMP nodes, OpenACC nodes etc.), which would be added as an extra child to those directives.

Does this sounds like a reasonable solution to this problem? @arporter @rupertford @sergisiso 

EdiT: I can probably start having a look at this this week if it seems reasonable.

LonelyCat124:
I have implemented the clause class as discussed.

I expect coverage not to be sorted yet, but I think all the current tests still succeed.

The `gen_code` support is minimal, I don't think its planned to support `gen_code` properly for clauses, instead it just uses the old `gen_code` functionality in the directives. I think the visitor does work correctly for other code generation methods though (a visitor exists).


LonelyCat124:
Coverage should now be done, need to have lunch and then look at pycodestyle/pylint and think about documentation.

LonelyCat124:
I think this is ready for review now. I'm not sure if I need to do more to get the documentation correct so let me know if so. I added the `clauses` member to where i thought the appropriate docs should be, and the rest I think is just autogenerated but I'm not sure if I'm missing something somewhere.

LonelyCat124:
I guess one question for the reviewer that I have is whether they're ok with this style of description for the valid children:
```
    _children_valid_format = ("Schedule, [GrainsizeClause | NumTasksClause],"
                              " [NogroupClause]")
```
In principle, this descriptor means there is an ordering:
1. `Schedule`
2. `GrainsizeClause | NumTasksClause | NogroupClause`
3. `NogroupClause` iff 2 is not a `NogroupClause`

OpenMP actually doesn't have any ordering restrictions on these clauses order, so it would be just as valid to swap the order of 2 and 3, however I'm not sure if/how I would describe this in a nice way, it'd end up being a bit gross like:
` Schedule, [ [ [NogroupClause], [GrainsizeClause | NumTasksClause] ] | [ [GrainsizeClause | NumTasksClause], [NogroupClause] ] ]`
I don't think its worth the additional complexity of this `_children_valid_format` just to unrestrict the order of the clauses, but am happy to make that change if you think it would be worthwhile.

src/psyclone/psyir/nodes/clause.py
```diff
@@ -0,0 +1,61 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author A. B. G. Chalk, STFC Daresbury Lab
+# -----------------------------------------------------------------------------
+
+''' This module contains the Clause abstract node implementation '''
+
+import abc
+from psyclone.psyir.nodes.node import Node
+import six
```
sergisiso, line 41: I see many Clause, Directive and its subclasses still use six a lot for marking the abstract base classes. Now that we don't support python2 anymore maybe we can delete "six" and inherit directly from ABC https://docs.python.org/3/library/abc.html , can you check is this is relatively easy to do.

(Unfortunately Node can't be properly ABC yet I think because it is instantiated for some dummy nodes in tests)

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -54,6 +54,11 @@
 from psyclone.psyir.nodes.directive import StandaloneDirective, \
     RegionDirective
 from psyclone.psyir.nodes.loop import Loop
+from psyclone.psyir.nodes.literal import Literal
+from psyclone.psyir.nodes.omp_clauses import GrainsizeClause, NowaitClause,\
+    NogroupClause, NumTasksClause
+from psyclone.psyir.nodes.schedule import Schedule
+from psyclone.psyir.symbols import INTEGER_TYPE
```
sergisiso, line 61: Again there is a lot of "six" use in this file than may be possible to delete now, but I am not entirely sure.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -283,8 +313,6 @@ def begin_string(self):
         '''
         result = "omp single"
 
-        if self._nowait:
-            result = result + " nowait"
```
sergisiso, line 287: Won't removing that disable nowait in the LFRic API that still uses gen_code?

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -599,6 +631,52 @@ def __init__(self, children=None, parent=None, grainsize=None,
...
+
+    @property
+    def clauses(self):
+        '''
+        :returns: the Clauses associated with this directive.
+        :rtype: List of :py:class:`psyclone.psyir.nodes.Clause`
+        '''
+        if len(self.children) > 1:
+            return self.children[1:]
+        return []
```
sergisiso, line 679: I think this should be in the super class RegionDirective. And a similar one in StandaloneDirective with just `return self.children`

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -652,15 +740,13 @@ def gen_code(self, parent):
...
-
         # Generate the string containing the required clauses
         extra_clauses = ", ".join(clause_list)
 
         parent.add(DirectiveGen(parent, "omp", "begin", "taskloop",
                                 extra_clauses))
 
-        for child in self.children:
-            child.gen_code(parent)
+        self.children[0].gen_code(parent)
```
sergisiso, line 749: User dir_body instead of children[0]

sergisiso (changes requested):
@LonelyCat124 In general looks good. I added a few comments to clean up aspects of the code. I haven't reviewed the test code yet.

Regarding
> OpenMP actually doesn't have any ordering restrictions on these clauses order, so it would be just as valid to swap the order of 2 and 3, however I'm not sure if/how I would describe this in a nice way, it'd end up being a bit gross like:
Schedule, [ [ [NogroupClause], [GrainsizeClause | NumTasksClause] ] | [ [GrainsizeClause | NumTasksClause], [NogroupClause] ] ]
I don't think its worth the additional complexity of this _children_valid_format just to unrestrict the order of the clauses, but am happy to make that change if you think it would be worthwhile.

The problem is that these are sibling constraints and the _valid_children format and method are more suited to parent-children constraints.
It could be enough to make it : `Schedule, [GrainsizeClause | NumTasksClause | NogroupClause]*`
and then add the no-repeat and the mutually excusive grainsize-numtasks in the _validate_global_constraints? It breaks a bit the fail-early design principle but I see the constructor and the transformation already check for the mutually exclusive constraint so it is already not easy to make it unless one explicitly constructs the tree. What do you think?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1439,6 +1439,32 @@ def nemokern_node(self, node):
...
+        string will be prefixed by a single space.
+
+        :param node: a Clause PSyIR node.
+        :type node: :py:class:`psyclone.psyir.nodes.Clause`
+
+        :returns: the Fortran code for this node.
+        :rtype: str
+
+        '''
+        result_list = [" "]
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1439,6 +1439,32 @@ def nemokern_node(self, node):
             result += self._visit(child)
         return result
 
+    def clause_node(self, node):
+        '''This method is called when a Clause instance is found in the
+        PSyIR tree. It returns the clause and its children as a string.  The
+        string will be prefixed by a single space.
```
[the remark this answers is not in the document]
LonelyCat124: Done I think. I dgot confused, I think the comma isn't needed.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1451,7 +1477,14 @@ def regiondirective_node(self, node):
         :rtype: str
 
         '''
-        result_list = ["{0}!${1}\n".format(self._nindent, node.begin_string())]
+        result_list = ["{0}!${1}".format(self._nindent, node.begin_string())]
+
+        clause_list = []
+        for clause in node.clauses:
+            clause_list.append(self._visit(clause))
+        result_list.append(",".join(clause_list))
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/clause.py
```diff
@@ -0,0 +1,61 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2020-2021, Science and Technology Facilities Council.
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/directive.py
```diff
@@ -107,14 +115,14 @@ def dir_body(self):
         :returns: the Schedule associated with this directive.
         :rtype: :py:class:`psyclone.psyir.nodes.Schedule`
 
-        :raises InternalError: if this node does not have a single Schedule as\
-                               its child.
+        :raises InternalError: if this node does not have a Schedule as\
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/directive.py
```diff
@@ -107,14 +115,14 @@ def dir_body(self):
...
+                               its first child.
         '''
-        if len(self.children) != 1 or not \
-           isinstance(self.children[0], Schedule):
+        if len(self.children) < 1 or not isinstance(self.children[0],
+                                                    Schedule):
             raise InternalError(
-                "Directive malformed or incomplete. It should have a single "
-                "Schedule as a child but found: {0}".format(
+                "Directive malformed or incomplete. It should have a "
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/directive.py
```diff
@@ -60,6 +60,14 @@ class Directive(Statement):
     _PREFIX = ""
     _colour = "green"
 
+    @property
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/node.py
```diff
@@ -394,6 +394,14 @@ def _validate_child(position, child):
...
         return False
 
+    @property
+    def clauses(self):
+        '''
+        :returns: the Clauses associated with this directive.
+        :rtype: List of :py:class:`psyclone.psyir.nodes.Clause`
+        '''
+        return []
+
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/omp_clauses.py
```diff
@@ -0,0 +1,145 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2021, Science and Technology Facilities Council.
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/omp_clauses.py
```diff
@@ -0,0 +1,145 @@
...
+# -----------------------------------------------------------------------------
+
+''' This module contains the implementations of the various OpenMP Directive
+nodes.'''
+
+from psyclone.psyir.nodes.clause import Clause
+from psyclone.psyir.nodes.literal import Literal
+
+
+class NowaitClause(Clause):
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/omp_clauses.py
```diff
@@ -0,0 +1,145 @@
...
+    '''
+    OpenMP nowait clause.
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/omp_clauses.py
```diff
@@ -0,0 +1,145 @@
...
+        :param int position: the position to be validated.
+        :param child: a child to be validated.
+        :type child: :py:class:`psyclone.psyir.nodes.Node`
+
+        :return: whether the given child and position are valid for this node.
+        :rtype: bool
+
+        '''
+        return False
+
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/nodes/clause.py
LonelyCat124: I added this for a few classes. We could make Node ABC and just have a DummyNode in the tests but I'm not sure if that works properly or not.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: I think the `gen_code` call still does this, it shoudln't need to be in begin_string right?

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: StandaloneDirective should return [] I think right? They have no clauses yet I think.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: Done

LonelyCat124:
I think I've addressed the comments in the review.

I think I'd rather keep the ordering (and mutual exclusion) over `[ListofClauses]*`, at least when there are few options. In the future it might be that the available type/number/exclusions of clauses is complex enough that the `_valid_children_format` might not be able to handle it, in which case just `[clauses]*` is sufficient, but where they are simple, and a slightly ordering can cover the mutual exclusion I think thats better for me.

doc/user_guide/psyir.rst
```diff
@@ -184,6 +184,7 @@ To solve this issue some Nodes also provide methods for semantic navigation:
    .. automethod:: psyclone.psyir.nodes.ArrayReference.indices()
 - ``RegionDirective``:
    .. automethod:: psyclone.psyir.nodes.RegionDirective.dir_body()
+   .. automethod:: psyclone.psyir.nodes.RegionDirective.clause()
```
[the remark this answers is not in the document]
LonelyCat124: I think sorted, let me know what you think.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -107,13 +111,32 @@ def _get_reductions_list(self, reduction_type):
...
+    @property
+    def clauses(self):
+        '''
+        :returns: the Clauses associated with this directive.
+        :rtype: List of :py:class:`psyclone.psyir.nodes.Clause`
+        '''
+        if len(self.children) > 1:
+            return self.children[1:]
+        return []
+
```
sergisiso, line 123: I resolved the previous comment because I couldn't reply but I still think this should be in RegionDirective. The behaviour will be the same in OpenACC even if we don't have examples of it yet.

Same for the OMPStandaloneDirective.clauses I think it could go to the generic StandaloneDirective

src/psyclone/psyir/nodes/omp_directives.py
sergisiso: Ah I see. Can you add the "# pylint: disable=no-self-use" to get rid of the pylint issue. I am right it thinking that in the long term (when all clauses have their class) this could be a static method or simply a string attribute of the class.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -587,9 +633,13 @@ class OMPTaskloopDirective(OMPRegionDirective):
                              a grainsize and num_tasks value \
                              specified.
     '''
+    _children_valid_format = ("Schedule, [OMPGrainsizeClause | "
```
sergisiso, line 636: > I think I'd rather keep the ordering (and mutual exclusion)

That's fine by me but please at a comment like: "This specification respects the mutual exclusion of OMPGrainsizeClause and OMPNumTasksClause but adds an ordering requirement. Other specifications to soften the ordering requirement are possible but need additional checks in the global constraints"

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -599,6 +649,43 @@ def __init__(self, children=None, parent=None, grainsize=None,
...
+        :rtype: bool
+
+        '''
+        if position == 0:
+            return isinstance(child, Schedule)
+        if position == 1:
+            return isinstance(child, (OMPGrainsizeClause, OMPNumTasksClause,
+                                      OMPNogroupClause))
+        if position == 2:
+            return (isinstance(child, OMPNogroupClause))
```
sergisiso, line 687: Pylint complaints about the unneeded outer parenthesis

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -2274,3 +2275,43 @@ def test_fw_comments(fortran_writer):
...
+    schedule = container.children[0]
+    loops = schedule.walk(Loop)
+    loop = loops[0].detach()
+    directive = OMPTaskloopDirective(children=[loop], num_tasks=32)
+    master = OMPMasterDirective(children=[directive])
+    parallel = OMPParallelDirective(children=[master])
+    schedule.addchild(parallel, 0)
+    assert '''!$omp parallel default(shared), private(i)
+  !$omp master
+  !$omp taskloop num_tasks(32)
```
sergisiso, line 2305: Please add nogroup clause here as well so we check that the clauses join is done with whitespaces.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -49,7 +49,9 @@
 from psyclone.psyir.nodes import OMPDoDirective, OMPParallelDirective, \
     OMPParallelDoDirective, OMPMasterDirective, OMPTaskloopDirective, \
     OMPTaskwaitDirective, OMPTargetDirective, OMPLoopDirective, Schedule, \
-    Return, OMPSingleDirective, Loop, Literal, Routine, Assignment, Reference
+    Return, OMPSingleDirective, Loop, Literal, Routine, Assignment,\
+    Reference, OMPNowaitClause, OMPGrainsizeClause, OMPNumTasksClause,\
+    OMPNogroupClause
```
sergisiso, line 54: Can you test the node.clauses property in some of this directives (we are currently missing the coverage of the clauses property with a StandaloneDirective which is just a check that it returns an empty string)

sergisiso (changes requested):
@LonelyCat124 This is almost ready now, just a few more comments regarding clean up but nothing major. I added a comment inline about this but make sure to hit 100% diff coverage in CodeCov.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1437,7 +1471,11 @@ def standalonedirective_node(self, node):
         :rtype: str
 
         '''
-        result_list = [f"{self._nindent}!${node.begin_string()}\n"]
+        result_list = [f"{self._nindent}!${node.begin_string()}"]
+
+        clause_list = []
+        result_list.append(",".join(clause_list))
+        result_list.append("\n")
```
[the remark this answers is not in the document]
LonelyCat124: Using the same code as for regiondirective is sensible, but untestable with coverage for now. I've added the code in commented out with a comment describing what neesd to change in the future.

I did just switch both to concatenation now instead.

src/psyclone/psyir/nodes/directive.py
```diff
@@ -43,14 +43,12 @@
 
 from __future__ import absolute_import
```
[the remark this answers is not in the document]
LonelyCat124: done

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: Moved these up to the parents then

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: Done

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: Fixed

src/psyclone/tests/psyir/backend/fortran_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
LonelyCat124: Added to OMPTaskloopDirective tests.

LonelyCat124:
Fixed up those comments now I think, should be ready to review if tests/cov pass now.

LonelyCat124:
`StandaloneDirective.clauses` is still not tested.

LonelyCat124:
Finally ready for review again

sergisiso (approved):
@LonelyCat124 All comments addressed. I still removed the body from the abstract class since I think by definition they should not have a body, and then removed the new test for it. Also fixed a couple of last typos.

All test pass with compilation, the documentation has been updated and build (after the typo correction), no TODOs pending for the current issue. This is ready to be merge.

LonelyCat124:
> @LonelyCat124 All comments addressed. I still removed the body from the abstract class since I think by definition they should not have a body, and then removed the new test for it. Also fixed a couple of last typos.

Ah thats just my fault, I didn't realise I could remove the body entirely from the function and not even need a `pass` so I was struggling with coverage.

Issue #1542, opened 2021-12-08
LFRic intergrid kernels: problem with continuous fields


tommbendall:
Alex Brown has noticed that there is a problem with some LFRic intergrid kernels that map continuous (W2) fields between fine and coarse meshes.

This seemed to happen when the model's halo depth is smaller than the number of fine cells per coarse cell. As the field is continuous, the model attempts to do a halo exchanges but we think they are not being updated to the correct depth.

For now our work-around (on a branch) is to change the meta-data to say that the mapped field is discontinuous. This prevents any halo exchanges calls from being inserted by PSyclone (as these kernels we don't actually need to do any halo exchanges!)

The offending kernels are:
[https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/gungho/source/kernel/prolong_w2_kernel_mod.F90](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/gungho/source/kernel/prolong_w2_kernel_mod.F90)
[https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/gungho/source/kernel/restrict_w2_kernel_mod.F90](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/gungho/source/kernel/restrict_w2_kernel_mod.F90)
[https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/gungho/source/kernel/weights_prolong_w2_kernel_mod.F90](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/gungho/source/kernel/weights_prolong_w2_kernel_mod.F90)

atb1995:
Two proposed solutions were:

1. Psyclone could allow GH_WRITE for continuous fields, writing to the shared DoFs twice. Essentially treating the continuous fields as discontinuous fields
2. Psyclone could prevent intergrid kernels from looping into the halo
region

arporter:
The problem is that the current PSyclone support for intergrid kernels was motivated by the multi-grid solver. As such it assumes that each grid in a hierarchy increases (or decreases) in resolution by a factor of 2 in both x and y directions. This therefore applies to the halo depths too. It seems that the requirements of the multi-grid solver do not quite line up with what is being done with using the intergrid support in order to run physics at different resolutions.

This needs to be explored further.

tommbendall:
[...]

PR #1594, opened 2022-01-28
[LFRic] script to remove halo exchanges (closes #1593)

### LINKED ISSUE #1593: [LFRic] script to remove halo exchanges
There has been discussion about PSyclone adding halo exchanges where they are not required. The reason for this is that there are some cases where a field may be continuous but within a kernel it only accesses the local dofs and annexed dofs of other arrays and therefore does not need a halo exchanges for these arrays - at least that is my understanding. In the long term this should be captured in the metadata somehow but in the short term a simple option would be for a PSyclone script to remove halo exchanges where they are known to not be required.

rupertford:


rupertford:
Ready for first review from @arporter assuming CI checks are successful

examples/lfric/scripts/rm_hex.py
```diff
@@ -0,0 +1,175 @@
...
+
+'''
+from psyclone.psyGen import Kern, HaloExchange
+from psyclone.psyir.transformations import TransformationError
+
+# Add all selected invoke/kernel, position/arg names here. The kernel
+# information has a name and a position, with the position being the
+# nth time this kernel is found in a schedule, as a kernel may be used
+# multiple times in a schedule.
+invoke_info = {
```
arporter, line 52: Presumably the idea is that there will be a copy of this script for each algorithm that needs modifying. It would therefore be better if the functionality was separate to the specification of the halo exchanges to delete. However, this is only an example so it can stay as is.

arporter (changes requested):
A very conservative (small 'c') implementation that is suitably restrictive. However, if this is going to go on to master then it should be exercised in one of the examples, ideally including compilation checking, and the respective documentation updated.

That said, given that in #1542 we have agreed to permit a kernel to specify a 'WRITE' access for a field on a continuous function space, we could just park this ticket.

rupertford:
Given that we have an agreed solution that uses the kernel metadata I don't see a need for this script. It was only to demonstrate that it is possible if needed and it is now not needed. I suggest that I withdraw this PR and issue #1593 and explain why in a comment. Do you agree @arporter?

arporter:
Yes, I agree that this can be closed.

rupertford:
As agreed, this script will not go on master as another better solution has been agreed.

Issue #1562, opened 2022-01-10
matmul 2 code transformation raises an exception if it is not the only thing on the rhs of an expression


rupertford:
j_ls_vorticity = wqp_h(qp1) * wqp_v(qp2) *             &
                        matmul( jac, ls_vorticity_at_quad )

rupertford:
Created branch [1562_matmul_rhs](https://github.com/stfc/PSyclone/tree/1562_matmul_rhs)

rupertford:
There are a couple of potential solutions here.

1) Determine the rank of the array returned by the matmul - probably by using `.datatype` as has been implemented for references. Then we can allocate a new array which we assign to using matmul. This requires us to know the size of the returned array as well as the dimension and could require allocation, which is not yet supported in PSyIR (so we would have to resort to fparser2).
2) If the rank and dimensions of the LHS and the matmul are the same then allocate matmul to the LHS and then increment it by the other values. For example: `a = b * matmul(c,d)` goes to `a = matmul(c,d); a = a*b`. I need to determine when this can be assumed. I think it is when everything is arrays with the same rank.

rupertford:
It may be that we should tackle implicit arrays being mapped to colon notation first in this case. e.g. `a` goes to `a(:)`

Issue #1561, opened 2022-01-10
matmul 2 code transformation raises an error if an argument is a function


rupertford:
jac = matmul(jac_inv,transpose(jac_inv))

PR #1607, opened 2022-02-04
add matmul 2 code transformation to psyad (closes 1559)

rupertford:


rupertford:
This is ready for first review by @arporter subject to CI

doc/psyad/user_guide/implementation.rst
```diff
@@ -374,17 +374,20 @@ it must be transformed to its associated adjoint form.
...
-
-.. note:: At the moment all ``dot_product`` instrinsics are transformed
-	  irrespective of whether they act on active variables or not.
+The only supported intrinsics at this time are ``dot_product`` and
+``matmul``.
+
+If a ``dot_product`` or ``matmul`` intrinsic is found in the
+tangent-linear code it is first transformed into equivalent inline
+code before the code is transformed to its adjoint form. The PSyIR
+``DotProduct2CodeTrans`` or ``Matmul2CodeTrans`` transformation
```
arporter, line 383: "...transformations are used by PSyAD to perform these transformations."

src/psyclone/psyad/transformations/preprocess.py
```diff
@@ -54,8 +55,14 @@ def preprocess_trans(kernel_psyir):
...
     '''
     dot_product_trans = DotProduct2CodeTrans()
+    matmul_trans = Matmul2CodeTrans()
 
     for oper in kernel_psyir.walk(BinaryOperation):
         if oper.operator == BinaryOperation.Operator.DOT_PRODUCT:
             # Apply DOT_PRODUCT transformation
             dot_product_trans.apply(oper)
+
+    for oper in kernel_psyir.walk(BinaryOperation):
```
arporter, line 65: You could do away with this separate loop (and thus its associated `walk`) and just have the `if` as a second child of the loop at line 60.

arporter (changes requested):
Good job. Just a few minor things to fix/tidy up.
Documentation has been updated and builds OK.
Associated test file gives 100% coverage of modified source file.

doc/psyad/user_guide/implementation.rst
rupertford: Better, thanks.

src/psyclone/psyad/transformations/preprocess.py
rupertford: As discussed privately I was concerned about the transformations affecting each other. However this is not the case, so I have done as you suggest. When we add support nested intrinsics (see #1561 and #1562 for example) we might need to revisit - and perhaps even have an outer `while intrinsics` loop. However, the tests associated with implementing solutions to those issues should pick up any problems.

arporter (approved):
Examples and tests OK with compilation.
All docs build OK modulo one minor error in kernel_tools which I've fixed.

PR #1619, opened 2022-02-11
[PSyAD] (Closes #1496) Add LFRic-specific support for test-harness generation

### LINKED ISSUE #1496: [PSyAD] Test harness needs to initialise dof map arrays correctly
Currently the PSyAD test harness initialises every kernel argument using pseudo-random numbers. However, LFRic kernels that iterate over cells are passed a dofmap array. This will need initialising with correct index data rather than random numbers.

arporter:


arporter:
I've brought this branch up-to-date with master. I now need to somehow merge it with what I have in 1555_lfrig_alg_gen.

arporter:
I've done the merge on a different branch (`1555_lfric_psyad_test_harness`) in case it all went pear shaped. It seems to be going well though. I'm in the process of changing the inoke-creation to use the new Functor classes etc.

arporter:
psyad/eg2 now builds, runs and passes. This is for the tl_hydrostatic_kernel_mod. I need to test with other kernels.

rupertford:
That's exciting!

arporter:
tl_rhs_eos_kernel_mod.F90 can't be compiled because it uses coordinate_jacobian module which is not in our cut-down infrastructure.
strong_curl_kernel_type compiles, runs and validates :-)

arporter:
I've updated the 'status' document in Teams but as of now I can validate three kernels (strong_curl, w2_to_w1_project and the tl_hydrostatic). Of the others, three have operator arguments which I haven't implemented test harness support for yet and two required extending the PSyclone LFRic test infrastructure. Having done that, the test harness builds and runs but produces NaNs for both kernels. It could be that the addtitional infrastructure requires additional setup that is not happening?

arporter:
I've now hacked my way through to supporting operators. This wasn't straightforward as we have no Builtins to initialise or manipulate operators. For now I've therefore extended PSyclone to permit operators as arguments to Builtins. Further, to avoid having to implement a new looping structure, I've allowed Builtins to operate on the 'domain' so that I can just use Fortran array notation when initialising/copying the contents of an operator. I've not included the operator arguments in the inner products that are calculated but this is currently OK as none of the operator arguments are active (in the current set of kernels). With all of that done, all of the kernels barring those requiring new infrastructure now pass the adjoint test :-)

arporter:
Closing this in favour of #1782,

Issue #1640, opened 2022-03-03
Release 2.2


arporter:
One of the deliverables for the ExCALIBUR project is a new release of PSyclone.
It would be great to get #1637 and #1554 on for this but it's not essential.

Probably we should also get the one outstanding PR on fparser onto master and make a release of that too as that is important in improving our ability to handle NEMOVAR.

Are there any other considerations @rupertford, @sergisiso and @TeranIvy ?

LonelyCat124:
Not sure which excalibur project this is in reference to ( I assume yours) - we also have to do a release each year for the deliverables for the Task-based one as well. I doubt any more tasking stuff will make it into this PSyclone release, however can we add release notes specifying the new major things in this release? And as part of that include something like:

- Added support for OpenMP's taskloop directive, and code to automatically compute data dependencies between taskloops and generate appropriate taskwaits as a separate transformation.

arporter:
@LonelyCat124, yes sorry, I meant the Marine Systems one. We don't normally do official release notes but perhaps we should (we do write an email summarising the significant changes and that's the same thing). We'll just have to decide where to put the notes.

sergisiso:
I see PSyclone github project doesn't show a section with the releases in the right side bar. In projects that have it there is a good place to put release notes. They can be created from https://github.com/stfc/PSyclone/releases

arporter:
That's a good point Sergi - I'd noticed that in the past but had not got any further in investigating it.

TeranIvy:
We definitely need mixed precision support on. That includes #1349 and also `r_tran` support, #1632.

`GH_WRITE` access for fields on continuous FS (#1608) is also highly desirable.

arporter:
This has got to be done by the 18th of March @TeranIvy so we can try to get those things on but if not it's no big deal - we can do another release once they are on.

arporter:
I used GitHub's rather nice release-management mechanism to create the release and have updated the 'making a release' instructions accordingly. You can see the new release at https://github.com/stfc/PSyclone/releases.
I've not emailed the LFRic list about this one as it is primarily aimed at the Marine Systems work.

PR #1657, opened 2022-03-16
Towards #1640 release 2.2

arporter:


arporter:
A few small changes ready for release 2.2. Ready for review from @rupertford.

rupertford (approved):
All looks good. After discussion with @arporter I've reverted one small change. I've updated the changelog and documentation but have not added the new release line to the changelog as we have agreed that this should be done on a release.

PR #1659, opened 2022-03-16
Fix kernel stub generator links in the LFRic tutorials and ignore binder links checks

nmnobre:


arporter:
@nmnobre you're making the changes in the wrong direction in this PR! psyclone-kern *is* the replacement for gen-kernel-stub, not the other way around. The problem is the target that the links are checked against. There's already an Issue for this I think.

PR #1668, opened 2022-03-22
(Closes #450) Ensure set_dirty/clean calls are outside ACC regions.

### LINKED ISSUE #450: [LFRic] set_dirty() and set_clean() calls in OpenACC kernels and parallel directives
At the moment any "set_dirty" or "set_clean" calls are added immediately after a loop. Currently these appear within acc kernels or acc parallel regions. In OpenMP these calls are protected by the "single" directive. However, in OpenACC it might be better to avoid calling these functions on a gpu and instead call them on the cpu. This still needs to be clarified (I will check with NVIDIA) but, if so, the set_dirty and set_clean placement would need to be re-written.

One option would be to add these calls just before a halo exchange. Another would be to add them as part of the PSyIR and them move them using transformations. The PSyIR option might seem the simplest but it would require similar dynamic logic to halo exchanges as the content and existance of such calls change as transformations are applied.

arporter:


arporter:
Turns out I'd not branched from master. Doh! Will create a new PR with the right commits.

PR #1669, opened 2022-03-22
(Closes #450) remove set_dirty/clean from ACC regions.

### LINKED ISSUE #450: [LFRic] set_dirty() and set_clean() calls in OpenACC kernels and parallel directives

arporter:
Fixes #450 #1694


arporter:
Code wise I think this is pretty much done now. However, I can't test with a compilable example because our version of the LFRic infrastructure does not compile with nvfortran (21.7 or 22.2). I'm waiting to hear back from @christophermaynard to see how he's doing his OpenACC experiments.

arporter:
On chatting with Rupert, I've realised I need to make sure that what I've done works for OpenMP offload.
I also need to update the relevant part of the LFRic tutorial.

arporter:
I could add a `gen_post_region_code` method to the OMPTargetDirective node. However, I'm wondering whether it might be better to move this functionality up to Schedule and have it find all HaloExchange nodes and insert all necessary set_dirty/clean calls before each one. Finding all `Loop`s between two `HaloExchange` nodes might be difficult though.

arporter:
I looked into doing this inside `DynInvokeSchedule.gen_code()` and it got complicated because it's not that easy to find the nodes before a halo exchange and, having done so, it's then not easy to insert f2pygen nodes in the correct place (because of course we don't use the PSyIR backend in LFRic yet). Therefore, I'm going to extend my original solution so that it works for OpenMP too.

arporter:
I started trying to generalise what I have so far and realised that, because it's all `gen_code` based, it won't work for NEMO and GOcean as they use the PSyIR backend. (Not that either of those APIs have this problem because neither have set-dirty/clean calls.)

arporter:
Implementation wise, this is now done. @hiker is kindly fixing the OpenACC example so once that's done I can proceed to check that this PR works with the example and then go for review.

arporter:
The fix for #1694 is quite simple so I've included that in this PR.
Currently, the code that we generate for lfric/eg14 still doesn't compile as we're not adding `!$ACC ROUTINE` to the user-supplied kernel code.

arporter:
Getting closer. Applying the ACCRoutineTransformation appears to work and the resulting code all still builds (without OpenACC). Pausing this while I go back to #1608.

arporter:
Makefile is currently not working correctly for the `compile` target. However, without DM I can now generate, compile and run on GPU on Isambard :-)

arporter:
We have a bug which means that the ACCRoutineTrans is not quite working properly (the `USE` statement in the PSy layer uses the name of the un-transformed module). If I fix that manually then we can run:
lfric_eg14_profile
The two builtin kernels are run in parallel while the coded kernel is currently run in serial (hence the big 'gap').

arporter:
Note that my attempt to use managed memory resulted in an illegal-address error when executing the coded kernel. However, since the plan is to use enter-data, I switched to that and everything seems to work OK.

examples/lfric/README.md
```diff
@@ -278,34 +278,19 @@ psyclone -s ./kernel_constants.py \
...
-   because calls to set_clean()/dirty() end up within parallel regions - TODO
-   #450.
-   ```sh
-   cd eg14/
-   psyclone -s ./acc_parallel_dm.py main.x90
-   ```
+The supplied Makefile defines a `compile` target that will build the
+transformed code. Currently the compilation will fail because the
+generated PSy-layer code does not contain the correct name for the
+transformed kernel module (issue #1724).
```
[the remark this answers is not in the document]
hiker: transformation.rst also still states about the routine transformation:

> This transformation is currently not supported for kernels in
the Dynamo0.3 API.

examples/lfric/eg14/acc_parallel_dm.py
```diff
@@ -64,4 +66,7 @@ def trans(psy):
             parallel_trans.apply(loop.parent.parent)
         enter_data_trans.apply(schedule)
 
+        for kernel in schedule.walk(CodedKern):
```
hiker, line 69: Add a comment here that makes it explicit that the kernel is modified (or a new kernel file is created).

src/psyclone/dynamo0p3.py
```diff
@@ -7655,9 +7653,13 @@ def gen_code(self, parent):
             return
 
         # Set halo clean/dirty for all fields that are modified
-        fields = self.unique_modified_args("gh_field")
+        if not self.unique_modified_args("gh_field"):
+            return
 
-        if not fields:
+        if self.ancestor((ACCRegionDirective, OMPRegionDirective)):
```
hiker, line 7659: Instead of hard-coding the classes here, would it make sense to add a method to the directive base class `can_include_class` (not sure about the name tbh, but you get the idea). So instead of searching for hard-coded classes, you would have to find a Directive as ancestor, and check if that disables calls inside of the routine.
Just a suggestion, I will still approve the PR without this.

src/psyclone/psyir/nodes/directive.py
```diff
@@ -133,6 +136,38 @@ def clauses(self):
...
+                    commented = True
+                    parent.add(CommentGen(parent, ""))
+                    parent.add(CommentGen(parent,
+                                          " Set halos dirty/clean for fields "
+                                          "modified in the above loop(s)"))
+                    parent.add(CommentGen(parent, ""))
+                loop.gen_mark_halos_clean_dirty(parent)
+
+        if commented:
+            parent.add(CommentGen(parent, ""))
```
hiker, line 169: Could you also add a comment stating that this is the end of the set dirty/clean section? (that might be useful for debugging if we ever should get halos created from different places?)

src/psyclone/tests/domain/lfric/transformations/dynamo0p3_transformations_test.py
```diff
@@ -6404,11 +6353,10 @@ def test_accenterdata_builtin(tmpdir):
 
     assert LFRicBuild(tmpdir).code_compiles(psy)
 
-    assert ("!$acc enter data copyin(0.0_r_def,f1_proxy,f1_proxy%data,"
-            "f2_proxy,f2_proxy%data,ginger,m1_proxy,m1_proxy%data,m2_proxy,"
-            "m2_proxy%data,map_aspc1_f1,map_w1,map_w2,map_w3,ndf_aspc1_f1,"
-            "ndf_w1,ndf_w2,ndf_w3,nlayers,undf_aspc1_f1,undf_w1,undf_w2,"
-            "undf_w3)" in output)
+    assert ("!$acc enter data copyin(f1_proxy,f1_proxy%data,"
```
hiker, line 6356: Great!

src/psyclone/tests/domain/lfric/transformations/dynamo0p3_transformations_test.py
```diff
@@ -6475,6 +6452,49 @@ def test_accparalleltrans():
         "      END DO\n"
         "      !$acc end parallel\n") in code
 
+    assert LFRicBuild(tmpdir).code_compiles(psy)
```
hiker, line 6455: Nice!

src/psyclone/tests/domain/lfric/transformations/dynamo0p3_transformations_test.py
```diff
@@ -7441,5 +7457,4 @@ def fake_validate(_1, _2, options=None):
                 else:
                     trans.validate(loop)
             assert "validate test exception" in str(err.value), \
-                "{0}.validate() does not call LoopTrans.validate()".format(
-                    name)
+                f"{name}.validate() does not call LoopTrans.validate()"
```
hiker, line 7460: Thanks a lot for all the code cleanup (f-strings, imports not in toplevel etc)!!

hiker (changes requested):
Great job. I have to admit that I don't like that we now create the halos from different places, but after scratching my head for some time, I agree with you that's the best solution. I made one suggestion to remove at least the list of directives from dynamop3.py file.

There is also what appears to be a minor conflict with trunk that needs to be fixed.
Main issue is to update the user guide, and maybe some makefile improvements.

examples/lfric/README.md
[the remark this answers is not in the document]
arporter: Oh yes, well spotted. I've add a mention and link to the doc for eg14 and have updated that text in transformation.rst to say that it almost works :-)

examples/lfric/eg14/Makefile
```diff
@@ -36,20 +36,21 @@
...
 # export F90=nvfortran
-# export F90FLAGS="-acc"
+# export F90FLAGS="-acc=gpu -Minfo=all -gpu=managed"
 
 PSYROOT=../../..
 
 include $(PSYROOT)/examples/common.mk
 
-GENERATED_FILES = *.o *.mod  $(EXEC) main_alg.f90 main_psy.f90
+GENERATED_FILES = *.o *.mod $(EXEC) main_alg.f90 main_psy.f90 testkern_w0_kernel_[0-9]_mod.f90
```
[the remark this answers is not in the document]
arporter: Wow you ran it a lot! Fixed and added.

examples/lfric/eg14/Makefile
```diff
@@ -59,33 +60,21 @@ LFRIC_LIB=$(LFRIC_DIR)/lib$(LFRIC_NAME).a
...
 	${PSYCLONE} -s ./acc_parallel_dm.py \
 	-opsy $*_psy.f90 -oalg $*_alg.f90 $<
 
-#TODO #1669 - the code currently does not compile
-#             set_dirty calls inside openacc region
-#TODO #1694 - the code currently does not compile
-#             incorrect variable names and constants
-#             when using builtin
-compile: transform_parallel_dm $(EXEC)
+testkern_w0_kernel_0_mod.f90: main_psy.f90
```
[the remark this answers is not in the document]
arporter: Yes, I've gone for (hopefully consistent) hardcoding throughout now.

examples/lfric/eg14/Makefile
```diff
@@ -97,8 +86,8 @@ $(LFRIC_LIB):
 	$(MAKE) -C $(LFRIC_DIR)
 
 # Dependencies
-main_psy.o:	testkern_w0_kernel_mod.o
-main_alg.o:	main_psy.o
+main_psy.o:	testkern_w0_kernel_0_mod.o
+main_alg.o:	main_psy.o testkern_w0_kernel_0_mod.o
```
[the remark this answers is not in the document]
arporter: Yep.

examples/lfric/eg14/acc_parallel_dm.py
arporter: Oh yes. In fact the various doc strings were outdated too.

src/psyclone/psyir/nodes/directive.py
arporter: Oh yes, good idea.

src/psyclone/dynamo0p3.py
arporter: Thinking about it, Transformations already have an `excluded_node_types` property and ideally we don't want to duplicate information. Information on forbidden node types could also be used when performing the global validation checks in the backend. Could we generate a Transformations `excluded_node_types` from similar properties from one or more Node types that we know the Transformation inserts?

arporter: I've opened #1755 to discuss this further.

examples/lfric/README.md
```diff
@@ -278,34 +278,19 @@ psyclone -s ./kernel_constants.py \
 ## Example 14: OpenACC
 
 This example shows how OpenACC directives can be added to the LFRic
-PSy-layer. This is work in progress so the resultant code is not
-expected to run correctly but it gives a starting point for
-evaluation.
+PSy-layer. It adds OpenACC enter data, parallel and loop directives in the
+presence of halo exchanges. It also transforms the (one) user-supplied
+kernel with the addition of an `ACC routine` directive.
```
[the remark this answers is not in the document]
hiker: Great!

src/psyclone/dynamo0p3.py
hiker: Am happy to postpone this.

hiker:
All issues have been addressed. In discussion with @arporter I have disabled the compilation of `examples/lfric/eg14`, which currently fails due to #1724 (to avoid our nightly compilation tests failing unnecessarily). I will merge this in as soon as the CI finishes.

hiker (approved):
All issues addressed, UG builds fine, and compilation tests finish without an error now.

PR #1672, opened 2022-03-24
Fix BOZ literal standard incompatibility

nmnobre:


arporter (approved):
Thanks for fixing this.
The copyright date needs updating but as that's the only outstanding thing I'll do it.

PR #1677, opened 2022-03-30
Ignore MyBinder HTTP503 in mlc GitHub Action

nmnobre:


arporter (approved):
I'm happy with the change and as it's a 1-line configuration change I'll merge it.
I note that we're now getting errors from virtualbox. I've tried accessing the site in a browser and it is very slow (and returned an error in Firefox). Not sure what we can do about that.

nmnobre:
It was actually very fortunate that VirtualBox's website was (still is?) down as the verbose output enabled me to confirm the behaviour for MyBinder is as expected. 😝

The solution is trivial: we wait 🙂

PR #1678, opened 2022-03-31
1667 add openacc compilation

hiker:
Should be ready to go. I noticed that the README in examples/lfric explicitly specifies the PSyclone command to run, but the old (and the new) Makefile contains separate makefile targets to trigger this command. Not sure if you want that changed or not.


hiker:
As far as I can see the failure was a problem reaching virtualbox (from tutorials/practicals/README.md) - it seems to be working now.

One point worth thinking about: I did not add any openacc flag to be used by default in the openacc eg14 example, which means any problems will not be noticed. On the plus side this means our nightly compilation test will keep on working. Not sure if you prefer to add some defaults (e.g. detect if compiler is gfortran, and then add -fopenacc etc)??

Otherwise ready to go.

examples/lfric/eg14/Makefile
```diff
@@ -31,34 +31,86 @@
...
 # ------------------------------------------------------------------------------
-# Author: A. R. Porter, STFC Daresbury Laboratory
-# Modified J. Henrichs, Bureau of Meteorology
+# Author: J. Henrichs, Bureau of Meteorology
+# Modifications: A. R. Porter, STFC Daresbury Laboratory
 
-include ../../common.mk
+# The compiler to use may be specified via the F90 environment
+#
+# export F90=gfortran
```
arporter, line 39: Probably worth changing this to specify how to build for OpenACC using nvfortran.

examples/lfric/eg14/Makefile
```diff
@@ -31,34 +31,86 @@
...
-transform: kernels parallel_no_dm parallel_dm
+transform_parallel:
+	${PSYCLONE} -nodm -s ./acc_parallel.py \
+	-opsy main_psy.f90 -oalg main_alg.f90 main.x90
 
-kernels:
-	${PSYCLONE} -s ./acc_kernels.py -nodm \
-  ../code/gw_mixed_schur_preconditioner_alg_mod.x90 \
-  -oalg /dev/null
+transform_parallel_dm:
```
arporter, line 73: Please put a TODO for #1669 here as the generated code is currently no correct.

examples/lfric/eg14/Makefile
```diff
@@ -31,34 +31,86 @@
...
-	@echo "No run targets for lfric/eg14"
+	./$(EXEC)
+
+$(EXEC): $(LFRIC_LIB) $(OBJ)
+	$(F90) $(F90FLAGS) $(OBJ) -o $(EXEC) -L$(LFRIC_DIR) -l$(LFRIC_NAME)
+
+$(LFRIC_LIB):
+	$(MAKE) -C $(LFRIC_DIR)
+
+$(EXTRACT_LIB): $(LFRIC_LIB)
```
arporter, line 93: I don't think we need this?

examples/lfric/eg14/README.md
```diff
@@ -0,0 +1,35 @@
+# Stand-alone LFRic Kernel OpenACC Example
+
+This directory contains a runnable example of kernel that uses
```
arporter, line 3: s/kernel/an LFRic mini-app/

examples/lfric/eg14/README.md
```diff
@@ -0,0 +1,35 @@
...
+you want to use:
+```shell
+export F90=gfortran
+export F90FLAGS="-Wall -g -fopenacc"
+make
+```
+
+## Running
+
+The binary can be executed using ``example_openacc`` without additional parameters:
```
arporter, line 30: Probably worth adding a sentence to say that if using NVIDIA hardware you can specify NV_ACC_NOTIFY=3 when running in order to see GPU-related activity.

examples/lfric/eg14/acc_kernels.py
```diff
@@ -54,6 +54,6 @@ def trans(psy):
         print("Transforming invoke '"+invoke.name+"'...")
         schedule = invoke.schedule
         kernels_trans.apply(schedule)
-        schedule.view()
+        # schedule.view()
```
arporter, line 57: Please rm this line.

examples/lfric/eg14/main.x90
```diff
@@ -0,0 +1,99 @@
+! BSD 3-Clause License
+!
+! Modifications copyright (c) 2020-2021, Science and Technology Facilities Council
```
arporter, line 3: Please update the copyright date range.

examples/lfric/eg14/acc_kernels.py
```diff
@@ -54,6 +54,6 @@ def trans(psy):
         print("Transforming invoke '"+invoke.name+"'...")
```
arporter, line 54: Please update the copyright date for this file.

arporter (changes requested):
Thanks very much for doing this Joerg. I've tried building it with nvfortran and unfortunately the LFRic infrastructure fails to compile:

    nvfortran -acc -I../../../src/psyclone/tests/test_files/dynamo0p3/infrastructure -c reference_element_mod.F90
    NVFORTRAN-S-0155-Cannot override PUBLIC type bound procedure with PRIVATE type bound procedure  (reference_element_mod.F90: 332)
    NVFORTRAN-W-0155-No overriding procedure specified for DEFERRED type bound procedure populate_outward_normals_to_vertical_faces (reference_element_mod.F90: 332)
    NVFORTRAN-S-0155-Cannot override PUBLIC type bound procedure with PRIVATE type bound procedure  (reference_element_mod.F90: 332)

@christophermaynard said "Oh yeah. This is the "private is too private" bug. A type cannot see its internal data which is clearly dumb. Set the data attributes of the type to public for now. I need to make a proper work around." Would you mind seeing whether this is a quick fix?

Apart from that, there are just a few minor things and the branch needs bringing up-to-date with master.

hiker:
> Not having OpenACC compilation tested automatically is, as you say, not good but not easy to fix either. Would it be possible to install e.g. the NVIDIA HPC SDK in your AWS instance? Could we then update the Makefile to use nvfortran by default and fall back to gfortran (without OpenACC) if unavailable?

We are not yet on AWS (hopefully soon), but on the NCI (research supercomputer) the nvidia SDK is luckily installed. But there are two problems:
1. The infrastructure itself does not compile. Do I assume correctly that you want me to remove all the private declarations in our copy of the infrastructure files so that it compiles?
2. We cannot simply switch compilers - the infrastructure is compiled once, and then used in **all** compilations we are doing. We would either need to support different compiled version of the infrastructure library, or add a third compile test (gnu, Intel, and nvfortran). The latter would be much simpler (just adding another pass to the compile test). Any preference?

examples/lfric/eg14/Makefile
hiker: Done. You are aware that there are also big problems with builtins?
```
!$acc& copyin(nlayers,field1_proxy,field1_proxy%data,0.0_r_def,ndf_aspc1_field1,undf_aspc1_field1,map_aspc1_field1,field2_proxy, &
!$acc& field2_proxy%data,1.0_r_def,ndf_aspc1_field2,undf_aspc1_field2,map_aspc1_field2)
      !
      !$acc parallel default(present)
      !$acc loop independent
      DO df=loop0_start,loop0_stop
        field1_proxy%data(df) = 0.0_r_def
      END DO
```
nlayers, (u)ndf... etc are all undefined in case of a builtin.

examples/lfric/eg14/README.md
hiker: Done.

examples/lfric/eg14/acc_kernels.py
hiker: Done

examples/lfric/eg14/Makefile
hiker: Done.

examples/lfric/eg14/Makefile
hiker: Indeed.

examples/lfric/eg14/README.md
hiker: Done.

examples/lfric/eg14/acc_kernels.py
hiker: Done.

examples/lfric/eg14/main.x90
hiker: Done

examples/lfric/eg14/testkern_w0_kernel_mod.f90
```diff
@@ -0,0 +1,82 @@
+! BSD 3-Clause License
+!
+! Copyright (c) 2017-2021, Science and Technology Facilities Council
```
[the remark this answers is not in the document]
hiker: Done.

hiker:
It was actually only one infrastructure file that failed, so I've fixed this file by commenting out the private statements (with a comment on why this was done). Not sure if we should add a TODO - it's not really a PSyclone thing :(

Compiling with two compilers (e.g. check if nvfortran is available, and if so switch to it for all openacc tests) might have an additional problem: on some module-based systems different compilers cannot be loaded at the same time (e.g. on Cray if you load the intel compiler, you can't also load the cray compiler). To avoid this, I think using a third compilation test would be much much easier (since the module load commands are site-specific, and I would really prefer not to have site-specific module commands anywhere).

I am also not sure about the openacc builtin issues - do you want a separate ticket, or are you aware of this already (I didn't find one when I looked).

hiker:
The link check fails because the sympy documentation is not found. I tried to fix this, but realised that it is a known issue (DMCA complain):
https://groups.google.com/g/sympy/c/yuJXymiJqws

For now, I would assume that this will get fixed soonish - at least that the link will be re-enabled.

Ready for next review.

examples/lfric/eg14/Makefile
arporter: Oh dear, no, I wasn't aware of that. If you could open an Issue that would be great.

arporter (changes requested):
Almost there now, thanks for fixing the infrastructure lib. I think there's just a new issue needed plus an associated TODO.
I think my preference would be to add a third compilation test with nvfortran with OpenACC enabled. There are OpenACC examples in both NEMO and GOcean so it should be possible to exercise those too when that's done.

examples/lfric/eg14/Makefile
hiker: #1694 
I've added a todo to the Makefile, as an explanation for why compilation fails. Couldn't think of a better place :)

hiker:
I've also tried to fix the missing link to yaxt. 

We will update our nightly builds to include a complete third run of everything using `nvfortran`, but atm we have to fix our inode quota issue first before we can do anything :)

Ready for next review.

arporter (approved):
All requested changes have been made and a new issue created (thanks).
Reference guide still builds fine.

Issue #1689, opened 2022-04-12
[LFRic] Incorrect stencil depth list for discontinuous stencil readers followed by discontinuous writers


TeranIvy:
When we have a kernel that has discontinuous stencil readers is followed by a kernel that has discontinuous writers in a single invoke, PSyclone generates incorrect code for stencil depth list string. E.g. as reported from [LFRic/branches/dev/ianboutle/r34895_bl_exp_split@35124](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/branches/dev/ianboutle/r34895_bl_exp_split?rev=35124) ([LFRic ticket 3205](https://code.metoffice.gov.uk/trac/lfric/ticket/3205)):
```
algorithm/bl_exp_alg_mod_psy.f90(493): error #6642: There are null arguments in an intrinsic function reference.   [MAX]
      IF (u_in_w3_proxy%is_dirty(depth=max(stencil_depth,))) THEN
---------------------------------------^
algorithm/bl_exp_alg_mod_psy.f90(494): error #6642: There are null arguments in an intrinsic function reference.   [MAX]
        CALL u_in_w3_proxy%halo_exchange(depth=max(stencil_depth,))
-----------------------------------------------^
algorithm/bl_exp_alg_mod_psy.f90(497): error #6642: There are null arguments in an intrinsic function reference.   [MAX]
      IF (v_in_w3_proxy%is_dirty(depth=max(stencil_depth,))) THEN
---------------------------------------^
algorithm/bl_exp_alg_mod_psy.f90(498): error #6642: There are null arguments in an intrinsic function reference.   [MAX]
        CALL v_in_w3_proxy%halo_exchange(depth=max(stencil_depth,))
-----------------------------------------------^
algorithm/bl_exp_alg_mod_psy.f90(501): error #6642: There are null arguments in an intrinsic function reference.   [MAX]
      IF (tile_fraction_proxy%is_dirty(depth=max(stencil_depth,))) THEN
---------------------------------------------^
algorithm/bl_exp_alg_mod_psy.f90(502): error #6642: There are null arguments in an intrinsic function reference.   [MAX]
        CALL tile_fraction_proxy%halo_exchange(depth=max(stencil_depth,))
```

[...]

PR #1690, opened 2022-04-12
 [LFRic] Bug fix for stencil depth list (for #1689)

TeranIvy:
Create correct stencil depth list when we have a kernel that has discontinuous stencil readers followed by a kernel that has discontinuous writers in a single invoke (more details in the issue).

Also cleaned up some `pylint` issues in `dynamo0p3.py`.

I generated PSy-layer code with optimisations for the `bl_exp_alg_mod.F90` in Ian's branch  [LFRic/branches/dev/ianboutle/r34895_bl_exp_split@35124](https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/branches/dev/ianboutle/r34895_bl_exp_split?rev=35124) from this bug fix branch, placed it into `um_physics/source/psy` directory and ran the `lfric_atm` test suite with it. It is all fine. Command-line build without optimisations also works fine.

### DIFF
--- a/src/psyclone/dynamo0p3.py
+++ b/src/psyclone/dynamo0p3.py
@@ -4773,7 +4773,7 @@ class DynBasisFunctions(DynCollection):
                 for var in self._qr_vars[shape]:
                     var_names.append(
                         self._symbol_table.find_or_create_tag(var+"_proxy")
-                                          .name)
+                        .name)
                 parent.add(
                     TypeDeclGen(
                         parent,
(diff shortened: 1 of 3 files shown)


src/psyclone/dynamo0p3.py
```diff
@@ -5956,19 +5956,24 @@ def _compute_halo_depth(self):
...
         depth_info_list = self._compute_halo_read_depth_info()
 
-        # if there is only one entry in the list we can just return
-        # the depth
-        if len(depth_info_list) == 1:
-            return str(depth_info_list[0])
-        # the depth information can't be reduced to a single
-        # expression, therefore we need to determine the maximum
-        # of all expresssions
+        # Create string representation of depth info list and filter
```
arporter, line 5963: I'm afraid I don't like this fix - it's fixing the symptom rather than the underlying disease. I've done a bit of digging and the problem seems to be that one of the HaloReadAccess objects returned by `_compute_halo_read_info` has a literal depth of zero and no other depth information (i.e. var_depth is None and `max_depth`, `max_depth_m1` and `annexed_only` are all `False`). I suspect this should never happen. If I were you @TeranIvy I'd pass this on to @rupertford now that you've created a reproducer :-)

arporter:
I've set status to 'reviewed with actions' as, although I haven't done a proper review, I think the fix needs to go deeper. You're welcome to tackle it @TeranIvy but I think it's really @rupertford 's bug :-)

TeranIvy:
> I've set status to 'reviewed with actions' as, although I haven't done a proper review, I think the fix needs to go deeper. You're welcome to tackle it @TeranIvy but I think it's really @rupertford 's bug :-)

I don't mind either. @rupertford, would you like to take this on?

rupertford:
> I don't mind either. @rupertford, would you like to take this on?

It probably makes sense for me to take a look seeing as I wrote the original logic.

TeranIvy:
Closing this as it is superseded by #1703.

Issue #1688, opened 2022-04-08
[Frontend] PR #1670 breaks the ability to process NEMO code


nmnobre:
Merged PR #1670, which simplifies the symbol tables of the nodes in the PsyIR, breaks the ability to process NEMO code. I suspect (but @sergisiso can probably correct me here) that this is due to existing bugs/inconsistencies that are simply resurfacing rather than the introduction of new bugs.

- [ ] Routines, such as Fortran functions, with symbol `RoutineSymbol<DeferredType>` lose tag `own_routine_symbol`. It appears that these routines initially get the tag associated with symbol `RoutineSymbol<NoType>` but later lose it when this is converted to a specific return type, e.g. `DataSymbol<Scalar<INTEGER, UNDEFINED>, Local>`. NEMO file `domvvl.f90` contains function `dom_vvl_alloc()` which triggers the behaviour when the PSyIR is generated.

- [ ] The letter case of the symbols added to the tables is inconsistent complicating string comparisons. I suggest storing them in their original case for symbols from case-sensitive languages and in lowercase for case-insensitive languages such as Fortran.

- [ ] Fortran USE statements broken into (at least) one continuation line using an ampersand are wrongly parsed/interpreted leading to errors due to unrecognised entities which PSyclone wrongly says are never imported. This happens when processing both `julian.f90` and `obs_conv.f90`. These files/modules can be skipped with no performance penalty since, even before merging PR#1670, no optimisation was being successfully applied.

- [ ] For some reason, when processing `obs_read_surf.f90`, PSyclone tries to swap a symbol in a nonexistent table. This file/module can also be skipped with no performance penalty since it is marked for profiling only.

sergisiso:
Can you check if #1692 fixes things. It did for me.

PR #1691, opened 2022-04-14
[Frontend] Lowercase routine names + tag return symbol (for #1688)

nmnobre:


rupertford:
Hi @nmnobre. Is there a particular reason why you plan to lower case the routine names?

nmnobre:
Hi @rupertford,

1) I'd say my major concern is that, in case-insensitive languages such as Fortran, the same entity can be referred to by a myriad of symbols: all the possible lower and uppercase combinations of the letters in its name. If the case encloses no information, it can (and perhaps should, see below) be dropped. If we ever build a frontend for case-sensitive languages, then it's my opinion the symbols should remain untouched.

2) As briefly explained in #1688, without said change, the return value of Fortran functions is being sometimes saved in a different case to the routine name, causing conflicts when transforming to NEMO-specific PSyIR constructs.

I'm open to suggestions 😊

rupertford:
Thanks for the explanation @nmnobre. The reason I asked is that my understanding is that the PSyIR is set up to be case insensitive and it is therefore the responsibility of the back-ends to deal with this if they need to.

nmnobre:
I see. That works as long as I'm not dealing with rogue developers in a case-sensitive language defining things like: 'int rupert;' and 'int RuPERt;' in the same compilation unit 😝 otherwise, mangling is always an option 🤓

rupertford:
Yes I think at some point we would need to mangle, unless we just switched to case-sensitive matching in the PSyIR and lower/upper-cased everything in the front-end as you originally suggested for this problem. If we did the latter, we could even keep an "original name" if we so desired, that any case insensitive back-ends could use to maintain case from input to output.

nmnobre:
Superseded by #1692.

PR #1693, opened 2022-04-20
1661 improve dep analysis

hiker:
Improve detection of dependencies for parallelisation.


hiker:
I think this should be ready for a first review. The one point I am not certain of is about error handling (i.e. sympy-writer throws an exception because of an invalid sympy expressions - which happens in case of array ranges). I caught that error in one sympy function, but am not certain where the best please is (symbolic_maths, dependency_tools, leaving it to the user script???)

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
         return True
 
     # -------------------------------------------------------------------------
-    def array_access_parallelisable(self, loop_variable, var_info):
-        '''Tries to determine if the access pattern for a variable
-        given in `var_info` allows parallelisation along the variable
-        `loop_variable`. The following messages might be provided
-        to the user using the message API:
+    @staticmethod
+    def get_flat_indices(component_indices, set_of_loop_vars):
```
sergisiso, line 241: I don't think the name captures exactly what this method does. Some ideas:

- This is a few levels down the call stack from when we deal with loops. I would call the second parameter just `variables`
- It is a static method that just operates on component_indices. Could this be moved to a method of component_indices?
- It does not completely flatten the structure and the name doesn't convey that it searches only for specific variables. Could this be named something like: `component_indices.get_variables_locations(variables)` or something similar?

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+            unique_vars = set(str(sig) for sig in index_vars.keys())
+            unique_vars = unique_vars.intersection(set_of_loop_vars)
+            indices.append(unique_vars)
+        return indices
 
-        :param loop_variable: symbol of the variable that is parallelised.
-        :type loop_variable: :py:class:`psyclone.psyir.symbol.DataSymbol`
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def partition(comp_ind1, comp_ind2, loop_vars):
```
sergisiso, line 271: loop_vars docstring is missing. And I would probably drop the  "loop_" part.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+        '''This function partitions the subscripts to the component indices
+        into disjunct sets. For example:
```
sergisiso, line 273: Maybe "into sets of minimal coupled groups" (following the book semantics)?

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+                partition_infos[first_use] = \
+                    (partition_infos[first_use][0].union(part_info[0]),
+                     partition_infos[first_use][1] + part_info[1])
+                del partition_infos[k]
+
+        return partition_infos
+
+    # -------------------------------------------------------------------------
+    @staticmethod
+    def independent_0_var(index_exp1, index_exp2):
```
sergisiso, line 340: I am not sure what "0_var" means in the name. Could this be named "are_indenpendent"?

But also, this method that is only used once and it is equivalent to:
 `SymbolicMaths.get().never_equal(index_exp1, index_exp2)`
maybe we don't need this whole method if there is the appropriate comment?

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+        :type index_written: :py:class:`psyclone.psyir.nodes.Node`
+
+        :returns: the dependency distance in loop iterations if it is a
+            independent integer value, and None otherwise.
+        :rtype: int or None.
+
+        '''
+        sym_maths = SymbolicMaths.get()
+        sympy_expressions, symbol_map = SymPyWriter.\
+            get_sympy_expressions_and_symbol_map([index_read, index_written])
```
sergisiso, line 401: A couple of Nemo files fail here with a ParseError. Can we put this inside a try/except and `return None` in case of a SymPy exception (meaning it could not guarantee independence)?

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+        # is a pair consisting of:
+        # - a set of all loop variables used in the subscript of
+        #   both accesses
+        # - list of all subscripts. Initially these lists contain
+        #   only one subscript, but they will be modified later
+        # Example: `a(i,j)` and `a(i,k)` -->
+        #          [ ({"i"}, [(0,0)]), ({"j","k"}, [(0,1)])]
+        partition_infos = []
+        for i, indx in enumerate(comp_ind1.iterate()):
+            partition_infos.append((indices_1[i].union(indices_2[i]), [indx]))
```
sergisiso, line 308: Some Nemo files fail here with a IndexOutOfRange. I can not see with the debugger which Fortran expression produces it but I can dig deeper if you need.

For the time being could you add some test parametrizations in src/psyclone/tests/psyir/tools/dependency_tools_test.py::test_partitions where "comp_ind1" and "comp_ind2" have different number of components. As there isn't any.
I would also like to see cases where the loop_vars make an empty set in some components of one component_indices but not the other one e.g. "a(n)%b(n)", "a(i)%b(n)"

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+        solutions = sym_maths.solve_equal_for(sympy_expressions[0],
+                                              sympy_expressions[1],
+                                              d_var)
+        if solutions == "independent":
+            # The solution is independent of the variable, i.e. every value
+            # of `d_var` is a solution. So there are certainly
+            # dependencies.
+            return None
+
+        if len(solutions) == 1:
```
sergisiso, line 431: Some Nemo files fail here with "TypeError: object of type 'ConditionSet' has no len()". As I said previously, if possible, it would be great to handle this complex sets inside the solve_equal_for and just return a standard python set.

src/psyclone/tests/psyir/tools/dependency_tools_test.py
```diff
@@ -202,116 +228,340 @@ def test_arrays_parallelise(parser):
...
+                                 integer i, j, k
+                                 integer, parameter :: n=10
+                                 real, dimension(n) :: a1
+                                 real, dimension(n,n) :: a2
+                                 real, dimension(n,n,n) :: a3
+                                 {expression} = 1
+                                 end program test''')
+    prog = parser(reader)
+    psy = PSyFactory("nemo", distributed_memory=False).create(prog)
+    assign = psy.invokes.get("test").schedule
```
sergisiso, line 284: Is there any reason for many of the DependencyTool tests to need the FortranStringReader and PSyFactory?

Ideally they should work with the more generic `psyir = fortran_reader.psyir_from_source(source)`

sergisiso (changes requested):
Submitting current comments, I will continue reviewing later today at home.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+                # Notice that distance 0 will only be returned if the
+                # subscript actually depends on the loop variable.
+                return True
+        # Additional tests could be added here (e.g. to see that
+        # `a(i,i)=a(i,i+1)+1` can be parallelised).
+
+        return False
+
+    # -------------------------------------------------------------------------
+    def is_array_access_parallelisable(self, loop_variables, write_access,
```
sergisiso, line 507: You can ignore this suggestion but maybe you could name this method: `is_loop_carried_dependency` since this seems to match the docstring and it will make it more differentiable from the next method (then also the returns docstring).

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+                        str_access = self._language_writer(write_access.node)
+                        self._add_message(f"The write access to "
+                                          f"'{str_access}' causes a "
+                                          f"write-write race condition.",
+                                          DACode.ERROR_WRITE_WRITE_RACE,
+                                          [str_access])
+                    else:
+                        str_write = self._language_writer(write_access.node)
+                        str_other = self._language_writer(other_access.node)
+                        self._add_message(f"The write access to '{str_write}' "
```
sergisiso, line 656: Now that I notice, these methods that do _add_message but don't have a starting _clean_messages() maybe ought to be private ( _ prefixed), so that the DependencyTool user don't call them directly because it would accumulate to previous error messages.

 In fact maybe only: 'get_all_messages', 'can_loop_be_parallelised' and the `get_*_parameters` should be the public API?

(* I am unsure about partition - discussed in another comment)

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -457,9 +752,9 @@ def can_loop_be_parallelised(self, loop, loop_variable=None,
         self._clear_messages()
 
         if not isinstance(loop, Loop):
-            raise TypeError("can_loop_be_parallelised: node must be an "
-                            "instance of class Loop but got '{0}'".
-                            format(type(loop).__name__))
+            raise TypeError(f"can_loop_be_parallelised: node must be an "
+                            f"instance of class Loop but got "
+                            f"'{type(loop).__name__}'")
         if not loop_variable:
```
sergisiso, line 758: This is not strictly touched in this PR but I had this question before:

is the `loop_variable` argument not always `loop.variable`? and the `var_accesses` argument, not always the result of `loop.reference_accesses(var_accesses)`?.  In what context do we need them to be provided as arguments?

I guess for "var_accesses" it could be a performance benefit to not traverse again the tree if we already have it, but I am still not convinced since any mutation of the tree could  produce a different var_access. And the user must know that is must not be re-used if anything has changed. So I think it would be safer to not provide this optional argument. But maybe there is some place where this is needed?

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -70,27 +71,24 @@ def validate(self, node1, node2, options=None):
...
         loop_var2 = node2.variable
 
         if loop_var1 != loop_var2:
-            raise TransformationError("Loop variables must be the same, "
-                                      "but are '{0}' and '{1}'".
-                                      format(loop_var1.name, loop_var2.name))
+            raise TransformationError(f"Loop variables must be the same, "
+                                      f"but are '{loop_var1.name}' and "
+                                      f"'{loop_var2.name}'")
         vars1 = VariablesAccessInfo(node1)
```
sergisiso, line 92: Maybe this is more generic than this transformation and could be mostly encapsulated in a new DependencyTool entry point (maybe `are_loops_fusable(loop1, loop2)` or `are_loops_independent(loop1, loop2)` ). It does feel this code belogn more on that domain and it would keep `partition` as private method. What do you think?

src/psyclone/domain/nemo/transformations/nemo_loop_fuse.py
```diff
@@ -174,35 +172,42 @@ def validate_written_array(var_info1, var_info2, loop_variable):
             inconsistent indices, e.g. a(i,j) and a(j,i).
 
         '''
-
+        # pylint: disable=too-many-locals
```
sergisiso, line 175: The implementation looks good but I am unsure where is this tested :)

The CI says its covered but `src/psyclone/tests/domain/nemo/transformations/nemo_loop_trans_test.py` doesn't seem to go thought this?

sergisiso (changes requested):
@hiker The new dependency analysis functionality looks great, it fixes the cases I had that the previous one couldn't and it opens the door to assist in the loop reductions in Nemo. (Unfortunately I could not produce yet a full NEMO run with it to provide the seedup numbers because there are a couple of unrelated issues in master - but I will try to fix them and do it on the next review)

I have asked for some methods renaming and code movements that I hope make things clearer. I am slowly going through the book but the topic is quite complex so I will rely mostly on the unit-tests to check the correctness.

In the Developer's guide, can you rename the "PSyIR Type, Symbols and Data Dependencies" section to "PSyIR Types and Symbols". And then in "Dependency Analysis Functionality in PSyclone" move the "Dependency Tools" subsection one level higher so it doesn't fall inside "Variable Accessess"

src/psyclone/tests/core/symbolic_maths_test.py
```diff
@@ -210,6 +211,97 @@ def test_symbolic_math_not_equal_structures(fortran_reader, expressions):
...
+
+
+@pytest.mark.parametrize("exp1, exp2, result", [("i", "2*i+1", set([-1])),
+                                                # Infinite solutions (i is any
+                                                # integer) are returned as
+                                                # string "independent"
+                                                ("i", "i", "independent"),
+                                                ("i*i", "2*i-1", set([1])),
+                                                ("i*i", "4", set([2, -2])),
+                                                ("2*i", "2*i+1", set()),
```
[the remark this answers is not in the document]
hiker: I've never seen that @sergisiso Could you give me the nemo example that returns a CondSet? Just the raw source code will probably be enough, don't need a breakdown to one access or so. 

And even worse, a CondSet is not compatible with a python `set`.

hiker:
OK, I think I have addressed nearly all issues, with one or two items open for discussion ;)

Only the link checker failed, so it should be ready to go!

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -41,16 +41,91 @@
     dependency analysis.'''
 
 from __future__ import absolute_import, print_function
+from enum import IntEnum
+
+from sympy import Symbol, Integer
```
sergisiso, line 46: Can you use "import sympy" and then use sympy.Symbol and sympy.Integer in the code.
Reading the code I was confused thinking you were creating PSyIR Symbols.

sergisiso (changes requested):
@hiker Some more suggestions, I reviewed all but the dependency_tools_test which I will do next time.

I also plan to test it with NEMO but I am still getting some issues which are not the fault of this PR, hopefully I will resolve it for next review.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -41,16 +41,91 @@
...
                            VariablesAccessInfo)
-from psyclone.errors import InternalError
+from psyclone.errors import InternalError, LazyString
 from psyclone.psyir.nodes import Loop
 from psyclone.psyir.backend.fortran import FortranWriter
+from psyclone.psyir.backend.sympy_writer import SymPyWriter
+from psyclone.psyir.backend.visitor import VisitorError
+
+
+class DTCode(IntEnum):
```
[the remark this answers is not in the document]
hiker: Yes, it was the *

src/psyclone/core/symbolic_maths.py
```diff
@@ -108,13 +165,88 @@ def equal(exp1, exp2):
...
+    def solve_equal_for(exp1, exp2, symbol):
+        '''Returns all solutions of exp1==exp2, solved for
+        the specified symbol. It restricts the solution domain to integer
+        values. If there is an infinite number of solutions, it returns
+        the string 'independent', indicating that the solution of exp1==exp2
+        does not depend on the specified symbol. This is done to avoid that
+        the SymPy instance representing an infinite set is used elsewhere
+        in PSyclone (i.e. creating a dependency in other modules to SymPy).
+        The FiniteSet returned otherwise is compatible with a Python set
+        and so does not introduce any dependencies.
```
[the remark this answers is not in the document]
hiker: Indeed.

src/psyclone/tests/core/component_indices_test.py
```diff
@@ -125,3 +127,47 @@ def test_component_indices_getitem_exceptions():
...
+                                 type(my_type) :: dv(10)
+                                 integer i, j, k
+                                 integer, parameter :: n=10
+                                 real, dimension(n) :: a1
+                                 real, dimension(n,n) :: a2
+                                 real, dimension(n,n,n) :: a3
+                                 {expression} = 1
+                                 end program test''')
+    prog = parser(reader)
+    psy = PSyFactory("nemo", distributed_memory=False).create(prog)
```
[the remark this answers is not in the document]
hiker: I can, thanks!

src/psyclone/psyir/tools/dependency_tools.py
hiker: Done.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+        the full structure  When we use the same function for other tests
+        (loop fusion etc), the subroutines here can be generalised to use
+        the dependency direction (i.e. "<", "=", ">") and instead of
+        checking for distance 0 then check for direction being only "="").
+
+        De-facto all that is happening atm is to try to find one subscript
+        that is guaranteed to be independent, which is all that is needed
+        for parallelisation.
+
+        :param str loop_variables: all loop variables involved.
```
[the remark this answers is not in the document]
hiker: All fixed (and thanks for the explanation on teams)

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -273,102 +224,443 @@ def _is_loop_suitable_for_parallel(self, loop, only_nested_loops=True):
...
+                if not self.is_array_access_parallelisable(loop_variables,
+                                                           write_access,
+                                                           other_access):
+                    # There is a dependency. Try to give precise error
+                    # messages:
+                    if write_access is other_access:
+                        # The write access has a dependency on itself, e.g.
+                        # a(3) = ...    or a((i-2)**2) = ...
+                        # Both would result in a write-write conflict
+                        str_access = self._language_writer(write_access.node)
```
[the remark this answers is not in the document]
hiker: OK, can't see why. I actually tried to actually produce the problem, but couldn't easily. I've given up to understand why you still get a warning, so I've just disabled it with an explanation.

src/psyclone/psyir/tools/dependency_tools.py
hiker: Done.

src/psyclone/psyir/tools/dependency_tools.py
hiker: Done.

hiker:
I think I've addressed all issues. Ready for next review.

src/psyclone/psyir/tools/dependency_tools.py
hiker: Done.

sergisiso (approved):
@hiker All comments have been addressed or there are new issues in the issue tracker covering them. This PR in a great enhancement to the dependency analysis utilities

Other than that
- All new code is covered and all tests pass with compilations.
- pyclodestyle is clean
- I fixed some remaining minor pylint issues. (I also updated some copyright dates)
- No TODOs referencing this PR or issue number.
- The documentation has been improved. All documentations build properly.


The PR is ready to merge.

Issue #1695, opened 2022-04-21
Compilation failures


hiker:
I just noticed that our nightly builds started to fail, potentially in various places. I'll use this ticket of listing all issues. So far I have verified:
- [ ] tutorial/practicals/LFRic/building_code/gungho_lib
```
-  joerg@Joerg-Surface:~/work/psyclone/tutorial/practicals/LFRic/building_code/gungho_lib$ make
gfortran -Wall -g -fcheck=bound -I../../../../../src/psyclone/tests/test_files/dynamo0p3/infrastructure -c base_mesh_config_mod.f90
base_mesh_config_mod.f90:300:15:

  300 |     geometry = emdi
      |               1
Error: Cannot convert INTEGER(4) to CHARACTER(1) at (1)
```
Looks like this was caused by e33cd05f2edce7eba564b6e4ec473ed02f2fb7fa

hiker:
The following patch fixes the compilation issue in the tutorials:
```
diff --git a/src/psyclone/tests/test_files/dynamo0p3/infrastructure/constants_mod.F90 b/src/psyclone/tests/test_files/dynamo0p3/infrastructure/constants_mod.F90
index 2e34041cd..e900d3059 100755
--- a/src/psyclone/tests/test_files/dynamo0p3/infrastructure/constants_mod.F90
+++ b/src/psyclone/tests/test_files/dynamo0p3/infrastructure/constants_mod.F90
@@ -214,7 +214,7 @@ module constants_mod
   integer  (i_def),     parameter :: IMDI  = -huge(0_i_def)   !< Value for integer numbers
   character(str_short), parameter :: CMDI  = 'unset'          !< Value for characters
   character(str_short), parameter :: UNSET_KEY  = CMDI        !< Chararater value for namelist enumerations
-  integer  (i_native),  parameter :: EMDI  = -1_i_native      !< Integer value for namelist enumerations
+  character(str_short), parameter :: EMDI  = 'missing data'       !< Chararater value for namelist enumerations
   !> @}
 
 end module constants_mod
```
It basically reverses the type change of EMDI that was done in e33cd05f2edce7eba564b6e4ec473ed02f2fb7fa. @rupertford , any comments on this change of type? I don't know if this fix is correct, or if it the gungho_lib should be modified instead.

hiker:
[...]

PR #1697, opened 2022-04-22
1695 fix tutorial compilation failures

hiker:
Fixed the broken compilation. Ticket #1695 has all details (I also added a .gitignore  - when I did the compile tests I noticed files that should be ignored).


arporter (approved):
Changes are small and self contained.
All tests, tutorials and examples are still OK with compilation.
No documentation update is required.
I'll proceed to merge.

PR #1699, opened 2022-04-26
Fix failing dynamopsy test by specifying dm=true explicitly.

arporter:
A small fix to one of the dynamopsy tests that was written assuming DM was True but didn't take any steps to ensure that it was.


arporter:
On thinking about this, this test was obviously failing because, on entry to it, DM was set to False. This means that some state has leaked from an earlier test. I'll see if I can find out which one...

arporter:
@mnobre, presumably running *just* the failing test in lfric_dynamopsy_test doesn't fail? Would you mind working backwards to see which test file has to be run before `lfric_dynamopsy_test.py` in order to trigger the error?

arporter:
It turns out that to reproduce this failure I had to mimic exactly what the CI does:

    pytest -n 2 --cov=psyclone --cov-report=xml

I then used `-v` to increase the verbosity so that I could see which tests each process had run prior to the failure. Finally, this let me narrow it down to:

    pytest -x domain/common/extract_driver_creator_test.py domain/lfric/lfric_dynamopsy_test.py

arporter:
This is ready for review now, perhaps by @hiker as he's looked at this too.

hiker:
Changes are small and self contained.
All tests, tutorials and examples are still OK with compilation.
No documentation update is required.
I'll proceed to merge.

hiker (approved):
All tests still passed after updating to latest master.

PR #1700, opened 2022-04-27
Improvements to the NEMO processing scripts

nmnobre:


arporter:
Hi @nmnobre, as discussed, please could you break-out the NEMO-script improvements (and the bug that assumed the directory is named `PSyclone`) as a separate PR?

nmnobre:
Hi @arporter, this is technically ready for review, in the sense that our usual configuration builds and runs with a single process and on a single GPU.
However, I think I should also try running with both the MPI and SI3 modules enabled, but our usual configuration seems to crash at runtime (even on master). Can you point me to a working configuration?

examples/nemo/scripts/kernels_trans.py
```diff
@@ -59,135 +58,76 @@
...
 
 
 class ExcludeSettings(object):
     '''
-    Class to hold settings on what to exclude from OpenACC KERNELS
-    regions.
+    Class to hold settings on what to exclude from OpenACC KERNELS regions.
 
-    :param dict settings: map of settings to override or None.
+    :param dict settings: map of settings to override or {}.
```
arporter, line 115: s/dict/Optional[dict]/ please and s/ or {}//

examples/nemo/scripts/kernels_trans.py
```diff
@@ -226,98 +166,50 @@ def valid_acc_kernel(node):
...
                 return False
-            if excluding.ifs_1d_arrays and \
-               any([len(array.children) == 1 for array in arrays]):
+            # When using CUDA Unified Memory, only allocated arrays reside in
+            # shared memory (including those that are created by compiler-
+            # -generated allocs, e.g. for automatic arrays). We assume that all
+            # arrays of rank 2 or greater are dynamically allocated, whereas 1D
+            # arrays are often static in NEMO. Hence, we disallow IFs where the
+            # logical expression involves the latter.
+            if any([len(array.children) == 1 for array in arrays]):
```
arporter, line 208: We could do better than this now for symbols defined in the current Fortran module. However, that can be for a subsequent PR.

arporter: Also, pylint suggests using a generator here (which means just removing the `[]` in my experience).

examples/nemo/scripts/kernels_trans.py
```diff
@@ -59,135 +58,76 @@
...
+                  "rho_air", "visc_air", "sbc_dcy", "glob_sum",
+                  "glob_sum_full", "ptr_sj", "ptr_sjk", "interp1", "interp2",
+                  "interp3", "integ_spline"]
 
 
 class ExcludeSettings(object):
     '''
-    Class to hold settings on what to exclude from OpenACC KERNELS
-    regions.
+    Class to hold settings on what to exclude from OpenACC KERNELS regions.
```
arporter, line 113: Since we're here, please could you update L111 so that `ExcludeSettings` does not inherit from `object`.

arporter (changes requested):
Looks good so far. I haven't actually tried it with NEMO code yet. Just a couple of minor things to fix and I'll check with NEMO once I'm awake again.

examples/nemo/scripts/kernels_trans.py
nmnobre: I'm guessing we could instead probe the static/dynamic status of arrays by checking for some keywords on the symbol table?

arporter: Yes although it's slightly clunky now that I come to look at it - you have to look at the `shape` of the datatype of the Symbol.

nmnobre: Ah... so I'd still be reasoning on the number of ranks of the involved arrays, not on keywords, like `allocatable` or `pointer`...

PR #1701, opened 2022-04-29
Frankenbranch: lfric psyad test harness

arporter:
This PR exists purely to help me manage the changes that are required to get this functionality to work. The expectation is that the changes themselves will be done as (much) smaller PRs.


arporter:
`field1` corresponds to `r_u` for this kernel which has intent `inout`.

arporter:
All of the three failing cases use the coordinate jacobian:

    (py3) kbc59144@laptop eg2$ grep -l jacobian *F90
    tl_kinetic_energy_gradient_kernel_mod.F90
    tl_rhs_eos_kernel_mod.F90
    tl_vorticity_advection_kernel_mod.F90

Perhaps that's the thing that isn't being initialised correctly? Do any of the mini-apps use it?

arporter:
Resorting to the faithful debugger (print), I've found that the determinant returned by the call to `pointwise_coordinate_jacobian` within the `tl_vorticity_advection_kernel` is sometimes exactly zero which then gives infinities in the inverse jacobian. I suspect this shouldn't be happening!

arporter:
Now that I'm not randomising the coordinate field(!) things are slightly better but the values for chi_3 gathered to the first element look odd:

     ipanel:           0
     chi_1_e:  -0.78539816339744817      -0.71994831644766089      -0.78539816339744828      -0.71994831644766089      -0.78539816339744828      -0.71994831644766100      -0.78539816339744828      -0.71994831644766100     
     chi_2_e:   0.71994831644766089       0.71994831644766089       0.78539816339744828       0.78539816339744817       0.71994831644766089       0.71994831644766100       0.78539816339744839       0.78539816339744828     
     chi_3_e:    0.0000000000000000        7.2759576141834259E-012   0.0000000000000000       -7.2759576141834259E-012   1000.0000000000073        1000.0000000000000        1000.0000000000073        1000.0000000000000     

This is for a run with one level. If I increase to 4 levels (in configuration.nml) then I get:

     ipanel:           0
     chi_1_e:  -0.78539816339744817      -0.71994831644766089      -0.78539816339744828      -0.71994831644766089      -0.78539816339744817      -0.71994831644766100      -0.78539816339744828      -0.71994831644766100     
     chi_2_e:   0.71994831644766089       0.71994831644766089       0.78539816339744828       0.78539816339744817       0.71994831644766089       0.71994831644766100       0.78539816339744839       0.78539816339744817     
     chi_3_e:    0.0000000000000000        7.2759576141834259E-012   0.0000000000000000       -7.2759576141834259E-012   250.00000000000000        250.00000000000728        250.00000000000000        250.00000000000000

arporter:
I realised that the Chi field is currently declared in kernel metadata as being on `any_space_9` when actually it is really on WChi so the basis and differential basis functions would not be right. Correcting that did not fix the problem however.

By comparing the various inputs to pointwise-coord-jacobian in compute_derham_matrices and in my test algorithm I found that the only difference was the `panel_id`. Just as for the chi field, I was randomising this input field but actually it holds geometry information and must be passed through from the initialisation code. Once I do that (and exclude both chi and it from the inner products) I get:

    Test of adjoint of 'tl_vorticity_advection_kernel_type' FAILED:    77057.183370836094        77057.182248782789        77106916.000000000     

which is pretty close but requires further investigation.

arporter:
Looked at moving this code back to using the cut-down infrastructure with PSyclone but that doesn't seem to include a way of getting the panel_id. In the mini-app we do:

    call init_fem( mesh, chi, panel_id )

which comes from `driver_fem_mod`. This doesn't exist in our cut-down infrastructure.

arporter:
In TL:

                    2                    1   2169638175.7990208     
    -2.7178216466045349E-004  -2.7364845804241839E-018  -2.1791517041842129E-003  -5.4997974410486049E-019   2.5948666665765488E-004  -2.2929705365366919E-003   2.4191629442040031E-004  -2.4303578493606616E-004  -2.4481797255768668E-003
    jac:    1.3238903853361965E-007  -5.8794316503285873E-008  -5.3108025052180923E-021  -5.8794316503285873E-008   1.2639972293259070E-007   2.2235334555554988E-020  -5.3108025052180923E-021   2.2235334555554988E-020   1.5999999999999945E-005
 
In Adjoint:

                    2                    1   2169638175.7990208     
    -2.7178216466045349E-004  -2.7364845804241839E-018  -2.1791517041842129E-003  -5.4997974410486049E-019   2.5948666665765488E-004  -2.2929705365366919E-003   2.4191629442040031E-004  -2.4303578493606616E-004  -2.4481797255768668E-003
    jac:    1.3238903853361968E-007  -5.8794316503285866E-008  -5.2939559203393771E-021  -5.8794316503285866E-008   1.2639972293259070E-007   2.2128735747018596E-020  -5.2939559203393771E-021   2.2128735747018596E-020   1.5999999999999945E-005
    Test of adjoint of 'tl_vorticity_advection_kernel_type' FAILED:    14896.748543429499        14896.748461231893        45188612.000000000     

So `jac_inv` (first row of numbers) matches exactly but the resulting `jac` doesn't.

arporter:
Realised that I hadn't updated the metadata of the adjoint kernel. With that fixed the test passes (insert very many celebratory emojis here).

arporter:
Note that I also discovered that my extension of the matmul2code transformation was getting the loop indices round the wrong way. This is fixed in #1716.

arporter:
Currently this PR introduces three new 'builtins':

 1. Setval_random - init a field with random data
 2. Setop_random - init an operator with random data
 3. Setop_X - Set an operator equal to another operator

The last two are problematic because we don't currently support built-ins that write to operators. This means we have no support for generating the appropriate (doubly-nested) dof loops. I've hacked my way around this by making these two kernels operate on the 'domain' and then using the Fortran `random_number` intrinsic which is polymorphic and will fill any shape of array passed to it. In order to make progress, it would probably be better if I introduced user-defined versions of these last two kernels.

rupertford:
In the longer term we should support builtins that iterate over cells (or whatever we call it) shouldn't we?

arporter:
Closing this in favour of #1782.

PR #1702, opened 2022-05-02
#1695-2 New fix for compilation failures, updating to newer lfric usage.

hiker:
Trying a new fix, using the new data structure from LFRic for `emdi`  (--> which becomes `unset_key`)


hiker:
Hi @arporter , this fix should work with your branch. I compared with the code on lfric trunk to see how the new data type are handled, and tried to fix it all. Not sure if it's worth asking @TeranIvy  to verify this (I basically replaced all `edmi` with `unset_key`).

arporter (approved):
Thanks @hiker. This all looks good and matches what I had to do in my branch.
Examples and tests are all OK with compilation.
`lfric/eg17/full_example` also runs OK.
I'll merge this once #1684 is on.

arporter:
Sphinx link-check failure is just Binder. All good.

PR #1730, opened 2022-05-26
(Closes #1347) allow for dependencies between parameter declarations.

### LINKED ISSUE #1347: Declaration ordering error

arporter:


arporter:
This is ready for first review now. One for @hiker or @sergisiso as it uses the dependence analysis.

sergisiso (approved):
@arporter Thanks for solving this, I see it removes TODOs, fixes xfails and fixes Nemo excluded files which is fantastic.

All new lines are covered and the tests pass with compilation.
Pycodestyle and pylint don't show new issues.
It is a bugfix so no new documentation is needed.
No TODOs left.

I tested it with NEMO and processing the new files work.

PR is approved for merging.

Issue #1748, opened 2022-06-08
Update setup.py to fparser 0.0.15


TeranIvy:
#1707 updated the PSyclone master to the lastest fparser submodule which is the 0.0.15 release. However, the `setup.py` in PSyclone still points to 0.0.14 as the prerequisite. This affects the installation of PSyclone from a working copy of the master by using the fparser from the `external` submodules. Installing fparser from the submodule downloads and installs 0.0.15 release first and then installs 0.0.14 from the PSyclone's `setup.py`.
I was a code reviewer for #1707 and overlooked this, so I am happy to fix it.

arporter:
Thanks for fixing Iva. When installing PSyclone for production use, I don't think you should be using the version of fparser in the submodule.

Issue #1707, opened 2022-05-09
PSyclone tests not working with latest fparser


arporter:
As described in https://github.com/stfc/fparser/issues/339, the latest fparser stores global state (symbol tables) and this must be cleared between unit tests.

arporter:
Currently the `parser` fixture is 'session' scoped so it is only called once for a single test session. We do it this way so as to avoid the repeated construction of the classes used in the parser. However, this then means that we can't easily clear parser state in this fixture.

arporter:
stfc/fparser#339 has been merged. All tests now pass on this branch with the latest version of fparser.

PR #1749, opened 2022-06-08
Update fparser version to 0.0.15 in setup.py (for #1748)

TeranIvy:
As outlined in #1748, when the PSyclone master was updated  to the latest fparser submodule (0.0.15 release) in  #1707, the `setup.py` eas omitted and it still points to fparser 0.0.14 as the prerequisite. When PSyclone and fparser are installed from a working copy of the master by using the fparser from the `external` submodules, fparser 0.0.15 is installed first and from the submodule and then fparser 0.0.14 from the PSyclone's `setup.py`.

I updated the fparser version to 0.0.15 in `setup.py`. I installed PSyclone locally from this branch and can confirm that only fparser 0.0.15 is now installed.

### DIFF
--- a/changelog
+++ b/changelog
@@ -83,6 +83,9 @@
 
 	29) PR #1679. Add a limit parameter to the PSyIR node ancestor method.
 
+	30) PR #1749 for #1748. Updates version of fparser specified in setup.py
+	to 0.0.15.
+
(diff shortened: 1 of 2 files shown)


arporter (approved):
Change is fine.

PR #1754, opened 2022-06-13
Don't use python 3.10.0 specifically, we should roll with minor version updates

LonelyCat124:
From dicussion on teams, PR is mostly to check it installs as we expect.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -94,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.8, 3.10.0]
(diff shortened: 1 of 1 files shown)


LonelyCat124:
Aha, this runs into the same issue (and why it specified 3.10.0) I think (which may have also been me). I will look into it.

arporter:
I was getting deja-vu but then realised that was because I did this in fparser: https://github.com/stfc/fparser/blob/4003e817982dc27ff0f9ded7cf8840f57d192ba5/.github/workflows/unit-tests.yml#L54

arporter (approved):
Change is fine. All tests still OK. No need to update anything else.

PR #1769, opened 2022-06-23
unsupported rhs form with literal (closes 1721)

rupertford:


rupertford:
Ready for first review subject to CI happiness.

src/psyclone/psyad/transformations/preprocess.py
```diff
@@ -95,6 +95,10 @@ def preprocess_trans(kernel_psyir, active_variable_names):
         else:
             sym_maths = SymbolicMaths.get()
             sym_maths.expand(assignment.rhs)
+            # The expand function does not expand if one of the terms
+            # is a literal e.g. 0.5*(a+b) so run the associativity
+            # function to catch any such cases.
```
arporter, line 100: Is this really true? I tried:
image

src/psyclone/psyad/transformations/preprocess.py
```diff
@@ -104,7 +108,8 @@ def associativity(assignment, active_variable_names):
     x*b and a*/x +- b*/x respectively.
 
     This function can be removed when support for Range nodes is added
-    to the SymbolicMaths expand function, see issue #1655.
+    to the SymbolicMaths expand function, see issue #1655, and if the
+    expand function can be modified so that it expands literals.
 
     :param assignment: the Assignment Node that we are looking at.
     :type assignment: :py:class:`psyclone.psyir.nodes.Assignment`
```
arporter, line 115: Since we're here, pylint complains about an unnecessary `elif` at L123. Please could you take a look?

src/psyclone/tests/psyad/transformations/test_preprocess.py
```diff
@@ -302,3 +302,27 @@ def test_preprocess_associativity4(fortran_reader, fortran_writer):
...
+        "end subroutine\n")
+    expected = (
+        "subroutine example(a, b, c)\n"
+        "  real :: a\n"
+        "  real :: b\n"
+        "  real :: c\n\n"
+        "  a = 0.5 * b + 0.5 * c\n\n"
+        "end subroutine example\n")
+    psyir = fortran_reader.psyir_from_source(code)
+    preprocess_trans(psyir, ["a", "b", "c"])
```
arporter, line 326: Worth adding a compilation check too?

arporter (changes requested):
Changes are pylint clean.
When I tried SymPy's `expand` it appeared to do the right thing. Please could you double check.
Apart from that there are just a couple of minor things.
There's no need to update documentation.

src/psyclone/psyad/transformations/preprocess.py
rupertford: You're quite right. sympy in Python does what is expected. I need to find out why it is not working via PSyclone.

rupertford:
You are right that sympy does expand such expressions. The problem is that the SymPyWriter.convert_to_sympy_expressions() method is actually expanding the expression if there is a literal. Then, when the actual expand function is called there is a test to see whether the output is different to the input (which it is not as it has already been expanded) and the original is not modified.

rupertford:
OK, I've found the problem. sympy's `parse_expr()`, which is used by PSyclone's SymPy wrapper utilities actually does simplification as well as parsing by default. This can be switched off with evaluate=False. Once this is done we get the expected result from PSyclone's `expand()` function.

src/psyclone/psyad/transformations/preprocess.py
rupertford: The problem was to do with a bug in the PSyclone expand function. I've fixed this now.

src/psyclone/psyad/transformations/preprocess.py
rupertford: Fixed.

src/psyclone/tests/psyad/transformations/test_preprocess.py
rupertford: Added.

rupertford:
The fix causes other problems. For some reason the sympy expression that comes out of `parse_expr()` differs when you use the `evaluate=False` option which causes an equality test to fail even though the string output is exactly the same. I've therefore gone with fixing the `expand()` function by checking equality of the generated result and the original rather than the output of `parse_expr` and the output of `expand`. We can then drop the `evaluate=False` as we don't mind if it does any of its own changes to the expression. By fixing this, `example/psyir/eg1` now fails as `expand` reorders `a-bc` to `-bc+a` which triggers the unary minus bug. More worryingly I am getting a symbol table error in a test and in psyir/eg2 which is to do with the expand function. I need to find out what is going on here.

rupertford:
The symbol table error was another minor bug in expand which was getting the wrong symbol table in certain cases. I've fixed this and all seems well.

rupertford:

rupertford:
Note, I fixed some f-string pylint errors in two files which now have no changes in them other than the f-string changes. They did at the time but I removed the changes them as they were no longer relevant. I thought I might as well keep the f-string changes though.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -309,7 +310,13 @@ def test_sym_writer_convert_to_sympy_expressions(fortran_reader):
     exp2 = psyir.children[0].children[1].rhs
     sympy_list = SymPyWriter.convert_to_sympy_expressions([exp1, exp2])
 
-    assert sympy_list[0] == parse_expr("a%a_b_1 + a%a_c(1) + i")
+    # Equals is failing here when checking that expr and sympy_list[0]
```
arporter, line 313: This comment seems to be out of date?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -32,13 +32,14 @@
 # POSSIBILITY OF SUCH DAMAGE.
 # -----------------------------------------------------------------------------
 # Author: J. Henrichs, Bureau of Meteorology
+# Modified: R. W. Ford, STFC Daresbury Lab
 
 ''' Module containing py.test tests the SymPy writer.'''
 
 from __future__ import print_function, absolute_import
```
arporter, line 39: Please rm this line.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -32,13 +32,14 @@
...
 
 import pytest
-from sympy import Function, Symbol
+from sympy import Function, Symbol, simplify
```
arporter, line 42: pylint says this import is unused.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -32,6 +32,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 # -----------------------------------------------------------------------------
 # Author: J. Henrichs, Bureau of Meteorology
+# Modified: R. W. Ford, STFC Daresbury Lab
```
arporter, line 36: The only changes to this file appear to be your name and a blank line. If that's so, please could you revert.

arporter (changes requested):
Looking good now.
Just a few minor things to address.
Tests and examples are fine with compilation.
Just one pylint issue to address.
Ref. guide builds with no new warnings.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
rupertford: Oops.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
rupertford: Removed.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
rupertford: Oops.

src/psyclone/psyir/backend/sympy_writer.py
rupertford: Oops, reverted.

rupertford:

arporter (approved):
Tests and examples are fine with compilation.
Will proceed to merge.

PR #1774, opened 2022-06-27
Fix error in test code (closes 1773)

rupertford:


rupertford:
Ready for first review. This is an easy two liner change so anyone could review.

Issue #1712, opened 2022-05-13
Additional info to store in messages from the DependencyAnalysis tool


hiker:
The newly integrated message system in the dependency analysis (#1693) stores additional information besides an error/warning code and the message strings. Initially these additional messages are the variable names involved, but Sergi suggested to store pointers to the psyir, e.g.:
```
self._add_message(LazyString(
    lambda: f"The write access to '{self._language_writer(write_access.node)}' causes a"
    f"write-write race condition."),
    DACode.ERROR_WRITE_WRITE_RACE,
    [write_access.node])
```
We should agree what exactly this 'additional information' is - variable names, variables in the index expressions, psyir info (either of variable, or only of the index expression), maybe store psyir info of the access plus ComponentIndex of the subscript involved, ...

hiker:
In #1751 a transformation is introduced that removes induction variables. Ideally this transformation should be executed before calling the dependency analysis, e.g. to remove:
```
do i=1, n
   im = i-1
   a(i) = a(im)+1
enddo
```
On the other hand, this transformation will change the code (especially it will move assignment to the variables after the loop), which might not be expected (and could affect other transformations). 
If on the other hand, we create a copy in the dependency tools, then any reference in the message object will now point to a copied not, not the original node anymore :(

PR #1778, opened 2022-06-28
1751 induction variables

hiker:
This partially implements handling of induction variables. Only missing are recursive ones (`k=k+3`), but Sergi indicated that this is the part that blocks him most, so here you go :)


hiker:
This is ready for a first review. Main question might be: how to integrate this with the dependency analysis. It would be ideal if this could be done in a copy of the tree inside the dependency_tools, but depending on how we implement #1712 (additional info in messages - a node or just a string) this might not work. Most sophisticated solution: the copy of a tree adds a pointer to the original node for each node :)   Though that's likely overkill.

For now I've added a note to the dependency tool section in the dev guide to indicate that this transformation should be run first before calling the dependency tools.

My suggestion: we add a ticket about how to best integrate the induction removal, and try to get some experience in actually using it. This might make it much clearer than just thinking about it :)

Related to the usage is also what to do with removed assignments inside of the loop. ATM, these assignments are added at the end of the loop (so if these values should be used, they will have appropriate values). But this can prevent loop fuse etc (though maybe these operations would run before???). We could add more checks (e.g. if an induction variable is only read after another write statement, it is not needed).

All in all, I think it would be great if someone (@sergisiso ??) could have a look to see where to take this, if more should be done (or postponed via additional tickets).

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+        :param psyir: the PSyIR in which to replace `original` with \
+            `replacement`.
+        :type psyir: :py:class:`psyclone.psyir.nodes.Node`
+        :param original: the reference which is to be replaced.
+        :type original: :py:class:`psyclone.psyir.nodes.Reference`
+        :param replacement: the node to insert.
+        :type replacement: :py:class:`psyclone.psyir.nodes.Node`
+
+        '''
+        # Just in case we avoid modifying the tree while walking it,
```
sergisiso, line 123: Add a comma after "Just in case".

And is this really a problem? I think it shouldn't be

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+      (e.g. `k = k + 3`).
+    * there must only be one assignment to the variable.
+
+    '''
+    def __str__(self):
+        return "Replaces all induction variables in a loop."
+
+    # ------------------------------------------------------------------------
+    @staticmethod
+    def _replace_references(psyir, original, replacement):
```
sergisiso, line 110: This could possibly be a generic node functionality, since the walk(Reference) could also be walk(Node) as the equals already check the types match. For instance:
`node.replace_all(original, replacement)`
or
`node.replace_subexpressions(original, replacement)`
but maybe this is for another PR.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+            return False
+
+        return True
+
+    # ------------------------------------------------------------------------
+    def apply(self, node, options=None):
+        '''Apply the ReplaceInductionVariables transformation to the
+        specified node. The node must be a loop. In case of nested
+        loops, the transformation might need to be applied several
+        time, from the inner-most loop outwards.
```
sergisiso, line 191: s/time/times/

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+
+        :param node: a Loop node.
+        :type node: :py:class:`psyclone.psyir.nodes.Loop`
+
+        '''
+        self.validate(node)
+        loop_var = node.variable.name
+
+        # Find assignments that are directly part of the loop (this avoid
```
sergisiso, line 200: s/avoid/avoids/ or prevents

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+        indx = 0
+        while indx < len(node.loop_body.children):
+            assignment = node.loop_body.children[indx]
+            # Only handle assignments, ignore if statements etc
+            if not isinstance(assignment, Assignment):
+                indx += 1
+                continue
+
+            # Assignment to arrays are ignored as well:
+            if isinstance(assignment.lhs, ArrayReference):
```
sergisiso, line 212: This won't catch Structure of arrays, e.g. "var.field(3) = 4" is this a problem/. In the transformation docstring you mention that this applies to "non-structure scalars" but I can't see where this is checked, shouldn't this be one of this checks with "indx+=1;continue"?

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
```diff
@@ -0,0 +1,188 @@
...
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# ----------------------------------------------------------------------------
+# Author: J. Henrichs, Bureau of Meteorology
+
+'''This module tests the ReplaceInductionVariables transformation.
+'''
+
+from __future__ import absolute_import, print_function
```
sergisiso, line 39: You should be able to remove this line without pylint complaining

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
```diff
@@ -0,0 +1,188 @@
...
+    assert "ic5 = (i - 1) * (i - 1) + 3 * SIN(i - 1)" in out_all
+    assert ("a(i + 1 + invariant + (i * i + 3 * SIN(i))) = 6 + "
+            "(i + 1 + invariant + (i * i + 3 * SIN(i)) + 1) * "
+            "(i + 1 + invariant + (i * i + 3 * SIN(i)))" in out_loop)
+    assert ("ic6 = i - 1 + 1 + invariant + ((i - 1) * (i - 1) + "
+            "3 * SIN(i - 1))" in out_all)
+
+
+# ----------------------------------------------------------------------------
+def test_riv_not_movable(fortran_reader, fortran_writer):
```
sergisiso, line 132: Can you create a similar version to this test but to check multiple structure and array combinations in the LHS.
E.g. Structure-scalar, array-accessed scalar, SoA, AoS.
I understand this will not be transformed right?

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
```diff
@@ -0,0 +1,188 @@
...
+    '''Tests assignments that cannot be moved.
+    '''
+    source = '''program test
+                integer i, ic1, ic2, ic3, ic4, ic5
+                real, dimension(10) :: a
+                do i = 1, 10
```
sergisiso, line 138: can we also add an induction variable with a function call in the RHS. This should not be transformed because we don't know if it has side effects that are sensitive to the number of times they are called. (Note this is not equivalent to the sin() in the previous test as sin should be a Operation and it is fine because all intrinsic operations that we support are pure functions)

sergisiso (changes requested):
@hiker The implementations looks good, I suggested some docstring fixes/changes and I ask for a bit more testing about what happens with induction variables that are arrays or structures.

Regarding:
> My suggestion: we add a ticket about how to best integrate the induction removal, and try to get some experience in actually using it. This might make it much clearer than just thinking about it :)

I agree

> But this can prevent loop fuse etc (though maybe these operations would run before???). We could add more checks (e.g. if an induction variable is only read after another write statement, it is not needed).

I would not do it in this PR  and let this be part of the "get some experience in actually using it" you mentioned before.

In my opinion this is a self-contained transformation and I would be happy to merge this, and let the integration with the dependency analysis be done separately when we have a better idea of how it should look.

src/psyclone/psyir/transformations/replace_induction_variables.py
hiker: Agreed with both points: could be useful, and happy to move this if we have the same requirement elsewhere, for now I just leave it here.

src/psyclone/psyir/transformations/replace_induction_variables.py
hiker: No idea what happened here - I was certain I had a a test for arrays, and somehow thought I tested for References (not instance of Reference).
I've fixed it all - it now supports structures, as long as there is no index used anywhere. Tests also added.
Note that I could not get the parser to handle a function call. So in the corresponding test for function calls on the RHS I had to manually modify the PSyIR to insert a 'Call' (which is a function call). Do we need a ticket for the issue with function calls?

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
hiker: I've added test cases for plain structures (that do work now) and arrays/structure, that do not work.

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
hiker: Done. As mentioned previously, I had to modify the PSyIR, I couldn't get it to read in a function (it ended up a codeblock only, not a call).

hiker (commented):
Good catches, especially the missing handling of Arrays and not-working handling of structures :(

doc/developer_guide/dependency.rst
```diff
@@ -598,6 +598,14 @@ uses `SymPy` internally to compare expressions symbolically.
     :members:
 
 
+.. note:: PSyclone provides :ref:`user_guide:replace_induction_variable_trans`,
```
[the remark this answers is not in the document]
hiker: Fixed.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: J. Henrichs, Bureau of Meteorology
+
+'''Module providing a transformation that removed induction variables from
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+a loop. '''
+
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
hiker: Indeed.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+    ...         im = i - 1
+    ...         ic = 2
+    ...         tmp(i) = ic * im
+    ...     enddo
+    ... end subroutine sub""")
+    >>> loop = psyir.walk(Loop)[0]
+    >>> ReplaceInductionVariables().apply(loop)
+    >>> print(FortranWriter()(psyir))
+
+    will generate:
```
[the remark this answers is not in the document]
hiker: I was actually thinking that, but when I checked some other files, none of them were setup to allow testing. Apparently I looked at the wrong files. I've fixed this now, so this comment can be tested.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+
+          do i = 1, 100, 1
+            tmp(i) = 2 * (i - 1)
+          enddo
+          ic = 2
+          im = i - 1 - 1
+
+        end subroutine sub
+
+    After the loop the replaced assignments to the induction variables are
```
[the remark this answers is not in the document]
hiker: Indeed it is.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,272 @@
...
+    variable:
+
+    * the variable must be a non-structure scalar.
+    * none of variables on the right-hand side can be written in the loop body
+      (the loop variable is written in the Loop statement, not in
+      the body, so it can be used).
+    * the assigned variable must not be read before the assignment.
+    * the assigned variable cannot occur on the right-hand side
+      (e.g. `k = k + 3`).
+    * there must only be one assignment to the variable.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/transformations/replace_induction_variables.py
hiker: After a lot of thinking I agreed, and I combined the two loops. The only thing I could think of is using an array twice (`a(a(i))`) somewhere. But even this shouldn't be a problem - the outer one will be replaced, and if it then handles the inner one (which is not not part of the tree anymore, who cares :)

src/psyclone/psyir/transformations/replace_induction_variables.py
hiker: Done.

src/psyclone/psyir/transformations/replace_induction_variables.py
hiker: Done.

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
hiker: Yep.

hiker:
That should be ready for next review.  I've opened #1788 to discuss integrating this with the dependency tools.

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
```diff
@@ -0,0 +1,266 @@
...
+    riv = ReplaceInductionVariables()
+    riv.apply(loop)
+    out = fortran_writer(psyir)
+    assert "a(i + 1) = 1 + (i + 1 + 1) * (i + 1)" in out
+    assert "ic1 = i - 5 + 1" in out
+    assert "ic2 = 3" in out
+
+
+# ----------------------------------------------------------------------------
+@pytest.mark.parametrize("array_expr", ["ic(i)", "ic(2)", "t(i)%b", "t(1)%b"])
```
sergisiso, line 202: For completeness can you add some SoA ("mys%a(i)" and "mys%b(i)%a") I did it myself and it works without any additional declaration because the unqualified import

sergisiso (changes requested):
@hiker Almost there now, just some last changed needed and then this can be merged.

I checked formatting, linting, tests, coverage and all pass. There is warning when building the developer guide (see inline comment).

src/psyclone/psyir/transformations/replace_induction_variables.py
hiker: This is apparently a known issue, so this is ok.

doc/developer_guide/dependency.rst
```diff
@@ -590,14 +590,22 @@ Dependency Tools
...
-to provide useful tools for dependency analaysis. It especially provides
+to provide useful tools for dependency analysis. It especially provides
 messages for the user to indicate why parallelisation was not possible. It
 uses `SymPy` internally to compare expressions symbolically.
 
 .. autoclass:: psyclone.psyir.tools.dependency_tools.DependencyTools
     :members:
 
 
+.. note:: PSyclone provides :ref:`user_guide:replace_induction_variable_trans`,
```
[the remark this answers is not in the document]
hiker: I believe this is because the label in the user guide is not on read-the-docs. When I grep'ed, the new label is defined in the users guide. Or is there anything I can do before this is merged?

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,279 @@
...
+        for node in psyir.walk(Reference):
+            if node == original:
+                copy = replacement.copy()
+                node.replace_with(copy)
+
+    # ------------------------------------------------------------------------
+    @staticmethod
+    def _is_induction_variable(assignment, accesses_in_loop_body):
+        '''Tests if the assignment is an induction statement that can be
+        replaced. An induction statements requires:
```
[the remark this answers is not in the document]
hiker: Good idea.

src/psyclone/psyir/transformations/replace_induction_variables.py
```diff
@@ -0,0 +1,279 @@
...
+
+    The following restrictions apply for the assignment to an induction
+    variable:
+
+    * the variable must be a scalar (i.e. no array access at all,
+      not even a constant like a(3) or a%b(3)%c)
+    * none of variables on the right-hand side can be written in the loop body
+      (the loop variable is written in the Loop statement, not in
+      the body, so it can be used).
+    * Only intrisic function calls are allowed on the RHS (since they
```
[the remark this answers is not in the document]
hiker: Indeed.

src/psyclone/tests/psyir/transformations/replace_induction_variables_test.py
hiker: I've added more tests here.

doc/developer_guide/dependency.rst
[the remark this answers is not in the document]
arporter: If the Sphinx link checking is passing in the test suite then this is OK.

sergisiso (approved):
@hiker All comments have been addressed, this is ready to merge

sergisiso:
@hiker I just noticed that the new transformation does not have the Trans postfix as all other transformations, I will rename it and its references before merging.

hiker:
> @hiker I just noticed that the new transformation does not have the Trans postfix as all other transformations, I will rename it and its references before merging.

Good point, thanks!

Issue #1779, opened 2022-06-29
Rename the NEMO API?


arporter:
@mnobre has suggested that we should rename the NEMO API. @hiker is in favour and suggested "EVOLUTION". We could go with "EVO" (as in Mitsubishi Evo which is fast :-)) or perhaps something else. What do people think?

nmnobre:
Bold, and perhaps too bold... what about no name at all? Dynamo and GOcean would simply be treated as exceptions to the "default" mode of operation on generic codes.
For instance, instead of:
```
if isinstance(sched, DynInvokeSchedule):
  from psyclone.dynamo0p3 import DynACCEnterDataDirective as AccEnterDataDir
elif isinstance(sched, GOInvokeSchedule):
  from psyclone.gocean1p0 import GOACCEnterDataDirective as AccEnterDataDir
elif isinstance(sched, NemoInvokeSchedule):
  from psyclone.nemo import NemoACCEnterDataDirective as AccEnterDataDir
else:
  raise InternalError()
```
we would now write:
```
from psyclone.psyir.nodes import ACCEnterDataDirective
if isinstance(sched, DynInvokeSchedule):
  from psyclone.dynamo0p3 import DynACCEnterDataDirective as ACCEnterDataDirective
elif isinstance(sched, GOInvokeSchedule):
  from psyclone.gocean1p0 import GOACCEnterDataDirective as ACCEnterDataDirective
```

arporter:
@sergisiso, @rupertford and I discussed this earlier and came down in favour (I think) of having no name as there's no actual API. Therefore, psyclone without `-api` would just do code transformation. This would break the MO build system as currently we have `dynamo0.3` as the default API and I think their script relies on that - i.e. it doesn't currently specify `-api`.

An alternative option we discussed was to have an entirely different script as entry point for the code-transformation path, e.g. `psytrans` instead of `psyclone` or similar.

The one thing we definitely concluded was that we need to take a look at those classes that have been specialised for 'NEMO' and see what functionality it is that they add.

arporter:
![image](https://user-images.githubusercontent.com/17409212/176451965-3eea3c66-5d0b-485c-9cc4-377daccde9a7.png)

arporter:
`NemoAllArrayAccess2LoopTrans` just uses `NemoArrayAccess2LoopTrans`. This uses the information from the configuration file to name the loop variable that is introduced (i.e. according to which dimension of the array is being accessed).

[...]

PR #1784, opened 2022-07-04
(towards #1779) Remove API bounds from Range2Loop transformations

sergisiso:


sergisiso:
@arporter This is ready for review, as noted in #1779 this does not make the transformation generic quite yet, it still creates NemoLoop and NemoKern. But I think a "towards" PR is appropriate here because we agreed to remove the bound naming, at least from the same transformation, but I am still not decided how to tackle the other things in the PR. And at the moment this is what is blocking other Nemo OpenMP work.

src/psyclone/tests/domain/nemo/transformations/nemo_arrayrange2loop_trans_test.py
```diff
@@ -71,48 +71,9 @@ def test_transform():
...
-        "    enddo\n"
-        "  enddo" in result)
-    assert Compile(tmpdir).string_compiles(result)
-
-
-def test_apply_fixed_bounds(tmpdir):
-    '''Check that the apply method uses bounds information from the range
-    node if it is supplied.
+    '''Check that the apply method uses the range bounds if these are specified
+    or the LBOUND, UBOUND expressions if they are not. Also check the the body
```
arporter, line 75: s/the the/that the/

src/psyclone/domain/nemo/transformations/nemo_arrayrange2loop_trans.py
```diff
@@ -139,75 +139,10 @@ def apply(self, node, options=None):
...
-        except IndexError:
-            lower_bound_info = None
-            upper_bound_info = None
-            loop_variable_name = symbol_table.next_available_name("idx")
-
-        # Lower bound
-        lower_bound = None
-        if not array_reference.is_lower_bound(array_index):
-            # The range specifies a lower bound so use it
-            lower_bound = node.start.copy()
```
arporter, line 163: This is functionality we need to keep - otherwise we'll always be generating loops over the full range of the array when that might not be what the original code did.

src/psyclone/tests/domain/nemo/transformations/nemo_arrayrange2loop_trans_test.py
```diff
@@ -71,48 +71,9 @@ def test_transform():
...
-        "            umask(ji,jj,jk,jt,idx) = vmask(ji,jj,jk,jt,idx) + 1.0\n"
-        "          enddo\n"
-        "        enddo\n"
-        "      enddo\n"
-        "    enddo\n"
-        "  enddo" in result)
-    assert Compile(tmpdir).string_compiles(result)
-
-
-def test_apply_fixed_bounds(tmpdir):
```
arporter, line 113: We still need this test and associated functionality.

arporter (changes requested):
I like the massive simplification (although slightly regret losing the appropriately-named loop variables as I think that affects readability of the generated code).
However, it's gone a bit too far such that I think we will generate incorrect code for e.g.

    ztmp(2:jpk-1,:,:) = 0.0

Please could you check.
Apart from that, all tests and examples are OK with compilation.
Does the documentation for the class (and or docstring) need updating?

src/psyclone/tests/domain/nemo/transformations/nemo_arrayrange2loop_trans_test.py
```diff
@@ -416,11 +274,14 @@ def test_apply_structure_of_arrays_multiple_arrays(fortran_reader,
     result = fortran_writer(psyir)
     print(result)
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/tests/domain/nemo/transformations/nemo_arrayrange2loop_trans_test.py
sergisiso: Again, I did merge both tests because I don't think this is a special case for this transformation, but note that the middle loop in the remaining test has fixed ranges and the generated code does respect them.

src/psyclone/domain/nemo/transformations/nemo_arrayrange2loop_trans.py
arporter: Oh I see, thanks.

arporter (changes requested):
All OK now and I was going to go ahead but then I checked and it turns out we do have an `ArrayRange2LoopTrans` but that this NEMO-specific version does not inherit from it! Please could you take a look? In particular, that generic transformation expects to be given an Assignment while this version expects to be given a Range.

arporter (approved):
We've agreed to punt the discussion about the duplication between `NemoArrayRange2LoopTrans` and `ArrayRange2LoopTrans` to issue #1779.

Issue #924, opened 2020-10-02
[nemo] Array-valued function needs to be inlined for OpenACC performance


arporter:
For example: sbcblk.f90, cp_air() function.

arporter:
e.g. if we have some code:

    my_array(:,:) = cp_air(array1(:,:)) * array2(:,:) + array3(:,:)

where `cp_air` is an array-valued function that performs compute then this is hard to parallelise/ensure it remains GPU resident. (If we put ACC KERNELS inside `cp_air` then we can't have it around this assignment.) Ideally we need to refactor this:

    my_tmp(:,:) = cp_air(array1(:,:))
    my_array(:,:) = my_tmp(:,:)*array2(:,:) + array3(:,:)

That then allows us to have KERNELS inside cp_air() and, separately, around the assignment.

Better still would be to in-line cp_air but we have to walk before we can run.

rupertford:
If I remember correctly, some of the existing kernel transformations (that replace intrinsics) have logic similar to the above. Perhaps the splitting out of an intrinsic/function should have its own transformation which can then be shared.

arporter:
Yes, I said as much to Chris when we were discussing this. Hopefully there's some reuse possible.

arporter:
Obviously, a step 0 for this work is to be able to identify array-valued functions. Currently, all RoutineSymbols that represent Fortran functions are given DeferredType (#1294).

arporter:
This is going quite well. I need to handle the case where there's a name clash for a `ContainerSymbol` as we can't simply rename it in that case.

arporter:
Note to self: I need to identify when a routine accesses variables made available from some outer scoping region as this prevents inlining.

arporter:
The first review and subsequent work on `validate` has thrown up a number of things that I'm adding TODOs for this time around:

  - [ ] `validate` will be much nicer once we can query the `datatype` of any `Reference` (#1799)
  - [ ] `Range` needs extending to return the PSyIR for the number of elements it represents
  - [ ] Support needs to be added for routines with named arguments.

sergisiso:
This is also one of the major issues in OpenMP for GPU now, mostly in the sbcblk region.
Some functions that will need to be inlined are:  q_sat, sbc_dcy, gamma_moist, cd_neutral_10m, psi_h, psi_m

arporter:
[...]

Issue #1799, opened 2022-07-14
Add support for querying the data type of a PSyIR expression.


arporter:
From #1696:

You maybe already hinted it here:

    It may be that we really want the routine (rank_of_subsection) to give us back the shape of any subsection

But this recursion method to find the slice rank is somewhat similar to what @rupertford was planning to implement in order to 'bubble up' the datatypes too all PSyIR expression (datanodes) nodes. I can't find the specific issue but we had a chat during Easter week and Rupert was about to started implementing it. (starting with operations but accessors must also be part of it for a complete solution)

Given that this .datatype property for all expression nodes will be more generic (and would avoid having the slightly confusing "rank_of_subsection" in nodes that have nothing to do with arrays like member and structurereference) I think the methods you add in this PR could at some point be superseded by .datatype and simply be:

rank_slice = len(top_node.datatype.shape)

Note that for a reference `a(3,:)`

    ref.symbol.datatype is ArrayType([3, Extent.ATTRIBUTE])

but

    ref.datatype is ArrayType([Extent.ATTRIBUTE])

which is the information you need.

sergisiso:
My comments regarding the initial "towards" PR of this issue:
- Currently resulting datatype of accesses to UknownTypes can become DeferredType,  my worry with deferredtype is that it could imply that the datatype could be resolved if the proper datatype information is brought into scope from the modules, but in this case it isn't true.
 - I don't see any problem in having arrays of DeferredType which we avoided in the first PR. It just expresses our current knowledge of the type. For more we need to import the module information.
- Currently the datatype comparisons strict. types are equal or not. But in reality we can have something like "is DeferredType equals to INTEGER_TYPE?" or "is 1D Array of length extend.attribute equal to 1D array of extend 3?" It seems it would be useful to have an interface to return True to this because they could be valid for our purposes.

arporter:
Currently, `datatype` for an ArrayMixin ignores the lower bound of each dimension (implicitly assuming it is unity). However, in #1973 I've realised that the lower bound is actually important (since Fortran allows it to be specified and that for a dummy argument can differ from the actual argument). I'm therefore wondering whether the `shape` returned by `datatype` should actually include bounds information rather than just extent? What do @sergisiso, @rupertford and @hiker think?

rupertford:
I think that shape also supports a tuple of values that allows both lower and upper bounds to be set.

arporter:
[...]

PR #1814, opened 2022-07-26
Initial inlining support (towards #924)

arporter:


arporter:
This is ready for first review now. One for @rupertford or @sergisiso.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+    ...     real, intent(inout) :: x
+    ...     x = 2.0*x
+    ...   end subroutine sub
+    ... end module test_mod"""
+    >>> psyir = FortranReader().psyir_from_source(code)
+    >>> call = psyir.walk(Call)[0]
+    >>> inline_trans = InlineTrans()
+    >>> inline_trans.apply(call)
+    >>> # Uncomment the following line to see a text view of the schedule
+    >>> # print(psyir.walk(Routine)[0].view())
```
sergisiso, line 80: Also, if possible I prefer finishing the doctest with e.g. `>>> print(FortranWriter()(psyir.walk(Routine)[0]))` and then have the output displayed (e.g like in src/psyclone/psyir/transformations/hoist_trans.py)

It makes for clear documentation and a complete doctest check

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+    (#924) the routine must also be in the same source file as the call.
+
+    '''
+    def apply(self, node, options=None):
+        '''
+        Takes the body of the routine that is the target of the supplied
+        call and replaces the call with it.
+
+        :param node: target PSyIR node.
+        :type node: subclass of :py:class:`psyclone.psyir.nodes.Routine`
```
sergisiso, line 93: I would remove "subclass of"

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+        :param options: a dictionary with options for transformations.
+        :type options: Optional[Dict[str, str]]
```
sergisiso, line 95: It doesn't matter much in this transformation but I think in general options should be Optional[Dict[str, Any]] as sometimes the value is a boolean or a int

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+        self.validate(node, options)
+
+        # The table we will copy symbols into.
+        table = node.scope.symbol_table
+        # Find the routine to be inlined.
+        orig_routine = self._find_routine(node)
+
+        if not orig_routine.children or isinstance(orig_routine.children[0],
+                                                   Return):
+            # Called routine is empty so just remove the call.
```
sergisiso, line 107: What would happen if this would be a function that only returned a value? I haven't seen any functions in the tests, have these been considered?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+            # Check whether this symbol is used to specify the precision of
+            # any literals.
+            if old_name in precision_map:
+                for lit in precision_map[old_name]:
+                    # A literal is immutable so create a new one with the
+                    # updated symbol as its precision.
+                    dtype = ScalarType(lit.datatype.intrinsic, old_sym)
+                    lit.replace_with(Literal(lit.value, dtype))
+
+    def validate(self, node, options=None):
```
sergisiso, line 281: Have you considered cases like the one below, where there is a remapping in the shape of a passed argument. I think this is valid Fortran and can't be inlined without also expressing the remapping.
```
subroutine main
  REAL A(100, 100)
  CALL S(A(26,2),N)
end subroutine

SUBROUTINE S(X,M)
  REAL X(*)
  DO I = 1, M
     X(I) = X(I) + M
  ENDDO
END
```

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+        if return_stmts:
+            if len(return_stmts) > 1 or not isinstance(routine.children[-1],
+                                                       Return):
+                # Either there is more than one Return statement or there is
+                # just one but it isn't the last statement of the Routine.
+                raise TransformationError(
+                    f"Routine '{name}' contains one or more "
+                    f"Return statements and therefore cannot be inlined.")
+
+        if routine.walk(CodeBlock):
```
sergisiso, line 329: Similar to these, have you thought if any of the Fortran symbol attributes that end up in UnknownFortranTypes could cause problems or are we OK accepting them all?

sergisiso (changes requested):
@arporter I just had time for a quick review today. It looks good and readable, I tried to come up with some cases that may fail and that the validation could catch. See inline comments.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+    >>> code = ("""
+    ... module test_mod
+    ... contains
+    ...   subroutine run_it()
+    ...   integer :: i
+    ...   real :: a(10)
+    ...   do i=1,10
+    ...     a(i) = 1.0
+    ...     call sub(a(i))
+    ...   end do
```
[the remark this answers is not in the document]
arporter: Oops.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,406 @@
...
+from psyclone.psyir.transformations.transformation_error import (
+    TransformationError)
+
+
+class InlineTrans(Transformation):
+    '''
+    This transformation takes a Call and replaces it with the body of the
+    target routine. It is used as follows:
+
+    >>> from psyclone.psyir.frontend.fortran import FortranReader
```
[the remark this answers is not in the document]
arporter: Ooh, I didn't think of trying that. Turns out I had a superfluous parenthesis.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Good idea and thanks for pointing me at an example.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Yes.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Oh yes, fixed.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Ah, not until now!

arporter: To keep things simple for now, I've just extended validate to reject this case.

src/psyclone/psyir/transformations/inline_trans.py
arporter: *Sigh* no, you're quite right, thanks for pointing this out. Again, for now I think I'll just reject such cases in the `validate` method.

arporter: It turns out PSyclone generates an error for an assumed-shape dummy argument so that's not possible at the moment. However, Fortran is quite happy with:

    program main
      real :: f1(100,100), f2(5,10)
      f1 = 1
      f2 = 1
      call kern(f1, f2, 10)
      write(*,*) f1(1:20,1)
    contains
      subroutine kern(field1, field2, n)
        integer, intent(in) :: n
        real, intent(inout) :: field1(n)
        real, intent(inout) :: field2(5,*)
        field1(:n) = 0.0
        field2(:,1) = 0.0
      end subroutine kern
    end program main

I seem to remember that @hiker discovered this recently too.

arporter: I've added a check and test for this. While doing this I realised that named arguments are also going to be a problem so I've excluded those too. I've also excluded cases where the call argument is an array subsection as that's not correctly handled yet.

src/psyclone/psyir/transformations/inline_trans.py
arporter: I guess I've partially handled this with my previous comment - named arguments are now checked for and excluded. I've added a test with a subroutine that has a pointer as argument (which works fine). Now that I write that though I realise I need to check for variables with initial values or the `SAVE` attribute.

arporter: Actually, I've added a test for `SAVE` and it works fine. I can't think of any other potential problems at the moment.

arporter:
Ready for next review from @sergisiso now.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,478 @@
...
+
+    @staticmethod
+    def _find_routine(call_node):
+        '''
+        Searches for the definition of the routine that is being called by
+        the supplied Call.
+
+        Currently only supports routines that are present in the
+        same source file - TODO #924.
+
```
sergisiso, line 453: argument docstring missing

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,478 @@
...
+                f"should be a Call but found '{type(node).__name__}'.")
+
+        name = node.routine.name
+
+        # Check that we can find the source of the routine being inlined.
+        routine = self._find_routine(node)
+
+        if not routine.children or isinstance(routine.children[0], Return):
+            # An empty routine is fine.
+            return
```
sergisiso, line 342: Can you move this after the check for return_symbol immediately below. Otherwise a function with only a return may be considered an empty routine.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,478 @@
...
+                if not sym.is_import:
+                    raise TransformationError(
+                        f"Routine '{routine.name}' cannot be inlined because "
+                        f"it accesses variable '{ref.symbol.name}' from its "
+                        f"parent container.")
+
+        # Check that the shape of any dummy array arguments are the same as
+        # those at the call site.
+        visitor = FortranWriter()
+        for dummy_arg, actual_arg in zip(routine_table.argument_list,
```
sergisiso, line 414: Maybe you can add a `# pylint: disable=cell-var-from-loop` here explaining that the the loop variables we are capturing are fine because the loop is ended after capturing them.

src/psyclone/tests/psyir/symbols/datatype_test.py
```diff
@@ -37,7 +37,6 @@
 
 ''' Perform py.test tests on the psygen.psyir.symbols.datatype module '''
```
sergisiso, line 38: s/psygen/psyclone/ and the order of the imports is not alphabetical.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -0,0 +1,969 @@
...
+from psyclone.psyir.transformations import (InlineTrans,
+                                            TransformationError)
+from psyclone.tests.utilities import Compile
+
+# init
+
+
+def test_init():
+    '''Test an InlineTrans transformation can be successfully created.'''
+    hoist_trans = InlineTrans()
```
sergisiso, line 52: s/hoist/inline/

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,478 @@
...
+        if not routine_sym.is_local:
+            raise TransformationError(
+                f"Routine '{name}' is imported and therefore cannot currently "
+                f"be inlined - TODO #924.")
+        table = routine_sym.find_symbol_table(call_node)
+        for routine in table.node.walk(Routine):
+            if routine.name == name:
+                return routine
+
+        raise InternalError(
```
sergisiso, line 472: I wonder if this should be a TransformationError instead, things like recursive, pure or elemental keywords will put routines in CodeBlocks and not be found here, but it is not necessarily invalid PSyIR.

sergisiso (changes requested):
@arporter The implementation looks good now, all limitations I can think off are caught in the validate and all is comprehensibly tested. All lines are covered and the documentations builds. There are no pycodestyle or pylint issues that need to be fixed.

There are just some final suggestions regarding docstrings and possible TODOs and bring it to master, then this can be merged.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -230,6 +229,17 @@ def __str__(self):
             precision_str = str(self.precision)
         return f"Scalar<{self.intrinsic.name}, {precision_str}>"
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
arporter: Good point. This will be fixed by #1799. I've added a TODO and warning in the dev. guide.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,478 @@
...
+        '''
+        Checks that the supplied node is a valid target for inlining.
+
+        Routines with a return value (AKA functions in Fortran) are currently
+        rejected: TODO #924.
+
+        :param node: target PSyIR node.
+        :type node: subclass of :py:class:`psyclone.psyir.nodes.Routine`
+        :param options: a dictionary with options for transformations.
+        :type options: dict of str:values or None
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -0,0 +1,478 @@
...
+        a(i) = 1.0
+        a(i) = 2.0 * a(i)
+      enddo
+    <BLANKLINE>
+    end subroutine run_it
+    <BLANKLINE>
+
+    The routine to be inlined must not contain any un-resolved symbols or
+    access any symbols that are defined in the parent container. Currently,
+    (#924) the routine must also be in the same source file as the call.
```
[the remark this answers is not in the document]
arporter: Good idea. Done and done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Oops. Good spot.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Oops - good spot.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Oh yes, I had to google that!

src/psyclone/tests/psyir/transformations/inline_trans_test.py
arporter: Oops, cut-n-paste strikes again.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Oh yes, I hadn't thought of that.

arporter:
Assuming CI happiness, this is ready for another look now.

sergisiso (approved):
@arporter All suggestions have been addressed and all test/formatting/documentation checks pass. There is a couple of docstring fixes needed but I will do them before merging. The PR is approved

PR #1830, opened 2022-08-09
(closes #1828) Fixes to enable explicit OpenACC Loop parallelism with UM

### LINKED ISSUE #1828: OpenACC Loop parallelism with managed memory
I tried using the NEMO OpenMP offloading script to insert OpenACC directive instead of OpenMP. It worked after fixing 2 issues:

- The ACCLoopDirective does not have a collapse attribute setter. It has all the infrastructure but currently it can only be set with the constructor parameter, but in NEMO the directive is first created and then the collpase number computed. We need a setter for this attribute.

- The ACCParallelDirective currently checks that any enter-data directive comes before this directive in the global constraints. @arporter I assume this is a constraint from a particular implementation of OpenACC we did in the pass but there is no reason to prevent the Parallel directive to be used with managed memory?

sergisiso:


sergisiso:
This is ready for review. Probably one for @arporter

src/psyclone/tests/psyir/nodes/acc_directives_test.py
```diff
@@ -61,7 +61,7 @@
 def setup():
     '''Make sure that all tests here use a new Config instance.'''
     Config._instance = None
-    yield()
+    yield
```
hiker, line 64: Out of curiosity - what is the difference? We have 22 `yield()`s and 7 `yield`s. Nothing wrong, I am just curious :)

hiker (changes requested):
The patch is nice and small. Besides my curious questions, there are a few pylint errors to be taken care of (and obviously they might not be caused by you ;) ).
Once this is done and the branch is up to master, I am happy to merge.
```
************* Module gocean1p0_transformations_test

domain/gocean/transformations/gocean1p0_transformations_test.py:1163:36: W0613: Unused argument 'fortran_writer' (unused-argument)
domain/gocean/transformations/gocean1p0_transformations_test.py:54:0: W0611: Unused MoveTrans imported from psyclone.transformations (unused-import)
domain/gocean/transformations/gocean1p0_transformations_test.py:54:0: W0611: Unused ACCDataTrans imported from psyclone.transformations (unused-import)
```
and
```
************* Module psyclone.psyir.nodes.acc_directives
psyir/nodes/acc_directives.py:86:8: R1725: Consider using Python 3 style super() without arguments (super-with-arguments)
psyir/nodes/acc_directives.py:343:4: R0913: Too many arguments (6/5) (too-many-arguments)
psyir/nodes/acc_directives.py:622:4: R0201: Method could be a function (no-self-use)
psyir/nodes/acc_directives.py:53:0: W0611: Unused Routine imported from psyclone.psyir.nodes.routine (unused-import)
```

src/psyclone/tests/psyir/nodes/acc_directives_test.py
sergisiso: Since I updated to python3.10 and pycodestyle 2.9 I was getting:
`src/psyclone/tests/psyir/nodes/acc_directives_test.py:64:10: E275 missing whitespace after keyword`

yield is like a return so I think `yield()` looked like a function but actually is returning a empty tuple without the whitespace between the keyword and the tuple, which I don't think is what we meant.

I updated all of them now.

hiker: Thanks! I used this as a reason to update my pylint, and get the same now, thanks (I didn't even wanted you to fix them all ... but this is certainly appreciated!!).

hiker:
I've took the liberty to disable two more pylint warnings and bringing this up to current master. AS soon as CI is finished, I'll merge it.

hiker:
I believe the failed coverage is caused by removing covered lines (--> fraction of uncovered lines slightly increases), all changes here appear to be fine. I'll merge this now.

PR #1838, opened 2022-08-12
(close #1704) replace_with takes into account named arguments

### LINKED ISSUE #1704: Rethink what happens with named arguments when we call replace_with in an argument
For example in Nemo dynzdf.f90:
```
CALL prt_ctl( tab3d_1=ua, clinfo1=' zdf  - Ua: ', mask1=umask,               &
    &         tab3d_2=va, clinfo2=       ' Va: ', mask2=vmask, clinfo3='dyn' )
```
has the node.argument_names:
[None, 'clinfo1', 'mask1', None, 'clinfo2', 'mask2', 'clinfo3']

sergisiso:


sergisiso:
This is ready for review. Probably for @rupertford . I took your suggestion from the issue discussion of modifying repalce_with but since its the parent type that matters I have not specialize it but used a `hasattr(self.parent,` instead.

sergisiso:
@rupertford I made an extra commit to fix a small bug that I found with the named_args. Now I ran it with the full Nemo codebase. It is still ready for review.

doc/user_guide/psyir.rst
```diff
@@ -726,9 +726,11 @@ methods.
 
 If an argument is inserted directly (via the children list) then it is
 assumed that this is not a named argument. If the top node of an
-argument is replaced then it is assumed that this argument is no
-longer a named argument. If arguments are re-ordered then the names
-follow the re-ordering.
+argument is replaced by removing and inserting a new node then it is
+assumed that this argument is no longer a named argument. If it is
+replaced with the `replace_with` method, it has a `keep_name_in_context`
+argument to chose the desired behaviour (defaults to True).
```
rupertford, line 732: chose -> choose

src/psyclone/psyir/nodes/call.py
```diff
@@ -233,7 +232,7 @@ def replace_named_arg(self, existing_name, arg):
                 f"in 'insert_named_arg' in the 'Call' node was not found "
                 f"in the existing arguments.")
         self.children[index] = arg
-        self._argument_names[index] = (id(arg), named_arg[1])
+        self._argument_names[index] = (id(arg), existing_name)
```
rupertford, line 235: This method no longer does what it says. It used to  replace one named argument with another. Now it replaces a named argument with an argument and keeps the original name irrespective of whether the replacing argument is named or not. Could you please reflect this change in the description. Is it also an error if the new argument is a named argument as we just throw the name away????

src/psyclone/psyir/nodes/node.py
```diff
@@ -1313,16 +1313,19 @@ def scope(self):
...
+    def replace_with(self, node, keep_name_in_context=True):
         '''Removes self, and its descendants, from the PSyIR tree to which it
         is connected, and replaces it with the supplied node (and its
         descendants).
 
         :param node: the node that will replace self in the PSyIR \
             tree.
         :type node: :py:class:`psyclone.psyir.nodes.node`
+        :param bool keep_name_in_context: whether to conserve the name \
+            referencing this node when it is within a parent with named nodes.
```
rupertford, line 1325: I'm not sure it makes sense to mention parent here as that is an implementation? Perhaps just remove (when it is ... named nodes.)

rupertford (changes requested):
Looks good Sergi. Pleasingly small. I've made a few comments inline. I'm guessing the documentation should be updated somewhere to describe what is going on, possibly in the psyir section of the user or developer guide? Is it also worth adding to the existing psyir examples to demonstrate named args if we don't already do so? Perhaps that is another PR.

doc/user_guide/psyir.rst
sergisiso: Fixed

src/psyclone/psyir/nodes/node.py
sergisiso: Done

src/psyclone/psyir/nodes/call.py
sergisiso: I don't think that's what it did. We just provided a new node also before, never a new name. I just changed name_arg[1] to existing_name because they already matched in the for loop and named_arg was producing a pylint error for being used outside the loop were it is defined.

> Is it also an error if the new argument is a named argument as we just throw the name away????

No, that's the id of the node that is being replaced.

That said, I modified the docstring to make this clear and also updated the equivalent Operation node.

sergisiso:
@rupertford This is ready for review again. I added documentation and examples as suggested. Also I think there was some confusion about the replace_named_arg method, I updated that docstring as well in both Call and Operation nodes.

rupertford (approved):
Requested changed have been made. All looks good. Approving.

PR #1846, opened 2022-08-22
Fix assorted docstrings

nmnobre:
In addition to these small typos I found while revising #1554, I also noticed that the [example](https://github.com/stfc/PSyclone/blob/ba35a86a1e13f1f266d2ba4e637621d1ae39939c/src/psyclone/transformations.py#L2606-#L2632) we give for ACCEnterDataTrans wouldn't actually add the directive by lack of a kernels/parallel construct in the schedule. Is this what we intended?


arporter:
> I also noticed that the [example](https://github.com/stfc/PSyclone/blob/ba35a86a1e13f1f266d2ba4e637621d1ae39939c/src/psyclone/transformations.py#L2606-#L2632) we give for ACCEnterDataTrans wouldn't actually add the directive by lack of a kernels/parallel construct in the schedule. Is this what we intended?

Oops, no! Would you mind fixing it in this PR?

src/psyclone/transformations.py
```diff
@@ -2836,17 +2837,17 @@ class ACCKernelsTrans(RegionTrans):
 
     For example:
 
-    >>> from psyclone.parse import parse
+    >>> from psyclone.parse.algorithm import parse
```
arporter, line 2840: Since we're here, it's worth doing this properly. Could you have a quick look and see whether you can get doctest to pass for this docstring (`python -m doctest transformations.py`). If it gets complicated we'll leave it as it's not really your problem.

src/psyclone/transformations.py
```diff
@@ -2950,17 +2951,17 @@ class ACCDataTrans(RegionTrans):
...
```
arporter, line 2954: Similarly, would you mind trying doctest on this docstring too?

src/psyclone/transformations.py
```diff
@@ -2950,17 +2951,17 @@ class ACCDataTrans(RegionTrans):
...
     >>> psy = PSyFactory(api).create(invokeInfo)
     >>>
     >>> from psyclone.transformations import ACCDataTrans
     >>> dtrans = ACCDataTrans()
     >>>
-    >>> schedule = psy.invokes.get('invoke_0').schedule
+    >>> schedule = psy.invokes.get('tra_adv').schedule
     >>> # Uncomment the following line to see a text view of the schedule
     >>> # print(schedule.view())
     >>> kernels = schedule.children[0].children[0].children[0:-1]
```
arporter, line 2967: As discussed, would you mind fixing the docstring for ACCEnterDataTrans too?

arporter (changes requested):
Thanks for these fixes Nuno.
For extra bonus points I've asked that you take a quick look to see whether you can get the docstrings you've fixed to work with doctest. Don't spend too much time on it though - if it gets complicated then we can leave it for now.

src/psyclone/transformations.py
nmnobre: Done... I've also added a kernels construct which I then surround with the data construct. This is strictly not needed and PSyclone will happily add the data construct around host code but... what's the point? Also, I've noticed the scalar 'r' isn't explicitly copied in, is that on purpose?

src/psyclone/transformations.py
nmnobre: I've tried but I don't really have any experience with the GOcean API so I'm not sure what I did makes sense.

arporter: No, I don't *think* so - if there were no compute constructs being added then that was definitely an oversight!

nmnobre: I've just confirmed the directive is indeed empty, I'll attempt a fix by inserting a kernels region around both generated loops. To clear this mess, after you are done reviewing, please do let me rewrite history to avoid the spurious commits.

nmnobre: Done.

nmnobre:
Should we change `SOURCE_FILE` to, say, `GOCEAN_SOURCE_FILE` and start using, say, `NEMO_SOURCE_FILE` instead of `tra_adv.F90` directly?

arporter:
> Should we change `SOURCE_FILE` to, say, `GOCEAN_SOURCE_FILE` and start using, say, `NEMO_SOURCE_FILE` instead of `tra_adv.F90` directly?

Yes, that's a good idea.

nmnobre:
> > Should we change `SOURCE_FILE` to, say, `GOCEAN_SOURCE_FILE` and start using, say, `NEMO_SOURCE_FILE` instead of `tra_adv.F90` directly?
> 
> Yes, that's a good idea.

Done. This was a bit harder than I thought it'd be!

arporter (approved):
Dev guide builds fine, as does Ref. guide.
I've found #1220 and mentioned the problems found here there (so to speak).
I'll proceed to merge.

src/psyclone/transformations.py
arporter: Yes, we don't need to copy scalars. OpenACC does that automatically.

PR #1847, opened 2022-08-25
Make a subSignature a Signature and allow Signature.to_language w/o arguments

nmnobre:


nmnobre:
Hey @hiker,

Sorry for the activity after marking it ready for review.
I just took advantage of the time zone difference :-)
This really is ready for review now!

Cheers,
-Nuno

src/psyclone/core/signature.py
```diff
@@ -118,19 +121,26 @@ def to_language(self, component_indices, language_writer=None):
...
         :raises InternalError: if the number of components in this signature \
             is different from the number of indices in component_indices.
         '''
 
+        # Avoid circular import
+        # pylint: disable=import-outside-toplevel
+        from psyclone.core import ComponentIndices
+
+        # By default, if component_indices is None, we assume a scalar access:
+        if component_indices == None:
```
hiker, line 141: This should be `is None`.

src/psyclone/core/signature.py
```diff
@@ -100,14 +100,17 @@ def __len__(self):
 
     # ------------------------------------------------------------------------
     def __getitem__(self, indx):
-        return self._signature[indx]
+        if isinstance(indx, slice):
+            return Signature(self._signature[indx])
+        else:
```
hiker, line 105: Another pylint complain: unnecessary `else` (i.e. remove else, and 'outdent' the return statement

hiker (changes requested):
>And one additional issue: you modified `to_language` - do you need this function (here is where a ticket would have helped)? Because I checked, and I believe this function is not needed anymore - I used it in the past to make it easier for some error messages, but with the improvements in the dependency analysis, these messages have been removed now, and there is no more usage of them in PSYclone. But I can see that this function could be useful for error messages, and I am not sure if someone else might be using it in scripts for nemo etc that are not part of psyclone. I would recommend to open a ticket so we can check this in the next telco (atm too many people are on leave :) ), and remove it later. Just add a #TODO to this function.

I've linked this PR on a comment in #1554, following a suggestion from @arporter to switch to `Signature.to_language`. Since that might be hard to find, here's the offending code:
https://github.com/stfc/PSyclone/blob/792a8cebbc1f3ab9e056bc166a09a40428b01449/src/psyclone/psyir/nodes/acc_directives.py#L886

hiker:
Hi @nmnobre ,

> Sorry for the activity after marking it ready for review. I guess I took advantage of the time zone difference. :-P This really is ready for review now!

Close call that ... I noticed it last night and nearly decided to quickly do it ... then I remembered my resolution to stop working unnecessarily that late at night and decided to wait :)

src/psyclone/core/signature.py
nmnobre: Done.

src/psyclone/core/signature.py
nmnobre: Done.

nmnobre:
> Thanks a lot for this PR. In general, it would be great to have a ticket, so I can actually compare what you are doing with the goal, while now I could only guess :)
> 
> Unfortunately there are a few minor things, which are actually not your fault, but since you touched it (and I noticed) ... :)
> 
> * I had two pylint message, see additional comments for these. Additionally (I didn't comment on them individually) there are a few strings that need to be converted to f-strings.
Done.
> * Both files need to update the (c) year to be ...-2022
Done.
> * You should add your name as modifier to both files.
Done.
> 
> And one additional issue: you modified `to_language` - do you need this function (here is where a ticket would have helped)? Because I checked, and I believe this function is not needed anymore - I used it in the past to make it easier for some error messages, but with the improvements in the dependency analysis, these messages have been removed now, and there is no more usage of them in PSYclone. But I can see that this function could be useful for error messages, and I am not sure if someone else might be using it in scripts for nemo etc that are not part of psyclone. I would recommend to open a ticket so we can check this in the next telco (atm too many people are on leave :) ), and remove it later. Just add a #TODO to this function.

nmnobre:
>And one additional issue: you modified `to_language` - do you need this function (here is where a ticket would have helped)? Because I checked, and I believe this function is not needed anymore - I used it in the past to make it easier for some error messages, but with the improvements in the dependency analysis, these messages have been removed now, and there is no more usage of them in PSYclone. But I can see that this function could be useful for error messages, and I am not sure if someone else might be using it in scripts for nemo etc that are not part of psyclone. I would recommend to open a ticket so we can check this in the next telco (atm too many people are on leave :) ), and remove it later. Just add a #TODO to this function.

I've linked this PR on a comment in #1554, following a suggestion from @arporter to switch to `Signature.to_language`. Since that might be hard to find, here's the offending code:
https://github.com/stfc/PSyclone/blob/792a8cebbc1f3ab9e056bc166a09a40428b01449/src/psyclone/psyir/nodes/acc_directives.py#L886

src/psyclone/core/signature.py
```diff
@@ -185,7 +195,7 @@ def to_language(self, component_indices, language_writer=None):
 
     # ------------------------------------------------------------------------
     def __repr__(self):
-        return "Signature({0})".format("%".join(self._signature))
+        return f"Signature({str(self)})"
```
hiker, line 198: Good catch this!

hiker (approved):
I fixed one more pylint complain (inheritance from object), but otherwise all issues have been fixed. No need to update the user guide, I will merge one CI is finished.

Issue #1840, opened 2022-08-12
Dependency Analysis bug: _partition method raises IndexError


sergisiso:
The dependency analysis raises a IndexError exception while analysing the following loop:

```
do widx3 = 1, loop_stop_120, 1
  loop_stop_121 = SIZE(tmask, 2)
  do widx2 = 1, loop_stop_121, 1
    loop_stop_122 = SIZE(tmask, 1)
    do widx1 = 1, loop_stop_122, 1
      if (tmask(widx1,widx2,widx3) == 1._wp .AND. tsn(widx1,widx2,widx3,2) == 0._wp) then
        tsn(widx1,widx2,widx3,2) = -99._wp
        tmask(widx1,widx2,widx3) = 0._wp
        umask(widx1,widx2,widx3) = 0._wp
        vmask(widx1,widx2,widx3) = 0._wp
      end if
    enddo
  enddo
enddo
```

the backtrace is:
```
    self.validate(node, options=options)
      File "/home/ssiso/workspace/stfc/PSyclone/src/psyclone/psyir/transformations/parallel_loop_trans.py", line 146, in validate
    only_nested_loops=False):
      File "/home/ssiso/workspace/stfc/PSyclone/src/psyclone/psyir/tools/dependency_tools.py", line 793, in can_loop_be_parallelised
    var_info)
      File "/home/ssiso/workspace/stfc/PSyclone/src/psyclone/psyir/tools/dependency_tools.py", line 628, in _array_access_parallelisable
    other_access):
      File "/home/ssiso/workspace/stfc/PSyclone/src/psyclone/psyir/tools/dependency_tools.py", line 521, in _is_loop_carried_dependency
    loop_variables)
      File "/home/ssiso/workspace/stfc/PSyclone/src/psyclone/psyir/tools/dependency_tools.py", line 283, in _partition
    partition_infos.append((indices_1[i].union(indices_2[i]), [indx]))
    IndexError: list index out of range
```

hiker:
This is caused by the `SIZE(tmask,...)` function registering a read access to `tmask` (with no indices). Fixing #1750 will fix this issue properly 

The following patch fixes the problem if you need an urgent fix:

```
diff --git a/src/psyclone/psyir/tools/dependency_tools.py b/src/psyclone/psyir/tools/dependency_tools.py
index 31ece4a74..39de6feb2 100644
--- a/src/psyclone/psyir/tools/dependency_tools.py
[...]

Issue #1750, opened 2022-06-09
Add Options to Variable Usage Functions


hiker:
#1742 shows that in some cases for example the variables used in defining the loop boundaries should not be included in the list of 'read' variables (`size(some_array,1)` - does not really read `some_array`). In other cases, the loop boundaries are required (kernel extraction), and/or should even include functions (`do i = 2, self%vspace%get_last_dof_owned()`).

My suggestion: we add an optional option-dictionary, similar to what we do in some of the `apply` and `validate` functions, where the calling function can influence the behaviour of the variable usage collection. E.g.:
``` 
reference.reference_accesses(var_access_info, {"loop-bounds": None})
reference.reference_accesses(var_access_info, {"loop-bounds": "variables"})
reference.reference_accesses(var_access_info, {"loop-bounds": "include-functions"})
```
The first call would not include any variables used in a loop boundary specification, the second one would be the current behaviour (get variables, but not function calls), the third one would also include function calls.

sergisiso:
We may be talking about 2 different things here:
- The reference_accesses of the whole loop, which includes a read of the boundaries at the beginning of the execution.
- The single-iteration reference access, which is what we need for the dependency analysis and in Fortran does not involve a read of the boundaries.

Maybe for the dependency analysis it should be doing `loop_body.reference_accesses(var_access_info)` and for the kernel extraction `loop.reference_accesses(var_access_info)`?

hiker:
I have started to implement the following option:
```
    :param bool array_shape_accesses_are_read: if True (default), an access \
        to an array using `size`, `lbound`, or `ubound` are considered read \
        accesses  to the array. If this is set to False, these accesses will \
        not be reported as 'read' accesses.
...    def __init__(self, nodes=None, array_shape_accesses_are_read=True):
```
So basically, if you set this variable to false, then the functions `size`, `lbound` and `ubound` will not report the array as being read anymore. This should fix this problem (and still allow the kernel extraction to be aware of the usage in the loop bounds).

I did discover that in https://github.com/stfc/PSyclone/pull/829/commits/b973396ba716d4178ad7c456ce7df25e78ed7360 code was added to disable reporting for `lbound` and `ubound` in the Reference object. This was apparently required for openacc. I am trying to remove this code and rely ont his new feature instead (the old way would not report a read in all cases, e.g. `lbound(array, my_dim)`, i.e. when the dimension is a variable).

hiker:
[...]

PR #1855, opened 2022-09-05
1840 fix da crash array without indices

hiker:
While the original crash reported in #1840 is caused by reporting a read access for `x` in `size(X...`, which will be fixed in #1750 , there are still mixed expressions (arrays with and without index used: `a(i) = max(a)`) that would trigger a crash.
This small PR fixes the crash and will not allow parallelisation in this case.

### DIFF
--- a/changelog
+++ b/changelog
@@ -50,7 +50,7 @@
 	19) PR #1809. A few fixes for code style issues.
 
 	20) PR #1803 for #1797. Adds support for logical variables to the
- 	PSyData	API.
+	PSyData API.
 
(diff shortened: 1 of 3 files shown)


sergisiso (approved):
@hiker Nice and short bug fix PR. I tested it with the 3 cases I had in NEMO and it now doesn't crash, it just refuses to parallelise the loop (I will monitor if #1852 does introduce parallelisation to them). All formatting is good, the case is tested and no documentation is needed. This is ready to merge.

Issue #1868, opened 2022-09-12
add reference2arraytrans to PSyAD script


rupertford:
Now that reference2arraytrans and the dotproducttrans fix are both on master we can add reference2arraytrans to the PSyAD script.

rupertford:
Created branch [1868_add_reference2arraytrans](https://github.com/stfc/PSyclone/tree/1868_add_reference2arraytrans)

PR #1869, opened 2022-09-12
add reference2arraytrans to PSyAD (closes issue #1868)

rupertford:


rupertford:
This is ready for first review, probably from @arporter seeing as it is mostly about PSyAD. It also deals with the fact that we don't want arrays within the LBOUND and UBOUND operators to be indexed.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1214,6 +1215,12 @@ def binaryoperation_node(self, node):
             fort_oper = self.get_operator(node.operator)
             if self.is_intrinsic(fort_oper):
                 # This is a binary intrinsic function.
+                # pylint: disable=unidiomatic-typecheck
+                if fort_oper in ["LBOUND", "UBOUND"] and \
```
arporter, line 1219: I don't like the look of this. Why is it necessary (given that the transformation now forbids modifying the argument to such operators)? Also, what happens if the argument to the operator is a reference to an array member of a structure?

src/psyclone/psyir/transformations/reference2arrayrange_trans.py
```diff
@@ -146,6 +146,16 @@ def validate(self, node, options=None):
...
                 f"The supplied node should be a Reference to a symbol "
                 f"that is an array, but '{node.symbol.name}' is not.")
+        if isinstance(node.parent, BinaryOperation):
+            if node.ancestor(Loop):
+                # This is an array reference within an LBOUND or
+                # UBOUND binaryoperator.
+                if node.parent.operator in [
+                        BinaryOperation.Operator.LBOUND,
+                        BinaryOperation.Operator.UBOUND]:
+                    raise TransformationError(
```
arporter, line 156: Please update the :raises: section of the doc string. Also, `SIZE` is a query function too.

arporter: Possibly, we should also disallow the case where the Reference is an argument to a Call?

src/psyclone/tests/psyad/transformations/test_preprocess.py
```diff
@@ -64,6 +64,30 @@ def test_preprocess_no_change():
...
+        "program test\n"
+        "real :: a(10), b(10), c(10)\n"
+        "a = b * c\n"
+        "end program test\n")
+    expected = (
+        "program test\n"
+        "  real, dimension(10) :: a\n"
+        "  real, dimension(10) :: b\n"
+        "  real, dimension(10) :: c\n\n"
+        "  a(:) = b(:) * c(:)\n\n"
```
arporter, line 82: How come we don't end up with an explicit loop here?

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -1336,6 +1336,50 @@ def test_fw_binaryoperator_matmul(fortran_writer, tmpdir, fortran_reader):
...
+
+    '''
+    code = (
+        "program test\n"
+        "  real, dimension(10,10,10) :: a\n"
+        "  real, dimension(10,10,10) :: b\n"
+        "  real, dimension(10,10,10) :: c\n"
+        "  integer :: idx\n"
+        "  integer :: idx_1\n\n"
+        "  do idx = LBOUND(a(1,1,1), 3), UBOUND(a(:,:,:), 3), 1\n"
```
arporter, line 1352: What about e.g. `UBOUND(a(1:3), 1)` where `a` actually has an extent `> 3`? The `datatype` property that I'm adding in #1850 will allow this to be sorted out.

src/psyclone/tests/psyir/transformations/reference2arrayrange_trans_test.py
```diff
@@ -240,6 +240,52 @@ def test_validate_structure(fortran_reader):
...
+        "  do idx = LBOUND(a, 3), UBOUND(a, 3), 1\n"
+        "    do idx_1 = LBOUND(a, 1), UBOUND(a, 1), 1\n"
+        "      a(:,:,:) = b(:,:,:) * c(:,:,:)\n"
+        "    enddo\n"
+        "  enddo\n\n"
+        "end program test\n")
+    psyir = fortran_reader.psyir_from_source(code)
+    trans = Reference2ArrayRangeTrans()
+    for idx, reference in enumerate(psyir.walk(Reference)):
+        if idx < 4:
```
arporter, line 278: I think this could be more robust if you checked/walked for BinaryOperation instead.

arporter: Also, please include `SIZE` in the test.

arporter (changes requested):
I have a couple of concerns about the implementation - particularly the modification to the backend (which might not actually be needed?).
Surprisingly the documentation does not need updating (it talks about "array notation" being replaced by loops which covers this case too).

rupertford:
I agree with you about the backend. I was going to take it back but you beat me to it. I think it should be a constraint when creating LBOUND and UBOUND operators, if you agree

arporter:
> I agree with you about the backend. I was going to take it back but you beat me to it. I think it should be a constraint when creating LBOUND and UBOUND operators, if you agree

Unless I've misunderstood you then I don't agree :-) It's valid (albeit strange) to pass a sub-section of an array to one of those query functions.

rupertford:
Well that would make life easier as I would not have to do anything. I was avoiding the following ...

`LBOUND(A(:,10),1)` for example and assuming it should be `LBOUND(A,1)`, but you think that is OK?

rupertford:
Perhaps we should disallow it in the PSyIR though, as it is confusing.

arporter:
> `LBOUND(A(:,10),1)` for example and assuming it should be `LBOUND(A,1)`, but you think that is OK?

Yes, I think that's valid. When you say "disallow it in the PSyIR", what would we do if we encountered some Fortran that did this?

rupertford:
Remove any array indexing when translating to PSyIR.

src/psyclone/psyir/backend/fortran.py
rupertford: Removed changes to backend.

src/psyclone/tests/psyad/transformations/test_preprocess.py
rupertford: This test has been replaced with one that shows an explicit loop being generated.

src/psyclone/tests/psyir/backend/fortran_test.py
rupertford: This test has been removed. The transformation does not modify arrays within LBOUND and UBOUND operators.

src/psyclone/tests/psyir/transformations/reference2arrayrange_trans_test.py
rupertford: This test has been removed. A new test for SIZE has been added.

src/psyclone/psyir/transformations/reference2arrayrange_trans.py
rupertford: Updated raises and added a test for SIZE.

rupertford: I'm not sure about Calls (and functions) as I think `call x(a)` and `call x(a(:)` are the same, as are `dotproduct(a,b)` and `dotproduct(a(:),b(:))` so it does not matter, and perhaps we want to be able to do this in some circumstances?.

Further, this PR is meant to be about adding the existing transformation to PSyAD, not modifying the transformation, so it could be a new issue if it is decided that is is an issue. I've had to make some changes to the existing transformation as integrating it into psyad showed that the transformation should not modify arrays within LBOUND, UBOUND and potentially (as you point out) SIZE, but I'm not sure we should be doing more here.

I'll leave it to the reviewer to decide the best way forward,

rupertford: I added a test, but left it at that.

rupertford:
I think this is ready for next review from @arporter.

src/psyclone/psyir/transformations/reference2arrayrange_trans.py
arporter: That sounds fine - I think I was getting confused.

arporter (approved):
All good now.
Ref. guide builds OK with 16 warnings but they are all the usual suspects.
There's one minor pylint issue which I'll fix (un-needed import).

Issue #1870, opened 2022-09-13
Typo in link causing linkcheck failures


hiker:
Ooops, I had a typo in a link, which causes the current link failures.

hiker:
Fixed in #1871.

Issue #1813, opened 2022-07-26
[PSyData] Kernel Extraction: implement stand-alone version of wrapper library that does not depend on NetCDF


hiker:
For compilation tests (and when migrating to really new platforms) it is useful to have an extraction and reading library that does not depend on NetCDF.
- [x] Implement basic writing using Fortran binary files
- [x] Implement LFRic version (that might likely just be the same as the NetCDF files)
- [x] Implement GOcean version (that might likely just be the same as the NetCDF files)
- [x] Implement Reader library
- [ ] Updated READMEs
- [ ] Update user guide/dev
- [ ] Add compilation tests(?)
- [ ] Add tests to make sure writer and driver write variables in the same order.

hiker:
This ticket can be closed. Compilation can be tested in one of the examples, it is documented, and the order of variables is identical (since they are taken from the same list).

PR #1871, opened 2022-09-13
#1870 Fix incorrect link.

hiker:
Fixes a typo introduced in #1813


arporter (approved):
Thanks for fixing this.
All checks (including the link checking) are now green.
It's a one-line fix in the documentation and that builds and renders fine.
Will proceed to merge...

PR #1875, opened 2022-09-14
(Closes #1874) rm pytest-pep257 from setup.py

### LINKED ISSUE #1874: pytest-pep257 only supports Python <= 3.4
At the moment the `setup.py` file lists `pytest-pep257` as one of the dependencies in the `test` section. However, it seems that no one uses it and it does not work for (at least) 3.10.6-7.

I therefore propose to remove this dependency unless @TeranIvy or @sergisiso object. (Rupert and Joerg have already said they don't use it.)

arporter:
Removes pytest-pep257 from the `test` extra_dependencies.

### DIFF
--- a/changelog
+++ b/changelog
@@ -115,6 +115,8 @@
 
 	42) PR #1871 for #1870. Fixes typo in link in User Guide.
 
+	43) PR #1875 for #1874. Remove pytest-pep257 dependency from setup.py
+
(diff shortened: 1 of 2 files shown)


arporter:
A one-liner. Ready for review from anyone :-)

sergisiso (approved):
@arporter PR approved, this setup change was decided in the associated issue.

PR #1906, opened 2022-10-06
(Closes #1896) Fix bug for parameter arrays in HoistLocalArraysTrans

### LINKED ISSUE #1896: HoistLocalArraysTrans does not reject arrays which are parameters
Asif has found that the HoistLocalArraysTrans transformation fails to spot that a local array is in fact a parameter and thus generates incorrect code:

    subroutine my_sub()
        integer, dimension(3), parameter :: flags = (/1,2,3/)
        ...
    end subroutine my_sub

arporter:
This PR also extends the validation to check for any ACCRoutine directives as these *usually* imply that the routine will be run in parallel. An option is implemented to permit this check to be disabled.

### DIFF
--- a/changelog
+++ b/changelog
@@ -154,6 +154,9 @@
 	55) PR #1807 towards #1806. Adds PSyIR support for LFRic kernel
 	metadata.
 
+	56) PR #1906 for #1896. Update HoistLocalArrayTrans to avoid hoisting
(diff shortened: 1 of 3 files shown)


sergisiso (approved):
@arporter The PR looks good, the new functionality has been added the HoistLocalArrayTans. The formating, pylint, builing documentation, tests and coverage all looks good. This is ready to merge.

Issue #1366, opened 2021-08-04
[psyir] Support allocate/deallocate and random_number intrinsics


hiker:
At this stage, an allocate statement becomes a code block. For the kernel-extraction driver creation, direct support for creating allocate statements would simplify the code (which currently parses a Fortran string to create a codeblock with the allocate statement) is used.

Note that there is also https://github.com/stfc/fparser/issues/298 to support the `mold=` argument. It would simplify the code further if `mold` would be supported.

arporter:
This support would also be useful when hoisting local, automatic arrays (#1634). Currently the required `allocated()` and `allocate()` calls have to be put into CodeBlocks.

arporter:
I'm thinking of taking this one on as it's needed for #1726. We should also address https://github.com/stfc/fparser/issues/298.

In Fortran we have:

    allocate(my_array(x,y,z), [mold=src_array, status=ierr])

and in C we have:

    my_array = malloc(x*y*z)

and one checks for success by examining whether `my_array` is `null`. We could support the 'mold' concept since in C that would simply be a `sizeof(src_array)/sizeof(src_array[0]).`

Are there other aspects to consider @hiker, @sergisiso and @rupertford?

arporter:
Naming: Fortran has `allocate` and `deallocate`, C has `malloc` and `free` where "malloc" really stands for "memory allocation." I favour having nodes `Allocate` and `Deallocate`. Although Fortran includes support for specifying the type of the entity (for OO style code), that moves us away from the functionality offered by `malloc` so we could simply exclude that?

arporter:
Just thought: @LonelyCat124 might have an opinion on this too.

LonelyCat124:
Is the plan for these to be a special type of Codeblock node? We should probably prevent code with allocate/deallocate statements from being inside parallel regions unless we have a good reason to allow it? There are certainly algorithms where memory allocation inside parallel regions is a thing, but computing things like shared/private might be not be worth allowing for now.

`Allocate`/`Deallocate` nodes seems reasonable to me, my only concern may be if a C++ backend (e.g. Kokkos) or something is planned how we handle that, though mostly I think C++ avoids explicit allocation now.

arporter:
> Is the plan for these to be a special type of Codeblock node?

No, I was planning on them being full `Statement` nodes. We would have to update the various parallel directive classes/transformations to ensure that they are excluded from being inside.

[...]

PR #1907, opened 2022-10-07
(Towards #1366) add support for IntrinsicCall (allocate, deallocate...)

arporter:


arporter:
Getting there now. I need to add support for `mold` and then either think about `deallocate` or stop and get this ready for review.

arporter:
This PR will require #1911 to go onto master as it needs the 'mold' support in fparser.

arporter:
This is now done but is waiting on #1911 to update fparser (as it requires support for the 'mold' argument to allocate).

sergisiso:
@arporter I merged the fparser PR, this should be unblocked now.

arporter:
Ready for first review now. Probably one for @sergisiso or @rupertford.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,178 @@
...
+        named arguments *must* come after any required arguments.
+
+        :param intrinsic: the Intrinsic being called.
+        :type intrinsic: py:class:`psyclone.psyir.IntrinsicCall.Intrinsic`
+        :param arguments: the arguments to this routine, and/or \
+            2-tuples containing an argument name and the \
+            argument. Arguments are added as child nodes.
+        :type arguments: List[ \
+            Union[:py:class:``psyclone.psyir.nodes.DataNode``, \
+                  Tuple[str, :py:class:``psyclone.psyir.nodes.DataNode``]]]
```
sergisiso, line 97: Any reason for the double black-quote? We usually use the single-backquote for type expressions, but I don't know if that matters.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,178 @@
...
+
+        :returns: an instance of cls.
```
sergisiso, line 99: "of this class." ?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,178 @@
...
+            optional_arg_names = []
+
+        # Validate the supplied arguments.
+        reqd_args = cls._required_args[intrinsic.name]
+        opt_args = cls._optional_args[intrinsic.name]
+        last_named_arg = None
+        pos_arg_count = 0
+        for arg in arguments:
+            if isinstance(arg, tuple):
+                name = arg[0].lower()
```
sergisiso, line 133: Maybe a TypeError if `not isinstance(arg[0], str)` here?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1629,6 +1629,9 @@ def call_node(self, node):
             else:
                 result_list.append(self._visit(child))
         args = ", ".join(result_list)
+        if isinstance(node, IntrinsicCall):
+            # An IntrinsicCall doesn't have 'call'.
+            return f"{self._nindent}{node.routine.name}({args})\n"
```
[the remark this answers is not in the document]
sergisiso: By the way, this in not directly tested in the fortran backend test file (although it is indirectly tested in other parts of the code). Can you add some simple cases in the fortran backend test file please.

sergisiso (changes requested):
@arporter The implementation of IntrinsicCalls looks great to me and it helps clean up some ugly CodeBlock code. I just made some inline comments.

Maybe what is missing is a bit more documentation:
- In the user_guide/psyir.rst:393 we list the Symbol subclasses, can you add InstrinsicSymbol (and the DataTypeSymbol which is missing).
- In developer_guide/psyir.rst "Selected Node Descriptions" section, maybe next to the "Operation Nodes" subsection add new subsection about the IntrinsicCall.
Specially what is the distinction between Operators and IntrinsicCalls.
  - were they needed to re-use the named/optional argument functionality of Calls?
and do we plan to move other intrinsics that now are Operations in to IntrinsicCalls or not?

  - Since Operators are accepted by offloading and parallelisation transformations while Calls are not (or shouldn't be) unless we guarantee they don't have side-effects (by being calls to pure routines/intrinsics or in PSyKAl APIs by being KernelsCalls)  and  we don't have a way to mark pure routines/intrinsics maybe leaving by design the ones that are likely to be valid to parallelise or run in a GPU as Operators could be a decision.

    but maybe this just complicate things and we should be moving them all to Intrinsic calls and just add the pure field as a property that we set for each intrinsic, what are your thoughts on this?
Anyway the developer guide should clarify why there is both and what is the picture we are going for.

arporter:
Thanks for the review Sergi. I should have written the documentation first as that always helps me to get things straight in my head. I've now drafted some:

image

and the first thing that becomes apparent is that actually, 'random' could still be an Operator as it has no optional arguments. I think it is the presence of the optional arguments that makes it very difficult to represent something as an Operator (unless we have a different Operator for each valid combination of optional arguments). `MAXVAL` and `MINVAL` will also be good fit for IntrinsicCall I think.

arporter:
I can see that there is an argument for moving all intrinsics from Operation to IntrinsicCall - it would clean up the fact that we already have multiple Operation nodes for e.g. MAX because it can have a variable number of arguments.
As you say, the parallelisation issues will probably require us to have a `pure` or `elemental` property for `IntrinsicCall`. Do we need to have a telco to discuss this with @rupertford?

rupertford:
Happy to discuss but my preference is for intrinsics to all be in the same place. I don't like some intrinsics being operators and some being IntrinsicCall as it is confusing and could also complicate things if we want to reason about other languages. I guess there is an argument that we could boil down intrinsics to ones without optional arguments and then build new ones (or generate code) for more complicated cases, but that is not the road we've gone down and I think understanding Fortran intrinsics (and perhaps C intrinsics in the future) in PSyIR fits in with its main purpose (performance).

src/psyclone/psyir/backend/fortran.py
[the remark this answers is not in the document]
arporter: Ah, you are right about the lack of "call" being specific to allocate/deallocate. I've updated the code and added a test to those for the Fortran backend.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,178 @@
...
+    ArgDesc = namedtuple('ArgDesc', 'min_count max_count types')
+    #: List of ArgDesc objects describing the required arguments, indexed
+    #: by intrinsic name.
+    _required_args = {}
+    #: Dict of optional arguments, indexed by intrinsic name. Each optional
+    #: argument is described by an entry in the Dict.
+    _optional_args = {}
+    # The positional arguments to allocate must all be References (or
+    # ArrayReferences but they are a subclass of Reference).
+    _required_args[Intrinsic.ALLOCATE.name] = ArgDesc(1, None, Reference)
```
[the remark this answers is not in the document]
arporter: Oh yes, that's nicer and it works.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,178 @@
...
+    _optional_args[Intrinsic.ALLOCATE.name] = {"mold": Reference,
+                                               "stat": Reference}
+    _required_args[Intrinsic.DEALLOCATE.name] = ArgDesc(1, None, Reference)
+    _optional_args[Intrinsic.DEALLOCATE.name] = {"stat": Reference}
+    _required_args[Intrinsic.RANDOM.name] = ArgDesc(1, 1, Reference)
+    _optional_args[Intrinsic.RANDOM.name] = {}
+
+    @classmethod
+    def create(cls, intrinsic, arguments):
+        '''Create an instance of class cls given the type of intrinsic and a
```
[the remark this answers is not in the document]
arporter: Yes, that's better thanks.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: No idea why I did that  - gone now.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: Done.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: Yes, done.

arporter:
I've (somewhat unilaterally) decided that it would be best if all intrinsics became IntrinsicCalls and have created #1987 to do this. I've updated the documentation to say this.

arporter:

doc/developer_guide/psyir.rst
```diff
@@ -490,6 +490,42 @@ var2, var3)` would be represented by a
 The PSyIR supports the concept of named arguments for operation
 nodes, see the :ref:`named_arguments-label` section for more details.
 
+IntrinsicCall Nodes
```
sergisiso, line 493: Thanks for adding the documentation, there is a bit about Symbols that also needs to be updated:
in the user_guide/psyir.rst:393 we list the Symbol subclasses, can you add InstrinsicSymbol (and the DataTypeSymbol which is missing).

doc/developer_guide/psyir.rst
```diff
@@ -490,6 +490,42 @@ var2, var3)` would be represented by a
...
+(`stat`) may be used to store a return value. It's therefore not clear
+what an allocation 'operation' would assign to (an `Operation`
+must be a child of a `Statement` and therefore could not be included
+in a `Schedule` on its own). Similarly, Fortran's `MAXVAL` and
+`MINVAL` intrinsics have optional `dim` and `mask` arguments. In order
+to represent these using `Operation` nodes, we would one for each
+of the four possible forms of each intrinsic.
+
+Therefore, to support intrinsic 'operations' that have optional
+arguments, the PSyIR has the :ref_guide:`IntrinsicCall psyclone.psyir.nodes.html#psyclone.psyir.nodes.IntrinsicCall` Node. This single class
```
sergisiso, line 508: Is it possible to break this line a bit? Maybe before and after the link.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1640,6 +1640,12 @@ def call_node(self, node):
             else:
                 result_list.append(self._visit(child))
         args = ", ".join(result_list)
+        if isinstance(node, IntrinsicCall):
+            if node.routine.name in ["ALLOCATE", "DEALLOCATE"]:
+                # An allocate/deallocate doesn't have 'call'.
+                return f"{self._nindent}{node.routine.name}({args})\n"
+            if node.routine.name == "RANDOM":
+                return f"{self._nindent}call RANDOM_NUMBER({args})\n"
```
sergisiso, line 1648: I prefer if you rename the PSyIR RANDOM to RANDOM_NUMBER (to not confuse them with other fuctions, like the intel specific https://www.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/a-to-z-reference/q-to-r/rand-random.html) and then update the backend as:
```
else:
    return f"{self._nindent}call {node.routine.name}({args})\n"
```

sergisiso: Or maybe remove the second condition... the rest of the visitor will already do the right thing, its just ALLOCATE, DEALOCATE that are exceptions.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,183 @@
...
+        :rtype: :py:class:`psyclone.psyir.nodes.IntrinsicCall`
+
+        :raises TypeError: if any of the arguments are of the wrong type.
+        :raises ValueError: if any optional arguments have incorrect names \
+            or if a positional argument is listed *after* a named argument.
+        :raises ValueError: if the number of supplied arguments is not valid \
+            for the specified intrinsic.
+
+        '''
+        if not isinstance(intrinsic, cls.Intrinsic):
```
sergisiso, line 109: pylint says "Second argument of isinstance is not a type". Would `intrinsic not in cls.Intrinsic` work for Enums?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,183 @@
...
+    subroutine). This can be found as a standalone statement
+    or an expression.
+
+    '''
+    # Textual description of the node.
+    _children_valid_format = "[DataNode]*"
+    _text_name = "IntrinsicCall"
+    _colour = "cyan"
+    _min_arg_count = 0
+    _max_arg_count = 0
```
sergisiso, line 58: Should these 2 lines be deleted? They are not used anywhere and probably they shouldn't be class variables anyway.

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -226,27 +226,25 @@ def apply(self, node, options=None):
...
                     IfBlock.create(
                         allocated_expr.copy(),
-                        [freader.psyir_from_statement(
-                            f"deallocate({sym.name})",
-                            node.symbol_table)]))
+                        [IntrinsicCall.create(
+                            IntrinsicCall.Intrinsic.DEALLOCATE,
+                            [Reference(sym)])]))
             body.append(
-                freader.psyir_from_statement(f"allocate({fwriter(aref)})",
```
sergisiso, line 240: Also remove the `fwriter = FortranWriter()` at line 157 and amend the comment.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -0,0 +1,183 @@
...
+
+from collections import namedtuple
+from enum import Enum
+
+from psyclone.psyir.nodes.call import Call
+from psyclone.psyir.nodes.reference import Reference
+from psyclone.psyir.symbols import IntrinsicSymbol
+
+
+class IntrinsicCall(Call):
```
sergisiso, line 47: Right now the IntrinsicCall constructor expects a RoutineSymbol (because it is not implemented and it uses the super class one) which is wrong.

I would not reimplement the constructor because this will still need to skip the super class constructor but set up the _argument_names (coupling both implementations). Instead I would simply add a _symbol_type class parameter and update the `call.__init__` to check against this. (since you are here you an also replace parent by kwargs in that constructor).

And add some IntrinsicCall constructor tests.

sergisiso (changes requested):
@arporter Thanks for the clarification about where this is going. I agree on the direction of moving Operations to IntrinsicCalls. See a few more inline comments but this is almost there now.

doc/developer_guide/psyir.rst
arporter: Done.

doc/developer_guide/psyir.rst
arporter: Yes, done.

src/psyclone/psyir/backend/fortran.py
arporter: Yes, I've simplified.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: That only works if `intrinsic` is a member of an Enumeration type. If it's of the wrong type then we get a type error. I'm suspecting that this is a bug in pylint as I found https://stackoverflow.com/questions/38393897/how-to-check-if-a-variable-is-of-type-enum-in-python and examined the input to isinstance:

    (Pdb) type(cls.Intrinsic)
    <class 'enum.EnumMeta'>
    (Pdb) type(Enum)
    <class 'enum.EnumMeta'>

and if I replace `cls.Intrinsic` with `Enum` in the call to `isinstance` then the pylint error goes away.

arporter: I've fixed it by first checking that it is an `Enum` and then checking that it is `in cls.Intrinsic`. I've also renamed the `intrinsic` argument back to `routine` to keep pylint happy.

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
arporter: Oops, good spot.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: Good spot again. I've done as you suggested and then added an additional test.

sergisiso (approved):
@arporter All the comments have been addressed. Formatting and linter are good now. All new functionality is covered in tests. Documentation has bee updated and builds without new issues. This PR is ready to merge.

sergisiso:
It seems like this closes #1366 even though it is marked as "towards" in the PR title. I have marked as resolved the issue on the changelog and will close the issue after merging. @arporter feel free to let me know and reopen the issue if I am missing something.

PR #1924, opened 2022-10-19
(Closes #1923) rm broken reference to KernelTrans in Dev Guide

### LINKED ISSUE #1923: Errors when building Dev Guide.
#1899 merged KernelTrans into KernelModuleInlineTrans but it is still referenced in the Dev. Guide. This needs to be updated.

arporter:
Small change to fix the Developer Guide which was broken by #1899 (and I failed to spot when reviewing).


arporter:
One for @sergisiso to check that he's happy that what I've done reflects what he did in #1899.

sergisiso (changes requested):
@arporter Ah thanks I also didn't notice the error. Since we are here I just noticed that the reference guide also shows this error:
`Module psyclone.domain.common.transformations doesn't have a element RaiseCall2InvokeTrans`
because:
`../../src/psyclone/domain/common/transformations/__init__.py` misnames the class in the all list
Can you fix this please?

arporter:
Oh yes, good spot. Fixed now. I was also starting to see Doxygen warnings about the number of nodes in some of our graphs so I've increased the limit for that. I've also updated conf.py to specify English as the previous value of `None` was generating a warning.

sergisiso (approved):
@arporter My comment was addressed. Most Developers guide and Reference Guide warning/errors are gone now. This is ready to merge.

PR #1940, opened 2022-11-08
Switch to Python 3.11 for CI tests

nmnobre:
I've had to slightly tweak a test due to a new feature in Python 3.11, which, when reporting an error (a `NameError` here), now draws an arrow below the location the error is found at.


arporter (approved):
Changes all look fine. CI is happy.
Will proceed to merge.

PR #1945, opened 2022-11-09
Improved sir backend (closes 1467)

rupertford:
Includes changes to work correctly with the latest version of dawn.

Does not fix the owner computes problem (where there is a write to an array with a loop index offset).
Does not fix the problem with scalars outside of nested loops.


arporter:
With Rupert's passing this isn't going to make it on now.

Issue #1883, opened 2022-09-19
Update kern_call_arg_list.py to use PSyIR


hiker:
In order to create a proper extraction driver for LFRic, (#1483)  we need to have the full PSyIR expression for all actual (implicit) kernel arguments. ATM this class only returns strings.

This has likely implications on other objects using this class as well.

arporter:
#1918 is merged now. Are all of the outstanding things represented by other issues now @hiker?

hiker:
Well, this is by far not complete. I have another PR ready that I just called 1883-part2, after that I suggest to close this (and leave the rest for the full conversion of LFRic to use PSyIR).

Issue #1295, opened 2021-06-18
PR 1264 re-introduces #1171


hiker:
PR 1264 changes the following line in `psyir.py`:
```
 MODULES = [
-    Module("constants_mod", ["i_def", "r_def", "l_def"])]
+    Module(LFRicConstants().UTILITIES_MOD_MAP["constants"]["module"],
+           ["i_def", "r_def", "l_def"])]

```
in dfc27653ff6c312a38ae2946781686a280d5ae4b. This causes the `LFRicConstants` object to be executed on import, causing the failures from #1171 again.

You can trigger this bug by executing:
```
~/work/psyclone/src/psyclone/tests$ pytest ./dependency_test.py 
____________________________________________________________ ERROR collecting src/psyclone/tests/dependency_test.py ____________________________________________________________
dependency_test.py:48: in <module>
    from psyclone.domain.lfric import KernCallAccArgList, KernStubArgList
../domain/lfric/__init__.py:52: in <module>
    from psyclone.domain.lfric.kernel_interface import KernelInterface
../domain/lfric/kernel_interface.py:44: in <module>
    from psyclone.domain.lfric import psyir as lfric_psyir
../domain/lfric/psyir.py:58: in <module>
    Module(LFRicConstants().UTILITIES_MOD_MAP["constants"]["module"],

```

arporter:
I can't reproduce this @hiker. Is it still a problem for you?

hiker:
Yes, I can't run all/some tests if I don't set PSYCLONE_CONFIG (note that I have a `pip -e` installation, which I believe is slightly different from a normal installation):
```
joerg@Joerg-Surface:~/work/psyclone/src/psyclone/tests$ pytest
============================================================================== test session starts ===============================================================================
platform linux -- Python 3.8.10, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: /home/joerg/work/psyclone, configfile: setup.cfg
plugins: forked-1.3.0, xdist-2.3.0, cov-2.11.1, anyio-3.5.0
collected 2131 items / 83 errors / 2048 selected                                                                                                                                 

===================================================================================== ERRORS =====================================================================================
_______________________________________________________________ ERROR collecting src/psyclone/tests/config_test.py _______________________________________________________________
config_test.py:54: in <module>
    from psyclone.domain.lfric import LFRicConstants
../domain/lfric/__init__.py:52: in <module>
[...]

PR #1950, opened 2022-11-14
1926 fix config file not propery used

hiker:
Fised the LFRic issue that a custom config file is not used as expected.

### DIFF
--- a/src/psyclone/generator.py
+++ b/src/psyclone/generator.py
@@ -98,6 +98,7 @@ def handle_script(script_name, info, function_name, is_optional=False):
         script function is called.
 
     '''
+    # pylint: disable=too-many-locals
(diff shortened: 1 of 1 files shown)


hiker:
This patch triggers a test failure (not sure why I didn't get this when testing ... maybe because I didn't run in parallel??)
The test `test_main_unexpected_fatal_error` in `generator_test.py` creates an unexpected error by patching `LFRicConstants()` with an invalid value. But since this patched object is now discarded, the error is not raised anymore, and the test fails.

We therefore need to find a better way to create this error - or properly solve #1295. Am setting this PR to invalid for now, since a proper fix for #1295 might be close once #1883 is merged.

hiker:
#2044 is the proper fix for this.

Issue #1955, opened 2022-11-16
Warnings after pylint upgrade


hiker:
I've recently upgraded to `pylint 2.15.5`. When running it in tests, I now get may startup warnings about  useless options in our config file:
```
.pylintrc:1:0: R0022: Useless option value for '--enable', 'import-star-module-level' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'oct-method' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'unpacking-in-except' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'parameter-unpacking' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'backtick' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'old-raise-syntax' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'dict-view-method' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'dict-iter-method' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'metaclass-assignment' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'next-method-called' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--enable', 'raising-string' was removed from pylint, see https://github.com/PyCQA/pylint/pull/4942. (useless-option-value)
[...]

PR #1956, opened 2022-11-17
#1955 Remove all now unsupported options from .pylintrc.

hiker:
I removed all options from the .pylintrc file that I was getting warnings for. I did not notice any bad results when running pylint on the sources.

Not sure if we should consider creating a new  .pylintrc file, the new format seems to be quite different. I didn't do it for now since we would have to all switch to the new format I'd guess.


arporter:
This looks fine to me and works OK too. Before I pull the trigger please could @TeranIvy check that these changes work as expected in the MO environment?

TeranIvy:
Ran `pylint`  2.12.2 on test files and mostly got the usual suspects of f-string formatting, imports outside of top level, unused imports and TODOs. 

I also got `(no-self-use)` and `(too-many-statements)` on some files, but that was a minority. All in all, the change seems fine.

arporter (approved):
Iva has confirmed she is happy with this.
Tests and examples are all still OK after merging with master (as they should be).
This is ready to go.

Issue #1731, opened 2022-05-26
Refactor Loop node


sergisiso:
The current Loop has the following attributes/methods
- children/create/semantic_navigation/eq
  These are the only things the core Loop should have, the rest should be moved to a new PSyLoop class.
  - [ ] Change 1 - Separate core PSyIR loop from PSy-layer Loop

- loop_type / valid_loop_types
  
  	gocean: inner/outer - encoded in GOceanConstants
  	lfric: dof/colours/colour/<empty_string>/null - Encoded in LFRicConstants
  	nemo: lon/lat/levels/tracers/unknown - from configuration file mapping-*
        (the variable name and loops bounds depend on it an all APIs, avoid inconsistencies )

  	- [ ] Change 2: rename gocean to lat and lon - this was agreed before, inner and outer are confusing when we swap the loops order or add tiling
  	- [ ] Change 3: remove the attribute, make the getter property computed and setter modify variable_name and upper/lower bounds. Remove LFRic NOT_INITIALIZED bounds.
  	- [ ] Change 4: Standarise <empty_string>/unknown/null - return "unknown" in property getter and invalid input in property setter)

  
- kernel 
  	Duplicated information: self._kern and self.walk(Kern)[0]

  	- [ ] Change 5: Remove the attribute and use computed property instead. What happens when fusing? Should it be a list of kernels or a kernel which is the sum of both? 
  	
- field/field_space/field_name
   There is a possible 4-way duplication: the self._attribute themselves, the self.field.attribute , the PSyIR tree loop bounds and the kern.arguments.
   - It is never that bad because in Gocean self.field is None, in LFRic the three bounds are NOT_INITIALISED until codegen, and in NEMO the 3 attributes are None
   - In LFRic  there are no setters (*well they are inherited without input checking! but they are not used
    - There is also invoke.fields which is of LFRicFields class and keeps track of all fields in an Invoke.
 
  	In NEMO this is not used, the 3 of them are always None (@arporter is the concept applicable to NEMO? even if we dont use it)
  
  - [ ] Change 6: Decide what is the source of truth and make the rest computed. I would say the source of truth is initially the kern.arguments provided to the constructor, during the constructor this is brought to the tree representation and then the metadata dropped, then the properties are computed looking at the tree and the setters modify the tree accordingly.
  
- iteration_space (iterates_over)
  In Gocean its initialized from kern.iterates_over and there is a setter that keeps tree consistent.
[...]

PR #1964, opened 2022-11-22
1961 lower to language for dynloop

### LINKED ISSUE #1961: Problems with driver creation - Index number of Loop can change / Fortran writer does not copy tags
The original problem was discovered during #1883. Originally, when calling `start_expr`, I added the corresponding Reference to the start variable, which is based on the loop index in the tree (e.g. `loop0_start`), to the children list. Next time `start_expr` was called, it would just return the value from the children list.

The problem with this approach is that if transformations are applied later, the index changes, and a newly introduced loop might have the same index (and therefore same variable name for start/stop) as the original loop had initially (the problem seen in #1883 is a compilation test after colouring is applied). As a short-term fix, I've removed adding the nodes to the children list.

Now, when creating the driver, I noticed that I get incorrect start/stop symbols. When the Fortran backend visits the routine, it creates a copy of the symbol table, but this copy does NOT copy the tags. The DynLoop is still in the tree, so when the DynLoop is visited to be converted to a string, it will use the symbol table without tags to find the start/stop symbols ... and will create new symbols since it uses the tags to look it up (and the copied symbol table does not have that tag, so a new symbol is created).

I don't know what's the proper way to fix this. I see a few options:
1. Maybe during `lower_to_language` the DynLoop should be converted to a Loop, meaning the start/stop expr will be evaluated once. This might work assuming that at this stage we still have a symbol table with tags. I checked with gocean, and it appears that GOLoop remain in the tree, i.e. are not converted to Loops. So not sure what the intended way is here.
2. We create a unique number for each loop, which is fixed, even if the loop is moved in the tree. This avoids the problems of inconsistent indices. @arporter  pointed out that a solution for this might be part of #1731. I had a look and didn't see this.
3. We copy the symbol table with tags in FortranWriter's routine_node. This seems to solve the problem for now, but the follow-up problem would be what if we have the same tag in several SymbolTable? Not sure if this would actually work in the existing code, but tags would be a problem :(

~~While 3. seems to work, I then spotted an additional problem: when I query the start/stop expression while visiting the DynLoop node, the code (after #1883) will verify if the `loop0_start` symbol has the right type, but it hasn't: somehow during copy the precision of the integer is lost - I try to lookup an `i_def` integer, but the symbol in the symbol table is now `Scalar<INTEGER, UNDEFINED>`. I will investigate this in more detail, and if required raise an additional ticket.~~
Edit: This was a bug in the driver and can be ignored.

I have tested 3. and this works for my examples.

hiker:
This is a small PR that fixes #1961, I just thought to split this off so that the driver creation code does not affect anything else, and is smaller.

This fixes #1961, which is a work around for until #1731 is done, which likely makes this lower_to_language unnecessary,


hiker:
I admit I don't understand this coverage failure - the uncovered lines were actually part of #1883 , and were covered in that PR. I would hope that once #1883 is on master, and this PR becomes as small as it should be, this error will just disappear.

hiker:
Lowering a DomainNode (which removed the DynLoop and replaces it with just the body) had the side effect that the handling of domain kernels in the FortranWriter was not tested anymore (since the FortranWriter lowered to language first, meaning the domain loop was removed, so the code that handles this is never executed).

Since we likely need this special handling of a domain kernel (once #1731 is done and we can remove the `lower_to_language_level` method from `DynLoop`, I didn't just remove the handling of domain kernels from the writer, instead I massaged the test case to trigger the appropriate line, and added 1731 so this can be simplified again once 1731 is done.

hiker:
Ready for review.

src/psyclone/tests/domain/lfric/lfric_loop_test.py
```diff
@@ -208,6 +215,30 @@ def test_mesh_name_intergrid():
...
+    sched.children[2].lower_to_language_level()
+    loop = sched.children[2]
+    assert not isinstance(loop, DynLoop)
+    assert isinstance(loop, PSyLoop)
+
+    _, invoke = get_invoke("25.0_domain.f90", TEST_API, idx=0)
+    sched = invoke.schedule
+    # Make sure we have the right node:
+    assert isinstance(sched[0], DynLoop)
+    # This call removes the loop and replaces it with the actual kernel
```
arporter, line 236: Please extend this test to check that the loop body is copied over correctly - I think that at the moment the statements will get reversed.

src/psyclone/dynamo0p3.py
```diff
@@ -7114,6 +7114,36 @@ def __init__(self, loop_type="", **kwargs):
...
+                                      self.step_expr.copy(),
+                                      self.loop_body.pop_all_children())
+            self.replace_with(psy_loop)
+        else:
+            # Domain loop, i.e. no need for a loop at all. Remove the loop
+            # node (self), and insert its children directly
+            pos = self.position
+            parent = self.parent
+            self.detach()
+            for child in self.loop_body.pop_all_children():
```
arporter, line 7144: I think you need to reverse this list in order to preserve the order of the body statements.

src/psyclone/tests/domain/lfric/lfric_loop_test.py
```diff
@@ -208,6 +215,30 @@ def test_mesh_name_intergrid():
     assert loops[0]._mesh_name == "mesh_field1"
 
 
+def test_lower_to_language():
+    ''' Tests that we can call lower_to_language_level a DynLoop.
```
arporter, line 219: "...level on a DynLoop."

src/psyclone/dynamo0p3.py
```diff
@@ -7114,6 +7114,36 @@ def __init__(self, loop_type="", **kwargs):
         self._upper_bound_name = None
         self._upper_bound_halo_depth = None
 
+    def lower_to_language_level(self):
```
arporter, line 7117: I can't quite remember the details of what this is fixing. We should probably have a test that shows what was wrong before you implemented this method? I've had a read of the associated issue and see that it was related to calling `gen` twice. Perhaps we need a test where we do this and check the output doesn't change?

arporter (changes requested):
Thanks for this Joerg.
Unless I'm missing something, I think we need a new test to demonstrate that the issue tackled by having the `lower_to_language` is now fixed. I also think there may be a bug in the way the loop body is copied over to the new loop - see inline.
No docs need updating.
Ref. guide is fine.

src/psyclone/tests/domain/lfric/lfric_loop_test.py
hiker: Very good catch. I've added a test that indeed showed that there was a bug, and then fixed the bug.

src/psyclone/tests/domain/lfric/lfric_loop_test.py
hiker: Done.

hiker:
Ready for next review. I hope the explanations help!

src/psyclone/tests/test_files/dynamo0p3/25.1_kern_two_domain.f90
```diff
@@ -0,0 +1,52 @@
...
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author A. R. Porter, STFC Daresbury Lab
+
+program single_invoke
+
+  ! Description: invoke of multiple kernels, one of which operates on the
```
arporter, line 38: s/one of which operates/all of which operate/

src/psyclone/dynamo0p3.py
```diff
@@ -7141,7 +7141,10 @@ def lower_to_language_level(self):
             pos = self.position
             parent = self.parent
             self.detach()
-            for child in self.loop_body.pop_all_children():
+            all_children_reverse = self.loop_body.pop_all_children()[::-1]
+            # Attach the children starting with the last, which
+            # preserves the origina order of the children.
```
arporter, line 7146: "original"

src/psyclone/dynamo0p3.py
```diff
@@ -7141,7 +7141,10 @@ def lower_to_language_level(self):
             pos = self.position
             parent = self.parent
             self.detach()
-            for child in self.loop_body.pop_all_children():
+            all_children_reverse = self.loop_body.pop_all_children()[::-1]
```
arporter, line 7144: I've never seen this before. More pythonic to use "...children().reverse()" perhaps?

arporter (changes requested):
Great - looks good now thanks Joerg.
Tests and examples are still all OK with compilation.
Just a few minor things that I couldn't let go and then this can be merged.

src/psyclone/tests/test_files/dynamo0p3/25.1_kern_two_domain.f90
hiker: Done (I've also updated the year and added my name as 'modified by')

src/psyclone/dynamo0p3.py
hiker: Done

src/psyclone/dynamo0p3.py
hiker: You would think so, wouldn't you? I did, too 😁  But that doesn't work: the result of `reverse` is `None`, so `all_children_reverse` becomes `None`, not the reverse of the list. I actually tried that first - the only way I could get it work was like this:
```
all_children_reverse = self.loop_body.pop_all_children()
all_children_reverse.reverse()
```
And I just hated the look of that: we assign the non-reverse list to a variable called `all_children_reverse`.

I did some googling, and found `reversed`, so I've changed it to:
```
all_children_reverse = reversed(self.loop_body.pop_all_children())
```
I hope that's an acceptable compromise.

hiker:
> Just a few minor things that I couldn't let go and then this can be merged.

Good points, all of them :)  But ready for review now.

src/psyclone/dynamo0p3.py
hiker: (And I finally changed the default setting of my editor to always enable spell checking - I hope that helps at least with the embarrassing typos)

src/psyclone/dynamo0p3.py
arporter: Yep, that's great.

arporter (approved):
Tests, examples and ref. guide still fine.
Proceeding to merge...

Issue #1698, opened 2022-04-25
Should SymbolTables have an equality operator?


sergisiso:
Currently any ScopingNode equality checks that `self.symbol_table == other.symbol_table` but since the SymbolTable doesn't have an `__eq__` method, it just checks that they are the exact same instance. For example two subroutines
```
Routine[name=my_sub]
    Return[]
Routine[name=my_sub]
    Return[]
```
won't be pick up as equal because they often have their own symbol table.

This also happens for loop_body/if_body, ... making the equality limited for subtrees of more than single statements.

arporter:
I would say that two symbols are equal if they:

 * have the same name;
 * have the same type (including shape if an array type);
 * have the same visibility;
 * are or are not both marked as arguments;
 
I agree that doing this check inside the symbol table sounds like a good way to go. The `argument_list` is itself a list of `Symbol`s so we would reuse the test that two symbols are equal for that.

sergisiso:
In #1899 I needed to compare routines (therefore schedules) and I implemented an initial version of the symbol_table equality operator (which is not complete). To finish it we need the full symbol/datatype/interfaces equality semantics.

Also, we need to consider what references to a symbol mean to be equal, e.g.:
```
subroutine a()
    real :: b
    b = b + 1
end subroutine
subroutine a()
    integer :: b
    b = b + 1
end subroutine
```
The equality of comparing both subroutines must return False (because their declaration is different).

But does the equality between the addition statements should be True or False?

LonelyCat124:
My vague feeling on this is that the addition statements should be False, but I could be persuaded either way. 

I think if the addition statements and symbol tables are equal, then the entire subroutines should be equal (as they are identical) since for code output there is no difference in any programming language. So as I see it the options are either:
1. Symbols are never equal to any other object
2. Symbols with the same class, type, name are equal.

Then the subroutines are either equal or not just by recursion.

hiker:
I am a bit late to the party, and it appears that we are getting off the actual subject (symbol/symbol table equality - now we are talking about statement/subroutine equality)  ... but just in case here another border cases:
```
subroutine a(x)
   integer x
   x=x+1
end subroutine a
subroutine a(y)
   integer y
   y=y+1
end subroutine a
```
Are they the same? They would (short of debug information) create the same machine code :)

My feeling is that we are better of having functions with an explanatory name (if possible) for comparison (semantically_equal, string_equal, ...) - but it appears that this is a bit late now :)

[...]

Issue #744, opened 2020-04-08
[PSyIR] Comparing datatypes


rupertford:
The approach to comparing datatypes in #728 is not nice. In #728 there was a discussion about how to improve this. This issue is the result of that discussion.

rupertford:
In #728 there were three options discussed

1) perform equality tests in the DataType classes
2) perform equality tests in a DataNode subclass of the relevant nodes (e.g. Literal)
3) use eq ne in the DataType classes to see if they are the same or not.

The problem with 1) is that (as far as I can tell) the base class would need to know about all of its subclasses to do the testing.

The problem with 2) is that a) it requires the datatype to already be attached to the node and b) it does not provide the methods to the DataSymbol class as it is not a Node. One solution to b) would be for DataNode and DataSymbol to use multiple inheritance.

The problem with 3) is that it only supports exact matching of datatypes and we might want to know if the datatype names (e.g. INTEGER) but not care about the precision. However this could be used in conjunction with the other options.

rupertford:
If we go for option 2 then the code could look something like the following ...

```
    def is_scalar(self, name=None, precision=None):
        '''Return True if the node is scalar and False otherwise. The optional
        name argument additionally checks whether the type-name
        matches and the optional precision argument additionally
        checks whether the precision matches.

        :param name: optional argument used to check whether this \
            instance has the same type name as the argument.
        :type name: :py:class:`psyclone.psyir.symbols.ScalarType.Name`

        :param precision: optional argument used to check whether this \
            instance has the same precision as the argument.
        :type name: :py:class:`psyclone.psyir.symbols.ScalarType.Precision`, \
            int or :py:class:`psyclone.psyir.symbols.DataSymbol`

        '''
        name_match = True
        if name:
            name_match = self._datatype.name == name
        precision_match = True
        if precision:
            precision_match = self._datatype.precision == precision

        return (name_match and precision_match and
                isinstance(self._datatype, ScalarType))
```

sergisiso:
> The problem with 1) is that (as far as I can tell) the base class would need to know about all of its subclasses to do the testing.

I don't see this. Consider the implementation below:
```
class DeferredType:
	def __eq__(self, other):
		return isinstance(other, DeferredType)

class ScalarType:
	def __eq__(self, other)
		return isinstance(other, ScalarType) and self.name == other.name \
                           and self.precision == other.precision

class ArrayType:
	def __eq__(self, other)
[...]

PR #1970, opened 2022-11-29
1966 enforce datatype of structure references

hiker:
Another small PR to help with the driver creation.


hiker:
I also added some tests and minor fixes for mixed precision handling, to make sure we get the right precision symbol when we enforce a user-defined type.

One change I did should be explained: in infer_argument I changed the type of the `r_def` etc defined from `INTEGER_SINGLE_TYPE` to `INTEGER_TYPE`. This is firstly consistent with the setting in `domain/lfric/psyir`:
```
    # Create the variables specified by the module (using PSyIR DataSymbols)
    for module_var in module.vars:
        exec(f"{module_var.upper()} = DataSymbol('{module_var}', INTEGER_TYPE,"
             f" interface=ImportInterface({MODULE_NAME.upper()}))")
```
and secondly (as far as I can tell) consistent with the specification of the `iso_fortran_env` module (see https://www.nag.com/nagware/np/r71_doc/iso_fortran_env.html):
```
INTEGER,PARAMETER :: real32 = SELECTED_REAL_KIND(6)
```
Note that we cannot use `domain.lfric.psyir.R_DEF` etc here, since then the corresponding module is not added to the correct symbol table.

I originally tried to compare the actual precision symbols in the tests, e.g. instead of:
```
    assert create_arg_list.psyir_arglist[3].datatype.precision.name == "r_def"
```
I tried:
```
    assert create_arg_list.psyir_arglist[3].datatype.precision == psyir.R_DEF
```
which feels a bit better to me :)  But that test doesn't work. I initially blamed the `INTEGER_SINGLE_TYPE` for the difference, but even after I fixed it (see above) and the two definitions appear to be the same, the test still failed:
```
  print(create_arg_list.psyir_arglist[2].datatype.precision, psyir.R_DEF,
          create_arg_list.psyir_arglist[2].datatype.precision == psyir.R_DEF)
r_def: DataSymbol<Scalar<INTEGER, UNDEFINED>, Import(container='constants_mod')>
r_def: DataSymbol<Scalar<INTEGER, UNDEFINED>, Import(container='constants_mod')> False
```
Not sure if this test is supposed to fail (since they are certainly different objects), or if the fact that the containers are likely different (objects) correctly triggers this to fail, or if we should have a working comparison method that would work in this case?? I decided not to touch this, and if required by the reviewer open a ticket.

hiker:
As additional explanation/summary: if the user defined type declarations are not found (which seems to be the normal case :grin:), we end up with 'deferred' declaration, and (in the normal case) no idea what type a user defined member access is. Since the KernCallArgList knows the type, I've added (in discussion with Sergi) a way to 'overwrite' the normal deferred type handling.

This allows me to define all the user defined access that are created for a kernel call, then the driver creation can pick up the right declarations without additional assumption, it is all contained in KernCallArgList, no need to change anything in the driver if a type should change.

sergisiso:
>print(create_arg_list.psyir_arglist[2].datatype.precision, psyir.R_DEF,
>          create_arg_list.psyir_arglist[2].datatype.precision == psyir.R_DEF)
>r_def: DataSymbol<Scalar<INTEGER, UNDEFINED>, Import(container='constants_mod')>
>r_def: DataSymbol<Scalar<INTEGER, UNDEFINED>, Import(container='constants_mod')> False
>
> Not sure if this test is supposed to fail (since they are certainly different objects), or if the fact that the containers are likely
> different (objects) correctly triggers this to fail, or if we should have a working comparison method that would work in this
> case?? I decided not to touch this, and if required by the reviewer open a ticket.

It shouldn't fail but we never decided the equality rules in https://github.com/stfc/PSyclone/issues/744 (should it be exact match or follow Fortran implicit casting rules?). How it is implemented will also affect https://github.com/stfc/PSyclone/issues/1698

src/psyclone/psyir/nodes/structure_reference.py
```diff
@@ -199,6 +220,8 @@ def _create(cls, symbol, symbol_type, members, parent=None):
             child_member = subref
         # Finally, add this chain to the top-level reference
         ref.addchild(child_member)
+        # pylint: disable=attribute-defined-outside-init
```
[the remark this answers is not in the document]
sergisiso: It may still complain about accessing a private member, this is OK, I prefer to keep this private for now, you can disable this one.

src/psyclone/psyir/nodes/structure_reference.py
```diff
@@ -241,8 +264,9 @@ def get_signature_and_indices(self):
     @property
     def datatype(self):
         '''
-        Walks down the list of members making up this reference to determine
-        the type that it refers to.
+        If no data type was enforced for this reference, walk down the list
```
sergisiso, line 267: To make it less visible maybe leave this as it was and add the addendum after the first sentence instead.

"If a overwrite datatype was given to this reference, this will return it instead of determining the type."

sergisiso (changes requested):
@hiker It looks good, I am mostly requesting some name changes and small fixes. It would be good to also add a couple of sentences to https://psyclone-dev.readthedocs.io/en/latest/psyir.html#references-to-structures-and-structure-members to explain what this option is and why was it needed.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -78,6 +78,19 @@ def __init__(self, kern):
...
+        self._map_fields_to_precision = {}
+        for field_info in const.DATA_TYPE_MAP.values():
+            field_type = field_info["type"]
+            # pylint: disable=no-member
+            if field_type in ["r_solver_field_type", "r_solver_operator_type"]:
+                self._map_fields_to_precision[field_type] = psyir.R_SOLVER
+            elif field_type == "r_tran_field_type":
+                self._map_fields_to_precision[field_type] = psyir.R_TRAN
+            else:
+                self._map_fields_to_precision[field_type] = psyir.R_DEF
```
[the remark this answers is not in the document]
hiker: Fair point ... done.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -143,27 +156,35 @@ def get_user_type(self, module_name, user_type, name, tag=None,
...
             user-defined type must be imported.
         :param str user_type: the name of the user-defined type.
+        :param member_list: the members used hierarchically.
+        :type member_list: List[str]
         :param str name: the name of the variable to be used in the Reference.
         :param Optional[str] tag: tag to use for the variable, defaults to \
             the name
-        :param shape: if specified, declare an array of user types
-        :type shape: List[:py:class:`psyclone.psyir.nodes.Node`]
+        :param enforce_datatype: the datatype for the reference, which will \
```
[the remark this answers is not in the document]
hiker: I agree :) Changed everywhere.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -143,27 +156,35 @@ def get_user_type(self, module_name, user_type, name, tag=None,
         return sym
 
     def append_user_type(self, module_name, user_type, member_list, name,
-                         tag=None):
+                         tag=None, enforce_datatype=None):
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -78,6 +78,19 @@ def __init__(self, kern):
         self._nqp_positions = []
         self._ndf_positions = []
 
+        # Create a mapping of the various fields type to the precision:
+        const = LFRicConstants()
+        self._map_fields_to_precision = {}
+        for field_info in const.DATA_TYPE_MAP.values():
+            field_type = field_info["type"]
+            # pylint: disable=no-member
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyir/nodes/structure_reference.py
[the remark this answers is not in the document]
hiker: I hope what I did was what you wanted.

src/psyclone/psyir/nodes/structure_reference.py
hiker: Done.

hiker:
> It shouldn't fail but we never decided the equality rules in #744 (should it be exact match or follow Fortran implicit casting rules?). How it is implemented will also affect #1698

I've added a TODO #744, so depending on how #744 is done, we can remove the string test and replace it with comparing the types.

hiker:
> @hiker It looks good, I am mostly requesting some name changes and small fixes. It would be good to also add a couple of sentences to https://psyclone-dev.readthedocs.io/en/latest/psyir.html#references-to-structures-and-structure-members to explain what this option is and why was it needed.

I actually found it hard to find a good place where to put this. After a while I settled with a section after the various 'member' sections,  but am happy if you find something better.

hiker:
Ready for next review.

src/psyclone/psyir/nodes/structure_reference.py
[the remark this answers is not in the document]
sergisiso: Almost, StructureReference should still have a positional "symbol" parameter which is just passed as a positional parameter to the super(). kwargs doesn't keep it (I think).

(it can also be done with an *args parameter but since it is just one maybe "symbol" end up with better documentation, but I don't mind which way it is implemented)

Also a constructor test in src/psyclone/tests/psyir/nodes/structure_reference_test.py would be good, even if the code is already covered indirectly.

doc/developer_guide/psyir.rst
```diff
@@ -744,6 +744,30 @@ subclass of ``Member``. Subsequent children represent the index expressions
 for the array access. The full API is given in the
 :ref_guide:`ArrayOfStructuresMember section of the reference guide psyclone.psyir.nodes.array_of_structures_member.html`.
 
+Data Type of a Structure Access
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+In order to get the actual data type of a structure reference, PSyclone
+needs to have access to the declaration of all structures involved
+in the access. In case of
```
sergisiso, line 751: I suggest adding:
"... in the accessor expression. However, these are often DeferredType if the module where they are declared has not been processed. In the case of some domain-API arguments added by PSyclone to a kernel call ..."

sergisiso: And update copyright authorship and date

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -194,6 +195,7 @@ def reference_accesses(self, var_accesses):
         var_accesses.next_location()
 
     def load(self, call, parent=None):
+        # pylint: disable=attribute-defined-outside-init, arguments-differ
```
sergisiso, line 198: I prefer not to disable those warnings. They are a reminder that this "load" pattern should be refactored.

src/psyclone/psyir/nodes/structure_reference.py
```diff
@@ -160,7 +188,7 @@ def _create(cls, symbol, symbol_type, members, parent=None):
                 f"symbol '{symbol.name}'")
 
         # Create the base reference to the symbol that is a structure
-        ref = cls(symbol, parent=parent)
+        ref = cls(symbol=symbol, parent=parent)
```
sergisiso, line 191: You could revert this named argument addition if you solve the positional argument issue I mentioned above.

(although I don't really mind how it is called here)

src/psyclone/tests/domain/lfric/kern_call_arg_list_test.py
```diff
@@ -106,6 +106,10 @@ def test_cellmap_intergrid(dist_mem, fortran_writer):
         'map_w2(:,cell)']
 
     check_psyir_results(create_arg_list, fortran_writer)
+    # pylint: disable=no-member
```
sergisiso, line 109: Again, you can bring this to the top since it happens multiple times in the file, and add a comment that this is because psyir classes are generated at runtime.

src/psyclone/tests/psyir/nodes/array_of_structures_reference_test.py
```diff
@@ -104,6 +106,22 @@ def test_asr_create(component_symbol):
...
+    # Test to enforce a type:
+    lbound = nodes.BinaryOperation.create(
+        nodes.BinaryOperation.Operator.LBOUND,
+        nodes.Reference(component_symbol), int_one.copy())
+    ubound = nodes.BinaryOperation.create(
+        nodes.BinaryOperation.Operator.UBOUND,
+        nodes.Reference(component_symbol), int_one.copy())
+    my_range = nodes.Range.create(lbound, ubound)
+    # pylint: disable=no-member
+    datatype = psyir.LfricRealScalarDataType()
```
sergisiso, line 119: Nitpicking but can you use a generic datatype for this test.

I would like the src/psyclone/test/psyir/* to be fully tested with src/psyclone/psyir/* instead of importing domain classes.

sergisiso (changes requested):
@hiker Almost there now, there is just minor things to address and a still open suggestion from the previous review.

src/psyclone/psyir/nodes/structure_reference.py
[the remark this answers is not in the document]
hiker: OK, I added 'symbol' as positional parameter back in.

doc/developer_guide/psyir.rst
hiker: That sounds better!

src/psyclone/domain/lfric/lfric_builtins.py
hiker: Fair enough, removed.

src/psyclone/psyir/nodes/structure_reference.py
hiker: Done.

src/psyclone/tests/domain/lfric/kern_call_arg_list_test.py
hiker: Done.

src/psyclone/tests/psyir/nodes/array_of_structures_reference_test.py
hiker: Good point, done.

src/psyclone/domain/lfric/lfric_builtins.py
hiker: (I just realised that I should remove both disable statements, not only the attribute-defined-outsite-init) ... done :)

src/psyclone/psyir/nodes/structure_reference.py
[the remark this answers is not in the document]
hiker: I also added a small test for the constructor.

sergisiso (approved):
@hiker All the comments have been addressed, this PR is ready to merge.

PR #1996, opened 2022-12-21
Addendum to #1952 to add missing tests

nmnobre:


arporter (approved):
Changes are pylint clean and the two missed lines are now covered.
Will proceed to merge.

Issue #888, opened 2020-09-09
-v command-line flag does not work unless a filename is supplied.


arporter:
Because we use `argparse.ArgumentParser` to handle command-line flags we automatically get an error if the user omits the (mandatory) filename argument. Unfortunately this means that simply running `psyclone -v` to get the version number fails.
I *think* what needs to be done is check for a solitary `-v` flag *before* we call `parse_args` in the `main()` routine in the generator.py file.

hiker:
I believe we discussed this when I added the -v flag 😄 I couldn't find an easy way of handling optional arguments (and still getting help at that time iirc). Googling now shows https://stackoverflow.com/questions/4480075/argparse-optional-positional-arguments - that would make the filename optional, but then we would have to check ourselves that filename is specified if no -v is given.

hiker:
Here it is: #175.

arporter:
Oh yes, thanks @hiker - I'd forgotten that!

arporter:
I've just found that `argparse` does know about 'version' (like 'help') and can take special action for it. See https://docs.python.org/3/library/argparse.html#action

Issue #175, opened 2018-06-04
Argument to display version string


MatthewHambley:
It's a small thing but it would help those of us who maintain development environments if PSyclone offered an argument which returned its version number. At the moment I can use ``which psyclone`` but that only works if it's been installed in a version specific location.

It might be worth making the version string available through an API in case PSyclone is ever used as a library by other programs.

Agreement must also be reached on the format of the string. I suggest the classic `<major>.<minor>.<release>[-<tag>]` with "tag" being "SNAPSHOT" at head of trunk and nothing for releases.

arporter:
PSyclone has the `-v` flag, introduced (by Joerg) with the 1.6.0 release. It prints the version number as specified in `psyclone.version` which is used by pypi and the documentation. Admittedly there's no API to access that information but it can easily be added to `version.py` should it ever be required.

I'll close this issue for now since there's no current need for an API.

MatthewHambley:
I am in the process of updating our PSyclone installation which is why this came up. I saw a `-v` option and tried it but got a "too few arguments" error so I assumed it was for something else.

rupertford:
Yes, the -v option is an optional addition so only outputs information if the required arguments are valid. This is not idea but it was easy to add in the first instance.

hiker:
That is an unfortunate side effect of the way the arguments are defined atm: psyclone must have one filename as parameter, without this argparse throws an error. For now you can use psyclone -v non_existing_file ... though this will print a warning that the file does not exist :(  I can quickly just abort psyclone if -v is specified, but changing the argument parsing to make sure there is a filename in most cases might be a bit more work :(

hiker:
Dang ... I should have checked if there already has been an answer first :)

PR #1997, opened 2023-01-03
Improve CLI --version interface

matthewrmshin:
Modify `psyclone --version` and `psyclone-kern --version` to behave like common commands, i.e. print version and exit, instead of continuing. (E.g., `git --version` would behave in this way.)

Before:
* `psyclone --version` does not work.
* `psyclone --version /fake/file` works but returns 1.
* `psyclone --version /path/to/real/filename` prints version and continue to process `/path/to/real/filename`.

After:
* `psyclone --version` works as expected.
* `psyclone --version /path/to/anyfile` prints version and exits, ignoring other opts/args.  <-- **Is this change OK?**

(I stumbled upon this problem while installing PSyclone in my local environment, and decided to have a stab at fixing it. I have since found two related issues after making this change #175 and #888. Please feel free to close down this PR if this is not the correct fix.)

### DIFF
--- a/src/psyclone/generator.py
+++ b/src/psyclone/generator.py
@@ -387,13 +387,11 @@ def main(args):
     parser.add_argument("--config", help="Config file with "
                         "PSyclone specific options.")
     parser.add_argument(
-        '-v', '--version', dest='version', action="store_true",
+        '--version', '-v', action='version',
+        version=f'PSyclone version: {__VERSION__}',
         help=f'Display version information ({__VERSION__})')
 
     args = parser.parse_args(args)
-
-    if args.version:
-        print(f"PSyclone version: {__VERSION__}")
 
     if args.profile:
         Profiler.set_options(args.profile)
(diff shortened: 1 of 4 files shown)


arporter:
Thanks very much @matthewrmshin - looks ideal to me. I'll do a proper review once the CI has run.

arporter (changes requested):
All looks good. Tests and examples are all fine still.
Ref. guide builds fine.
The only thing outstanding is the documentation - the 'getting going' (https://psyclone.readthedocs.io/en/stable/getting_going.html#run) and 'PSyclone command' (https://psyclone.readthedocs.io/en/stable/psyclone_command.html#running) sections of the User Guide need tweaking to reflect the slight changes in output. Are you OK to do that or would you like one of us to take it on?

matthewrmshin:
@arporter I have updated the user guide page you have indicated to reflect the output I now get. (There appears to be some unrelated changes to the help output though.)

matthewrmshin:
(CI failure unrelated. For some reason, the links to www.virtualbox.org are no longer working.)

arporter (approved):
Requested changes have been made (thanks).
Documentation builds fine and looks good.
Will proceed to merge.

PR #1976, opened 2022-12-05
1974 literals in lfric variable access info

hiker:
Fixes #1974. I tested for the `GH_INC` issue, and could not reproduce it, `GH_INC` and `GH__READINC` both seems to be reported correctly as output variables.
The other two issues (literals and array element access as variable names) are fixed.

PR #2015, opened 2023-01-18
#2012 Fix crash in access information collection when an LFRic domain loop is used.

### LINKED ISSUE #2012: NanTestTrans fails if applied to a 'null' loop
We get an error saying that the loop variable is `None`. Without looking at the code, I think this case should probably be supported so I'm marking this as a bug.

hiker:
This should fix #2012 . I looked at a few options of implementing this, but in `loop.py`, there is no `loop_type` or `iteration_space` defined that I could use (some tests failed because they did pass in a  plain `Loop`, not a `PSyLoop`). And I didn't want to test `isinstance(self, DynLoop)` or so in the base class `loop.py`. Only alternative would be to re-implement the `reference_access` in `DynLoop`, but this leads to some bad code duplication (the handling of the loop body would then need to be done in `DynLoop` and some other classes - PSyLoop or Loop or GOLoop/NemoLoop ... ). In short, I couldn't make up my mind about the best way, and this way should be safe, and has no code duplication. Let me know if you prefer a different implementation.


hiker:
Ready for review.

arporter (approved):
Thanks for doing this Joerg - I like the fix.
Ref. guide builds fine.
Will proceed to merge.

arporter:
This is just waiting on #1976 to go onto master.

PR #2029, opened 2023-01-31
(Closes #1567) correct visibility of imported symbols

### LINKED ISSUE #1567: [fparser2-FE] Visibility of symbols imported from other modules is lost
The various type definitions imported via USE statements in the following code:

    MODULE jb_def
     USE bge_ctl_def, ONLY :  fbge_ctl_typ             !: Control parameters for background error covariance
     USE bge_def, ONLY :  fbge_typ                 !: Background error covariance data structure
     USE ens_ctl_def, ONLY : fens_ctl_typ             !: Control parameters for ensemble estimation
     USE ens_typ, ONLY :  fens_typ                 !: Ensemble estimation data structure
     USE eof_ctl_def, ONLY :  feof_ctl_typ             !: Control parameters for EOFs
     USE eof_typ, ONLY :  feof_typ                 !: EOF data structure
     USE bal_ctl_def, ONLY :  fbal_ctl_typ             !: Balance operator control
     USE bal_typ, ONLY :  fbal_typ                 !: Balance operator data structure
     USE control_vectors, ONLY :  ctlvec
     USE variables, ONLY :  fvariables_typ            !: Control vector variables

     IMPLICIT NONE

     !! * Routine accessibility
     PRIVATE

     PUBLIC  fjb_typ                  !: Jb data structure

     PUBLIC  fbge_ctl_typ,  fbge_typ, fens_ctl_typ, fens_typ, fbal_ctl_typ, fbal_typ

     TYPE fjb_typ

        TYPE(fbge_ctl_typ)   ::  fbge_ctl             !: Control parameters for background error covariance
        TYPE(fbge_typ    )   ::  fbge                 !: Background error covariance data structure
        TYPE(fens_ctl_typ)   ::  fens_ctl             !: Control parameters for ensemble estimation
        TYPE(fens_typ    )   ::  fens                 !: Ensemble estimation data structure
        TYPE(feof_ctl_typ)   ::  feof_ctl             !: Control parameters for EOFs
        TYPE(feof_typ    )   ::  feof                 !: EOF data structure
        TYPE(fbal_ctl_typ)   ::  fbal_ctl             !: Balance operator control
        TYPE(fbal_typ)       ::  fbal                 !: Balance operator data structure
        TYPE(fvariables_typ) ::  fbkg_vars
        TYPE(fvariables_typ) :: fbge_vars

        TYPE(ctlvec), POINTER :: x_bkg               !: Background control vector

     END TYPE fjb_typ

    END MODULE jb_def

should be public but when we re-generate the code we lose this information. Possibly this is because there are two `PUBLIC` statements and we only spot the first?

arporter:


arporter:
Note to self: I need to test for the specific situation of `nv_ctl_opn` in `lib_mpp.F90` - this is an unresolved `Symbol` that is not referenced anywhere in the module apart from in an accessibility statement in the module declaration section.

arporter:
This is just waiting on Victoria now to see whether this branch fixes her issues with NEMOVAR.

arporter:
Victoria has now tested this branch with NEMOVAR and it fixes the issue.
Marking this as ready for review. Probably one for @sergisiso I think as it is all about handling Symbols in the PSyIR backend.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -744,63 +744,65 @@ def gen_access_stmt(self, symbol_table):
...
                 # dealing with the declarations in a given module which itself
                 # uses a public symbol from some other module, the
                 # accessibility of that symbol is determined by the
                 # accessibility statements in the current module.
                 if symbol.visibility == Symbol.Visibility.PUBLIC:
-                    public_routines.append(symbol.name)
+                    public_symbols.append(symbol.name)
                 elif symbol.visibility == Symbol.Visibility.PRIVATE:
-                    private_routines.append(symbol.name)
+                    private_symbols.append(symbol.name)
```
sergisiso, line 792: Would it be worth/safe skipping the statement it this is already the default?
```
if symbol.visibility == Symbol.Visibility.PUBLIC and not symbol_table.default_visibility in [None, Symbol.Visibility.PUBLIC]:
    return self._nindent + "public\n"
elif symbol.visibility == Symbol.Visibility.PRIVATE and symbol_table.default_visibility == Symbol.Visibility.PRIVATE: 
    private_symbols.append(symbol.name)
```
Just to keep it similar to the input code (although that not a primary requirement)

src/psyclone/tests/psyir/backend/fortran_gen_decls_test.py
```diff
@@ -253,14 +253,17 @@ def test_gen_decls_routine_unresolved(fortran_writer):
...
-    Container.
+    of a routine symbol is not an ImportInterface, unless there's a wildcard
+    import from a Container.
 
     '''
-    # Check a user-defined routine symbol with an (unsupported)
-    # ArgumentInterface
     symbol_table = SymbolTable()
+    # Check that a RoutineSymbol representing an intrinsic is OK
+    symbol_table.add(RoutineSymbol("nint", interface=UnresolvedInterface()))
```
sergisiso, line 262: This check passes because a nint is not yet an IntrinsicSymbol (used by an IntrinsicCall), but ultimately if we had a RoutineSymbol("nint") it would be something redeclaring the intrinsic (which is valid Fortran) and would need the declaration or import statement for it, right?
If this is the case we may want a TODO here (or better in the code that lets this happen). Do we have an issue to port all intrinsics to proper IntrinsicSymbols?

sergisiso (changes requested):
@arporter This is a bugfix. The issue has been solved and new test for it have been added. I looks good but I just have a couple of comments/questions about code that has been modified. The formatting and coverage are green.

src/psyclone/psyir/backend/fortran.py
arporter: I thought about doing that and then decided not to. However, since you raise it I've now done it as I think it's much nicer. Caused quite a few changes in the tests though.

src/psyclone/tests/psyir/backend/fortran_gen_decls_test.py
arporter: Actually, no. The comment was misleading - this is nothing to do with intrinsics. Since Fortran permits the resolution of symbols to be deferred until link time then all is fine as long as the routine is made available in some other file scope. (@rupertford changed the PSyIR backend to allow for this situation a couple of months ago.) I've therefore updated the comment but left the code unchanged.

arporter: In fact, the first part of this test is already covered by the previous one so I've updated this test to purely check for the exception that we raise if the interface to the RoutineSymbol is something unexpected.

arporter:

sergisiso (approved):
@arporter All my previous comments have been addressed or clarified. All tests and formatting checks pass. This PR is ready to merge.

Issue #1529, opened 2021-11-28
update PSyclone internal cut-down LFRic infrastructure to support mixed precision


rupertford:
The mixed precision example (lfric/eg19) won't compile as symbols such as `r_solver` and `r_solver_field_type` are not yet supported.

hiker:
I have fixed this as part of #1978 (though #1979 still remains, i.e. the fact that `r_solver` and `r_def` atm is still both 64 bit, so it's not real mixed precision).

hiker:
I am closing this ticket for now, since #2030 has just been merged. There is also #1980 (which might be a duplicate of this), and it does introduce the updated infrastructure files. I did reopen #1979 (which has been closed in between), since #1979 did NOT change the value of `r_solver` in `constants_mod.F90` to be 32 bit. I've added the required patch (2 lines) to #1979, and once #1980 is merged, we can quickly fix that.

Issue #1978, opened 2022-12-07
Update Infrastructure files and psydata to supply mixed precision


hiker:
Trying to extract data with gungho with mixed precision (r_def 64 bit, r_solver_def 32 bit) does not compile anymore, since there is no implementation for 32-bit fields in the psydata (which would affect all PSyData libraries).

Solution:
- [ ] Update the field mod in our copy of the infrastructure files to be up with latest LFRic. This will provide `field_r32_type` and `field_r64_type`, the current `field_type` and `field_solver_type` are just aliased to these types.
- [ ] Look into 32 and 64 bit integer as well
- [ ] Update all psydata LFRIc libraries to support these new types. This will likely involve:
- [ ] Use jinja for the lfric files (atm they are simple Fortran) to avoid code duplication

hiker:
I tried to update a larger fraction of the infrastructure files (than just scalar and field 32/64) ... and ended up in the usual dependency hell and gave up. So I am only adding the minimal file for the new field and scalar structures, and manually modified some other files to support 32/64 bit versions of certain functions as minimally required.

PR #2030, opened 2023-02-01
#1529 update eg19 Makefile so that compilation works

arporter:
A small PR that just updates eg19 so that it compiles. As noted in the issue, it's not true mixed precision as the various symbols actually boil down to the same type but I don't think that matters for our purposes.

### DIFF
--- a/changelog
+++ b/changelog
@@ -285,6 +285,9 @@
 	97) PR #2023 for #1975. Extends DynReferenceElement to ensure that
 	it adds related kernel arguments to KernCallArgList as PSyIR.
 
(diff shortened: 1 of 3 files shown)


arporter:
A small one ideally suited to @hiker I think :-)

hiker:
All good. I've took the liberty to add a .gitignore file in eg19 so that the files created by PSyclone are ignored.
No reference to 1529 was left.

hiker (approved):
I should have ticked the 'approve' checkbox :)

hiker:
Dang, dang, dang - I had this already done in #1978, which is still under review, and forgot all about it. Well, I guess it's up to me to fix the conflicts now :(

PR #2035, opened 2023-02-05
2033 split variables access info

hiker:
Gee, that was more involved than expected.


hiker:
Closing this in favour of #2036, which is based on master.

Issue #2058, opened 2023-02-20
Driver Creation for NEMO


hiker:
#2057 has shown some issues for the NEMO kernel creation:

- #1611  I believe I actually have this fixed for LFRic, so when doing #2049 that might get fixed.
- Apparently the (GOcean-based) driver creation gets confused with the data types?? It declares all real arrays as scalar reals (might be because GOcean only handles its own fields???). I wonder if the LFRic kernel extraction might fare better here. If so, #2049 might fix this as well.

Issue #1992, opened 2022-12-15
Kernel extraction should support MPI, and potentially several different files per kernel


hiker:
The restriction on not supporting distributed memory should be easy to lift, by just adding a rank number to the output files. The driver would then not read only one, but several files (one per process) and execute the kernel.
Once we have this, we might also want to create more than one file per kernel (per process). For example maybe the first and last time that the kernel is called, or store `X` data files somehow equally distributed across time, ...
Again, this would add a similar number to the output files to the rank, causing similar issues in the driver as outlines below.

- [ ] Enable distributed memory support for kernel extraction by adding a rank number to the output files.
    - [ ]  Make the driver either accept a rank number to read the right file, or automatically read the file for all ranks and execute the kernel. Also think of #1989 , how to present this information in a short summary.
- [ ] Enable that a kernel can dump files more than once, again by adding an index number (and similar to the MPI number let the driver handle this). We might use techniques from managing backup tapes to decide which files to delete to keep a kind of 'equidistant' set of files. Maybe an overall max size/kernel could be specified, or a number/kernel??

arporter:
Some of this has now been done (#2647) but the allowing several different files per extracted region is still outstanding.

PR #2059, opened 2023-02-20
2057 extraction for nemo

hiker:
This adds standalone and NetCDF extraction libraries for NEMO, plus a runnable example that extracts data.

Driver creation is done separately in #2058.


hiker:
The link check failures are expected, I have added links to the new library files, which do not yet exist. That should work once this is merged.

examples/nemo/README.md
```diff
@@ -133,3 +135,8 @@ data regions.
 SIR generation and transformation to CUDA using Dawn with simple
 examples and a cut down version of the tracer advection (tra_adv)
 benchmark.
+
+## Example 5
+
+Extraction of kernel data. Using the tra_adv benchmark, this examples
```
arporter, line 141: Ooh!

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2018-2021, Science and Technology Facilities Council.
```
arporter, line 4: 2023?

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
...
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# ------------------------------------------------------------------------------
+# Author: A. R. Porter, STFC Daresbury Lab
+# Modified J. Henrichs, Bureau of Meteorology
+
+# Makefile for the 5th NEMO example. Uses PSyclone (which must
+# be installed) to generate Fortran for the tracer-advection benchmark.
+# The script provided here instrument the kernels
```
arporter, line 39: "instruments each of the kernels"

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
...
+# export F90=gfortran
+# export F90FLAGS="-g -O0"
+
+
+include ../../common.mk
+
+TYPE ?= standalone
+
+GENERATED_FILES += psy.f90 psy.o traadv-$(TYPE).exe output.dat psyclone.log \
+  traldf_iso.F90
```
arporter, line 53: I don't think you need traldf_iso?

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
...
+
+F90FLAGS ?= -fopenmp
```
arporter, line 55: This example doesn't use OMP.

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
...
+
+# The compiler to use may be specified via the F90 environment variable,
+# it defaults to:
+# export F90=gfortran
+# export F90FLAGS="-g -O0"
+
+
+include ../../common.mk
+
+TYPE ?= standalone
```
arporter, line 50: Please add a comment here.

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
...
+F90FLAGS += -I$(EXTRACT_DIR) -I$(READ_DIR)
+LIB_NAME = lib_extract.a
+
+
+.PHONY: allclean
+
+compile: traadv-$(TYPE).exe
+
+run: compile
+	IT=2 JPI=10 JPJ=10 JPK=5  ./traadv.exe
```
arporter, line 81: Good idea having the env. vars set here.

examples/nemo/eg5/Makefile
```diff
@@ -0,0 +1,103 @@
...
+
+traadv-$(TYPE).exe: psy.o $(EXTRACT_DIR)/$(LIB_NAME)
+	$(F90) psy.o -o traadv-$(TYPE).exe $(EXTRACT_DIR)/$(LIB_NAME) $(LDFLAGS)
+
+transform: kernels
+
+# Need `-l all` to ensure line-lengths in generated code are less than the
+# standard-mandated 132 chars. Also kernel extraction does not support
```
arporter, line 89: Since this is NEMO you don't need the -nodm flag and associated comment.

examples/nemo/eg5/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+```sh
+psyclone -nodm -l all -api nemo -s ./extract_kernels.py ../code/tra_adv.F90
+```
+
+Executing this will output the transformed Fortran code with the 
+kernel extraction code added. Note that some of the lines in this
+Fortran code will exceed the 132-character limit. This may be remedied
+by supplying the `-l all` flag to PSyclone (as is done in the Makefile).
+Kernel extraction does not support distributed memory (TODO #1992), so
```
arporter, line 20: As in the Makefile, the NEMO API doesn't have explicit support DM so you don't need '-nodm'.

examples/nemo/eg5/Makefile
arporter: However, you need to update the name of the executable to include `${TYPE}` otherwise `TYPE=netcdf make run` doesn't work.

examples/nemo/eg5/domain_setup.sh
```diff
@@ -0,0 +1,42 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2019, Science and Technology Facilities Council.
```
arporter, line 4: -2023

examples/nemo/eg5/extract_kernels.py
```diff
@@ -0,0 +1,105 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: R. W. Ford, A. R. Porter, S. Siso and N. Nobre, STFC Daresbury Lab
+
+'''A transformation script that seeks to apply OpenACC DATA and KERNELS
```
arporter, line 36: This doc string needs updating :-)

examples/nemo/eg5/extract_kernels.py
```diff
@@ -0,0 +1,105 @@
...
+'''
+
+from psyclone.nemo import NemoLoop
+from psyclone.transformations import TransformationError
+from psyclone.psyir.transformations import ExtractTrans
+
+
+def trans(psy):
+    '''A PSyclone-script compliant transformation function. Applies
+    OpenACC 'kernels' and 'data' directives to NEMO code.
```
arporter, line 76: Please update.

examples/nemo/eg5/extract_kernels.py
```diff
@@ -0,0 +1,105 @@
...
+            continue
+
+        count = 0
+        for kern in sched.children:
+            count += 1
+            if not isinstance(kern, NemoLoop):
+                continue
+            try:
+                extract.apply(kern,
+                              options={"COLLECT-ARRAY-SHAPE-READS": False})
```
arporter, line 101: Please add a comment explaining what the "COLLECT-ARRAY-SHAPE-READS" option setting is for.

lib/extract/netcdf/extract_netcdf_base.jinja
```diff
@@ -218,7 +219,8 @@ contains
 
         ! Open the NetCDF file
         retval = CheckError(nf90_create(module_name//"-"//region_name//".nc", &
-                                        NF90_CLOBBER, this%ncid))
+                                        IOR(NF90_NETCDF4,NF90_CLOBBER),       &
```
arporter, line 222: What's the NF90_NETCDF4 for?

lib/extract/netcdf/nemo/kernel_data_netcdf.f90
```diff
@@ -0,0 +1,59 @@
...
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author J. Henrichs, Bureau of Meteorology
+
+!> This module implements a simple NetCDF writer using the PSyData
+!! interface. It is specific to the generic NEMO API, i.e. it does 
+!! actually not provide any API-specific types like fields, only standard
+!! Fortran data types. Therefore, it is just an empty wrapper to around
```
arporter, line 39: s/to around/around/

examples/nemo/README.md
arporter: "this example"

hiker: Done.

examples/nemo/eg5/Makefile
hiker: Indeed.

examples/nemo/eg5/Makefile
hiker: Done.

examples/nemo/eg5/Makefile
hiker: OOps ... and neither do I need psyclone.log.

examples/nemo/eg5/Makefile
hiker: Fixed.

examples/nemo/eg5/Makefile
hiker: Done.

examples/nemo/eg5/Makefile
hiker: Done.

examples/nemo/eg5/Makefile
hiker: Well, I actually do need it ... but I totally agree that this is stupid. So I've fixed this, the distributed memory flag is now only tested if the API is not NEMO. I've also added tests to see it works as expected with LFRic/GOcean, and a completely new test for nemo. I didn't add any more nemo specific tests for nemo - I expect that Nemo will always work with the base class (since it's pure PSyIR). At this stage I can't decide if we really want a new `nemo_extract_trans.py` file that just adds this one test (the test for DM is actually important at this stage for the extraction libraries, they do not support MPI). We would then have tests for distributed memory duplicated in any API-specific implementation, which is ugly as well. I suggest we leave it till the extraction libraries support MPI (TODO #1992), then it will just go away.

examples/nemo/eg5/README.md
hiker: Removed.

examples/nemo/eg5/domain_setup.sh
hiker: Done.

examples/nemo/eg5/extract_kernels.py
hiker: Oops - totally.

examples/nemo/eg5/extract_kernels.py
hiker: Oops again.

examples/nemo/eg5/extract_kernels.py
hiker: Done - but tbh, we might need to look at a better option here.  FWIW, reason is the expression:
```
zwx(:,:,jpk) = 0.e0
```
which (I assume) is lowered to an expression using `shape(zwx, ..)` on the RHS, which then gives a problem with a combined read/write access on the left hand side. Of course, jf we should have code lik:
```
something_else(shape(zwx, 1),jpk) = 0.e0
```
then we do need the read access (or we get smart and only store the result of `shape`. But that will require a lot of additional work in the driver later, which I wouldn't like ... )? I'd guess converting array expressions to explicit loops would also solve this issue.

lib/extract/netcdf/extract_netcdf_base.jinja
hiker: I've added a comment (we need netcdf 4 Format in order to use 64-bit integer in NF90_INT_64 or so).

lib/extract/netcdf/nemo/kernel_data_netcdf.f90
hiker: Done.

hiker:
As before, the link check errors are expected, since I've added new locations which do not yet exist.

examples/nemo/eg5/extract_kernels.py
```diff
@@ -91,12 +92,17 @@ def trans(psy):
...
 
-        count = 0
         for kern in sched.children:
-            count += 1
             if not isinstance(kern, NemoLoop):
                 continue
             try:
+                # The example contains array expressions, e.g.:
+                # zwx(:,:,jpk) = 0.e0
+                # PSyclone will convert them using the shape of zwx. But
```
arporter, line 101: "PSyclone represents this internally using `Range` with `LBOUND` and `UBOUND` intrinsics and currently this results in several..."

lib/extract/netcdf/extract_netcdf_base.jinja
arporter: Great, thanks.

arporter (changes requested):
Very nearly there now.
Just a couple of corrections needed.
Also, I realised the UG section on examples needs updating (file:///home/kbc59144/Projects/PSyclone/doc/user_guide/_build/html/examples.html#nemo).
Ref. guide builds fine.

examples/nemo/eg5/Makefile
hiker: Oops.

examples/nemo/eg5/extract_kernels.py
hiker: Done.

hiker:
I've updated #2080 to look at using `is_full_access` (I had a look, that should be simple, just a bit too much to sneak this in as part of this PR :) ), and then added a TODO for #2080 to the example, so we can update the comment.
Ready for next review.

arporter (approved):
UG has been updated and builds fine.
Examples are still all OK with compilation.
Will proceed to merge.

PR #2063, opened 2023-02-22
Improve handle_script() to avoid masking ImportError exceptions

nmnobre:


nmnobre:
After the adding the test, I actually feel we should only do:

```
try:
    transmod = __import__(filename)
except Exception as error:
    raise GenerationError(
        f"generator: attempted to import '{filename}' but a problem "
        f"was encountered: {error}") from error
```

src/psyclone/tests/generator_test.py
```diff
@@ -147,23 +145,6 @@ def test_script_file_wrong_extension():
             "extension" in str(error.value))
 
 
-def test_script_file_not_found_relative():
-    '''Checks that handle_script() in generator.py raises the expected
```
arporter, line 151: Ah, so we already supported searching the Python path for the script. I guess this is a consequence of the use of `__import__` but what happens if there is a file with the same name somewhere on the original Python path? Perhaps @rupertford can remember why we do things this way?

src/psyclone/tests/generator_test.py
```diff
@@ -175,11 +156,9 @@ def test_script_invalid_content():
...
             os.path.join(BASE_PATH, "dynamo0p3", "1_single_invoke.f90"),
             api="dynamo0.3",
-            script_name=os.path.join(
-                BASE_PATH, "dynamo0p3", "error.py"))
-    assert "attempted to import 'error' but script file " in str(error.value)
-    assert ("src/psyclone/tests/test_files/dynamo0p3/error.py' is "
-            "not valid python" in str(error.value))
+            script_name=os.path.join(BASE_PATH, "dynamo0p3", "error.py"))
+    assert ("attempted to import 'error' but a "
+            "problem was found: " in str(error.value))
```
arporter, line 161: I'd expect to see a new test here to cover one of the situations that used to get masked. e.g. if a transformation script imported a Python module that couldn't be found?

src/psyclone/generator.py
```diff
@@ -99,15 +99,14 @@ def handle_script(script_name, info, function_name, is_optional=False):
     '''
     sys_path_appended = False
     try:
-        # a script has been provided
+        # a script file has been provided, check it exists
+        if not os.path.isfile(script_name) and not any(
+               os.path.isfile(os.path.join(p, script_name)) for p in sys.path):
```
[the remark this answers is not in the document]
arporter: To answer my own question, I see that the code has always behaved this way (and I've either forgotten or never noticed). In which case I'm happy (although is there a utility that searches the `sys.path` so that we don't have to?).

arporter (changes requested):
OK, so now I've got over my confusion I'm happy with the changes.
All that's needed is at least one extra test to demonstrate that the poor behaviour we used to have has been fixed by this PR.

src/psyclone/generator.py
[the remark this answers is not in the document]
nmnobre: I've rewritten `handle_script()` so you no longer need the `./`.
`__import__()` reads the system path, so that's why we check it.
I'm not aware of any such utility.

src/psyclone/tests/generator_test.py
nmnobre: We now prepend instead of append.

src/psyclone/tests/generator_test.py
nmnobre: Done.

src/psyclone/generator.py
```diff
@@ -98,29 +98,35 @@ def handle_script(script_name, info, function_name, is_optional=False):
...
-            sys.path.append(filepath)
-        filename, fileext = os.path.splitext(filename)
+        module_name, fileext = os.path.splitext(filename)
+        # the file must either be:
+        # a) at the given path or, given no path, in the current directory; or
+        # b) given no path, in the system path
+        if not (os.path.isfile(script_name) or
+                not filepath and any (os.path.isfile(os.path.join(p, filename))
+                                      for p in sys.path)):
+                raise GenerationError(
```
arporter, line 111: pylint doesn't like the indentation here. (e.g. you need to indent the above two lines an extra 4 spaces).

src/psyclone/tests/generator_test.py
```diff
@@ -129,10 +129,9 @@ def test_script_file_no_extension():
 
 def test_script_file_wrong_extension():
     '''Checks that handle_script() in generator.py raises the excepted
```
arporter, line 131: Not yours but since we're here s/excepted/expected/

src/psyclone/generator.py
```diff
@@ -98,29 +98,35 @@ def handle_script(script_name, info, function_name, is_optional=False):
...
+        sys_path_prepended = True
+        sys.path.insert(0, filepath)
         try:
-            transmod = __import__(filename)
+            transmod = __import__(module_name)
         except Exception as error:
             raise GenerationError(
-                f"generator: attempted to import '{filename}' but a problem "
-                f"was found: {error}") from error
+                f"generator: attempted to import '{module_name}' but a "
```
arporter, line 128: Not yours but looking at the tests, I think a bit more context would be useful in this message, e.g.: "...attempted to import specified PSyclone optimisation script '{module_name}' but...." and again for the error messages below.

src/psyclone/tests/test_files/dynamo0p3/error_import.py
```diff
@@ -0,0 +1 @@
+import non_existent
```
arporter, line 1: Please add a comment here explaining the purpose of this file.

arporter (changes requested):
Looking good now, thanks for those improvements Nuno.
Just a few things to tidy up and please could you extend the appropriate bit of the User Guide (https://psyclone.readthedocs.io/en/stable/transformations.html#sec-transformations-script) to say that ${PWD} is included in the search path if no path is supplied to the script.
Tests and examples are all still OK with compilation.
Ref. guide still builds without new warnings.

src/psyclone/tests/generator_test.py
nmnobre: Done.

src/psyclone/generator.py
nmnobre: Done.

src/psyclone/generator.py
nmnobre: Done.

src/psyclone/tests/test_files/dynamo0p3/error_import.py
nmnobre: `error.py` now `error_syntax.py` never did either...

arporter: While that might be true, it doesn't mean it's good ;-) Please add an appropriate comment to that file too. (It only needs a single line.)

arporter (changes requested):
Just a couple of comments needed and then this can go on.

arporter (approved):
Tests and examples still OK.
Will proceed to merge.

PR #2079, opened 2023-03-20
Create PSyIR kernel arguments from new metadata classes (closes #2078)

### LINKED ISSUE #2078: Create LFRic PSyIR kernel arguments from new metadata
A new metadata to kernel arguments base class has been implemented in PR #2048 which uses the new metadata classes. This new base class should be subclassed to create LFRic PSyIR kernel arguments and should replace the existing `kern_stub_arg_list.py` and the partially implemented `kernel_interface.py`, both of which use the old style metadata, with the former using old-style string names and the latter using LFRic PSyIR.  Once fully implemented it can be used for kernel stub generation instead of `kernel_stub_arg_list.py`. Also see issue #928 which should be closed once this issue is addressed.

rupertford:
This PR creates a subclass of the new `metadata_to_argument_rules.py` base class being implemented in PR #2048. This new subclass will create LFRic PSyIR kernel arguments. This is similar to the existing partial implementation in `kernel_interface.py` but uses the new metadata and will be complete, so will replace this code. It will also replace the existing `kern_stub_arg_list.py` which uses the old style string arguments and the old style metadata and is used when generating kernel stub code. Once complete this new class can then be used to generate kernel stub code by replacing the existing implementation in  `gen_kernel_stub.py`.

### DIFF
--- a/src/psyclone/domain/lfric/formal_kernel_args_from_metadata.py
+++ b/src/psyclone/domain/lfric/formal_kernel_args_from_metadata.py
@@ -1,0 +1,952 @@
+''' xxx '''
+
+from psyclone.domain.lfric import LFRicTypes, KernelArgOrder
+from psyclone.psyir.symbols import ArgumentInterface
+from psyclone.psyir.symbols import SymbolTable
+from psyclone.psyir.symbols import ArrayType, DataSymbol
+from psyclone.psyir.nodes import Reference
+
+# TODO: many symbols here are not declared in LFRicTypes
+
+class FormalKernelArgsFromMetadata(KernelArgOrder):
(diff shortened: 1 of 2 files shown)


rupertford:
As mentioned above this PR depends on PR #2048 being on master before it can proceed.

arporter:
There's quite a long way to go to get the tests to pass flake8 - they're very much a WIP with variables assigned to but never used.

arporter:
Following what Rupert wrote in the description of this PR, I've modified gen_kernel_stub so that it uses the new functionality in this PR rather than the old, fparser1-based version. This should help me understand what this PR is doing.

PR #2084, opened 2023-03-23
(Closes #1801) support assumed-shape arrays with specified lower bound

### LINKED ISSUE #1801: psyir support for arrays declared with indexed bounds
We need to add support for arrays declared like a(0:n-1). At the moment they are stored as unknown fortran type, however this caused psyad to fail in poly2d_w3_reconstruction_kernel_mod.F90

arporter:


arporter:
Ready for review. One for @rupertford or @sergisiso.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -550,12 +550,13 @@ def _validate_data_node(dim_node):
...
-                    elif not isinstance(dim, int):
+                    elif not isinstance(dim, (int, ArrayType.Extent)):
                         raise TypeError(
                             f"A DataSymbol shape-list element specifying lower"
                             f" and upper bounds must be a 2-tuple containing "
-                            f"either int or DataNode entries but '{dimension}'"
-                            f" contains '{type(dim).__name__}'")
+                            f"either int, DataNode or ArrayType.Extent entries"
+                            f" but '{dimension}' contains "
+                            f"'{type(dim).__name__}'")
```
sergisiso, line 559: Could we also validate that the upper can not have a value if the lower doesn't? For instance now this is accepted:
` array_type = ArrayType(INTEGER_TYPE, [(ArrayType.Extent.ATTRIBUTE, 10)])`

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -1251,6 +1281,18 @@ def test_process_not_supported_declarations():
...
+    # this is invalid so we have to break the parse tree it produces for
+    # a valid case.
+    reader = FortranStringReader("integer :: l12(:)")
+    fparser2spec = Fortran2003.Specification_Part(reader).content[0]
+    # Break the parse tree
+    aspec = walk(fparser2spec, Fortran2003.Assumed_Shape_Spec)[0]
+    aspec.items = (None, Fortran2003.Int_Literal_Constant('4', None))
+    processor.process_declarations(fake_parent, [fparser2spec], [])
+    l11sym = fake_parent.symbol_table.lookup("l12")
+    assert isinstance(l11sym.datatype, UnknownFortranType)
```
sergisiso, line 1294: Is this equivalent to "integer :: l12(:4)" ? In this case gfortran does not accept it. Is it valid Fortran? Should it be a GenerationError instead.

sergisiso (changes requested):
@aporter The changes look good, I am just asking for a bit more testing that we don't permit the `dimension(:value)` case.

Also, if I understand correctly we were already supporting `dimension(value1:value2)` this just adds support for the case where the upper bound is assumed. So it should not affect transformations like inline or range2loop because the behavior is the same?

However, can you add some of these cases in src/psyclone/tests/psyir/transformations/inline_trans_test.py line ~520

and maybe to the two range2loop implementations that we have: Nemo and generic (even if it is just to check that it returns the appropriate error message)

The  issue 1801 also says:
> in WW3 code - it prevents the application of HoistLocalArrayTrans to quite a number of arrays.

Was this only because it end up being a UnknownFortranType or is there any difficulty with this transformations? Can you put an example similar to the ones that were causing trouble before in the HoistLocalArrayTrans test?

src/psyclone/psyir/symbols/datatypes.py
arporter: I thought about this and concluded that it's never valid for a lower limit to be of `ArrayType.Extent` since it can only be this if the upper limit is and in that case there should just be a single `ArrayType.Extent.xxx` rather than a tuple of values. I've simplified the checking code.

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: No, it's not valid Fortran and yes, that's a good idea - I've changed it to a GenerationError now.

arporter:
Found #2125 while extending the InlineTrans test. I resisted the urge to fix it here :-)

arporter:
I think the confusion arises because #1801 wasn't originally about assumed-size arrays but arrays with shapes specified using range notation. The HoistLocalArraysTrans already works with these now:
image

arporter:
> So it should not affect transformations like inline or range2loop because the behavior is the same?

That's correct. This is all about understanding the declaration. Those transformations always use `LBOUND` and `UBOUND` intrinsics to get the loop ranges (unless an explicit range is given in the Array accessor) so this PR has no effect. I have extended one test but, as expected, it makes no difference.

arporter:
Ready for another look now (assuming CI happiness).

arporter:
CI is not happy :-( Taking this back for now.

arporter:
Updated to cover the missed line. I also realised that `datatype_test.py` wasn't giving 100% coverage of `datatypes.py` so I've fixed that too. Really should be ready for review again now.

sergisiso (approved):
@arporter Thanks for adding the extra checks, I agree with the simplifications done in ArrayBound.

Linters, tests, documentation all pass (I took the liberty of improve one docstring)

This is ready to merge.

Issue #1618, opened 2022-02-10
Use PSyIR for the LFRic algorithm layer


rupertford:
The gocean1.0 API algorithm-layer transformation now uses PSyIR. The LFRic algorithm layer transformation should now be migrated to using PSyIR as well. This is expected to stress the PSyIR so may result in other issues.

rupertford:
Created branch [1618_lfric_algorithm_psyir](https://github.com/stfc/PSyclone/tree/1618_lfric_algorithm_psyir)

rupertford:
- [ ] Allow LFRic invoke calls to be unresolved
- [ ] Fix generated algorithm module and subroutine names
- [ ] Try with lfric code and fix any problems
- [ ] Update PSyclone tests and examples

rupertford:
Created branch to fix the problem of incorrect module and subroutine names being generated [1618_lfric_alg_psyir_names](https://github.com/stfc/PSyclone/tree/1618_lfric_alg_psyir_names)

rupertford:
Created branch to fix the problem of the `qr` argument being added at the wrong place [1618_lfric_alg_psyir_qr](https://github.com/stfc/PSyclone/tree/1618_lfric_alg_psyir_qr)

rupertford:
In the existing (pre-PSyIR) implementation that creates LFRic arguments, it is the job of the PSy-layer classes to generate them. This is first done in a generic way in the invoke base class (setting the `_alg_unique_args` list), ignoring any QR or stencil arguments and then the stencil instance extends this list with extent and direction args and finally the qr arguments are determined from the `QR_Rules` structure and added. This information is originally determined from the kernel metadata before being captured in the PSy-layer classes.

rupertford:
I don't think we should be interfacing to the old way of doing things so am thinking of writing something that replicates the QR and Stencil support, but in a more modular way, by reading the kernel metadata and then perhaps having Stencil and QR objects that can be queried (but I'm not sure yet).

arporter:
Currently we have `InvokeCallTrans` which converts the PSyIR representing `call invoke(...)` into Algorithm-specific PSyIR that captures that this is not just a standard subroutine call. We need another transformation that converts an `AlgorithmInvokeCall` back into a `Call` to the appropriate PSy-layer routine. (Currently this happens in the `lower_to_language_level` method but we've agreed that it shouldn't.)

arporter:
[...]

PR #2104, opened 2023-04-12
Example code should use a function (towards #1618)

rupertford:
The example code `examples/lfric/code/gw_mixed_schur_preconditioner_alg_mod.x90` is not valid as if you have an interface with the same name as a type then the associated routines must be functions and in the example the routine is a subroutine. For example, the following is invalid:

```
module test_mod
  type, public :: test_type
  end type
  interface test_type
     module procedure constructor
  end interface
contains
  subroutine constructor
  end subroutine
end module
```
```
rupert@xxx$ gfortran -c interface.f90
interface.f90:2:27:

    2 |   type, public :: test_type
      |                           1
Error: In generic interface 'test_type' at (1) procedures must be all FUNCTIONs as the generic name is also the name of a derived type
```

and the following is OK:

```
module test_mod
  type, public :: test_type
  end type
  interface test_type
     module procedure constructor
  end interface
contains
  function constructor() result(self)
  end function
end module
```


rupertford:
Ready for review. This is a very small change in example code. Probably best suited for @arporter or @TeranIvy

TeranIvy:
@rupertford, happy to take it if you're OK with me doing the built-ins PR first. How did you encounter the bug? When I run `make compile` with `gfortran` 6.1.0 in `examples/lfric` everything works fine!

rupertford:
> @rupertford, happy to take it if you're OK with me doing your built-ins PR first. How did you encounter the bug? When I run `make compile` with `gfortran` 6.1.0 in `examples/lfric` everything works fine!

Thanks @TeranIvy. Yes I'm happy for this to be done after the built-ins PR. `make compile` does not compile this code, presumably because it contains lfric infrastructure calls, which is why we don't see any errors. The code is used in egs 12, 13 and 15 I think. I encountered it through a circuitous route and it does not actually affect anything in PSyclone but it is better to have more correct code even if it is not compiled. The version of gfortran I used is 9.3.0.

TeranIvy (commented):
Can confirm that the GW preconditioner constructor on LFRic trunk is function and not subroutine.

Ran full battery of tests with compilation, LFRic examples with compilation and documentation building.

Updated changelog and user guide. CI permitting, proceeding to merge.

PR #2110, opened 2023-04-18
(Closes #2109) update to 0.1.0 of fparser

### LINKED ISSUE #2109: Move to fparser 0.1.1
fparser 0.1.0 has been released today. We need to move over to using this in GHA and check that LFRic works correctly with it.

arporter:


arporter:
Ready for review. One for @TeranIvy as this will need to be tested with LFRic.

TeranIvy:
> Given the issue found with fparser, I suggest we delete this PR and create a new one once 0.1.1 is out. Is that OK @TeranIvy?

Yes, @arporter, that's a good idea! Thank you :) I'll create the fparser issue now.

PR #2060, opened 2023-02-20
LFRic algorithm layer transformation using PSyIR (towards #1618)

rupertford:
Now that the new kernel metadata parsing is on master this can be used to make the LFRic algorithm layer transformation using PSyIR create the correct arguments when there are stencils and quadrature.

PR #2112, opened 2023-04-18
Split LFRic example algorithm file (towards #1618)

rupertford:
The upcoming PR #2060 includes a check that LFRic algorithm files only contain a single module or program (although this logic will be hidden behind a switch for the time being). The Met Office would like to have this check in LFRic so it will be kept when the new code goes live (see #1618). LFRic example 11 falls foul of this check as it has an algorithm file that includes a program as well as a module. This PR splits this file in two.


rupertford:
This is ready for review and is suitable for anyone.

arporter (approved):
Modified example compiles fine.
No other changes have been made.
Will proceed to merge.

PR #2115, opened 2023-04-19
(Closes #2114) ensure allocate with type-spec becomes a CodeBlock

### LINKED ISSUE #2114: [nemo] Bug in ALLOCATE handling when type-spec is present
Processing NEMO 4.2.xxx with master fails because:

    subroutine mpp_bcast_nml(cdnambuff, kleng)
      CHARACTER(LEN = :), ALLOCATABLE, INTENT(INOUT) :: cdnambuff
      ALLOCATE( CHARACTER(LEN=kleng) :: cdnambuff )

is reproduced as:

    ALLOCATE(cdnambuff)

which is wrong as it no longer has any extent information.

image

arporter:
Small bug fix. I've elected just to go for a CodeBlock for now rather than attempt to handle the type-specification - I don't know how we'd do that for an argument to a Routine. Perhaps it would need a DataType argument?


arporter:
A small bug fix for anyone in NEMO land so perhaps @rupertford or @sergisiso?

src/psyclone/tests/psyir/frontend/fparser2_alloc_handler_test.py
```diff
@@ -129,6 +129,25 @@ def test_alloc_with_mold(fortran_reader):
...
+program test_alloc
+  character(len=*)   :: oh_dear = "not good"
+  real, allocatable, dimension(:, :) :: var1
+  allocate(var1(5,5), errmsg=oh_dear)
+end program test_alloc
+'''
+    psyir = fortran_reader.psyir_from_source(code)
+    calls = psyir.walk(IntrinsicCall)
+    assert len(calls) == 1
+    assert calls[0].argument_names == [None, "ERRMSG"]
```
sergisiso, line 147: The "errmsg" argument name is not listed here: https://psyclone-dev.readthedocs.io/en/latest/psyir.html#intrinsiccall-nodes or in src/psyclone/psyir/nodes/intrinsic_call.py _optional_args

This may be fine (as this test proves that it already works) but can you remind me why did we needed to specify the mold and stats optional arguments in the class constructor? Was it to check the type or there was some ordering issue when some name was missing? 

And would a single literal string can cause the ordering issues if the "errmsg" name is not specified?

Also, to be complete can you consider also the "source" allocate argument.

sergisiso (changes requested):
@arporter I agree that type-spec being a CodeBlock for now is the right approach. But I need a reminder of why other parameters are specified in intrinsic_call and if errmsg and source should also be there.

Other than this, this PR looks good.

src/psyclone/tests/psyir/frontend/fparser2_alloc_handler_test.py
arporter: Thanks for asking this. I've now got to grips with the various arguments that ALLOCATE can accept. It turns out that `errmsg` is an *OUTPUT* variable and thus a literal string is not accepted (fparser2 correctly gives a SyntaxError). I'll add support for `source` too...

arporter: Unfortunately, the additional arguments raise a new problem: `errmsg` only makes sense if `stat` is supplied and `source` and `mold` are mutually exclusive. As it stands, we have no way of enforcing these restrictions. I think there are 4 options:

 1. leave it as it is which allows a user to create PSyIR which will result in invalid Fortran;
 2. add extra dicts such as `_mutually_exclusive` and `_arg_requires` to capture additional information on a per-intrinsic basis;
 3. add a per-intrinsic call-back that performs these checks if required;
 4. change the implementation so that each Intrinsic is implemented as a subclass of IntrinsicCall (which is the proper way of doing option 3).

What do you think @sergisiso and @rupertford?

arporter: In fact, our type checking is already rather lax - we simply check that certain arguments are `Reference`s but really we should check that the Symbol they refer to is of the correct intrinsic type.

Issue #2094, opened 2023-04-03
get_private_clauses for OMPParallelDirective failing when accessing structures


LonelyCat124:
After merging master I found that suddenly this test code is failing in `task1.lower_to_language_level()`, when calling `get_private_clauses` the code fails at a lookup of a structure: ` KeyError: "Could not find 'rval%nx' in the Symbol Table."`

Of course it shouldn't look up `rval%nx` but just `rval`. I don't think any of the tests from #2061 cover structure accesses.

The code for my test is here, the relevant line is presumable (in Fortran, as opposed to my test which constructs it all in PSyIR) `tmp2 = rval%nx` which is inside a parallel do loop. The task stuff in this test is not relevant to this failure I think.
```
    # Check inout StructureReference non-array dependency
    subroutine = Routine("testsub")
    parallel = OMPParallelDirective.create()
    subroutine.addchild(parallel)
    sing = OMPSingleDirective()
    tmp = DataSymbol("tmp", INTEGER_SINGLE_TYPE)
    tmp2 = DataSymbol("tmp2", INTEGER_SINGLE_TYPE)
    tmp3 = DataSymbol("tmp3", INTEGER_SINGLE_TYPE)
    rval = DataSymbol("rval", grid_type)
    subroutine.symbol_table.add(tmp)
    subroutine.symbol_table.add(tmp2)
    subroutine.symbol_table.add(tmp3)
    subroutine.symbol_table.add(rval)

    do_dir = OMPDoDirective()
    assign1 = Assignment.create(
        Reference(tmp2), StructureReference.create(rval, ["nx"])
    )
    loop1 = Loop.create(
        tmp,
        Literal("1", INTEGER_SINGLE_TYPE),
        Literal("128", INTEGER_SINGLE_TYPE),
        Literal("1", INTEGER_SINGLE_TYPE),
        [assign1],
    )
    do_dir.children[0].addchild(loop1)

    task1 = DynamicOMPTaskDirective()
    loop1 = Loop.create(
        tmp,
        Literal("1", INTEGER_SINGLE_TYPE),
        Literal("128", INTEGER_SINGLE_TYPE),
        Literal("32", INTEGER_SINGLE_TYPE),
        [task1],
    )
    assign1 = Assignment.create(
        StructureReference.create(rval, ["nx"]), Reference(tmp2)
    )
    subloop1 = Loop.create(
        tmp2,
        Reference(tmp),
        Literal("32", INTEGER_SINGLE_TYPE),
[...]

Issue #598, opened 2019-11-21
Improved handling of scalar in OpenMP


hiker:
Handling of scalar variables in omp parallel region needs to be improved. Example:
```
!$omp parallel
   a=100
   b=100
   c=100
!$omp do
   do i=1, b
     x(i) = a*i*c
   end do
   c=2305
!$omp do
   do i=1, b
      a = x(i)*x(i)*c
     x(i) = sin(a)
   end do
!$omp end parallel
...
    do i=1, b
      a = x(i)*x(i)
      y(i) = sin(a)
```
The variable 'b' must be a shared variable (since it is assigned once and then read outside of the parallel region - but this is only known after analysing the usage of 'b' after the parallel region. The initial usage of 'a' could be shared (in which case the assignment to a and b could be moved outside of the parallel block), but because of the usage in the second omp loop, it must be declared a private. 'c' on the other hand can be shared.

Issues: if variables are shared, they might need to be enclosed in atomic statements. It might also be better to rename some variable, e.g. 'a' in the second omp loop could be renamed 'a1', in which case the usage of 'a' in the first loop could become shared.

Atm assignments to 'b' etc after omp parallel are declared to be shared - which works as long as they are assigned the same value by all threads, though protecting it with atomic might be better.

Possible outcomes:
- a new transform that can rename scalars depending on usage, so that there is no 'mixed' usage of one variable as shared in one loop, and then private in the next (all in the same omp parallel region).
- this transform could also change the order of assignments - in the example above the assignment to b could be moved up, in which case it would not need to be done inside of the parallel region at all.

arporter:
@LonelyCat124 has correctly pointed out that #2041 is largely a duplicate of this one so I shall close that one and summarise @nmnobre 's finding here. The specific problem occurs in NEMO's `icedyn_rhg_evp` module:

![image](https://user-images.githubusercontent.com/17409212/217856875-d1fddeea-dd59-4d60-b64b-6e39e3dc97d2.png)

In this case, if `ln_aEVP` is `.false.` then every thread will need the value of `z1_alph2` that is assigned before the loop. i.e. `z1_alph2` needs to be `firstprivate` and not `private`.

sergisiso:
Note that this is also the same issue reported for OpenACC in https://github.com/stfc/PSyclone/issues/423#issuecomment-943099981

sergisiso:
I will look at this unless somebody else is already on it

PR #2124, opened 2023-05-03
(closes #2101) Refactor OpenMP inference of private symbols and report symbols needing sync

### LINKED ISSUE #2101: get_private_clauses of OMPParallelDirective incorrectly flagging non-array accesses as firstprivate
I have the following test in the task branch which now fails (and before the `firstprivate` changes did not). I've slightly adjusted it as previously it was easy to argue "k wasn't set", but now k is an input parameter and PSyclone doesn't change its behaviour.

```
    code = '''
    subroutine my_subroutine(k)
        integer :: i, ii
        integer :: j, jj
        integer :: k
        do i = 1, 320, 32
            do j = 1, 320, 32
                do ii=i, i+32
                    do jj = j,j+32
                        k = k + ii
                        k = k * jj
                    end do
                end do
            end do
        end do
    end subroutine
    '''
    tree = fortran_reader.psyir_from_source(code)
    ptrans = OMPParallelTrans()
    strans = OMPSingleTrans()
    tdir = DynamicOMPTaskDirective()
    loops = tree.walk(Loop)
    loop = loops[1].children[3].children[0]
    parent = loop.parent
    loop.detach()
    tdir.children[0].addchild(loop)
    parent.addchild(tdir, index=0)
    strans.apply(loops[0])
    ptrans.apply(loops[0].parent.parent)
    correct = '''subroutine my_subroutine(k)
  integer :: k
  integer :: i
  integer :: ii
  integer :: j
  integer :: jj


  !$omp parallel default(shared), private(i,ii,j,jj)
  !$omp single
  do i = 1, 320, 32
    do j = 1, 320, 32
      !$omp task private(ii,jj), firstprivate(i,j), shared(k), \
depend(in: k), depend(out: k)
      do ii = i, i + 32, 1
        do jj = j, j + 32, 1
          k = k + ii
          k = k * jj
        enddo
      enddo
      !$omp end task
    enddo
  enddo
  !$omp end single
  !$omp end parallel

end subroutine my_subroutine\n'''
    assert fortran_writer(tree) == correct
```

For some reason, the access `k = k + ii` is being flagged as a `firstprivate` access - I think because its a non-array access, however since its an assignment to `k`, it should be `shared` - if this was in an `OMPDoDirective` then we'd obviously have a race condition (which is not allowed), however the data concurrency issue is handle by the dependency on the task.

The output from `fortran_writer(tree)` is instead:
```
subroutine my_subroutine(k)
  integer :: k
  integer :: i
  integer :: ii
  integer :: j
  integer :: jj

  !$omp parallel default(shared), private(i,ii,j,jj), firstprivate(k)
  !$omp single
  do i = 1, 320, 32
    do j = 1, 320, 32
      !$omp task private(ii,jj), firstprivate(i,j), shared(k), depend(in: k), depend(out: k)
      do ii = i, i + 32, 1
        do jj = j, j + 32, 1
          k = k + ii
          k = k * jj
        enddo
      enddo
      !$omp end task
    enddo
  enddo
  !$omp end single
  !$omp end parallel

end subroutine my_subroutine
```

I think this affects a lot of my tests, and are instances with PSyclone generating incorrect code.

@sergisiso @nmnobre What were the rules decided for determining if a variable in PSyclone is `firstprivate` as opposed to `shared`? Maybe this needs to be smarter - perhaps variable accesses inside OMPSerialRegions need difference rules?

sergisiso:


sergisiso:
@arporter @rupertford This is ready for a first review.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -714,28 +762,27 @@ def _get_private_clauses(self):
...
+                        if has_been_read:
+                            loop_pos = loop_ancestor.loop_body.abs_position
+                            if last_read_position > loop_pos:
+                                # It is read in the same loop_body, so it is
+                                # a race condition
+                                need_sync.add(symbol)
+                            else:
+                                fprivate.add(symbol)
+                        elif conditional_write:
                             fprivate.add(symbol)
```
arporter, line 774: I can see that the logic here is essentially unchanged but I can't remember the details: why does a conditional write (without a read) imply first private? Is it because it may subsequently be read within the loop? Could/should we check for such a read? Please could you add a comment to explain.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -101,7 +102,8 @@ def test_ompparallel_changes_begin_string(fortran_reader):
     assert isinstance(pdir.children[3], OMPFirstprivateClause)
     priv_clause = pdir.children[2]
 
-    # Make acopy of the loop
+    # If the code inside the region changes after lowering, the next lowering
+    # will update the clauses appropriately
```
arporter, line 106: Is this your preferred mechanism for keep clauses up-to-date? I ask because in #2157 I've introduced an 'update' mechanism to do this: changes to ChildrenList trigger each ancestor node to update (if necessary).

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -152,20 +177,27 @@ def test_ompparallel_changes_gen_code(monkeypatch):
     assert "private(cell,k)" in code
 
     # Monkeypatch a case with private and firstprivate clauses
-    pclause = OMPPrivateClause(children=[Reference(Symbol("a"))])
-    fpclause = OMPFirstprivateClause(children=[Reference(Symbol("b"))])
-    monkeypatch.setattr(pdir, "_get_private_clauses",
-                        lambda: (pclause, fpclause))
+    monkeypatch.setattr(pdir, "_infer_sharing_attributes",
```
arporter, line 180: I like this way of testing - it's better than having to give it different Fortran input :-)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -558,27 +621,57 @@ def test_directive_get_private_lfric():
...
+    pvars, fpvars, sync = directive._infer_sharing_attributes()
+    assert len(pvars) == 1
+    assert list(pvars)[0].name == 'i'
+    assert len(fpvars) == 0
+    assert len(sync) == 0
 
     # Example with private and firstprivate variables on OMPParallelDoDirective
+    # In this case scalar1 is firstprivate because it has a conditional-write
+    # that it is not guaranteed to happen on each iteration, so the private
+    # variable needs to be initialised with the value it has before the loop.
```
arporter, line 674: But only if it is read within the loop?

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -616,17 +710,138 @@ def test_directive_get_private(fortran_reader):
...
+        end subroutine''')
+    omptrans = OMPParallelTrans()
+    routine = psyir.walk(Routine)[0]
+    omptrans.apply(routine.children)
+    directive = psyir.walk(OMPParallelDirective)[0]
+    pvars, fpvars, sync = directive._infer_sharing_attributes()
+    assert len(pvars) == 1
+    assert len(fpvars) == 1
+    assert len(sync) == 0
+    assert sorted(pvars, key=lambda x: x.name)[0].name == 'i'
```
arporter, line 748: You don't need these two sorts because there's only one symbol in each list.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -616,17 +710,138 @@ def test_directive_get_private(fortran_reader):
...
+    assert sorted(fpvars, key=lambda x: x.name)[0].name == 'scalar2'
+
+    # Similar but with a OMPParallelDoDirective and the firstprivate is
+    # in the same loop but before the loop body
+    psyir = fortran_reader.psyir_from_source('''
+        subroutine my_subroutine()
+            integer :: i, scalar1
+            real, dimension(10) :: array
+            do i = 1, 10, scalar1
+                scalar1 = array(i)
```
arporter, line 758: I don't think this is valid Fortran - you can't change a variable that defines the iteration space of a loop within that loop. However, since this is testing `_infer_sharing_attributes` we could ignore that problem and just note it in a comment.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -616,17 +710,138 @@ def test_directive_get_private(fortran_reader):
...
+        end subroutine''')
+    omplooptrans = OMPParallelLoopTrans()
+    loop = psyir.walk(Loop)[0]
+    omplooptrans.apply(loop, options={'force': True})
+    directive = psyir.walk(OMPParallelDoDirective)[0]
+    pvars, fpvars, sync = directive._infer_sharing_attributes()
+    assert len(pvars) == 1
+    assert len(fpvars) == 1
+    assert len(sync) == 0
+    assert sorted(pvars, key=lambda x: x.name)[0].name == 'i'
```
arporter, line 770: You don't need to sort these.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -616,17 +710,138 @@ def test_directive_get_private(fortran_reader):
...
+    may return different results than the original code.
+    '''
+
+    # In this example the result will take the value of the last i in
+    # sequential order, but an arbitrary i in parallel.
+    psyir = fortran_reader.psyir_from_source('''
+        subroutine my_subroutine()
+            integer :: i, result
+            do i = 1, 10
+               result = i
```
arporter, line 833: Oh yes, this is the example I was worrying about earlier - if `result` was read after the loop then we'd have a problem. Do we have an Issue for this?

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -616,17 +710,138 @@ def test_directive_get_private(fortran_reader):
...
+            enddo
+        end subroutine''')
+    omplooptrans = OMPParallelLoopTrans()
+    loop = psyir.walk(Loop)[0]
+    omplooptrans.apply(loop, options={"force": True})
+    directive = psyir.walk(OMPParallelDoDirective)[0]
+    pvars, fpvars, sync = directive._infer_sharing_attributes()
+    assert len(pvars) == 1
+    assert sorted(pvars, key=lambda x: x.name)[0].name == 'i'
```
arporter, line 842: No need to sort here.

arporter (changes requested):
Looks good Sergi. Mostly minor things apart from shared variables that live-out from a loop. Hopefully we already have an issue for this? I'm also not sure that this PR solves @LonelyCat124 's comment: https://github.com/stfc/PSyclone/issues/2101#issuecomment-1513116534
Just one minor pylint issue (see inline).
Ref. guide builds fine.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -591,14 +600,31 @@ def lower_to_language_level(self):
         :returns: the lowered version of this node.
         :rtype: :py:class:`psyclone.psyir.node.Node`
 
+
```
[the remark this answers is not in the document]
sergisiso: Done, and happy to get rid of the backslashes :)

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -591,14 +600,31 @@ def lower_to_language_level(self):
...
+        :raises GenerationError: if the OpenMP directive needs some \
+            synchronisation mechanism to create valid code. This are not \
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -591,14 +600,31 @@ def lower_to_language_level(self):
...
+        # Create data sharing clauses (order alphabetically to make generation
+        # reproducible)
+        private, fprivate, need_sync = self._infer_sharing_attributes()
+        private_clause = OMPPrivateClause.create(
+                            sorted(private, key=lambda x: x.name))
+        fprivate_clause = OMPFirstprivateClause.create(
+                            sorted(fprivate, key=lambda x: x.name))
+        if need_sync:
+            raise GenerationError(
+                f"Lowering {type(self).__name__} does not support symbols that "
```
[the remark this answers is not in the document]
sergisiso: Fixed

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -671,14 +717,16 @@ def _get_private_clauses(self):
...
+            # We consider potential private variables the ones that are written
+            # in every iteration of a loop.
+            # If one such scalar is read before it is written, it will be
+            # considered firstprivate.
             has_been_read = False
+            last_read_position = 0
             for access in accesses:
                 if access.access_type == AccessType.READ:
                     has_been_read = True
+                    last_read_position = access.node.abs_position
```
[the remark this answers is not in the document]
sergisiso: I don't think so, see below

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -714,28 +762,27 @@ def _get_private_clauses(self):
                             limit=loop_ancestor,
                             include_self=True)
 
-                        # If a previous value might be needed we mark it as
-                        # firstprivate
-                        if has_been_read or conditional_write:
+                        if has_been_read:
+                            loop_pos = loop_ancestor.loop_body.abs_position
```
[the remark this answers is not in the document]
sergisiso: That's right, access.location uses its own numbering different from the nodes numbering. So I need the node location to compare to other nodes.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: lastprivate is not execution order (which I always thought it was), the standard says "value from the sequentially last iteration of the associated loops" so it could solve it for certain cases, however, I don't know what happens when the last iteration has the value inside a conditional and the conditional is not true.

I am not too concerned about those because the transformations would not let you parallelise the loop by default, it needs the `option={'force': True}` before reaching the inference of sharing attributes, so it is already explicit that we are forcing the parallelism into it.

src/psyclone/psyir/nodes/omp_directives.py
sergisiso: > Is it because it may subsequently be read within the loop? 

yes, and if the write didn't happen we need its original value. I extended the comment to make this clearer.

> Could/should we check for such a read?

I don't think so because we may want to maintain the original value also after the loop. This may not happen yet because we don't deal well with accesses after the parallel region (this is tracked in #598) but this is independent from this logic (e.g. OpenMP allows being `firstprivate(x) lastprivate(x)` at the same time). Also I put a TODO for write-only accesses before this point that may also make a check here unecessary.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -714,28 +762,27 @@ def _get_private_clauses(self):
...
+                                # a race condition
+                                need_sync.add(symbol)
+                            else:
+                                fprivate.add(symbol)
+                        elif conditional_write:
                             fprivate.add(symbol)
                         else:
                             private.add(symbol)
+                    else:
+                        # If we find it outside a loop
```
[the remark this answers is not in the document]
sergisiso: Added more comments and removed conditions (in exchange for breaks) as suggested below.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: I don't particularly like how this is currently implemented.

I prefer your "sending an update signal upwards to invalidate/recalculate cached values" solution. (I would prefer even more not caching/duplicating information at all, but I know it is not always simple and your solution will be useful in several places when the duplication exists.)

We can update this after your PR is on master.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -558,27 +621,57 @@ def test_directive_get_private_lfric():
...
+    assert list(pvars)[0].name == 'cell'
 
     directive.children[1] = OMPDefaultClause(
             clause_type=OMPDefaultClause.DefaultClauseTypes.NONE)
     with pytest.raises(GenerationError) as excinfo:
-        _ = directive._get_private_clauses()
+        _ = directive._infer_sharing_attributes()
     assert ("OMPParallelClause cannot correctly generate the private clause "
             "when its default data sharing attribute in its default clause is "
             "not shared." in str(excinfo.value))
```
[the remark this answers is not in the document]
sergisiso: Done

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Deleted

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: This is valid Fortran :) as you say the variable can't change, but it is read before starting the loop. This is the same as the previous example but making sure that read accessors in the loop bounds are also considered (as being outside the loop).

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Removed sort

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Yes, I added a TODO now, but also see test below with a more clear use-after-loop and a pytest.xfail for the appropriate issue.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: removed

sergisiso:
@arporter This is ready for another review.

> I'm also not sure that this PR solves @LonelyCat124 's comment: https://github.com/stfc/PSyclone/issues/2101#issuecomment-1513116534

@LonelyCat124 can correctly if I am wrong but I think we agreed that the issue referenced in that comment was not a big deal, it just produces a warning but the code is still right. Removing the warning/copy of uninitialized data would be hard for the generic case. I suggest ignoring that warning for now (we can close the issue with this PR).

However, what is solved is the issue mentioned in the first comment, which is the one that is a blocker. With this PR the `k` variable is not reported as firstprivate anymore, but as a `need_sync`, and @LonelyCat124 can ignore it or verify that the depend clauses already deal with it.

I added a test case for that specific tiling code from the first comment.

src/psyclone/psyir/nodes/omp_directives.py
arporter: OK. Thanks for extending the comments and adding TODOs.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -729,57 +758,54 @@ def _infer_sharing_attributes(self):
...
-                                # It is read in the same loop_body, so it is
-                                # a race condition
-                                need_sync.add(symbol)
-                            else:
-                                fprivate.add(symbol)
-                        elif conditional_write:
+                    # Otherwise, the assignment to this variable is inside a
+                    # loop (and it will be repeated for each iteration), so
+                    # we declare it as private or need_synch
+                    name = str(signature).lower()
```
arporter, line 781: What happens if the variable in question is some sort of structure access, e.g. `a%b`? Will this still work?

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -697,13 +705,26 @@ def _infer_sharing_attributes(self):
...
+        # of variables that we currently declare as private. This should be
+        # lastprivate.
+        # e.g:
+        # !$omp parallel do <- will set private(ji, my_index)
+        # do ji = 1, jpk
+        #   my_index = ji+1
+        #   array(my_index) = 2
+        # enddo
+        # #end do
+        # call func(my_index) <- my_index does not have a value
```
arporter, line 721: Strictly speaking, won't it have the value given it by the last thread to write to it?

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -729,57 +758,54 @@ def _infer_sharing_attributes(self):
...
+                    # loop (and it will be repeated for each iteration), so
+                    # we declare it as private or need_synch
+                    name = str(signature).lower()
+                    symbol = access.node.scope.symbol_table.lookup(name)
+
+                    # If it has been read before we have to check if ...
+                    if has_been_read:
+                        loop_pos = loop_ancestor.loop_body.abs_position
+                        if last_read_position < loop_pos:
+                            # .. it was before the loop, so it is fprivate
```
arporter, line 788: Nice comments, thanks :-)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
arporter: Great. Please could you add a TODO for #2157.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
arporter: Ah! I live and learn :-) At least it let me look knowledgeable when answering @LonelyCat124 's question on Teams ;-)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
arporter: Oh yes, thanks.

arporter (changes requested):
Very nearly there now. Please could you add/extend the tests so that they also cover the case where we have to analyse the accesses to a StructureReference rather than just a Reference.
I've triggered the integration workflow and will check the results before the next review.

arporter:
Just realised that the potential issue with structures is probably what @LonelyCat124 is addressing in #2094. Probably best to check with him.

src/psyclone/psyir/nodes/omp_directives.py
hiker: Just use `signature[0]` or `signature.var_name` (if the latter doesn't exceed line length :)  ). I don't think lower is required, symbol table `_normalize`s the name it looks up, which means it converts it to lower case.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Done

src/psyclone/psyir/nodes/omp_directives.py
sergisiso: Not if it is private, at least the standard doesn't say so. I have updated to "my_index has not been updated"

src/psyclone/psyir/nodes/omp_directives.py
sergisiso: I used the suggested line. Although this just considers the top-level component of the derived type accessor. I added a TODO for the more complex types.

sergisiso:
@arporter If CI passes this is ready for another look.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -848,6 +850,65 @@ def test_directive_infer_sharing_attributes(fortran_reader):
...
+            enddo
+        end subroutine''')
+    omptrans = OMPParallelTrans()
+    routine = psyir.walk(Routine)[0]
+    omptrans.apply(routine.children)
+    directive = psyir.walk(OMPParallelDirective)[0]
+    pvars, fpvars, sync = directive._infer_sharing_attributes()
+    assert len(pvars) == 2
+    assert sorted(pvars, key=lambda x: x.name)[0].name == 'i'
+    assert sorted(pvars, key=lambda x: x.name)[1].name == 'scalar1'
```
arporter, line 879: Should this be an `xfail` too as `scalar1` is actually a member of `mt1` so doesn't exist in isolation?

arporter (changes requested):
Very, very nearly there. I think there's just one test that should also be `xfail` as the result of the _infer_sharing_attributes only gives the name of a member of a structure.
I've triggered the integration tests so will check on them at a future point.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: No, scalar1 is also an independent scalar variable.

`mt1` (which has the scalar1 component) is shared because it is just written outside the loop (hence once)

sergisiso:
@arporter I put this ready for review again without changes, I think the example is correct and should not xfail.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
arporter: Oh yes, sorry, I completely missed that.

arporter:
Integration tests are green.

arporter (approved):
All good now (well, in fact is was all good before but I wasn't seeing things correctly).

PR #2131, opened 2023-05-07
Improve error handling when builtin symbol is a call (toward #1618)

rupertford:
If a builtin is used in the algorithm layer then its name should not be used as variable, subroutine, module name etc. This PR improves PSyclone's error handling when this is the case.


rupertford:
Ready for first review from @arporter or @sergisiso

src/psyclone/tests/domain/lfric/transformations/raise_psyir_2_lfric_alg_trans_test.py
```diff
@@ -208,6 +208,32 @@ def test_codeblock_invalid(monkeypatch, fortran_reader):
...
+        "  call invoke(setval_c(field, value))\n"
+        "end subroutine setval_c\n")
+    psyir = fortran_reader.psyir_from_source(code)
+    invoke_trans = RaisePSyIR2LFRicAlgTrans()
+    with pytest.raises(TransformationError) as info:
+        invoke_trans.validate(psyir.children[0][0])
+    assert ("The arguments to this invoke call are expected to be kernel "
+            "calls which are represented in generic PSyIR as CodeBlocks "
+            "or ArrayReferences, but 'setval_c(field, value)' is of type "
+            "'Call'. Has 'setval_c' also been used as a routine name in "
```
arporter, line 233: Can we do better? i.e. presumably it ends up as a `Call` because we already have a `RoutineSymbol` named 'setval_c'. It might be better if we could say this as well as saying which SymbolTable we found it in?

src/psyclone/tests/generator_test.py
```diff
@@ -745,6 +745,46 @@ def test_main_expected_fatal_error(capsys):
...
+    filename = str(tmpdir.join("alg.f90"))
+    with open(filename, "w", encoding='utf-8') as my_file:
+        my_file.write(code)
+    with pytest.raises(SystemExit) as excinfo:
+        main([filename])
+    # the error code should be 1
+    assert str(excinfo.value) == "1"
+    _, output = capsys.readouterr()
+    expected_output = (
+        "Transformation Error: Error in RaisePSyIR2LFRicAlgTrans "
```
arporter, line 779: As I said earlier, I'm not convinced that a call to 'generate' should raise a TransformationError but we can discuss. Perhaps we need a new `RaisingError` or something?

arporter (changes requested):
Thanks for improving things. You're not going to like me saying this but I think the error message could be further improved - we should be able to report quite a lot more information rather than saying "have you...?". I'm also not sure that calling `generate` should be permitted to result in a `TransformationError` - that (to me) is quite counter intuitive. Can we catch the error lower down and then re-raise as something else?
Tests with compilation are still broken but that's not your fault (waiting on #2136).
I'll check the examples next time.
We should document the rule that name clashes with LFRic Builtins are not permitted - probably in the Builtins section of the UG.

src/psyclone/generator.py
```diff
@@ -64,6 +64,7 @@
 from psyclone.domain.lfric.transformations import (
     LFRicAlgTrans, RaisePSyIR2LFRicKernTrans, LFRicAlgInvoke2PSyCallTrans)
 from psyclone.errors import GenerationError, InternalError
+from psyclone.psyir.transformations import TransformationError
```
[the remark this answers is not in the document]
rupertford: soop.

src/psyclone/tests/domain/common/transformations/raise_psyir_2_alg_trans_test.py
```diff
@@ -340,8 +340,9 @@ def test_arg_error(fortran_reader):
     invoke_trans = RaisePSyIR2AlgTrans()
     with pytest.raises(TransformationError) as info:
         invoke_trans.validate(psyir.children[0][0])
-    assert ("The arguments to this invoke call are expected to be a "
-            "CodeBlock or an ArrayReference, but found 'Literal'."
+    assert ("The arguments to this invoke call are expected to be kernel "
+            "calls which are represented in generic PSyIR as CodeBlocks "
+            "or ArrayReferences, but ''hello'' is of type 'Literal'."
```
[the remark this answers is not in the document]
rupertford: As discussed, this is because literal strings keep the original quotes atm so I've added a comment referencing the issue.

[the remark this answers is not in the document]
rupertford: We use the debug writer and it outputs strings with quotes, hence the extra quotes here. If it is not a string then we only get single quotes. I've changed the example to an integer literal but that does not fix the above problem. We would have to strip quotes from the output if they exist I think. I can do this if you think it needs doing.

src/psyclone/tests/generator_test.py
rupertford: Modified as agreed.

rupertford:

rupertford:
I'm afraid the user-guide pdf is re-written in this PR even though it is not needed as I updated it (and the changelog by accident). I could have reverted but this would be on a merge and I wasn't confident enough to choose what merges to what so have undone the changelog modification but kept the updated pdf.

src/psyclone/tests/domain/common/transformations/raise_psyir_2_alg_trans_test.py
```diff
@@ -323,26 +323,34 @@ def test_array_reference(fortran_reader):
...
         invoke_trans.validate(psyir.children[0][0])
-    assert ("The arguments to this invoke call are expected to be kernel "
-            "calls which are represented in generic PSyIR as CodeBlocks "
-            "or ArrayReferences, but ''hello'' is of type 'Literal'."
+    if arg == "alg(field)":
+        assert ("Error in RaisePSyIR2AlgTrans transformation. The invoke "
+                "call argument 'alg' has been used as a routine name. This "
+                "is not allowed." in str(info.value))
+    else:
+        print(info.value)
```
arporter, line 349: Please rm this print.

src/psyclone/tests/generator_test.py
```diff
@@ -775,14 +775,16 @@ def test_generate_trans_error(tmpdir, capsys, monkeypatch):
...
-        "the code?\n")
-    assert output == expected_output
+    # The output is split as the location of the algorithm file varies
+    # due to it being stored in a temporary directory by pytest.
+    expected_output1 = "Generation Error: In algorithm file '"
+    expected_output2 = (
+        "alg.f90':\nTransformation Error: Error in RaisePSyIR2LFRicAlgTrans "
+        "transformation. The invoke call argument 'setval_c' has been used as"
+        " a routine name. This is not allowed.\n")
+    print(output)
```
arporter, line 785: Please rm this `print`.

arporter (changes requested):
Thanks for updating.
Examples are all fine with compilation.
I get one test failure:

    E       assert "The arguments to this invoke call are expected to be kernel calls which are represented in generic PSyIR as CodeBlock...ferences, but 'setval_c(field, value)' is of type 'Call'. Has 'setval_c' also been used as a routine name in the code?" in "Transformation Error: Error in RaisePSyIR2LFRicAlgTrans transformation. The invoke call argument 'setval_c' has been used as a routine name. This is not allowed."
    E        +  where "Transformation Error: Error in RaisePSyIR2LFRicAlgTrans transformation. 
    The invoke call argument 'setval_c' has been used as a routine name. This is not allowed." = str(TransformationError())
    E        +    where TransformationError() = <ExceptionInfo TransformationError() tblen=2>.value

    domain/lfric/transformations/raise_psyir_2_lfric_alg_trans_test.py:230: AssertionError

I've taken the liberty of reverting the UG pdf back to the most recent version on master.
Having said that, I think it will change anyway as we still need to make sure that the UG
clearly states that having a routine with the same name as a Builtin is forbidden - please could you take a look?
Changes are pylint clean. Ref. guide builds without new warnings.

src/psyclone/tests/domain/common/transformations/raise_psyir_2_alg_trans_test.py
rupertford: Oops

src/psyclone/tests/generator_test.py
rupertford: Oops again

rupertford:
Fixed test error and added a note to the user guide.

rupertford:

arporter (approved):
Updated UG builds fine. Ref. guide builds fine.
Tests and examples are OK with compilation.
Will proceed to merge once #2170 is on.

PR #2133, opened 2023-05-09
(Closes #2126) add YAML config files to fix RTD builds

### LINKED ISSUE #2126: Dev Guide and PSyAD UG builds failing on RTD
The Sphinx builds are failing with, e.g.:

    Extension error:
    Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168)

arporter:


arporter:
Contrary to what I said originally, this PR uses a new 'experimental' feature of RTD that lets you specify the location of a YAML configuration file for your document. This then means we can have a separate configuration file for each document and use that to use a more recent build environment which does away with the problem of urllib3. I've left the existing `pip_requirements.txt` files unchanged in case this 'experimental' feature disappears again in the future.
Ready for review from anyone :-)

rupertford (approved):
Adds 4 new yaml files for RTD. I don't know much about the required format for these but the code looks good. Approving.

Issue #2132, opened 2023-05-09
Self-hosted runner to perform integration tests


sergisiso:
A self-hosted runner is useful to do integration tests with our local hardware (HPC-grade CPUs, GPUs) and installed compiler toolchains, that would take a lot of compute resources on the free GH cloud runners.

#967 adds two GH workflows that:
1. on reviewer's demand, push the branch to a private repo
2. a self-hosted runner executes integration tests workflow on the private repo commits.

The current solution has some limitations:
- It only executes pytest compilation tests, but we could also do PSycloneBench, NEMO, LFRic, ...
- The private repo is on @arporter account and username, it could be more generic.
- The self-hosted environment is currently custom made (not captured/replicable). It would be good to have a container file / spack script in this repository that set up the self hosted environment, so other people/orgs could run it locally if they wanted to. It would also document how to create it for the next time we need to do it.
- The private repo. currently runs all of the same actions that the public repo. does - this wastes GHA compute resources.

arporter:
#2106 and #2107 have added integration tests for NEMO. However, we have not yet migrated everything over to the relevant user on the self-hosted runner. @sergisiso commented:

I have not moved the models and software(perl and intel netcdf). The former are quite big and I am unsure if we want to replicate them, the latter should maybe be ported to the module system, or find a reproducible (container?) way to be able to move the runner to other servers.

arporter:
Of the original points in this Issue, only:

"The self-hosted environment is currently custom made (not captured/replicable). It would be good to have a container file / spack script in this repository that set up the self hosted environment, so other people/orgs could run it locally if they wanted to. It would also document how to create it for the next time we need to do it."

remains. Sergi has made quite a lot of progress towards using Spack to do this and work will continue in the background so I'm closing this issue.

PR #2150, opened 2023-05-22
(Towards #2132) change private repo to be in stfc organisation.

arporter:
A small PR that changes the private repo used to launch heavier compilation testing etc. Previously this was under my personal github account but now it is under the stfc organisation. I've also taken the opportunity to improve things slightly by making the necessary secrets (for ssh) only available in the `deployment` environment.

I've also altered the other action configuration files so that all the other tests *do not* get launched by the mirror repo. (Currently they get duplicated which is very inefficient.)


arporter:
I've also been slightly naughty and updated the various RTD confg. files to ensure that graphviz gets installed as I noticed we are not getting the inheritance diagrams produced.

arporter:
This is ready for a first look now from @hiker, @rupertford or @sergisiso. It addresses some of the limitations identified in #2132, primarily by changing the private repo that is used for the compilation tests to one that is in the stfc organisation. I'll make a new document on Teams to record the internal details.

arporter:
(Note that, as with the first PR that added compilation testing, the button to launch it will only be updated once this hits master. For testing purposes, you can change the `on` setting to be `push`.)

sergisiso (changes requested):
@arporter It looks good and I can see in the commit history that it triggers the jobs in the new repo successfully. I just have a couple of comments/suggestions:

- Instead of `if: ${{ github.repository == 'stfc/PSyclone' }}` we could use `if: ${{ github.repository != 'stfc/PSyclone-mirror' }` to enable regular tests. Some people fork the repo and work on it until they have a PR ready and they would expect the CI to work.

- I noticed that the compile tests are done with python3.6 (glados default), but this version is not supported/tested by psyclone. We could be in a situation where the compilation test fails because of a difference btw 3.6 and 3.7. One solution it would be to update the run target to do `python3.8 -m venv .runner_venv` which is also available or instead do a `module load python` before the venv (which uses 3.11).

arporter:
Thanks @sergisiso and good spot on the Python version. I've done as you suggested and altered it to load the python/3.11.3 module before running. Also a good spot on the issue of running actions from forks so I've changed to `!= stfc/PSyclone-mirror` as you suggested.

arporter:
Using the Python 3.11.3 module on Glados doesn't seem to work. I get an error related to sqlite when attempting to run the tests. I can run them OK on my laptop with this python version so it must be something about the environment setup on Glados.

arporter:
OK, working with 3.11.3 now. Ready for another look.

Issue #1802, opened 2022-07-17
sympy expression errors in psyad


rupertford:
When trying to process poly1d_vert_adv_kernel_mod.F90 with psyad sympy expression errors occur.

hiker:
Any details? Can you perhaps provide the file?

arporter:
An update, the problem here is triggered by:

    stencil(p+1) = k - floor(real(vertical_order,r_def)/2.0_r_def) + p

which gives:

    File "/home/kbc59144/.pyenv/versions/3.10.7/envs/psyclone310/lib/python3.10/site-packages/sympy/parsing/sympy_parser.py", line 907, in eval_expr
    expr = eval(
    File "<string>", line 1
    k -Function ('FLOOR' )(Function ('REAL' )(Symbol ('vertical_order' ),Symbol ('r_def' ))/Float ('2.0' )Symbol ('_r_def' ))+p 
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    SyntaxError: invalid syntax. Perhaps you forgot a comma?

arporter:
You can reproduce this failure by doing:

    cd PSyclone/examples/psyad/lfric
    psyad -api lfric -a advective -- tangent_linear/tl_poly1d_vert_adv_kernel_mod.F90

hiker:
Issue is that `floor` is not supported in PSyIR (it ends up as a code block).
Three changes required:
1. 
``` 
--- a/src/psyclone/psyir/frontend/fparser2.py
+++ b/src/psyclone/psyir/frontend/fparser2.py
@@ -962,6 +962,7 @@ class Fparser2Reader():
         ('.not.', UnaryOperation.Operator.NOT),
         ('abs', UnaryOperation.Operator.ABS),
         ('ceiling', UnaryOperation.Operator.CEIL),
+        ('floor', UnaryOperation.Operator.FLOOR),
         ('exp', UnaryOperation.Operator.EXP),
         ('log', UnaryOperation.Operator.LOG),
         ('log10', UnaryOperation.Operator.LOG10),
```
3.
```
 src/psyclone/psyir/nodes/operation.py
index f50a4f85b..59c5da980 100644
--- a/src/psyclone/psyir/nodes/operation.py
[...]

Issue #1655, opened 2022-03-13
sympy wrappers work with range nodes


rupertford:
Range nodes currently cause the sympy wrappers to fail.

Issue #1537, opened 2021-12-03
Sympy comparison fails if array notation is used


arporter:
The PSyIR supports array slicing using ':' notation. However, this is causing test failures in #1526 because Sympy doesn't understand it:

    psyad/transformations/test_assignment_trans.py:574: 
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    psyad/transformations/test_assignment_trans.py:105: in check_adjoint
        trans.apply(assignment)
    ../psyad/transformations/assignment_trans.py:94: in apply
        if sym_maths.equal(ref, node.lhs):
    ../core/symbolic_maths.py:111: in equal
        str_exp1 = parse_expr(self._writer(exp1))
    /home/kbc59144/.python_envs/py3/lib/python3.8/site-packages/sympy/parsing/sympy_parser.py:1026: in parse_expr
        raise e from ValueError(f"Error from parse_expr with transformed code: {code!r}")
    /home/kbc59144/.python_envs/py3/lib/python3.8/site-packages/sympy/parsing/sympy_parser.py:1017: in parse_expr
        rv = eval_expr(code, local_dict, global_dict)
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    code = "Function ('a' )(Integer (2 ):Integer (3 ))", local_dict = {}
    global_dict = {'Abs': Abs, 'AccumBounds': <class 'sympy.calculus.util.AccumulationBounds'>, 'Add': <class 'sympy.core.add.Add'>, 'Adjoint': <class 'sympy.matrices.expressions.adjoint.Adjoint'>, ...}

This problem occurs in psyad/transformations/assignment_trans.py when checking whether a reference on the RHS matches that on the LHS of an assignment (in order to identify whether it is performing an increment).

arporter:
I've implemented a workaround in #1526 for now (with a TODO to this issue). Possibly a slightly better version of what I've done could be included in the `equal()` method.

sergisiso:
This is also a problem in one nemo file that has the expression: `p2d(ji:ji2,jj:jj2)`
This is checked for dependencies when trying to add OpenMP and fails with:
```
[...]

PR #2155, opened 2023-05-30
2120 array and array expressions in sympy

### LINKED ISSUE #2122: SymPy - array access with and without index in one expression
The following two expressions causes an exception:
```
 sym_maths.equal(psyir_of("A"), psyir_of("A(i)"))
 sym_maths.equal(psyir_of("A(i)"), psyir_of("A"))
```
trigger two problems:
1. TypeError: 'Symbol' object is not callable (for the first example)
2. TypeError: unsupported operand type(s) for -: 'a' and 'UndefinedFunction'

Caused by inconsistencies in using 'a' as scalar or as a sympy function. @rupertford indicated that enforcing a symbol to be a sympy function if there is at least one array access helps.

hiker:
This implements support for array expressions in SymPy, and should fix #2122, #1537, and #1655. Not sure about #1802, I need additional details for that.

Big question: psyad relies on `SymPy.expand`-ing an expression, then parsing it again as Fortran and inserting it into PSyIR. With this solution (replacing each index with three parameters for the corresponding sympy function), this does not work anymore.

Two possible solutions:
1. After expanding and parsing, we replace all triple-indices with the corresponding single index again. Sounds a bit messy imho :)
2. We add an option to disable the replacement of each index with triple function parameters (which means disabling support for array expressions) in SymPyWriter. PSyAD could then use this to work as good as it does before this PR.

@rupertford , @arporter , @sergisiso  - feedback required :)


hiker:
I am closing this since it had the wrong branch number (2120 is a different ticket).
New branch: 1537_sympy_array_expressions

Issue #2165, opened 2023-06-07
Add support for ASSOCIATE blocks


arporter:
Fortran 2003 has the ASSOCIATE block:
![image](https://github.com/stfc/PSyclone/assets/17409212/0c907558-aced-4124-bbc9-6aff4d2856cb)
and this is used quite heavily in NEMOVAR.
Essentially, it boils down to syntactic sugar in order to simplify the use of complex expressions, e.g.:
![image](https://github.com/stfc/PSyclone/assets/17409212/8a504b16-dd11-44c2-92cd-110d77258e76)

arporter:
I thought about this some more last night and decided it wouldn't actually be that easy - as illustrated by the last two examples above, the associate-name can be a reference to a structure access. This is then effectively the same problem that we've already solved in the in-lining support. That might need some refactoring to support this case.

arporter:
There is a subtlety:
![image](https://github.com/stfc/PSyclone/assets/17409212/1a196345-b84a-47fe-918a-816ebe6fddc3)
this means that if we simply replace an associate-name with its corresponding expression, we have to make sure that none of its constituent references are written to before the substitution.

arporter:
I now have a first working version that simply creates an instance of `InlineTrans` and uses its internal method to perform the necessary substitutions. This works out of the box for quite complex examples :-) I now need to:
  - [ ] tidy the code in InlineTrans;
  - [ ] move that code to a different location (maybe in tools?);
  - [ ] allow for the subtlety above by using dependence analysis to check we can't break anything;

arporter:
I started implementing functionality to assign the associate-expressions to local temporaries and immediately hit (yet) another problem: associate-names can appear on the *LHS* of assignments (as in the examples in the description of this Issue). Obviously it then makes no sense to have a local temporary :-(

arporter:
Unfortunately, if an associate-name is passed to a Call then we don't know whether or not it is written to.

arporter:
Our normal approach to this sort of problem is to give the user the nuclear option (i.e. tell us that it is safe). However, this functionality is in the PSyIR front end and so we have no current mechanism for exposing such an option (and then it would have to be on a per-file basis anyway).

An alternative approach would be to leave the front end as is and supply a Transformation that attempts to convert a CodeBlock containing an ASSOCIATE block into code. That could then be given a `force` option. What do @rupertford and @sergisiso think?

rupertford:
[...]

PR #2167, opened 2023-06-09
(Towards #2165) add support for associate constructs.

arporter:
This is an initial cut that we can improve upon if we need to. In particular, the solution chosen here replaces instances of associate-name with the corresponding expression. However, since that expression must be evaluated at the start of the construct, if any of the variables it references are written to within the construct then this substitution cannot be performed. We could refine this by checking the precise location at which any such writes are performed - if they happen to only occur *after* lines where we need to perform the substitution then this would still be OK.

Alternatively, we could attempt to determine the type of the expression and then store its value in a new variable at the start of the construct.


arporter:
I need to test this branch with some real NEMOVAR code before putting it as ready for review.

arporter:
Of course, the very first NEMOVAR I tried this on had some WRITE stmts inside the associate construct so we're no further forward. I can refine by checking which symbols are referenced by the resulting CodeBlock but that may not help. If that's  so then I think the only option is to remove the WRITES?

arporter:
I could refine what I have to check that any Call or Write is not the last statement in the construct. Beyond that, I could check whether there are any reads of any associate names after a Call or write.

arporter:
The ASSOCIATE constructs in the NV code we looked at today all contained calls and so this approach won't work. We could allow a user to permit calls within associate blocks but that feels dangerous and very hard to test. I think therefore that the only remaining option is to create temporary variables and that means we have to be able to determine their type.

arporter:
I'm going to close this for now but keep the branch.

PR #2170, opened 2023-06-11
Fix bug in datatypes.py (closes #2169)

### LINKED ISSUE #2169: datatype.py should include Declaration_Construct from Fortran2003
`datatype.py` currently included Declaration_Construct from Fortran2008. This is incorrect as it is declared in Fortran2003. This needs to be changed as this will not work when branch fparser PR [417](https://github.com/stfc/fparser/pull/417) goes onto master.

rupertford:
There was a small bug in datatypes.py that is triggered by an update to fparser that is about to go on master. This PR fixes the problem.


rupertford:
Anyone can review this but I'll put down @arporter and @sergisiso

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -193,13 +193,13 @@ def type_text(self):
         # pylint:disable=import-outside-toplevel
         from fparser.common.readfortran import FortranStringReader
         from fparser.common.sourceinfo import FortranFormat
-        from fparser.two import Fortran2008
+        from fparser.two import Fortran2003
```
arporter, line 196: I'm not sure whether or not we should make classes that are common to both F08 and F03 available via either module. However, that's not something to worry about here.

arporter (approved):
Small fix which works with latest fparser. No need to update docs.
Tests all OK with compilation, as are examples.

Issue #1312, opened 2021-06-24
Pass read-only arrays of scalars to kernels


TeranIvy:
We have had repeated requests for passing arrays of scalars to kernels. The design is outlined below.

***Note:** The code in the examples was successfully tested with Intel 19.0.1 and Gnu 11.2.0 compilers.*

### Kernel layer

Full example: [scalar_arrays_kernel_mod.f90.txt](https://github.com/user-attachments/files/15925584/scalar_arrays_kernel_mod.f90.txt)

#### Metadata and argument access

- As for scalars, only read-only user-defined scalar arrays are allowed.
- Keyword to denote the scalar arrays is ~~`GH_ARRAY`~~ `GH_SCALAR_ARRAY`. We also need to specify the number of ranks as an integer. ~~We broadly agreed on `NRANKS` as a keyword for ranks (`rank` is a Fortran intrinsic so cannot be used here).~~ The new keyword for scalar array will need to be added to `argument_mod.F90` where LFRic stores metadata IDs.
- An example of ~~suggested~~ metadata design updated after the metadata review that compiles is as follows:
```
     type(arg_type), dimension(3) :: meta_args = (/          &
          arg_type(GH_SCALAR_ARRAY, GH_REAL,    GH_READ, 2), &
          arg_type(GH_SCALAR_ARRAY, GH_LOGICAL, GH_READ, 1), &
          arg_type(GH_SCALAR_ARRAY, GH_INTEGER, GH_READ, 4)  &
          /)
```
***Note:** We are not allowing kernels with no fields so this example is just for illustration.*

#### Arguments

- For every scalar array, pass:
  - ~~An `integer` scalar of kind `i_def` for the number of ranks, `nranks_<array_name>`;~~ *This is not required as it is read from the metadata.*
  - A rank-1 `integer` array of kind `i_def` and size `nranks_<array_name>` containing the upper bounds for each rank, `dims_<array_name>` (lower bound is assumed to be 1 as this is how Fortran passes array slices to subroutines by default).
- Pass array of data type and kind as inferred from the metadata.

The argument list for the metadata above would look something like:
```
  subroutine scalar_arrays_code(dims_rarray, real_array,    &
                                dims_larray, logical_array, &
                                dims_iarray, integer_array)
```
and the declarations
```
    integer(i_def), intent(in), dimension(2) :: dims_rarray
    integer(i_def), intent(in), dimension(1) :: dims_larray
[...]

PR #2173, opened 2023-06-12
#1312: metadata for read-only arrays

mo-lottieturner:
.


mo-lottieturner:
#1312

mo-lottieturner:
tests all green except codecov

mo-lottieturner:
Just documentation left to do

mo-lottieturner:
just did a double check of pycodestyle and pylint, after noticing the above modifier line had slipped through - pycodestyle clear in `src/psyclone`, but pylint has quite a few 'Access to protected member' in the test files - should I ignore these?
```
tests/domain/lfric/kernel/array_arg_metadata_test.py:78:36: W0212: Access to a protected member _get_metadata of a client class (protected-access)

tests/domain/lfric/kernel/common_meta_arg_metadata_test.py:126:11: W0212: Access to a protected member _datatype of a client class (protected-access)
tests/domain/lfric/kernel/common_meta_arg_metadata_test.py:127:11: W0212: Access to a protected member _access of a client class (protected-access)
tests/domain/lfric/kernel/common_meta_arg_metadata_test.py:158:11: W0212: Access to a protected member _datatype of a client class (protected-access)
tests/domain/lfric/kernel/common_meta_arg_metadata_test.py:159:11: W0212: Access to a protected member _access of a client class (protected-access)
tests/domain/lfric/kernel/common_meta_arg_metadata_test.py:299:11: W0212: Access to a protected member _datatype of a client class (protected-access)
tests/domain/lfric/kernel/common_meta_arg_metadata_test.py:303:11: W0212: Access to a protected member _access of a client class (protected-access)

tests/domain/lfric/lfric_array_mdata_test.py:87:16: W0212: Access to a protected member _inits of a client class (protected-access)
tests/domain/lfric/lfric_array_mdata_test.py:89:8: W0212: Access to a protected member _init_array of a client class (protected-access)
tests/domain/lfric/lfric_array_mdata_test.py:160:17: W0212: Access to a protected member _inits of a client class (protected-access)
tests/domain/lfric/lfric_array_mdata_test.py:169:8: W0212: Access to a protected member _init_scalar of a client class (protected-access)
```

arporter:
> but pylint has quite a few 'Access to protected member' in the test files - should I ignore these?

For the tests, `pylint` should be run in the `tests` directory. That way it picks up some test-specific settings (one of which is to turn off such warnings).

doc/user_guide/dynamo0p3.rst
```diff
@@ -1343,6 +1378,15 @@ checks (when generating the PSy layer) that any kernels which read
 operator values do not do so beyond the level-1 halo. If any such
 accesses are found then PSyclone aborts.
 
+.. _lfric-array-sizes:
+
+Array sizes
```
arporter, line 1383: I think this should be included in the same example where the fourth and fifth entries for fields and operators are described.

Also, I may be missing something here but why `NRANKS*n`? I think that is confusing as it mixes the notation used for a field vector with something different. Can we not just have "n"? I suppose we could have `GH_ARRAY*n` by analogy with a field vector and do away with this extra entry altogether. In fact, maybe it should be `GH_SCALAR*n` as that really does then line up with the way we do field vectors?

doc/user_guide/dynamo0p3.rst
```diff
@@ -112,13 +116,13 @@ Please see the :ref:`algorithm-layer` section for a description of the
...
 a particular data structure. These data structures are represented by the
-five LFRic (Dynamo 0.3) API argument types: :ref:`scalar <lfric-scalar>`,
-:ref:`field <lfric-field>`, :ref:`field vector <lfric-field-vector>`,
-:ref:`operator <lfric-operator>` and :ref:`column-wise operator
-<lfric-cma-operator>`. All of them except the field vector are
-represented in the above example. ``qr`` represents a quadrature object
-which provides information required by a kernel to operate on fields
-(see section :ref:`dynamo0.3-quadrature` for more details).
+six LFRic (Dynamo 0.3) API argument types: :ref:`scalar <lfric-scalar>`,
+:ref:`array <lfric-array>`,:ref:`field <lfric-field>`, :ref:`field vector
```
arporter, line 120: This isn't rendering correctly for me:
image

doc/user_guide/dynamo0p3.rst
```diff
@@ -423,6 +437,20 @@ associated kernel metadata description and their precision:
 +--------------------------+---------------------------------+-----------+
```
arporter, line 437: Please mention arrays at L409.

doc/user_guide/dynamo0p3.rst
```diff
@@ -423,6 +437,20 @@ associated kernel metadata description and their precision:
...
 | LOGICAL(L_DEF)           | GH_SCALAR, GH_LOGICAL           | L_DEF     |
 +--------------------------+---------------------------------+-----------+
+| REAL(R_DEF)              | GH_ARRAY, GH_REAL               | R_DEF     |
```
arporter, line 440: I think these new lines can be merged with the corresponding lines for scalars by altering the latter to say "GH_SCALAR/GH_ARRAY, GH_INTEGER" etc. In fact, once we've discussed my suggestion of `GH_SCALAR*n` these changes might not be required at all.

arporter (changes requested):
Thanks for this @mo-lottieturner. I should have said, but I've come to realise that for something like this it's always best to just do the documentation first and then, once we're all agreed, proceed to the implementation.

As it is, I've only looked at the documentation so far. I'm not too sure about the form of the new metadata as it seems to me to be a bit more complex than is necessary (although I'm happy to be corrected). I think there are two choices, extend `GH_SCALAR` so that we also support `GH_SCALAR*n`. This has the advantage that it follows what we do for field vectors and is therefore intuitive. It also reduces the confusion about why a field is not an array. Failing that, my second suggestion would be to simplify the fourth metadata entry for a `GH_ARRAY` so that it is just `nranks` - there's no need to use the `*` notation.

Happy to discuss with all interested parties :-)

doc/user_guide/dynamo0p3.rst
mo-lottieturner: missing space issue, should be fixed

mo-lottieturner:
For posterity: If memory serves, we agreed to stick with `GH_ARRAY`, but that `NRANKS*n` can be switched to just `n`

doc/user_guide/dynamo0p3.rst
mo-lottieturner: For posterity: the reason behind not using `GH_ARRAY*n` (or `GH_SCALAR*n`) is because `n` here does not represent the number of rows in a vector, but the number of dimensions in an array (so `n=1` is a vector, `n=2` is a 2d-array, etc)

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -0,0 +1,159 @@
...
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author L. Turner, Met Office
+
+'''Module containing the ArrayArgMetadata class which captures the metadata
+associated with an array argument. Supports the creation, modification
+and Fortran output of an Array argument.
```
arporter, line 38: Sorry to be a pain but I keep thinking we need to qualify our mentions of  'array' to make it clear it is something special (after all, a field boils down to an array and I think it could lead to confusion). Therefore, I'm wondering whether `ScalarArray` (and so `GH_SCALAR_ARRAY`) would be a better name for this concept - it signals that we really do mean something other than a field?
Something to discuss with at least @TeranIvy and possibly others.

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -0,0 +1,159 @@
...
+    # The name used to specify an array argument in LFRic metadata.
+    form = "gh_array"
+    # The relative positions of LFRic metadata. Metadata for an array
+    # argument is provided in the following format 'arg_type(form,
+    # datatype, access, function_space)'. Therefore, for example, the
+    # index of the form argument (form_arg_index) is 0.
+    form_arg_index = 0
+    datatype_arg_index = 1
+    access_arg_index = 2
+    array_size_arg_index = 3
```
arporter, line 64: I suggest replacing "size" with "rank" here.

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -0,0 +1,159 @@
...
+    # The name to use for any exceptions.
+    check_name = "array"
+    # The number of arguments in the language-level metadata (min and
+    # max values).
+    nargs = (4)
```
arporter, line 69: Should be `(4, 4)` perhaps but I guess I'll find out later.

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -0,0 +1,159 @@
...
+from psyclone.domain.lfric.kernel.scalar_arg_metadata import ScalarArgMetadata
+
+
+class ArrayArgMetadata(ScalarArgMetadata):
+    '''Class to capture LFRic kernel metadata information for an array
+    argument.
+
+    :param str datatype: the datatype of this array (GH_INTEGER, ...).
+    :param str access: the way the kernel accesses this array (GH_READ).
+    :param str function_space: the function space that this array is \
```
arporter, line 51: This should be documenting `array_ndims`.

doc/user_guide/dynamo0p3.rst
```diff
@@ -1388,6 +1410,15 @@ checks (when generating the PSy layer) that any kernels which read
...
+.. _lfric-array-sizes:
+
+Array sizes
+^^^^^^^^^^^
+
+The size of a :ref:`scalar array <lfric-array>` is described by ``NRANKS*<n>``,
+where *n > 0* is the number of Fortran ranks representing the dimension of the
+array.
+
 .. _lfric-function-space:
```
arporter, line 1422: The rules for the arguments to a Kernel subroutine also need to be updated. Presumably, an Array argument will need to be followed by `ndims` and then possibly an integer, rank-1 array of size `ndims` containing the size of each dimension? Or, will `ndims` always be a literal (known) value such that we can each of the dimension sizes as an explicit subroutine argument? (This will allow us to explicitly declare the shape of the `array` argument in the PSy layer which is something we always aim to do.)

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -0,0 +1,159 @@
...
+        :rtype: str
+        '''
+        return self._array_ndims
+
+    @array_ndims.setter
+    def array_ndims(self, value):
+        '''
+        :param str value: set the function space to the
+            specified value.
+        '''
```
arporter, line 143: Please document (after a blank line) the various :raises:

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def fortran_string(self):
+        '''
+        :returns: the metadata represented by this class as Fortran.
+        :rtype: str
+        '''
+        return (f"arg_type({self.form}, {self.datatype}, {self.access}, "
+                f"{self.array_ndims})")
+
+    @staticmethod
```
arporter, line 104: Unless I'm missing something, this whole method can be removed as it's identical to the one in the class from which this one inherits.

arporter (changes requested):
Thanks for those changes @mo-lottieturner. The docs are looking pretty good now although I realise that the section describing the detailed rules for the subroutine arguments of a kernel also need to be updated.
I've now gone through the code changes but not the test files. Generally speaking it looks good although you'll see that I'm suggesting (another?) name change for clarity as I keep getting confused about distinguishing an 'array' from a 'field'. We'll need to agree any name change with at least @TeranIvy and whoever else she thinks might care!

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
TeranIvy: I think you have a point there, @arporter, as we can have arrays of fields.

doc/user_guide/dynamo0p3.rst
TeranIvy: Yes, we need to pass `ndims` and rank-1 array containing sizes. The proposed design is in the description of #1312, together with examples of code for Kernel, Algorithm and PSy layers (all of them are compilable as far as I remember).

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
TeranIvy: Discussed this with @mo-lottieturner and she's agreed to change it.

mo-lottieturner: does this include changing the file name from array_arg_metadata to scalar_array_arg_metadata?

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -270,43 +275,83 @@ def _validate_vector_size(self, separator, arg_type):
...
+        try:
+            array_ndims = int(arg_type.args[3])
+        except ValueError as err:
+            raise ParseError(
+                f"In the LFRic API, the array notation must be in "
+                f"the format 'n' where 'n' is an integer, "
+                f"but '{arg_type.args[3]}' was found in "
+                f"'{arg_type}'.") from err
+
+        # ... or it is less than 1 (1 is the default for all fields)...
```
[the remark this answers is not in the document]
mo-lottieturner: This section was copypasted from _validate_vector_size - the same comment is there (less than 2, 1 the default). Not sure I follow it either - should I remove the brackets section?

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -599,7 +644,65 @@ def _init_scalar(self, arg_type):
...
                 f"'{self._data_type}' data type was found in '{arg_type}'.")
 
-        # Scalars don't have vector size
+        # Scalars don't have vector size or array size
+        self._vector_size = 0
+        self._array_ndims = 0
+
+    def _init_array(self, arg_type):
+        '''
+        Validates metadata descriptors for scalar array arguments and
```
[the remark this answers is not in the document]
mo-lottieturner: as in ScalarArray instead of scalar array?

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -599,7 +644,65 @@ def _init_scalar(self, arg_type):
...
+        initialises scalar array argument properties accordingly.
+
+        :param arg_type: LFRic API scalar array argument type.
+        :type arg_type: :py:class:`psyclone.expression.FunctionVar`
+
+        :raises InternalError: if argument type other than an array is
+                               passed in.
+        :raises ParseError: if there are not exactly 4 metadata arguments.
+        :raises InternalError: if an array argument has an invalid data type.
+        :raises ParseError: if array arguments do not have read-only access.
```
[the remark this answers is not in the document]
mo-lottieturner: not sure this is right - think it should be 'if the array argument has an invalid access type'

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
mo-lottieturner: It feels inconsistent for scalar arrays to be the only ones to use camel case

doc/user_guide/dynamo0p3.rst
```diff
@@ -1242,6 +1273,8 @@ the :ref:`LFRic fields <lfric-field>`):
 +========================+=================================+
```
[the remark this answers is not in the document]
mo-lottieturner: I'm not clear on what needs changing where for this comment, sorry

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
mo-lottieturner: this is an artefact of distinguishing between `NRANKS*<n>` and `n` - I'll have a look at what else might need changing of this sort

mo-lottieturner: I've changed this to `array_ndims_arg_index` to match with everything else that refers to this argument, including the property

doc/user_guide/dynamo0p3.rst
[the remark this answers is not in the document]
mo-lottieturner: I think I've corrected what needs correcting here

src/psyclone/domain/lfric/kernel/common_meta_arg_metadata.py
```diff
@@ -200,6 +202,28 @@ def get_vector_length(cls, fparser2_tree):
         vector_length = components[1].strip()
         return vector_length
 
+    @classmethod
+    def get_array_ndims(cls, fparser2_tree):
+        '''Retrieves the array ndims metadata value found within the
+        supplied fparser2 tree and checks that it is valid.
```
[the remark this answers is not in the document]
mo-lottieturner: The validation is done in the setter - should some or all of these be moved to the getter?

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
mo-lottieturner: @arporter could I have an answer to this please?

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
mo-lottieturner: @arporter and an answer to this too please

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
arporter: If the class name has changed then yes please.

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
arporter: In this case, I think "ScalarArray" is what we want as this indicates it is a special, named quantity in the API.

mo-lottieturner:
@arporter not all review comments have been addressed yet, but as I'm handing this over to Alistair, it would be helpful for you to mark the addressed comments as resolved, if you have a moment?

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
```diff
@@ -101,59 +100,53 @@ def fortran_string(self):
...
+        '''Retrieves the array ndims metadata value found within the
+        supplied fparser2 tree and checks that it is valid.
+
+        :param fparser2_tree: fparser2 tree capturing the required metadata.
+        :type fparser2_tree: :py:class:`fparser.two.Fortran2003.Part_Ref`
+
+        :returns: the array ndims value extracted from the fparser2 tree.
+        :rtype: str
+
+        :raises TypeError: if the array ndims is not a string.
```
arporter, line 130: This method only actually raises ValueErrors. Also, I think this checking should probably be done in the array_ndims setter. I'm not sure whether we should duplicate - we'd need to look at what we do for other properties.

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
arporter: I think we just remove the brackets section.

arporter:
OK @mo-lottieturner and @mo-alistairp, I've gone through and resolved everything that has been done.

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
mo-alistairp: Hi @arporter, I've had a little look into this and I think I've understood (apologies if not!). 

I agree that it makes more sense to validate this in the setter rather than the getter. I had a little look, and it seems in both `field_vector_arg_metadata.py` and `evaluator_targets_metadata.py` the setter validates the data (although admittedly neither seem to have a specific getter function from the fparser2 tree). 

When you talk about duplication, do you mean validating the data in both the getter and setter?

mo-alistairp: As an extension to what I said above, I;ve just noticed that in `inter_grid_arg_metadata.py` there is some checking in both `@mesh_arg.setter` and in the static `get_mesh_arg` (which gets the value from the fparser2 tree), which seems more similar to our case here. So I'm not enirely sure what the best course of action is.

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -599,7 +644,65 @@ def _init_scalar(self, arg_type):
...
+                f"{nargs_array} arguments if its first argument is of "
+                f"{const.VALID_ARRAY_NAMES} type, but found {self._nargs} in "
+                f"'{arg_type}'.")
+
+        # Check whether an invalid data type for an array argument is passed
+        # in. Valid data types for arrays are valid data types in LFRic API.
+        if self._data_type not in const.VALID_ARRAY_DATA_TYPES:
+            raise InternalError(
+                f"Expected one of {const.VALID_ARRAY_DATA_TYPES} as the "
+                f"array data type but got '{self._data_type}'.")
```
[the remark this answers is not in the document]
mo-alistairp: Can I check, with this and similar comments (s/size/rank/) are you wanting me to simply replace those exactly?

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
arporter: In that case, I think we live with the duplication of the validation code and do it in both. Perhaps just a little utility method unless that's over the top.

mo-alistairp: Ok thanks, I'll try and come up with a good implementation

mo-alistairp: Simple code duplication seemed like a better solution for this as the tests aren't both raising the TypeError. If you disagree, I'm happy to change it

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -713,6 +820,19 @@ def vector_size(self):
...
         return self._vector_size
 
+    @property
+    def array_ndims(self):
+        '''
+        Returns the array size of the argument. This will be 1 if ``*n``
+        has not been specified for all argument types except scalars
+        (their array size is set to 0).
+
+        :returns: array size of the argument.
```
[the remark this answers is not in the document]
mo-alistairp: Done

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -713,6 +820,19 @@ def vector_size(self):
         '''
         return self._vector_size
 
+    @property
+    def array_ndims(self):
+        '''
+        Returns the array size of the argument. This will be 1 if ``*n``
```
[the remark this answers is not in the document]
mo-alistairp: Done

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
mo-alistairp: Done

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
mo-alistairp: Done

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
mo-alistairp: I believe I've changed all the occurances of array to ScalarArray

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
mo-alistairp: Removed the bracketed section

src/psyclone/domain/lfric/kernel/common_meta_arg_metadata.py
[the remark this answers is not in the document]
mo-alistairp: Added testing to the getter and setter in `scalar_array_arg_metadata.py`

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
mo-alistairp: I believe I've changed all occurances of array to ScalarArray and I've changed file names

mo-alistairp:
I've addrsesed all the open comments. It's possible I've been a little overzealous with the changes from array to ScalarArray, so just let me know. Otherwise, this is ready for review

doc/user_guide/dynamo0p3.rst
```diff
@@ -1388,6 +1410,15 @@ checks (when generating the PSy layer) that any kernels which read
 operator values do not do so beyond the level-1 halo. If any such
 accesses are found then PSyclone aborts.
 
+.. _lfric-array-sizes:
+
+Array sizes
+^^^^^^^^^^^
+
+The size of a :ref:`scalar array <lfric-array>` is described by ``NRANKS*<n>``,
```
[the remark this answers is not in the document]
arporter: This just needs to be updated with an example now.

src/psyclone/domain/lfric/kernel/common_meta_arg_metadata.py
```diff
@@ -186,7 +187,7 @@ def get_vector_length(cls, fparser2_tree):
         :returns: the vector length value extracted from the fparser2 tree.
         :rtype: str
```
arporter: The changes to this file could be reverted entirely. (`git checkout master ./common_meta_arg_metadata.py` I think.)

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
```diff
@@ -0,0 +1,172 @@
...
+from psyclone.domain.lfric.kernel.common_arg_metadata import CommonArgMetadata
+from psyclone.domain.lfric.kernel.scalar_arg_metadata import ScalarArgMetadata
+
+
+class ScalarArrayArgMetadata(ScalarArgMetadata):
+    '''Class to capture LFRic kernel metadata information for a ScalarArray
+    argument.
+
+    :param str datatype: the datatype of this ScalarArray (GH_INTEGER, ...).
+    :param str access: the way the kernel accesses this Scalar Array (GH_READ).
```
arporter, line 53: "... (e.g. GH_READ)."

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
```diff
@@ -0,0 +1,172 @@
...
+    :param str array_ndims: the rank (number of dimensions) of this
+                            ScalaArrray.
```
arporter: "ScalarArray"

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
```diff
@@ -0,0 +1,172 @@
...
+
+    '''
+    # The name used to specify a ScalarArray argument in LFRic metadata.
+    form = "gh_scalar_array"
+    # The relative positions of LFRic metadata. Metadata for a ScalarAarray
+    # argument is provided in the following format 'arg_type(form,
+    # datatype, access, function_space)'. Therefore, for example, the
```
arporter: A ScalarArray doesn't have a function_space I think?

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
```diff
@@ -0,0 +1,172 @@
...
+        :rtype: str
+        '''
+        return (f"arg_type({self.form}, {self.datatype}, {self.access}, "
+                f"{self.array_ndims})")
+
+    @property
+    def array_ndims(self):
+        '''
+        :returns: the number of dimensions for this ScalarArray argument.
+        :rtype: str
```
arporter: Since the setter (below) stores the value as an `int`, this needs updating to say it returns an int. Please could you do this the new way, i.e.:

```python
def array_ndims(self) -> int:
```
and then you can remove the `rtype` line entirely.

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
```diff
@@ -0,0 +1,172 @@
...
+        separately).
+
+        :param fparser2_tree: fparser2 tree containing the metadata
+            for this argument.
+        :type fparser2_tree: :py:class:`fparser.two.Fortran2003.Part_Ref` |
+            :py:class:`fparser.two.Fortran2003.Structure_Constructor`
+
+        :returns: a tuple containing the datatype, access and array ndims
+            metadata.
+        :rtype: Tuple[str, str, str]
```
arporter: Again (see below), ndims is stored as an int. Please could you update using typehinting:
```python
def _get_metadata(cls, fparser2_tree: Union[Fortran2003.Part_Ref, Fortran2003.Structure_Constructor]) -> Tuple[str, str, int]:
```
and then remove the :type and :rtype lines of the docstring.

arporter: (you may need to add an import of Union and Tuple from the `typing` module.)

src/psyclone/domain/lfric/kernel/array_arg_metadata.py
arporter: That looks fine, thanks.

src/psyclone/domain/lfric/lfric_arg_descriptor.py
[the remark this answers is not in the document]
arporter: Having made my way through this review I've now remembered that read-only *is* the only valid access type so this should read "if the array argument does not have read-only access".

doc/user_guide/dynamo0p3.rst
arporter: @mo-alistairp this one is still outstanding - it's https://psyclone.readthedocs.io/en/latest/user_guide/lfric.html#rules-for-general-purpose-kernels that needs to be updated.

arporter (changes requested):
Thanks for taking this on @mo-alistairp. It's getting there now.

Somewhere in my comments I requested that you add an "e.g." before GH_READ but I now realise that GH_READ is the only permitted access for a ScalarArray so please don't  make that change. In fact, possibly we should emphasise that restriction instead.

I've realised that we don't have an example of the use of this new kernel argument but I suggest you do that as a separate issue/pr so that we can get this one over the line. Perhaps add it as something to do in #1312 and make this a 'towards' PR.

I'll fire off the integration tests just to be on the safe side.

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
mo-alistairp: I'm not sure it does actually set it as an `int` - `value` is what gets assigned to `self._array_ndims` which I believe is a `str` containing an `int`. I think this means that the later comment should also be type `Tuple[str,str,str]` still. Of course if you'd rather it was just stored as an `int`, I can change that too?

Has typehinting become the PSyclone standard then?

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
mo-alistairp: I've tried to avoid using the typing module, as I believe it's been deprecated since Python 3.9. However, I wasn't able to use the native union operator for the fparser2_tree. Are you happy with me using the native tuple typehinting or would you rather I used the typing module for consistency?

Similar to above, I believe the ndims is stored as a `str` containing an `int`. I can change that instead if you'd rather?

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
mo-alistairp: I've ignored this, as you requested

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
mo-alistairp: Done

src/psyclone/domain/lfric/kernel/common_meta_arg_metadata.py
mo-alistairp: Reverted the changes

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
mo-alistairp: I believe you're right and instead that should have been the number of dimensions of the array - I've changed this.

doc/user_guide/lfric.rst
```diff
@@ -2062,6 +2062,9 @@ conventions, are:
       freedom for the ``to`` and ``from`` function spaces,
       respectively. Again the intent is determined
       from the metadata (see :ref:`lfric-api-meta-args`).
+   5) If the current entry is a ScalarArray then include the Fortran
+      variable in the argument list. The ScalarArray must be denoted
+      with intent ``in`` to match its read-only nature.
```
arporter: We will also need to pass in the size of each of the n dimensions of this array. I suggest n scalar arguments immediately following the ScalarField argument if @TeranIvy agrees. Is there an example of a kernel of this type somewhere?

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
arporter: It doesn't seem to have changed?

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
arporter: Oh yes, you're quite right - I failed to spot that, having converted it to an integer, we still store the str value. I think it would be more natural to store an `int` so please could you update this (although, if this causes significant problems elsewhere then we can leave it).

src/psyclone/domain/lfric/kernel/scalar_array_arg_metadata.py
arporter: Since we still officially support 3.9, using the `typing` module is fine I think. If you can get away without it though then that's even better.

doc/user_guide/lfric.rst
```diff
@@ -1399,6 +1425,15 @@ checks (when generating the PSy layer) that any kernels which read
...
 accesses are found then PSyclone aborts.
 
+.. _lfric-array-sizes:
+
+Array sizes
+^^^^^^^^^^^
+
+The size of a :ref:`scalar array <lfric-array>` is described by ``<n>``,
+where *n > 0* is the number of Fortran ranks representing the dimension of the
+array.
```
arporter: Please provide an example here, e.g:
```
...array, e.g. a logical, scalar array of rank three would be specified as:

::

    arg_type(GH_SCALAR_ARRAY, GH_LOGICAL, GH_READ, 3)

doc/user_guide/lfric.rst
arporter: Apologies, I found a previous comment where @TeranIvy spelt-out what needed to be done. The specification is here https://github.com/stfc/PSyclone/pull/2173/files - for each ScalarArray argument we need to pass an additional integer array holding the size of each dimension of the array.

mo-alistairp: I was using that doumentation, but just misread - so apologies from me too! That should be updated now.

doc/user_guide/lfric.rst
mo-alistairp: I can't seem to respond to each individual comment, but I believe I've addressed each of your requests. I've -

- changed it over so it saves `array_ndims` as an `int`, and changed the relavent tests and documentation to reflect that 
- updated a couple of other functions to have typehinting 
- added the example you requested
- fixed a couple of typos

arporter: Thanks.

PR #2190, opened 2023-06-23
(Closes #1431) Initial changes to add the canonical loop trans abstract class

### LINKED ISSUE #1431: Add OMPLoopTrans parent class (subclass of ParallelLoopTrans)
The OpenMP loop transformations supported in PSyclone (`OMPTaskloopTrans` and `OMPLoopTrans`) should both check that the loop is of the canonical loop form (https://www.openmp.org/spec-html/5.0/openmpsu40.html#x63-1260002.9.1)

For Fortran style loops this involves (written by Sergi in #1415 ):

- Only BinaryOperation.[ADD, SUB, MUL] + Literals + References are valid in in the incr
- Loop invariant with respect to the outermost associated loop. As discussed the other day I think for a single loop in Fortran, the incr is loop invariant by definition, but I don't think this is true for nested loops then and could be checked in the validate.

An additional check (which we can't always do if the loop increment is not a constant, at which point it isn't PSyclone's issue I think) is that the loop increment is not 0.

I think this should sit as a different class (and not part of `OMPLoopTrans`) as `OMPTaskloopTrans` also uses this, but I don't believe they should be subclasses - perhaps something like `OMPCanonicalLoopTrans`.

One other slight thing about the name `OMPLoopTrans` in general - OpenMP has a loop construct (https://www.openmp.org/spec-html/5.0/openmpsu44.html) - I think its unlikely PSyclone would support this construct (as I'm currently unclear what the use-case of it is), the docstring for OMPLoopTrans should probably also be adjusted from "Adds an OpenMP directive to a loop" to something like "...OpenMP do (or for) directive to a loop"

LonelyCat124:
Having some issues with the testing here, so I wanted to make this PR for potential discussion on why tests are failing if I can't work it out.

The canonical loop validation testing is not yet implemented either.

### DIFF
--- a/src/psyclone/psyir/transformations/omp_canonical_loop_trans.py
+++ b/src/psyclone/psyir/transformations/omp_canonical_loop_trans.py
@@ -1,0 +1,254 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2017-2022, Science and Technology Facilities Council.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+#
+# * Neither the name of the copyright holder nor the names of its
+#   contributors may be used to endorse or promote products derived from
+#   this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT                                                                                                        # LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: A. B. G. Chalk, STFC Daresbury Lab
+
+''' Base OpenMP Loop transformation to validate directives that require
+canonical loop form.'''
+
+import abc
+
+from psyclone.core import VariablesAccessInfo, Signature
+from psyclone.psyir.nodes import (
+    Reference, BinaryOperation, Literal)
+from psyclone.psyir.symbols import DataSymbol, INTEGER_TYPE, ScalarType
+from psyclone.psyir.transformations.parallel_loop_trans import \
+    ParallelLoopTrans
+
+class OMPCanonicalLoopTrans(ParallelLoopTrans, metaclass=abc.ABCMeta):
+
+    def _check_valid_format(self, node, loop_var):
+        '''
+        TODO
+        '''
+        # We know node contains a Reference to loop_var to reach this call.
+
+        if isinstance(node, Reference):
+            return True
+
+        # All other allowed cases are binary operations
+        if not isinstance(node, BinaryOperation):
+            return False
+        # We allow var + a2, a2 + var
+        if (node.operator == BinaryOperation.Operator.ADD and
+                isinstance(node.children[0], Reference) and
+                isinstance(node.children[1], Reference)):
+            return True
+
+        # We allow var - a2 and a2 - var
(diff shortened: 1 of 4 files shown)


LonelyCat124:
Ok, I missed some cases I didn't think about.
In one of the Nemo examples, somewhere we get a NemoLoop with a start variable that is formed of:
`BinaryOperation(LBOUND(zwx, 2))`

The loop is created from:
```
zwx(:,:,jpk) = 0.e0
```

and transformed by PSyclone to:
```
do idx = LBOUND(zwx, 2), UBOUND(zwx, 2), 1
  do idx_1 = LBOUND(zwx, 1), UBOUND(zwx, 1), 1
    zwx(idx_1,idx,jpk) = 0.e0
  enddo
enddo
```

This looks to be ok w.r.t canonical loop form, but is currently not handled correctly.
I think I need to be smarter checking invariance somehow, since the `LBOUND` and `UBOUND` statements should not need to worry about the array they act upon being written to, and is a common use-case.
I could ignore BinaryOperations containing UBOUND/LBOUND explicitly, but I'm not sure if that is the correct approach.

arporter:
> I think I need to be smarter checking invariance somehow, since the `LBOUND` and `UBOUND` statements should not need to worry about the array they act upon being written to, and is a common use-case.

This is something we've hit elsewhere. `Call` nodes now have the `is_pure` and `is_elemental` properties. @sergisiso has suggested we add another one, something like `enquiry_fn` that tells us that the routine doesn't actually access/care about the data in its argument. That could then naturally be handled in the dependence analysis rather than having to special case.

src/psyclone/psyir/transformations/omp_canonical_loop_trans.py
```diff
@@ -0,0 +1,254 @@
...
+
+        # Check for a1*var - a2 or var*a1 - a2
+        if (node.operator == BinaryOperation.Operator.SUB and
+                isinstance(node.children[1], Reference) and
+                node.children[1].symbol != loop_var and
+                isinstance(node.children[0], BinaryOperation) and
+                node.children[0].operator == BinaryOperation.Operator.MUL and
+                (node.children[0].children[0].symbol == loop_var or
+                 node.children[0].children[1].symbol == loop_var)):
+            return True
```
hiker, line 121: These tests still don't feel right. Too many very similar tests, and I am not even sure that they are complete (what about -a1*var-a2 or +a1*var+a2, which from memory adds another UnaryOperation node).
SymPy offers a polynomial class, which gives you access to the coefficients, and for the example above at least it will convert them all to a similar structure:
```
>>> sp.Poly(a1 * var - a2, var).all_coeffs()
[a1, -a2]
>>> sp.Poly(a2 - a1 * var,var).all_coeffs()
[-a1, a2]
```
etc. Unfortunately, it does still first simplify the expressions, so
```
>>> sp.Poly(a2 - a1 * var + 2*var,var).all_coeffs()
[2 - a1, a2]
```
Could we perhaps in a case like this rewrite the PSyIR with the equivalent expression (psyad is already doing this when expanding terms). Or you could additionally count the number of references and symbols in the original PSyIR, which should be less than or equal to 4?

hiker:
> > I think I need to be smarter checking invariance somehow, since the `LBOUND` and `UBOUND` statements should not need to worry about the array they act upon being written to, and is a common use-case.
> The loop is created from:
> 
> ```
> zwx(:,:,jpk) = 0.e0
> ```
> 
> and transformed by PSyclone to:
> 
> ```
> do idx = LBOUND(zwx, 2), UBOUND(zwx, 2), 1
>   do idx_1 = LBOUND(zwx, 1), UBOUND(zwx, 1), 1
>     zwx(idx_1,idx,jpk) = 0.e0
>   enddo
> enddo
> ```
> 
> This looks to be ok w.r.t canonical loop form, but is currently not handled correctly. I think I need to be smarter checking invariance somehow, since the `LBOUND` and `UBOUND` statements should not need to worry about the array they act upon being written to, and is a common use-case. I could ignore BinaryOperations containing UBOUND/LBOUND explicitly, but I'm not sure if that is the correct approach.

Not sure if I understand the exact problem - in the example eg1 this loop does not even reach your new code, since the loop type is `unknown`, it is skipped because the user script only omp-parallelises `levels` loops:
```
       for loop in invoke.schedule.walk(Loop):
            if loop.loop_type == "levels":

```
I somehow feel that I missed something ;)

hiker:
> This is something we've hit elsewhere. `Call` nodes now have the `is_pure` and `is_elemental` properties. @sergisiso has suggested we add another one, something like `enquiry_fn` that tells us that the routine doesn't actually access/care about the data in its argument. That could then naturally be handled in the dependence analysis rather than having to special case. EDIT: in fact, I think the dependence analysis has a flag that can control what it does about these statements. @hiker will know if he's awake at the moment.
> 
> See e.g. https://psyclone-ref.readthedocs.io/en/latest/_static/html/classpsyclone_1_1psyir_1_1tools_1_1dependency__tools_1_1DependencyTools.html#a619bd6530f6be092fc4c873b017e332c

Yes, anything that uses `VariablesAccessInfo` will by default ignore shape-related accesses, and so will not be reported (you have to explicitly specify a flag if you want these accesses reported).

src/psyclone/psyir/transformations/omp_canonical_loop_trans.py
LonelyCat124: I looked again and my interpretation of the OpenMP standard was also wrong, in that `a1` and `a2` simply need to be "loop invariant integer expressions", and not simply variables.
I think I should have already checked that any `Reference` objects in these terms are loop invariant in `validate` before reaching this function, so what actually needs to be done is:
1. Check we have some structure that could be handled as one of the allowed structures.
2. Check that the `a1` and `a2` expressions (if present) are integer expressions.
I wonder if 2 would have been handled already during input/tree construction or if we can just ignore it.
I don't have a good idea how to check the structure conforms though, maybe the sp.Poly class you recommend will work straight out the box? But I'm not clear how to use this - could you provide a quick example?

I'm trying to work out what this actually disallows, and I think any operation involving all integer values/references, and only `+-*` operations is allowed (since you could transform `x*y*var_outer` into `a1*var_outer` since `a1=x*y`. Equally division (maybe any integer-safe) operations should be ok provided they are not performed on `var_outer`. Does that make sense or have I missed something?

LonelyCat124:
The changes fix the known errors in the examples, so now the tests I had failing locally are also failing on github.
I'm unclear on the cause at all, they're all lfric tests which shouldn't have their behaviour changed as far as I'm aware?

arporter:
Closing this as has served its purpose.

Issue #2205, opened 2023-06-30
PSyclone losing some subroutines from a subroutine contains statement


LonelyCat124:
I don't have a good understanding of what is happening in this file in Socrates.

File in question is `socrates_set_diag.f90`, which contains a module containing a subroutine containing some additional subroutines.

This could therefore be related directly to #2200 - but in this case PSyclone execution is completed and outputting invalid Fortran code.

Rough code structure is
```
Module socrates_set_diag
....
contains
subroutine set_diag(...)
...
call sum_flux_channels(.., surf=diag%...)
...
contains
subroutine sum_flux_channels(...)
subroutine ... (more than 1)
end subroutine set_diag
end module
```

The output code from Psyclone outputs the subroutine without the `contains` block, so compilation fails as the compiler has no idea what these functions are.

rupertford:
I don't think we've had to deal with code that has subroutines within subroutines before, so you're probably picking up bugs associated with that.

LonelyCat124:
Yeah - it seems quite commonplace in socrates (I'd guess at least 10-15 files may have this behaviour) -  some files will compile even though the contained subroutines are lost, but then fail to link.

arporter:
It's something that happens in NEMOVAR but (AFAIK), we do successfully create CodeBlocks there. There must be something else going on here.

LonelyCat124:
I closed #2200 in favour of this.

From the original issue:
I managed to construct a mini psyclone test that replicates this behaviour:
```
def test_fail_case(fortran_reader):
    code = '''subroutine a(b, c, d)
    real b, c, d

    b = my_func(c, d)

    contains
    real function my_func(a1, a2)
    real a1, a2
    my_func = a1 * a2
    end function
    end subroutine'''
    psyir = fortran_reader.psyir_from_source(code)
```

Test fails here:
```
../../../psyir/frontend/fortran.py:79: in psyir_from_source
    psyir = self._processor.generate_psyir(parse_tree)
../../../psyir/frontend/fparser2.py:1155: in generate_psyir
    self.process_nodes(node, [parse_tree])
../../../psyir/frontend/fparser2.py:2534: in process_nodes
    psy_child = self._create_child(child, parent)
../../../psyir/frontend/fparser2.py:2597: in _create_child
    return handler(child, parent)
../../../psyir/frontend/fparser2.py:4604: in _program_handler
    self.process_nodes(file_container, node.children)
[...]

PR #2208, opened 2023-07-03
(Towards 2205) Workaround solution for subroutine contains

LonelyCat124:
This has the workaround and test for converting subroutines and functions with `contains` into `CodeBlocks`.

This does not give support for `PSyclone` to do anything with those functions however, so a longer term solution to #2205 may be more useful in the future.


arporter (commented):
Good job. This is a bug fix so no need to update docs. Testing is thorough.
Coverage is fine.
Integration tests are underway. Once/if they are green this can be merged.

arporter (approved):
Integration tests are all green.
Will proceed to merge.

PR #2211, opened 2023-07-03
use kernel names in algorithm harness code (closes 2183)

rupertford:
@DrTVockerodtMO requested that the names for fields, operators etc in the generated LFRic (algorithm layer) harness code are taken from the names used in the tangent-linear code, rather than using generic names. This should not matter if everything is automated, but at the moment manual tweaks to the generated harness code are required and it would help when modifying and/or debugging.


rupertford:

arporter (approved):
All looks good. Tests and integration tests all fine.
Changes are pylint clean.
Ref. guide is all fine.
Will proceed to merge.

PR #2212, opened 2023-07-04
(Closes #2203) Fix for do loop inside while loop polluting the while loop creation

### LINKED ISSUE #2203: PSyclone incorrectly converting `DO` statement to `Do j = ` loop resulting in compiler issue
Socrates error again, the file in question is `make_block_1.f90`

Inside the original file, there is this code loop:
```
  IF (l_range) THEN
    i = 0
    DO
      WRITE(iu_stdout, '(A)') &
        'Enter units followed by lower and upper limits and increment:'
      DO
      .... some EXIT condition
     END DO
      range_bands = NINT( (range_high - range_low) / range_inc )
      IF (i + range_bands > SpBasic%n_band) THEN
        WRITE(iu_err, '(a)') 'Specified range exceeds number of bands.'
        CYCLE
      END IF
      DO j = 1, range_bands
        i = i + 1
        ....
      END DO
      IF (i == SpBasic%n_band) EXIT
    END DO
```

As far as I understand it, this should be something like:
```
do while (.true.)
      WRITE(iu_stdout, '(A)') &
        'Enter units followed by lower and upper limits and increment:'
      do while(.true.)
     ... some exit condition
     end do
   ....
   do j = 1, range_bands, 1
      i = i + 1
......
```

However, the output from PSyclone goes very wrong, and instead starts with: `do j = 1, range_bands, 1`, which is one of the inner loops - not sure what is causing this but it seems to be confused.

This bug also occurs for `make_block_17.f90`, `read_solar_spectrum.f90`, `map_sub_bands_mod.f90`, `tidy_90.f90`, `corr_k_single.f90`, `conjugate_gradient_90.f90`, `exponent_fit_90.f90`, `optimal_k.f90`, `read_pt_line_90.f90`, `set_g_point_90.f90`, `decompose_phf_90.f90`, `scatter_average_90.f90`

LonelyCat124:
This appears to be another very small bug, but I could be wrong. 
As far as I understand it, the `Fortran2003.Nonlabel_Do_Stmt` will always contain the `Fortran2003.Loop_Control` if there is one, so I changed instead to search that instead of the full content for the Loop control. The previous code would find a `Do i = 1...` loop control block from a child loop if the loop itself is meant to be a while loop, resulting in the error found in #2203


src/psyclone/tests/psyir/frontend/fparser2_do_handler_test.py
```diff
@@ -181,3 +182,37 @@ def test_undeclared_loop_var(fortran_reader):
...
+      end if
+  end do
+
+  end subroutine'''
+
+    psyir = fortran_reader.psyir_from_source(code)
+    loops = psyir.walk((WhileLoop, Loop))
+    assert isinstance(loops[0], WhileLoop)
+    assert isinstance(loops[1], WhileLoop)
+    assert isinstance(loops[2], Loop)
```
arporter, line 218: Although this is a FE test, I think it would be good to use the backend here too and just check that the generated Fortran is correct (and compiles if poss).

arporter (changes requested):
Looks good @LonelyCat124 - thanks. I just have a request to extend the test to check that we get back the Fortran we expect.
Changes are pylint clean.
No need to update docs.

LonelyCat124:
Added the extra code check and compilation check to the test now.

arporter (approved):
Requested change has been made (thanks).
Tests and integration tests are all green.
Will proceed to merge.

PR #2222, opened 2023-07-13
Add support for intrinsics tanh(x) & float(x)

svalat:
Add support for the `tanh(x)` and `float(x)` intrinsic ops required to pass on CROCO Fortran files.


src/psyclone/psyir/nodes/operation.py
```diff
@@ -319,11 +319,11 @@ class UnaryOperation(Operation):
         # Logical Operators
         'NOT',
         # Trigonometric Operators
-        'COS', 'SIN', 'TAN', 'ACOS', 'ASIN', 'ATAN',
+        'COS', 'SIN', 'TAN', 'TANH', 'ACOS', 'ASIN', 'ATAN',
         # Other Maths Operators
         'ABS', 'CEIL',
         # Casting Operators
-        'REAL', 'INT', 'NINT'
+        'REAL', 'FLOAT', 'INT', 'NINT'
```
LonelyCat124, line 326: Is there a functional difference between `float(x)` and `real(x)` in Fortran? If not, I think the `FLOAT` operator shouldn't exist, and PSyclone should just convert `float(x)` to `real(x)` internally, and output `real(x)` since `FLOAT` is an extension.
From my reading `REAL(A) is converted to a default real type if A is an integer or real variable.` is the same as `FLOAT(A)` (with the exception that `REAL` also converts non-integer values, so is a superset of functionality) so I don't think we need the seperate operator.

The `real` operator already outputs a float cast for C backends so I think its not needed.

sergisiso:
@svalat This PR may be superseded  by #2298 which adds support for all intrinsics up to Fortran2018. If you want to test it note that they won't be Operations anymore, they are now IntrinsicCalls which in turn inherit from Call.

svalat:
I just moved on the PR you mention to try and it works for me, so as you propose I think we can close this PR which is now useless.

arporter:
Closing as @svalat agrees it has been superseded.

PR #2230, opened 2023-07-21
Improve error message when a kernel functor is not explicitly included in a use statement (closes #2228)

### LINKED ISSUE #2228: LFRic alg layer error when using PSyIR: unknown kernel symbol
The version of `mixed_schur_preconditioner_alg_mod` on LFRic trunk is causing LFRic PSyIR alg-layer parsing to fail as `dg_inc_matrix_vector_kernel_type` is not imported in subroutine `back_substitute`. This is an error in the LFRic code which I've passed on but I need to tidy the error produced by PSyclone to make it clearer what the problem is.

rupertford:
Improves the error message provided to the user when a kernel functor has not been explicitly specified in a use statement when using the new PSyIR approach to parsing the algorithm layer.


rupertford:
Ready for first review from anyone really, but I've specified @sergisiso and @arporter

Issue #2234, opened 2023-07-26
Performance degradation in NEMO OpenMP GPU


sergisiso:
There has been a performance degradation (form 11s to 19s) in the NEMO OpenMP GPU performance, if I look at the CI recorded timings it was first seen in the .eqv. vs .eq. fix (#2209)

2206_fix 9a3a55f2b64bcc056bae97b4171bc717f7a37e07 11.481 11.481
1419_loop_var_unknowntype 4f56807878f7b13be8dddbddfb222e7b0c54bb1c 11.420 11.420
2183_psyad_alg_names 39c0d8469a301d71bedfc437f33dca90a3d4d166 11.419 11.419
2204_fix 592ce49646ba0abb7200f2e9b7fb88a5c3ab7d8a 19.282 19.282
2171_extend_unknownfortrantype 6634a81ca8615be3d05c223698358b4df021bec7 19.242 19.242

Now if I profile this and the the one merge before in master I see that some GPU gaps are introduced: e.g. in tra_ldf
Before:
![image](https://github.com/stfc/PSyclone/assets/1449555/b5c1a058-1da2-4b2e-a99a-719aee859df3)
After:
![image](https://github.com/stfc/PSyclone/assets/1449555/e79898bf-ecfb-4411-85ff-0edfc744cbc7)

Now the strange thing is that if I diff the output step.f90, and ldftra.f90 (which has ldf_tra and ldf_eiv) there is no changes, but somehow the second version does not even produce the nvptx profile records !?

The are changes in other output files (CASE statements with booleans), but I don't see how this have influence here

sergisiso:
There is only 6 files with differing outputs and examining them the most relevant difference is that expressions like:
```
 SELECT CASE (nldf_dyn)
      CASE (np_lap_i)
```
Are now not cannonicalized to if-statements and therefore the body is not parallelised. This is because nldf_dyn and np_lap_i are defined in other modules and we don't have the datatype.

Copying the files over shows that restituting these fixes the performance problem.

Unfortunately I can not simply add the resolve_import at the nemo utils.py as I do in other similar cases because this is a frontend decision and by the time I resolve_imports it is already too late.

sergisiso:
I want to solve this because it is a substantial performance decrease. Also Select/Case string comparisons create lots of problems in the OpenACC and maybe we can also solve them (I need a simple reproducer of this!).

[...]

PR #2240, opened 2023-08-01
Tweak PSyclone's OpenACC example script for NEMO

nmnobre:
Hey,

Since implementing `do while` loops in PSyclone, I don't think I ever tested the example script on the MetOffice's code - bad Nuno 👎 - and we were now hitting a new compiler error which this patch avoids.

Cheers,
Nuno

### DIFF
--- a/changelog
+++ b/changelog
@@ -538,6 +538,9 @@
 	consider that UnresolvedInterface symbols can come from the body of
 	a Fortran Interface.
 
+	182) PR #2240 for #2234. Fix WhileLoop issue in the NEMO OpenACC kernels
+	script.
+
 release 2.3.1 17th of June 2022
 
 	1) PR #1747 for #1720. Adds support for If blocks to PSyAD.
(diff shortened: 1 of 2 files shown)


sergisiso:
Thanks @nmnobre . I have been able to compile it and I will try to run it when the Glados is free.

It doesn't matter much since it is an _init subroutine, but do you know what is the thing that the compiler doesn't like?, both do while loop bodies look fairly simple

sergisiso:
@nmnobre Or is it the do while itself that is the problem (with its conditional exit). Doing some grepping it seems other do whiles are not inside a kernels region.

In this case we should add the "do while" as a condition of "valid_acc_kernel" instead of an ignore_routine?

sergisiso (approved):
@nmnobre Thanks for spotting and fixing this. Since this is a bugfix on a script most of the review tests do not apply. I have also been able to compile and run this WhileLoop excluded version. This is ready to merge.

(since no issue number is mentioned, I will consider it towards #2234)

Issue #446, opened 2019-07-18
Supporting dependence analysis for Call nodes.


hiker:
At the moment calls are just stored as CodeBlocks, which means they can't be further analysed (e.g. if no information about the calls is available, the variable access API still has to mark those accesses as 'UNKNOWN').

While likely not all Fortran statements will be supported by the PSyIR, it was agreed that calls will need to be supported.

arporter:
@hiker, I think this can be closed now as we do represent calls as `Call` nodes?

hiker:
@arporter, is there an implementation of `reference_accesses` in the `Call` class? Or will the arguments be properly marked as 'read-write' (I think that's the best we can do, unless we know more about the subroutine) from one of the base class implementation??

arporter:
Ah, no, it doesn't have such an implementation. I *think* it will use the one in `Node` which in turn just calls `reference_accesses` for each of the child References - these will currently return just 'READ' so that's not correct. I think the simplest solution is to add a `reference_accesses` method to Call and, as you say, set them to 'read-write' for the moment. We could do better by attempting to find the source of the routine that is being called but we won't always be able to do that. (e.g. if it's in a library.)

arporter:
I've just hit this again while doing #1396. I can confirm that we do need to implement `Call.reference_accesses` ! I'll add an x-failing test in the associated PR.

arporter:
In fact, getting test coverage was going to be a pain so I decided I'd do this first. It will only be a 'towards' (if the reviewer agrees) as I shan't attempt to resolve the interface of the called routine. However, it should be good enough for most purposes.

LonelyCat124:
I'm guessing the long-term plan for this is to have the call look up the routine symbol and check the intents (if declared) of the arguments to compute the accesses? At the moment PSyclone is conservative and just makes an assumption that the arguments are all ReadWrite, however for example for NemoLite2D we know the arguments, e.g. if I find the routine (using InlineTrans._find_routine for now since its implemented) corresponding to a Call - I can check the symbol_table and see that `jj` is `Argument(Access.READ)` - the only thing I'm unsure on how to do is check which argument corresponds to which symbol in the symbol_table, but I think by checking the argument vs the symbol_table.argument_list (and then look up the argument_list element in the symbol_table and checking if it has a declared Access?).

This is currently causing false positives for me when attemping to apply a chunk loop trans on a loop containing these kernel calls, as the loop variable is "written to" inside the call, though we know this is not really true.

[...]

PR #2262, opened 2023-08-15
(Towards #446) add reference_accesses() method to Call

arporter:


arporter:
This is just waiting on CI and full integration tests and if they are OK I'll mark it as ready for review.

arporter:
Everything is green. Ready forreview from either @sergisiso , @hiker or @rupertford .

src/psyclone/psyir/nodes/call.py
```diff
@@ -294,6 +296,44 @@ def _validate_child(position, child):
...
+                sig, indices_list = arg.get_signature_and_indices()
+                var_accesses.add_access(sig, default_access, arg)
+                # Any symbols referenced in any index expressions are READ.
+                for indices in indices_list:
+                    for idx in indices:
+                        idx.reference_accesses(var_accesses)
+            else:
+                # This argument is ultimately pass-by-value so any
+                # inputs to it are READ.
+                arg.reference_accesses(var_accesses)
```
sergisiso, line 335: I was almost ready to approve the PR but this comment is wrong, even though the implementation is right.

An expression.reference_access will decide its access type, it is not necessarily READ. For instance:
`call my_pure_sub(a, b, my_impure_increment_and_return_func(c))` would make a:READ, b:READ, c: READWRITE

Maybe you can add this nested example in your `test_call_reference_accesses`

sergisiso (changes requested):
@arporter Almost, I just complain about a comment :)

src/psyclone/psyir/nodes/call.py
arporter: I think this is already covered by:

https://github.com/stfc/PSyclone/blob/2b6e8dfc90608a560fb3b33efc42e8ce8c43f2cb/src/psyclone/tests/psyir/nodes/call_test.py#L423-L430

arporter: I've updated the Comment now though (well spotted).

sergisiso (approved):
@arporter Thanks for the comment change, everything else looks good. Tests and formatting pass (ignoring #2274). The issue is part of TODOs but it will not be closed yet. This is ready to merge.

PR #2270, opened 2023-08-16
(Closes #1575) Remove indentation if no other way to output valid fortran line.

### LINKED ISSUE #1575: PSyclone generates empty continuation lines in output Fortran
Simon has reported:
 
PSyclone can generate very long Fortran source-code lines, for example through indentation of the nested if-then-else constructs that result from if constructs with many "else if" sections in the original source code, so that limiting the length of generated source-code lines is often necessary. The option "-l output" of the 'psyclone' command that provides this, however, produces empty continuation lines if the indentation is 132 characters or more, which turn out to be rejected by at least one compiler. Therefore, I'd like to suggest to filter out empty continuation lines when "-l output" is in use.

Assigning to @rupertford as he wrote the line-length limiter :-)

LonelyCat124:
@arporter I believe this is the solution we discussed in #1575 , should be ready for review if there are no issues with tests.


src/psyclone/tests/line_length_test.py
```diff
@@ -384,6 +384,20 @@ def test_long_lines_allocate(fortran_reader, fortran_writer):
     assert correct in out
 
 
+def test_long_lines_indentation():
+    '''Test that if we have too much initial indentation that we still
+    can output a result.'''
+    long_string = '''                                                     \
```
arporter, line 390: Please could you extend this test to add a case where, even after `ltrim`, the line needs to be broken.

arporter (changes requested):
Thanks for taking this Aidan. I think (but may be wrong) that there's a slightly nicer way to do it - see what you think. Apart from that, the test just needs extending (or a new test adding) where a line is still too long, even dedented (if that's a word).

src/psyclone/line_length.py
```diff
@@ -131,6 +130,25 @@ def process(self, fortran_in):
                 c_end = self._cont_end[line_type]
                 key_list = self._key_lists[line_type]
 
+                # Check if we can output this line with current indentation
+                # level correctly.
+                lstripped_line = line.lstrip()
```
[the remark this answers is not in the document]
LonelyCat124: I did consider this solution, but I wasn't sure which approach to go with. I'll swap to that implementation instead to avoid the duplication instead.

src/psyclone/tests/line_length_test.py
LonelyCat124: Added

LonelyCat124:
Should be all sorted now, will wait for tests to pass and mark as ready if so.

LonelyCat124:
Should be ready for another look.

arporter (approved):
All looks good now. This version of PSyclone works with my artificially-constructed massive if-else-if block.
Will run the integration tests and then proceed to merge if green.

arporter:
Integration tests are all green. Ready to be merged once final CI has run.

Issue #2216, opened 2023-07-10
Inlining `Kern` and subclasses.


LonelyCat124:
Here is a "standalone" test that fails on `master` to do with inlining with `GOKern`s (it can be all just added to the end of the inline transformation test file and will fail).

```
GOCEAN_BASE_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                os.pardir, os.pardir, "test_files",
                                "gocean1p0")
from psyclone.parse.algorithm import parse
from psyclone.psyGen import PSyFactory, CodedKern
from psyclone.domain.common.transformations import KernelModuleInlineTrans

def test_inline_failure():
    ''' Test the apply method of the InlineTrans with gocean. '''
    _, invoke_info = parse(os.path.join(GOCEAN_BASE_PATH, "single_invoke.f90"),
                           api="gocean1.0")
    inline_trans = InlineTrans()
    kintrans = KernelModuleInlineTrans()
    psy = PSyFactory("gocean1.0", distributed_memory=False).\
        create(invoke_info)
    schedule = psy.invokes.invoke_list[0].schedule

    calls = schedule.walk(CodedKern)
    kintrans.apply(calls[0])
    calls[0].lower_to_language_level()
    calls = schedule.walk(Call)
    inline_trans.apply(calls[0])

    code = str(psy.gen)
```
To be able to apply the inline transformation here, I need to get the `GOKern` and convert it into a `Call` (which is usually done at lowering, but I force it to happen early both here, and inside the omp task transformation apply function).

The test breaks at the end of `_find_routine` as the `routine_symbol` is none of `is_modulevar`, `is_unresolved`, nor `is_import`. In fact, if I look at `call_node.root.children[:]` it turns out the routine is `call_node.root.children[1]`, so in the current scope as a `resolved` symbol, however this is not searched and just falls through to the `InternalError`.

I'm not sure if its sensible to search the current scope for any `Routine` and if it is the appropriate routine, to return that routine? Notably in the future we need to check the scope is correct (e.g. for a Routine contained inside a Routine that does not contain the `call_node` we need to not find it and still fall through to the error).

Does this seem like a reasonable idea or am I just using `InlineTrans` badly here?

@arporter @sergisio

sergisiso:
[...]

PR #2291, opened 2023-08-22
(Closes #2290) Add the appropriate interface into `RoutineSymbol` when doing `KernelModuleInlineTrans`

### LINKED ISSUE #2290: `KernelModuleInlineTrans` doesn't result in the `RoutineSymbol` having `is_modulevar` return `True`
After performing a `KernelModuleInlineTrans` transformation, the resulting `RoutineSymbol` cannot be used in an `InlineTrans` as `is_modulevar` returns `False`.

If you instead output the code, and then read in the resulting code then `InlineTrans` will succeed, as the reader correctly identifies the `RoutineSymbol` is `is_modulevar` and then the `InlineTrans` can be done as excpected.

Can duplicate this bug with this script for NemoLite2D:
```
def trans(psy):
    '''Transformation entry point'''
    config = Config.get()

    schedule = psy.invokes.get('invoke_0').schedule
    inline_trans = InlineTrans()

    module_inline_trans = KernelModuleInlineTrans()

    # Inline all kernels in this Schedule
    for kernel in schedule.kernels():
        module_inline_trans.apply(kernel)
        kernel.lower_to_language_level()

    for call in schedule.walk(Call):
        inline_trans.apply(call)
```

I think this might be something to do with how the `RoutineSymbol` is added to the `Container`:
`node.ancestor(Container).symbol_table.add(RoutineSymbol(name))`
For the `RoutineSymbol` to have `is_modulevar` to be `True`, it needs to have `isinstance(self._interface, DefaultModuleInterface)`, so probably something like
`node.ancestor(Container).symbol_table.add(RoutineSymbol(name, interface=DefaultModuleInterface()))`

I'm not sure if that would cause any other edge cases though as I'm not that familiar with those sections of the code.

@sergisiso @arporter does this fix seem reasonable?

LonelyCat124:


LonelyCat124:
Small bugfix, I came to the same realisation as Sergi did when discussing #2216 previously - should be ready for review

arporter (commented):
This is a small (but important) bug fix.
Changes are fine.
Docs don't need updating.
I've fired off the integration tests and if they are green then this is good to go.

Issue #1990, opened 2022-12-15
Kernel extraction driver: handle values imported from other modules


hiker:
Several 'real life' kernels in LFRic Gungho import values from other modules directly, they are not passed in as parameter (esp. if the usage is several level down the call tree: a calls b calls c, which needs the planet_radius).

1. A very simple useful help would be to detect this situation, and print an appropriate message (or abort driver creation, or add the message to the created driver).
2. Better: we collect the list of all these variables, and add them to the extracted file and the driver.

Both solutions would require us to find whatever module is used, parse that module to get the required information.

Some very quick suggested solutions:

1. We use the kernel directories (command line `-d`) to find source files. This feels a bit misusing existing code, but would make it easier to implement: we rely on coding style to define the mapping from module names to file names, and use this to find the file.
2. We provide a new command line parameter listing all files (or all directories), or even a mapping of module names to file names. This is then used as suggested in 1. to find all files, parse them to get additional variables required
3. We could try to utilise a kind of 'caching' mechanism: we provide a list of directories (or filenames), when we have to lookup a name we rely on the naming scheme to find it, but store the value in some kind of cache, so on future invocations we can find the name quickly. That might be difficult to implement correctly (e.g. PSyclone might be running in parallel, so we need to lock accesses to the cache; we also need to make sure the cache is up-to-date, e.g. when LFRic is updated files might be moved to a new directory, ...)

hiker:
Ideally most of the work should be in a stand-alone module/class, so that the same could be used for handling openacc, which has the same issue with values used directly from other modules. It is not certain if the OpenACC problem can actually be solved, but let's aim to make the required code as generic as possible.

hiker:
Some of the modules to be inlined might contain pre-processor directives, which can't be handled by PSyclone (e.g. with LFRic's constant_mod.F90 we get multiple definition of the same symbol). 
- FParser itself seems to handle the file. We could just use fparser since all we need is to find which symbols are imported from which modules.
- We can still just append the plain text to the driver to make other modules available, but we would then need to examine the Fortran source code directly for use statements, which is ugly.

PR #2299, opened 2023-08-31
2297 option for original name in variables access

hiker:
A small PR that is part of #1990 (kernel extraction with non-local variables). It introduces an option to VariableAccessInfo to return the original name of a renamed (`use some_mod, a=>b`) variable name. 
It is not actually used, though should already be fully covered in the tests.

### DIFF
--- a/src/psyclone/core/variables_access_info.py
+++ b/src/psyclone/core/variables_access_info.py
@@ -64,9 +64,13 @@ class VariablesAccessInfo(dict):
         to a True value, arrays used as first parameter to the PSyIR query \
         operators lbound, ubound, or size will be reported as 'read'.
         Otherwise, these accesses will be ignored.
+    :param Any options["USE-ORIGINAL-NAMES"]: if this option is set to a \
+        True value, an imported symbol that is renamed (``use mod, a=>b``)
+        will be reported using the original name (``b`` in the example).
+        Otherwise these symbols will be reported using the renamed name
+        (``a``).
 
     '''
-
     # List of valid options and their default values. Note that only the
     # options method checks this, since it is convenient to pass in options
     # from the DependencyTools that might contain options for these tools.
--- a/src/psyclone/core/variables_access_info.py
+++ b/src/psyclone/core/variables_access_info.py
@@ -67,25 +67,31 @@ class VariablesAccessInfo(dict):
 
     '''
 
-    # List of valid options. Note that only the options method checks this,
-    # since it is convenient to pass in options from the DependencyTools
-    # that might contain options for these tools.
-    _VALID_OPTIONS = ["COLLECT-ARRAY-SHAPE-READS"]
+    # List of valid options and their default values. Note that only the
+    # options method checks this, since it is convenient to pass in options
+    # from the DependencyTools that might contain options for these tools.
+    # COLLECT-ARRAY-SHAPE-READS: controls if access to the shape of an array
+    #     (e.g. ``ubound(a)`` are reported as read or not at all. Defaults
+    #     to True.
+    # USE-ORIGINAL-NAMES: if set this will report the original names of any
+    #     symbol that is being renamed (``use mod, renamed_a=>a``). Defaults
+    #     to False.
+    _DEFAULT_OPTIONS = {"COLLECT-ARRAY-SHAPE-READS": False,
+                        "USE-ORIGINAL-NAMES": False}
 
     def __init__(self, nodes=None, options=None):
         # This dictionary stores the mapping of signatures to the
         # corresponding SingleVariableAccessInfo instance.
         dict.__init__(self)
 
+        self._options = VariablesAccessInfo._DEFAULT_OPTIONS.copy()
         if options:
             if not isinstance(options, dict):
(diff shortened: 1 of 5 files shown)


src/psyclone/core/variables_access_info.py
```diff
@@ -64,28 +64,38 @@ class VariablesAccessInfo(dict):
         to a True value, arrays used as first parameter to the PSyIR query \
         operators lbound, ubound, or size will be reported as 'read'.
         Otherwise, these accesses will be ignored.
+    :param Any options["USE-ORIGINAL-NAMES"]: if this option is set to a \
+        True value, an imported symbol that is renamed (``use mod, a=>b``)
+        will be reported using the original name (``b`` in the example).
+        Otherwise these symbols will be reported using the renamed name
+        (``a``).
```
arporter, line 72: Please document the InternalError here.

src/psyclone/core/variables_access_info.py
```diff
@@ -161,11 +171,11 @@ def options(self, key=None):
```
arporter, line 171: Not the fault of this PR but I've just noticed the doc string (L160) says that this method returns None if the key is not specified in the options. That doesn't look right to me?

src/psyclone/core/variables_access_info.py
```diff
@@ -161,11 +171,11 @@ def options(self, key=None):
 
         '''
         if key:
-            if key not in VariablesAccessInfo._VALID_OPTIONS:
+            if key not in VariablesAccessInfo._DEFAULT_OPTIONS:
+                valids = list(VariablesAccessInfo._DEFAULT_OPTIONS.keys())
+                valids.sort()
```
arporter, line 176: Probably worth a comment explaining that `sort` is needed for reproducible test output.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -172,6 +172,10 @@ def reference_accesses(self, var_accesses):
 
         '''
         sig, all_indices = self.get_signature_and_indices()
+        if isinstance(self.symbol.interface, ImportInterface) and \
```
arporter, line 175: I think you can just do `if self.symbol.is_import and ...` here? If so, you can get rid of the import of ImportInterface again.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -172,6 +172,10 @@ def reference_accesses(self, var_accesses):
...
+                var_accesses.options("USE-ORIGINAL-NAMES") and \
+                self.symbol.interface.orig_name:
+            sig = Signature(self.symbol.interface.orig_name, sig[1:])
```
arporter, line 178: Comment please.

src/psyclone/tests/psyir/nodes/reference_test.py
```diff
@@ -182,6 +183,20 @@ def test_reference_accesses():
...
+    symbol.interface = ImportInterface(ContainerSymbol("my_mod"),
+                                       orig_name="orig_name")
+    reference.symbol = symbol
+    var_access_info = VariablesAccessInfo()
+    reference.reference_accesses(var_access_info)
+    assert "renamed_name: READ" in str(var_access_info)
+
+    var_access_info = VariablesAccessInfo(options={"USE-ORIGINAL-NAMES": True})
+    reference.reference_accesses(var_access_info)
+    assert "orig_name: READ" in str(var_access_info)
```
arporter, line 198: Nice.

arporter (changes requested):
Nice work Joerg.
Updated dev guide builds fine.
Ref. guide builds without new warnings.
Just a little bit of tidying to do.
I'll check integration tests next time (just to be on the safe side).

src/psyclone/core/variables_access_info.py
hiker: Done.

src/psyclone/core/variables_access_info.py
hiker: I've updated the docs, I think this is correct, but not properly explained (it returns none if a valid key is queried that is not specified).

src/psyclone/core/variables_access_info.py
hiker: Done.

src/psyclone/psyir/nodes/reference.py
hiker: Oh, that's nice.

src/psyclone/psyir/nodes/reference.py
hiker: Done.

hiker:
I might need some instructions about the compilation tests. I've triggered them, but they are not running. I'll ask in the telco tonight :)
Otherwise this is ready to go.

src/psyclone/core/variables_access_info.py
arporter: Thanks.

arporter (approved):
CI is all green.
Ref. guide builds fine.
Just waiting on integration tests.

Issue #2319, opened 2023-09-19
Reformat DynStencil into LFRicArgStencil namedtuple


oakleybrunt:
Sub-issue of #2235 
Parallel issue to #2265

After splitting DynStencil from Dynamo0p3 and renaming to LFRicStencil, APorter sees that there is no verification of LFRicStencil setter and getter methods. APorter proposes to replace LFRicStencil with a `namedtuple` LFRicArgStencil. This is to stay in Dynamo0p3.py.

arporter:
#2321 is on master.

Issue #2235, opened 2023-07-27
Classes to be split from dynamo0p3.py


mo-lottieturner:
Umbrella issue for the splitting of dynamo0p3.py into separate files in `domain/lfric/`. Smaller, more detailed issues to be made and linked back to this issue - PRs should only move one class out each.

LFRicLoopBounds: appears in `domain/lfric/lfric_invoke.py` and `tests/domain/lfric/lfric_loop_bounds_test.py`, no name change needed - uses LFRicCollection, so unlikely to have a circular dependency

#### Others unlikely to have circular dependencies:
LFRicScalarArgs: appears in two files and two test files (`grep -rwn LFRicScalarArgs`)
DynStencil: only appears in `dynamo0p3.py`
DynamoInvokes: three files inc `dynamo0p3.py`

#### in `pylint: disable=import-outside-toplevel` markers
DynReferenceElement
DynStencils - lots of places to change this, but only 7 different files  
DynCMAOperators
LFRicMeshProperties
DynBasisFunctions
DynLoop
FSDescriptors, DynKernelArguments
DynInvokeSchedule
LFRicBuiltIn
(LFRicScalarArgs, DynFunctionSpaces, DynDofmaps, LFRicFields, DynLMAOperators, 
DynMeshes, DynBoundaryConditions, DynProxies, LFRicRunTimeChecks,
DynCellIterators, DynReferenceElement, LFRicLoopBounds, DynGlobalSum)

mo-lottieturner:
Table

| Class | rename? | called under a `disable=import-outside-toplevel`? | uses another class in `dynamo0p3.py`? |  number of files | test file situation |
| ----- | :------: | :-----------------------------------------: | :------------------------: | --- |--- |
| DynFuncDescriptor03      | yes |         |  | 4   | split | 
| RefElementMetaData       |  ?  |         |  | 3   | move  | 
| MeshProperty             |  ?  |         |  | 2   | good  | 
| MeshPropertiesMetaData   |  ?  |         |  | 2   | none? | 
| DynKernMetadata          | yes |         |  | 29? | | 
| DynamoPSy                | yes |         |  |   |  | 
| DynamoInvokes            | yes |         |  |   |  | 
| DynStencils              | yes | yes - 15|  |   |  | 
| LFRicMeshProperties      |     | yes - 4 |  |   |  | 
| DynReferenceElement      | yes | yes - 3 |  |   | | 
[...]

PR #2321, opened 2023-09-19
#2319 Refactored DynStencil into a dataclass LFRicArgStencil

oakleybrunt:
PR for #2319 
Part of #2235 

Decided to refactor into a dataclass (part of the dataclasses library) for clarity.

Calls to LFRicArgStencil may not list all parameters which causes an error. I think the defualt value setting in a `namedtuple` is less readable than a `dataclass`. 

A dataclass effectively outsources some methods like `__init__` and `__repr__` to the `@dataclass` decorator: 
https://docs.python.org/3.7/library/dataclasses.html

However, it does involve importing `dataclass` from the `dataclasses` library. 

NOTE: Can you check that the default types are suitable, please?

@arporter, let me know what you think of this.


arporter:
Nice, I didn't know about this. So long as it's standard and works for all our supported versions of Python (and I see that all the tests still pass) then I'm very happy, especially if it lets us document it properly (generating documentation for namedtuples is one of their main weaknesses). Please add a class docstring and see if you can document the various properties so that the reference guide builds correctly and has useful info. in it.

sergisiso:
dataclasses (python>3.7) are fantastic, as well as and typing (python>3.5) which I also see imported in this PR, and it would be great to start using them both!

But given that a release is imminent, maybe we can wait to merge this until after the release is done so we can point to the lastest version that somebody could try using an older python?

One question I have is if we start using typing, e.g. in this PR: `name: str`, do we still need to specify the types in the docstrings, or could we remove them as its duplicated information? Supposedly Sphinx use the type annotations when generating the docs?

TeranIvy:
> > But given that a release is imminent, maybe we can wait to merge this until after the release is done so we can point to the lastest version that somebody could try using an older python?
> 
> Yes, that sounds wise to me so long as @TeranIvy is happy too?

That sounds good, @arporter and @sergisiso! We already have issues with ancient libraries on Crays so I agree with waiting for this after the release. Good work using dataclasses, @oakleybrunt 😄

arporter:
I realised this is in fact the one I should look at next. @oakleybrunt would you mind fixing the conflict and then I'll review.

oakleybrunt:
@arporter it's all yours :)

src/psyclone/dynamo0p3.py
```diff
@@ -8474,14 +8464,16 @@ def __init__(self, call, parent_call, check=True):
...
+                    dyn_argument.descriptor.stencil['type']
+                    )
                 dyn_argument.stencil = stencil
 
                 if dyn_argument.descriptor.stencil['extent']:
                     raise GenerationError("extent metadata not yet supported")
                     # if supported we would add the following
                     # line. However, note there is currently no setter
-                    # for extent in DynStencil so this would need to
+                    # for extent in LFRicArgStencil so this would need to
```
arporter, line 8476: I don't think this comment is true anymore as a dataclass allows its attributes to be set directly? Therefore, I think you can just delete the sentence beginning "However, ...".

arporter (changes requested):
Looks good. I'll fire off the integration tests. There's just one comment that needs updating I think.

src/psyclone/dynamo0p3.py
oakleybrunt: Indeed, no need for setters with a dataclass.

arporter:
The integration tests threw up the fact that we were inadvertently using Python 3.6 for some of them and that failed here because `dataclass` is only valid in Python >=3.7. Therefore, this PR is now blocked by #2345.

arporter:
As discussed in the meeting this morning, please could you update the implementation of LFRicArgStencil to use a `frozen` dataclass and remove the documentation of the types from the class doc string (since we now have type hints for those).

src/psyclone/dynamo0p3.py
arporter: Well, we could but in this case the code changes required are minimal so I wouldn't bother with implementing setters. (Unless of course we modify any of the members elsewhere too.)

oakleybrunt:
Git sullied my `srcpsyclone/dynamo0p3.py` file when I used `git pull` before pushing. That is why youre seeing three commits with essentailly the same message and a fourth where DynStencil is removed again.

arporter (approved):
Changes are all good. Integration tests are green.
Thanks for your patience and persistence with this :-)
Will proceed to merge.

PR #2325, opened 2023-09-19
(closes #1947) Add OMP/ACC Atomics and OMP Simd directive

### LINKED ISSUE #1947: Add support for OpenMP/ACC atomics and OMP SIMD
Adding support for atomics (https://www.openmp.org/spec-html/5.0/openmpsu95.html) will provide more options on how to parallelize loops and kernels.

sergisiso:
This is just the directives nodes, it does not include transformations or automatic additions to solve dependency analysis issues which will be explored in a subsequent PR.

At the moment there is some replication between OMPAtomicDirective and ACCAtomicDirective. I have not made a common subclass because the restrictions (is_valid_atomic_statement method) of each standard are different, and I am unsure if they will end up with the same implementation in the future or not, but the reviewer may disagree.


src/psyclone/tests/psyir/nodes/acc_directives_test.py
```diff
@@ -495,3 +496,83 @@ def test_accdatadirective_update_data_movement_clauses(fortran_reader,
...
+
+    code = '''
+    subroutine my_subroutine()
+        integer, dimension(10, 10) :: A = 1
+        integer, dimension(10, 10) :: B = 2
+        integer :: i, j, val
+
+        A(1,1) = A(1,1) ** 2
+        A(:,1) = A(:,1) / 2 + 3 - 5
+        A(1,1) = MOD(A(1,1), 3)
```
arporter, line 528: How come we exclude MOD? Is there some particular reason?

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -45,21 +45,24 @@
...
-from psyclone.transformations import Dynamo0p3OMPLoopTrans, OMPParallelTrans, \
-    OMPParallelLoopTrans, DynamoOMPParallelLoopTrans, OMPSingleTrans, \
-    OMPMasterTrans, OMPTaskloopTrans, OMPLoopTrans
+from psyclone.transformations import (
+    Dynamo0p3OMPLoopTrans, OMPParallelTrans,
+    OMPParallelLoopTrans, DynamoOMPParallelLoopTrans, OMPSingleTrans,
+    OMPMasterTrans, OMPTaskloopTrans, OMPLoopTrans)
 from psyclone.tests.utilities import get_invoke
 
 BASE_PATH = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(
```
arporter, line 68: Not yours but would you mind fixing:

    psyir/nodes/omp_directives_test.py:573:0: C0325: Unnecessary parens after '=' keyword (superfluous-parens)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -1577,3 +1580,135 @@ def test_omploop_equality():
...
+    atomic.validate_global_constraints()
+
+    # If the statement is invalid (for any reason already tested in a previous
+    # test), it raises an error
+    monkeypatch.setattr(atomic, "is_valid_atomic_statement", lambda _: False)
+    with pytest.raises(GenerationError) as err:
+        atomic.validate_global_constraints()
+    assert "is not a valid OpenMP Atomic statement." in str(err.value)
+
+    # If it doesn not have an associated statement
```
arporter, line 1650: "does"

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -1577,3 +1580,135 @@ def test_omploop_equality():
...
+    atomic.dir_body[0].detach()
+    with pytest.raises(GenerationError) as err:
+        atomic.validate_global_constraints()
+    assert ("Atomic directives must always have one and only one associated "
+            "statement, but found: " in str(err.value))
+
+
+def test_omp_atomics_srtings():
```
arporter, line 1658: "strings"? and again in the test below.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -1577,3 +1580,135 @@ def test_omploop_equality():
...
+    end subroutine
+    '''
+    tree = fortran_reader.psyir_from_source(code)
+    routine = tree.walk(Routine)[0]
+    stmt = routine.children[0]
+    simd = OMPSimdDirective()
+    simd.dir_body.addchild(stmt.detach())
+    routine.children.insert(0, simd)
+
+    # If it doesn not have an associated loop
```
arporter, line 1695: "does"

arporter (changes requested):
Thanks for this Sergi.
Mostly good but I'm a bit worried that the validation will let things through that it shouldn't - see inline.
CodeCov says there are two lines not covered - please could you take a look?
Ref. guide builds without new warnings.
No need to update the User Guide at this point.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -902,8 +906,90 @@ def _sig_set_to_string(sig_set):
...
+            return False
+
+        # Not all rules are checked, just that:
+        # - operands are of a scalar intrinsic type
+        if not isinstance(stmt.lhs.datatype, ScalarType):
+            return False
+
+        # - operators are one of:
+        #   +, *, -, /, AND, OR, EQV, NEQV
+        if isinstance(stmt.rhs, BinaryOperation):
```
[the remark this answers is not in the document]
sergisiso: No need to walk, it just needs to check the top-level operation, think of it as a combined special operation that exists with the assignment, but what the other operand is doesn't matter (I tested this just in case and in fact the compiler accepts anything).

However, your example make me realize that we at least need to check that one of the top-levels operands is the same as the lhs, so both:
`a = a + (something_not_atomic)`
and
`a = (something_not_atomic) + a`
is supported, but something like:
`a = b + (a + 3)`
is not (I also checked this, the compilers do not rearrange this associative math operation, and instead fails with:
`Error: !$OMP ATOMIC assignment must be var = var op expr or var = expr op var at (1)`)

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -902,8 +906,90 @@ def _sig_set_to_string(sig_set):
     return ",".join(sorted(names))
 
 
+class ACCAtomicDirective(ACCRegionDirective):
+    '''
+    OpenACC directive to represent that the memory accesses in the associated
+    assignment must be performed atomically.
```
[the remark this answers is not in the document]
sergisiso: yes it can, done

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -1678,9 +1682,129 @@ def validate_global_constraints(self):
...
+        :rtype: str
+
+        '''
+        return "omp end simd"
+
+    def validate_global_constraints(self):
+        ''' Perform validation of those global constraints that can only be
+        done at code-generation time.
+
+        :raises GenerationError: if the OMPSimdDirective has an associated
```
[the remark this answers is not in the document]
sergisiso: I am unsure that we need it here, as the code is valid OpenMP (simd is just a hint and does not change the code semantics).

But we may want to do it in the transformation, at least to notify, but this is not part of this PR.

src/psyclone/tests/psyir/nodes/acc_directives_test.py
sergisiso: It is not in the list of supported atomic instrinsics :) And I tried with the compiler just in case and they certainly don't like it.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Done

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: fixed

src/psyclone/psyir/nodes/omp_directives.py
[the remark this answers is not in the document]
arporter: Ah, I hadn't realised it was just a hint.

src/psyclone/tests/psyir/nodes/acc_directives_test.py
arporter: Aha!

arporter (changes requested):
We just need to be clear about how many child statements we support for OpenACC atomic and then this is good to go.

src/psyclone/psyir/nodes/acc_directives.py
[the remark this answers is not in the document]
sergisiso: I clarified the docstring.

sergisiso:
@arporter I fixed the suggested docstrings, this is ready for another review.

(I had to do a git pull --merge as I didn't realise the branch was brought to master before I did the changes and this has confused the github PR commit history, but the end diff is still the right one.)

arporter (changes requested):
Thanks for clarifying the docstrings - looks good now.
Unfortunately, this has clashed with Aidan's changes for tasking. Please could you take a look?

sergisiso:
@arporter Brought to master, ready for review again.

arporter (approved):
Changes are all good now and CI is green.
Just waiting on the integration tests.

arporter:
Integration tests were all green. Will proceed to merge...

Issue #2316, opened 2023-09-15
Update fparser version in setup.py


TeranIvy:
PSyclone `setup.py` still requires fparser 0.1.1 release despite the newer 0.1.2 release being on PyPI. 

When installing PSyclone from the master or a branch both fparser 0.1.1 release and 0.1.2 submodule are pulled in. This complicates testing on our end, so I would like to have it fixed for the final testing of PSyclone master before finalising the release.

For info, `lib/python3.7/site-packages` directory of PSyclone test installation built from the current master or a branch looks something like this

```
drwxrwsr-x.  PSyclone-2.3.1.dist-info
drwxrwsr-x.  fparser
drwxrwsr-x.  fparser-0.1.1.dist-info
drwxrwsr-x.  fparser-0.1.2.dev29+gca11947.dist-info
drwxrwsr-x.  psyclone
```

**Update:** We agreed to release fparser 0.1.3 and update `setup.py` to point to that version, as well as point PSyclone master to the latest fparser. The relevant fparser issue is https://github.com/stfc/fparser/issues/427.

arporter:
fparser 0.1.3 is out! I'll do this next...

TeranIvy:
@arporter, I'm happy to finish this :) I'll update the `setup.py` and pull in the most recent fparser master as submodule.

PR #2328, opened 2023-09-20
For #2316: Update setup.py to fparser 0.1.3

TeranIvy:
For #2316. Also updates the fparser submodule to the latest master.


TeranIvy:
@arporter and @rupertford, I think this is ready for review. I am getting some warnings and errors when I run tests with compilation using GFortran 11.2.0. I believe this is because it's stricter on syntax than 6.1.0 that we used before. I'm happy to open another issue after this is merged to fix the syntax.

arporter (approved):
Changes are good thanks.
CI and Integration tests are all green.
Surprisingly, the CI is already set to use the release version of fparser (as opposed to the submodule).
Will proceed to merge.

PR #2329, opened 2023-09-21
(Towards 2205, Closes 2315) Programs now handle contains statements to result in code blocks

LonelyCat124:
Not sure if this needs adding for _program_handler as well as _main_program_handler or not as I'm not clear on the difference.

This is again a workaround similar to that in place for functions and subroutines - it results in a codeblock but I think this is better than the current disappearing act. Future fixes for #2205 will also need to work for Program statements.


src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4574,7 +4574,18 @@ def _main_program_handler(self, node, parent):
...
+        :raises NotImplementedError: if the node contains a Contains clause.
         '''
+        try:
+            _first_type_match(node.children,
+                              Fortran2003.Internal_Subprogram_Part)
+            has_contains = True
+        except ValueError:
+            has_contains = False
+        if has_contains:
+            raise NotImplementedError("Psyclone doesn't yet support 'Contains'"
```
arporter, line 4586: Please could you do away with the `has_contains` and just move this raise directly into L4582. L4594 would then just be `pass` (please also add a comment along the lines of "The routine does not have a Contains section").

src/psyclone/tests/psyir/frontend/fparser2_main_program_handler_test.py
```diff
@@ -94,3 +94,19 @@ def test_main_program_handler(parser, code, expected):
...
+
+    contains
+    Subroutine TestSubroutine()
+    End Subroutine
+    End Program'''
+
+    psyir = fortran_reader.psyir_from_source(code)
+    cblock = psyir.children[0]
+    assert isinstance(cblock, CodeBlock)
+    assert "SUBROUTINE" in str(cblock.get_ast_nodes[0])
```
arporter, line 112: I would have expect the whole Program to be in the CodeBlock? Does re-generating the code (use the `fortran_writer` fixture) give the correct output?

arporter (changes requested):
Thanks for this @LonelyCat124. Everything that helps towards a successful pass-through is very valuable!

I think the change itself could be tidied slightly and I'm a bit worried that the CodeBlock isn't what I expected. I'm therefore requesting that you extend the test just to make sure.
Apart from that this is a small bug fix so no need to update docs.
I'll check the integration tests next time (but don't anticipate any problems).

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Done, also fixed typo in the exception.

src/psyclone/tests/psyir/frontend/fparser2_main_program_handler_test.py
LonelyCat124: Yeah - I should probably check more here, my main thing was to check the subroutine wasn't lost but maybe just checking the output from fortran_writer would also be a good idea.

LonelyCat124:
Should have fixed those issues.

arporter (approved):
Just waiting on the integration tests now before I proceed to merge.

arporter:
Integration tests were green. Will proceed to merge.

Issue #2318, opened 2023-09-18
LFRic Kernel Extraction Driver compilation example is not working


hiker:
```
cd examples/lfric/eg17/full_example_extract
make compile
make driver-main-update
 ./extract.standalone 
 Mesh has           5 layers.
20230918202638.830+1000:INFO : Min/max minmax of field1 =   0.10000000E+01  0.80000000E+01
 ./driver-main-update 
At line 2482 of file driver-main-update.F90
Fortran runtime error: Attempting to allocate already allocated variable 'field1'

Error termination. Backtrace:
#0  0x7fcabebaf960 in ???
#1  0x7fcabebb04d9 in ???
#2  0x7fcabebb0ad6 in ???
#3  0x563bd9445143 in main_update
	at /home/joerg/work/psyclone/examples/lfric/eg17/full_example_extract/driver-main-update.F90:2482
#4  0x563bd944581e in main
	at /home/joerg/work/psyclone/examples/lfric/eg17/full_example_extract/driver-main-update.F90:2446
```
Reason is that a read-write variable is detected as written only. Simple patch:
```
diff --git a/src/psyclone/domain/lfric/lfric_extract_driver_creator.py b/src/psyclone/domain/lfric/lfric_extract_driver_creator.py
index 21cb73201..cfb4bf075 100644
--- a/src/psyclone/domain/lfric/lfric_extract_driver_creator.py
+++ b/src/psyclone/domain/lfric/lfric_extract_driver_creator.py
@@ -578,7 +578,7 @@ class LFRicExtractDriverCreator:
             # variables have References, and will already have been declared
             # in the symbol table (in _add_all_kernel_symbols).
             orig_sym = original_symbol_table.lookup(signature[0])
-            is_input = signature in read_write_info.read_list
+            is_input = any(signature == i[1] for i in read_write_info.read_list)
             if orig_sym.is_array and orig_sym.datatype.intrinsic.name in \
                     self._all_field_types:
                 # This is a field vector, so handle each individual field
```

- [x] Add test that shows the problem
- [x] Fix the problem
- [x] Check if there are similar tests elsewhere that rely on the assumption that the read or write list is signatures only (since they are now tuples module name/signature). Maybe add a convenience function to ReadWriteInfo

hiker:
[...]

PR #2332, opened 2023-09-25
#2318 Small change to fix 2318 before 1990 is merged.

hiker:
Small fix for #2318 (driver runtime errors), taken from #1990.


arporter:
Change looks good. I've fired-off the integration tests...

arporter (approved):
Changes are good.
All integration tests pass.
eg17/full_example_extract builds and runs :-) (We should put this into the integration tests at some point.)
No other changes needed because this is a small bugfix.

PR #2342, opened 2023-09-29
(Closes #2340) fix bug in VERIFY intrinsic.

### LINKED ISSUE #2340: `VERIFY` IntrinsicCall has its `name` attribute set incorrectly
```
        VERIFY = IAttr(
            'TRIM', True, True, False,
            ArgDesc(2, 2, DataNode), {"back": DataNode, "kind": DataNode})
```

This results in `VERIFY` from read Fortran turning into `TRIM` when output.

arporter:


arporter:
To protect us from similar problems in the future, I've added a 'nemo passthrough' job to the NEMO integration tests. The only files excluded from it are those with an INCLUDE of mpif.h and two files where we fail to handle a WHERE correctly. I've added a new script for this - a simplified version of 'process_nemo.py'. It could be merged with that script if the reviewer thinks it would be better.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -35,7 +35,12 @@
 # Modified: R. W. Ford and S. Siso, STFC Daresbury Lab
 # -----------------------------------------------------------------------------
 
-'''This module contains pytest tests for the IntrinsicCall node.'''
+'''
+This module contains pytest tests for the IntrinsicCall node.
+
+TODO #x2341 - tests need to be added for all of the supported intrinsics.
```
sergisiso, line 41: Is the 'x' in the issue number a typo?

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -392,3 +397,105 @@ def test_reference_accesses_bounds(operator, fortran_reader):
...
+    # Should have 4 VERIFY and 2 KIND
+    assert len(psyir.walk(IntrinsicCall)) == 6
+    result = fortran_writer(psyir).lower()
+    assert "if (verify(clname(ind1:ind2), '0123456789') == 0) then" in result
+    assert ("if (verify(clname(ind1:ind2), '0123456789', back=.true.) "
+            "== 0) then" in result)
+    assert ("if (verify(clname(ind1:ind2), '0123456789', kind=kind(1)) "
+            "== 0) then" in result)
+    assert ("if (verify(clname(ind1:ind2), '0123456789', kind=kind(1), "
+            "back=.true.) == 0) then" in result)
```
sergisiso, line 501: This tests are good, but in addition, instead of testing them all, we can add a:

```
    def test_enumerator_name_matches_name_field():
        for intrinsic_entry in IntrinsicCall.Intrinsic:
             assert intrinsic_entry._name_ == intrinsic_entry.name
```

Which, as you will discover, it spots another 2 errors 😓

sergisiso (changes requested):
Thanks @arporter, this will help improve the testing. I added some comments inline.

Another suggestion this is that, regardless of whether we merge the passthrough.py with the process_nemo.py (which I think it would be a good idea), couldn't we at least remove the EXLUDED_FILES that were not necessary anymore from process_nemo.py. 
Or move them to PROFILE_ONLY if make the perf slower or inside the script (e.g.  the one regarding HoistLocatArraysTrans to skip this particular transformation in only this file)?

examples/nemo/scripts/Makefile
```diff
@@ -58,6 +58,8 @@ $(TARGETS): % : $(addprefix psycloned-%/, $(notdir $(SRC_FILES)))
 	@echo "Finished generating $@"
 
 # PSyclone instructions for each target type
+psycloned-passthrough/%.f90: ${ROOT_SRC}%.f90 psycloned-passthrough
+	${PSYCLONE_NEMO_DIR}/passthrough.py -I ${ROOT_SRC} -o psycloned-passthrough $<
 psycloned-openmp_cpu/%.f90: ${ROOT_SRC}%.f90 psycloned-openmp_cpu
```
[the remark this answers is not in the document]
arporter: Oops, done.

examples/nemo/scripts/passthrough.py
```diff
@@ -0,0 +1,138 @@
...
+                sys.exit(1)
+            FAILED_FILES.append(ffile)
+        else:
+            print(f"Time taken for {file_name}: {tstop - tstart:.2f}s")
+
+        print("--------------------\n")
+
+    print("All done.")
+    if FAILED_FILES:
+        print(f"PSyclone failed on the following source files: {FAILED_FILES}")
```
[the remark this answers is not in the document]
arporter: Oh yes, good point. Added to process_nemo.py (and I've removed this file now).

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
arporter: Oops.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
arporter: Oh yes! Good test, thanks.

arporter:
I've rationalised so that we only have the `process_nemo.py` script again. This script now does a passthrough on any files that aren't being transformed so we have better coverage. I've fired off the integration tests.

I'll investigate whether I can slim down the list of excluded files now...

arporter:
The NEMO integration test (ECMWF OMP CPU build) is now failing because it turns out that PSyclone processing of `diaharm.f90` was always failing but wasn't being picked up (due to the lack of `sys.exit(1)` spotted by Sergi).

arporter:
I've found the source of the problem: in `validate` we weren't telling the dependence analysis to check *all* variables so it was returning with the first problem it found which happened to be one that we then proceeded to ignore (single-write-to-a-scalar). I've fixed this by telling it to check every variable. That then broke some of the LFRic tests where we parallelise loops over dofs containing reductions. However, I was able to fix this in DynLoop.independent_iterations so that recent addition worked well :-)

sergisiso (approved):
@arporter All comments have been addressed. The intrinsics testing and the NEMO integration test have been improved. There are no TODOs references, linters are clean (although pycodestyle finds issues but not related to this PR). Test and code coverage pass. This is ready for merging.

PR #2344, opened 2023-10-01
maxval, minval and sum 2 code support for array argument expressions and constant indices (closes #2323)

### LINKED ISSUE #2323: Maxval2CodeTrans with operator/intrinsics as top-level argument
I tried applying the new Maxval2CodeTrans to the NEMO missing GPU region.
The intrinsics that we aim to transform look as following:
```
stpctl.f90:
119:      zmax(1) = MAXVAL(ABS(sshn(:,:) + ssh_ref * tmask(:,:,1)))
121:      zmax(1) = MAXVAL(ABS(sshn(:,:)))
123:    zmax(2) = MAXVAL(ABS(un(:,:,:)))
134:      zmax(3) = MAXVAL(-tsn(:,:,:,jp_sal), mask=llmsk)
135:      zmax(4) = MAXVAL(tsn(:,:,:,jp_sal), mask=llmsk)
137:        zmax(5) = MAXVAL(-tsn(:,:,:,jp_tem), mask=llmsk)
138:        zmax(6) = MAXVAL(tsn(:,:,:,jp_tem), mask=llmsk)
140:          zmax(9) = MAXVAL(cu_adv(:,:,:), mask=llmsk)
151:            zmax(8) = MAXVAL(ABS(wi(:,:,:)), mask=wmask(:,:,:) == 1._wp)
```

See output below, some good and bad news. 1 of them (line 140) has been converted and the dependency analysis properly reports "maxval_var" as a reduction 🥳   The the rest are refused because there is an intrinsic (ABS) or UnaryOperator (MINUS) as top-level node or a fixed dimension in the array accessor:
```
Transformation Error: Maxval2CodeTrans only supports arrays or plain references for the first argument, but found 'IntrinsicCall'.
MAXVAL(ABS(sshn(:,:) + ssh_ref * tmask(:,:,1)))
Transformation Error: Maxval2CodeTrans only supports arrays or plain references for the first argument, but found 'IntrinsicCall'.
MAXVAL(ABS(sshn(:,:)))
Transformation Error: Maxval2CodeTrans only supports arrays or plain references for the first argument, but found 'IntrinsicCall'.
MAXVAL(ABS(un(:,:,:)))
Transformation Error: Maxval2CodeTrans only supports arrays or plain references for the first argument, but found 'UnaryOperation'.
MAXVAL(-tsn(:,:,:,jp_sal), mask=tmask(:,:,:) == 1._wp)
Transformation Error: Maxval2CodeTrans only supports arrays with array ranges, but found a fixed dimension in 'tsn(:,:,:,jp_sal)'.
MAXVAL(tsn(:,:,:,jp_sal), mask=tmask(:,:,:) == 1._wp)
Transformation Error: Maxval2CodeTrans only supports arrays or plain references for the first argument, but found 'UnaryOperation'.
MAXVAL(-tsn(:,:,:,jp_tem), mask=tmask(:,:,:) == 1._wp)
Transformation Error: Maxval2CodeTrans only supports arrays with array ranges, but found a fixed dimension in 'tsn(:,:,:,jp_tem)'.
MAXVAL(tsn(:,:,:,jp_tem), mask=tmask(:,:,:) == 1._wp)
Transformation Error: Maxval2CodeTrans only supports arrays or plain references for the first argument, but found 'IntrinsicCall'.
MAXVAL(ABS(wi(:,:,:)), mask=wmask(:,:,:) == 1._wp)
Loop not parallelised because: Transformation Error: Dependency analysis failed with the following messages:
Warning: Variable 'maxval_var' is read first, which indicates a reduction.
Loop not parallelised because: Transformation Error: Dependency analysis failed with the following messages:
Warning: Variable 'maxval_var' is read first, which indicates a reduction.
Loop not parallelised because: Transformation Error: Dependency analysis failed with the following messages:
Warning: Variable 'maxval_var' is read first, which indicates a reduction.
```

We could allow operators and elemental intrinsiccalls as arguments in this transformation because we can assume that the array inside them will give use the shape that we need, or we could go a more generic route and continue the #685, since this seems the simpler case for node datatype inference. What do you think @rupertford @arporter ? Has any of you started #685?

rupertford:
Extends maxval, minval and sum intrinsics so that the array argument can be an expression. Also adds support for array argument indices to have a scalar value. This additional functionality is needed in NEMO. Note, a more general solution would be to replace expression with a temporary variable but that would require knowing the type, dimensions, etc for the expression and we don't support that yet.

### DIFF
--- a/src/psyclone/psyir/transformations/intrinsics/mms_base_trans.py
+++ b/src/psyclone/psyir/transformations/intrinsics/mms_base_trans.py
@@ -42,7 +42,7 @@ from abc import ABC, abstractmethod
 
 from psyclone.psyir.nodes import (
     Assignment, Reference, Literal, Loop, ArrayReference, IfBlock, Range,
-    IntrinsicCall)
+    IntrinsicCall, Node, UnaryOperation, BinaryOperation)
 from psyclone.psyir.symbols import (
     DataSymbol, INTEGER_TYPE, ScalarType, ArrayType)
 from psyclone.psyGen import Transformation
@@ -61,13 +61,13 @@ class MMSBaseTrans(Transformation, ABC):
     @staticmethod
     def _get_args(node):
         '''Utility method that returns the minval, maxval or sum arguments,
-        (array reference, dimension and mask).
+        (expression (containing array reference), dimension and mask).
 
         :param node: a minval, maxval or sum intrinsic.
         :type node: :py:class:`psyclone.psyir.nodes.IntrinsicCall`
 
         returns: a tuple containing the 3 arguments.
-        rtype: Tuple[py:class:`psyclone.psyir.nodes.reference.Reference`,
+        rtype: Tuple[py:class:`psyclone.psyir.nodes.DataNode`,
             py:class:`psyclone.psyir.nodes.Literal` |
             :py:class:`psyclone.psyir.nodes.Reference`,
             Optional[:py:class:`psyclone.psyir.nodes.Node`]]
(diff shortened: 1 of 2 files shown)


rupertford:
Given that this is actually going to have to solve a few problems to work for all the cases outlined be Sergi

1) expressions (working)
2) multiple arrays in an expression (not yet implemented)
3) scalar dimensions (not yet implemented)

I thought I should review whether it makes sense to wait for "determining the result datatype from an expression" to be added to the PSyIR. However, I think these are independent solutions and this one should be more efficient as it involves less copying, so am planning to continue.

The obvious thing to do if we know the result datatype is the following:

```
x  = maxval(a(:)+b(:))
```
becomes
```
allocate(c(n))
c(:) = a(:) + b(:)
x  = maxval(c(:))
```

which can then be used with the existing transformation.

Whereas the solution in this PR would be to do the following

```
result = tiny() ! might need changing depending on expression
do i = 1, n
  if (result < (a(i)+b(i)) then
    result = a(i) + b(i)
  end if
end do
x = result
```

The above solution removes the need for an array copy so I will continue.

sergisiso:
@rupertford I had a look at this again, and the resulting "inlined" max expression is probably not what we want, because we then still have to annotate the operation as a reduction or atomic, and both of this support min/max as an operator, but not the multi-line condition, so instead of:
```
  if (result < (a(i)+b(i)) then
    result = a(i) + b(i)
  end if
```
we want to simply leave it at
```
MAX(result, a(i) + b(i))
```

sergisiso:
Also, looking at the implementation I feel like we are making things more complicated that it needs to be, the array range to loop for expressions is already implemented, and it took some time to do it well, so we should re-use it:

Step 1) from "x  = maxval(a(:)+b(:))", extract the expr and put it here:
```
a(:) = a(:) + b(:)
```
first array on the lhs because the transformation below checks for an array in the lhs and it uses it for the bounds

Step2) call arrayrange2loop_trans.py on the assignment and get
```
do idx = LBOUDN(a,1), UBOUND(a,1)
   a(idx) = a(idx) + b(idx)
enddo
```

Step3), convert again the assignment to its
```
do idx = LBOUDN(a,1), UBOUND(a,1)
   MAX(x, a(idx) + b(idx))
enddo
```

Step 4) Add x initialisation.
```
x = tiny()
do idx = LBOUDN(a,1), UBOUND(a,1)
   MAX(x, a(idx) + b(idx))
enddo
```

rupertford:
This PR has been closed as a different implemented is being done (see issue #2323)

PR #2346, opened 2023-10-02
(closes #2345) change NEMO and LFRic integration tests to use module version of Python

### LINKED ISSUE #2345: Integration tests currently using Python 3.6
The virtual env. used by the integration tests appears to be based on Python 3.6. This is causing the LFRic tests to fail for #2319 as that issue is using `dataclasses` which only exists in Python >= 3.7.

arporter:


arporter:
This is a small fix to ensure that all integration tests use the same, recent (supported) version of Python. It is required by #2319 as that introduces a dependence on Python >= 3.7 while some of the tests are using the (EOL) Python 3.6.

arporter:
Link check is a random failure with the NVIDIA web site - I can visit the page without problems. All the integration tests are now green so this is ready for review. Probably one for @sergisiso as he's most familiar with the integration tests.

sergisiso (approved):
@arporter Thanks for fixing this. This is github actions fix, I can see that the integration test have passed successfully, no code has been touched so no tests or documentation is needed.

PR #2350, opened 2023-10-03
update fparser version on PSyclone master (closes #2349)

### LINKED ISSUE #2349: update fparser version on master
There is a new version of fparser on master that fixes a bug and allows PSyclone to inline a kernel. We should therefore make use of this new version by default, rather than 0.1.3

rupertford:
This PR adds in the master version of fparser to external/fparser and updates the dependencies so that 0.1.3 or newer versions can be used. It may be better to only allow newer versions than 0.1.3 but I've not done that.


rupertford:
Ready for review from @arporter or @sergisiso I guess, but anyone can take it.

arporter (changes requested):
Almost perfect. You just need to update `.github/workflows/python-package.yml` so that it uses the submodule version of fparser rather than the one from pypi.

(I note that this setup is duplicated across all our yml files which is not nice but I'm not sure how easy it would be to fix.)

rupertford:
I'm getting this ...

 ! [remote rejected]     2349_update_fparser -> 2349_update_fparser (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/python-package.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/stfc/PSyclone.git'

Perhaps I've not given my token enough privileges.

rupertford:
Ready for second review from @arporter

arporter (approved):
Requested change has been made.
This is good to go.

PR #2355, opened 2023-10-04
multi-precision metadata support (closes #2347)

### LINKED ISSUE #2347: Add mixed precision metadata support to PSyIR in LFRic
Now that #2255 is on master we no longer lose values in structure types in the PSyIR. This means that we can now add mixed precision metadata support to the PSyIR for LFRic metadata (as mixed precision metadata is captured as a structuretype in fparser2).

Note, there is a potentially separate problem of how we deal with the associated interface. I'm not yet sure what is best to do here. We could add it as metadata but it might need to be added back in a particular place when lowering. We could have it as a separate high level LFRic PSyIR concept, kept in its original location. Or something else!

rupertford:
It turns out that now we do not lose the "extends" keyword for the metadata then multi-precision PSyIR ends up being an unknownfortrantype, not a StructureReference, so the sasme code works for both single and multi-precisions kernels. This  PR just adds some tests to demonstrate this. Note, the issue of interfaces has not yet been addressed. That will be anther PR.

### DIFF
--- a/src/psyclone/tests/domain/lfric/transformations/raise_psyir_2_lfric_kern_trans_test.py
+++ b/src/psyclone/tests/domain/lfric/transformations/raise_psyir_2_lfric_kern_trans_test.py
@@ -70,6 +70,32 @@ PROGRAM = (
     f"end module dummy\n")
 
 
+METADATA_MULTI = (
+    "type, extends(kernel_type) :: testkern_type\n"
+    "   type(arg_type), dimension(1) :: meta_args =  (/     &\n"
+    "           arg_type(gh_field,    gh_real, gh_inc,  w1) &\n"
+    "         /)\n"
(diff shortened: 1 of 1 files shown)


rupertford:
Ready for first review. Probably @arporter or @sergisiso

arporter:
Hi @rupertford, am I correct in thinking that this change doesn't affect the user yet?

rupertford:
> Hi @rupertford, am I correct in thinking that this change doesn't affect the user yet?

It should already help, but is a result of a previous PR #2255. This PR just checks it works. I think some of the inlining done by @sergisiso should already benefit. Basically, whenever there was the error can't parse StructureType, it should no longer do that.

arporter (approved):
This is just a change to the testing so no doc update or integration testing required.
Standard CI is all green.
Will proceed to merge.

PR #2361, opened 2023-10-11
(Closes #2360) Fixes for LFRic dependency analysis

### LINKED ISSUE #2360: [LFRic] Dependence analysis issues
Transforming Gravity Wave for GPU using `parallel` and `loop` directives has revealed a couple of problems:

 1. ParallelLoopTrans does not take into account whether the loop is to be run in serial (e.g. ACC LOOP seq);
 2. `DynLoop.independent_iterations` says that all loops over dofs can be parallelised whereas some Builtins perform reductions that will require special treatment.

arporter:
A couple of small fixes identified when doing OpenACC for LFRic.


arporter:
Small fixes that change the way DA is done for LFRic loops. These were needed to do ACC PARALLEL/LOOP for LFRic.

src/psyclone/dynamo0p3.py
```diff
@@ -7348,11 +7348,22 @@ def independent_iterations(self,
...
+        if self.loop_type == "dof":
+            # The generic DA can't see the PSyIR of this Builtin (because it
+            # hasn't been lowered to language level) so we use
+            # domain-specific knowledge about its properties.
+            if self.kernel.is_reduction:
+                dtools._add_message(
+                    f"Builtin '{self.kernel.name}' performs a reduction",
+                    DTCode.WARN_SCALAR_REDUCTION)
+                return False
+            return True
```
sergisiso, line 7365: Just to check, is this different from the condition below, does bulitins have "has_inc_arg"? and does the "has_inc_arg" below show put a similar message in dtools?

sergisiso (changes requested):
@arporter it looks good, I just have some general questions about how these fits with other LFRic transformations/kernels. But feel free to disregard if it leads to complications.

src/psyclone/transformations.py
```diff
@@ -681,8 +681,11 @@ def validate(self, node, options=None):
                 raise TransformationError(
                     f"Error in {self.name} transformation. The kernel has an "
                     f"argument with INC access. Colouring is required.")
-
-        super().validate(node, options=options)
+        # As this is a domain-specific loop, we don't perform general
+        # dependence analysis.
+        local_options = options.copy() if options else {}
+        local_options["force"] = True
+        super().validate(node, options=local_options)
```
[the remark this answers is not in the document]
arporter: No, the DA still rejects the loop for valid reasons - it's just that we know that come code-generation time we will put the correct clauses on the OpenMP directive. I've extended the comment. I've checked and the has_inc_arg is not done in either ParallelLoopTrans.validate() or LoopTrans.validate(). It has to be said that our use of domain-specific information is not well encapsulated in the various transformation classes at the minute - we have a few `if isinstance(node, PSyLoop)...` in the validates.

src/psyclone/tests/domain/lfric/lfric_loop_test.py
```diff
@@ -1023,31 +1022,12 @@ def test_loop_independent_iterations(monkeypatch, dist_mem):
...
+        loops[1].independent_iterations()
     assert "loop of type 'broken' is not supported" in str(err.value)
     monkeypatch.undo()
-    # Test that DA warnings about a scalar variable are ignored if the loop is
-    # over DoFs.
-    monkeypatch.setattr(DependencyTools, "get_all_messages",
-                        lambda _1: [Message("just a test",
-                                            DTCode.WARN_SCALAR_REDUCTION)])
-    assert loop.independent_iterations()
     # Test when the DA returns True. Since this currently never happens for an
```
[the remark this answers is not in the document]
arporter: I tried and we miss a line if we don't do it. I think the key thing here is that the LFRic kernel is over cell columns rather than dofs. I've extended the comment.

src/psyclone/dynamo0p3.py
arporter: No, Builtins do not have `has_inc_arg` as they do not perform INC operations (which only happen when we iterate over cells and write to shared dofs). However, they can perform reductions (e.g. MIN, MAX, SUM).
I've extended the "has_inc_arg" so it too adds a message to dtools. BTW, I'm currently being naughty and doing `dtools._add_message`. Should I make the `add_message` method public? Is @hiker happy with that?

hiker: Gee, that's an odd case. To some degree I wouldn't want to make this public (since this functionality should not be used by someone using the DT), but it makes sense here :(
TBH, I don't mind either. The `independent_iterations` function is kind of part of the DT, so I would be ok leaving it this way (i.e. 'private'), but I would think that you also need to disable a warning here(?).The moment a user-script would add its own messages (e.g. in case of a slightly-higher-level script that's being used by a PSyclone transformation), this should then become public.

sergisiso (approved):
@arporter Thanks for expanding the comments, the rest of the PR looks good. There are no TODOs remaining, linters don't show new issues, test and coverage pass. I am running integration but unless there is some error there, this is ready to merge.

PR #2363, opened 2023-10-11
(Closes #1254) catch unsupported statements in implicit-part

### LINKED ISSUE #1254: [psyir] fparser2 frontend currently ignores implicit statements
To date we have always (implicitly) assumed that the Fortran we are dealing with is using `implicit none`. The fparser2 psyir frontend needs extending to check that this is the case and to abort if not.

arporter:
This fixes a long-standing issue whereby we were silently ignoring things that could appear in the 'implicit-part' of a specification-part. We now catch things like `implicit real (a-h, o-z)` and the awfulness that is ENTRY statements.


arporter:
Integration tests are all green. This is ready for review from @rupertford, @sergisiso or @hiker.

Issue #2120, opened 2023-04-26
Kernel Extraction - Handle failures


hiker:
If there are issues with parsing a kernel or finding a symbol, atm a message is printed, but otherwise this is ignored. The message might also easily get lost in a long list of output.

We should have an option to abort in case of a problem with the kernel extraction (e.g. symbol cannot be accessed since it's private in an external module) or driver creation. But in some cases it is also useful to keep on going (since otherwise testing on LFRic is very hard if psyclone fails for one file).

Suggest solutions:
1. Add a command line or config file parameter to control if PSyclone should abort or just continue
2. Add information to the driver about these problems - so either the driver does not compile, or it compiles, but would write a warning, error, or even abort at runtime with an appropriate error code.
3. The code likely causing the problem should either be added, or maybe added as a comment (so the user can manually fix it??)

Issue #2054, opened 2023-02-15
Make Kern be a mixin to attach functionality/meaning to other nodes instead of a Node itself


sergisiso:
Currently kernel calls are CodedKern that are in turn are Kern and in it inherits from Node.
This means that:
- `tree.walk(Call)` does not return calls to kernels, which is unintuitive
- also that generic call transformations  can not be applied to CodedKern
- and that after lowering a CodedKern to a Call, other optimization that are encoded in the gen_code path don't work anymore.
(this is a temporal issue since ideally this optimizations have to be brought to an earlier step when everything is still in the higher abstraction nodes)

This could be solved if we define CodedKern as inheriting from both Kern(baseclass) and Call (Node). Similar with BuiltIn and InlindedKernel (this inheriting from Schedule - instead of having a schedule)

PR #2367, opened 2023-10-17
2296 add routine info

hiker:
Adds a separate `RoutineInfo` object to the module manager.


hiker:
Besides adding the `RoutineInfo` object it also adds support to ignore modules (in the config file), supports generic interfaces (by creating individual info objects for each function listed in the generic interface, then providing a GenericRoutineInfo object that returns the union of the information. At some stage we might try to identify exactly which version is called here). I added quite a few example .f90 files originally used for the improved kernel extraction, but I needed them for test coverage).
I tried to remove any other modification that was in the original PR #1990.

arporter:
Hi @hiker, @sergisiso and I have some concerns about this. It feels as though it is implementing another layer on top of the PSyIR which is not ideal. A lot of the functionality looks as though it is either already present in the `Routine` node or could be. Would you mind taking a look to see whether you can achieve the same goals purely using PSyIR?

hiker:
Thanks, @arporter and @sergisiso . I had a look, and it would seem possible to add most of the functionality to the Routine class. The one somewhat difficult part is the support for generic interfaces, which atm is done by a specialisation of the RoutineInfo object, which makes it really easy to handle in the dependency analysis (which is not part of this PR). I am not even sure if we have an existing node we could use for generic interfaces (it's all int he symbol table afaik), but I *think* it might just be possible to get a list of objects for a symbol from the ModuleInfo object (so if there is no generic interface, it will be a list of length 1, otherwise one PSyIR node for each implementation of the interface - I would just have to move the loops out of the RoutineInfo into the DA. Not neat, but will work.

There was a long-term vision in my implementation though (which I never really shared ... would be so much easier if I were on site :)  ): I would assume at some stage we will come to the conclusion that we need to support some kind of caching mechanism, since we might need to parse imported modules for type information (or if there are generic imports, and we need to find where a symbol is coming from) frequently, and fparser/PSyclone parsing is slow. My idea was that PSyclone's command line could be extended to take list of files to process (scripts would be specified using e.g. a colon-separated list of format strings: `optimisation/{SITE}/{RELATIVE_PATH}/{BASENAME}.py:optimsation/{SITE}/global.py` or so, which would support the current LFRic way and more). Then the module manager could cache all required information, and manage memory as well (e.g. discard infrequently used AST and PSyIRs, but keep the module information about symbols). We could even try to add a memory limit for this kind of caching (if required). Doing this in the PSyIR won't work .... since we then cannot discard the PSyIRs which would store the cached information in the first place (maybe ... throw away the children??? ... Scary)

Also, it would make the PSyIR nodes 'heavier' (I would expect that e.g. routine.py kind of doubles in size, though still under 1000 lines). E.g. if I move this code into the `Routine` class, now the routine class test also needs to test the handling of local and non-local symbols, which is kind of unrelated to the information in the  PSyIR node. While atm all the external handling is in this new file (and corresponding tests). I actually like the idea to keep the actual code for the PSyIR nodes as small as possible, and write tools around the PSyIR which use them, but I am aware that you seem to prefer 'all in one big node' approach (e.g. our discussion about the domain-specific DA, where I wanted a small callbacks added to the API-specific nodes, but you preferred to move thee actual call to the DA into the nodes). 

So, I assume that all of this is not strong enough to convince you, one question:

Should I still cache the information (e.g. about non-local symbols)? I assume that part of the reason to store this in the PSyIR is that the PSyIR might change, and we then might have outdated info. That could become quite expensive, e.g. thinking of driver extraction, nearly any module in the full call tree will use `constants_mod.F90`. So we would traverse its PSyIR over and over and over to collect its symbols, since none of the info (e.g. which external symbols does it use) is cached.

sergisiso:
> I am not even sure if we have an existing node we could use for generic interfaces

Generic interfaces are declarations, currently fall into RoutinesSymbols with UnknownFortranType (I think). We could implement a better Symbol Type for these, but they are not nodes, right?

> I would assume at some stage we will come to the conclusion that we need to support some kind of caching mechanism

I did thought about that, and in the case of a single execution (like you are proposing with multiple files - I am undecided about this) I think the logical place is at src/psyclone/psyir/frontend/fortran.py::psyir_from_file we have the filename as the identified and all file parsing should go through here, so it can memorize previous results and return them immediately when requested.

> Doing this in the PSyIR won't work .... since we then cannot discard the PSyIRs which would store the cached information in the first place 

I am not sure why not, the PSyIR should not store the caching itself, but a top dictionary with {filename: psyir_root}, we can discard items from here if memory is a concern (which I don't think it is at the moment).

> It would make the PSyIR nodes 'heavier'

This is a concern, and there is a balance to it. But the problem with splitting things that are tightly connected is that then they are hard to discover and we end up replicating functionality. Also, when modifying the core node is easier to break the dependent external functionality, e.g. knowing that changes would invalidate a cached symbol list in an external class.

This is why in general I am against this local caching (because fparser is really the bottleneck, not traversing again the psyir_tree), but at least in psyir we now have the "update_signal" to invalidate caches/precomputed values.

> Should I still cache the information (e.g. about non-local symbols)? 

I wouldn't, see my previous response :) 

> That could become quite expensive, e.g. thinking of driver extraction, nearly any module in the full call tree will use constants_mod.F90

In addition to the top-level file caching, we could explore the idea of having an fparser pass that only parses the module declarations and routine signatures because this is all what it is exposed externally. fparser is slow when it starts recursing the parser rules for each expression, but it may be much faster if we just do the module declarations part. Maybe we should explore this route.

hiker:
> > I am not even sure if we have an existing node we could use for generic interfaces
> 
> Generic interfaces are declarations, currently fall into RoutinesSymbols with UnknownFortranType (I think). We could implement a better Symbol Type for these, but they are not nodes, right?

Yes, that is kind of the problem - I need 'something' that stores the generic interface details (i.e. list of routine names, so I can collect all dependencies from all routines atm). But I hope using a list of routines will suffice. Not as neat as my encapsulated solution though.

> 
> > I would assume at some stage we will come to the conclusion that we need to support some kind of caching mechanism
> 
> I did thought about that, and in the case of a single execution (like you are proposing with multiple files - I am undecided about this) I think the logical place is at src/psyclone/psyir/frontend/fortran.py::psyir_from_file we have the filename as the identified and all file parsing should go through here, so it can memorize previous results and return them immediately when requested.

We still need a mapping from module name to file name. Though the ModuleManager already provides that (actually, it maps also module name to PSyIR immediately). This is already on master, so not worth discussing here :)

> 
> > Doing this in the PSyIR won't work .... since we then cannot discard the PSyIRs which would store the cached information in the first place
> 
> I am not sure why not, the PSyIR should not store the caching itself, but a top dictionary with {filename: psyir_root}, we can discard items from here if memory is a concern (which I don't think it is at the moment).

The top dictionary with filename -> PSyIR mapping is basically what the RoutineInfo object implements (well, it maps subroutine name instead of filename, which I think is more useful, since typically we have the subroutine name, but not the file name :)

Well, we can actually ignore this, if we don't cache data (with this I mean the list of external symbols used by a function).

> 
> > It would make the PSyIR nodes 'heavier'
> 
> This is a concern, and there is a balance to it. But the problem with splitting things that are tightly connected is that then they are hard to discover and we end up replicating functionality. Also, when modifying the core node is easier to break the dependent external functionality, e.g. knowing that changes would invalidate a cached symbol list in an external class.

Exactly the 'tightly connect' part - is function 'A' tightly connected with a variable in module D, because A calls B which calls C which uses D? If we modify C, then variable "D" is not connected to A anymore. I would not consider this a tight coupling. For me that is more a derived value, not part of the node. But I agree, this is more a coding preference style :) And ... now that I think about it, if we don't cache, this is OK. Sorry, seeing the time it takes to create the drivers, I am a bit obsessed with caching intermediate results:
In LFRic you can get long call trees ... A calls B calls C calls D, so each time we ask for the list of all external symbols for A, we need to go through all these PSyIRs again, instead of just returning the cached info. And you will ask for e.g. `log_mod` (and field and ...) quite often in LFRic :)

Let's say I am not happy about this, but will accept your majority decision (though I might try to do a speed test :)  )

> 
> This is why in general I am against this local caching (because fparser is really the bottleneck, not traversing again the psyir_tree), but at least in psyir we now have the "update_signal" to invalidate caches/precomputed values.
> 
> > Should I still cache the information (e.g. about non-local symbols)?
> 
> I wouldn't, see my previous response :)

OK :(


> 
> > That could become quite expensive, e.g. thinking of driver extraction, nearly any module in the full call tree will use constants_mod.F90
> 
> In addition to the top-level file caching, we could explore the idea of having an fparser pass that only parses the module declarations and routine signatures because this is all what it is exposed externally. fparser is slow when it starts recursing the parser rules for each expression, but it may be much faster if we just do the module declarations part. Maybe we should explore this route.

For driver creation I do need more, since I have to follow the call tree. So I need the full tree (e.g. to find if a variable is accessed that comes from a different module, and if this variable is read or written).

sergisiso:
@hiker Maybe before you do any change on this PR (even what we suggested) we should have a call to understand it better. I did not get before that you needed the transitive call dependencies inside each routine on LFRic.. that is in fact a lot of code to parse!

hiker:
> @hiker Maybe before you do any change on this PR (even what we suggested) we should have a call to understand it better. I did not get before that you needed the transitive call dependencies inside each routine on LFRic.. that is in fact a lot of code to parse!

I'm always happy to have a chat with you, but I did some performance tests (I took the driver creation for fem_constants_mod.x90, which runs for around 55 seconds on my laptop). It has I think the most number of kernel calls (83).
1. Removing the cached non_local_var data might cause a slowdown of 1% - very well within the runtime variation.
2. Additionally removing the variable access computation made more or less no measurable difference.
3. Removing the cached PSyIR (which means it will convert the fparser tree to PSyIR ... unless there is some hidden caching happening, I just call `generate_psyir` with the fparser tree over and over), caused 5 to 10% overhead.

Caching removed the need for computing non-local-variables and variable access data for 167 calls. So all in all, as long as we have the fparser tree, I'd think we are fine with computing this on-demand. Of course, if we consider multiple-file processing there might be more benefit :)

While I still like my approach (no surprise there :) ), it seems that caching is just not necessary at all, which means I can move the computation function into the `Routine` class.

sergisiso:
> 3. Removing the cached PSyIR (which means it will convert the fparser tree to PSyIR ... unless there is some hidden caching happening, I just call generate_psyir with the fparser tree over and over), caused 5 to 10% overhead.

Note that, if it wasn't clear, I am not against this top-level caching from file(or module)name to PSyIR , in fact I think we need need to rely more on finding the declarations that are often imported in order to do safer transformations, which will put more stress on this. What I am worried about is the caching of internal node properties instead of looking up the tree/symbol_tables, which I think make the code more complex and harder to mutate.

> > I did thought about that, and in the case of a single execution (like you are proposing with multiple files - I am undecided about this) I think the logical place is at src/psyclone/psyir/frontend/fortran.py::psyir_from_file we have the filename as the identified and all file parsing should go through here, so it can memorize previous results and return them immediately when requested.

> We still need a mapping from module name to file name. Though the ModuleManager already provides that (actually, it maps also module name to PSyIR immediately). This is already on master, so not worth discussing here :)

I did not appreciate that we already had this functionality, this is indeed similar to what I was thinking! This would be for another PR but potentially we can use the ModuleManager on src/psyclone/psyir/symbols/containersymbol.py:import_container ? 

I am still unsure about its location under src/psyclone/parse though - this directory has the remaining parts of psyclone parsing fparser directly instead of using the PSyIR and I had the own goal to end up deleting this directory (tbh this is why I initially didn't like new code added here). But this is not what the ModuleManager does. Would it make sense to move the ModuleManager files to src/psyclone/psyir/frontend (they seem frontend functionality to me)? In the long term I wouldn't want fparser dependencies to escape this folder. (This can also be another PR if this one already does enough)

hiker:
> > 3. Removing the cached PSyIR (which means it will convert the fparser tree to PSyIR ... unless there is some hidden caching happening, I just call generate_psyir with the fparser tree over and over), caused 5 to 10% overhead.
> 
> Note that, if it wasn't clear, I am not against this top-level caching from file(or module)name to PSyIR , in fact I think we need need to rely more on finding the declarations that are often imported in order to do safer transformations, which will put more stress on this. What I am worried about is the caching of internal node properties instead of looking up the tree/symbol_tables, which I think make the code more complex and harder to mutate.

Since I am obviously a victim of Knuth's  "premature optimisation" (embarrassing), I think we agree now :)


> 
> > > I did thought about that, and in the case of a single execution (like you are proposing with multiple files - I am undecided about this) I think the logical place is at src/psyclone/psyir/frontend/fortran.py::psyir_from_file we have the filename as the identified and all file parsing should go through here, so it can memorize previous results and return them immediately when requested.
> 
> > We still need a mapping from module name to file name. Though the ModuleManager already provides that (actually, it maps also module name to PSyIR immediately). This is already on master, so not worth discussing here :)
> 
> I did not appreciate that we already had this functionality, this is indeed similar to what I was thinking! This would be for another PR but potentially we can use the ModuleManager on src/psyclone/psyir/symbols/containersymbol.py:import_container ?

You mean using the ModuleManager to find things? Yes, that was the whole idea of it. Long term I would hope that any parsing is done via calls to the ModuleManager (i.e. it will manage paths etc).

> 
> I am still unsure about its location under src/psyclone/parse though - this directory has the remaining parts of psyclone parsing fparser directly instead of using the PSyIR and I had the own goal to end up deleting this directory (tbh this is why I initially didn't like new code added here). But this is not what the ModuleManager does. Would it make sense to move the ModuleManager files to src/psyclone/psyir/frontend (they seem frontend functionality to me)? In the long term I wouldn't want fparser dependencies to escape this folder. (This can also be another PR if this one already does enough)

Oh, we totally agree here - I only put it there since all parsing functions were there, and that's mostly what it is doing. I didn't know that this whole directory was scheduled for removal :) I am not sure about the best location - it seems a  bit hidden under psyir/frontend. I would even consider putting it in the root, since it might become a rather essential part of scripts (that need to find information about modules), but tbh I don't care too much. I am happy to move the ModuleManager to a different location anytime.

hiker:
As discussed, I've moved the information from RoutineInfo to Routine. I didn't bother renaming this branch (which now clearly does not add RoutineInfo anymore :) )

src/psyclone/parse/module_info.py
```diff
@@ -35,16 +35,23 @@
...
+                                     Interface_Stmt, Procedure_Stmt,
+                                     Subroutine_Subprogram, Use_Stmt)
 from fparser.two.parser import ParserFactory
-from fparser.two.utils import walk
+from fparser.two.utils import FortranSyntaxError, walk
 
-from psyclone.errors import PSycloneError
+from psyclone.errors import InternalError, PSycloneError
+from psyclone.psyir.nodes import Container, FileContainer, Routine
+from psyclone.psyir.frontend.fparser2 import Fparser2Reader
```
arporter, line 53: Alpha-order please.

src/psyclone/parse/module_info.py
```diff
@@ -92,6 +103,44 @@ def __init__(self, name, filename):
         # each module, indexed by the module names: Dict[str, Set(str)].
         self._used_symbols_from_module = None
```
arporter, line 104: No this PR but s/None/{}/ as it's a dict?

src/psyclone/parse/module_info.py
```diff
@@ -92,6 +103,44 @@ def __init__(self, name, filename):
...
 
+        # This variable will be a set that stores the name of all routines
+        # (based on fparser), # so we can test is a routine is defined
```
arporter, line 107: "...can test whether a routine is defined"

arporter: Given that the costly bit is the creation of the fparser tree, could we not just create PSyIR and examine the resulting symbol table(s)? Or are you after the extra bit of robustness/coverage gained by only relying on fparser2 and not having to convert to PSyIR?

src/psyclone/parse/module_info.py
```diff
@@ -92,6 +103,44 @@ def __init__(self, name, filename):
...
+        # without having to convert the AST to PSyIR. It is initialised with
+        # None so we avoid trying to parse a file more than once (parsing
+        # errors would cause routine_names to be empty, so we can test
+        # if routine_name is None vs if routine_names is empty)
+        self._routine_names = None
+
+        # This dictionary stores the mapping of routine name to the list
```
arporter, line 114: Again, could we make this a property of `RoutineSymbol`, e.g. we could add a `get_schedule()` method to it?

src/psyclone/parse/module_info.py
```diff
@@ -92,6 +103,44 @@ def __init__(self, name, filename):
...
+        # of PSyIR Routine nodes. It is a list since in case of a generic
+        # interface several Routine nodes are required.
+        self._psyir_of_routines = None
+
+        # This map contains the list of routine names that are part
+        # of the same generic interface.
+        self._generic_interfaces = {}
+
+        # This is a dictionary that will cache non-local symbols used in
```
arporter, line 123: Again, the PSyIR symbol table already has this functionality (or easily could have), albeit without the caching.

arporter (commented):
I think I must be seeing things differently. There appears to be a lot of duplication with the PSyIR here. I think we need a discussion on what we all want and how best to implement it.  We'll need to include @sergisiso and @rupertford too. My (current!) view is that if something can be done in the PSyIR then that's where it should be done, otherwise we get a proliferation of layers and tools.

As an example, currently ContainerSymbol contains a method that will go away and find the PSyIR of the associated Container. I was thinking (in #924) that we should do something similar for RoutineSymbol. How each of those methods works out which file to examine is then a more general problem which I thought perhaps ModuleInfo would help to solve. Perhaps what is needed is a layer that encapsulates all that is known about all of the source encountered by PSyclone. (This would also be a step towards #1859.) This layer should really only have pointers to top-level Container nodes and perhaps inquiry methods that loop over those and interrogate them as appropriate?

hiker:
> I think I must be seeing things differently. There appears to be a lot of duplication with the PSyIR here. I think we need a discussion on what we all want and how best to implement it. We'll need to include @sergisiso and @rupertford too. My (current!) view is that if something can be done in the PSyIR then that's where it should be done, otherwise we get a proliferation of layers and tools.

I don't think that's necessary. While I do not agree, if that's the way you want it, that's the way I do it.

_{some slight irony on}_
You want to move all required functionality into the nodes, i.e. make them bigger and more powerful. My view is that `psyclone.psyir` should not import anything outside of `psyclone.psyir` (with the exception of errors, configuration etc - thinks I would consider 'core' functionality, though I was not successful moving all of this into `core` :) ). Anything in `psyir.nodes`  is just a tree, which needs tools to do anything with it. To me moving more functionality into a tree node is the same issue we have/had with dynamo0p3.py, just adding methods to classes instead of adding classes to a file - moving more and more unrelated stuff in there. E.g. currently `routine.py` imports 7 symbols from 5 modules, none of which causing circular dependencies. After moving the routine_info functionality into this class, it now has 17 imports from 12 modules, including 2 circular dependencies. That is a clear indication that this is wrong imho.

But let's agree to disagree. A node should be able to display its own id, name in a graph (text or graphics), which it does :) And similarly, it should define its own implementation in Fortran and C, provide transformations that are applicable to it  (e.g. `openmp_do_me` :)  ), return if it can be executed in parallel, and return information about other modules. Hang on, only the last functionality is what you want to be moved into the node, why not the rest?
_{some slight irony off}_

Just let me know that a majority of you UK people agree (it's easier for you to quickly chat), I would then go ahead and remove the ModuleInfo class, and  merge its functionality that into the Container class instead.

I assume we keep the ModuleManager still? 

> 
> As an example, currently ContainerSymbol contains a method that will go away and find the PSyIR of the associated Container. 

Yes, ContainerSymbol duplicates code in the ModuleManager :) I didn't notice that, or I would have commented on it :) Let's have a look. First, this function causes a circular dependency, of its 7 imports, 4 are unnecessary if the ModuleManager would be used, including one circular dependency. I couldn't easily identify where this property is used, but I can't see a problem using the ModuleManager with `container.name` wherever it is needed (or the ModuleManager could also support a container symbol directly). This code in containersymbol also duplicates code in parse/kernel.py (find files for a given module name), and I don't think that can easily be avoided, since kernel.py has not yet a containersymbol. But both functions could just use the ModuleManager, and then there would be only one location where we implement how to find the Fortran file for a module name.

> I was thinking (in #924) that we should do something similar for RoutineSymbol. How each of those methods works out which file to examine is then a more general problem which I thought perhaps ModuleInfo would help to solve. Perhaps what is needed is a layer that encapsulates all that is known about all of the source encountered by PSyclone. (This would also be a step towards #1859.) This layer should really only have pointers to top-level Container nodes and perhaps inquiry methods that loop over those and interrogate them as appropriate?

That thing is called the ModuleManager, which we could also rename to file manager?? I found it more useful to use module names mostly - typically I don't need to find a file given a file name, I always have a module name/symbol and need to find its source code (yes, internally it will obviously need file info, but the tool using it doesn't care). RoutineSymbol would need import information to find the module though. The ModuleManager could obviously find a routine based on a name, but it would require to parse all files :) If you have the module name (or symbol, which includes the name), you can immediately get the information from the ModuleManager.

I have not used the ModuleManager in other places (e.g. to find kernel source code), since I didn't want to interfere with what other people are doing. I was going to raise this in a follow up ticket (e.g. atm I use the -d directive for kernels to instruct the ModuleManager where to look for kernel sources. But e.g. in order to work on LFRic, I need to specify a lot more directories than just kernel in order to find the modules, so I might also slightly change the semantics). While it worked for me in LFRic extraction, I wouldn't replace the existing mechanism to find kernel modules without discussion.

src/psyclone/parse/module_info.py
hiker: Yes, though I am not using it atm. But I have a problem with current LFRic, where log_mod.F90 contains a symbol that is only used in a write statement, i.e. a code block. So my code can't find if this variable is read or written.  In theory, we could use fparser to access these details (though for now it's much easier to work around). I am happy to convert this to PSyIR instead.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+from psyclone.psyGen import BuiltIn, Kern
+from psyclone.psyir.nodes import (Call, Container, IntrinsicCall, Reference)
+from psyclone.psyir.symbols import (ArgumentInterface, DefaultModuleInterface,
+                                    ImportInterface)
+from psyclone.psyir.tools.read_write_info import ReadWriteInfo
+
+
+# pylint: disable=too-few-public-methods
+class CallTreeUtils():
+    '''This class provides functions functions to analyse the sequence of
```
arporter, line 51: "functions to analyse"

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author J. Henrichs, Bureau of Meteorology
+# -----------------------------------------------------------------------------
+
+''' This module provides tools analyse the sequence of calls
```
arporter, line 37: "tools to analyse"

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+
+
+        '''
+        non_locals = []
+
+        outer_module = routine.ancestor(Container)
+
+        for access in routine.walk((Kern, Call, Reference)):
+            # Builtins are certainly not externals, so ignore them.
+            if isinstance(access, BuiltIn):
```
arporter, line 84: A nicety but since BuiltIn and IntrisicCall are similar concepts, you could handle them both here `isinstance(access, (Builtin, IntrinsicCall))`

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                continue
+
+            if isinstance(access, Kern):
+                # A kernel is a subroutine call from a module:
```
arporter, line 88: Please add a TODO here for #2054 as hopefully that will remove the need for this branch (it will be a `Call`).

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                    # A function defined in the same module
+                    non_locals.append(("routine", outer_module.name,
+                                       Signature(sym.name)))
+                    continue
+
+                # We don't know where the subroutine comes from
+                non_locals.append(("routine", None, Signature(sym.name)))
+                continue
+
+            sym = access.symbol
```
arporter, line 117: Comment please, e.g. "Access must be a Reference"

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+        self.get_output_parameters(read_write_info, node_list, variables_info)
+        return read_write_info
+
+    # -------------------------------------------------------------------------
+    def get_non_local_read_write_info(self, node_list, read_write_info):
+        '''Returns the information about non-local variables that are read
+        or written.
+
+        '''
+
```
arporter, line 262: Please rm this blank line.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                try:
+                    mod_info = mod_manager.get_module_info(kernel.module_name)
+                except FileNotFoundError:
+                    print(f"Could not find module '{kernel.module_name}' - "
+                          f"ignored.")
+                    continue
+
+                all_possible_routines = mod_info.resolve_routine(kernel.name)
+                for routine_name in all_possible_routines:
+                    psyir = mod_info.get_psyir(routine_name)
```
arporter, line 288: I think that `get_psyir` on a ModuleInfo object should return the psyir of the Module. That can then be queried for routines.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+        1. replacing all subroutine calls with the list of their corresponding
+            non-local symbols.
+        2. Resolving unknown types to be either function calls (which then
+            get resolved as above), or variables.
+        This is done recursively until only variable accesses remain.
+
+        The actual non-local accesses will then be added to the ReadWriteInfo
+        object.
+
+        :param todo: the information about symbol type, module_name,
```
arporter, line 304: I think this needs a less confusing name - perhaps `outstanding_nonlocals` or something?

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+            symbol_name and access information
+        :type todo: List[Tuple[str,str, str,\
```
arporter, line 306: No need for line continuation.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                continue
+            done.add(info)
+            external_type, module_name, signature, access_info = info
+            if module_name in mod_manager.ignores():
+                continue
+            if external_type == "routine":
+                if module_name is None:
+                    # We don't know where the subroutine comes from.
+                    # For now ignore this
+                    print(f"Unknown routine '{signature[0]} - ignored.")
```
arporter, line 332: Please add a TODO for #11 here (logging).

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                    continue
+                try:
+                    mod_info = mod_manager.get_module_info(module_name)
+                except FileNotFoundError:
+                    print(f"Cannot find module '{module_name}' - ignored.")
```
arporter, line 337: #11 here too please. In fact, I think these print statements ought to be removed as they don't have any contextual information and will be confusing to a user. Alternatively, please could you prefix them with e.g. "Attempting to resolve non-local accesses [in xxx] - cannot find ...."

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                    continue
+                try:
+                    # Add the list of non-locals to our todo list:
+                    todo.extend(self.get_non_local_symbols(
+                        mod_info.get_psyir(signature[0])))
+                except KeyError:
+                    print(f"Cannot find symbol '{signature[0]}' in module "
```
arporter, line 344: More context and #11 please.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+                          f"'{module_name}' - ignored.")
+                continue
+
+            if external_type == "unknown":
+                # It could be a function (TODO #1314) or a variable. Check if
+                # there is a routine with that name in the module information:
+                try:
+                    mod_info = mod_manager.get_module_info(module_name)
+                except FileNotFoundError:
+                    print(f"Cannot find module '{module_name}' - ignoring "
```
arporter, line 354: More context and #11 please.

arporter (changes requested):
I appreciate this is now going to get caught up in #2462 but I think it's looking pretty good. There's still some Routine-related stuff that I think should be delegated/left to the PSyIR but apart from that I'm happy. My feeling from this PR and #2413 is that a `Container` (or module) is as far up as the PSyIR can go. Above that then sits the ModuleManager that lets us navigate between the different modules/containers in the application. I guess the ModuleManager could cache some information about a Container such as the Routines it defines. Or, we give the Container node a method that returns these? @sergisiso ?

src/psyclone/parse/module_manager.py
```diff
@@ -307,11 +330,14 @@ def sort_modules(module_dependencies):
             # Take a copy so we can modify the original set of dependencies:
             dependencies_copy = dependencies.copy()
             for dep in dependencies_copy:
-                if dep not in todo:
+                if dep in todo:
+                    continue
+                # Print a warning if this module is not supposed to be ignored
+                if dep not in self.ignores():
+                    # TODO 2120: allow a choice to abort or ignore.
                     print(f"Module '{module}' contains a dependency to "
```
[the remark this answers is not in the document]
hiker: This message is only printed if we know that 'mod' depends on 'dep', but we do not have 'dep' in our list of modules. If we would ignore this here, the sorting later would not work (since this module can never have all its dependencies satisfied). A module that has no dependencies will not print this messages (since the outer loop `for dep in dependencies_copy` has an empty list).

src/psyclone/parse/module_manager.py
```diff
@@ -243,6 +264,9 @@ def get_all_dependencies_recursively(self, all_mods):
         while todo:
             # Pick one (random) module to handle:
             module = todo.pop()
+            # Ignore any modules that we were asked to ignore
+            if module in self.ignores():
```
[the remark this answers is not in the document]
hiker: Very observant. I had noticed that, and fixed it in the 'next' PR (to avoid modifying this while a review is in progress :) ), but I've backported it now :)

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,415 @@
...
+        # First collect all non-local symbols from the kernels called. They
+        # are collected in the todo list. This list will initially contain
+        # unknown accesses, since at this stage we cannot always differentiate
+        # between function calls and array accesses. In the resolve step
+        # that is following the corresponding modules will be queried and
+        # the right accesses (functions or variables) will be used.
+        todo = []
+        mod_manager = ModuleManager.get()
+        for node in node_list:
+            for kernel in node.walk(Kern):
```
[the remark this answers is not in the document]
hiker: I am not sure, do we need to distinguish between language-level and high-level PSyIR? In some case (elsehwere, can't remember where now, kernel extraction likely), I handle kernels and calls in the same branch. So if this can be maintained (I would expect that we need to adjust things a bit when we use nemo with the module manager), I would suggest to leave this for now. And then add a separate 'renaming/moving' PR, since we also agreed that the ModuleManager etc should be elsewhere, not in parse. Let's sort the proper location out during this PR, which should make the history a lot easier to read.

src/psyclone/parse/module_info.py
```diff
@@ -134,12 +183,63 @@ def get_parse_tree(self):
...
+
             reader = FortranStringReader(self.get_source_code())
             parser = ParserFactory().create(std="f2008")
             self._parse_tree = parser(reader)
+
+            # First collect information about all subroutines/functions.
+            # Store information about generic interface to be handled later
+            # (so we only walk the tree once):
+            all_generic_interfaces = []
+            for routine in walk(self._parse_tree, (Function_Subprogram,
```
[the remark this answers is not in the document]
hiker: As indicated below, we can only do this once we have generic interface support. I urgently need this (for LFRic), so I've added a #TODO here.

src/psyclone/parse/module_info.py
```diff
@@ -134,12 +183,63 @@ def get_parse_tree(self):
...
+                                                   Interface_Block)):
+                if isinstance(routine, Interface_Block):
+                    all_generic_interfaces.append(routine)
+                else:
+                    routine_name = str(routine.content[0].items[1])
+                    self._routine_names.add(routine_name)
+
+            # Then handle all generic interfaces and add them to
+            # _generic_interfaces:
+            for interface in all_generic_interfaces:
```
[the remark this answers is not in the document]
hiker: Agreed, #todo added.

src/psyclone/parse/module_info.py
```diff
@@ -134,12 +183,63 @@ def get_parse_tree(self):
...
+                # Get the name of the interface from the Interface_Stmt:
+                name = str(walk(interface, Interface_Stmt)[0].items[0]).lower()
+                self._routine_names.add(name)
+
+                # Collect all specific functions for this generic interface
+                routine_names = []
+                for proc_stmt in walk(interface, Procedure_Stmt):
```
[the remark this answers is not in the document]
hiker: I admit I didn't look at the standard. Again, #todo - this problem should go away with generic interface support in PSyIR.

src/psyclone/parse/module_info.py
```diff
@@ -134,12 +183,63 @@ def get_parse_tree(self):
...
+                    # Convert the items to strings:
+                    routine_names.extend([str(i) for i in
+                                          proc_stmt.items[0].items])
+                self._generic_interfaces[name] = routine_names
+
         return self._parse_tree
 
+    # ------------------------------------------------------------------------
+    def contains_routine(self, routine_name):
```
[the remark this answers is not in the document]
hiker: Nothing much I can do about this while #2413 is not on trunk. #todo added.

src/psyclone/parse/module_info.py
```diff
@@ -210,3 +309,88 @@ def get_used_symbols_from_modules(self):
...
+
+        :returns: PSyIR representing this module.
+        :rtype: List[:py:class:`psyclone.psyir.nodes.Node`]
+
+        '''
+        if self._psyir is None:
+            try:
+                self._psyir = \
+                    self._processor.generate_psyir(self.get_parse_tree())
+            except (KeyError, SymbolError, InternalError, FortranSyntaxError):
```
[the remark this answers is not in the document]
hiker: I have a separate ticket for error handling. I totally agree that we should not ignore errors. But for now this is extremely useful for me, since I often run the kernel extraction across all of LFRic to find errors, and since there are files with 10, 20 or more kernels, it is really useful for me to get the errors printed, but the script keeps on testing ALL kernels at once. I think I have all locations marked with TODO #2120 for this.

src/psyclone/parse/module_info.py
```diff
@@ -210,3 +309,88 @@ def get_used_symbols_from_modules(self):
...
+            self._psyir_of_routines = {}
+            for routine in self._psyir.walk(Routine):
+                self._psyir_of_routines[routine.name.lower()] = routine
+
+        if routine_name is not None:
+            return self._psyir_of_routines[routine_name.lower()]
+        return self._psyir
+
+    # ------------------------------------------------------------------------
+    def resolve_routine(self, routine_name):
```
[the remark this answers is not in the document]
hiker: Again, #todo added.

src/psyclone/parse/module_info.py
```diff
@@ -210,3 +309,88 @@ def get_used_symbols_from_modules(self):
...
+        routine_name = routine_name.lower()
+        if routine_name not in self._generic_interfaces:
+            return [routine_name]
+
+        # If a generic interface name is queried, return a copy
+        # of all possible routine names that might be called:
+        return self._generic_interfaces[routine_name][:]
+
+    # ------------------------------------------------------------------------
+    def get_symbol(self, name):
```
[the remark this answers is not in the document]
hiker: Neither am I .... it's actually not used. I'd guess it comes with one of the next PRs, but I have removed it, and will remove its usage wherever it might be :)

src/psyclone/parse/module_info.py
```diff
@@ -148,16 +248,15 @@ def _extract_import_information(self):
...
-                                  f"twice.")
         # Initialise the caches:
         self._used_modules = set()
         self._used_symbols_from_module = {}
 
-        parse_tree = self.get_parse_tree()
+        try:
+            parse_tree = self.get_parse_tree()
+        except FortranSyntaxError:
+            # Hide syntax errors
```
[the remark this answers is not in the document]
hiker: Asa above, TODO #2120 added.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Indeed.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: I thought we had agreed that tools would act on the PSyIR, and not adding these functions to the PSyIR (to keep it small). But no problem, I've changed it by moving this function into Container.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done. It also has #2120 (for error handling)

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

src/psyclone/psyir/tools/call_tree_utils.py
hiker: Done.

hiker:
I had to do a bit of additional refactoring, to better prepare for supporting multiple modules in one file. It should be ok now.

I am not sure if you expect me to do more about the potential overlap we have - e.g I've left `contains_routine` in ModuleInfo, though it could be done using `Container.get_routine_psyir(NAME)` instead. Though the latter required the creation of the PSyIR, while I think atm that part at least works with fparser info only. Not sure if it makes much of a difference (well, it might actually make a difference if we have PSyIR parsing problems e.g. a namelist statement turns the whole file into a codeblock, so we can't get anything atm - but with the fparser based test we do get information about names).

Ready for next review.

sergisiso:
> I guess the ModuleManager could cache some information about a Container such as the Routines it defines. Or, we give the Container node a method that returns these? @sergisiso ?

I would stay away of caching these, I manage to convince @hiker about this from his first proposal because all cost is on the parsing part, traversing the tree lookig for information is relatively cheap.
At some point we need to look at caching between psyclone executions or executing a psyclone command with multiple files to process, this will make a difference but is a different PR altogether.

src/psyclone/parse/module_info.py
arporter: I meant that the lines should be ordered by the modules they import from but I don't care that much :-)

src/psyclone/parse/module_info.py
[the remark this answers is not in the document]
arporter: OK, thanks.

src/psyclone/parse/module_info.py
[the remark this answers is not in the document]
arporter: I spy no #todo?

src/psyclone/psyir/nodes/container.py
```diff
@@ -170,6 +170,23 @@ def node_str(self, colour=True):
     def __str__(self):
         return f"Container[{self.name}]\n"
 
+    def get_routine_psyir(self, name):
```
arporter, line 173: Thanks for adding this.

src/psyclone/psyir/tools/call_tree_utils.py
arporter: I meant s/todo/outstanding_nonlocals/ :-) I was quite happy with the routine name - please could you revert it.

src/psyclone/psyir/tools/call_tree_utils.py
arporter: Sorry to be so picky but could you add the missing space after the first "str," in that line.

src/psyclone/parse/module_manager.py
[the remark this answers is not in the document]
arporter: Oh I see. I thought I must be missing something.

src/psyclone/tests/conftest.py
```diff
@@ -233,3 +234,22 @@ def fixture_fortran_reader():
 def fixture_fortran_writer():
     '''Create and return a FortranWriter object with default settings.'''
     return FortranWriter()
+
+
+@pytest.fixture(scope='function')
+def clear_module_manager_instance():
+    ''' The tests in this module all assume that there is no pre-existing
```
arporter, line 241: docstring needs re-writing slightly now that it's in the top-level confteset.py.

src/psyclone/tests/parse/module_info_test.py
```diff
@@ -180,3 +175,120 @@ def test_mod_info_get_used_symbols_from_modules():
...
+    # Test that a file that can't be converted to PSyIR returns an
+    # empty FileContainer.
+    mod_man.add_search_path(dyn_path, recursive=False)
+    # The file 'broken_builtins_mod.f90' contains invalid Fortran and
+    # cannot be parsed:
+    broken_builtins = mod_man.get_module_info("broken_builtins_mod")
+    broken_builtins_psyir = broken_builtins.get_psyir()
+
+    # We should still get an empty Container with a dummy Container:
+    assert isinstance(broken_builtins_psyir, Container)
```
arporter, line 210: I know you want to run across all of LFRic without aborting with an error but I'm a bit concerned that returning an apparently valid `Container`, even when the code couldn't be parsed, could be confusing - I can imagine a user thinking that the module has been found and a Container created when in fact it's just an empty container because of a previous error. Presumably, in such a case one of the error messages I looked at earlier should be printed out? If so, please could you check for this here. If not, we need to think about how to help a user in this situation!

src/psyclone/tests/parse/module_info_test.py
```diff
@@ -180,3 +175,120 @@ def test_mod_info_get_used_symbols_from_modules():
...
+
+
+# -----------------------------------------------------------------------------
+@pytest.mark.usefixtures("change_into_tmpdir", "clear_module_manager_instance",
+                         "mod_man_test_setup_directories")
+def test_generic_interface():
+    '''Tests that a generic interface works as expected.
+    '''
+    mod_man = ModuleManager.get()
+    mod_man.add_search_path("d1")
```
arporter, line 223: Worth a comment explaining that this relies on the directories and files setup by the `mod_man_test_setup_directories` fixture.

src/psyclone/tests/parse/module_info_test.py
```diff
@@ -180,3 +175,120 @@ def test_mod_info_get_used_symbols_from_modules():
...
+    assert mod_info.contains_routine("myfunc1")
+    assert mod_info.contains_routine("myfunc2")
+
+    assert "myfunc" in mod_info._generic_interfaces
+
+
+# -----------------------------------------------------------------------------
+@pytest.mark.usefixtures("change_into_tmpdir", "clear_module_manager_instance",
+                         "mod_man_test_setup_directories")
+def test_resolve_routine():
```
arporter, line 239: This needs a TODO as it will be removed/moved when Container has functionality to query for a Routine.

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
```diff
@@ -0,0 +1,443 @@
...
+             "dummy_module_variable"),
+            ('routine', 'testkern_import_symbols_mod', "local_func"),
+            ("routine", "module_with_var_mod", "module_subroutine"),
+            ("unknown", "module_with_var_mod", "module_var_a"),
+            ("routine", "testkern_import_symbols_mod", "local_subroutine"),
+            ("routine", None, "unknown_subroutine")]
+            )
+    assert non_locals_without_access == expected
+
+    # Check the handling of a symbol that is not found: _compute_non_locals
```
arporter, line 190: "_compute_all_non_locals"

arporter (changes requested):
Getting there now. There's going to be a lot of sorting out needed once #2413 and #2435 make it on but that at least will be confined to the tooling you're introducing here.
You'll see I've asked for a lot more comments, especially in the new Fortran test files.
I'll check building the various docs next time around.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/module_with_builtin_mod.f90
```diff
@@ -0,0 +1,57 @@
...
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!-------------------------------------------------------------------------------
+! Author J. Henrichs, Bureau of Meteorology
+
+module module_with_builtin
+
+contains
+
```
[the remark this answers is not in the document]
hiker: Yes, my tests are a bit of a mess, because I tried to pull out a smaller part of the actual remote-variable support (e.g. I believe there might even be some files here that are not used, but they are used in the the upcoming #1990 - likely during the refactoring (do you remember how this started??) I had to remove tests that used them, but they should still be required in #1990.

So in short, in #1990 there is an additional test that verifies the remote symbols used here, so that's why I want to have this specific case here.

config/psyclone.cfg
```diff
@@ -53,6 +53,10 @@ VALID_PSY_DATA_PREFIXES = profile, extract, read_only_verify, nan_test
 # the mpirun/mpiexec ranks_per_node parameter must match this number.
 OCL_DEVICES_PER_NODE = 1
 
+# Symbols imported from the following modules will be ignored when parsing
```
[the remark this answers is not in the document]
hiker: Done. I also added some other missing stuff.

src/psyclone/configuration.py
```diff
@@ -379,6 +380,14 @@ def load(self, config_file=None):
         # unique name (within the specified kernel-output directory).
         self._kernel_naming = Config._default_kernel_naming
 
+        ignore_modules = self._config['DEFAULT'].getlist("IGNORE_MODULES", [])
+        # Avoid circular import
+        # pylint: disable=import-outside-toplevel
+        from psyclone.parse import ModuleManager
```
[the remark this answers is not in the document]
hiker: I believe we agreed to move this at some later stage (to make diffs easier)

src/psyclone/parse/module_info.py
```diff
@@ -35,16 +35,23 @@
 
 '''This module contains the ModuleInfo class, which is used to store
 and cache information about a module: the filename, source code (if requested)
-and the fparser tree (if requested).
+and the fparser tree (if requested), and information about routine it
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/parse/module_info.py
hiker: I am using this to distinguish between 'not yet filled' (None) and 'no symbols' ({}).

src/psyclone/parse/module_info.py
hiker: Removed.

src/psyclone/parse/module_info.py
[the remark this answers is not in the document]
hiker: Ah yes, I had added this in the code above where the variable was defined. I've added one more ... that's now the script with the most TODO's :)

src/psyclone/psyir/tools/call_tree_utils.py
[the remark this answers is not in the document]
hiker: I've added a new issue for bringing the LFRic functionality back into gocean and 'generic' in #2494.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -0,0 +1,430 @@
...
+                    mod_info = mod_manager.get_module_info(kernel.module_name)
+                except FileNotFoundError:
+                    # TODO #11: Add proper logging
+                    # TODO #2120: Handle error
+                    print(f"[CallTreeUtils.get_non_local_read_write_info] "
+                          f"Could not find module '{kernel.module_name}' - "
+                          f"ignored.")
+                    continue
+
+                all_possible_routines = mod_info.resolve_routine(kernel.name)
```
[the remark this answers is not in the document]
hiker: todo added

src/psyclone/tests/parse/module_info_test.py
hiker: I've printed the error message now. Proper handling of errors will be done (for the existing issue)

src/psyclone/tests/parse/module_info_test.py
hiker: Done.

src/psyclone/tests/parse/module_info_test.py
hiker: Done.

src/psyclone/tests/parse/module_info_test.py
```diff
@@ -180,3 +175,120 @@ def test_mod_info_get_used_symbols_from_modules():
...
+
+# -----------------------------------------------------------------------------
+@pytest.mark.usefixtures("change_into_tmpdir", "clear_module_manager_instance",
+                         "mod_man_test_setup_directories")
+def test_module_info_extract_import_information_error():
+    '''Test handling of files that cannot be parsed in
+    _extract_import_information.
+    '''
+    mod_man = ModuleManager.get()
+    mod_man.add_search_path("d2")
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/module_call_tree_mod.f90
```diff
@@ -0,0 +1,100 @@
...
+! FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author: J. Henrichs, Bureau of Meteorology
+
+module module_call_tree_mod
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/module_renaming_external_var_mod.f90
```diff
@@ -0,0 +1,44 @@
...
+! FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author: J. Henrichs, Bureau of Meteorology
+
+module module_renaming_external_var_mod
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/module_with_name_clash_mod.f90
```diff
@@ -0,0 +1,46 @@
...
+! DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+! FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author: J. Henrichs, Bureau of Meteorology
+
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/module_with_var_mod.f90
```diff
@@ -0,0 +1,49 @@
...
+! FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author: J. Henrichs, Bureau of Meteorology
+
+module module_with_var_mod
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/testkern_import_symbols_error_mod.f90
```diff
@@ -0,0 +1,89 @@
...
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Authors: R. W. Ford and A. R. Porter, STFC Daresbury Lab
+! Modified: I. Kavcic, Met Office
+! Modified: J. Henrichs, Bureau of Meteorology
+
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/testkern_import_symbols_mod.f90
```diff
@@ -0,0 +1,107 @@
...
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Authors: R. W. Ford and A. R. Porter, STFC Daresbury Lab
+! Modified: I. Kavcic, Met Office
+! Modified: J. Henrichs, Bureau of Meteorology
+
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/testkern_import_symbols_name_clash_mod.f90
```diff
@@ -0,0 +1,89 @@
...
+! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Authors: R. W. Ford and A. R. Porter, STFC Daresbury Lab
+! Modified: I. Kavcic, Met Office
+! Modified: J. Henrichs, Bureau of Meteorology
+
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/configuration.py
```diff
@@ -379,6 +380,14 @@ def load(self, config_file=None):
...
+        mod_manager = ModuleManager.get()
+        for module_name in ignore_modules:
+            mod_manager.ignore_module(module_name)
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/parse/module_info.py
hiker: That's a todo for now till we have support for interfaces.

src/psyclone/parse/module_info.py
hiker: That also need support for interfaces to fully replace the current functionality, todo is there.

src/psyclone/parse/module_manager.py
```diff
@@ -142,6 +143,23 @@ def _add_all_files_from_dir(self, directory):
...
+
+        :param str module_name: name of the module to ignore.
+
+        '''
+        self._ignore_modules.add(module_name.lower())
+
+    # ------------------------------------------------------------------------
+    def ignores(self):
+        ''':returns: the set of modules to ignore.
+        :rtype: Set[str]
```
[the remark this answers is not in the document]
hiker: Done - likely not all of them in every file I touched though, but I started :)

src/psyclone/tests/config_test.py
```diff
@@ -781,3 +783,33 @@ def test_config_class_initialised(monkeypatch):
...
+    ModuleManager. '''
+
+    mod_manager = ModuleManager.get()
+    monkeypatch.setattr(mod_manager, "_ignore_modules", set())
+    config_file = tmpdir.join("config")
+    get_config(config_file, _CONFIG_CONTENT)
+
+    assert mod_manager.ignores() == {'mpi', 'netcdf'}
+
+    # Make sure if works if IGNORE_MODULES is not specified at all by checking
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/conftest.py
hiker: Done.

src/psyclone/tests/parse/module_manager_test.py
```diff
@@ -271,30 +255,54 @@ def test_mod_manager_get_all_dependencies_recursively(capsys):
...
     deps = {"a": {"b", "c"}, "b": set(), "c": {"b"}}
-    assert ModuleManager.sort_modules(deps) == ["b", "c", "a"]
+    assert mod_man.sort_modules(deps) == ["b", "c", "a"]
 
     deps = {"a": {"b", "c"}, "b": set(), "c": {"netcdf", "b"}}
-    deps_sorted = ModuleManager.sort_modules(deps)
+    deps_sorted = mod_man.sort_modules(deps)
     assert deps_sorted == ["b", "c", "a"]
     out, _ = capsys.readouterr()
     assert ("Module 'c' contains a dependency to 'netcdf', for which we "
```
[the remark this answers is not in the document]
hiker: I've rewritten the warning, I think it's better now.

src/psyclone/tests/parse/module_manager_test.py
```diff
@@ -308,3 +316,31 @@ def test_mod_man_sort_modules(capsys):
...
+    mod_man.add_search_path("d1")
+    mod_man.add_search_path("d2")
+
+    # First finds a_mod, which will parse the first directory
+    mod_man.ignore_module("a_mod")
+    mod_info = mod_man.get_module_info("a_mod")
+    assert mod_info is None
+    assert "a_mod" in mod_man.ignores()
+
+    mod_info = mod_man.get_module_info("d_mod")
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/tests/psyir/nodes/container_test.py
```diff
@@ -38,15 +38,14 @@
...
 
-from __future__ import absolute_import
 import pytest
-from psyclone.psyir.nodes import Container, Return, KernelSchedule, \
-    FileContainer
-from psyclone.psyir.symbols import SymbolTable, DataSymbol, REAL_SINGLE_TYPE
+
 from psyclone.errors import GenerationError
 from psyclone.psyir.backend.fortran import FortranWriter
+from psyclone.psyir.nodes import colored, Container, FileContainer, \
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/psyir/nodes/container_test.py
```diff
@@ -180,3 +179,21 @@ def test_container_children_validation():
     assert ("Item 'Return' can't be child 1 of 'Container'. The valid format"
             " is: '[Container | Routine | CodeBlock]*'."
             "" in str(excinfo.value))
+
+
+def test_container_get_psyir():
+    '''Test that get_psyir works
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/invoke_kernel_with_imported_symbols_error.f90
```diff
@@ -0,0 +1,50 @@
...
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!-------------------------------------------------------------------------------
+! Author J. Henrichs, Bureau of Meteorology
+
+program invoke_kernel_with_imported_symbols
+
+  ! Description: single function specified in a named invoke call
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/test_files/dynamo0p3/driver_creation/invoke_kernel_with_imported_symbols.f90
```diff
@@ -0,0 +1,52 @@
...
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!-------------------------------------------------------------------------------
+! Author J. Henrichs, Bureau of Meteorology
+
+program invoke_kernel_with_imported_symbols
+
+  ! Description: single function specified in a named invoke call
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
```diff
@@ -0,0 +1,443 @@
...
+                     'module_var:READ(0)'),
+                    ("reference", "g_mod", Signature("module_var"),
+                     'module_var:WRITE(0)')])
+    # Convert the access info to a string for easy comparison:
+    assert (set((i[0], i[1], i[2], str(i[3])) for i in all_non_locals) ==
+            expected)
+
+
+# -----------------------------------------------------------------------------
+def test_call_tree_utils_inout_parameters_nemo(fortran_reader):
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
```diff
@@ -0,0 +1,443 @@
...
+        subroutine myfunc2() ...
+            module_var = p + module_var_2
+        end subroutine myfunc2
+    So they both use the module variable module_var, but myfunc1 reads it,
+    myfunc2 writes it. '''
+    mod_man = ModuleManager.get()
+    mod_man.add_search_path("d2")
+    mod_info = mod_man.get_module_info("g_mod")
+    ctu = CallTreeUtils()
+    # ctu.get_non_local_symbols(mod_info.get_psyir("myfunc"))
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
hiker: Oops.

hiker:
I think  have addressed all issues so far, though I couldn't reply to some of the comments, no idea why, there was just no option.

A lot things added up as TODO's, with the interface support still coming (and me needing interfaces for current LFRic work), I couldn't see a way around that. Also the error handling is still a TODO - just a side-effect of this being part of a 'work-in-progress' PR :(

Additionally, I think some of the test files under `test_files/dynamo/driver` are not used atm - but I couldn't see an easy way of removing them, without removing them then also in the next PR, which will need them.

src/psyclone/parse/module_info.py
arporter: Please could you add a TODO for #2435

src/psyclone/parse/module_info.py
arporter: TODO #2435 please.

src/psyclone/parse/module_info.py
```diff
@@ -96,15 +96,15 @@ def __init__(self, name, filename):
         self._psyir = None
 
         # A cache for the module dependencies: this is just a set
-        # of all modules used by this module. Type: Set[str]
+        # of all modules used by this module. Type: set[str]
         self._used_modules = None
 
         # This is a dictionary containing the sets of symbols imported from
-        # each module, indexed by the module names: Dict[str, Set(str)].
+        # each module, indexed by the module names: dict[str, set[str].
```
arporter, line 103: Missing closing `]` please.

src/psyclone/tests/parse/module_manager_test.py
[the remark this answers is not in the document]
arporter: Thanks :-)

arporter (changes requested):
Thanks for all of that Joerg. It's pretty much there now - just a few minor things.
Integration tests are running.
All docs build fine.

src/psyclone/parse/module_info.py
hiker: Done.

src/psyclone/parse/module_info.py
hiker: Done ... I think. The `_psyir_of_routines` has been removed, I don't know why it is still shown here. But I believe that todo also belongs to the `_generic_interfaces` below, so I have added it there.

src/psyclone/parse/module_info.py
hiker: Done.

hiker:
Integration tests triggered and all passed, and CI here as well.
Ready for next review.

arporter (approved):
Will proceed to merge.

arporter:
Link-check failure was just bitbucket having a moment.

Issue #1960, opened 2022-11-17
Canonicalise WHERE with reduction intrinsics, which currently are placed into CodeBlocks


sergisiso:
In ECMWF NEMO sbccpl.f90 this WHERE construct.
```
WHERE( picefr(:,:) > 1.e-10 ) ; zevap_ice(:,:,1) = frcv(jpr_ievp)%z3(:,:,1) * SUM( a_i_last_couple, dim=3 ) / picefr(:,:)
ELSEWHERE                     ; zevap_ice(:,:,1) = 0._wp
END WHERE
```
is converted to:
```
do widx2_2 = 1, SIZE(picefr, 2), 1
   do widx1_2 = 1, SIZE(picefr, 1), 1
      if (picefr(widx1_2,widx2_2) > 1.e-10) then                                                                                        1               
          zevap_ice(widx1_2,widx2_2,1) = frcv(jpr_ievp)%z3(widx1_2,widx2_2,1) * SUM(a_i_last_couple, dim=3) / picefr(widx1_2,widx2_2)
      else
          zevap_ice(widx1_2,widx2_2,1) = 0._wp
      end if
   enddo
enddo
```
Which produces the compiler error: `The shapes of the array expressions do not conform.   [ZEVAP_ICE]`

I am not sure what the error is, but it is a nested WHERE (#1651 ?) with array elements ( #717 ?) and it uses 1, SIZE instead of L/UBound which could cause troubles with non-standard bounds.

UPDATE: This doesn't produce invalid code anymore, but it is still a CodeBlock

arporter:
Simon has also reported:

    where(sum(v1(:), dim=2) > 0.0)   ! -> Issue 3: "do widx1 = 1, SIZE(v1, 1), 1" has incorrect bounds, it should be
                                     !             "do widx1 = lbound(v1, 1), rbound(v1, 1), 1"
       v1(:) = 1.0
    end where

    where(sum(v2(:,:), dim=2) > 0.0) ! -> Issue 4: the reduction "sum(v2(:,:), dim=2)" is ignored
      v3(:) = 1.0
    end where

arporter:
Looking at the intrinsics in fparser, I realise we are going to have similar problems with MAXVAL, MINVAL and PRODUCT.

sergisiso:
As we talk about, these issues are also present in the arrayrange2loop transformation, although we correctly refuse them in the validate because they contain non-elemental intrinsics.

If we could convert
```
[...]

Issue #949, opened 2020-10-23
PSyIR frontend support for array declarations that are DataNodes (expressions).


rupertford:
PR #947 introduced support for expressions in array declarations. However the fortran front end needs to be updated to make use of this new functionality

arporter:
I'm closing this as it's a duplicate of #2138.

PR #2372, opened 2023-10-19
(Towards #1960) put WHEREs containing reductions into CodeBlocks and fix non-unity lower-bound bug.

arporter:


arporter:
I've asked Simon if he'd mind testing this branch with the files on master of NEMO that have been giving him trouble.

arporter:
There's a separate bug that causes us to generate incorrect loop bounds if the array being iterated over has a non-default lower bound. We currently always generate, e.g.:

          do widx2 = 1, SIZE(sjk, 2), 1
            do widx1 = 1, SIZE(sjk, 1), 1
              if (sjk(widx1,widx2,jn) /= 0._wp) then
                r1_sjk(widx1,widx2,jn) = 1._wp / sjk(widx1,widx2,jn)
              end if
            enddo
          enddo

while we *should* generate e.g. `do widx2 = LBOUND(sjk,2), UBOUND(sjk,2), 1`

arporter:
Just checked with SPITZ12 (as that includes ice) and, with master, we get a compilation failure:

     nemo/cfgs/SPITZ_TEST/BLD_SCT_PSYCLONE/obj/iceistate.f90
NVFORTRAN-S-0074-Illegal number or type of arguments to sum - keyword argument array (/home/aporter/NEMO/nemo/cfgs/SPITZ_TEST/BLD_SCT_PSYCLONE/obj/iceistate.f90: 297)
  0 inform,   0 warnings,   1 severes, 0 fatal for ice_istate

arporter:
I've realised that even moving to `LBOUND`/`UBOUND` isn't guaranteed to work because the mask expression only has to have the same *shape* as any arrays assigned to within a `where-assignment-stmt`:
image
and shape is defined as:
image
i.e. the shape does *not* include the lower bounds of the rank(s) of an array.

arporter:
Therefore, the generated loops should be over the *shape* of the mask expression. Whenever we index into an array (in a `where-assignment-statement`), that indexing should take into account the lower bound of the array.

arporter:
Tests are now happy (incl. with compilation). Need to test NEMO passthrough now.

arporter:
NEMO integration tests are green and performance is good :-)

arporter:
I can build GYRE_PISCES and SPITZ12 on NEMO main with the excludes that mention 'where' commented out of `mk/sct_psyclone.sh`. The GYRE_PISCES case also runs fine (single MPI process). I don't have input files for SPITZ12 for this version of NEMO.

arporter:
Realised that `process_nemo.py` had a TODO for this issue for `sbccpl.f90` - update script (to include this file) and re-run integration tests...

arporter:
Ready for first review, probably from @sergisiso. Fixes bugs in the WHERE handling and also improves the `datatype` method of ArrayReference so that it does a bit better when lacking intrinsic type information. This will need another round of integration tests as I've changed it a bit since the last lot were green. In fact, I'll fire them off now.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3650,14 +3654,31 @@ def _array_syntax_to_indexed(self, parent, loop_vars):
...
+                            [base_ref.copy(),
+                             ("dim", Literal(str(idx+1), INTEGER_TYPE))])
+                    else:
+                        lbound = shape[range_idx].lower.copy()
+                    # Create the index expression:
+                    #    idx-expr = array-lower-bound + loop-idx - 1
+                    expr = BinaryOperation.create(
+                        add_op, lbound, Reference(symbol))
+                    expr2 = BinaryOperation.create(sub_op, expr, one.copy())
+                    array.children[idx] = expr2
```
sergisiso, line 3681: Although this is technically correct, most of the time array-lower-bound is 1 and we end up with "1 + loop-idx - 1". Could we create a special case for this to avoid the "+1-1" clutter

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -327,7 +327,11 @@ def test_where_within_loop(fortran_reader):
     assert isinstance(where_loop.loop_body[0], IfBlock)
     assign = where_loop.loop_body[0].if_body[0]
     assert isinstance(assign, Assignment)
-    assert [idx.name for idx in assign.lhs.indices] == ["widx1", "jl"]
+    assert assign.lhs.indices[0].debug_string() == "1 + widx1 - 1"
+    assert assign.lhs.indices[1].debug_string() == "jl"
+    assert where_loop.start_expr.value == "1"
+    assert (where_loop.stop_expr.debug_string() ==
+            "(UBOUND(var, dim=1) - LBOUND(var, dim=1)) / 1 + 1")
```
sergisiso, line 334: Again, to improve the readability of the output, can we make the stop_expr `SIZE(var, dim=1)` when the lbound is known and it is "1"?

And I would like to see a separate test with a where statement with different lower bounds (known at compile time and not known), to see how the loop lower and upper bound are set then.

Also, with the mask dimension (which I think is used for the loop bounds) and the array in the body of the where with the range shifted (different lower bounds), as I think is one of the issues mentioned in the PR.

sergisiso (changes requested):
@arporter I like the new implementation, I just want to see some more explicit tests of dealing with lower bounds and (only if its feasable) improve a bit the output code for simple cases.

examples/nemo/scripts/process_nemo.py
```diff
@@ -80,8 +80,6 @@
 NOT_WORKING = [
     # Array accessed inside WHERE does not use array notation
     "diurnal_bulk.f90",
-    # TODO #1960: There is a bug on a WHERE fparser2 frontend processing
-    "sbccpl.f90",
 ]
```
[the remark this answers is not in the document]
arporter: Good point. It's #717 (I'll add a TODO) and I've reviewed that list of issues and managed to close one :-)

src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -129,7 +129,9 @@ def datatype(self):
         '''
         shape = self._get_effective_shape()
         if shape:
-            if isinstance(self.symbol.datatype, ArrayType):
+            if type(self.symbol) is Symbol:
+                orig_shape = []
```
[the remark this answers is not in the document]
arporter: It's not nit picking if it's hard to understand what I wrote :-)

src/psyclone/psyir/nodes/structure_reference.py
```diff
@@ -325,9 +325,10 @@ def datatype(self):
                 cursor_type = cursor_type.intrinsic
             if isinstance(cursor_type, DataTypeSymbol):
                 cursor_type = cursor_type.datatype
-            cursor_type = cursor_type.components[cursor.name].datatype
-            if isinstance(cursor_type, (UnknownType, DeferredType)):
-                return DeferredType()
+            if not isinstance(cursor_type, (DeferredType, UnknownType)):
+                # Once we've hit a Deferred/UnknownType the cursor_type will
```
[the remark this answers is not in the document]
arporter: No worries. It wasn't too bad :-)

src/psyclone/psyir/frontend/fparser2.py
arporter: Done.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
arporter: Oh yes, now that I've done your suggestion and special-cased the "+1-1", I realise I don't have a test for anything else!

arporter: I also realise that I've complicated things unnecessarily here:

    integer, dimension(4:5,-2:8) :: array

    write (*,*) 'First rank size = ', size(array, 1)
    write (*,*) 'Second rank size = ', size(array, 2)

gives:

    First rank size =            2
    Second rank size =           11

so I can use `SIZE` rather than `UBOUND` and `LBOUND`.

arporter: In fact, it turned out that adding some tests revealed a quite significant bug - I'd been using the lower bound given by the `shape` of the datatype of the access and that is always either unknown or unity. (Because `shape` doesn't capture lower bound information.) I've fixed this now but also rediscovered #949 which means that we can't yet query a declaration if it has non-standard lower bounds. This means we have to resort to LBOUND more than we would like but that can be improved in future.

arporter:
I've run the integration tests and they were all green and performance was fine.

arporter:
I just need to check again with HEAD of nemo now...

arporter:
GYRE builds and runs and SPITZ12 builds :-) Need to look at coverage now.

arporter:
CI is happy now. Ready for another look from @sergisiso.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -384,9 +385,173 @@ def test_where_array_subsections():
...
+
+
+def test_where_mask_starting_value(fortran_reader, fortran_writer):
+    '''
+    Check handling of a case where the mask array is indexed from values other
+    than unity.
+
+    # TODO #949 - we can't currently take advantage of any knowledge of the
+    # declared lower bounds of arrays because the fparser2 frontend doesn't yet
+    # capture this information (we get an UnknownFortranType).
```
sergisiso, line 402: "UnsupportedFortranType"

sergisiso (changes requested):
@arporter The WHERE implementation looks good now, the integration test pass and also the formatting checks. I made some final minor comments. The only bigger comment is on the array_mixin get_effective_shape, but this is getting a bit out of scope so I would be happy to moving them to another issues if they are not simple to solve.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -546,13 +546,31 @@ def _num_elements(expr):
...
-                                          div, Literal("1", INTEGER_TYPE))
-            return plus
+            if (isinstance(start, IntrinsicCall) and
+                    start.intrinsic == IntrinsicCall.Intrinsic.LBOUND and
+                    isinstance(stop, IntrinsicCall) and
+                    stop.intrinsic == IntrinsicCall.Intrinsic.UBOUND and
+                    isinstance(start.children[0], Reference) and
+                    isinstance(stop.children[0], Reference) and
+                    start.children[0].symbol is stop.children[0].symbol and
+                    start.children[1] == stop.children[1]):
```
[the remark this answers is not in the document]
arporter: Oh yes, apologies - I meant to go back to this code and tidy it up once I'd got everything working. (Hence the lack of comments.) I think I did mean `is_full_range`. I've refactored now and got rid of this duplication and it's a lot nicer (I claim).

src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -47,7 +47,7 @@
 from psyclone.psyir.symbols import (DataSymbol, UnresolvedType,
                                     UnsupportedFortranType, UnsupportedType,
                                     DataTypeSymbol, ScalarType, ArrayType,
-                                    INTEGER_TYPE)
```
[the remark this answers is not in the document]
arporter: Fixed.

src/psyclone/tests/psyir/nodes/array_mixin_test.py
```diff
@@ -491,7 +491,7 @@ def test_aref_get_full_range_unknown_size(extent):
 # _get_effective_shape
 
 
-def test_get_effective_shape(fortran_reader, fortran_writer):
+def test_get_effective_shape(fortran_reader):
```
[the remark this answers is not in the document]
arporter: Good idea (and apologies for missing the previous suggestion). Added and they all work :-) The output could still be improved in a few cases but that's a nicety (and I haven't done it here because it's slightly complicated making sure it works for members of structure accesses too.)

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -530,10 +530,10 @@ def _num_elements(expr):
             is given by (stop - start)/step + 1.
 
             :param expr: the range for which to compute the number of elements.
-            :type expr: :py:class:`psyclone.psyir.nodes.Range` or \
+            :type expr: :py:class:`psyclone.psyir.nodes.Range` or
```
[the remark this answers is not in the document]
arporter: Fixed.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
arporter: Oops. Old habits die hard.

arporter:
SPITZ12 still builds OK, as does ORCA2_ICE_PISCES (in passthrough mode). Integration tests are running. Ready for another look now.

sergisiso (approved):
Thanks @arporter for addressing all comments and refactoring the ArrayMixin utilities. This is a good improvement. The formatting checks pass, the integration test pass with expected results and the remaining where issues are properly marked with TODOs, so this is ready to merge.

PR #2375, opened 2023-10-26
Fix #2374 by using the proper NetCDF variable names for array of fields

### LINKED ISSUE #2374: [PSyData] LFRic driver creation creates incorrect code when using array of fields
After the switch to use pointers for the parameters, the driver now contains:
```
  call extract_psy_data%ReadVariable('chi_1_data', chi_1_data)
  call extract_psy_data%ReadVariable('chi_2_data', chi_2_data)
  call extract_psy_data%ReadVariable('chi_3_data', chi_3_data)
```

But the name used when writing is `chi%1` etc.

hiker:
Fixes #2374.


hiker:
Quick explanation: ATM an array of fields is passed to the extraction library as one parameter, and the extraction library writes them as individual fields by adding a number for each index. But the driver reads in individual fields (since this is what the PSy layer uses). When using pointer variables, the name for array indices which is passed to the extraction library was changed (and all tests where changed, too :)   ). This affects only the NetCDF extraction library (it uses name, the stand-alone library only relies on the right order of the variables, and doesn't actually use the name, so it wasn't detected during CI, which I believe only tests the standalone build).

Not sure if we want to modify the compilation tests to also test NetCDF - in general we are not using the NetCDF version for tests (so it all defaults to use standalone).

hiker:
This is a very small bug fix, unrelated to any other extraction work (but would be good to get this in, since atm the LFRic extraction example does not work when using NetCDF extraction).

arporter (commented):
Apologies for breaking this @hiker. At one stage I had it very broken and I thought I'd fully reverted those changes but it appears I missed this. All looks fine to me.
I've triggered the integration tests and if those are fine then this is good to go.

PR #2051, opened 2023-02-14
Lower LFRic built-ins (towards #1525)

TeranIvy:

PR #2376, opened 2023-10-26
Lower remaining LFRic built-ins (for #1525)

### LINKED ISSUE #1525: Apply lower_to_language in all LFRic built-ins

TeranIvy:
Closes #1525. Supersedes #2051.

Summary of changes:
* Lowered reduction built-ins (inner product and sum). This required changes in psyGen;
* Lowered conversion built-ins using PSyIR representation of `REAL` and `INT` intrinsics. This led to the class `LFRicXKern` and the associated tests being redundant;
* Separated the modified tests into their own modules in `src/psyclone/tests/domain/lfric/builtins`;
* Tidied up the affected tests. The good news is that all built-ins are now supported in PSyKE driver
    creation (`lfric_extract_driver_creator_test.py`);
* Tidied up documentation.


TeranIvy:
CI all fine, passing the PR to @sergisiso for review as agreed.

src/psyclone/psyGen.py
```diff
@@ -1130,13 +1140,16 @@ def reprod_reduction(self):
...
-        name = self.ancestor(InvokeSchedule).symbol_table.\
-            find_or_create_tag(tag, "l_" + tag).name
-        return name
+        reductions with reproducible reductions.
+
+        :returns: local reduction variable name.
+        :rtype: str
+
+        '''
+        return "l_" + self.reduction_arg.name
```
sergisiso, line 1152: Any reason you removed the symbol_table operation? If you know for sure that the symbol already exist you can use
`...symbol_table.lookup_with_tag(tag).name` so if there is a symbol with a clashing name on the original code this would resolve it. If the symbol may not exist we still need the `find_or_create_tag`

src/psyclone/tests/domain/lfric/builtins/X_innerproduct_X_test.py
```diff
@@ -0,0 +1,159 @@
...
+            "      !\n")
+        assert output_dm in code
+        assert "      USE scalar_mod, ONLY: scalar_type" in code
+        assert "      REAL(KIND=r_def), intent(out) :: asum\n" in code
+        assert "      TYPE(scalar_type) global_sum\n" in code
+
+    # Test compilation of generated code
+    assert LFRicBuild(tmpdir).code_compiles(psy)
+
+
```
sergisiso, line 141: Is there any CodeCov missed lines if we remove all this?

This now should already be tested by the code lowering (below) and the backend (in another file), which are better encapsulated for unit testing.

If we mange to delete these it would be great because sometimes we want to change the Fortran formatting in the backend and at the moment we need to update these assert all over the tests, which is quite time consuming.

The same applies to all similar tests under src/psyclone/tests/domain/lfric/builtins/

src/psyclone/tests/domain/lfric/lfric_integer_builtins_test.py
```diff
@@ -44,11 +44,7 @@
 '''
 
 import os
-import pytest
-
```
sergisiso, line 48: This file also has many calls to `code = str(psy.gen)` followed by asserts to the resulting code.

The psy.gen for builtins is now a single line: `parent.add(PSyIRGen(parent, self))` which in turn calls the `lower_to_language_level()`

It would be better to test these just with the "lower_to_language_level()" instead of the psy.gen because here we shouldn't be concerned about anything else that the psy-layer generates, just the built-in resulting code.

Would you mind doing the change? It is mostly deleting unneeded lines.

src/psyclone/tests/domain/lfric/lfric_extract_driver_creator_test.py
```diff
@@ -396,53 +396,6 @@ def test_lfric_driver_operator():
...
-            f"'{name}' is not supported" in out)
-
-    # write_driver prints an error message, and does not return an error
-    # ------------------------------------------------------------------
-    driver_creator.write_driver(invoke.schedule, read_write_info,
-                                "extract", "_post", ("region", "name"))
-    assert (f"Cannot create driver for 'region-name' because:\nLFRic builtin "
-            f"'{name}' is not supported" in out)
-
-
```
sergisiso, line 445: This has left some lines uncovered, see https://app.codecov.io/gh/stfc/PSyclone/pull/2376/indirect-changes

But I think this checks can be removed now since from the lfric_extract_driver_creator.py since there are no more builtins without lowering?

Similarly the LFRicBulitIn.array_ref method is not used anymore, can we remove it?

sergisiso (changes requested):
Well done @TeranIvy, this is a good step to use the new backends on LFRic code and remove old code.

My remaining suggestion is to check if we can remove the remaining psy-layer code assert when testing built-ins, since these are not the responsibly of the built-ins unit tests. Only that the builtin has been lowered to the expected code is necessary there.

There are also some new uncovered lines due to these changes (see inline comments), but I think we can get rid of them.

src/psyclone/psyGen.py
```diff
@@ -1107,19 +1109,27 @@ def reference_accesses(self, var_accesses):
...
+        contains one and `None` otherwise.
+        :rtype: :py:class:`psyclone.psyGen.KernelArgument` or `NoneType`
+
+        '''
         return self._reduction_arg
 
     @property
     def reprod_reduction(self):
-        '''Determine whether this kernel/builtin is enclosed within an OpenMP
+        '''Determine whether this kernel/built-in is enclosed within an OpenMP
```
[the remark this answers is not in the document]
TeranIvy: Done, and added typed option for return type.

src/psyclone/psyGen.py
TeranIvy: I removed the operation of creating a symbol from here because the symbol for local reduction is only required if we are doing reproducible reductions. Otherwise the existing symbol associated with `reproduction_arg` is enough. Besides, the removed code here did not generate the correct data symbol for an array of dimensions `(1, omp_thread_index)`.

I moved the symbol generation to `if self.reprod_reduction` branch of `_reduction_reference` function, as we need to return one Reference from two options: a single-valued argument for non-reproducible reductions and an array argument for reproducible reductions. The array data symbol is now created in `reduction_reference` for reproducible reactions, i.e. `symtab.find_or_create_tag(tag=self.local_reduction_name, ...)`.

I appreciate you would like to check for a possible clash, but then I think creating the symbol here only makes sense if it is done under `if self.reprod_reduction` condition, and it is not quite clear to me what the property would return if that condition is not met. I am refactoring global reduction infrastructure anyway, so we can perhaps decide on this there?

src/psyclone/tests/domain/lfric/lfric_integer_builtins_test.py
TeranIvy: Ok, I significantly pruned down this test. I left the full-blown checks with distributed memory and annexed DoFs in the first built-in, though, as we need to check for the correct looping limits and possible halo exchanges on "write" field. Otherwise I modified all other tests for a single run with distributed memory turned on (default for LFRic API), check for only the built-in code and compilation. Hopefully this is what you had in mind!

src/psyclone/tests/domain/lfric/lfric_extract_driver_creator_test.py
TeranIvy: Oops, forgot these! Removed.

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -2379,31 +2324,25 @@ def metadata(cls):
...
 
-    def gen_code(self, parent):
+    def lower_to_language_level(self):
         '''
-        Generates LFRic API specific PSy code for a call to the
-        X_innerproduct_Y Built-in.
-
-        :param parent: Node in f2pygen tree to which to add call.
-        :type parent: :py:class:`psyclone.f2pygen.BaseGen`
-
```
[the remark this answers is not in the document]
TeranIvy: Oops, apologies for forgetting these! Fixed.

src/psyclone/tests/domain/lfric/builtins/X_innerproduct_X_test.py
TeranIvy: I removed the majority of this and left the key differences between serial code and distributed memory implementation. Similar for other tests mentioned above, with the exception of one test in `reduction_builtins_test.py` , where it seems to me that the intention is to showcase the entire PSy layer.

TeranIvy:
Made changes for code review and refreshed the branch. Ready for another look with @sergisiso.

src/psyclone/psyGen.py
sergisiso: I don't feel too good about how this is implemented. For example I would expect that a method called *_name returns the name of the symbol but it now returns the tag (the symbol name may actually be different after doing the lookup).

And in general I think generation of systems should happen earlier, at psy-layer generation time, or at transformation time if it comes from a modification introduced by applying a transformation. The node methods should only retrieve information already existing on the tree.

However, I do agree that fixing this seem out of scope for this PR. And since the CI run shows we are getting correct results for LFRic, and I am happy to leave this for a future PR/discussion.

src/psyclone/tests/domain/lfric/lfric_extract_driver_creator_test.py
sergisiso: There are still uncovered lines on `fric_extract_driver_creator.py`, but it seems we shouldn't delete any further. Specifically these:
```
 except NotImplementedError as err:
     print(f"Cannot create driver for '{region_name[0]}-"
           f"{region_name[1]}' because:")
     print(str(err))
     return ""
```

Maybe @hiker can advice if there are other ways to trigger this error message now that we don't have any Bultins that don't have a lowering method implemented?

src/psyclone/tests/domain/lfric/builtins/X_innerproduct_X_test.py
```diff
@@ -0,0 +1,122 @@
...
+BASE_PATH = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(
+        os.path.abspath(__file__))))),
+    "test_files", "dynamo0p3")
+
+# The PSyclone API under test
+API = "dynamo0.3"
+
+
+# pylint: disable=invalid-name
```
sergisiso, line 60: I see you had to disable "invalid-name" in all this test files, but this still doesn't remove the pylint issue:
` Module name "X_innerproduct_X_test" doesn't conform to snake_case naming style`
Coming from the filename itself.

I would prefer to conform to python guidelines here and rename the filenames and the testnames in this folder to use lowercase instead of disabling pylint. (To be clear, not the classes or Fortran filenames, this are already properly capitalised)

sergisiso: And not the fault of this PR but pycodestyle finds some issues, would you mind fixing them as they are easy fixes:
```
src/psyclone/tests/psyir/transformations/transformations_test.py:113:11: E275 missing whitespace after keyword
src/psyclone/tests/psyir/transformations/transformations_test.py:116:1: W293 blank line contains whitespace
src/psyclone/psyir/nodes/intrinsic_call.py:969:1: E305 expected 2 blank lines after class or function definition, found 1
```

sergisiso (changes requested):
@TeranIvy Almost there now, I am happy with the implementation and I have now built the documentation, executed tests and CI and run the linters.

See the remaining pylint and pycodestyle issues in the comments, but also the missed CodeCov in a response to a comment from the previous review.

src/psyclone/psyGen.py
TeranIvy: That is a valuable feedback on implementation, thank you! I linked your comments in #2381 and I am happy to add the relevant TODO to this issue to make sure this is dealt with.

TeranIvy: Added TODOs in `local_reduction_name` and `_reduction_reference`.

src/psyclone/tests/domain/lfric/lfric_extract_driver_creator_test.py
hiker: I believe the following code triggers that exception:
```
        for node in nodes:
            for builtin in node.walk(LFRicBuiltIn):
                # If the lower_to_language function is not overwritten from
                # the implementation in Node, the builtin is not yet supported:
                if type(builtin).lower_to_language_level == \
                        Node.lower_to_language_level:
                    raise NotImplementedError(
                        f"LFRic builtin '{builtin.name}' is not supported")
```
So I check if there is an overwritten `lower_to_language_level` function is available, and if not (i.e. only the base class) it raises an exception. With this PR, when all builtins are lowered, this test in the driver (and then the corresponding test in the tests directory) can be removed.

TeranIvy:
Ready for another review from @sergisiso.

src/psyclone/tests/domain/lfric/lfric_extract_driver_creator_test.py
hiker: What happens if this line is triggered, will psyclone completely abort? 

I added a print statement only so I could easily test as much of LFRic as possible, even if (say) a certain intrinsic is not supported. With some files containing 20+ invokes, if one kernel has a problem that the driver extraction cannot handle, we miss out a lot of additional tests.

I have opened #2120 to find a better way (command line option or so, so the user can control if it should abort or just continue with one missing driver - I think I need more command line options in general, so I'll do this as a separate commit), so from that point of view I would prefer the old behaviour (just print the message and return an empty string), and use #2120 to properly control what should happen in case of an error.

sergisiso: I agree with @hiker on not changing the behavior on this PR.

The problem is that the error that was triggering this lines on the test (the builtin does not have a lowering method) does not exist anymore, so we need an alternative NotImplementedError for triggering them.

Alternatively, if it would be easier for hiker to work on a better solution, I would be ok on commenting out the uncovered lines (the print) and add a TODO to #2120

sergisiso:
Back to you @TeranIvy , there is just one conversation left to resolve.

src/psyclone/tests/domain/lfric/lfric_extract_driver_creator_test.py
hiker: Yes, I'm happy with that as well, thanks!

TeranIvy: OK, I restored the `NonImplementedError`, commented out `print` statements and added TODO as requested. Please note that I had to remove `as err` because `pylint` complained about `err` not being used after commenting out `print` statements. However, that does not solve the problem of coverage as the `except NonImplementedError` line is not being tested. I'm happy to restore `print` statements and live with a bit less coverage, TBH.

sergisiso (approved):
@TeranIvy All comments have been addressed, all tests pass, the missing coverage is marked with a todo, there were a few pycodestyle isssues probably introduced by a merge but I took the liberty of fix them. This is ready to merge.

PR #2377, opened 2023-10-27
Add `depth` kwarg to `Node.walk`

### LINKED ISSUE #2368: Add `depth` kwarg to `walk`
It'd be useful to be able to only `walk` over nodes with a specified `depth` level.

jwallwork23:
Closes #2368.

This additional constraint has been useful in my work. I think implementing it in PSyclone is as simple as what I've done here. Note that the depth restriction has to be done once at the end, rather than within the recursion.

This could be straightforwardly extended to accept a tuple of depths, if that would be useful.


src/psyclone/tests/psyGen_test.py
```diff
@@ -2219,6 +2219,14 @@ def test_walk():
     binary_op_list = invoke.schedule.walk(BinaryOperation)
```
arporter, line 2219: It's not your fault but this test is in the wrong place. As we have a rule about not modifying things and moving them at the same time (and to make your life easier), I suggest moving the new bit into a new `test_walk_depth` test in `tests/psyir/nodes/node_test.py`. As Sergi has suggested elsewhere, you can use the `fortran_reader` test fixture to create some PSyIR from some Fortran and then test `walk` on that.

src/psyclone/tests/psyGen_test.py
```diff
@@ -2219,6 +2219,14 @@ def test_walk():
...
     assert len(binary_op_list) == 3
 
+    # Test depth restriction
+    depth_12_list = invoke.schedule.walk(BinaryOperation, depth=12)
+    assert len(depth_12_list) == 1
```
arporter, line 2224: If possible, when you move and re-write this test, could you make the check a bit more careful to ensure that the node that has been returned is definitely the one you expect and that it has the right depth (especially if my suggestion of implementing an efficiency saving comes to fruition).

src/psyclone/psyir/nodes/node.py
```diff
@@ -1092,6 +1095,10 @@ class or a tuple of classes) is found. This can be used to avoid
             return local_list
```
arporter, line 1095: Although this is simple, I find it hard to pass-up the opportunity for efficiency savings. If `depth` is passed into the recursive call to `walk` then it will always be available (if the user has supplied a value) and we can stop walking down a particular branch of the tree as soon as a node's depth exceeds the target depth?

src/psyclone/psyir/nodes/node.py
```diff
@@ -1076,6 +1077,8 @@ class or a tuple of classes) is found. This can be used to avoid
         :type my_type: type | Tuple[type, ...]
         :param stop_type: class(es) at which recursion is halted (optional).
         :type stop_type: Optional[type | Tuple[type, ...]]
+        :param depth: the depth value the instances must have (optional).
+        :type depth: Optional[type | int]
```
arporter, line 1081: Should just be `Optional[int]` I think?

src/psyclone/psyir/nodes/node.py
arporter: If this is a goer then please could you also update the method docstring to mention this functionality.

arporter (changes requested):
Thanks very much for this Joe. It's fine as is but I'm wondering whether we could also implement an efficiency saving to curtail the walk if a depth has been supplied.

src/psyclone/psyir/nodes/node.py
jwallwork23: Thanks, addressed this in 579d09517ba612522dc45e3f0e01c167e160f280.

src/psyclone/tests/psyGen_test.py
jwallwork23: Hopefully the updated version in 53d88af1aecdb71f3734bdda20deba31cedcdea6 is better suited.

src/psyclone/tests/psyGen_test.py
jwallwork23: Addressed in 53d88af1aecdb71f3734bdda20deba31cedcdea6.

src/psyclone/tests/psyGen_test.py
jwallwork23: [Now https://github.com/stfc/PSyclone/pull/2377/commits/c1a4126f33df226a4559f8d762e3f0f814161f67]

src/psyclone/tests/psyGen_test.py
jwallwork23: [Now https://github.com/stfc/PSyclone/pull/2377/commits/c1a4126f33df226a4559f8d762e3f0f814161f67]

src/psyclone/psyir/nodes/node.py
jwallwork23: [Now https://github.com/stfc/PSyclone/pull/2377/commits/8c90d967dfc4d2112b9ac05eac408b549a7b83b3]

jwallwork23:
I don't think the failed CI test is anything to do with the changes in this PR (`Server Error: Internal Server Error for url`).

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -1626,3 +1626,42 @@ def test_siblings(fortran_reader):
...
+    '''Test the depth restriction functionality of Node's walk method.'''
+
+    code = '''subroutine test_depth()
+    integer :: i, j, k
+    integer :: arr(2,2,2)
+
+    do i = 1, 2
+      do j = 1, 2
+        do k = 1, 2
+          arr(i,j,k) = 0
```
arporter, line 1641: For my peace of mind, would you mind increasing the complexity of this a bit so that we exercise going down different branches of the tree. e.g. change the innermost loop body to be:

    if(something)then
      arr(i,j,k) = 0
    else
      arr(i,j,k) = -1
    end if

so that we have two assignments at the same depth but in different branches.

src/psyclone/psyir/nodes/node.py
```diff
@@ -1072,36 +1072,45 @@ def sameParent(self, node_2):
...
+        :type depth: Optional[int]
 
         :returns: list with all nodes that are instances of my_type \
                   starting at and including this node.
         :rtype: List[:py:class:`psyclone.psyir.nodes.Node`]
 
         '''
         local_list = []
-        if isinstance(self, my_type):
+        if isinstance(self, my_type) and (depth is None or self.depth == depth):
```
arporter, line 1099: pylint says this line is too long (and would `depth in [None, self.depth]` work?)

arporter (changes requested):
Thanks for those improvements. I think it's really nice now. I was all set to give it the green light but then I thought that perhaps we should increase the complexity of the test so that the PSyIR tree has multiple branches. Once that's added this is good to go.
Note that I've brought the branch up to master so you'll need to pull it before you make changes.

src/psyclone/tests/psyir/nodes/node_test.py
jwallwork23: Good idea. Addressed in 9b7917c75.

src/psyclone/psyir/nodes/node.py
jwallwork23: Applied this suggestion in 9d96edf37.

arporter (commented):
I should have done this before but I've only now kicked off the integration tests. If they are green then I shall proceed to merge.

arporter (approved):
Integration tests were green.
Will proceed to merge.

PR #2378, opened 2023-10-27
Add `Node.siblings` property

### LINKED ISSUE #2369: Add `get_siblings` function/method
It'd be useful to have a way to get all 'siblings' of a node, i.e., all of its parents' children.

jwallwork23:
Closes #2369.

A simple addition to determine a node's siblings, i.e., parent's children. I implemented it as a property, for consistency with `children` and `parent`. However, it does mean that the siblings include the node itself. Maybe this is fine.


rupertford (approved):
Changes look good. This helps simplify tree traversal. Approving.

PR #2380, opened 2023-10-27
Account for `siblings` special case

### LINKED ISSUE #2369: Add `get_siblings` function/method

jwallwork23:
Closes #2369.

@sergisiso pointed out a special case I missed in #2378 - the case of a root node. Recommended fix applied.

### DIFF
--- a/src/psyclone/tests/psyGen_test.py
+++ b/src/psyclone/tests/psyGen_test.py
@@ -2247,3 +2247,8 @@ def test_siblings():
         siblings = loop.siblings
         assert loop in siblings
         assert len(siblings) == (1 if loop.ancestor(Loop) else 3)
+
+    # Special case of a root node
+    root_siblings = invoke.schedule.siblings
+    assert len(root_siblings) == 1
+    assert root_siblings[0] is invoke.schedule
--- a/src/psyclone/tests/psyGen_test.py
+++ b/src/psyclone/tests/psyGen_test.py
@@ -2226,29 +2226,3 @@ def test_walk():
 
     binary_op_list = invoke.schedule.walk(BinaryOperation, Kern)
     assert not binary_op_list
-
-
-def test_siblings():
-    '''Tests the siblings functionality.'''
-
-    # This function contains an integer assignment followed by two loops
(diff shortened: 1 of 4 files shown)


src/psyclone/tests/psyGen_test.py
```diff
@@ -2247,3 +2247,8 @@ def test_siblings():
         siblings = loop.siblings
         assert loop in siblings
         assert len(siblings) == (1 if loop.ancestor(Loop) else 3)
+
+    # Special case of a root node
+    root_siblings = invoke.schedule.siblings
+    assert len(root_siblings) == 1
+    assert root_siblings[0] is invoke.schedule
```
sergisiso, line 2254: We typically require that the tests must be in the same directory path (but under /tests) than the file they are testing, so they are easier to find.

Tests for `src/psyclone/psyir/nodes/node.py`
must be in: `src/psyclone/tests/psyir/nodes/node_test.py`

Can you move test_sibiling there please?

Also, for testing core psyir functionality we try to avoid using any of the APIs (e.g. nemo) and avoid using external files, so instead of:
`_, invoke = get_invoke("explicit_do_two_loops.f90", "nemo", 0)`

You could use the fortran_reader fixture to create the specific psyir that you need for the test:
```
def test_siblings(fortran_reader):
    code = '''subroutine test()
        ! Code needed for this test
    end subroutine'''
```
You can see an example in the last test on that file (`test_path_from`)

sergisiso (changes requested):
@jwallwork23 Thanks for quickly fixing the bug. I have one more suggestion to make the test follow our current testing practices.

(You will notice that I changed the label to "Reviewed with actions", our review process is that you now make the changes and respond to the conversation (or push back if you disagree) - but without resolving it, and once it is ready for another look change the label again to "Ready for review".)

src/psyclone/tests/psyGen_test.py
jwallwork23: Thanks, have now addressed this in https://github.com/stfc/PSyclone/pull/2380/commits/af68f179ba073e688619a8b9f4c7c5c59159d6c8.

Issue #2256, opened 2023-08-11
Add PSyclone support for Fortran do concurrent


sergisiso:
fparser is adding support for Fortran do concurrent in https://github.com/stfc/fparser/pull/423

PSyclone could simply cannonicalise it to a/multiple loop nodes, we would lose the "iterations are independent" knowledge but we can figure it out again by the dependency analysis.

Or we could store the "iterations are independent" as a Boolean attribute of Loop, which would allow to skip the dependency analysis, but we need a way to invalidate this assumption if there are changes in the loop body. Maybe the tree update signal introduced in https://github.com/stfc/PSyclone/pull/2157

arporter:
There was a fair bit of discussion in the #2383 PR about whether or not we should annotate loops that were originally 'do concurrent'. A couple of the major points were:

https://github.com/stfc/PSyclone/pull/2383#discussion_r1386409729

and

https://github.com/stfc/PSyclone/pull/2383#discussion_r1386784437

arporter:
Perhaps a screenshot is better:
![image](https://github.com/stfc/PSyclone/assets/17409212/b62ec471-66d1-4ae2-990f-ea8b53a9af12)

PR #2383, opened 2023-10-30
(towards #2256) Add support for Fortran do concurrent

sergisiso:


sergisiso:
@arporter @rupertford This is ready for a first review

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2981,60 +3025,51 @@ def _do_construct_handler(self, node, parent):
...
+            loop = self._create_bounded_loop(parent, loop_var, limits_list)
+            loop.ast = node
+            loop_body = loop.loop_body
+            loop_body.ast = node
+        elif ctrl[0].items[3] is not None:
+            triplet = walk(ctrl[0].items[3], Fortran2003.Forall_Triplet_Spec)
+            loop = None
+            for expr in triplet:
+                variable, start, stop, step = expr.items
+                new_loop = self._create_bounded_loop(parent, variable,
```
arporter, line 3055: Could/should we annotate this loop so that we know it was marked as having independent iterations? We might have to remove such an annotation if the loop body is subsequently altered though. Not essential for this PR.

src/psyclone/tests/psyir/frontend/fparser2_do_handler_test.py
```diff
@@ -255,3 +259,71 @@ def test_do_inside_while(fortran_reader, fortran_writer, tmpdir):
...
+    assert loops[0].step_expr == Literal("1", INTEGER_TYPE)
+
+    # Do concurrent with nested loops and step values
+    code = '''
+    subroutine test
+      integer, parameter :: N =10, M=20
+      integer, dimension(N,M) :: array
+      integer i, j
+
+      do concurrent(i=1:N:3, j=1:M:2)
```
arporter, line 295: Can a `do concurrent` itself contain a nested, explicit loop? e.g.

    do concurrent(i=1:N)
      do j=1,10
         a(i,j) = 1.0
      end do
    end do

Please add a test for this if so. (Also, when `do concurrent` is nested inside another loop of a different kind.)

arporter (changes requested):
Good work Sergi. Just needs a few more comments and some additional tests.
This is a towards PR so I'll not check for the issue number.
I'll trigger the integration tests now.

arporter:
Integration tests were green.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2981,60 +3025,51 @@ def _do_construct_handler(self, node, parent):
                     raise NotImplementedError()
 
         ctrl = walk(nonlabel_do, Fortran2003.Loop_Control)
-        # do loops with no condition and do while loops
-        if not ctrl or ctrl[0].items[0]:
+        # In fparser Loop_Control has 4 children, all but one should be None,
+        # the one which is not None defines the loop boundaries style:
+        # LoopCtrl(While_Loop, Counter_Loop, Concurrent_Loop, Concurrent_Loop)
```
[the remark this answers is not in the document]
sergisiso: I am not quite sure what the third one is maybe @rupertford can help?
The fparser rule says the fields are:
```
    F2003: While_Loop_Cntl: scalar-logical-expression, delim
    F2003: Counter_Loop_Cntl: var, lower, upper, [step], delim
    F2008: Concurrent_Loop_Cntl: conc_expr, delim
    F2018: Concurrent_Loop_Cntl: conc_expr, local_x, delim
```
but looking at the code it seems to put the "optional_delimiter" comma from the Fortran2018 R1123 rule in:
 `[ , ] CONCURRENT concurrent-header concurrent-locality`
which seems to be irrelevant for PSyIR.

I am not sure I understand this properly, but for now I updated the comment.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2981,60 +3025,51 @@ def _do_construct_handler(self, node, parent):
...
-        self.process_nodes(parent=loop, nodes=[limits_list[1]])
+        elif ctrl[0].items[1] is not None:
+            # CounterLoops, its children are: Loop variable and Loop Limits
+            loop_var, limits_list = ctrl[0].items[1]
+            loop = self._create_bounded_loop(parent, loop_var, limits_list)
+            loop.ast = node
+            loop_body = loop.loop_body
+            loop_body.ast = node
+        elif ctrl[0].items[3] is not None:
+            triplet = walk(ctrl[0].items[3], Fortran2003.Forall_Triplet_Spec)
```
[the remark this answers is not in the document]
sergisiso: Improved the comment

src/psyclone/psyir/frontend/fparser2.py
sergisiso: This is the reason that this is just a towards PR. An annotation feels wrong to me because it would not be true that we can ignore it (it needs to be invalidated), but maybe a local property. I added a TODO now to explore this.

sergisiso: @arporter I am having second thoughts on this. The way concurrent is defined is that there can still be reductions and dependencies but the compiler is responsible of producing the correct code to respect these. So this annotation/property won't prevent the need to run our independent_iteratons/dependency analysis, because we still need to react to the dependencies or leave it serial. So, I propose not recording it and close the issue with this PR if you agree.

src/psyclone/psyir/frontend/fparser2.py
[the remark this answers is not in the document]
rupertford: If it is a concurrent loop then the arguments are [None, None, optional comment, forall_header()]. Is that what you were asking? It is all a bit convoluted to avoid changing the fparser tree for the 2003 version of loop control.

src/psyclone/psyir/frontend/fparser2.py
arporter: Ah. Yes, in that case it's no different from any other loop as far as the PSyIR is concerned - it becomes a matter for the validation step of any transformation.

arporter: Would it be worth keeping a note though as we might want to target those particular loops with transformations? It should only be used for information though - it wouldn't imply that the dependence analysis wasn't needed for instance.

rupertford: I think we should record that a loop has been declared as independent. It's something that @sergisiso and I disagree on - hopefully you're on my side @arporter :-)

arporter: By the sounds of it, it's no different from ACC KERNELS - it seems to say to the compiler to try to parallelise it but it's up to the compiler to decide? I'll need to read the standard which I will do in a bit...

rupertford: "The DO CONCURRENT construct provides a means for the program to specify that individual loop
iterations have no interdependencies"

arporter: > "The DO CONCURRENT construct provides a means for the program to specify that individual loop
> iterations have no interdependencies"

That doesn't seem to match what Sergi said: "The way concurrent is defined is that there can still be reductions and dependencies but the compiler is responsible of producing the correct code to respect these."?

rupertford: I've looked through the 2008 and 2018 specs and there is nothing about reductions and dependencies. The way the spec reads is that you are simply specifying that all iterations are independent. The example they give is an indirection where the user knows that the lookups are different for different iterations.

The only potentially relevant words I could find are the following but it is unclear to me if they rule out reductions but they do seem to rule out dependencies.

`A variable that is referenced in an iteration shall either be previously defined during that iteration, or shall not be defined or become undefined during any other iteration.`

Where did Sergi find his quote from? Perhaps the compiler vendors have gone further than the spec?

src/psyclone/tests/psyir/frontend/fparser2_do_handler_test.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: For now I reverted it back to adding a TODO and keeping this PR a towards, since I only want to stop the parser breaking preventing any further processing.

We can explore this on the issue, but the only thing that the standard says is "The execution order of the iterations can be left indeterminate (DO CONCURRENT);" but is says nothing regarding if iterations can run in parallel or not. My understanding is that this is for the compiler to figure out. (there are some extra constraints F18 C1136 to C1141 that makes it easier for the compiler).
I believe that:
```
do concurrent (i=1:100)
    a = a  + array(i)
enddo
```
should not be parallelised without dealing with he a race condition. The standard just says that the order doesn't matter. (There is a default(shared) additional keyword that may make the loop independent, see `https://github.com/j3-fortran/fortran_proposals/issues/62` for a discussion on this)

sergisiso:

src/psyclone/psyir/frontend/fparser2.py
rupertford: I would like to (not at all humbly) point out that the quote I gave before and repeat below is from the 2008 spec

> "The DO CONCURRENT construct provides a means for the program to specify that individual loop iterations have no interdependencies"

arporter: I think we can leave this as a TODO for now and argue about it later :-)

arporter (approved):
All requested changes have been made and integration tests are green.
Will proceed to merge.

src/psyclone/tests/psyir/frontend/fparser2_do_handler_test.py
```diff
@@ -255,3 +259,104 @@ def test_do_inside_while(fortran_reader, fortran_writer, tmpdir):
...
+          end do
+        end do
+      end do
+    end subroutine test
+    '''
+    psyir = fortran_reader.psyir_from_source(code)
+    loops = psyir.walk(Loop)
+    assert len(loops) == 4
+    output = fortran_writer(psyir)
+    print (output)
```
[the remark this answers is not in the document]
arporter: I've taken the liberty of fixing this.

PR #2387, opened 2023-11-03
(Towards #1799) Support querying the datatype of a unary/binary operation.

arporter:


arporter:
I need to handle cases where we have a DeferredType, a DeferredType with a partial datatype and an UnknownType.

arporter:
I was thinking of implementing functionality to allow us to create a new UnknownFortranType given an old one and a shape. However, that's quite a lot of work and involves importing a lot of stuff into the datatypes module. Therefore I'm going to leave that for now. As far as operations are concerned, if we have partial datatype information then that is sufficient as I believe that none of the unsupported declaration attributes will affect the type of the result (and, for those that will, we can't have a partial datatype anyway).

arporter:
I've triggered the CI and integration tests. If everything is green then this will be ready for review.

arporter:
Ready for review. This is pretty critical PSyIR stuff so one for @sergisiso or @rupertford I think.

src/psyclone/psyir/nodes/call.py
```diff
@@ -355,6 +355,14 @@ def is_available_on_device(self):
...
         return False
 
+    @property
+    def datatype(self):
+        '''
+        :raises NotImplementedError: TODO #1799
+        '''
+        raise NotImplementedError("datatype property not yet implemented for "
+                                  "Call nodess")
+
```
sergisiso, line 365: I think this should be one level up, in the DataNode class, as all DataNodes will have datatypes.

And again, I am wondering it it will be easier to return DeferredType instead of error because we keep inside the type system semantics, and for example:
`a == call(b) + c` we could end with a bool even if a, b, c are DeferredType/UnknwonType without dealing with any errors.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -42,8 +42,11 @@
 from abc import ABCMeta
 from enum import Enum
 
-from psyclone.errors import GenerationError
+from psyclone.errors import GenerationError, InternalError
 from psyclone.psyir.nodes.datanode import DataNode
+from psyclone.psyir.symbols.datatypes import (
+    ArrayType, BOOLEAN_TYPE, DataTypeSymbol, DeferredType, ScalarType,
+    UnknownFortranType, UnknownType)
```
sergisiso, line 49: pylint says unused InternalError and DatatypeSymbol

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +230,137 @@ def create(operator, lhs, rhs):
...
+                        return UnknownFortranType("")
+
+            argtypes.append(dtype)
+
+        base_type = None
+
+        if self.operator not in self._numeric_ops:
+            # Must be a relational or logical operator. Intrinsic type of
+            # result will be boolean.
+            base_type = BOOLEAN_TYPE
```
sergisiso, line 313: This could be at the top with `return BOOLEAN_TYPE` so we don't need to recurse to the children datatypes nor deal with their complications.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +230,137 @@ def create(operator, lhs, rhs):
...
+                        f"same intrinsic type but different precision "
+                        f"('{argtypes[0].precision}' and "
+                        f"'{argtypes[1].precision}').")
+                # Operands are of the same type so that is the type of the
+                # result.
+                base_type = argtypes[0]
+            elif argtypes[0].intrinsic == ScalarType.Intrinsic.REAL:
+                base_type = argtypes[0]
+            elif argtypes[1].intrinsic == ScalarType.Intrinsic.REAL:
+                base_type = argtypes[1]
```
sergisiso, line 330: Another TODO to 1590 will be useful here, as this will need to be updated when complex are added as they take priority.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +230,137 @@ def create(operator, lhs, rhs):
...
+
+        if all(isinstance(atype, ArrayType) for atype in argtypes):
+            # Both operands are of array type.
+            if len(argtypes[0].shape) != len(argtypes[1].shape):
+                raise TypeError(
+                    f"Binary operation '{self.debug_string()}' has operands "
+                    f"of different shape: '{self.children[0].debug_string()}' "
+                    f"has rank {len(argtypes[0].shape)} and "
+                    f"'{self.children[1].debug_string()}' has rank "
+                    f"{len(argtypes[1].shape)}")
```
sergisiso, line 357: Shouldn't this return the larger shape (with the same assumption that extends match)? e.g:
```
real, dimension(:,:) :: A
real, dimension(:) :: B

A + B
```
result is equal to type of A

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +230,137 @@ def create(operator, lhs, rhs):
...
+            # In general there is no way we can check that the extents of each
+            # dimension match so we have to assume that they do.
+        shape = (argtypes[0].shape if isinstance(argtypes[0], ArrayType) else
+                 argtypes[1].shape)
+        return ArrayType(base_type, shape=shape)
+
 
 # For automatic API documentation generation
 __all__ = ["Operation", "UnaryOperation", "BinaryOperation"]
```
sergisiso, line 366: Unless I missed something, the UnaryOperation is not implemented yet and it seems trivial, its always `self.children[0].datatype` regardless of operator. Maybe you could add it in this PR?

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +230,137 @@ def create(operator, lhs, rhs):
...
+        '''
+        :returns: the datatype of the result of this BinaryOperation.
+        :rtype: :py:class:`psyclone.psyir.symbols.DataType`
+
+        :raises InternalError: if the types of the operands are not permitted
+                               for the given operator.
+        :raises NotImplementedError: if the operands are of the same intrinsic
+                                     type but different precisions.
+        :raises TypeError: if the operands are both arrays but are of different
+                           shapes.
```
sergisiso, line 275: Again, instead of `UnknownFortranType("")` consider `DeferredType()`, I am not so sure about the errors, returning Deferred will make things easier by removing special casing of the caller and better chaining, but we will lose the information about why it happens (we don't have logging yet :/)

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +230,137 @@ def create(operator, lhs, rhs):
...
+        else:
+            # We have a numerical operation.
+            if argtypes[0].intrinsic == argtypes[1].intrinsic:
+                if argtypes[0].precision != argtypes[1].precision:
+                    raise NotImplementedError(
+                        f"Cannot determine the type of expression "
+                        f"'{self.debug_string()}' involving arguments of the "
+                        f"same intrinsic type but different precision "
+                        f"('{argtypes[0].precision}' and "
+                        f"'{argtypes[1].precision}').")
```
sergisiso, line 323: In Fortran'08 page 149 it says the expression type is that of the largest operand precision, is this something we want to check the the precisions are known, or we leave this for another PR?

sergisiso (changes requested):
@arporter Good start to implementing expression datatypes, you will see inline that my main suggestion is asking if it would be easier to return DeferredType instead of dealing with errors and dummy unknowntypes. Other than that it looks good.

doc/developer_guide/psyir.rst
```diff
@@ -521,7 +521,32 @@ sections of the reference guide.
...
         implementations for each possible datatype.
-            
+
+Data Type of an Operation Node
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Table 7.2 of the Fortran2008 standard specifies the rules governing
+the types of operands and their results. The PSyIR follows these rules
+with the exception that there is no support for symbols of complex
+(imaginary) type.  For a numeric intrinsic binary operation, these
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -128,9 +129,37 @@ def datatype(self):
...
         if shape:
-            return ArrayType(self.symbol.datatype, shape)
+            if isinstance(self.symbol.datatype, UnknownType):
+                # Even if an Unknown(Fortran)Type has partial type
+                # information, we can't easily use it here because we'd need
+                # to re-write the original Fortran declaration stored in the
+                # type. We could manipulate the shape in the fparser2 parse
+                # tree if need be but, at this point, we wouldn't know what
+                # the variable name should be (TODO #2137).
+                base_type = UnknownFortranType("type(TODO_2137) :: some_var")
```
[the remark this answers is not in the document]
arporter: I agree that renaming `UnknownType` to `UnsupportedType` would be very helpful. At the moment I thought our convention was that `DeferredType` was something that we *could* know if we traced the import of a Symbol. Possibly that particular case should be `UnresolvedType`?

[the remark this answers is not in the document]
arporter: For the moment I've changed it to return `DeferredType` as that does feel better.

src/psyclone/psyir/nodes/call.py
arporter: Ooh, nice. Done.

src/psyclone/psyir/nodes/operation.py
arporter: Oops.

src/psyclone/psyir/nodes/array_reference.py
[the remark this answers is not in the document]
sergisiso: I created #2419 for this. I thing that whether the Deferred/Unresolved type can be resolved is a property of the symbol (interface), not the type itself. But we can move the discussion to the new issue as this is out of scope for this PR.

src/psyclone/psyir/nodes/operation.py
arporter: Yes, I had it that way originally but then realised that one can have an operation that results in an array of booleans.

src/psyclone/psyir/nodes/operation.py
arporter: Done.

src/psyclone/psyir/nodes/operation.py
arporter: This isn't supported because it is ambiguous. e.g. gfortran says:

        6 |   write(*,*) A + B
          |            1     2
    Error: Inconsistent ranks for operator at (1) and (2)

src/psyclone/psyir/nodes/operation.py
arporter: Oh yes, added.

src/psyclone/psyir/nodes/operation.py
arporter: The UnknownFortranTypes are now all DeferredType. I think we need the `InternalError` and `TypeError` (could also be InternalError?) because, in those situations, we are confident that something has gone wrong. I take your point about the `NotImplementedError` though and will change that.

src/psyclone/psyir/nodes/operation.py
arporter: Ah! I didn't find that so thank you. I've extended the precision-handling support now.

arporter:
Ready for another look now. CodeCov is claiming that one line has become uncovered but I've just checked on master and it isn't covered there either.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -740,8 +740,15 @@ def create(components):
...
-        :type components: list of 4-tuples
+        :param components: the name, type, visibility and initial value (if
+                           any) of each component.
+        :type components: List[tuple[
+            str,
+            :py:class:`psyclone.psyir.symbols.DataType` |
+                 :py:class:`psyclone.psyir.symbols.DataTypeSymbol`,
+            :py:class:`psyclone.psyir.symbols.xxx`,
+            Optional[:py:class:`psyclone.psyir.symbols.Symbol.Visibility`]
+            ]]
```
sergisiso, line 751: This doesn't seem right, Visibility and IntitialValue order is switched and InvitialValue should be a Optional[DataNode], not a symbol. You could take the docstring of the add method in this class, which is really that these components must match.

doc/developer_guide/psyir.rst
```diff
@@ -521,7 +521,42 @@ sections of the reference guide.
         datatype of the operands (e.g. in the select-case canonicalisation).
         A solution to this is to create an abstract interface with appropriate
         implementations for each possible datatype.
-            
+
+Data Type of an Operation Node
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For unary operations, the type of the result is just that of
+the operand.
```
sergisiso, line 529: Given that this are also part of table 7.2, it feels like this sentence should be part the paragraph below, maybe move it to before "For a numeric, binary operation..."

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +242,209 @@ def create(operator, lhs, rhs):
...
         binary_op.addchild(rhs)
         return binary_op
 
+    def _get_result_precision(self, precisions):
+        '''
+        Compares the two precisions to determine the precision of the result
+        of the operation.
+
+        If the two precisions are the same, then that value is returned.
+        Otherwise, Section 7.1.9.3 of the Fortran standard says that in this
```
sergisiso, line 282: The Section numbers change for different versions of the standard, so you need to mention that this is for 2008 standard.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +242,209 @@ def create(operator, lhs, rhs):
...
+                f"'{self.children[1].debug_string()}'")
+
+        # We can't reason about the precision of the result.
+        return ScalarType.Precision.UNDEFINED
+
+    def _get_result_scalar_type(self, argtypes):
+        '''
+        Examines the two operand types to determine the base type of the
+        operation. If the type cannot be determined then an instance of
+        `DeferredType` is returned.
```
sergisiso, line 328: Maybe mention that this follows "Section 7.2 of the Fortran2008" as in the equivalent precision method docstring.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -224,38 +242,209 @@ def create(operator, lhs, rhs):
...
+    def datatype(self):
+        '''
+        Determines the datatype of this operation. If it cannot be determined
+        for any reason then an instance of DeferredType is returned.
+
+        :returns: the datatype of the result of this BinaryOperation.
+        :rtype: :py:class:`psyclone.psyir.symbols.DataType`
+
+        :raises TypeError: if the operands are both arrays but are of different
+                           shapes.
```
sergisiso, line 393: I agree with the previous response that I resolved about this being an error an not a DeferredTpe, but I believe this should be a InternalError.

sergisiso (changes requested):
@arporter Just some minor comments remaining. I agree that the missed Codecov does not seem related to this PR. Formatters and tests all are green.

src/psyclone/psyir/symbols/datatypes.py
arporter: Oh yes. That 'xxx' indicates I hadn't finished writing it! Hopefully fixed now.

doc/developer_guide/psyir.rst
arporter: Oh yes, that's better, thanks.

src/psyclone/psyir/nodes/operation.py
arporter: Oops.

src/psyclone/psyir/nodes/operation.py
arporter: Done.

src/psyclone/psyir/nodes/operation.py
arporter: Yes, done.

arporter:
Hopefully ready for another look once the CI has finished.

sergisiso (approved):
@arporter All the remaining comments have been addressed. And all the test, doc building and formatting still pass. This is ready to merge.

PR #2391, opened 2023-11-08
generator.py exception message bug fix (closes #2390)

### LINKED ISSUE #2390: generator.py message error
`generator.py` raises an exception when trying to construct a message that will be used to raise an exception. The reason for this is that there is an assumption that the result of applying `.scope` to a node will provide a node with a name. However, the node returned could be a `Schedule` node which has no name. The simplest solution is to find the first ancestor that is a Routine node instead as what is wanted is the name of the enclosing subroutine. Also see issue #2343 that is collating issues to do with parsing the LFRic algorithm layer with PSyIR of which this issue is one.

rupertford:
`generator.py` raises an exception when trying to construct a message that will be used to raise an exception. The reason for this is that there is an assumption that the result of applying .scope to a node will provide a node with a name (in particular a `Routine`). However, the node returned could be a `Schedule` node which has no name. This PR fixes this bug by finding the ancestor `Routine` node which is what is wanted (the name of the enclosing subroutine).


rupertford:
Ready for first review from anyone who wants it. It's a very small change. So could be @arporter, @sergisiso @TeranIvy or @hiker

arporter (commented):
Code changes all look fine.
I've fired off the integration tests.

arporter (approved):
Integration tests were green.
Will proceed to merge.

PR #2406, opened 2023-11-20
(closes #2404) Fix error in setop_random

### LINKED ISSUE #2404: bug in PSyAD harness generation
@DrTVockerodtMO has reported that the PSyAD harness code fails on full debug as the bounds used for the kernel `setop_random_kernel_code` are `loop_stop = mesh%get_last_halo_cell(1)` but should go to `op_proxy%ncell_3d`. The former bounds have a lower value than the latter leading to NANs in some of the output.

It is not clear what the problem is because the code giving the problem is generated by PSyclone which is extensively tested and used.

rupertford:
setop_random was only iterating over the horizontal and not over the vertical. This meant that only some of the operator values were being written to. This caused problems when testing adjoints. This PR fixes this problem.


rupertford:
Ready for review. A simple change. One for @arporter, @TeranIvy or @hiker I think

arporter:
Oops. I didn't mean to 'request changes'. I'll sort that out once the integration tests have reported back.

arporter:
Integration tests were all green. Will proceed to merge.

arporter (approved):
Integration tests are green.
All fine.

PR #2412, opened 2023-11-23
(Closes #2411) remove NemoKern

### LINKED ISSUE #2411: Remove NemoKern
A grep shows NemoKern is used in many places, but most of them seem selfreferential - only needed because it exists and it must be inserted/tested/.... I don't think its used in any optimization logic since we use NEMOLoops.loop_type for these.  

This has two goals:
- Continue moving NEMO towards generic code transformation
- Leave room/simplify the refactoring of InlinedKern with the properties needed to fit with the other components of PSyKAl (e.g. inline Kernels without breaking DM)

@arporter @rupertford Is this what we agreed? Who wants to take this?

arporter:


arporter:
Removes some 800 lines :-) Altering the automatic profiling revealed some edge cases that I've also addressed. It also highlights some issues with InlinedKern but that's OK since one of the reasons for doing this was to get ready to make InlinedKern into a more useful concept (i.e. represent a CodedKern that *has* been inlined into the PSy layer).

I claim this is ready for review but will fire off the integration tests now...

One for @sergisiso or @rupertford I think.

sergisiso (approved):
@arporter Sorry that it took so long to review. This is a good step to move towards a generic code transformation mode decoupled from NEMO. All tests, codecov, documentation building and formatting pass. Also integration tests (but I will re-run after before merging to make sure the latest changes on master don't break anything).

One slight regression that we talked about is that we can not profile "kernels" anymore in Nemo, but we typically do "invokes" (now "routines") anyway, however, for applications like the tracer_advections (single routine) that is not very interesting. Maybe in the future we could add more "loop" or "offload/parallel regions" to better profile single routine applications.

Issue #2296, opened 2023-08-29
Add separate routine info object to module manager


hiker:
Part of #1990.

hiker:
Good point - we ended up not doing it this way, but yes, closing it.

PR #2413, opened 2023-11-23
Towards #924 - support module inline in the 'nemo' API.

arporter:


arporter:
Realised that I currently can't handle a routine that is called via a generic interface. We create a RoutineSymbol for that interface but it is of UnknownFortranType. Without inspecting and reasoning about the type information within that type, there's nothing I can do and doing that is effectively implementing something like a `RoutineInterfaceSymbol` which is a different issue altogether. I'll put in a x-failing test for now.

arporter:
I think I have this all working and tested now. I'm going to try extending the integration test to actually make use of it...

arporter:
Of course, NEMO rarely uses `USE xxx, ONLY: yyy` so generally this still isn't going to work. I tried manually adding an `ONLY` in lbclnk.f90 but that then hit the interface problem:

    Module inlining of 'lbc_nfd' failed: Transformation Error: KernelModuleInlineTrans failed to retrieve PSyIR for routine 'lbc_nfd' due to: RoutineSymbol 'lbc_nfd' exists in Container 'lbcnfd' but is of UnknownFortranType:
    interface lbc_nfd
      module procedure lbc_nfd_2d, lbc_nfd_3d, lbc_nfd_4d
      module procedure lbc_nfd_2d_ptr, lbc_nfd_3d_ptr, lbc_nfd_4d_ptr
      module procedure lbc_nfd_2d_ext
    end interface
    Cannot currently module inline such a routine.

I also found a bug that was exercised when attempting to module-inline a routine that was already present in the module. I've added a separate validation check for this.

arporter:
I've created #2422 for the 'interface problem'.

arporter:
Ready for a first look now.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -83,34 +85,75 @@ def validate(self, node, options=None):
...
+            kern_or_call = "Kernel"
+        elif isinstance(node, Call):
+            if isinstance(node, IntrinsicCall):
+                raise TransformationError(
+                    f"Cannot module-inline a call to an intrinsic (got "
+                    f"'{node.debug_string()}')")
+            routine_sym = node.routine
+            kname = routine_sym.name
+            kern_or_call = "routine"
+        else:
```
sergisiso, line 116: What's the plan of this (even if not implemented in this PR)?

If its supported for any generic calls it seems this transformation should be moved to src/psyclone/psyir/transformations but then this mentions CodedKern (which is not generic).  Some options are:
- do instead a specialisation of this transformation for the CodedKern (not ideal because we need to remember to import the specialised one when we are in the appropriate API)
- Make CodedKern inherit from Call and make kernel conform to the Call methods (there is an issue about this that I started a while ago but I never finished)

Also do you plan to drop the "Kernel" from the name of the transformation? And if we go through the pain of renaming, there has been some confusion in the past about the ambiguous meaning with "Inline a kernel module" instead of "Module Inline a Kernel", maybe we can improve the name to avoid this? (Should be Module or Container?, is Inline the right word for it?)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -83,34 +85,75 @@ def validate(self, node, options=None):
...
             raise TransformationError(
                 f"Target of a {self.name} must be a sub-class of "
-                f"psyGen.CodedKern but got '{type(node).__name__}'")
+                f"psyGen.CodedKern or psyir.nodes.Call but got "
+                f"'{type(node).__name__}'")
+
+        parent_container = node.ancestor(Container)
+        if isinstance(parent_container, FileContainer):
```
sergisiso, line 123: Do we need `... or not parent_container)` in case it is a detached subtree?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -250,68 +296,127 @@ def _prepare_code_to_inline(code_to_inline):
...
+        return (caller_name, code_to_inline)
+
+    def apply(self, node, options=()):
         ''' Bring the kernel subroutine into this Container.
 
         :param node: the kernel to module-inline.
         :type node: :py:class:`psyclone.psyGen.CodedKern`
         :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
+        :type options: Optional[Iterable]
```
sergisiso, line 330: Unfortunately I don't think the tuple/Iterable trick works here to remove the:
```
        if not options:
            options = {}
```
This worked when we only iterate the list/dict on the body of the method, in general the options argument is something that we access by name and therefore we use the Mapping interface (https://docs.python.org/3/glossary.html#term-mapping), but I don't think we have an immutable object that implements the Mapping interface that we could provide as default argument (unless we import some non-standard class, but I don't think its worth it)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -250,68 +296,127 @@ def _prepare_code_to_inline(code_to_inline):
...
+                # TODO #924 - we can't currently resolve a subroutine if its
+                # name doesn't match that in the caller (as will be the case
+                # if it's being called via an Interface in Fortran). Once
+                # we can, we will need to use the newly obtained RoutineSymbol
+                # to create a new Call here to replace node.
+                # sym = container.symbol_table.lookup(callee_name)
+                raise NotImplementedError(
+                    f"Cannot module-inline call to '{caller_name}' because its"
+                    f" name does not match that of the callee: "
+                    f"'{callee_name}'. TODO #924.")
```
sergisiso, line 419: I know that we can't convert this to a TransformationError, otherwise we may return normally from an inconsistently applied transformation, but could this condition be found in the validate? If we can then this can be a TransformationError there and prevent uncatched error in user scripts.

sergisiso (changes requested):
@arporter I made an initial review of the PR, I have a few high-level concerns, especially for the get_routine method that I want to discuss before reviewing it in more detail.

src/psyclone/psyir/symbols/routinesymbol.py
```diff
@@ -151,6 +152,70 @@ def is_elemental(self, value):
                             f"'{type(value).__name__}'")
         self._is_elemental = value
 
+    def get_routine(self, container=None):
+        '''
+        Recursively searches for the implementation of this RoutineSymbol by
+        following Container imports.
```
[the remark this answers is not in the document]
arporter: Thanks Sergi. Just a quick note so that I don't forget: once we support routine interfaces, we will need to support the case where we get more than one possible result for the routine being called (assuming we can't disambiguate using the argument lists).

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: My preference would be to make `CodedKern` inherit from `Call` and then we would drop the distinction here.
I hadn't thought about renaming but, yes, I agree it needs to happen. Perhaps it should be "ModuleInlineTrans" or, "ContainerInlineTrans"? I don't know what the "right" word is I'm afraid. We could go for "move" or "insert" perhaps so "InsertRoutineTrans" might be clearer as we're keeping the Routine, just moving it (as opposed to 'inlining' where the Routine disappears).

src/psyclone/psyir/symbols/routinesymbol.py
```diff
@@ -151,6 +152,70 @@ def is_elemental(self, value):
...
+                f"'{container.name}' but is of UnknownFortranType:\n"
+                f"{self.datatype.declaration}\n"
+                f"Cannot currently module inline such a routine.")
+
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.routine import Routine
+        for routine in container.walk(Routine, stop_type=Routine):
+            if routine.name.lower() == self.name.lower():
+                kernel_schedule = routine
+                return kernel_schedule
```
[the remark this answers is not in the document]
arporter: Oh yes, good spot.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: What do @rupertford, @TeranIvy and @hiker think about the name we should use?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes, added.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes. I've just reverted this change.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes - moved it to the new `get_psyir_to_inline()` method. I've kept a safety catch here to make future extension easier.

arporter:
I've been bold and added a new `inline_calls` method to `examples/nemo/scripts/utils.py` that attempts to inline everything. I've added a call to this from the `acc_kernels_trans.py` script so we'll see what the integration tests make of it...

arporter:
As usual, actually trying things for real has revealed a couple of problems - one to do with generic interfaces and another with the `psyclone_internal_cmp` routine that we introduce in some situations. I think I've fixed both of those now and have relaunched the integration tests.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
sergisiso: @arporter Since we are bringing it into the local scope and the implementation is different/copied (changes here will not affect the other callers in other moduels) maybe it can use "localization" or "privatization" e.g "PrivatizeRoutineToLocalContainer"

Which made me think that whatever is called we need to mark the new symbol as "private", which we currently don't do and could potentially cause clashes from other modules importing the one where the caller is!

arporter:
Integration tests failed but I can't reproduce it interactively. Therefore, I've fired them off again but requested that the failing build be done in serial to help me see what's going on...

arporter:
Issue was that a local routine that we were trying to inline was in a CodeBlock and thus not found. I was raising an InternalError in this case (because I thought it should never happen) but now I realise it can legitimately occur I've changed this to a TransformationError.

arporter:
Finally have integration tests working again. Looking at test coverage I now realise that I have duplication between `InlineTrans._find_routine()` and the new (in this PR) `Call.get_callees()`.

arporter:
Most of the duplication is gone now. I've started trying to integrate with the `ModuleManager` but have hit the issue that it's currently LFRic specific. I confess that I reviewed it originally but I now think that it should be general purpose and sub-classed for LFRic. The general solution should (I think) parse every Fortran source file so as to robustly identify the modules that it contains. The LFRic sub-class could optimise this out as the coding standards mean we know that there is only one module per file and that it's name is essentially given by that of the file.

arporter:
Integration tests failed as it turns out that the Fortran frontend wasn't passing the include paths when creating a new parser instance. If I fix that then processing a real NEMO file is incredibly slow, probably because we now attempt to resolve all wildcard imports and that means a lot of parsing!

hiker:
> Most of the duplication is gone now. I've started trying to integrate with the `ModuleManager` but have hit the issue that it's currently LFRic specific. 

Dang. I tried to keep it general, but reading the rest I realise what you mean :(

> I confess that I reviewed it originally but I now think that it should be general purpose and sub-classed for LFRic. The general solution should (I think) parse every Fortran source file so as to robustly identify the modules that it contains. The LFRic sub-class could optimise this out as the coding standards mean we know that there is only one module per file and that its name is essentially given by that of the file.

So it's just to determining which modules are contained in which file name? I have moved this into a separate function, which could be overwritten by domain-specific implementation (or a config setting, or command line option??). That function was designed to return a list of module names, so it's independent of the LFRic coding style (though atm it only returns a one-element list).
This function could in theory parse the files to return the actual full list, though that would be prohibitively expensive :( We could try to do a quick regular expression search in the files, not fully parsing them? Please open a ticket and let me know exactly what the problem is (i.e. do you need a quick parsing of a lot of files, or implement a different rule, and if so, which one).

arporter:
It turns out that the OpenMPI header files include quite a few `EXTERNAL` statements. If the header is INCLUDEd at the top level of a MODULE this results in the whole module being put in a CodeBlock and then we can't find routines to inline. We report that we can't find the routine but that is confusing as it is clearly in the source that we say we've checked - we need a better message.

arporter:
I've fired off the integration tests for this. Note that it currently still includes the regex workaround/extension to the ModuleManager that I think @hiker was going to revisit?

arporter:
Integration tests (processing NEMO) fail because of:

    "psyclone/parse/module_manager.py", line 219, in get_modules_in_file
    source_code = file_in.read()
                  ^^^^^^^^^^^^^^
      File "<frozen codecs>", line 322, in decode
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd0 in position 5446: invalid continuation byte

I need to look at what we do in fparser to workaround this problem.

hiker:
> I've fired off the integration tests for this. Note that it currently still includes the regex workaround/extension to the ModuleManager that I think @hiker was going to revisit?

Yes, I am - but note that I have around 5 PRs that would need to go in first. Looking at them, maybe not all of them. but at least #2296 and #1990 (since they both likely touch the module manager, and I want to avoid getting more conflicts in this sequence of PRs, which has caused me grief in the past)

arporter:
Am getting test compilation errors because (in the kernel-extraction driver tests) I think we're getting the wrong module definition inserted into the generated code - it has an empty `testkern_code` subroutine with no arguments. The problem here is that the Fortran files contained in the `tests/test_files` directories do not define a single application and therefore sometimes define entities with the same name, eg.:

    (psyclone310) mr@laptop tests$ grep -rl testkern_code test_files/dynamo0p3/
    test_files/dynamo0p3/tl_testkern_mod.F90
    test_files/dynamo0p3/testkern_short_name_mod.f90
    test_files/dynamo0p3/testkern_mod.F90
    test_files/dynamo0p3/testkern_no_datatype_mod.f90
    test_files/dynamo0p3/testkern_wrong_mod_name.F90
    test_files/dynamo0p3/testkern_invalid_fortran_mod.f90

Possiby the simplest solution is to chagne these so that they use distinct names?

arporter:
It turned out I only needed to rename one module in one test source file to fix that problem so I've done that. To fix the problem in a real application it may be necessary to collect the source that *is* going to be compiled into some 'build' directory first?

arporter:
If a routine accesses a variable brought into module scope by a wildcard import, then attempting to module inline a second routine that does the same thing (for the same variable) ends up with the first variable being renamed!

arporter:
Can I break-out my changes to `ModuleManager` and `Call` into separate PRs as this is currently rather large?

hiker:
> Can I break-out my changes to `ModuleManager` and `Call` into separate PRs as this is currently rather large?

I'm strongly in favour of this - allows me to easily test with kernel extraction.

arporter:
NEMO integration tests fail with:

```
Inlining of 'ctl_warn' failed:
Transformation Error: Routine 'ctl_warn' contains one or more CodeBlocks and therefore cannot be inlined.
Inlining of 'dia_hsb_rst' failed:
Transformation Error: Routine 'dia_hsb_rst' contains one or more CodeBlocks and therefore cannot be inlined.
Adding profiling to non-OpenACC regions in dia_hsb_init
Generation Error: generator: specified PSyclone transformation module 'acc_kernels_trans'
raised the following exception during execution...
{
Invokes found:
dia_obs_init
dia_obs
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/examples/nemo/scripts/acc_kernels_trans.py", line 431, in trans
    inline_calls(sched)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/examples/nemo/scripts/utils.py", line 166, in inline_calls
    inline_trans.apply(call)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/inline_trans.py", line 136, in apply
    self.validate(node, options)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/inline_trans.py", line 655, in validate
    routine = node.get_callees()[0]
              ^^^^^^^^^^^^^^^^^^
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/nodes/call.py", line 504, in get_callees
    for name in container.resolve_routine(rsym.name):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/gh_Inlining of 'ctl_warn' failed:
Transformation Error: Routine 'ctl_warn' contains one or more CodeBlocks and therefore cannot be inlined.
Inlining of 'dia_hsb_rst' failed:
Transformation Error: Routine 'dia_hsb_rst' contains one or more CodeBlocks and therefore cannot be inlined.
Adding profiling to non-OpenACC regions in dia_hsb_init
Generation Error: generator: specified PSyclone transformation module 'acc_kernels_trans'
raised the following exception during execution...
{
Invokes found:
dia_obs_init
dia_obs
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/examples/nemo/scripts/acc_kernels_trans.py", line 431, in trans
    inline_calls(sched)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/examples/nemo/scripts/utils.py", line 166, in inline_calls
    inline_trans.apply(call)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/inline_trans.py", line 136, in apply
    self.validate(node, options)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/inline_trans.py", line 655, in validate
    routine = node.get_callees()[0]
              ^^^^^^^^^^^^^^^^^^
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/nodes/call.py", line 504, in get_callees
    for name in container.resolve_routine(rsym.name):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/nodes/container.py", line 279, in resolve_routine
    raise TypeError(
    TypeError: Expected 'ctl_warn' to correspond to either a RoutineSymbol or a GenericInterfaceSymbol but found 'Symbol'
}
please check your scriptrunner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/nodes/container.py", line 279, in resolve_routine
    raise TypeError(
    TypeError: Expected 'ctl_warn' to correspond to either a RoutineSymbol or a GenericInterfaceSymbol but found 'Symbol'
}
please check your script
```

arporter:
Performance is now back to the level it is on master (no faster unfortunately). However, this has required that I turn off the automatic addition of profiling calipers to non-OpenACC regions in the `acc_kernels_trans.py` script. I'm not sure what the solution to this is. If at any point we finish the work to extend PSyclone to retain any directives in existing code then we could hit this problem again as there may be directives inside a routine that we wish to inline. The logical thing to do at that point would be to strip-out the directives from the inlined version on the basis that they only applied when it was a standalone routine. We could do the same thing with profiling calls.

hiker:
> Performance is now back to the level it is on master (no faster unfortunately). However, this has required that I turn off the automatic addition of profiling calipers to non-OpenACC regions in the `acc_kernels_trans.py` script. 

With that you meant the automatically applied psydata-based profiling regions (using `--profile kernels/invokes`)? It is strange that this would cause a huge overhead (maybe verify that they are indeed correctly applied outside of the loops? Maybe something has changed?). I am aware that this is NEMO, but still if profiling has a huge overhead, it's worth checking that the transformations did not somehow moved profiling into a loop.

arporter:
image

arporter:
Currently, NEMO performance is unaffected because we don't manage to inline anything :-( There are a variety of reasons for this:
 * Not finding the source of the routine;
 * The routine contains a CodeBlock;
 * The routine has an unsupported type of argument (e.g. `character(len=*)`  or `optional`).

arporter:
This PR is big enough and is only a 'towards' so I'll leave the issue of inlining routines called via an interface for now. I've updated the utils::inline_calls routine so that it 'forces' the inlining of a few routines that are only prevented by a CodeBlock containing a CLOSE statement. The integration tests for this are re-running now.
For LFRic, the issue of inlining will come to the fore once we have integration tests for GPU offload.

arporter:
Ready for a look now from @sergisiso.

src/psyclone/psyir/nodes/container.py
```diff
@@ -171,21 +171,85 @@ def node_str(self, colour=True):
...
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.routine import Routine
+        from psyclone.psyir.symbols.symbol import Symbol, SymbolError
+        # pylint: enable=import-outside-toplevel
+        for node in self.children:
+            if isinstance(node, Routine) and node.name.lower() == rname:
+                # Check this routine is public
+                routine_sym = self.symbol_table.lookup(node.name)
+                if (allow_private or
+                        routine_sym.visibility == Symbol.Visibility.PUBLIC):
```
sergisiso, line 205:
```
else:
    return None
```
As presumably Fortran does not let private subroutine names clash with possible wildcard imported ones, so we can stop without searching further? (it may be good to verify this)

src/psyclone/psyir/nodes/container.py
```diff
@@ -171,21 +171,85 @@ def node_str(self, colour=True):
...
+            container = None
+            try:
+                container = child_cntr_sym.container(local_node=self)
+            except (SymbolError, FileNotFoundError):
+                pass
+            if not container:
+                # TODO #11 log that we didn't find the Container from which
+                # the routine is imported.
+                return None
+            return container.get_routine_psyir(rname)
```
sergisiso, line 236: Should we pass the same "check_wildcard_imports" value down?

src/psyclone/psyir/nodes/container.py
```diff
@@ -171,21 +171,85 @@ def node_str(self, colour=True):
...
+            # Routine does not exist in the SymbolTable.
+            routine_sym = None
+
+        if routine_sym and isinstance(routine_sym, GenericInterfaceSymbol):
+            # The routine is actually an interface to one or more routines.
+            # This should be handled outside this routine (e.g. by using
+            # `resolve_routine` first).
+            return None
+
+        if routine_sym and routine_sym.is_import:
```
sergisiso, line 224: The `if (allow_private or routine_sym.visibility == Symbol.Visibility.PUBLIC)` also needs to be done here and with the wilcard imports below because we can have:
```
module my_mod
   use modA, only: sub1
   use modB  ! contains sub2
   private :: sub1, sub2
end module
```
We want get_routine_psyir() by default to not find sub1 and sub2

src/psyclone/psyir/symbols/containersymbol.py
```diff
@@ -125,15 +125,28 @@ def copy(self):
...
         return new_symbol
 
-    @property
-    def container(self):
+    def container(self, local_node=None):
         ''' Returns the referenced container. If it is not available, use
         the interface to import the container
 
+        :param local_node:
+        :type local_node:
```
sergisiso, line 133: missing docstring

src/psyclone/psyir/symbols/containersymbol.py
```diff
@@ -125,15 +125,28 @@ def copy(self):
         new_symbol.is_intrinsic = self.is_intrinsic
         return new_symbol
 
-    @property
-    def container(self):
+    def container(self, local_node=None):
```
sergisiso, line 128: Since it's not a property anymore I would name it "get_container_psyir" to be equivalent to the introduced "get_routine_psyir".

And in my opinion, in both cases, "search" (instead of "get") would imply more clearly that they are potentially expensive operation that can return "None". But I will accept both.

sergisiso (changes requested):
That's a huge improvement from last time @arporter, thanks for doing the API refactoring. I found that this time the implementation was easy to follow (which is great considering the complexity of the problem) and my suggestions/doubts are more localised.

examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -85,7 +86,8 @@
 PROFILE_TRANS = ProfileTrans()
 
 # Whether or not to add profiling calls around unaccelerated regions
-PROFILE_NONACC = True
+# N.B. this can inhibit PSyclone's ability to inline!
+PROFILE_NONACC = False
```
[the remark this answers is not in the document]
arporter: It's possible. I'll create another issue - #2682

examples/nemo/scripts/utils.py
```diff
@@ -150,6 +151,57 @@ def enhance_tree_information(schedule):
...
+
+    where each step is dependent upon the success of the previous one.
+
+    TODO #924 - this could be InlineAllCallsTrans.apply(schedule,
+                                                        excluding={})
+
+    :param schedule: the schedule in which to search for Calls.
+    :type schedule: :py:class:`psyclone.psyir.nodes.Schedule`
+
+    '''
```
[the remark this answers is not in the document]
arporter: Yes, you're right. This is going to be a problem. I've added it as a task item in #924 and linked to this comment.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -36,21 +36,25 @@
 #         J. Henrichs, Bureau of Meteorology
 # Modified I. Kavcic, Met Office
 
-''' This module provides the KernelModuleInlineTrans transformation. '''
+''' This module provides the KernelModuleInlineTrans transformation.
+TODO #924 - rename this to PrivatiseRoutineToLocalContainerTrans and move it
+to psyir/transformations/. '''
```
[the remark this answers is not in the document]
arporter: Done (#2683)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -64,57 +68,92 @@ class KernelModuleInlineTrans(Transformation):
 
 
     .. warning ::
-        Not all kernel subroutines can be module-inlined. This transformation
-        will reject attempts to in-line kernels that access global data in the
-        original module.
+        Not all Routines can be privatised. This transformation will reject
+        attempts to privatise routines that access global data in the
+        original Container.
```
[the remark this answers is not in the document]
arporter: Yes :-)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -64,57 +68,92 @@ class KernelModuleInlineTrans(Transformation):
...
+                    raise TransformationError(
+                        f"{kern_or_call} '{kname}' cannot be module inlined "
+                        f"into Container '{parent_container.name}' because "
+                        f"there is no explicit import of it ('USE ..., ONLY: "
+                        f"{kname}' in Fortran) and a Routine with that name "
+                        f"is already present in the Container.")
 
         # Check that the PSyIR and associated Symbol table of the Kernel is OK.
         # If this kernel contains symbols that are not captured in the PSyIR
         # SymbolTable then this raises an exception.
```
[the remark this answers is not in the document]
arporter: Oh yes, that was old.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -123,49 +162,57 @@ def validate(self, node, options=None):
             if isinstance(symbol, IntrinsicSymbol):
                 continue
             if not symbol.is_import:
-                try:
-                    var.scope.symbol_table.lookup(
-                        symbol.name, scope_limit=kernel_schedule)
-                except KeyError as err:
+                _tmp = var.scope.symbol_table.lookup(
+                    symbol.name, scope_limit=kernel_schedule, otherwise=None)
+                if not _tmp:
```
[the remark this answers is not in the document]
arporter: Oh yes, not sure what I was thinking there!

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -64,57 +68,92 @@ class KernelModuleInlineTrans(Transformation):
...
+                f"{self.name} failed to retrieve PSyIR for {kern_or_call} "
+                f"'{kname}' due to: {error}"
+            ) from error
 
         # We do not support kernels that use symbols representing global
-        # variables declared in its own parent module (we would need to
+        # variables declared in their own parent module (we would need to
         # create new imports to this module for those, and we don't do
         # this yet).
         # These can only be found in References, Calls and CodeBlocks
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -255,59 +344,105 @@ def apply(self, node, options=None):
...
+                # The routine symbol already exists, and we know from the
+                # validation that it's a Routine. Now check if they are
+                # exactly the same.
+                for routine in container.walk(Routine, stop_type=Routine):
+                    if routine.name == caller_name:
+                        # This TransformationError happens here and not in the
+                        # validation because it needs the symbols_to_bring_in
+                        # applied to effectively compare both versions.
+                        # This will be fixed when module-inlining versioning is
+                        # implemented.
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -73,12 +75,29 @@ def test_validate_inline_error_if_not_kernel():
...
+    with pytest.raises(TransformationError) as err:
+        inline_trans.apply(call)
+    assert ("Cannot module-inline a call to an intrinsic (got 'SIN(b)')"
+            in str(err.value))
 
 
 def test_validate_with_imported_subroutine_call():
     ''' Test that the module inline transformation supports kernels with
-    call nodes that reference and imported symbol. '''
+
```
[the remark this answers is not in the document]
arporter: Oops.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -650,3 +720,188 @@ def test_module_inline_with_interfaces(tmpdir):
...
+        intrans.validate(call)
+    assert ("failed to retrieve PSyIR for routine 'my_sub' due to: Failed to "
+            "find the source code of the unresolved routine 'my_sub' - looked "
+            "at any routines in the same source file and attempted to resolve "
+            "the wildcard imports from ['my_mod', 'other_mod']. However, "
+            "failed to find the source for ['my_mod', 'other_mod']. The module"
+            " search path is set to []. Searching for external routines that "
+            "are only resolved at link time is not supported."
+            in str(err.value))
+
```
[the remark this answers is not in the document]
arporter: Good idea, done.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -650,3 +720,188 @@ def test_module_inline_with_interfaces(tmpdir):
...
+    '''
+    sym = RoutineSymbol("my_sym")
+    rout = Routine.create("my_sym", SymbolTable(), [])
+    node = Call.create(sym)
+    # For simplicity we just monkeypatch Call.get_callees() so that it appears
+    # to return more than one Routine.
+    monkeypatch.setattr(node, "get_callees", lambda: [rout, rout])
+    with pytest.raises(TransformationError) as err:
+        KernelModuleInlineTrans._get_psyir_to_inline(node)
+    assert ("The target of the call to 'my_sym' cannot be inserted because "
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -131,10 +131,9 @@ def structure(self):
     @property
     def get_ast_nodes(self):
         '''
-        :returns: the list of nodes associated with this code block in \
-        the original AST.
-        :rtype: list of subclass of \
-        `:py:classfparser.two.Fortran2003.Base`
+        :returns: the nodes associated with this code block in
+                  the original AST.
```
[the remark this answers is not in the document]
arporter: Done (plus some others).

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -333,21 +333,36 @@ def generate_symbols(self, symbol_table):
...
             # this Container as part of a PSyData transformation.
             csym = ContainerSymbol(self.fortran_module)
             symbol_table.add(csym, tag=self.fortran_module)
 
+        if not isinstance(csym, ContainerSymbol):
+            raise InternalError(
+                f"Cannot add PSyData module '{self.fortran_module}' because "
+                f"another Symbol already exists with that name and is a "
+                f"{type(csym).__name__} rather than a ContainerSymbol.")
+
```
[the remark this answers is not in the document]
arporter: Hmm. This was part of https://github.com/stfc/PSyclone/commit/34521a68a2. It looks like I was hitting problems with the code attempting to add the same ContainerSymbol multiple times while lowering code for PSyDataNodes. Now we add it just once.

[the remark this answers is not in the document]
arporter: I think this is because each PSyDataNode sets-up the necessary Symbols required so this step can be repeated.

src/psyclone/psyir/nodes/container.py
arporter: You're quite right - it seems that Fortran doesn't permit this (I expected that it would). Updated code and comments.

src/psyclone/psyir/nodes/container.py
arporter: We could, but that then means we always fully recurse and that was prohibitive for NEMO where everything is a wildcard import. This way, we only recurse one level and then stop. I've made this more explicit by setting `check_wildcard_imports=False` in this call. To do better we'd need to keep track of the depth.

src/psyclone/psyir/nodes/container.py
arporter: Good point. I've added the check after we lookup the RoutineSymbol. Further down, before we look at wildcard imports, I've added a check on the default visibility of the table as, without an explicit Symbol entry, that will determine whether an imported symbol is visible. I'll need to add tests for this too.

src/psyclone/psyir/symbols/containersymbol.py
arporter: Oops.

src/psyclone/psyir/symbols/containersymbol.py
arporter: That's a good idea. I've gone with `find_{container,routine}_psyir` and I think it makes things a lot more readable.

arporter:
Integration tests were green. Ready for another look now.

sergisiso (approved):
@arporter All the comments have been addressed. Test, codecoverage and integration tests all pass. User/developer/reference al build. This is ready to merge.

Issue #2427, opened 2023-12-04
Fix bug in maxval2loop transformation


rupertford:
At the moment the output of the `maxval2loop` transformation uses the `TINY()` intrinsic as an initial value. However, this is the smallest positive number possible so if there were only negative numbers the result would be the value of `TINY()` which is incorrect. The initial value should instead be `HUGE() * -1.0`

rupertford:
Creating branch [2427_maxval2loop_bug](https://github.com/stfc/PSyclone/tree/2427_maxval2loop_bug)

PR #2428, opened 2023-12-04
fix maxval2loop bug (issue #2427)

rupertford:
`maxval2loop` uses `TINY()` as its initial value which does not take negative numbers into account. This PR replaces `TINY()` with `-HUGE()`.


rupertford:
Ready for first review from @sergisiso

arporter:
Integration tests were green.

PR #2439, opened 2023-12-14
(closes #2105) Use maxval2loop and atomics to parallelise NEMO stpctl.f90

### LINKED ISSUE #2105: Execute maxval intrinsic in NEMO stpctl.f90 on the GPU
#2085 added support for minval, maxval and sum IntrinsicCalls but this does not yet execute the new intrinsics on the GPU and is still one of the manual tweak necessary to improve the performance.

To run them on the GPU for the OpenACC kernels we need to:
- Introduce a minval2code, maxval2code and sum2code transformations
or
- Refactor the `if not have_loop:` in the try_kernel_trans method of examples/nemo/scripts/kernels_trans.py 

To run them on the GPU for explicit OpenACC or OpenMP we need to:
-  Add OpenMP reduction clause in addition to the maxval2code transformation

sergisiso:
Updates the nemo scripts with the new maxval2loop and atomic psyclone functionality, this makes it ~10% faster (for my 10its elapsed time test) - although there was a lot of variability today in Glados.

If I look with the profiler:
Before:
image
After:
image
The benefit comes from keeping the arrays in the GPU, not so much from stpctl itself as it is dominated by a syscall (IO) and sending this new data to the GPU.


sergisiso:
@arporter @rupertford This is ready to review. (if the reviewer doesn't do it before I will launch the integration tests at night to try to get GLaDos emptier as this PR should give a new baseline for perf comparisons)

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -112,6 +114,25 @@ def trans(psy):
...
+            for loop in invoke.schedule.walk(Loop):
+                # Skip if an outer loop is already parallelised
+                if loop.ancestor(Directive):
+                    continue
+                omp_loop_trans.apply(loop, options={"force": True})
+                omp_target_trans.apply(loop.parent.parent)
+                assigns = loop.walk(Assignment)
+                if len(assigns) == 1 and assigns[0].lhs.symbol.name == "zmax":
+                    stmt = assigns[0]
+                    if OMPAtomicDirective.is_valid_atomic_statement(stmt):
```
arporter, line 129: This should really be done with a Transformation. For now I'll accept a TODO though (unless there's some issue with having a transformation - I very vaguely remember that there might be?).

examples/nemo/scripts/utils.py
```diff
@@ -35,13 +35,15 @@
 
 ''' Utilities file to parallelise Nemo code. '''
```
arporter, line 36: Not for this PR but I'm thinking it would be good to move as much of this as we can into `domain/nemo/` and then we can get it unit tested.

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -41,7 +41,8 @@
     enhance_tree_information, add_profiling
 
 from psyclone.psyGen import TransInfo
```
arporter, line 43: Please could you clean up for pylint in this file - it doesn't like the ordering of imports and says `Schedule` is unused.

arporter (changes requested):
Change looks OK, always good to go faster! I'd prefer a Transformation but that can be a separate Issue :-)
We also need to see about moving code out of `examples` and into e.g. `domain/nemo` so that we can add tests. Again, a separate Issue.
I note that the integration tests failed last time but I think that's because of #2440 so will have another go once that's on.

examples/nemo/scripts/utils.py
sergisiso: I agree this is getting bigger than it should, in addition to `domain/nemo` I think the openmp collapsing can be an option to the generic omp_loop_trans/acc_loop_trans. The normalisation utility could also be a generic metatransformations once array2loop have been refactored, because it isn't NEMO specific anymore.

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: I created #2446 and added the TODO. The problem is that the loop_trans needs to be aware of the atomic (same with reductions) otherwise it can not guarantee that it is valid (this is why for this file I pass the 'force' option). I think the best solution is to do both from inside the loop_transformation but I prefer to discuss this in the new issue.

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: Done

sergisiso:
> I note that the integration tests failed last time but I think that's because of https://github.com/stfc/PSyclone/pull/2440 so will have another go once that's on.

Unfortunately it was not/only this, it is because a maxval2loop in the normalisation step fails for the esmwf version of nemo.
```
Invokes found in psy_bdyini_psy:
bdy_def
raised the following exception during execution...
{
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/examples/nemo/scripts/omp_cpu_trans.py", line 92, in trans
    normalise_loops(
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/examples/nemo/scripts/utils.py", line 179, in normalise_loops
    Maxval2LoopTrans().apply(intr)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/intrinsics/array_reduction_base_trans.py", line 224, in apply
    reference2arrayrange.apply(reference)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/reference2arrayrange_trans.py", line 177, in apply
    self.validate(node, options=None)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/reference2arrayrange_trans.py", line 149, in validate
    if not node.symbol.is_array:
           ^^^^^^^^^^^^^^^^^^^^
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/symbols/symbol.py", line 440, in is_array
    raise ValueError(f"No array information is available for the "
    ValueError: No array information is available for the symbol 'nb_bdy'.
```
To fix it I modified the generic implementation to return False, with matches the specialised implementation in that if doesn't have enough information (e.g. DeferedType, no partial datatype) it also returns false.

sergisiso:
There is also this issue:
```
    Maxval2LoopTrans().apply(intr)
      File "/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.12/site-packages/psyclone/psyir/transformations/intrinsics/array_reduction_base_trans.py", line 349, in apply
    outer_loop.parent.children.insert(outer_loop.position, assignment)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'NoneType' object has no attribute 'children'
```

sergisiso:
Also fails in ecmwf's `zcflnow(1) = MAX( zcflnow(1), MAXVAL( ABS( pv_ice(:,:) ) * rdt_ice * r1_e2v(:,:) ) )` with the compile time error `The shapes of the array expressions do not conform. [ZCFLNOW] zcflnow(1) = MAX(zcflnow(1), tmp_var)`

where zcflnow is:
`REAL(wp), DIMENSION(1)  ::   zcflprv, zcflnow`

sergisiso:
@arporter This is ready for another review. It was more complex to fix than I thought. I could have disabled the maxval2loop on the ecmwf version since the CPU will not parallelise them, but I think its good to have them on to expose the potential issues.

rupertford:
> The one that fails is `jpbdta = MAXVAL(nblendta(:,:nb_bdy))` Also when doing: `x(3) = maxval(a)` it initialises with `x(3) = -HUGE(x)` but should be `-HUGE(x(3))`?

This shouldn't matter. The use of a variable or array should result in the same value of HUGE as they have the same datatype and it is the precision of the datatype that determines the value of HUGE.

sergisiso:
> > The one that fails is jpbdta = MAXVAL(nblendta(:,:nb_bdy)) Also when doing: x(3) = maxval(a) it initialises with x(3) = -HUGE(x) but should be -HUGE(x(3))?

> This shouldn't matter. The use of a variable or array should result in the same value of HUGE as they have the same datatype and it is the precision of the datatype that determines the value of HUGE.

These were 2 different problems:
- The first is that nblendta(:,:nb_bdy) was silently not converted to a loop by the arrayrange2loop transformation and the apply continued with the wrong nodes. This is now catch and an error is raised.
- The second is it true that for an array HUGE does the right thing, although it is not documented in the standard, but I still think it had to be fixed because the lhs can be a complex expression e.g. "mystryct(3)%myvar" so the huge needs to be initialised by type of the resulting access. And the change was quite simple.

arporter:
> * nblendta(:,:nb_bdy) was silently not converted to a loop by the arrayrange2loop transformation

Hi @sergisiso, can you remember why this was? It seems like something we should fix (in a separate PR)?

arporter (commented):
All looks good now, thanks.
I'll re-run the integration tests, just to make sure.

arporter:
Integration tests were fine.

arporter:
I noticed that CodeCov was reporting newly-missed lines. This was because Symbol.is_array no longer raises an exception so I've removed the try...except around the call to it. I'll re-run the integration tests to be on the safe side.

arporter:
Happily, integration tests were all fine.

PR #2454, opened 2024-01-04
(#Closes 2418) Improvements to fix 2418

LonelyCat124:
Small PR to fix the error caused by intrinsiccall nodes now they inherit from `Call`.


LonelyCat124:
Ready for review now :)

PR #2456, opened 2024-01-05
(Closes #1744 ) - Otter via PSyData implementation PR

### LINKED ISSUE #1744: Otter transformations

LonelyCat124:
Currently only contains initial modifications for the Task Directive nodes to support PSyData additions from transformations.


LonelyCat124:
Closing this PR as this is not the approach I expect us to use.

Issue #2004, opened 2023-01-10
Rationalise ArrayRange2Loop transformations.


sergisiso:
Currently there is a generic:
- ArrayRange2Loop transformation

and a more advanced:
- NemoArrayRange2Loop

The Nemo version still adds NemoLoops and NemoKern, but everything else which is generic should be moved to the generic transformation.

Also Nemo has:
- NemoAllArrayRange2LoopTrans
- NemoOuterArrayRange2LoopTrans

Similar generic transformations should exist.

PR #2465, opened 2024-01-11
(towards 2004) Move same_range functionality to ArrayMixin and other improvements

sergisiso:
This is some work in preparation of merging the two implementations we have of ArrayRange2Loop in #2004, but it is big enough to be a standalone PR:
- It moves the same_range implementation outside the arrayrange transformation, because it seems this functionality is more generic.
- It allows psyir_from_expression/statement methods to work without a symbol_table argument, because it is convenient for tests where we do not care about the symbol declarations.
- It modernises method signatures using type hints
- Adds some xfail test to demonstrate cases when this functionality is still lacking


sergisiso:
This is now ready for review, one for @arporter or @hiker

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -598,6 +598,121 @@ def get_outer_range_index(self):
...
+        is the same but it can not be proved, so we return False.
+
+        :param index: the index indicating the location of a range node in
+            this array.
+        :param array2: the array accessor that we want to compare it to.
+        :param index2: the index indicating the location of a range node in
+            array2.
+
+        :returns: True if the ranges are the same and False if they are not
+        the same, or if it is not possible to determine.
```
hiker, line 622: The continuation line needs to be indented, otherwise the ref-guide complains.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -598,6 +598,121 @@ def get_outer_range_index(self):
                 return child.position
         raise IndexError
 
+    def same_range(self, index: int, array2, index2: int) -> bool:
+        ''' This method compares the range length of this array node
+        at a given index with the range length of a second array at
+        a second index. This is useful to verify if array operations
+        are valid, e.g.: A(3,:,5) + B(:,2,2).
+
+        Note that this check supports symbolic comparisons, e.g.:
```
hiker, line 607: Without looking at the actual transformation, but shouldn't this compare `upper-lower` of the two ranges and the `step` (as a simplified approach ... or potentially `(upper-lower)/step`, though that likely needs a round to integer somewhere :) )?

That should allow us to compare the ranges in `a(i:i+2)` with `b(1:3)`. I am ok to leave this if this is a restriction of the transform that we don't want to change in a refactoring PR, but maybe add this as an example to the ticket #2485 (or document this, looking at the docs of the arrayrange2loop this seems not to be mentioned.

src/psyclone/psyir/transformations/arrayrange2loop_trans.py
```diff
@@ -78,127 +78,6 @@ class ArrayRange2LoopTrans(Transformation):
 
     '''
```
hiker, line 79: In line 44 the import of symbolic maths can be removed.

src/psyclone/tests/psyir/frontend/fortran_test.py
```diff
@@ -149,29 +149,24 @@ def test_fortran_psyir_from_expression_invalid(fortran_reader):
...
     assert ("Must be supplied with a valid SymbolTable but got 'str'" in
             str(err.value))
-    with pytest.raises(TypeError) as err:
-        fortran_reader.psyir_from_expression("3.0 + sin(a)", None)
-    assert ("Must be supplied with a valid SymbolTable but got 'NoneType'" in
-            str(err.value))
-    table = SymbolTable()
     # OK
-    fortran_reader.psyir_from_expression("return", table)
+    fortran_reader.psyir_from_expression("return")
```
hiker, line 160: Nice, it is really convenient for testing indeed.

src/psyclone/tests/psyir/transformations/arrayrange2loop_trans_test.py
```diff
@@ -267,105 +267,6 @@ def test_transform():
     assert ArrayRange2LoopTrans()
```
hiker, line 267: Not your problem, but I get:
```
./psyir/transformations/arrayrange2loop_trans_test.py:273:1: E302 expected 2 blank lines, found 1
```

src/psyclone/tests/psyir/nodes/array_mixin_test.py
```diff
@@ -564,3 +564,127 @@ def test_get_outer_range_index_error():
     array = ArrayReference.create(symbol, [Literal("2", INTEGER_TYPE)])
     with pytest.raises(IndexError):
         _ = array.get_outer_range_index()
+
```
hiker, line 567: I get:
```
psyir/nodes/array_mixin_test.py:569:0: R0915: Too many statements (59/50) (too-many-statements)
```
Maybe split this test into two, one for cases raising an exception, one for the rest?

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -598,6 +598,121 @@ def get_outer_range_index(self):
                 return child.position
         raise IndexError
 
+    def same_range(self, index: int, array2, index2: int) -> bool:
```
hiker, line 601: I get:
```
psyir/nodes/array_mixin.py:601:4: R0912: Too many branches (14/12) (too-many-branches)
```
I would be happy to disable this warning with a pylint directive.

If you are touching this file anyway again, could you also add the missing closing ` in lines 136 and 137 (I know it's not your problem, and I am also aware that this function is not documented, since its private, but my check script pointed these lines out to me :) )

hiker (changes requested):
Nice job - I like the type annotations  :) I had a question about the implementation, and otherwise mostly style issues.

src/psyclone/psyir/nodes/array_mixin.py
sergisiso: Done

src/psyclone/psyir/nodes/array_mixin.py
sergisiso: Good point, it currently can not be done because the arrayrange2loop transformation uses a single index variable for both ranges, if the array declaration lbound is different then we need to add another expression on one of them to account for the lbound difference between both range expressions.

I will look at this when refactoring the arrayrange2loop, so I added a TODO for that one instead.

This is also similar to what @arporter is doing for WHERE cannonicalisation, so at some point we need to see if there are common functionality to re-use.

src/psyclone/psyir/transformations/arrayrange2loop_trans.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/arrayrange2loop_trans_test.py
sergisiso: Fixed

src/psyclone/tests/psyir/nodes/array_mixin_test.py
sergisiso: Done

src/psyclone/psyir/nodes/array_mixin.py
sergisiso: Done both

sergisiso:
@hiker This is ready for next review

src/psyclone/psyir/nodes/array_mixin.py
hiker: Great, thanks!

src/psyclone/tests/psyir/nodes/array_mixin_test.py
hiker: Thanks!

hiker:
All good. Am waiting for CI to finish, then I'll merge it in

hiker (approved):
CI is green, good to go.

Issue #2224, opened 2023-07-20
Sphinx linkcheck fails for an anchor in a GH-generated page


arporter:
Recently the Sphinx link checker has been failing for:

https://github.com/stfc/PSyclone/tree/master/tutorial/practicals#user-content-netcdf-library-lfric-examples

I found https://github.com/sphinx-doc/sphinx/issues/9255 and established that Sphinx has recently been updated so that it 'knows' about GitHub and automatically tries prepending the "user-content-" to anchors. However, this particular page is a GH-generated rendering of a file in the source tree. If I 'wget' the URL then I can see code that defines the anchor:

![image](https://github.com/stfc/PSyclone/assets/17409212/f629cfd1-ac04-49d0-88fb-41ec73b19478)

but the link checker doesn't find it. It's not clear to me whether this is due to a recent change in Sphinx or in the way GitHub generates the page. For now, we can simply workaround this problem by configuring linkcheck to ignore this particular anchor so that's what I'll do in this issue.

PR #2470, opened 2024-01-12
#2468 Fixed anchor to README

### LINKED ISSUE #2468: Sphinx linkcheck stating Anchors are broken
In the sphinx-link checks, the anchors on https://github.com/stfc/PSyclone are flagging as broken despite being navigable to.

Currently, the only anchor in the docs is https://github.com/stfc/PSyclone#user-content-try-it-on-binder which _should_ be #try-it-on-binder but both direct to the same place.

oakleybrunt:
Closes #2468


sergisiso:
Unfortunately, GitHub anchors do not pass linkcheck (this is a known problem of linkcheck), in the past prefixing them with user-content did the trick but it seems this doesn't work anymore. Also in `doc/user_guide/conf.py` there is the ` linkcheck_anchors_ignore` parameter that maybe can be configured to ignore some of these patterns. Maybe @arporter has looked at this in the past?

But then we need to keep track of them. For instance, `https://puma.nerc.ac.uk/trac/GOcean` is ignored by `linkcheck_ignore` but it is broken. @arporter Is there something else where we can reference for GOcean?

oakleybrunt:
> in `doc/user_guide/conf.py` there is the ` linkcheck_anchors_ignore` parameter that maybe can be configured to ignore some of these patterns. Maybe @arporter has looked at this in the past?

I did try to add `linkcheck_anchors_ignore_for_url` in `doc/user_guide/conf.py`:

> A list or tuple of regular expressions matching URLs for which Sphinx should not check the validity of anchors. This allows
skipping anchor checks on a per-page basis while still checking the validity of the page itself. Default is an empty tuple ()

but it didn't work.

arporter:
A GOcean link could be: https://gtr.ukri.org/projects?ref=NE%2FL01209X%2F1&pn=0&fetchSize=25&selectedSortableField=date&selectedSortOrder=ASC

arporter:
I think I did previously investigate. After some searching I found: https://github.com/stfc/PSyclone/issues/2224

arporter:
If the `linkcheck_anchors_ignore` really doesn't work then we should open a ticket on Sphinx. Is it possible to see what changed that caused this check to start failing? Presumably GHA upgraded the version of Sphinx it was installing?

sergisiso:
There are a few related issues already opened in the Shpinx tracker and they have struggled with this for years. We could ignore any link to the PSyclone Readme that has an anchor. I made a commit directly in this branch with it, I hope it ok. What do you think of this solution?

arporter:
> There are a few related issues already opened in the Shpinx tracker and they have struggled with this for years. We could ignore any link to the PSyclone Readme that has an anchor. I made a commit directly in this branch with it, I hope its ok. What do you think of this solution?

It looks OK to me. As you say, it's not ideal but it seems the best we can do at the moment and at the end of the day, it's only links after all.

sergisiso:
If @oakleybrunt is ok with it we could put this for review as other branches are always failing the checks because of these links. @arporter I see there is links to the GOcean puma server in some fortran copyright files, do you think we also need to change those?

oakleybrunt:
Go ahead @sergisiso

arporter:
> I see there is links to the GOcean puma server in some fortran copyright files, do you think we also need to change those?

Yes please. Might as well get it fixed now.

sergisiso:
As discussed, the remaining references are in license headers and to Gung-Ho, so we will do them on a separate PR.

Therefore, this is ready for review @arporter

doc/user_guide/conf.py
```diff
@@ -372,11 +372,11 @@
...
 # MyBinder fails on a very regular basis so we skip those links.
-# The puma site no longer exists but the GOcean documentation needs to be
-# updated (TODO #237).
-# Despite its name 'collab.metoffice.../twiki' requires authenticated access.
 linkcheck_ignore = [r'^https://mybinder.org/v2/gh/stfc/psyclone',
-                    r'https://puma.nerc.ac.uk/trac/GOcean',
+# Shpinx has problems wiht Github anchors, so we skip the links to anchors to
+# the main README
+                    r'^https://github.com/stfc/PSyclone#',
+# Despite its name 'collab.metoffice.../twiki' requires authenticated access.
```
arporter, line 379: To be fair to them, it's no longer called 'collab' so we could just shorten (to "Requires authentication") but I'll probably do that if I'm happy with the rest of the PR.

doc/user_guide/gocean1p0.rst
```diff
@@ -64,9 +64,9 @@ The GOcean Library
 
 The use of PSyclone and the GOcean 1.0 API implies the use of a
 standard set of data types and associated infrastructure. This is
-provided by version 1.0 of the GOcean Library (GOLib v.1.0).
-Currently this library is distributed separately from PSyclone and is
-available from https://puma.nerc.ac.uk/trac/GOcean.
+provided by the dl_esm_inf library. Currently this library is distributed
```
arporter, line 67: Thanks for updating this. Please could you also change the section header and the paragraph above so that they refer to the "GOcean infrastructure library - dl_esm_inf".

arporter (changes requested):
Thanks for this fix.
Link check is now happy.
I just have a couple of small requests and then this can go on.

doc/user_guide/conf.py
sergisiso: Done

doc/user_guide/gocean1p0.rst
sergisiso: Replaced all GOLib references

sergisiso:
I made the requested changes, this is ready to review again @arporter

arporter (approved):
All good now, thanks for making those changes.
User guide builds without any warnings and update sections look good.
This is only a documentation change so no need to run the integration tests.
Will proceed to merge.

PR #2474, opened 2024-01-12
#2409 Comments for PSy-Layer Builtins

### LINKED ISSUE #2409: LFRic PSy-layer comments for builtins
At the moment it can be difficult to determine the type of builtin that is coded in the PSy-layer. If a comment were added before each builtin stating what type of builtin it is this could make the code more readable to the scientist.

oakleybrunt:
Closes #2409

This change implements a new class method in the `LFRicBuiltIn` class which is optionally called by it's children in their `lower_to_language_level` method. The class method uses the child's attributes to fill a preceding comment with a summary of the builtin's use, e.g. `! Built-in: setval_c (set a real-valued field to a real scalar value)`.

This will also, indirectly, allow much faster searching of tests using a specific builtin since the assertion will (usually) include a check against a builtin's output.


src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -557,6 +577,8 @@ def lower_to_language_level(self):
         rhs = BinaryOperation.create(BinaryOperation.Operator.ADD,
                                      arg_refs[1], arg_refs[2])
         assign = Assignment.create(arg_refs[0], rhs)
+        # Add a preceding comment to the Assignment
+        assign = LFRicBuiltIn.add_preceding_comment(self, assign)
```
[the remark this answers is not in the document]
arporter: It's unfortunate (and not your fault) that the current design means we get this exact line duplicated in every `lower_to_language_level`. I see in fact that we already have the last three or so statements duplicated in every such method. Therefore, I think that instead of `add_preceding_comment` we have a 'private' `_create_assignment(lhs, rhs)` and then we can replace everything after (and including) `assign = Assignment.create(lhs, rhs)` with `return self._create_assignment(lhs, rhs)`. It's a slightly bigger change but should result in fewer lines of code overall.

arporter (changes requested):
Thanks for this @oakleybrunt, it's a good change and well executed. However, it highlights that we already have a fair bit of code duplication in the various `lower_to_language` methods of the LFRicBuiltin subclasses. I've therefore made a suggestion to replace your new method with another one (sorry!) that does a bit more. This will reduce code duplication in the `lower_to_language` methods. Note that you need to check your code with pytcodestyle/pylint as there are a few lines that are too long.
I've not looked at any of the tests or checked the reference guide yet.

src/psyclone/domain/lfric/lfric_builtins.py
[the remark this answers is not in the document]
oakleybrunt: I was able to do this for all but one builtin - LFRicSetvalRandomKern. This is because there is no assignment node here, only IntrinsicCall.

src/psyclone/tests/domain/lfric/lfric_builtins_test.py
```diff
@@ -123,6 +123,7 @@ class Dummy3(Dummy2):
 # ------------- Tests for built-ins methods and arguments ------------------- #
 
 
+# pylint: disable=invalid-name
```
arporter, line 126: This should not be required as long as you run pytest from within the `src/psyclone/tests` directory. Please could you double-check?

src/psyclone/tests/domain/lfric/builtins/reduction_builtins_test.py
```diff
@@ -135,11 +135,13 @@ def test_multi_builtin_single_invoke(tmpdir, monkeypatch, annexed, dist_mem):
             "      asum = 0.0_r_def\n"
             "      !\n"
             "      DO df=loop0_start,loop0_stop\n"
+            "        ! Built-in: X_innerproduct_Y (real-valued fields)\n"
```
arporter, line 138: This looks really good.

arporter (changes requested):
Very nearly there now.
Ref. guide builds with no new warnings. I've just triggered the integration tests.

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -494,13 +494,30 @@ def get_scalar_argument_references(self):
...
         '''
         return [arg.psyir_expression() for arg in self._arguments.args
                 if arg.is_scalar]
 
+    def _create_assignment(self, lhs, rhs):
+        '''
+        Creates an assignment from left-hand- and right-hand-side ArrayReference
+        nodes, then replaces the Builtin with the newly created Assignment
+        node.
+
```
arporter, line 509: You need to document the `lhs` and `rhs` arguments.

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -3252,16 +3135,14 @@ def lower_to_language_level(self):
...
             [arg_refs[1], ("kind", Reference(r_precision))])
-        assign = Assignment.create(arg_refs[0], rhs)
-        # Add a preceding comment to the Assignment
-        assign = LFRicBuiltIn.add_preceding_comment(self, assign)
-        # Finally, replace this kernel node with the Assignment
-        self.replace_with(assign)
-        return assign
+
+        # Create assignment and replace node
+        return self._create_assignment(lhs, rhs)
```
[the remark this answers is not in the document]
oakleybrunt: Yes, I agree. Let's get as self-descriptive as we can here.

src/psyclone/tests/domain/lfric/lfric_builtins_test.py
oakleybrunt: Apologies, I added this to filter out all of the `invalid-name` warnings from `pylint` to see if there were any warnings I had created. It is definitely not needed.

src/psyclone/domain/lfric/lfric_builtins.py
arporter: I've taken the liberty of correcting what you had for rhs (it might not be an array reference).

arporter:
I was all set to merge this but the (newly-automated) pylint checks are highlighting some problems. Please could you fix them. Note you'll need to pull this branch before you modify it as I've updated it.

oakleybrunt:
Codecov is not uploaded but this is ready for review

arporter (approved):
Looking good now. Just waiting on CI and integration tests.

arporter:

arporter:
Just waiting on final CI run now.

Issue #2463, opened 2024-01-11
Handling of Namelist files


hiker:
Trying to use PSyclone with the Nemo API on UM physics code showed that the handling of namelist files need to be improved. ATM, if a namelist statement is found in the source code, the whole module is converted to a code block, so we are getting a useless PSyIR (of just one node, instead of just having the namelist declaration as an Unknown Fortran).

While the namelist statement can be fixed in UMN-LFRic physics (as far as I can tell no namelists are actually read by LFRic, e.g. the reading of the namelist file is #ifdef'ed out for LFRic), this need to be handled better, since it takes a long way to find out why a file is not parsed. fparser itself handles it just fine, and most of PSyclone as well (as one big code block :) ), only when accessing symbol table etc things go wrong.

FWIW, the original exception is raised in line 2536 of fparser2, with node being:
```
NAMELIST /run_bl/ i_bl_vn, 
```
a `Namelist_Stmt` (and the rest of the code, e.g. the actual subroutines is not even touched anymore).

Even if this is not improved, we have to print a proper error message to guide the user to the problem. Identifying which line is causing fparser2 to fail can take a long time!

arporter:
At least part of the answer to this is logging.

PR #2477, opened 2024-01-15
#2463 Support namelist in PSyIR (as an UnknownFortranType).

hiker:
This basically mirrors the handling of common blocks - adding a namelist as an UnknownFortranType (with `_PSYCLONE_INTERNAL_NAMELIST` as name). For @sergisiso I'd guess.


hiker:
Hmm - I found a potential bug(?) here, but am not sure what happens. UM contains something like:
```
real my_Var = rdmi    ! Imported value to indicate invalid value
...
namelist /bla/ my_far, ...
```
Then the Fortran backend throws an error, that an initiatlisation needs a static interface (while my_var now has namelist interface). I tried a simple patch:
```
diff --git a/src/psyclone/psyir/backend/fortran.py b/src/psyclone/psyir/backend/fortran.py
index 1e1c793ab..67e4e8058 100644
--- a/src/psyclone/psyir/backend/fortran.py
+++ b/src/psyclone/psyir/backend/fortran.py
@@ -52,7 +52,7 @@ from psyclone.psyir.symbols import (
     ArgumentInterface, ArrayType, ContainerSymbol, DataSymbol, DataTypeSymbol,
     DeferredType, RoutineSymbol, ScalarType, Symbol, IntrinsicSymbol,
     SymbolTable, UnknownFortranType, UnknownType, UnresolvedInterface,
-    StructureType)
+    StructureType, NamelistInterface)
 
 
 # Mapping from PSyIR types to Fortran data types. Simply reverse the
@@ -607,7 +607,8 @@ class FortranWriter(LanguageWriter):
                 symbol.initial_value):
             result += " = " + self._visit(symbol.initial_value)
         elif isinstance(symbol, DataSymbol) and symbol.initial_value:
-            if not symbol.is_static:
+            if not symbol.is_static and \
+                    not isinstance(symbol.interface, NamelistInterface):
                 raise VisitorError(
                     f"{type(symbol).__name__} '{symbol.name}' has an initial "
                     f"value ({self._visit(symbol.initial_value)}) and "
```
That seems to work, but I am not sure if this is a proper fix - e.g I believe Fortran should allow a variable to be initialised without parameter. Is this a PSyIR assumption? The other option might be to make NamelistInterface inherit from StaticInterface ... but that feels totally wrong :)

arporter:
Hi @hiker, static is correct here because, in Fortran, providing an initial value to a local variable then implies it has the SAVE attribute. I think `my_var` should still have a static interface. The problem is that the namelist doesn't actually *declare* the variable, it just says its value can be populated from a namelist. i.e. we need both the declaration and the namelist and I don't think we can do that with an UnknownFortranType?

hiker:
> Hi @hiker, static is correct here because, in Fortran, providing an initial value to a local variable then implies it has the SAVE attribute.

Dang - I just realised that I read this a while ago (Fortran pitfalls or so). Who on earth came up with this language in the first place??? ;) Yes, I see the history ... no recursion, so it was easiest not to bother with a stack, so all was static, and that became the semantic when assigning a value :) I would have loved to listen to the committee discussion about that :)

> I think `my_var` should still have a static interface. The problem is that the namelist doesn't actually _declare_ the variable, it just says its value can be populated from a namelist. i.e. we need both the declaration and the namelist and I don't think we can do that with an UnknownFortranType?

I also just realised that `is_static` actually tests the interface (have I mentioned I hate properties??). I can think of four solutions:
1. We add two new interfaces: NamelistInterface, StaticNamelistInterface ... though I am the first one to admit that sucks :)
2. We allow a symbol to have several interfaces ... and that just sounds like a very bad idea :)
3. We store the namelist information in an additional attribute (e.g. `self._namelist_name`, defaulting to None meaning no namelist or so). We can't easily get the namelist name out of the UnkownFortranType (unless we fix the corresponding ticket), otherwise we could just point to the 'artificial' namelist symbol.
4. We just ignore the information that a variable is in a namelist, ie. I remove the whole NamelistInterface, and don't change the interface of the symbol.

I kind of tend to prefer 4 - it is KISS, and given that we don't really handle namelist in the first place, I don't think we need that information anyway. It would certainly be good enough for my use case (just to avoid that a file with a namelist statement ends up as one big CodeBlock). Would that be ok?

hiker:
I've gone the KISS way, basically just removing the `NamelistInterface`, so I don't store any details to the symbol that it is part of a namelist. I am happy to add this if someone tells me the best way. My idea would be to add an attribute and store info there, and this attribute would just be None otherwise. I am not sure if I should check if all symbols in the namelist statements are defined, and ... what to do if they are not ... raise an exception I'd guess? Or declare following Fortran deprecated implicit declarations (shiver)?????

I've noticed it is marked as under review, I've added 'ready for review', to make it more confusing :)

sergisiso:
No worries I just started reading the code yesterday and still didn't make any comments. I was also leaning to propose to go with option 4 so I am happy with the new changes. I will restart/continue the review today.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2647,38 +2661,39 @@ def _process_common_blocks(nodes, psyir_parent):
...
+                            raise NotImplementedError(
+                                f"Symbol '{sym.name}' has an initial value"
+                                f" ({sym.initial_value.debug_string()}) "
+                                f"but appears in a common block. This is "
+                                f"not valid Fortran.")
+                        sym.interface = CommonBlockInterface()
+            except KeyError as error:
+                raise NotImplementedError(
+                    f"The symbol interface of a common block variable "
+                    f"could not be updated because of {error}.") from error
```
sergisiso, line 2696: Isn't this the same? I think I prefer to leave it as it was.

src/psyclone/psyir/symbols/__init__.py
```diff
@@ -33,32 +33,35 @@
...
 from psyclone.psyir.symbols.symbol_table import SymbolTable
 from psyclone.psyir.symbols.typed_symbol import TypedSymbol
-from psyclone.psyir.symbols.datatypes import DataType, ScalarType, \
-    ArrayType, TYPE_MAP_TO_PYTHON, REAL_TYPE, REAL_SINGLE_TYPE, \
-    REAL_DOUBLE_TYPE, REAL4_TYPE, REAL8_TYPE, INTEGER_TYPE, \
-    INTEGER_SINGLE_TYPE, INTEGER_DOUBLE_TYPE, INTEGER4_TYPE, INTEGER8_TYPE, \
-    BOOLEAN_TYPE, CHARACTER_TYPE, DeferredType, UnknownType, \
-    UnknownFortranType, StructureType, NoType
+from psyclone.psyir.symbols.datatypes import \
+    (ArrayType, BOOLEAN_TYPE, CHARACTER_TYPE, DeferredType, DataType,
```
sergisiso, line 54: I think its a bit nicer to avoid the backslash by starting the parenthesis in the first line, then there can be the linebreak just after the parenthesis.

src/psyclone/tests/psyir/symbols/interfaces_test.py
```diff
@@ -41,10 +41,10 @@
 
 import pytest
 
-from psyclone.psyir.symbols.interfaces import SymbolInterface, \
-    AutomaticInterface, ArgumentInterface, ImportInterface, \
-    UnresolvedInterface, StaticInterface, DefaultModuleInterface, \
-    CommonBlockInterface, UnknownInterface
+from psyclone.psyir.symbols.interfaces import \
+    (AutomaticInterface, ArgumentInterface, CommonBlockInterface,
```
sergisiso, line 45: Same here, you can get rid of the backslash by bringing the parenthesis up.

sergisiso (changes requested):
@hiker I am happy with this initial support of Namelist, I just made some formatting comments. Similar to commonblocks this won't survive symbol renaming or complex transformations but it is good enough to preserve the original statement without putting the whole routine in a codeblock, which is often all we need.

arporter:
> Similar to commonblocks this won't survive symbol renaming or complex transformations but it is good enough to preserve the original statement without putting the whole routine in a codeblock, which is often all we need.

Does this mean we need to extend some of the transformations (e.g. inlining) to abort if they encounter namelist symbols?

src/psyclone/psyir/frontend/fparser2.py
hiker: This requires a pylint disable, since the nesting is too deep. I often try to reduce nesting level (since I find it easier to read), but I'll fix it up.

src/psyclone/psyir/symbols/__init__.py
hiker: Just to double check: if I add the parenthesis at the line above, then all following imported symbols must be aligned to the opening parenthesis, or else we get a pycodestyle error (continuation line under-indented). Given that the `from ... import` is already long, this means the 4 or 5 lines of import we have atm, become 12 lines. Similarly, the 3 lines above importing from interface become 8 lines.

TBH, I don't really care. I personally prefer to have the import statements more compact. So, unless you tell me to leave it, I'll fix this according to your review (tomorrow).

hiker (commented):
Just two quick style questions before I work on this (tomorrow) :)

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Ah I see, but I it seems simpler to have positive conditions and no continues.

I also prefer not disabling this kind of pylint warnings, it is a reminder than the code can be improved and is the developer/reviewer who decide what's best. Although we never be able to go to 0 warnings, but I thinks that fine.

src/psyclone/psyir/symbols/__init__.py
sergisiso: I mean like in `src/psyclone/psyir/nodes/__init__.py`, it is just substituting the backslash by the opening parenthesis, all the imports are still at the same place as now.

sergisiso:
@hiker Unfortunately this NEMO integration test failed with `PSyclone failed on the following source files: ['/home/gh_runner/NEMO/cfgs/SPITZ12_serial/BLD/ppsrc/nemo/stopar.f90']`. If you don't have access to the files I will explore what it is this afternoon

sergisiso:
> Can you inline if you have any UnknownFortranType? Each of them will contain symbols that can't be renamed.
Additionally, with the current approach you can't easily tell if a symbol is in a namelist (since for now I've implemented this to not add anything to the symbols) - so you would have to parse the UnknownFortranTypes. I can easily add it in (since I've done it in my original commit :) ). Alternatively, I'd guess I could also implement a UnknownNamelistFortranType (giving us access to the name of the namelist and symbols etc)... but my feeling is we don't want to go that way, since this would extend the PSyIR with something that is very Fortran specific.

I wouldn't complicate things for now and leave it as it is. I hope the inline does not allow UnknownTypes as there are many ways that this could make it fail, not just namelists.

src/psyclone/tests/psyir/symbols/interfaces_test.py
hiker: Gee, you are right, I had no idea. I always had at least one term still in the previous line. Totally agreed, that's a much better solution, thanks!

hiker:
> @hiker Unfortunately this NEMO integration test failed with `PSyclone failed on the following source files: ['/home/gh_runner/NEMO/cfgs/SPITZ12_serial/BLD/ppsrc/nemo/stopar.f90']`. If you don't have access to the files I will explore what it is this afternoon

I have a version of that file, not sure if it is the same, but it does crash in the same way. But I *THINK* it's actually the omp script that is broken, it was just not discovered previously, because the whole subroutine was turned into a codeblock. If I take my stopar.F90 file on current master and run:
```
joerg@Joerg-Surface:~/work/psyclone/src/psyclone$ psyclone -api nemo -s ../../examples/nemo/scripts/omp_gpu_trans.py ./stopar.F90  | grep ^sto_
sto_par
sto_rst_read
sto_rst_write
sto_par_white
sto_par_flt
sto_par_flt_fac
```
But if I comment out the namelist statement in sto_par_init, on master I get:
```
<skipping error trace as in CI>
sto_par
sto_par_init
```
`sto_par_init` was NOT listed in the previous output. And I get the same output (including `sto_par_init` that is) when I use the file **with** the namelist statement and my branch.

I think the script never ran on `sto_par_init` (because the subroutine turned out to be a codeblock). Now it does, and it crashes. I'd think that the script has an error.

sergisiso:
Ah, this is not a performance sensitive region, you could add a: `if invoke.name == "sto_par_init": continue` check in the script. (probably the omp_cpu_trans and acc_loops_trans will need the same)

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2532,6 +2534,18 @@ def process_declarations(self, parent, nodes, arg_list,
                     raise NotImplementedError(
                         f"Error processing implicit-part: implicit variable "
                         f"declarations not supported but found '{node}'")
+
+            elif isinstance(node, Fortran2003.Namelist_Stmt):
+                # Place the declaration statement into a using an internal
```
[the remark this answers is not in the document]
hiker: Done

src/psyclone/psyir/frontend/fparser2.py
hiker: OK, I reverted that back to the original design, but without the pylint

src/psyclone/psyir/symbols/__init__.py
hiker: Thanks, I never realised that this was possible :)

src/psyclone/tests/psyir/symbols/interfaces_test.py
hiker: Done

hiker:
I've tried to 'properly' fix the script (i.e. without hardcoding a name). It seems to work, but the nemo compilation is still failing, but as far as I can see with compiler errors:
```
nvfortran -c -o bdytides.o -I/home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/BLD/inc -i4 -Mr8 -g -O3 -Mflushz -Mfma -Minline -fastsse -Mcray=pointer -Mvect=prefetch,tile,fuse -Mpre -mp=gpu -gpu=managed -I/home/gh_runner/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/lib/profiling/nvidia/ -I/home/aporter/NEMO/xios_trunk/inc -I/not/defiled/build/lib/mct -I/not/defiled/build/lib/psmile.MPI1 -I/apps/packages/netcdf_fortran/4.6.1-nvidia-hpcsdk/include -c /home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/WORK/bdytides.f90
NVFORTRAN-S-0074-Illegal number or type of arguments to ubound - keyword argument array (/home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/WORK/bdytides.f90: 185)
NVFORTRAN-S-0074-Illegal number or type of arguments to lbound - keyword argument array (/home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/WORK/bdytides.f90: 186)
NVFORTRAN-S-0082-Illegal substring expression for variable filtide (/home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/WORK/bdytides.f90: 190)
```

My version of that file is processed fine with PSyclone, but I don't have a compile environment for nemo :( As far as I can tell it appears a similar problem: a subroutine is now handled (i.e. modified), which wasn't before. So I am again not sure if this is the problem of this PR. Could you have a check of what the problem is?

arporter:
The problem here is almost certainly the character manipulation within an offload region - that just doesn't work with NVIDIA. We should catch this but don't currently. I have a small patch to the `acc_kernels_script` that checks for character literals in assignments but that should probably(?) go into the relevant `validate` methods as it clearly applies to OMP offload too.

hiker:
> The problem here is almost certainly the character manipulation within an offload region - that just doesn't work with NVIDIA. We should catch this but don't currently. I have a small patch to the `acc_kernels_script` that checks for character literals in assignments but that should probably(?) go into the relevant `validate` methods as it clearly applies to OMP offload too.

It sounds like the best option for me to progress this PR is to start excluding certain subroutine names :(

arporter:
If I was you @hiker, I'd extend the scripts so that they don't attempt to do anything with any routine ending in `_init` (or that have a namelist in it perhaps?).

hiker:
> If I was you @hiker, I'd extend the scripts so that they don't attempt to do anything with any routine ending in `_init` (or that have a namelist in it perhaps?).

I don't feel comfortable making the first decision (I don't know if there isn't an `_init` that can make use of threading. And we should certainly not disable anything with namelist in them, since that has nothing to do with the actual failure (just that often something using namelist uses strings). But the good news is I do have all compile CIs fixed (well, subroutines excluded :) ). And brought it up to latest master :)

src/psyclone/psyir/symbols/__init__.py
```diff
@@ -33,32 +33,35 @@
...
 from psyclone.psyir.symbols.containersymbol import ContainerSymbol
 from psyclone.psyir.symbols.data_type_symbol import DataTypeSymbol
-from psyclone.psyir.symbols.interfaces import AutomaticInterface, \
-    ImportInterface, ArgumentInterface, UnresolvedInterface, StaticInterface, \
-    DefaultModuleInterface, UnknownInterface, CommonBlockInterface
+from psyclone.psyir.symbols.interfaces import (
+     ArgumentInterface, AutomaticInterface, CommonBlockInterface,
+     DefaultModuleInterface, ImportInterface,
+     StaticInterface, UnknownInterface, UnresolvedInterface)
 from psyclone.psyir.symbols.intrinsic_symbol import IntrinsicSymbol
```
sergisiso, line 48: Not your fault but I noticed that IntrinsicSymbol is missing from the `__all__` list

src/psyclone/psyir/frontend/fparser2.py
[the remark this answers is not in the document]
sergisiso: The response for this and the next comment seems that you addressed the issues, but I don't see any code changes, is there a commit missing?

sergisiso (changes requested):
@hiker This is almost ready, I marked a couple of things that I could have fixed myself. But the main reason to not approve it yet is that this may also break (with no fault of its own) some of the other NEMO, NEMOVAR, wavewatch, ... projects that are waiting for a PSyclone release.

We could:
- First make the release and then merge this.
or
- Merge this and also fixes to the omp_target/acc_kernel/... validations to check for strings in that region before the release

But I want to discuss this with @arporter first and experiment with the string exclusions to know what performance impact it has before deciding.

src/psyclone/psyir/frontend/fparser2.py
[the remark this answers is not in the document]
hiker: No idea what happened, I am pretty certain I added this ... anyway, I've done it now ... I hope :)

src/psyclone/psyir/symbols/__init__.py
hiker: Done.

src/psyclone/psyir/frontend/fparser2.py
hiker: Again, no idea where this ended up. I've changed that back.

hiker:
I can't answer your comment ... not the first time, am I doing something wrong? Anyway

> The response for this and the next comment seems that you addressed the issues, but I don't see any code changes, is there a commit missing?

No idea, I added the missing words. And similarly, the other change was the change in indentation that I did? I've also removed this (and again, I was sure I did that before :) ).

Re delay: I don't mind if this is merged after the release, I am not in a hurry, and I think keeping this up-to-date with master should be reasonable easy (as long as no one starts playing with fparser :)  )

sergisiso:
@hiker Now that the release is out we can proceed with this, I brought it to master and launched a final integration test to see if everything is still fine.

sergisiso:
@hiker  Unfortunately the integration test still fails with:
 `NVFORTRAN-F-0155-Compiler failed to translate accelerator region (see -Minfo messages): Character local variable unsupported on device home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/WORK/sbccpl.f90: 503)`

hiker:
Really strange - I checked, I had it pass the nemo integration test. I did notice though that in that passing test I had the message:
``` 
Skipping /home/gh_runner/NEMO/cfgs/SPITZ12_serial/BLD/ppsrc/nemo/sbccpl.f90 entirely.
```
Ah, and there was a comment that the 'where' processing causes this to fail. IIrc a where-improvement was recently added, so it would be a combination of new where processing and namelist support triggering this now. OK, I'll try to fix (i.e. ignore) stuff now :)

hiker:
Looks like it's fixed now. Ready to go again :)

sergisiso (approved):
@hiker the integration tests are now green, thanks for your patience on this. The PR is now ready to be merged.

PR #2481, opened 2024-01-17
(Closes #1193) unpin jsonschema

### LINKED ISSUE #1193: Drop Python 3.5?
The notebook tests are failing for Python 3.5. Possibly this is because Jupyter has dropped support for Python 3.5 (https://github.com/jupyter/notebook/pull/5962). Apparently 3.5 was EOL last September.

As discussed at https://github.com/jupyter/notebook/issues/5961#issuecomment-768662236), we could choose to pin the version of Jupyter we use but I think a better option is to drop 3.5.

arporter:
While we were still supporting Python 3.5 we had to pin jsonschema (to 3.0.2). However, that's now causing me problems with more recent versions of other packages so this PR simply removes the pin.


arporter:
I've fired off the integration tests...

arporter:
CI all fine (apart from link check) and integration tests all fine. Ready for review from anyone!

sergisiso (approved):
@arporter I am happy with this since it passes the tests of the currently supported python versions. This is ready to merge

PR #2484, opened 2024-01-19
(towards #2062) Add Node.origin_string method to retrieve original filename and line number information

### LINKED ISSUE #2062: New node method to report where nodes come from
During my recent meetings I received some more than once the suggestion that we could improve the error messages, I think the problem is that we often just print the node (view()) causing the error but it difficult to know where it comes from.

In #1587 I proposed the `Node.debug_string()` to improve some error output, but in addition we could also introduce a `Node.origin_string()` which which could:
- for nodes that come from a source file and have a _ast link print:
```
Node ArrayReference[name:'array1'] from line XX of mysource.f90:
> array(i,j) = scalar + array1(i,j)
                        ^^^^^^ 
```
(Looking at fparser node I found there is the item with the line number and start/end position on this line, so the previous message should be possible. However, I have not found the filename on the fparser node, @arporter @rupertford do you know if the filename is stored somewhere?)

- For nodes dynamically created we could use the python inspect capabilities to say where they were created (but I would try to make them user-facing by reporting something in the stack that the user understands) e.g:
`Node 'ArrayReference[name:'array1']' created in TransformationXXX`
or
`Node 'ArrayReference[name:'array1']' created in 'psyclone_script.py' line XX`

This is quite a bit of work: getting the information, checking that the _ast links are not broken (they aren't tested), updating error messages in a sensible way... but I think we should start moving on this direction.

@rupertford @arporter @hiker @TeranIvy I will appreciate any opinions on this?

sergisiso:
It does not work for all cases (it returns <unknown> fields when some information is unavailable), but it is already useful for many cases, so it would be good to have it on the master branch.


sergisiso:
@arporter This is also ready for review.

src/psyclone/psyir/nodes/node.py
```diff
@@ -1629,6 +1629,31 @@ def debug_string(self):
...
+
+        :returns: a string specifing the origin of this node.
+        :rtype: str
+        '''
+        name = self.coloured_name(False)
+        line_span = "<unknown>"
+        original_src = "<unknown>"
+        filename = "<unknown>"
+        from psyclone.psyir.nodes.statement import Statement
+        node = self if isinstance(self, Statement) else self.ancestor(Statement)
```
arporter, line 1645: You can just do `node=self.ancestor(Statement, include_self=True)`

src/psyclone/psyir/nodes/node.py
```diff
@@ -1629,6 +1629,31 @@ def debug_string(self):
...
+        # Try to populate the line/src/filename
+        if node and node._ast:
```
arporter, line 1647: Although this is OK as a first cut, it brings in intimate details of the fparser parse tree which would probably best be isolated in the `fparser2` frontend.  That way, any frontend could provide a `get_origin_string()` or similar. We could query the type of `._ast` to determine which frontend we should use?

arporter (changes requested):
Thanks for this Sergi. Pleasingly simple :-) Only query I have is whether the details of how to query the fparser2 parse tree should be isolated somewhere, perhaps the fparser2 frontend? Please also bring up to master.

src/psyclone/psyir/nodes/node.py
sergisiso: Done

src/psyclone/psyir/nodes/node.py
sergisiso: @arporter I agree with the problem, but I am unsure about the solution.

Currently I am using hasattr and the knowledge of fparser tree composition, which is not ideal.

> We could query the type of ._ast to determine which frontend we should use?

However, if I do that the logic will be in fparser2.py but I will have to import fparser2 library here to compare the type of the _ast node, creating a real dependency to fparser2, which I want to avoid.

I also want to avoid creating the origin string pre-emptively and store it in the node because that is a lot of string manipulation for something that only is needed very occasionally.

Maybe the solution is to add an optional _generate_origin_string attribute to the PSyIR node which is initialised with a Callable (a closure) that generates the string and will allow us to remove the _ast itself. This is a significant change since it requires to update all the fparser2.py PSyIR node creations and it starts complicating things a bit, so I wanted to check with you what do you think

arporter: > However, if I do that the logic will be in fparser2.py but I will have to import fparser2 library here to compare the type of the _ast node, creating a real dependency to fparser2, which I want to avoid.

Could we tag the generated PSyIR with the identity of the frontend used to create it?

sergisiso: Implementing this would need a bit of experimentation, but Grenoble already uses this feature and I would like them to follow master and not a specific branch, so for the time being a renamed this PR to a towards and added a TODO with where we are in this discussion. Let me know if thats ok.

sergisiso:
@arporter This is ready for next review

src/psyclone/psyir/nodes/node.py
arporter: Yes, I think it's OK for now - it's not as though we have a lot of frontends :-)

arporter:
This is just waiting on the integration tests now...

arporter (approved):
Integration tests are all fine.
Will proceed to merge.

Issue #257, opened 2018-12-20
Loop Fusion Transform can create incorrect code


hiker:
The Fuse transformation does not take a potential stencil operation into account. Since now we have stencil information for gocean, this can be tested. Example:
```
  do j
     do i
        f1(i, j, a)    ! Assume a pointwise write kernel for a
     enddo
  enddo
  do j
     do i
        f2(i, j, a)    ! Assume a (010, 010, 010) read-write stencil access for a
     enddo
  enddo
```
In f2 when computing a(i,j) the indices a(i, j+-1) are used, which was defined in f1. But if the outer loops are fused:
```
  do j
     do i
        f1(i, j, a)    ! Assume a pointwise write kernel for a
     enddo
     do i
        f2(i, j, a)    ! Assume a (010, 010, 010) read-write stencil access for a
     enddo
  enddo
```
when now computing a(i,j), again a(i, j+-1) is read. But now a(i,j+1) has not yet been computed in f1 when the value is needed in f2.

Two solutions:
1) we abort the fuse transformation if the above conditions is detected - though needs a careful look at the dependencies.
2) we do some kind of 'loop peeling', which also needs a close look at the dependencies. If required, we could do:
```
  ! Pre-compute two rows
  do j=1, 2
     do i
        f1(i, j, a)    ! Assume a pointwise kernel for a
     enddo
  enddo

  do j
    if(j+2 <= j_max) then
       do i
            f1(i, j+2, a)    ! Assume a pointwise kernel for a
       enddo
     endif
     do i
          f2(i, j, a)    ! Assume a (010, 010, 010) stencil access for a --> now a(i, j+-1) are defined
     enddo
  enddo
```
Or instead of the if test we could run the second j loop from 1 to jmax-2, and have another peeled loop for f2 at the end. The problem with this loop peeling is that it might not work well when parallelising the loop: we might not have enough iteration in the first peeled loop if we parallelise along j, or if we parallelise over i instead of j, we might access data on a remote socket if otherwise parallelisation is along j.

I'm happy to look at this next year.

hiker:
In #2489 some improvements are made to detect invalid loop fusion configuration. As indicated by:
https://github.com/stfc/PSyclone/pull/2489#discussion_r1474362325

These tests are actually skipped in case of a GoLoop (a test for PSyLoop is done before, and these tests are only executed if it is not a PSyLoop).

Unfortunately, we run into a bug in VariableAccesses in case of a kernel. In gocean1p0.py:
[...]

PR #2489, opened 2024-01-25
(Closes#2488) Updating base loop fusion code towards 2488

LonelyCat124:
Added some changes to the base loop fusion code to be a generic (and correct) loop fusion option. There is some stuff I've intentionally skipped over for now, we can either open a TODO or I can add them to this PR if we think its worthwhile, its just a bit messier.


LonelyCat124:
It looks like some of the specialisations in the Nemo implementation (which we felt were probably generic) are breaking examples.

Could I add a force option to skip some of the checks? @arporter @sergisiso

LonelyCat124:
I don't plan to remove the Nemo implementation at this time (it has more restrictions than the generic) - also to avoid breaking any code that is dependent upon it - we could deprecate it at some point though?

sergisiso:
> It looks like some of the specialisations in the Nemo implementation (which we felt were probably generic) are breaking examples.

The generic implementation does not check dependencies at all, I think the proper implementation is the one in NEMO and should be make the dependency.

What are the breaking examples? Are they some false negative generic examples, or some cases in other APIs?

> Could I add a force option to skip some of the checks?

I am ok with the force option.

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -110,6 +113,155 @@ def validate(self, node1, node2, options=None):
                         f"{node1.view()}\n"
                         f"{node2.view()}"))
 
+        vars1 = VariablesAccessInfo(node1)
+        vars2 = VariablesAccessInfo(node2)
+
+        # Check if the loops have the same loop variable
+        loop_var1 = node1.variable
+        loop_var2 = node2.variable
```
sergisiso, line 121: Could you try indenting the part that you added here inside the previous "else:", if the "if isinstance(node1, PSyLoop) and isinstance(node2, PSyLoop):" condition is taken we are in PSyKAL domain and the fact that there are no dependencies is guaranteed by the definition and not code analysis.

LonelyCat124: Done, didn't break anything so probably its fine.

LonelyCat124:
> > It looks like some of the specialisations in the Nemo implementation (which we felt were probably generic) are breaking examples.
> 
> The generic implementation does not check dependencies at all, I think the proper implementation is the one in NEMO and should be make the dependency.
> 
> What are the breaking examples? Are they some false negative generic examples, or some cases in other APIs?

It was breaking gocean example 1, the `validate_written_scalar` function (which in its current implementation is very naive) was failing as one loop read and the other had write accesses, which is forbidden by this method. I'm not sure if they were PSyLoops so having a force option to skip that seemed like an easy initial alternative.

@hiker has issue #257 to look at better dependency analysis for loop fusion - unless this is an immediate need I'd leave that out of this PR for now I think.

LonelyCat124:
I think this is ready for a proper look - happy to add more to it if the reviewer feels its necessary.

LonelyCat124:
There is also potentially some test restructuring I'm looking at. Almost all the tests in the `loop_fusion_test` directory don't actually test that, but instead test the nemo implementation.

arporter:
Sergi and I are both of the opinion that we should try quite hard to remove the Nemo specialisation of the loop-fuse trans in favour of the generic one by moving its functionality over. You say that gocean eg1 fails if you do that. Can you use the `force` option to resolve that for now?

LonelyCat124:
> Sergi and I are both of the opinion that we should try quite hard to remove the Nemo specialisation of the loop-fuse trans in favour of the generic one by moving its functionality over. You say that gocean eg1 fails if you do that. Can you use the `force` option to resolve that for now?

@arporter I think I've already done that - at this point the Nemo specialisation is more restrictive than the generic implementation in this branch, I've only left it in the code to avoid breaking any existing scripts outside of PSyclone - internally at least the only place `NemoLoopFuseTrans` is used is in its own tests, some of which fail for the generic transformation as it allows renaming of loop variables.

sergisiso:
We can remove it already, at the moment is barely used by anybody and we are progressively removing any Nemo specialisation to make it generic.

> Nemo specialisation is more restrictive than the generic implementation in this branch

What restrictions would we be missing with the generic?

Also I noticed the implementation uses:
```
vars1 = VariablesAccessInfo(node1)
vars2 = VariablesAccessInfo(node2)
```
could you see if
```
vars1 = VariablesAccessInfo()
vars2 = VariablesAccessInfo()
node1.reference_accesses(var1)
node2.reference_accesses(var2)
```
gives the same result?, as this is the proper way to allow higher-abstraction nodes to provide domain knowledge to the dependency analysis.

LonelyCat124:
> > Nemo specialisation is more restrictive than the generic implementation in this branch
> 
> What restrictions would we be missing with the generic?

Just no variable renaming if the loops have different loop variables - which i don't think is a good restriction but i didn't want to just remove it in case it was important somewhere.

arporter:
> Just no variable renaming if the loops have different loop variables - which i don't think is a good restriction but i didn't want to just remove it in case it was important somewhere.

I think we can get rid of that. Presumably that was for when we were relying on the loop variable name to tell us the iteration space but we want to be more general than that now.

LonelyCat124:
It should be removed now.

LonelyCat124:
Hmm, some coverage now broke in a openMP directive I didn't touch - I'll take a look.

arporter:
> Hmm, some coverage now broke in a openMP directive I didn't touch - I'll take a look.

Don't worry - I have exactly the same problem in one of my PRs. I suspect it might be something to do with the python bytecode having optimised out the `if this: pass` line as I've tried locally and that line doesn't appear to be hit during testing.

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -129,6 +300,17 @@ def apply(self, node1, node2, options=None):
...
+
+        # If the loop variable isn't the same, we change all uses of the
+        # second loop's variable to reference the first loop's variable.
+        if loop_var1 != loop_var2:
+            references = node2.loop_body.walk(Reference)
+            for ref in references:
+                if ref.symbol == loop_var2:
+                    ref.symbol = loop_var1
+
```
sergisiso, line 313: I was going to ask that we could do an assignment of "Assignment.create(Reference(loop_var2), Reference(loop_var1))", but this would defeat the purpose of what you want to do because then the additional assignment won't allow to collapse it with the parent loop. But maybe its still worth adding it a comment here and at the transformation docstring that this has the side effect that loop_var2 does not have the same value after the loop with the expectation that it isn't used.

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -66,19 +71,31 @@ def validate(self, node1, node2, options=None):
...
                                      a :py:class:`psyclone.psyir.nodes.Loop`.
         :raises TransformationError: if one or both Nodes are not fully-formed.
         :raises TransformationError: if the Nodes do not have the same parent.
-        :raises TransformationError: if the Nodes are not next to each \
+        :raises TransformationError: if the Nodes are not next to each
                                      other in the tree.
-        :raises TransformationError: if the two Loops do not have the same \
+        :raises TransformationError: if the two Loops do not have the same
                                      iteration space.
+        :raises TransformationError: if either Loop contains a reference to
```
sergisiso, line 88: Maybe it would be clearer as: "if the loops don't share the iteration variable but make use of the other loop's variable in their body"

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -51,6 +54,8 @@ class LoopFuseTrans(LoopTrans):
     PSyIR of a Schedule after performing validity checks for the supplied
     Nodes. Examples are given in the descriptions of any children classes.
 
+    If loops have different named loop variables, the loop variable of the
+    second loop will be renamed if possible before fusion.
```
sergisiso, line 58: Looking at the number of remaining TODOs I think it would be good to mention in the docstring that:
"Note that the validation of this transformation has still several shortcomings, specially for domain API loops, use at your own risk."

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -110,6 +127,160 @@ def validate(self, node1, node2, options=None):
...
+                # loop accesses the other loops variable symbol.
+                # If so, then for now we disallow this merge (though we could
+                # in theory allow using the unused one unless both use each others)
+                for signature in vars1:
+                    var_name = str(signature)
+                    if var_name == loop_var2.name:
+                        raise TransformationError(
+                            f"Error in {self.name} transformation. First "
+                            f"loop contains accesses to the second loop's "
+                            f"variable.")
```
sergisiso, line 147: Maybe: `f"variable: '{var_name}'.")`

And the same below

sergisiso (changes requested):
@LonelyCat124 It's great that we could get rid of the NEMO version, I made some minor comments and also suggested that the docstring of this transformation should make clear of the current limitations of the transformation.

sergisiso:
And can this PR be a "closes" instead of a "towards", since other issues are already captured elsewhere?

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -110,6 +127,160 @@ def validate(self, node1, node2, options=None):
...
+    
+                symbol = symbol_table.lookup(signature.var_name)
+                # TODO #1270 - the is_array_access function might be moved
+                is_array = symbol.is_array_access(access_info=var_info1)
+                if not ignore_dep_analysis:
+                    if not is_array:
+                        LoopFuseTrans.validate_written_scalar(var_info1,
+                                                              var_info2)
+                    else:
+                        LoopFuseTrans.validate_written_array(var_info1,
```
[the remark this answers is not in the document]
LonelyCat124: I thought this should be either `LoopFuseTrans` or just nothing - since they're static methods I thought accessing them through self was bad practice? Happy to change it though if I'm wrong

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -110,6 +127,160 @@ def validate(self, node1, node2, options=None):
                         f"{node1.view()}\n"
                         f"{node2.view()}"))
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -66,19 +71,31 @@ def validate(self, node1, node2, options=None):
         :type node2: :py:class:`psyclone.psyir.nodes.Node`
         :param options: a dictionary with options for transformations.
         :type options: Optional[Dict[str, Any]]
+        :param bool options["force"]: whether to force fusion of the
+                                      target loop (i.e. ignore any dependence
+                                      analysis). This only skips a limited
```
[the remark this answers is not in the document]
LonelyCat124: Fixed.

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -66,19 +71,31 @@ def validate(self, node1, node2, options=None):
...
         :raises TransformationError: if the Nodes do not have the same parent.
-        :raises TransformationError: if the Nodes are not next to each \
+        :raises TransformationError: if the Nodes are not next to each
                                      other in the tree.
-        :raises TransformationError: if the two Loops do not have the same \
+        :raises TransformationError: if the two Loops do not have the same
                                      iteration space.
+        :raises TransformationError: if either Loop contains a reference to
+                                     the other Loop's variable and the Loops
+                                     don't share a loop variable.
```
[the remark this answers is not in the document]
LonelyCat124: This isn't raised by this function I don't think, but instead by the static functions - should it still be in the docstring for this function?

src/psyclone/psyir/transformations/loop_fuse_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/loop_fuse_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/loop_fuse_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/loop_fuse_trans.py
LonelyCat124: Added (though I think it should be `{loop_varX.name}` so I've gone with that).

examples/gocean/eg1/runme_loop_fuse.py
```diff
@@ -82,20 +82,26 @@
...
+                 SCHEDULE.children[0].loop_body[1],
+                 options={"force": True})
 FUSE_TRANS.apply(SCHEDULE.children[0].loop_body[0],
-                 SCHEDULE.children[0].loop_body[1])
+                 SCHEDULE.children[0].loop_body[1],
+                 options={"force": True})
 FUSE_TRANS.apply(SCHEDULE.children[0].loop_body[0],
-                 SCHEDULE.children[0].loop_body[1])
+                 SCHEDULE.children[0].loop_body[1],
+                 options={"force": True})
```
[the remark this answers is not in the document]
LonelyCat124: Done

LonelyCat124:
@sergisiso I've made most of the changes, a couple of points I've asked for clarification - let me know what you think.

src/psyclone/psyir/transformations/loop_fuse_trans.py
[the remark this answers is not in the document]
sergisiso: I don't have a strong opinion here. self allow them to be re-implemented by a subclass, but I am not sure if we want this or not here. I will accept it either.

src/psyclone/psyir/transformations/loop_fuse_trans.py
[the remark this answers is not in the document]
sergisiso: I like Transformation validate methods to document all possible fail conditions, since then is easy to read in the documentation what is supported and what not in a single place. So I would still add it here.

Also, arguably the two utility static methods should be private (underscore prefixed) and the Transformation user be unaware that they exist.

src/psyclone/psyir/transformations/loop_fuse_trans.py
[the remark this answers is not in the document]
LonelyCat124: I will leave them as static for now since they don't use either self or cls, but I see your point. I think we can be smarter about calling them using `type(self).` instead of `LoopFuseTrans` so they can be re-implemented by a subclass.

src/psyclone/tests/domain/nemo/transformations/nemo_loop_trans_test.py
```diff
@@ -41,8 +41,9 @@
 import pytest
 from psyclone.psyir.nodes import Loop
 from psyclone.psyir.transformations import LoopTrans
-from psyclone.domain.nemo.transformations import NemoLoopFuseTrans
-from psyclone.tests.utilities import get_invoke
+from psyclone.psyir.transformations.transformation_error import \
```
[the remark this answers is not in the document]
LonelyCat124: Ok , I wasn't sure, removed now.

src/psyclone/psyir/transformations/loop_fuse_trans.py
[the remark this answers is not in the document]
LonelyCat124: Ok, I'll change accordingly.

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -103,18 +125,184 @@ def validate(self, node1, node2, options=None):
...
+            # Check if the loops have the same loop variable
+            loop_var1 = node1.variable
+            loop_var2 = node2.variable
+            if loop_var1 != loop_var2:
+                # If they don't have the same variable, find out if the one
+                # loop accesses the other loops variable symbol.
+                # If so, then for now we disallow this merge (though we could
+                # in theory allow using the unused one unless both use each
+                # others)
+                for signature in vars1:
```
hiker, line 152: Sorry for adding to an ongoing review ... but I just noticed that the loops can be removed by using:
```
if Signature(loop_var2.name) in vars1:
   raise ...
```
(and similarly for the next loop)

src/psyclone/psyir/transformations/loop_fuse_trans.py
```diff
@@ -103,18 +125,184 @@ def validate(self, node1, node2, options=None):
...
                 raise TransformationError(LazyString(
                     lambda node1=node1, node2=node2:
                         f"Error in {self.name} transformation. Loops do not "
                         f"have the same iteration space:\n"
-                        f"{node1.view()}\n"
-                        f"{node2.view()}"))
+                        f"{node1.debug_string()}\n"
+                        f"{node2.debug_string()}"))
+
+            vars1 = VariablesAccessInfo(node1)
```
hiker, line 140: Doesn't this need to be outdented by one level? All these tests are skipped if the loop is a PSyloop. While this might (??) imply that the variable names are identical (since it's code created by PSyclone), this will also skip the variables tests at around line 167.

LonelyCat124: @sergisiso said to do this, I'm not sure

LonelyCat124: Since no tests fail I think its probably fine to have this at this level on indentation?

hiker: The indentation is wrong ... BUT: if you try to fix this (and actually execute this code with e.g. an gocean example). you will get a crash in the dependency tools, since it turns out in case of a kernel we get incorrect data types (I am debugging this atm, string instead of PSyIR nodes). Add a TODO #257, and I am putting some details into that ticket.

sergisiso: I am aware this needs to be outdented at some point. Ideally we want to even be able to fuse domain and generic loops. But this was out of scope for this PR, we just want to maintain the same functionality we had in the nemo loop but without the specialized transformation. I asked @LonelyCat124 to add a note at the transformation docstring about the limitations but we can also add a TODO comment here.

hiker: I've just submitted #2497, which replaces the string in the access information in case of a GOcean kernel with proper PSyIR expressions, so that the partition call (used here) does not fail anymore if it would be outdented.

There is no need to change anything here for now, I am happy to fix this once this PR is merged as part of my PR (I am still fixing CI, so I think you will be in way before me) - it would just be great if you could ping me when merging this PR in case that I miss it :)  Thanks for your work on fusion!

sergisiso (approved):
Thanks @LonelyCat124, all comments are not resolved, I made a minor docstring change and deleted the NemoLoopFuse from the User Guide. The formatting and tests are green and all new code is covered by tests. This is ready to merge.

PR #2493, opened 2024-01-29
(Closes #2492) nemoouterarrayrange bug

### LINKED ISSUE #2492: NemoOuterArrayRange2LoopTrans bug when array-index expression includes array reference
Processing a reference of the form `this(jp)%that(:, map(ij) - np, :)` causes the apply method of `NemoOuterArrayRange2LoopTrans` to fail. The culprit is:

     lhs_array_ref = node.lhs.walk(ArrayMixin, stop_type=Range)[-1]

where we're using the `walk` to find the 'deepest' array reference (i.e. allowing for an access to a member of a derived type). However, that `walk` ends up giving us `map(ij)` in this case which is wrong.

arporter:
Bug fix for the named transformation (required when an array reference itself uses an array reference in an index expression) and also a minor fix to the `utils.py` file in `examples/nemo/scripts` that prevents us attempting to specialise the same Symbol more than once. (Found while processing latest NEMO source.)

### DIFF
--- a/changelog
+++ b/changelog
@@ -147,6 +147,8 @@
 	53) PR #2439 for #2105. Use maxval2loop and atomics to parallelise
 	NEMO stpctl.f90.
 
+	54) PR #2493 for #2492. Fix bug in NemoOuterArrayRange2LoopTrans
+
 release 2.4.0 29th of September 2023
 
 	1) PR #1758 for #1741. Splits the PSyData read functionality into a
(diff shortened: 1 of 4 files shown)


arporter:
Integration tests are running now...

sergisiso (changes requested):
@arporter This fixes a couple of bugs and the implementation is simple, I am happy with the code changes.

But Codecov shows an indirect line miss. Can we explore what is it? Also the integration tests all pass but it may be worth running it again when nobody is on glados to make sure the indirect changes don't have any performance impact.

arporter:
The missed line is weird and @LonelyCat124 has just hit the same thing in his PR. I've tried running the test suite for master on my laptop and that line is not hit. Possibly it is hit in an older version of Python (I have 3.12)? Anyway, it doesn't seem to be a result of this particular PR.

sergisiso (approved):
@arporter Re-running the integration tests shows that all performance match their best numbers. So we can proceed to merge this PR.

PR #2506, opened 2024-02-16
Update CI software dependencies

### LINKED ISSUE #2348: pylint 3.0 does not work with pytest
The tests fail with:
```
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/pytest_pylint/pylint_util.py", line 5, in <module>
    from pylint.interfaces import IReporter
ImportError: cannot import name 'IReporter' from 'pylint.interfaces' (/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/pylint/interfaces.py)
```

nmnobre:
We can also try version 24.1 of nvfortran for the NEMO tests, but I'd rather first make sure these minor updates work before trying to do that, ~since it most likely won't work.~ It doesn't, we'll have to keep using version 23.7.

Closes #2348.


sergisiso:
The CI for NEMO failed with:
```
Lmod is automatically replacing "nvidia-hpcsdk/23.7" with "oneapi/2024.0.1".

Lmod is automatically replacing "nvidia-hpcsdk/23.7" with "oneapi/2024.0.1".

Note: the module "nvidia-hpcsdk" cannot be unloaded because it was not loaded.

Lmod has detected the following error: These module(s) or extension(s) exist
but cannot be loaded as requested: "oclfpga", "mpi", "compiler", "compiler-rt",
"tbb"
   Try: "module spider oclfpga mpi compiler compiler-rt tbb" to see how to load
the module(s).
```

nmnobre:
> The CI for NEMO failed with:
> 
> ```
> Lmod is automatically replacing "nvidia-hpcsdk/23.7" with "oneapi/2024.0.1".
> 
> Lmod is automatically replacing "nvidia-hpcsdk/23.7" with "oneapi/2024.0.1".
> 
> Note: the module "nvidia-hpcsdk" cannot be unloaded because it was not loaded.
> 
> Lmod has detected the following error: These module(s) or extension(s) exist
> but cannot be loaded as requested: "oclfpga", "mpi", "compiler", "compiler-rt",
> "tbb"
>    Try: "module spider oclfpga mpi compiler compiler-rt tbb" to see how to load
> the module(s).
> ```

Yes, of course, I forgot to generate the oneAPI module files, should be fixed now. :)

setup.py
```diff
@@ -170,8 +170,7 @@ def get_files(directory, install_path, valid_suffixes):
             'psydata': ["Jinja2"],
             # TODO #2348: pytest currently fails with pylint >=3
             'test': ["pep8", "flake8", "pylint<3.0", "pytest-cov",
-                     "pytest-pep8", "pytest-pylint", "pytest-flakes",
-                     "pytest-xdist"],
+                     "pytest-pylint", "pytest-flakes", "pytest-xdist"],
```
sergisiso, line 173: We don't use pep8, flake8, pytest-pylint, pytest-flakes I think (@arporter can correct me). The wiki and developer guide mentions pep8, but we actually use pycodestyle. Would it make sense to remove them all and maybe add pycodestyle and pytest-pycodestyle instead?

(althought I tried the second and was not working for me)

sergisiso: Also remove the < 3.0 from pylint?

nmnobre: We do use flake8?

sergisiso: Sorry my grep missed the folder starting with .

@arporter So then we the rule is that we should use flake8 instead of pycodestyle?

nmnobre:
Closes #2348.

setup.py
sergisiso: I saw the setup.cfg but didn't know what was it needed for.

"pylint" is still needed as test dependency since the wiki says the reviewer should use it.

nmnobre: Okay, I'll put it back. I was just trying to see how much faster I could make the CI tests by skipping it. :P

arporter:
Could we run `pycodestyle` on the whole source tree as part of the CI - we've had errors making it onto `master` lately (and I have to admit that sometimes I forget to check myself).

sergisiso:
But we already check it because flake is a superset of pycodestyle, but we have these "--select=E9,F63,F7,F82" that let issues reach master, we just need to remove this flag instead of adding pycodestyle.

@nmnobre You will need to wait for 2507_inline_trans_ignore_arg_clash that remove current pycodestyle/fake8 issues in master

sergisiso:
Oh you were quick on fixing them! no need to revert now :)

nmnobre:
> But we already check it because flake is a superset of pycodestyle, but we have these "--select=E9,F63,F7,F82" that let issues reach master, we just need to remove this flag instead of adding pycodestyle.

Yes, it is indeed a superset, but removing the filter as you suggest causes a myriad of other warnings/errors to show...
So I thought I'd just call pycodestyle independently.

sergisiso:
Ok, I also realise that other PR are blocked until this is on master due to the pep8 issue affecting everything

sergisiso (approved):
The integration tests on GlaDos are green and the linkcheck fail is from mybinder which at the moment is down and does not need to be fixed. This PR is ready to merge

sergisiso:
I also create #2513 for further looking into the linters

PR #2509, opened 2024-02-22
(Closes #2394) add support char logical in psydata

### LINKED ISSUE #2394: PSyData missing support for logical and character types
At the least the NAN test does not support logical and character types. While long term we might want to update the transformations to not supply any non-real argument in the first place (which should now be possible, since I believe we have the required type information), a short terms solution is to just add these functions (as empty) to the generic interfaces.

hiker:


lib/read_only/Makefile
```diff
@@ -50,8 +50,8 @@ PSYDATA_LIB_DIR ?= ./..
 
 # The read-only verification library is implemented for int, real and
```
arporter, line 51: Please update the comment too.

lib/value_range_check/Makefile
```diff
@@ -50,8 +50,8 @@ PSYDATA_LIB_DIR ?= ./..
 
 # The nan-test library is implemented for int, real and
 # double scalars and 2-dimension arrays
```
arporter, line 52: Please update the comment too.

lib/value_range_check/value_range_check_base.jinja
```diff
@@ -33,9 +33,11 @@
    when mould is larger than value.) #}
 
 {% if ALL_TYPES is not defined %}
-   {% set ALL_TYPES = [ ("Double", "real(kind=real64)",   64),
-                        ("Real",   "real(kind=real32)",   32),
-                        ("Int",    "integer(kind=int32)", 32) ] %}
+   {% set ALL_TYPES = [ ("Double",  "real(kind=real64)",   64),
+                        ("Real",    "real(kind=real32)",   32),
+                        ("Char",    "character",            8),
+                        ("Logical", "logical",              8),
```
arporter, line 39: I happened to be reading about this today and, in Fortran, the storage of a logical has to be the same as the storage of a default integer. I don't know what it is for character. I can't remember how this 'number of bits' is used here though?

lib/value_range_check/value_range_check_base.jinja
```diff
@@ -431,7 +433,7 @@ contains
         {{type}}, dimension({{DIMENSION}}), intent(in) :: value
 
         {# IEEE_SUPPORT_DATATYPE does not even compile for int data types #}
```
arporter, line 435: Comment needs updating?

arporter (changes requested):
Nice, small PR :-)
Would it be possible to extend one of the examples to make use of/exercise this new functionality? Otherwise it's completely untested :-(
Also, I have a concern about the 'number of bits' being specified for logical and character variables. I can't remember how you use this information?

lib/read_only/Makefile
hiker: Done.

lib/value_range_check/Makefile
hiker: Done.

lib/value_range_check/value_range_check_base.jinja
hiker: OK, that cost me some time. Turns out, the bit size was only used for the read-only-verification, so I decided to remove these unnecessary value from the other files.
But since it was (potentially/likely) also set when calling `process.py`, so I changed read-only to  set the bit size depending on name instead. That made this patch a bit bigger, but these unnecessary values are gone now.

hiker:
> Nice, small PR :-) Would it be possible to extend one of the examples to make use of/exercise this new functionality? Otherwise it's completely untested :-( 

That was a bit of a headache, since neither gocean nor lfric support character strings. So I've added a new generic read-only verification library, and added an example as nemo/eg6. I couldn't add an examples that will actually throw an error (since PSyclone processes the file and does not support `loc` or `sizeof`, which was the trick I used in gocean). There is documentation how you can manually modify the psy file to trigger an error (but admittedly that's kind of stupid). Suggestions welcome.

> Also, I have a concern about the 'number of bits' being specified for logical and character variables. I can't remember how you use this information?

I have also refactored all of that, so no more bits specified. Unfortunately, all of that that made this PR somewhat bigger :)

doc/user_guide/psy_data.rst
```diff
@@ -230,6 +231,25 @@ An executable example for using the GOcean read-only-verification
...
+Compilation of the library is done by invoking ``make`` and setting
+the required variables:
+
+.. code-block:: shell
+
+    make F90=ifort F90FLAGS="--some-flag"
+
+This will create a library called ``lib_read_only.a``.
+An executable example for using the generic read-only-verification
+library is included in ``examples/gocean/eg6/``.
```
arporter, line 251: nemo/eg6 rather than gocean?

examples/nemo/eg6/Makefile
```diff
@@ -0,0 +1,84 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# ------------------------------------------------------------------------------
+# Author: A. R. Porter, STFC Daresbury Lab
+# Modified J. Henrichs, Bureau of Meteorology
+
+# Makefile for the 5th NEMO example. Uses PSyclone (which must
```
arporter, line 37: First sentence is wrong (and could just be removed).

examples/nemo/eg6/README.md
```diff
@@ -0,0 +1,98 @@
+# PSyclone NEMO Example 6 - Read-only Verification
```
arporter, line 1: Please add a brief section to examples/nemo/README.md to mention this new example.

doc/user_guide/tutorials_and_examples.rst
```diff
@@ -783,6 +783,14 @@ compilation instructions are in the ``README.md`` file, including how
 to switch from using the stand-alone extraction library to the NetCDF-based
 one (see :ref:`extraction_libraries` for details).
 
+Example 6: Read-only Verification
```
arporter, line 786: I realise we have duplication between this text and the markdown README files. If you're feeling keen, we could experiment with removing this section and replacing it with an include of the appropriate README. Sphinx's `include` directive supports a 'parser' option which apparently allows this (https://docutils.sourceforge.io/docs/ref/rst/directives.html#include-options). However, now that I write all that, I think perhaps that would be a separate issue.

examples/nemo/eg6/README.md
```diff
@@ -0,0 +1,98 @@
...
+ PSyData: checked variable char_var
+ PSyData: checked variable logical_var
+ PSyData: PostEnd dummy r0
+   3.00000000     F
+```
+
+If you copy the lines 32 and 33 from ``dummy.f90`` into ``psy.f90``,
+the code will modify ``logical_var`` (by using out-of-bound array accesses.
+Or you could just manually set ``logical_var = .true.``). If you then
+compile again, an error will be produced.
```
arporter, line 45: This worked for me. Good stuff. Only wrinkle is that doing `make compile` again ran PSyclone and over-wrote my modified psy.f90 file. I had to cut-n-paste the compile and link lines to build with the modified psy.f90 file. Not a biggie. Either mention it here or perhaps it's just a small tweak to the Makefile?

examples/nemo/eg6/dummy.f90
```diff
@@ -0,0 +1,35 @@
...
+         else
+            umask(ji,jj) = 3
+         endif
+       end do
+
+    ! VERY NAUGHTY CODE AHEAD - there be dragon!!
+    ! We modify the values of some read-only fields by using offsets
+    ! in a written array. This will abort if the code is compiled
+    ! with array bound check of course.
+    ! TODO: Unfortunately, PSyclone does not support loc or sizeof,
```
arporter, line 25: Please could you clarify here. I tried it for myself and see that it's the generation of the PSyData calls that is broken. If we don't already have an Issue please could you create one and refer to it here. Generating code that doesn't compile isn't a good look for us :-(

examples/nemo/eg6/dummy.f90
```diff
@@ -0,0 +1,35 @@
+program dummy
```
arporter, line 1: Please add the ususal license and a brief comment summarising the purpose of this file.

lib/read_only/generic/Makefile
```diff
@@ -0,0 +1,102 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: J. Henrichs, Australian Bureau of Meteorology
+
+
```
arporter, line 36: Please add a comment saying that this Makefile is for the read-only checking lib for generic Fortran code.

lib/read_only/generic/README.md
```diff
@@ -0,0 +1,90 @@
+# Read-only Verification Library for Generic Fortran Code
+
+This library implements the [PSyData API](
+https://psyclone.readthedocs.io/en/latest/psy_data.html#read-only-verification-library-for-gocean)
```
arporter, line 4: This link is for gocean. Is that deliberate?

lib/read_only/generic/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+The locations of the ReadOnly and PSyData base classes are specified
+using the environment variables ``$LIB_TMPLT_DIR`` and ``$PSYDATA_LIB_DIR``,
+respectively. They default to the relative paths to the
+[``lib/read_only``](./../) and top-level [``lib``](./../../) directories.
+
+The compilation process will create the wrapper library ``lib_read_only.a``.
+
+Similar to compilation of the [examples](
+https://psyclone.readthedocs.io/en/latest/examples.html#compilation), the
```
arporter, line 39: Should be `tutorials_and_examples.html` I think.

lib/value_range_check/value_range_check_base.jinja
arporter: Sorry. That was quite a lot of work. I think it's an improvement though.

lib/value_range_check/value_range_check_base.jinja
arporter: :-)

examples/nemo/eg6/dummy.f90
arporter: You could take the opportunity to say this is really just about exercising aspects of the read-only verification library.

arporter (changes requested):
Thanks very much Joerg. Just a few minor things now.
In testing the compilation of the examples I discovered that the xdsl Makefile was broken (spaces instead of tabs) so I've taken the liberty of fixing that and pushed it on your branch.

examples/nemo/eg6/README.md
hiker: So many places to modify :)  Done, thanks for spotting that.

examples/nemo/eg6/Makefile
hiker: Done, and I slightly reworded the next sentence to read a bit better.

doc/user_guide/psy_data.rst
hiker: Indeed.

doc/user_guide/tutorials_and_examples.rst
hiker: I have it nearly all sorted out, then I realised that the rst has a link target from the dev guide, which would now need to be defined in the README.md :cry: ... I kind of ran out of time to see if this is possible, and opened #2847 with the patch I have done so far.

examples/nemo/eg6/README.md
hiker: Good point. I think I've fixed that now.

examples/nemo/eg6/dummy.f90
hiker: Not really a PSyclone problem (at this stage at least :) ). The two functions `loc` and `sizeof` are non-standard extensions, and as such are not detected by fparser (and then end up as being assumed to be arrays).  We have an fparser issue to support additional (standard) intrinsics:
https://github.com/stfc/fparser/issues/434

So I've added these two in with the comment that they are non-standard. I've also updated the comment to better indicate why it doesn't work.

If you prefer, I could handle `loc` and `sizeof` as special case in the transformation, so that they wouldn't get 'extracted' ... but of course that would mean if a user has those names used as variables, it would fail as well :(

examples/nemo/eg6/dummy.f90
hiker: Done

lib/read_only/generic/Makefile
hiker: Done.

lib/read_only/generic/README.md
hiker: Nope - good catch, thanks!

lib/read_only/generic/README.md
hiker: Indeed - that's broken in all three versions of the library  - oops, it's actually broken in all READMEs in the lib directory. All fixed now.

hiker:
OK, all issues addressed I believe.

hiker:
Linkspector failures are caused by the new example not yet in git (master). I've started the compilation tests.

doc/user_guide/psy_data.rst
hiker: I meant ... you are right, it was indeed wrong.

doc/user_guide/tutorials_and_examples.rst
arporter: Thanks Joerg :-)

examples/nemo/eg6/dummy.f90
arporter: No, that's good now, thanks.

arporter (approved):
All good now thanks Joerg.
I'll proceed to merge.

PR #2516, opened 2024-03-05
Fix all issues flagged by flake8

### LINKED ISSUE #2513: Resolve flake8 and pycodestyle duplication
Currently we execute both in the CI because the first has some error exclusions. Ideally we want to resolve the exclusions, just run one of the linters (flake8 is a superset of pycodestyle), and update the wiki/developer guide to mention the exact steps we expect.

nmnobre:
Closes #2513.

Could the reviewers please be particularly diligent regarding:

~1) 9fabecd, which might break doxygen, how can I check this?~
2) ac7bb26, which differentiates between tutorial exercises for which the unused imports are needed to solve the exercise (ignored in `setup.cfg`) and those for which the imports are indeed unused (just removed the imports).


sergisiso (approved):
@nmnobre The integration test all pass with expected results. The changes all look good and the imports left in the tutorials are necessary. This is ready to merge.

(I removed one noqa because the entire statement could be removed. I will also update the github wiki page after merging the PR)

PR #2521, opened 2024-03-19
Fix line-length example

### LINKED ISSUE #2520: line_length example is failing
The CI test uses `make`, which works, but the instructions also state that you should be able to run `runme.py`, but that doesn't work anymore:

```
/psyclone/examples/line_length$ python ./runme.py 
  File "/home/joerg/work/psyclone/examples/line_length/./runme.py", line 7, in <module>
    ALG, PSY = generate(
TypeError: generate() got an unexpected keyword argument 'kernel_path'
```

hiker:
Fixes #2520  by removing the kernels directive.

### DIFF
--- a/changelog
+++ b/changelog
@@ -29,6 +29,8 @@
 	11) PR #2511 for #2510. Array2Loop transformations avoid character arrays
 	unless requested in the transformation options.
 
+	12) PR #2521 for #2520. Fix issue with line_length runme example.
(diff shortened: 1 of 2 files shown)


sergisiso (approved):
@hiker I agree in removing kernel_path from this script as the kernel is looking for is already in the same directory.

PR #2535, opened 2024-04-04
Otter profiling code for #1774

LonelyCat124:


LonelyCat124:
Waiting on #2537 and #2538 to continue this.

LonelyCat124:
Preprocessor support added so code generation can now work correctly.

I want to force some names to remain fixed, and I want to have a unit test for the OmpSerialRegion as well.

LonelyCat124:
@arporter I think this is ready for a first look - there are some statement orderings that are slightly different to what I did by hand (when the code adds taskwaits in itself) but they don't change the functionality so I have no issue leaving them as is.

LonelyCat124:
We don't need this in a release for the end of the tasking project - so there's no reason to rush completion of this PR. I think the logic is all here for this and we can decide how we want to fit this into PSyclone (perhaps as Joerg suggests in the examples, though I think we still end up with some extensions of the transformation to be able to do this correctly inside the example script)

arporter:
> We don't need this in a release for the end of the tasking project - so there's no reason to rush completion of this PR. I think the logic is all here for this and we can decide how we want to fit this into PSyclone (perhaps as Joerg suggests in the examples, though I think we still end up with some extensions of the transformation to be able to do this correctly inside the example script)

Great. I think we go for an example then.

arporter:
Since this modifies core PSyIR code for a tool (Otter) that is currently experimental/in development, we've agreed that this PR is not the way to go. However, we'll keep the branch in case it becomes useful in future.

Issue #1057, opened 2021-01-18
Symbols should have a generic method to check if they match with a set of symbol properties.


sergisiso:
#1042 adds to the symbol table the ability to provide specialized parameter arguments to generic methods. For example:

```
symbol_table.symbol_from_tag("tag", DataSymbol, constant_value=3)
```

Currently, if the symbol is already found in the SymbolTable the additional arguments (in the previous case `constant_value=3`) are ignored. But it would be better if this additional arguments are checked if they match on the symbol found.

We can implement this check polymorphically inside each symbol (because it knows about its own properties):
```
symbol.match_properties(**keyword_arguments)
```

This may also be beneficial for broader symbol comparisons.

PR #2540, opened 2024-04-05
(Closes #2537) Added exact_name requirement to the symbol_table find_and_create_tag …

### LINKED ISSUE #2537: only_exact_name parameter for find_or_create_tag on SymbolTable
When creating variables which need a tag (to find later) and we need to guarantee that the created symbol will have the correct name (e.g. if they are an import from a module) an `only_exact_name` parameter would be helpful.

If this cannot be met I think we just throw an exception and fail for now.

LonelyCat124:
…function


LonelyCat124:
First draft of this also ready for review @sergisiso - the only thing I maybe think we should add is if exact_name is requested and we find a tagged variable, should it have the same name also?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -427,6 +428,8 @@ def find_or_create_tag(self, tag, root_name=None, **new_symbol_args):
         :param str tag: tag identifier.
         :param str root_name: optional name of the new symbol if it needs \
                               to be created. Otherwise it is ignored.
+        :param bool exact_name: whether to disallow renaming of the
+                                     created symbol.
```
sergisiso, line 432: When moving it to new_symbol, I would prefer it to be "allow_renaming" with the docstring "whether the allow the generated symbol name to be different that the provided root_name. Defaults to True."

Tbh, it would also be useful to leave the docstring descriptions here together with the new_symbol "shadowing" and "symbol_type", otherwise it's difficult to know that they exist.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -454,6 +460,15 @@ def find_or_create_tag(self, tag, root_name=None, **new_symbol_args):
         except KeyError:
             if not root_name:
                 root_name = tag
+            if exact_name:
+                symbols = self.get_symbols()
+                existing_names = set(symbols.keys())
+                if root_name in existing_names:
```
sergisiso, line 466: Maybe in a one-liner "if root_name in self.get_symbols().keys()" will be clearer.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -454,6 +460,15 @@ def find_or_create_tag(self, tag, root_name=None, **new_symbol_args):
...
+                    raise SymbolError(
+                        f"Attempted to create a symbol with name "
+                        f"'{root_name}' but a symbol with that name already "
+                        f"exists, and using the exact name was required."
```
sergisiso, line 470: Maybe "Can not create symbol '{root_name}' without renaming it because it already exists in this scope"

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -454,6 +460,15 @@ def find_or_create_tag(self, tag, root_name=None, **new_symbol_args):
         except KeyError:
             if not root_name:
                 root_name = tag
+            if exact_name:
```
sergisiso, line 463: In new_symbol,  can you also add the check (or import-renaming logic) that I mentioned in https://github.com/stfc/PSyclone/issues/2537#issuecomment-2042818518

sergisiso: Also make it fail for UnsupportedType which should not support renaming.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -2025,6 +2025,17 @@ def test_find_or_create_tag():
...
+                                     exact_name=False)
+    assert tag5.name == "var_1"
+    # Check that it fails if exact_name was specified but we couldn't create
+    # a variable with that name.
+    with pytest.raises(symbols.SymbolError) as err:
+        symtab.find_or_create_tag("tag6", root_name="var",
+                                  exact_name=True)
+    assert ("Attempted to create a symbol with name 'var' but a "
+            "symbol with that name already exists, and using the exact "
+            "name was required." in str(err.value))
```
sergisiso, line 2038: This should work the same after moving to new_symbol (maybe with a different error message) so you should keep it here. But also add the same test for new_symbol.

sergisiso (changes requested):
@LonelyCat124 I think this functionality could be moved one level down the stack because it is a more generic problem. Could you try moving it see if everything still works? Also you could add checks to rename ImportInterface using orig_name and prevent UnsupportedTypes (see issue comment). 

> [...] if exact_name is requested and we find a tagged variable, should it have the same name also?

Yes that would make sense, if fact I think we should check that all provided attributes match but this is part of #1057

src/psyclone/tests/psyir/symbols/symbol_table_test.py
LonelyCat124: Added some new tests for the changes in new_symbol.

LonelyCat124:
@sergisiso Adding something to disallow renaming of UnsupportedFortranTypes has broken how common blocks are created by the fparser2 frontend - is this something I should/can workaround or should we remove that restriction for now?

LonelyCat124:
@sergisiso Ready for a nother look, should be a lot closer to a real implementation now

sergisiso (changes requested):
Almost there now @LonelyCat124 , you just need to change the location of one of the added asserts.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -340,30 +340,36 @@ def _has_same_name(cls, first, second):
...
         By default it creates a generic symbol but a symbol_type argument
         and any additional initialization keyword arguments of this
         symbol_type can be provided to refine the created Symbol.
 
-        :param root_name: optional name to use when creating a new \
-            symbol name. This will be appended with an integer if the name \
+        :param root_name: optional name to use when creating a new
+            symbol name. This will be appended with an integer if the name
             clashes with an existing symbol name.
         :type root_name: str or NoneType
```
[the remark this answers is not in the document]
LonelyCat124: Done.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -2050,6 +2074,17 @@ def test_rename_symbol():
...
+    # and disallow renaming.
+    with pytest.raises(symbols.SymbolError) as err:
+        schedule_symbol_table.new_symbol("array",
+                                         symbol_type=symbols.DataSymbol,
+                                         datatype=array_type,
+                                         allow_renaming=False)
+    assert ("Cannot create symbol 'array' as a symbol with that name already "
+            "exists in this scope, and renaming is disallowed."
+            in str(err.value))
+
```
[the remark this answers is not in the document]
LonelyCat124: Added a new test for this code.

LonelyCat124:
@sergisiso Should be ready for a (hopefully) final look.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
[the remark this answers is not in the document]
sergisiso: The old test can be removed, but I will do it myself before merging.

sergisiso (approved):
@LonelyCat124 All comments have been addressed, the integration tests passed, the code coverage is complete, and all new functionality is int he docstrings. This is approved for merging.

PR #2545, opened 2024-04-10
#2544 Accept -c as shortcut for --config on command line.

### LINKED ISSUE #2544: Accept -c on command line (as short form of --config)
It is unintuitive that there is no shortcut for `--config` (which just caused some confusion).

It would also be nice if any errors in the command line would be reported when `-v` is specified ... `-v` was used as verbose, and while it then printed the version number, it did not report an error for the invalid `-c` command line. A quick look at the argparse manual seems to indicate that this is not possible, argparse exits after printing the version number. But if someone knows more, please let me know.

hiker:
Trivial change, fixes #2544


hiker:
I've also added `-c` for kernel_tools, update the user guide (no --config in dev guide), and fixed some other documentation issues (warning during build, duplicated arguments, tabs instead of spaces).

hiker:
Ready for review (ignoring the missing codecov). And checking with the ticket, the problem was that there was no error message for an incorrect `-c` when the user added `-v` (assuming that was triggering verbosity). So not an issue with the missing file argument (as I originally stated), and I think it's accepted behaviour that when the version is requested, no checking of arguments is done.

sergisiso (approved):
Thanks @hiker, I agree with the change as -c being short for --config is expected behaviour. This PR is approved for merging.

Issue #2499, opened 2024-02-08
Scalarization transformation


LonelyCat124:
In parts of the physics codes for LFRiC we come across loop patterns such as this:
```
do i = ...
  do l = ...
    temp_in(l) = array(l,i) * array2(l,i)
  end do
  call exp_v(n, temp, temp_in)
  do l = ...
   !do some based on temp(l)
  end do
end do
```

Once we inline and fuse this loop structure, we get loops like this:
```
do i = ...
  do l = ...
    temp_in(l) = array(l,i) * array2(l,i)
    temp(l) = exp(temp_in(l))
   !do some based on temp(l)
  end do
end do
```

For cases such as this, `temp_in` and `temp` can be scalarised providing that nothing outside the loop depends on their values (which would already be a strange implementation choice, since it would only be for the final value of `i`). This would help us remove some false dependencies, as there is a write-write dependency on `temp(l)` if we use `collapse` on this loop, however these are not necessary since `temp` can just be a local scalar instead.

The goal of this transformation would be to take code like the above (post all the other inline and loop fusion transformations) and generate:
```
do i = ...
  do l = ...
   temp_in_scalar = array(l,i) * arary2(l,i)
   temp_scalar = exp(temp_in_scalar)
   !do something based on temp_scalar
  end do
end do
```

At this point, we can apply `target` + `loop` with `collapse` which will lead to less kernel launches and synchronization, and probably better performance on GPU.

LonelyCat124:
First step - on Reference: Find next (and previous?) Reference to this symbol.

LonelyCat124:
> > First step - on Reference: Find next (and previous?) Reference to this symbol.
> 
> The VariableAccess information already contains all accesses in order.

Yeah - the plan is to use that data with this transformation.

LonelyCat124:
@hiker I'm a bit confused about how to use the VariablesAccessInfo - is there a linkage between VariablesAccessInfo and the node for a given read/write access? E.g. For a given routine if I wanted to find (in order) all the accesses/dependencies on a given symbol (slash signature) can I do that with the VariablesAccessInfo? I can find the sequence of reads/writes but if I wanted to refer back to the relevant `Reference` is that currently possible?

hiker:
[...]

PR #2553, opened 2024-04-19
(Towards #2499) Initial implementation of next_access function on Reference

LonelyCat124:
@sergisiso I have an initial implementation of `next_access` using the VariablsAccessInfo that we expect to be useful for GPU-based pull requests. I think its ready for a first look.

If this looks ok to you I'll also implement `previous_access` in a similar way. I was not sure if the `all_access` function (or in general the members from VariablesAcessInfo) were necessarily in order, hence the loop at Reference.py:219 . I also feel like using next_access a lot on a large Routine could be computationally expensive, and wonder if we can cache VariablesAccessInfo and clear this cache in `_update_node` (or update_signal? but we need to ensure super() is called if other nodes implement either of these functions).


LonelyCat124:
I guess one thing I'm not sure on is whether the `location` of a VariablesAccessInfo is related to its position in the psyir tree? I think there's no guarantee (especially if you use things like `.merge`) so perhaps I need to use node.abs_position in my checks, I was just hoping to avoid that if possible.
Again - if we need to use abs_position It would be nice to cache this, but I think its less feasible to do so (since adding a node somewhere in the tree needs to invalidate the cache'd abs_position of every node in the tree that appears "after" the change, whilst being able to ignore all other sections of the tree.

hiker:
> I guess one thing I'm not sure on is whether the `location` of a VariablesAccessInfo is related to its position in the psyir tree? 

No, not directly. The position is basically a sequential number of statements. So the left-most statement in a tree should have position 1 etc.

I had a very brief look - I think you can't really avoid getting the accesses in the whole subroutine (unless you want to walk the tree explicitly). And yes, I think you then need to look up the position of the reference you have in the list of all accesses (that might be a nice method in the SingleVarAccesses ... `get_index_of` or so). Once you have the index `ind`, then `ind+1` in that list would immediately give you the next access (including the reference) ... excluding conditionals :(

> I think there's no guarantee (especially if you use things like `.merge`) so perhaps I need to use node.abs_position in my checks, I was just hoping to avoid that if possible. Again - if we need to use abs_position It would be nice to cache this, but I think its less feasible to do so (since adding a node somewhere in the tree needs to invalidate the cache'd abs_position of every node in the tree that appears "after" the change, whilst being able to ignore all other sections of the tree.

I did a lot of caching ... and removed it mostly again. For what I did, there was no measurable speed difference, so re-computing means that it will work even if the tree gets modified in between.

LonelyCat124:
> Once you have the index ind, then ind+1 in that list would immediately give you the next access (including the reference) ... excluding conditionals :(

How do conditionals affect it? We probably need to find the next one including conditionals (which the current implementation finds but if there's an easier way that would be good.

hiker:
> > Once you have the index ind, then ind+1 in that list would immediately give you the next access (including the reference) ... excluding conditionals :(
> 
> How do conditionals affect it? We probably need to find the next one including conditionals (which the current implementation finds but if there's an easier way that would be good.

Oops - bad wording. I meant VariableAccesses do not handle conditionals, so the 'next' reference found could be inside a conditional that is not ever executed. But yes, References in conditionals will certainly be found.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
...
+        :returns: the previous reference to the symbol accessed by this
+                  Reference.
+        :rtype: :py:class:`psyclone.psyir.nodes.Node` or None
+        '''
+        # Avoid circular import
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.routine import Routine
+        # The scope is as far as the Routine that contains this
+        # Reference.
+        var_access = VariablesAccessInfo(nodes=self.ancestor(Routine))
```
sergisiso, line 207: I would like this be generic to any PSyIR subtree, even when not connected to a Routine. So, I agree that in general we should look for the ancestor(Routine), but when this is not found (e.g we have a standalone statement `a = b + a + b` ) we could start from the expression root.

Also add a test with a disconnected statement, as currently they are all full source trees.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
...
+        signature, indices = self.get_signature_and_indices()
```
sergisiso, line 208: `indices` is unused, replace with `_`

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
...
+        single_accesses_to_this = var_access[signature]
+        all_accesses = single_accesses_to_this.all_accesses
```
sergisiso, line 210: `single_access_to_this` is a confusing name (because it stores all accesses to this symbol), I would skip it altogether and just do:
`all_accesses = var_access[signature].all_accesses`

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
...
+        index = sys.maxsize
```
sergisiso, line 211: You can do `index = -1` here and `index = len(all_accesses)` in the next one, and we don't need to import sys.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
             return UnresolvedType()
         return self.symbol.datatype
 
+    def previous_access(self):
+        '''
+        :returns: the previous reference to the symbol accessed by this
+                  Reference.
```
sergisiso, line 199: maybe simpler: `:returns: the previous reference to the same symbol.`

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
...
+        :rtype: :py:class:`psyclone.psyir.nodes.Node` or None
```
sergisiso, line 200: Use ``Optional[:py:class:`psyclone.psyir.nodes.Node`]``

src/psyclone/psyir/nodes/reference.py
```diff
@@ -191,6 +193,57 @@ def datatype(self):
...
+                index = i
+                break
+
+        if index > 0:
+            return all_accesses[index-1].node
+        return None
+
+    def next_access(self):
+        '''
+        :returns: the next reference to the symbol accessed by this Reference.
```
sergisiso, line 224: Same considerations as in previous method, e.g. `:returns: the next reference to the same symbol.`, ...

src/psyclone/tests/psyir/nodes/reference_test.py
```diff
@@ -214,3 +214,221 @@ def test_reference_can_be_copied():
...
+
+    # Check if statements don't affect
+    code = '''subroutine my_sub()
+   integer :: a
+   logical :: b = .false.
+   a = 1
+   if(b) then
+     a = 0
+   end if
+   a = 2
```
sergisiso, line 299: I am curious about what happens with an else block:
```
a = 1
if b then
  a = 2
else
  a = 3
endif
a = 4
```

Does next_access of a(2) points to a(3) or a(4)?

I don't know what is the right behaviour, we can refine that on upcoming PRs that actually use this functionality, but it would be good for now to have an assert here to demonstrate what it currently does.

src/psyclone/tests/psyir/nodes/reference_test.py
```diff
@@ -214,3 +214,221 @@ def test_reference_can_be_copied():
...
+    assert a_3.previous_access() is a_2
+
+
+def test_reference_previous_access_with_codeblock(fortran_reader):
+    code = '''subroutine my_sub()
+    character, dimension(100) :: a
+    write(a, "A") "mytest"
+    a = "test"
+    end subroutine
+    '''
```
sergisiso, line 427: Could you also test what happens with initialisation expressions, e.g.:
```
module my_mod
   use external_mod, only: a
   contains
   subroutine my_sub()
      integer :: b = a + a +1
      b = a + 3
   end subroutine my_sub
end module my_mod
```
I expect that the 'a' in the statement won't have a previous_access to the initialisation expression (because that's not part of the tree, which is the right behaviour unless later we find the contrary is needed in the transformations)

And if you relaxed the `ancestor(Routine)` you can do next/previous_access from the two references in the initialisation expression, but not jump into the body of the subroutine. Which I also think is the right behaviour for now.

sergisiso (changes requested):
Thanks @LonelyCat124, this will be useful for upcoming transformations. I just made minor comments for the implementation and requested to extend some tests to document the more unclear behaviour.

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Added the code to do it, will work on a test as well.

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Changed.

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Changed

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Fixed - I think this is mostly a leftover from an old implementation so good fix.

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Done

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Done

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Should be changed.

src/psyclone/tests/psyir/nodes/reference_test.py
LonelyCat124: It will point to a = 3.
I'm not sure I agree with that behaviour but I'm also not sure whether we should change it here not - perhaps more of a discussion when we create thing using this code.

src/psyclone/tests/psyir/nodes/reference_test.py
LonelyCat124: It doesn't find in the initialisation expression.
I'll try to do add the same for the initialisation expression if I can work it out.

LonelyCat124: @sergisiso Is this test sufficient for the non-routine access case or do you still want another standalone case?

LonelyCat124:
@sergisiso should be ready for another look.

sergisiso (approved):
@LonelyCat124 All looks good now (but a couple of small things that I fixed myself), we can refine the specific behaviour when we use it. This is ready to merge.

PR #2563, opened 2024-04-30
(Closes #2499) scalarization transformation implementation

### LINKED ISSUE #2499: Scalarization transformation

LonelyCat124:
First implementation is here. I have unit tests for the private routines, I need to test it with code examples doing the full transformation properly but in theory its implemented here for now.


LonelyCat124:
@sergisiso this is ready for a first look.
I didn't add anything to user or dev guide for now, let me know if you think I should.

I'm waiting for hiker to clarify my question before I remove the code at lines 123 in the transformation, but I expect to remove those lines as I think they're unreachable.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,212 @@
...
+from psyclone.psyGen import Kern
+from psyclone.psyir.nodes import Assignment, Call, CodeBlock, IfBlock, \
+        Reference, Routine
+from psyclone.psyir.symbols import DataSymbol
+from psyclone.psyir.transformations.loop_trans import LoopTrans
+
+
+class ScalarizationTrans(LoopTrans):
+
+    def _find_potential_scalarizable_array_symbols(self, node, var_accesses):
```
sergisiso, line 50: I found this method hard to understand until I read the apply method. I think its name is too generic and it does too many things, so I suggest breaking it up (see name suggestions below).

I also notice that they all iterate over potential var_accesses and select/filter the ones passing the conditions. Could we use the python filter functionality? If the apply would be something like this:
```
var_accesses = VariablesAccessInfo(nodes=node.loop_body)
var_accesses = filter(self._is_local_array, var_accesses)
var_accesses = filter(self._have_same_unmodified_index, var_accesses)
var_accesses = filter(self._first_access_is_write, var_accesses)
var_accesses = filter(self._not_used_after_loop, var_accesses)
```
It would be very readable (so much so that I think it doesn't need inline comments) and makes each method simpler (to output array, no iteration, ...). Also the methods don't use self or node (except the last one but see next comment), should they be static methods or even just inner functions of apply?

Could you explore if this would work.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,212 @@
...
+
+from psyclone.core import VariablesAccessInfo
+from psyclone.psyGen import Kern
+from psyclone.psyir.nodes import Assignment, Call, CodeBlock, IfBlock, \
+        Reference, Routine
+from psyclone.psyir.symbols import DataSymbol
+from psyclone.psyir.transformations.loop_trans import LoopTrans
+
+
+class ScalarizationTrans(LoopTrans):
```
sergisiso, line 48: When you add the docstring, can you bring what you wrote in apply here and add a simple example, e.g. like `src/psyclone/psyir/transformations/hoist_trans.py` . So then we just need to list the autoclass under the available transformations section of `doc/user_guide/transformations.rst`  and can be tested with `python -m doctest src/psyclone/psyir/transformations/scalarization_trans.py`. The apply docstring can be a oneliner then.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,212 @@
...
+
+        for signature in potentials:
+            # Find the last access of each signature
+            last_access = var_accesses[signature].all_accesses[-1].node
+            # Find the next access to this symbol
+            next_access = last_access.next_access()
+            # If we don't use this again then its valid
+            if next_access is None:
+                potential_arrays.append(signature)
+                continue
```
sergisiso, line 105: I was expecting the check for next accesses to only be this. Why do we need the rest, is it because the next_access implementation is currently missing these? E.g. if we use next_accesses in other places, e.g. when deciding where to place a halo_exchage to be safe, wouldn't we need the same logic as below? In this case shouldn't this all be inside the next_access implementation?

Also I see that next access only returns one reference, was this the right choice?, given that potentially is more than one:
```
a = 3
if condition:
    a = 1
else:
    a = 1
```
Also the `a` in the inner loop has 2 potential next_acesses
```
do i=1, 10
     a = 1
     do j 1=1, 10
          a = 2
     endo
enddo
a = 3
```

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,212 @@
...
+            ancestor_assign = next_access.ancestor(Assignment)
+            if (ancestor_assign is not None and
+                    ancestor_assign.lhs is not next_access):
+                continue
+
+            # If it has an ancestor that is a CodeBlock or Call or Kern
+            # then we can't guarantee anything, so we remove it.
+            if (next_access.ancestor((CodeBlock, Call, Kern))
+                    is not None):
+                continue
```
sergisiso, line 141: I find hard to see what each of these are checking (without going to the tests), could you add a short code snipet showing the dependency that each is looking for, inlcuding the commented one.

sergisiso (changes requested):
Good initial implementation @LonelyCat124, I am just asking if the code can be slightly refactored to be more readable and I am wondering if some logic should leave inside the next_access methods, I am happy to discuss this one.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I deliberately kept the next_access function simple and tried to avoid the complicated cases (and rely on the dependency analysis tooling) like this, and just assume these are ok.

I can try to make a new issue and PR to improve that functionality if you think its necessary/useful. I feel like there can always be an incredibly complex structure of nested loops that makes this difficult.

sergisiso: But wouldn't this complex logic be for everywhere that we want to prove if a symbol is going to be accessed next or not, not just for this transformation?

sergisiso: > I deliberately kept the next_access function simple and tried to avoid the complicated cases (and rely on the dependency analysis tooling)

I am not sure if it's next access or the VariablesAccessInfo tooling itself that needs to provide this.

@hiker and @JulienRemy I am trying to understand this. Is this related to the logic you needed on building the DAG? That the VariablesAccessInfo returns ALL following dependencies under a node (but not looking upwards - like we need in my loop snippet above) and the next_accesses just one is code order?

But what we need (and maybe next_access should be) is a method that return any "directly reachable" accesses to the same symbol, which could be more than one, and will be equivalent to the arrows leaving from a node in the DAG?

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Its doable I think, I've moved them to be static methods (I don't like inner functions), they do still need var_accesses (and node in the final functions case), so there is a lambda in the filter calls, but it seems to work. I will need to rewrite tests for this and maybe expand the apply testing.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I think this is a separate issue (though I'm not sure about the loop case, we might need to gracefully fail somehow for some case, I'll try to remember to put a test in for that).

We otherwise ignore the if case here (which could be improved if next_acccess gave all options, either through the access info or otherwise) and return False instead of trying to work it out.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: The commented one is removed because its not reachable with how the next_access/variable_access info gives results for assignments.

LonelyCat124:
@sergisiso I've addressed most of the comments here - I think the next_access one is a bigger issue than just here (and we need to decide if the variable access tooling should handle this or if we just do it in next_access, but I think we can do an implementation for this in the mean time anyway).

I need to still expand the tests for the apply routine and add documentation before another review though I think.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: @sergisiso I think I could probably do most of this for next access, butt I'm not sure on if there is an easy way to tell whether a node is inside an else if or an else or an if, for example for a case like (this is how we have the if/elseif behaviour in Psyclone)
```
a = 1
if cond1 then
   a = 2
else
  if cond2 then
    a = 3
    a = 6
  else
    if cond3 then
       a=4
end if 
end if
end if
a = 5
```

We currently would find the `a=2` case, and the proposal is that we need to also find the `a=3`, `a=4`, and `a=5` cases. The trick is knowing that the `a=3` and `a=4` is inside an else if case, so to continue trying to find more "next accesses" that aren't inside the same branch of the else/if structures. We also need different behaviour for `else if` vs `else` behaviour (and all cases are complicated by the "one branch doesn't contain a next_access")

sergisiso:
@LonelyCat124 I don't mind the order of these (if you put the associated TODOs), but wouldn't want to go very far with some logic that then will need to be deleted.

LonelyCat124:
I brought this up to master now and made the tests pass, but I'll start reworking the logic here now.

LonelyCat124:
Ok, I have an implementation of this now - I will try to add some more functionality tests and finish the docs next.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Added this documentation now

LonelyCat124:
@sergisiso Now we have the new functionality i think this is ready for a new look - I added more tests, updated the logic and added some more documentation.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
...
+          !  - Unsupported statement: Print_Stmt
+          PRINT *, b(i)
+        end if
+      enddo
+    <BLANKLINE>
+    end program test
+    '''
+
+    @staticmethod
+    def _is_local_array(signature, var_accesses):
```
sergisiso, line 97: Just to let you know that now we allow type hints e.g. 
`def _is_local_array(signature: Signature, var_accesses: VariablesAccessInfo) -> bool:`

instead of the ":type:" docstrings. Although not mandatory for now it is more readable.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
...
+        '''
+        :param signature: The signature to check if it is a local array symbol
+                          or not.
+        :type signature: :py:class:`psyclone.core.Signature`
+        :param var_accesses: The VariableAccessesInfo object containing
+                             signature.
+        :type var_accesses: :py:class:`psyclone.core.VariablesAccessInfo`
+        :returns bool: whether the symbol corresponding to signature is a
+                       local symbol or not.
```
sergisiso, line 106: "local array symbol"

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
...
+        :returns bool: whether all the array accesses to signature use the
+                       same index, and whether the index is unmodified in
+                       the code region.
+        '''
+        array_indices = None
+        scalarizable = True
+        for access in var_accesses[signature].all_accesses:
+            if array_indices is None:
+                array_indices = access.component_indices
+            # For some reason using == on the component_lists doesn't work
```
sergisiso, line 133: Is this comment outdated, we recently improved this, also the `[:]` shouldn't be needed (or is this what the comment is about?)

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2017-2024, Science and Technology Facilities Council.
```
sergisiso, line 4: Only "2024-2025" as date

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
```diff
@@ -0,0 +1,542 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2022-2024, Science and Technology Facilities Council.
```
sergisiso, line 4: Only "2024-2025" as date

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
```diff
@@ -0,0 +1,542 @@
...
+  enddo
+  do i = 1, 100, 1
+    if (c(i) > 50) then
+      b(i) = c(i)
+    else
+      b(i) = b(i) + c(i)
+    end if
+  enddo'''
+    out = fortran_writer(psyir)
+    print(out)
```
sergisiso, line 540: Remove leftover prints

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
...
+        :returns bool: whether the value computed in the loop containing
+                       sig is read from after the loop.
+        '''
+        # Find the last access of the signature
+        last_access = var_accesses[sig].all_accesses[-1].node
+        # Find the next accesses to this symbol
+        next_accesses = last_access.next_accesses()
+        for next_access in next_accesses:
+            # next_accesses looks backwards to the start of the loop,
+            # but we don't care about those accesses here.
```
sergisiso, line 183: If we don't have it yet, we could benefit from a "node.is_descendant_of(node2)" that returns a bool, and I would use a while .parent recursion with an "is" comparison instead of abs_position to implement it.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
...
+                return False
+
+            # If next access is in an IfBlock condition then it reads the
+            # value.
+            ancestor_ifblock = next_access.ancestor(IfBlock)
+            if ancestor_ifblock:
+                conditions = ancestor_ifblock.condition.walk(Node)
+                for node in conditions:
+                    if node is next_access:
+                        return False
```
sergisiso, line 199: I understand the "we don't care if it's in the same loop" but I don't understand why we need the rest instead of simply "return False"

When we have a next_accesses(), I thought that it only pointed to potential following reads, so for:
```
  a(i) = 1
  b(i) = 1
  c(i) = 1
enddo
d = a(i)  ! Guaranteed read
d = call(b(i)) ! Potential read
c(i) = d  ! Guaranteed not read
```
the 'a' and 'b' next_accesses() will point to the statements below, but 'c' wouldn't because it is written next. Is this not correct?

sergisiso (changes requested):
@LonelyCat124 This has improved a lot, bringing the dependency logic inside next_accesses and splitting each part into separate checks has made it much easier to follow it. I have some  more minor comments now

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: No - the next access finds the write as well so `c`.next_accesses() would point to the `c(i) = d` lhs.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: The comment is about the [:] - when I removed it this didn't work

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,324 @@
...
+    ... real :: a(100), b(100)
+    ... do i = 1,100
+    ...   a(i) = i
+    ...   b(i) = a(i) * a(i)
+    ... end do
+    ... do j = 1, 100
+    ...  if(b(i) > 200) then
+    ...    print *, b(i)
+    ...  end if
+    ... end do
```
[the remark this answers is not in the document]
LonelyCat124: I added the second loop to distinguish between the scalarisable and non-scalarisable variables. I'll try to work out the missing BLANKELINE - its not clear to me where thats missing.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I've added them alongside the type docstrings - I can remove them if you want but I wasn't sure if we were replacing or using alongside.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I tested it, if I don't have the `[:]` it still fails

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Fixed

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
LonelyCat124: Fixed

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
LonelyCat124: Removed, only one I found.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Ok - I'll add this as part of this PR if we don't have it.

LonelyCat124: I'm going to add this, but in principal we could not - `path_from` sort of does this (you can try/except on ValueError and ignore the result).

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: (So if we find a write we don't exit here, only if we find reads).

src/psyclone/psyir/transformations/scalarization_trans.py
sergisiso: Sorry for not clarifying it, we are replacing them, so you can delete the docstrings types unless is something currently not representable as typehint

src/psyclone/psyir/transformations/scalarization_trans.py
sergisiso: Ok, this is strange but not for this PR, thanks for checking

src/psyclone/psyir/transformations/scalarization_trans.py
sergisiso: I see the new method, can you use it here?

src/psyclone/psyir/transformations/scalarization_trans.py
sergisiso: Ok, but then all this method logic "_value_unused_after_loop" seems that will need to be implemented every time we check for this. E.g. in the "async/nowait clause" or in the "omp value can be private", so it seems a candidate to be put in a more "shared" location (it feels to me part of the DefinitionUseChain, maybe we can pass a parameter to `next_accesses(only_potential_reads=True)` and then `find_forward_accesses(only_potential_reads=True)`?)

Although I also think it could be the default behaviour, and we could do a `include_next_guaranteed_write` optional argument instead

What do you think?

If you prefer, we can delay this to a TODO and follow up PR.

src/psyclone/psyir/nodes/node.py
```diff
@@ -1782,6 +1782,23 @@ def update_parent_symbol_table(self, new_parent):
...
 
+    def is_descendent_of(self, potential_ancestor) -> bool:
+        '''
+        Checks if this node is a descendant of the `potential_ancestor` node.
+
+        :param potential_ancestor: The Node to check whether its an ancestor
+                                   of self.
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+
+        :returns bool: whether potential_ancestor is an ancestor of this node.
```
sergisiso, line 1793: You can delete the "bool" from here.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,334 @@
...
+
+from psyclone.core import VariablesAccessInfo, Signature
+from psyclone.psyGen import Kern
+from psyclone.psyir.nodes import Assignment, Call, CodeBlock, IfBlock, \
+        Loop, Node, Reference, Routine, WhileLoop
+from psyclone.psyir.symbols import DataSymbol
+from psyclone.psyir.transformations.loop_trans import LoopTrans
+
+
+class ScalarizationTrans(LoopTrans):
```
sergisiso, line 49: Remember to list it somwhere in `doc/user_guide/transformations.rst`

sergisiso (changes requested):
@LonelyCat124 See new comments and responses to previous review comments. There is only some documentation/typing work left, and I have some doubts about the location of some logic.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Docstring types removed where appropriate.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Yes - I've also added the loop as a parameter to `_value_unused_after_loop` to avoid computing it over and over.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I think this could be a standalone routine, but I'm not sure where it belongs and we should decide where to put it in part of a follow-up PR that also uses it, as its not clear to me where it should go (since I don't think it is part of DefinitionUseChain as its specific to Loops). It could go in Loop or Reference or some node, but its a bit messy there too.

I think that potentially this code is out of date  though - we added `is_read` and `is_write` to `Reference` previously, and I should probably just use that information instead. I will rewrite this to use that instead. I think my preference would be to leave next_accesses as is, and just make use of `is_write`/`is_read` to determine the status of the access (with the slight caveat that people have to check if they have had a non-Reference returned).

LonelyCat124: Ok - this code is now much simpler and mostly just uses the `is_read` functionality.

src/psyclone/psyir/nodes/node.py
LonelyCat124: Done

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I added it to the list of available transformatiosn - let me know if you think it should be somewhere more specific.

LonelyCat124:
@sergisiso Should be good for another look - there is one thing unsorted from the comments re: whether it should be a TODO or have a discussion next week and decide where to put it - I'll leave that to you to decide.

src/psyclone/psyir/transformations/scalarization_trans.py
sergisiso: Aha, that's way better, if it is down to these few lines then we don't need a shared place to put it (scrap my previous comment).

One more thing, instead of checking for `(CodeBlock, Call, Kern)` could we simplify it further doing:
```
   if not next_access.is_write():
       return False
```
And move the CodeBlock and Kern check inside the "is_write", so when we improve their functionality, we don't have to go fix each transformation? (The write already assumes worst case for argument calls - I would also add a "potentially writing" in the docstring)

(And for a separate PR but I don't see is_read/is_write used in DefinitionUseChain, could we benefit from using it there?)

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,282 @@
...
+        base_symbol = var_accesses[signature].all_accesses[0].node.symbol
+        if not base_symbol.is_automatic:
+            return False
+
+        return True
+
+    @staticmethod
+    def _have_same_unmodified_index(signature: Signature,
+                                    var_accesses: VariablesAccessInfo) \
+            -> bool:
```
sergisiso, line 120: This is pedantic but since I still have an open comment above, we try to avoid backslashes, could you do:
```
    def _have_same_unmodified_index(
            signature: Signature,
            var_accesses: VariablesAccessInfo
    ) -> bool:
```
And similar in the backslashes in signatures below.

sergisiso (changes requested):
It is really almost ready @LonelyCat124 , but I just want to ask a last question about the previous review remaining comment.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: Discussed in teams - this is not as simple as next_access may not be a reference, in whic case we can't call `is_write` and this check is to handle those cases.

LonelyCat124:
@sergisiso If I want to run this as part of integration tests (i.e. applying it to every loop in NEMO or something) do I just need to modify omp_gpu_trans.py or similar and then run the tests? If so I'l try to set that going over the weekend

LonelyCat124:
Ok - integration tests find a couple of issues.

NEMO 5 fails due to an error I'm not sure inside the dependency code, where something isn't created correctly. I'll need to check the processed file to fix that.

NEMO 4 fails due to a failure to resolve a MAXVAL signature, which I need to test myself. I'm a bit confused why we get a signature for MAXVAL but I'll investigate.

LonelyCat124:
Will run the integration tests again - only NEMO4 though for now.

LonelyCat124:
Remaining issues seem to be a bug in DefinitionUseChains. I'm going to try to work this out - once I find it am I ok to fix it here or shall i create an issue and separate pull request @sergisiso ?

LonelyCat124:
Probable bug: What happens if we find an element of a structure that we think can be scalarised (e.g. `mystruct%array(i)`) - we shouldn't scalarise? Or we scalarise only if we can work out the type of the element of the structure and then we need to come up with a name. @sergisiso what are your thoughts on how this should be handled?

LonelyCat124:
NEMO passthrough gives the wrong result - I need to investigate.

LonelyCat124:
NEMO4 now works, NEMO5 gets the wrong answer, I'll try to investigate next week

LonelyCat124:
First bug from NEMO5, that i'm not exactly sure how to resolve.

PSyclone scalarizes this first loop:
```
    if (MOD(kt, nn_trd) == 0 .OR. kt == nit000 .OR. kt == nitend) then
      do jk = 1, jpkm1, 1
        do jj = ntsj - 0, ntej + 0, 1
          do ji = ntsi - 0, ntei + 0, 1
            zkx(ji,jj,jk) = 0._wp
            zky(ji,jj,jk) = 0._wp
            zkz(ji,jj,jk) = 0._wp
            zkepe(ji,jj,jk) = 0._wp
          enddo
        enddo
      enddo
      call eos(ts(:,:,:,:,kmm), rhd, rhop)
      zkz(:,:,1) = 0._w
      do jk = 1, jpk, 1
        do jj = ntsj - 0, ntej + 0, 1
          do ji = ntsi - 0, ntei + 0, 1
            zkz(ji,jj,jk) = xcof * e1e2t(ji,jj) * ww(ji,jj,jk) * (rhop(ji,jj,jk) + rhop(ji,jj,jk - 1)) * tmask_i(ji,jj)
          enddo
        enddo
      enddo
```

It believes the write to `zkz(ji,jj,jk)` can be scalarized, because the next access (`zkz(:,:,1)`) overwrites it with 0. This means the values of `zkz(:,:,2:)` will be undefined in this example - the next access is also a write and maybe that is why but I need to check.

LonelyCat124:
Ok - replicating this seems to be difficult.
Part of this is because the next_accesses is finding the References inside the LHS object, since the ArrayReference here is not so straightfoward:
```
Assignment[]
    ArrayReference[name:'arr']
        Range[]
            IntrinsicCall[name='LBOUND']
                Reference[name:'LBOUND']
                Reference[name:'arr']
                Literal[value:'1', Scalar<INTEGER, UNDEFINED>]
            IntrinsicCall[name='UBOUND']
                Reference[name:'UBOUND']
                Reference[name:'arr']
                Literal[value:'1', Scalar<INTEGER, UNDEFINED>]
            Literal[value:'1', Scalar<INTEGER, UNDEFINED>]
        Range[]
            IntrinsicCall[name='LBOUND']
                Reference[name:'LBOUND']
                Reference[name:'arr']
                Literal[value:'2', Scalar<INTEGER, UNDEFINED>]
            IntrinsicCall[name='UBOUND']
                Reference[name:'UBOUND']
                Reference[name:'arr']
                Literal[value:'2', Scalar<INTEGER, UNDEFINED>]
            Literal[value:'1', Scalar<INTEGER, UNDEFINED>]
        Literal[value:'1', Scalar<INTEGER, UNDEFINED>]
    Literal[value:'0.0', Scalar<REAL, UNDEFINED>]
```

It finds the References to `arr` inside the IntrinsicCalls. Probably I need to change the DefinitionUseChains to be smarter about enquiry functions (I hate inquiry functions...), we should ignore inquiries I think. I think I should also change `Reference.is_read` to know about inquiry functions as well.
@sergisiso Are you ok for me to make these changes in this PR or would you prefer a separate PR?

LonelyCat124:
I have some fixes to handle these cases now, assuming everything passes for older PSyclone versions and coverage is ok i will run the integration tests again

LonelyCat124:
Ok - this is now waiting for NEMO integration tests to be fixed to run integration again. I think I've fixed the issues I found - I have some commented out code that I thought I needed but unless there are issues in NEMO integration tests still they will be removed.

LonelyCat124:
@sergisiso I think this is ready for a new look now. The integration tests all pass - I will revert the changes to the transformation scripts once we're close to this being accepted.

.github/workflows/nemo_tests.yml
```diff
@@ -96,6 +96,13 @@ jobs:
         module load perl/${PERL_VERSION}
         make -j 4 passthrough
         make -j 4 compile-passthrough
+        make run-passthrough |& tee output.txt
```
sergisiso, line 99: @LonelyCat124 Note that this won't run automatically because we added `if: ${{ github.run_attempt != '1' }}`, we don't have any passthrough only file, so this is duplicating what following actions already do. To run this you need to explicitly press the "Re-run action" button on the mirror.

The other problem that we had that was fixed in that "tee" hides some issues because the return value is always correct (even if the make target fails), and for some specific scenarios the g

You may want to test it adding
```
    for subroutine in psyir.walk(Routine):
        scalartrans = ScalarizationTrans()
        for loop in subroutine.walk(Loop):
            scalartrans.apply(loop)
```
in the omp_cpu_trans.py or omp_gpu_trans.py probably after normalise_loop step to check both nemov4 and 5. Specially because we mostly care about 5 nowadays

src/psyclone/psyir/nodes/reference.py
```diff
@@ -102,10 +102,17 @@ def is_read(self):
...
+        from psyclone.psyir.nodes.intrinsic_call import IntrinsicCall
         parent = self.parent
         if isinstance(parent, Assignment):
             if parent.lhs is self:
                 return False
+
+        # If we have an intrinsic call parent then we need to check if its
+        # an inquiry. Inquiry functions don't read from their arguments.
+        if isinstance(parent, IntrinsicCall) and parent.is_inquiry:
+            return False
```
sergisiso, line 114: They read from some arguments, e.g. var1 will be read. If that's not a problem for now just put a TODO to 2302 as we need more information to do it better.
`UBOUND(ARRAY, DIM=var1)`

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -449,6 +458,11 @@ def _compute_forward_uses(self, basic_block_list):
                     if defs_out is not None:
                         self._defsout.append(defs_out)
                     return
+                # If its parent is an inquiry function then its neither
+                # a read nor write.
+                if (isinstance(reference.parent, IntrinsicCall) and
+                        reference.parent.is_inquiry):
```
sergisiso, line 464: This will also ignore different arguments, maybe it will be safer to just say that the first argument is neither a read nor a write? (also in the comment above)

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -699,6 +711,11 @@ def _compute_backward_uses(self, basic_block_list):
                 abs_pos = reference.abs_position
                 if abs_pos < self._start_point or abs_pos >= stop_position:
                     continue
+                # If its parent is an inquiry function then its neither
+                # a read nor write.
+                if (isinstance(reference.parent, IntrinsicCall) and
+                        reference.parent.is_inquiry):
+                    continue
```
sergisiso, line 718: Same?

sergisiso (changes requested):
@LonelyCat124 I will continue the review when the new results are in, could you also look if the intrinsiccalls logic can be made more robust.

LonelyCat124:
A bug that NEMO5 has revealed is a case like this:
```
    code = '''
    subroutine test
    use mod
    integer :: i
    real, dimension(1:1000) :: arr
    do i = 1, 1000
        arr(i) = 0.0
    end do

    do i = 1, 1000
      if(zeqn > 0.0) then
        arr(i) = zh_prev
      else
        arr2(i) = zh_prev
      endif
        zh = sqrt(arr(i))
    end do
    end subroutine
    '''
```
With `arr(i)` in the second loop being only written to conditionally we shouldn't scalarise it. I might just restrict this more agressively and if the first write has an ancestor `IfBlock` that is a child of the `Loop` we're scalarizing then just refuse to do that variable.

LonelyCat124:
NEMO5 is passing now so that last change was the required fix I think.
I want to reverse the order of application of scalarization transformation and see if it leads to any performance changes - i think any performance changes will be very minor compared to the time required to performance the scalarization transformation over the whole code, but we can always do a diff and have a files_to_scalarize if it is meaningful.

I'll finish tidying up the PR as well tomorrow.

src/psyclone/psyir/nodes/reference.py
LonelyCat124: @sergisiso is it always the case that argument 1 is the inquiry and the other arguments are read-only?

LonelyCat124: Ah I just read 2302 - yeah its not causing issues for now - if we have no idea on which argument is the inquiry then its difficult and I have concerns about this.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done

LonelyCat124:
Setting the integration tests going one more time, but assumping they're ok this will be ready for review.

LonelyCat124:
@sergisiso this is ready for review now - only remaining thing I know I need to change is to revert the transformation script change.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,451 @@
...
+    ... real :: a(100), b(100)
+    ... do i = 1,100
+    ...   a(i) = i
+    ...   b(i) = a(i) * a(i)
+    ... end do
+    ... do j = 1, 100
+    ...  if(b(i) > 200) then
+    ...    print *, b(i)
+    ...  end if
+    ... end do
```
sergisiso, line 70: I think I prefer to remove this second loop and the resulting output below. We do not document cases that fail to validate in other transformation examples and it is unclear why this is here in the generated documentation without context.

src/psyclone/psyir/transformations/scalarization_trans.py
```diff
@@ -0,0 +1,451 @@
...
+        return True
+
+    @staticmethod
+    def _get_index_values_from_indices(
+            node: ArrayMixin, indices: List[Node]) -> Tuple[bool, List[Node]]:
+        '''
+        Compute a list of index values for a given node. Looks at loop bounds
+        and range declarations to attempt to convert loop variables to an
+        explicit range, i.e. an access like
+        >>> do i = 1, 100
```
sergisiso, line 207: The ">>>" makes `python -m doctest src/psyclone/psyir/transformations/scalarization_trans.py` fail. If you want a fortran snippet you can use the rst notation `.. code-block:: fortran` and it will appear properly rendered in the generated documentation

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -687,3 +687,58 @@ def test_definition_use_chains_backward_accesses_nonassign_reference_in_loop(
     assert reaches[0] is routine.children[1].loop_body.children[1].children[1]
     assert reaches[1] is routine.children[1].loop_body.children[0].lhs
     assert reaches[2] is routine.children[0].lhs
+
+
+def test_definition_use_chains_backward_accesses_empty_schedules(
+    fortran_reader,
+):
+    '''Coverage to handle the case where we have empty schedules inside
```
sergisiso, line 695: s/Coverage to handle/Test/ ?

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -687,3 +687,58 @@ def test_definition_use_chains_backward_accesses_nonassign_reference_in_loop(
...
+    assert len(reaches) == 3
+    assert reaches[0] is routine.children[4].rhs.children[1]
+    assert reaches[1] is routine.children[4].rhs.children[0]
+    assert reaches[2] is routine.children[0].lhs
+
+
+def test_definition_use_chains_backward_accesses_inquiry_func(
+    fortran_reader,
+):
+    '''Coverave to handle the case where we have an inquiry function
```
sergisiso, line 726: s/Coverage to handle/Test/ ?

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
```diff
@@ -937,3 +937,56 @@ def test_definition_use_chains_forward_accesses_multiple_routines(
     )
     reaches = chains.find_forward_accesses()
     assert len(reaches) == 0
+
+
+def test_definition_use_chains_forward_accesses_empty_schedules(
+    fortran_reader,
+):
```
sergisiso, line 944: Same s/Coverage to handle/Test/  in this 2 docstrings?

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
```diff
@@ -0,0 +1,967 @@
...
+'''This module tests the scalarization transformation.
+'''
+
+from psyclone.core import VariablesAccessInfo
+from psyclone.psyir.nodes import Loop
+from psyclone.psyir.transformations import ScalarizationTrans
+from psyclone.tests.utilities import Compile
+
+
+def test_scalararizationtrans_is_local_array(fortran_reader):
```
sergisiso, line 45: Docstring missing

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
```diff
@@ -0,0 +1,967 @@
...
+    # Test filter behaviour same as used in the transformation
+    local_arrays = filter(
+            lambda sig: ScalarizationTrans._is_local_array(sig, var_accesses),
+            var_accesses)
+    local_arrays = list(local_arrays)
+    assert len(local_arrays) == 1
+    assert local_arrays[0].var_name == "local"
+
+
+def test_scalarizationtrans_have_same_unmodified_index(fortran_reader):
```
sergisiso, line 115: Docstring missing

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
```diff
@@ -0,0 +1,967 @@
...
+    unmodified_indices = filter(
+            lambda sig: ScalarizationTrans._have_same_unmodified_index(
+                sig, var_accesses),
+            local_arrays)
+    unmodified_indices = list(unmodified_indices)
+    assert len(unmodified_indices) == 1
+    assert unmodified_indices[0].var_name == "a"
+
+
+def test_scalarizationtrans_check_first_access_is_write(fortran_reader):
```
sergisiso, line 163: Docstring missing

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
```diff
@@ -0,0 +1,967 @@
...
+
+    first_write_arrays = filter(
+            lambda sig: ScalarizationTrans._check_first_access_is_write(
+                sig, node, var_accesses),
+            unmodified_indices)
+    first_write_arrays = list(first_write_arrays)
+    assert len(first_write_arrays) == 2
+
+
+def test_scalarizationtrans_value_unused_after_loop(fortran_reader):
```
sergisiso, line 218: Docstring missing

sergisiso (changes requested):
Good job @LonelyCat124 this is a complex transformation but I like its thorough testing. Its also good to see the DefUseChain tested in real NEMO code, that will be useful for upcoming work.

The remaining comments are mostly about cleaning up comments and documentation.

examples/nemo/scripts/omp_cpu_trans.py
```diff
@@ -72,6 +75,13 @@ def trans(psyir):
     :type psyir: :py:class:`psyclone.psyir.nodes.FileContainer`
 
     '''
+    # TODO Remove
+    scalartrans = ScalarizationTrans()
+    for subroutine in psyir.walk(Routine):
+        loops = subroutine.walk(Loop)
+        loops.reverse()
+        for loop in loops:
+            scalartrans.apply(loop)
```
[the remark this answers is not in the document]
LonelyCat124: I could - my concern is how long it takes to apply the scalarization transformation to the full code.
I think adding it to normalise_loops with default set to False might be the best option for now.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -250,16 +250,20 @@ def find_forward_accesses(self):
...
-                    chain = DefinitionUseChain(
-                        self._reference.copy(),
-                        body,
-                        start_point=ancestor.abs_position,
-                        stop_point=sub_stop_point,
-                    )
-                    chains.insert(0, chain)
+                    # If we have a basic block with no children then skip it,
+                    # e.g. for an if block with no code before the else
+                    # statement, as is found in NEMO4.
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -835,6 +852,11 @@ def find_backward_accesses(self):
             # Now add all the other standardly handled basic_blocks to the
             # list of chains.
             for block in basic_blocks:
+                # If we have a basic block with no children then skip it,
+                # e.g. for an if block with no code before the else
+                # statement, as is found in NEMO4.
```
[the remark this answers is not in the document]
LonelyCat124: Removed them all

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: It was to show a case that didn't scalarise, but I can remove the references to b

LonelyCat124: Actually I don't like removing it because it no longer has any context behind why the transformation does anything. If you still want me to I can remove it but I think showing a case where it only scalarizes some elements makes more sense conceptually. I'll add some more documentation explaining it.

src/psyclone/psyir/transformations/scalarization_trans.py
LonelyCat124: I think I've fixed that now.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
LonelyCat124: Done

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
LonelyCat124: added

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
LonelyCat124: added

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
LonelyCat124: added

src/psyclone/tests/psyir/transformations/scalarization_trans_test.py
LonelyCat124: added

.github/workflows/nemo_tests.yml
```diff
@@ -96,6 +96,13 @@ jobs:
...
+        # Check the output is as expected for the first 6 digits (if not exit with error message)
+        tail -n 1 output.txt | grep -q " it :      10"  || (echo "Error: 'it :      10' not found!" & false)
+        tail -n 1 output.txt | grep -q "|ssh|_max:  0.259483" || (echo "Error: '|ssh|_max:  0.259483' not found!" & false)
+        tail -n 1 output.txt | grep -q "|U|_max:  0.458515" || (echo "Error: '|U|_max:  0.458515' not found!" & false)
+        tail -n 1 output.txt | grep -q "S_min:  0.482686" || (echo "Error: 'S_min:  0.482686' not found!" & false)
+        tail -n 1 output.txt | grep -q "S_max:  0.407622" || (echo "Error: 'S_max:  0.407622' not found!" & false)
```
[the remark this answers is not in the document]
LonelyCat124: Changed.

examples/nemo/scripts/omp_cpu_trans.py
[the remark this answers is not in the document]
LonelyCat124: Done

LonelyCat124:
@sergisiso I've fixed the changes for the review now so back to you.

src/psyclone/psyir/transformations/scalarization_trans.py
sergisiso: ah, then it's ok

sergisiso (approved):
@LonelyCat124 All comments have been addressed, I fixed a couple of small things with the passthrough and updated the version since it is the first PR after a release. I will launch the tests once more since we touched the scripts although Scalarization should be disable by default.

While updating the changelog I also realised that we use "Scalarization" while other parts of the code  use the "ise" instead of "ize". @LonelyCat124 @arporter should this be renamed as Scalarisation or is it fine as it is?

arporter:
I would prefer to keep with the British spelling :-)

sergisiso:
ok I've done a global s/calariz/calaris/ that I think should cover all instances. At least grepping I can't find anything wrong and still passes all tests.

LonelyCat124:
Ah I just did that too - I will not push then.

Issue #2011, opened 2023-01-17
Combine ModuleManager and search/kernel path


hiker:
In #1991 a ModuleManager is added, which helps with the driver creation by allowing copying all required source files for a driver into the driver source file, creating truly stand-alone drivers without outside dependencies (otherwise, the driver will depend on the kernel, which because of meta-data declaration will always depend on constanst_mod, kernel_mod, argument_mod).

We should investigate if the current path handling (-d/-I) could be combined with the ModuleManager. The module manager could maintain two separate list of search paths - one for -d, one for -I - but at the same time add them to its internal cache. This should make the finding of kernels faster, and avoids code duplication (and frequent accesses to the file system).

The one big problem that I noticed when looking at this:
ATM the kernel_path handling will detect if there is more than one file that defines the same module, and raises an exception. This would be extremely difficult to implement: if there are two files with the same name in different paths, searching would stop the moment the first file is found, and since the information is cached, it would not be detected that there is a duplicate. It might happen later, when a different kernel requires the ModuleManager to look at more paths that a clash is detected. Or we would need to parse all files initially, but that can be a huge overhead (e.g. if we want to be able to find all LFRic files, and we pass the root of the working directory as path, which will add over 700 files for gungho ... each time we start PSyclone :(  ).

It seems unlikely that we will ever have a duplicated file (since even the compiler will not complain, though at link time an error might be raised if the same symbol is defined in different .o files ... but if the files are in library, I don't think the linker would even complain). Additionally, imho it should be the task of the build system to detect this, not PSyclone (if I am not mistaken, the dependency analysis that the LFRic build system runs to detect module dependencies, will crash if the same module exists, since it would try to add a duplicated key into the DB), so I would think that this condition would be detected at some stage (not sure if this is before or after PSyclone, and how this might change with FAB).

@arporter , @sergisiso, @rupertford , @TeranIvy  - feedback welcome. If we agree to ignore duplicated modules, the implementation is really trivial, but if we agree that we want PSyclone to detect this, the overhead seems not to be worth the effort (and I would close this ticket as invalid).

PR #2564, opened 2024-04-30
(Closes #2462) generalise ModuleManager

### LINKED ISSUE #2462: Generalise ModuleManager?
In working on #2413 (and unifying the way we search for Containers in particular) I thought it would be good to use the `ModuleManager` as an efficient way of only parsing the files that we have to and those only once. (This is needed for instance when attempting to 'module-inline' called routines. It's likely that we're going to end up querying the same modules repeatedly.)

However, it turns out the `ModuleManager` is currently LFRic specific so we should probably make it generic and then subclass it for LFRic. Initially I thought that perhaps we should use fparser to determine which modules a source file contains but, as @hiker predicted, this is prohibitively expensive. (Especially since we'll have to do it every time psyclone is run.) A better solution may therefore be to use a regex to identify modules. The LFRic sub-class could optimise this out as the coding standards mean we know that there is only one module per file and that its name is essentially given by that of the file.

EDIT: another selling point for using the `ModuleManager` is that it finally frees us from the assumption that a module is found in a source file of the same name.

arporter:


arporter:
My concern is not LFRic but everything else. However, I take your point about potentially, repeatedly reading 6800 files! I've taken a look at the NEMO source and the rule there *appears* to be that a file <some_name>.f90 contains `module some_name`. In WaveWatchIII that rule largely holds but there are exceptions. @LonelyCat124 and @sergisiso, could you comment on the other sources that you've been looking at? As @hiker said earlier in #2462, a good first test would be `if module_name in file_name`.

The "where to put the file reading" problem is a bit of a chicken-and-egg. `ModuleManager` has to read a file to confirm that it contains a given module (unless this is strictly enforced as it is in LFRic) and it is only then that it creates a `ModuleInfo` object for that module. However, currently it is the `ModuleInfo` that contains the functionality for reading (and parsing, processing) the source file.

arporter:
Using:

    from difflib import SequenceMatcher
 
    score = SequenceMatcher(None, str1, str2).ratio()

gives a floating-point measure of how similar two strings are (1.0 for identical, 0.0 for nothing in common).

arporter:
> I love ❤️ the similarity one! Can we try to test various module names with the basename of the files? We could define a threshold (again, in the config file?) for which files are similar enough.

Yes, I think that might be a good, general-purpose solution that doesn't require too much work. (I need this in a hurry really.)

LonelyCat124:
Socrates is ~ 700 files - it seems to be either module_name_mod.f90 or module_name.f90. There are some non-module containing files but I guess this just doesn't deal with those.

arporter:
The problem at the moment is that the ModuleManager expects to process all files it encounters as it searches for a particular module (in `get_module_info`) whereas we really need it to search through all files and identify the most likely candidates before actually processing them. We may need to keep a record of all files we have seen but not processed in order to do this?

LonelyCat124:
Can it (or indeed PSyclone at all) deal with multiple modules inside a single file?

hiker:
> The problem at the moment is that the ModuleManager expects to process all files it encounters as it searches for a particular module (in `get_module_info`) whereas we really need it to search through all files and identify the most likely candidates before actually processing them. We may need to keep a record of all files we have seen but not processed in order to do this which is quite a big change to how it currently works?

Yes. ATM we immediately store a mapping from the (unverified/assumed) module names to ModuleInfo object, which (initially) only stores the path to the file. So, if we can't deduce the mod name (based on coding style), we should store this information to avoid having to back to the file system again.

hiker:
> Can it (or indeed PSyclone at all) deal with multiple modules inside a single file?

Somewhat. It can map several module information to the same file. But (from memory), if you then request the source code (or fparser or psyir info), each module info will independently read and parse the source code again. It shouldn't be hard to fix that though, each module info could keep a list of other modules (and then use the module manager to update the other objects. I feel there might be an even better solution ... maybe we can share some state info between these module info objects that come from the same file??

arporter:
> I feel there might be an even better solution ... maybe we can share some state info between these module info objects that come from the same file??

Should we have an intermediate e.g. `FileInfo` object that then stores references to related `ModuleInfo` instances if they've been created for it (or an empty list if the contents of the file are yet to be examined)? These `FileInfo` objects would then keep track of the files we've found and their full path (+...?). I did wonder whether this should be handled in `fparser` but decided that fparser should probably just stick to parsing.

hiker:
> Should we have an intermediate e.g. `FileInfo` object that then stores references to related `ModuleInfo` instances if they've been created for it (or an empty list if the contents of the file are yet to be examined)? These `FileInfo` objects would then keep track of the files we've found and their full path (+...?). I did wonder whether this should be handled in `fparser` but decided that fparser should probably just stick to parsing.

That was actually my very first thought, but then I expected you to say that this should be done in the PSyIR - just make FileContaier to store that info (and be populated later on parsing time) :)  That was pretty stupid reasoning, so yes, I think that would make sense, and the very clean solution to handle multiple modules in one file.

sergisiso:
@arporter I am not sure if I understand why do you need to extend the setter? I would expect the Module manager to use the same include_dir (what is passed with the -I option). So it just needs to consult the property of this value. Or is this different?

arporter:
The issue is that currently, the ModuleManager and Config objects are both singletons and have their search paths set separately. On reflection, probably ModuleManager should *not* store these paths at all and simply look them up from the Config object?

sergisiso:
@arporter Does it cache if the regex has found the modules? I am surprised this is necessary, the regex should be very fast.

arporter:
No, it caches which directories it has searched. (The concern I think is recursively exploring the directories.) It also caches the contents of a file once it has been read (which obviously it has to be to use the regex).

sergisiso:
> No, it caches which directories it has searched. (The concern I think is recursively exploring the directories.)
Still, I can grep for a string in all LFRic trunk in 0.2s

I understand doing it once and getting a map of module_name -> file_where_it_lives. But why do we have to repeat this (since the include_path won't change in a psyclone execution), couldn't we do this once and the beginning and then just look up the map?

I have to admit that I am also unsure about the recursive functionality, because then it diverges from the -I flag classic of compilers.

sergisiso:
> It isn't done repeatedly. The search just expands until it finds the requested module. The files it finds on the way are cached so that the next time a module is requested they don't have to be found again.

Would it impact to much the performance to remove the current stop and let all the -I path files be mapped at this stage?

arporter:
> Would it impact the performance to remove the current stop and let all the -I path files be mapped at this stage?

I suppose the issue is that if we do that then we pay the full cost of the mapping every time we launch PSyclone (that's probably only a small percentage of the total PSyclone run time though). Apart from a reduction in code complexity, I'm not sure what we gain by doing this? Would it be that the ModuleManager then stores a map of files rather than a list of directories (and thus has a clearer distinction from the info. held by Config)?

hiker:
> > Would it impact the performance to remove the current stop and let all the -I path files be mapped at this stage?
> 
> I suppose the issue is that if we do that then we pay the full cost of the mapping every time we launch PSyclone (that's probably only a small percentage of the total PSyclone run time though). Apart from a reduction in code complexity, I'm not sure what we gain by doing this? Would it be that the ModuleManager then stores a map of files rather than a list of directories (and thus has a clearer distinction from the info. held by Config)?

I would very much prefer to stop asap instead of doing a full scan. At this stage with the Fab build system we have no way of communicating this mapping from the worker processes back (we might look at that some time in the future), and the same holds for the old makefile-based builds. Which means we potentially have dozens of PSyclone processes (48 on our machine) running at the same time, each one then doing a full scan of all source files (at least when driver creation is requested, we need to look at all directories, including jules, ukca, um, ...). I am not sure how happy lustre will be about this. I am uncertain if the information will be/can be cached, since potentially new files will be created in the tree (psy and alg files) - and this is mostly meta-data info, which can be expensive.

hiker:
> > I have to admit that I am also unsure about the recursive functionality, because then it diverges from the -I flag classic of compilers.
> 
> True, it's optional but the default is to do it.
> 
> All of this is really a question of how the ModuleManager was originally designed and isn't something I was proposing to change here. Perhaps @hiker can comment?

First of all, the behaviour was implemented to mirror existing behaviour of PSyclone when specifying the kernel path - it searches recursively as well. Secondly, we don't really want to constantly ask users to update the build system when adding a new directory (alternatively, the build system would need to scan for directories and provide them on the command line).

hiker:
> > No, it caches which directories it has searched. (The concern I think is recursively exploring the directories.)
> 
> Still, I can grep for a string in all LFRic trunk in 0.2s

Are you sure that the data is not cached by the OS? Here what I see on our research computer on a lustre filesystem:

[jxh903@gadi-login-07 lfric]$ time grep -ir "where is this" .

real	1m1.211s
user	0m0.693s
sys	0m3.839s
[jxh903@gadi-login-07 lfric]$ time grep -ir "where is this" .

real	0m4.780s
user	0m0.589s
sys	0m2.192s

So >10 times faster the second time around. And I got similar timings on a different node (I don't think I can drop cache myself)

hiker:
> The issue is that currently, the ModuleManager and Config objects are both singletons and have their search paths set separately (albeit, the ones passed in to ModuleManager are looked up from Config). On reflection, probably ModuleManager should _not_ store these paths at all and simply look them up from the Config object?
> 

They should be only in one place, I agree. The reason why I didn't do more towards this was that I think the kernel path and module manager path should be combined, but since module manager will contain more files (kernel path only includes `lfric.../kernels`, while the module manager will contain `lfric.../*`) - and that could in theory change the behaviour, so instead I opened #2011 to discuss this :) So I guess, this PR will also close #2011.

arporter:
I really don't want to do #2011 here :-) I want this one to go on as soon as possible because it's needed for the LFRic GPU work. I can add some TODOs to the relevant places for now.

sergisiso:
> Are you sure that the data is not cached by the OS?

Aha, yes it was (but also for potentially consecutive psyclone invocations). In my laptop it goes from 5s (first) to 0.2s (following). But I do see the problem with Luster now, which is a common target and agree with the need for caching/stop searching.

I still see some problems with the recursive behaviour for generic code, as I would expect to put the same INCLUDE_PATH as its provided to the compiler, but since it acts differently in may get a different result if it has another module with the same file in a subdirectory. I suppose we at least do a BFS search so we potentially always pick the one in the top directory?

hiker:
> I still see some problems with the recursive behaviour for generic code, as I would expect to put the same INCLUDE_PATH as its provided to the compiler, but since it acts differently in may get a different result if it has another module with the same file in a subdirectory. I suppose we at least do a BFS search so we potentially always pick the one in the top directory?

It uses os.walk to find the directories, which (by default) goes top to bottom. It then searches all files in the first directory found, then the next etc - so yes, BFS.

Same file name in the build tree is a recipe for disaster :) I don't think we need to be concerned about that. FWIW, Fab will immediately abort if it finds the same file name twice (so there's a lot of cherry-picking involved then importing um, jules, ukca, ... packages, since they often have the same utilities file ... in different versions).

I just realised that recursive is a flag when adding the search path (defaulting to True), so it would be fairly easy to switch this off (e.g. different command line options could add a path either recursive or not).

arporter:
Ready for review now from either @hiker or @sergisiso . It generalises the ModuleManager so that it no longer assumes a strict mapping between filename and the module that it contains. More significantly, it also changes the interface-resolving mechanism to use the ModuleManager. Integration tests are running.

src/psyclone/parse/file_info.py
```diff
@@ -0,0 +1,122 @@
...
+
+    :returns: 2-tuple containing replacement for bad chars (an empty string
+              and the position from where encoding should continue).
+    :rtype: tuple[str, int]
+
+    '''
+    return ("", err.end)
+
+
+codecs.register_error("file-error-handler", log_decode_error_handler)
```
hiker, line 58: I still don't like code execution at import time. Esp. since this will have a global impact (make the function a static member, and just register it before and unregister it after the statement.
But, admittedly I can't see any real disadvantage to change this globally, so I am happy to leave it this way.

src/psyclone/parse/file_info.py
```diff
@@ -0,0 +1,122 @@
...
+        :returns: the contents of the file (utf-8 encoding).
+        :rtype: str
+
+        '''
+        if self._source_code is None:
+            # Error handler is defined at the top of this file. It simply skips
+            # any characters that result in decoding errors. (Comments in a
+            # code may contain all sorts of weird things.)
+            with open(self._filename, "r", encoding='utf-8',
+                      errors='file-error-handler') as file_in:
```
hiker, line 119: Hmm - ok, I wrote above that I am happy to ignore the global state change (at a not well defined time, considering that it happens on import time) ... but what about:
```
            with open(self._filename, "r", encoding='utf-8',
                      errors='ignore') as file_in:
```
That seems to work for me (for the given tests), simplify the code, and avoid a global change.

src/psyclone/parse/module_info.py
```diff
@@ -315,63 +248,76 @@ def get_psyir(self):
...
+        for cntr in self._psyir.walk(Container):
+            if cntr.name.lower() == self.name:
+                return cntr
+
+        # We failed to find the Container - double-check the parse tree
+        for mod_stmt in walk(self.get_parse_tree(), Fortran2003.Module_Stmt):
+            if mod_stmt.children[1].string.lower() == self.name:
+                # The module exists but we couldn't create PSyIR for it.
+                # TODO #11: Add proper logging
+                print(f"File '{self.filename}' does contain module "
```
hiker, line 290: That's a great addition!!

src/psyclone/parse/module_manager.py
```diff
@@ -39,12 +39,21 @@
...
+import re
 
 from psyclone.errors import InternalError
+from psyclone.parse.file_info import FileInfo
 from psyclone.parse.module_info import ModuleInfo
 
 
+# regex to find Fortran module names. Have to be careful not to match
+# e.g. "module procedure :: some_sub".
+_MODULE_PATTERN = re.compile(r"^\s*module\s+([a-z]\S*)\s*$",
```
hiker, line 53: Could this be a class variable initialised with None, and in `get` only compile it? While afaik python will only execute the `compile` the first time this module is imported, I don't think there is a need  to expose this symbol unnecessary.

src/psyclone/parse/module_manager.py
```diff
@@ -53,6 +62,10 @@ class ModuleManager:
     # Class variable to store the singleton instance
     _instance = None
 
+    # How similar a file name must be to a module name to trigger reading
+    # of the file.
+    _threshold_similarity = 0.7
```
hiker, line 67: That is a really great idea. I can't wait to see how this works with LFRic and UM :)

src/psyclone/parse/module_manager.py
```diff
@@ -179,50 +221,63 @@ def get_module_info(self, module_name):
...
-        function can be implemented in a derived class to actually parse the
-        source file if required.
+    def get_modules_in_file(self, finfo):
+        '''
+        Uses a regex search to find all modules defined in the file with the
+        supplied name.
 
-        :param str filename: the file name for which to find the list \
-            of modules it contains.
+        :param str filename: the fully-qualified name of the file to check for
```
hiker, line 266: The parameter is now `finfo`, not `filename`.

src/psyclone/psyir/nodes/container.py
```diff
@@ -187,6 +188,37 @@ def get_routine_psyir(self, name):
                 return routine
         return None
 
+    def resolve_routine(self, name):
+        '''This function returns a list of function names that might be
+        actually called when the routine `name` is called. In most cases
+        this is exactly name, but in case of a generic subroutine the
```
hiker, line 194: Backticks around `name`.

src/psyclone/psyir/nodes/container.py
```diff
@@ -187,6 +188,37 @@ def get_routine_psyir(self, name):
...
+        name might change. For now (since we cannot compare routine
+        signatures yet), we return the list of all possible functions that
+        might be called.
+
+        :param str name: the name of the routine to resolve
+
+        :returns: the names of those routines that may actually be invoked
+            when the routine `name` is called or an empty list if there is no
+            routine with that name in this container.
+        :rtype: list[None | psyclone.psyir.nodes.Routine]
```
hiker, line 204: I believe that should be `rtype: list[str]`

src/psyclone/psyir/symbols/containersymbol.py
```diff
@@ -213,43 +212,36 @@ class FortranModuleInterface(ContainerSymbolInterface):
...
-            f"Module '{name}' (expected to be found in '{name}.[f|F]90') not "
-            f"found in any of the include_paths directories "
-            f"{Config.get().include_paths}.")
+        # pylint: disable-next=import-outside-toplevel
+        from psyclone.parse import ModuleManager
+        mod_manager = ModuleManager.get()
+
+        # TODO #2011 - rationalise how this interacts with the kernel search
+        # path set in generate().
+        mod_manager.add_search_path(Config.get().include_paths)
```
hiker, line 233: Great change, I love the use of the ModuleManager (of course I am biased :) )

src/psyclone/psyir/symbols/containersymbol.py
```diff
@@ -213,43 +212,36 @@ class FortranModuleInterface(ContainerSymbolInterface):
 
     @staticmethod
     def import_container(name):
```
hiker, line 214: Not this PRs fault, but I wonder if there isn't a better name? `import_container`  does not really tell me what this function does (esp. not that it returns something, I would expect that something is added to a symbol table??).
Maybe `get_container_psyir`?

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -278,6 +278,12 @@ def get_non_local_read_write_info(self, node_list, read_write_info):
         '''Returns the information about non-local variables that are read
         or written.
 
+        :param node_list: list of nodes containing Kernel calls to interrogate.
```
hiker, line 281: Gee, embarrassing - thanks for fixing the missing parameters.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -332,9 +348,10 @@ def _resolve_calls_and_unknowns(self, outstanding_nonlocals,
         object.
 
         :param outstanding_nonlocals: the information about symbol type,
-            module_name, symbol_name and access information
-        :type outstanding_nonlocals: list[tuple[str, str, str,
-                              :py:class:`psyclone.core.Signature`,str]]
+            module_name, symbol_name, signature and access information.
+        :type outstanding_nonlocals: list[tuple[
+            str, str, str, :py:class:`psyclone.core.Signature`,
+            :py:class:`psyclone.core.SingleVariableAccessInfo`]]
```
hiker, line 354: Thanks.

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -373,29 +390,28 @@ def _resolve_calls_and_unknowns(self, outstanding_nonlocals,
...
-                    else:
+                cntr = mod_info.get_psyir()
+                if not cntr:
+                    print(f"[CallTreeUtils._resolve_calls_and_unknowns] "
+                          f"Cannot get PSyIR for module '{module_name}' - "
+                          f"ignoring unknown symbol '{signature[0]}'.")
+                    continue
+                for routine_name in cntr.resolve_routine(signature[0]):
+                    routine = cntr.get_routine_psyir(routine_name)
+                    if not routine:
```
hiker, line 401: Much neater, thanks

src/psyclone/psyir/tools/call_tree_utils.py
```diff
@@ -411,25 +427,36 @@ def _resolve_calls_and_unknowns(self, outstanding_nonlocals,
...
+                    # symbol table won't have been populated)
+                    sym_tab = cntr.symbol_table
+                    try:
+                        sym = sym_tab.lookup(signature[0])
+                        if sym.is_constant:
+                            continue
+                    except KeyError:
+                        print(f"Unable to check if signature '{signature}' "
+                              f"is constant.")
+                        sym = None
```
hiker, line 459: Not your fault, but do we need `sym` actually? Same for line 435

src/psyclone/tests/domain/gocean/transformations/gocean1p0_transformations_test.py
```diff
@@ -1485,9 +1487,21 @@ def test_accroutinetrans_module_use():
...
-            "which is imported. If this symbol "
+    assert ("accesses the symbol 'rdt: Symbol<Import(container='model_mod')>'"
+            " which is imported. If this symbol "
             "represents data then it must first" in str(err.value))
+    # Tell the ModuleManager where to find the module that is being USED by
+    # the kernel.
+    mod_man = ModuleManager.get()
+    mod_man.add_search_path(get_base_path("gocean1.0"))
+    # Now that we can resolve the symbols, we know that `rdt` is a parameter
+    # (and is not a problem) but that `magic` is a variable.
```
hiker, line 1498: Great improvement.

hiker (commented):
Will continue later.

hiker (changes requested):
Nice PR, it is good to see the module manager becoming more useful.
There are a few minor issues mentioned in the comments.

I also realised that we now have two FileInfo classes (the other one in ./parse/algorithm.py), not exactly ideal :(

Looking at the documentation, I wonder if the new FileInfo object should be mentioned together with the module manager in the dev guide. And that section sits atm in the psy_data.rst, now that the module manager is used elsewhere, maybe it should go in a different section (not sure ... maybe modules.rst)?

src/psyclone/parse/file_info.py
arporter: Gone now.

src/psyclone/parse/file_info.py
arporter: Oh yes. Can't believe I'd missed that. Thanks very much.

src/psyclone/parse/module_manager.py
arporter: Oh yes. In fact I've just made it an instance variable since this is a Singleton anyway.

src/psyclone/parse/module_manager.py
arporter: Fingers crossed! At least the integration tests are happy.

src/psyclone/psyir/nodes/container.py
arporter: Done.

src/psyclone/psyir/nodes/container.py
arporter: Oops again.

src/psyclone/psyir/symbols/containersymbol.py
arporter: I'll take praise where I can get it :-)

src/psyclone/psyir/tools/call_tree_utils.py
arporter: Well, I think I reviewed it so it's embarrassment shared :-)

src/psyclone/psyir/tools/call_tree_utils.py
arporter: Good spot, gone now.

src/psyclone/tests/parse/module_info_test.py
```diff
@@ -50,12 +53,12 @@
                          "mod_man_test_setup_directories")
 def test_module_info():
     '''Tests the module info object.'''
-    mod_info = ModuleInfo("a_mod", "file_for_a")
+    mod_info = ModuleInfo("a_mod", FileInfo("file_for_a"))
     assert mod_info.filename == "file_for_a"
-    assert mod_info._source_code is None
+    assert mod_info.name == "a_mod"
```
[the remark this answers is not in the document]
arporter: Done (I think).

src/psyclone/tests/parse/module_manager_test.py
```diff
@@ -147,20 +134,24 @@ def test_mod_manager_add_files_from_dir():
     tmp/d2/d_mod.X90
     tmp/d2/d4/e_mod.F90
     tmp/d2/d4/f_mod.ignore
-    '''
 
+    '''
```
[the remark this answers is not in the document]
arporter: Oh yes, good spot.

src/psyclone/tests/psyir/nodes/container_test.py
```diff
@@ -197,3 +197,52 @@ def test_container_get_routine_psyir():
...
+    '''
+    Test that the resolve_routine() method works as expected for:
+
+      * an individual routine;
+      * a generic interface;
+      * a name that does not correspond to a routine;
+      * a name that does not exist.
+
+    '''
+    psyir = fortran_reader.psyir_from_source('''
```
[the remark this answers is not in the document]
arporter: I'm not sure which fixture you mean? AFAICT, there's no existing fixture in scope here that would help and we don't want to read from file - we just need some Container PSyIR to test.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -2452,6 +2452,7 @@ def test_rename_codeblock_error(fortran_reader):
 
 # resolve_imports
 
+@pytest.mark.usefixtures("clear_module_manager_instance")
```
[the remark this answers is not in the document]
arporter: Oh, that's an interesting idea. However, doesn't `scope=class` mean that all tests in a class will *share* the fixture rather than running it for each one as we need? In the past, I've created a new fixture that simply makes use of the fixture that I want to `autouse` and then marked my new fixture as `autouse` for all test functions in that module.

arporter:
Thanks for that Joerg. I've restructured the docs (which made me realise some doc strings were still wrong) by adding a new `Module Manager` section and referring to it appropriately. Good spot on the second (or first) `FileInfo`. I've renamed it `AlgFileInfo` as that seems appropriate and didn't require very much effort :-)

src/psyclone/tests/parse/module_info_test.py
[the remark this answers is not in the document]
hiker: Thanks!

src/psyclone/tests/psyir/symbols/symbol_table_test.py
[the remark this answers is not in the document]
hiker: I admit I am not entirely certain how fixtures works in test classes. To be investigated at some time in the future :)

src/psyclone/tests/psyir/nodes/container_test.py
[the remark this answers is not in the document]
hiker: Oops, yes, you are right, that's only in parse - apologies.

hiker:
The documentation builds fine, all issues were addressed. We have one CI failures, but that appears to be related to a missing/incorrect OpenMPI module:
``` 
Lmod has detected the following error: The following module(s) are unknown:
"openmpi/5.0.2"
```

Additionally, we have one question for @sergisiso.

@arporter , let me know if you are happy to go ahead with a merge anyway (and bring it up to master, there are some conflicts), or if yuo prefer me to wait for feedback and CI to be fixed.

src/psyclone/psyir/symbols/containersymbol.py
sergisiso: Yeah, I like the new proposed name

arporter: I've actually gone with `get_container`: since we're in the PSyIR here, I felt the `_psyir` was superfluous. Is that OK @sergisiso and @hiker?

hiker: I'm happy with that. In the context of PSyclone, Container makes it obvious what it is.

hiker:
All issues are addressed, tests and CI passes; documentation builds fine. No need to update user guide. Progressing to merge

hiker (approved):
CI passed, progressing to merging.

Issue #2156, opened 2023-05-30
Add support for marking routines as inquiry 


sergisiso:
#2147 is adding a flag that marks subroutines as elemental and pure.

Similarly it would be good to mark inquiry routines (that have arrays as arguments but the data of the array is not accessed).

This impacts the way the array2loop transformations treats them (currently there are manual exceptions for LBOUND, UBOUND, ...) and how the GPU offloading considers dependencies to their data (again we have manual exceptions)

arporter:
Although we do now support these concepts, the fparser2 frontend is still ignoring them, e.g. Simon reports:

    module m

    contains

      impure elemental subroutine s     ! -> Issue 1: both "impure" and "elemental" disappear
        ...

LonelyCat124:
I think this is a pretty small fix (though I'm not super comfortable with the fparser2 frontend).
I added an else and print statement for now:
```
        if prefix:
            for child in prefix.children:
                if isinstance(child, Fortran2003.Prefix_Spec):
                    if child.string not in SUPPORTED_ROUTINE_PREFIXES:
                        raise NotImplementedError(
                            f"Routine has unsupported prefix: {child.string}")
                    else:
                        print("SUPPORTED BUT WE DIDNT DO ANYTHIGN")
                else:
                     base_type, _ = self._process_type_spec(routine, child)
```
and the SUPPORTED BUT WE DIDNT DO ANYTHING happens for impure and elemental. I can try to draft a fix together but I might need help putting together (unit) tests from someone more familiar with fparser2's frontend (unless you're actively working on this already @arporter )

LonelyCat124:
Turns out (as usual with fparser2) i didn't understand it properly.

The above code is in subroutine handler, which I think would come up if we had a subroutine outside a module. I'm unclear what in the fparser2 frontend creates the routine symbol for such a scoping.

If we have a module containing it, then we have `_process_routine_symbols`, which does pull out is_elemental or is_pure and add them to the routine symbol. This is not dropped by fparser (if a module is containing it) but actually by the fortran backend, which just has this code:

```
            args = [symbol.name for symbol in node.symbol_table.argument_list]
            suffix = ""
            if node.return_symbol:
                # This Routine has a return value and is therefore a Function
                routine_type = "function"
                if node.return_symbol.name.lower() != node.name.lower():
                    suffix = f" result({node.return_symbol.name})"
            else:
                routine_type = "subroutine"
[...]

Issue #2201, opened 2023-06-30
[PSyIR] Support for Fortran files without modules?


arporter:
In processing Socrates, @LonelyCat124 has been encountering source files that just contain program units (subroutines) without a module. At the minute, PSyclone does not create/populate a SymbolTable for a FileContainer - this is explicitly rejected by the Fortran backend.

We need to consider how we want to support 'bare' program units. It seems sensible that we would populate a SymbolTable associated with the FileContainer but I can't remember why we don't do that. Perhaps @sergisiso, @rupertford and @hiker have some thoughts?

rupertford:
I'm not sure I understand. A FileContainer can contain multiple program units. In Fortran these could be a program, a module, a subroutine, a function (and some data stuff we don't support). So a subroutine would be a Routine within a FileContainer. If we only have one program unit then it is also valid to not have a FileContainer at all. I'm pretty sure there are lots of examples in our tests that only use subroutines, functions or programs and have no modules.

arporter:
That's all quite true. However, we don't put Symbols for the various routines into a SymbolTable and therefore when one routine calls another one, it can't find a Symbol for it. (@LonelyCat124 hit this while trying to workaround a statement function by introducing a new FUNCTION into the source file instead.)

rupertford:
Ah, I see. Yes, it makes sense for Routine and Container symbols to be added to parent Container symbol tables including a FileContainer.

LonelyCat124:
A simple standalone testcase again:
```
def test_fail_case(fortran_reader):
    code = '''
    real function my_func(a1, a2)
    real a1, a2
    my_func = a1 * a2
    end function
    subroutine a(b, c, d)
    real b, c, d

    b = my_func(c, d)

    contains
    end subroutine'''
    psyir = fortran_reader.psyir_from_source(code)
```

LonelyCat124:
I tried to see if I could have a quite look at how to add a SymbolTable to a FileContainer but I think this will be more complex than just that.

I think fparser2 relies on there being a `module` present to create the `RoutineSymbol` - since `_process_routine_symbols` is only called from `generate_container` (which I cannot find a single call to other than tests? Any idea what this is for @arporter?) and from `_module_handler`, which we only reach when finding a Module.

I guess the bit I can't really find is where the FileContainer is created for non-programs - for `_program_handler` it clearly creates its own FileContainer, but doesn't process any contained routine symbols - I assume since the base assumption is there will be a module.

[...]

Issue #1200, opened 2021-04-09
Routine and Container own symbol names issues


sergisiso:
Routine and Containers (maybe also other nodes) keep a "_name" string attribute. Routine also adds a symbol tagged "own_routine_symbol" and externally are both referenced by their symbols. The issue is that if the symbol or the internal name are renamed, they are not kept consistent and can produce wrong code.

I think both need to keep just a reference to the symbol defining its name and this should be the only source of truth. Then the name property must follow this reference to return the name string.

There is also another issue in that the Routine puts its RoutineSymbol into its own scope. This is not completely right as a Routine also acts as a declaration of this symbol to its parent scope, but this may still be unknown when creating the routine.

PR #2566, opened 2024-05-02
(Towards #2156) Changes to routine_copy and fortran backend for elemental and impure support

LonelyCat124:
A few small changes required to make the PSyIR layer (mostly the backends) not lose information on pure and elemental subroutines.

This might solve some of the NEMO issues reported, but only if the subroutines are contained within modules.


LonelyCat124:
Also I realise that this doesn't technically fix the stuff for the issue 2156 but just andy's comments within.

LonelyCat124:
@arporter @sergisiso Should we also ask to test this branch with NEMO before the NEMO party?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1157,6 +1157,23 @@ def routine_node(self, node):
             result = f"{self._nindent}program {node.name}\n"
             routine_type = "program"
         else:
+            # Find RoutineSymbol
+            container = node.ancestor(Container)
+            rsym = None
+            if container:
+                rsym = container.symbol_table.get_symbols().get(
```
arporter, line 1164: Not for this PR but this idiom makes me think it would be nice if `SymbolTable.lookup()` let you say that a result of `None` was acceptable, rather than having to do do a `try...except` for a KeyError.

src/psyclone/psyir/symbols/routinesymbol.py
```diff
@@ -151,6 +152,22 @@ def is_elemental(self, value):
                             f"'{type(value).__name__}'")
         self._is_elemental = value
 
+    def copy(self):
+        '''Create and return a copy of this object. Any references to the
+        original will not be affected so the copy will not be referred
+        to by any other object.
+
+        :returns: A symbol object with the same properties as this \
```
arporter, line 160: No need for line-continuation char (`\`).

src/psyclone/tests/psyir/symbols/routinesymbol_test.py
```diff
@@ -142,3 +143,16 @@ def test_routinesymbol_str():
...
+    '''
+    Test the RoutineSymbol.copy() method.
+    '''
+    asym = RoutineSymbol("a", is_elemental=True, is_pure=True)
+    new_sym = asym.copy()
+    assert new_sym is not asym
+    assert new_sym.name == asym.name
+    assert new_sym.visibility == asym.visibility
+    assert new_sym.is_elemental
+    assert new_sym.is_pure
```
arporter, line 158: Please could you extend this with an example where the original RoutineSymbol does not have `is_elemental` and `is_pure` specified in the constructor.

arporter (changes requested):
Thanks Aidan, this looks good to me.
Just a couple of very minor things to tidy up. I can't yet run the integration tests as the python modules on the runner need fixing so that will have to wait until next time.
Ref guide still builds fine.

src/psyclone/psyir/backend/fortran.py
LonelyCat124: I feel like it would be better if `lookup` returned `None` always instead of `KeyError` and the `try...except` pattern could just be `if result is None` (or similar pythonic implementation), though I imagine thats probably a decent amount of code needing rewriting.

src/psyclone/psyir/symbols/routinesymbol.py
LonelyCat124: Removed

src/psyclone/tests/psyir/symbols/routinesymbol_test.py
LonelyCat124: Done

sergisiso:
Thanks @LonelyCat124 to implement this, looking at NEMO upstream it seems that they all are elemental functions, I think it will work with your implementation but it would be good to add some tests with functions to make sure it all goes right(frontend-attribute-backend).
The signatures in nemo are:
`ELEMENTAL FUNCTION SIGN_SCALAR( val1, val2 )`
and there is one:
`impure ELEMENTAL FUNCTION pres_temp_sclr( val)`

LonelyCat124:
> Thanks @LonelyCat124 to implement this, looking at NEMO upstream it seems that they all are elemental functions, I think it will work with your implementation but it would be good to add some tests with functions to make sure it all goes right(frontend-attribute-backend). The signatures in nemo are: `ELEMENTAL FUNCTION SIGN_SCALAR( val1, val2 )` and there is one: `impure ELEMENTAL FUNCTION pres_temp_sclr( val)`

I'm guessing there's some return type on these functions too? It works if I use `real elemental function` but I get a codeblock from something else if its just `elemental function x`

sergisiso:
In nemo functions the type is in the declarations (with the same name as the function), not the function line:
```
   ELEMENTAL FUNCTION SIGN_SCALAR( pa, pb )
      !!-----------------------------------------------------------------------
      !!                  ***  FUNCTION SIGN_SCALAR  ***
      !!
      !! ** Purpose : overwrite f95 behaviour of intrinsinc sign function
      !!-----------------------------------------------------------------------
      REAL(wp), INTENT(in) :: pa,pb          ! input
      REAL(wp)             :: SIGN_SCALAR    ! result
      !!-----------------------------------------------------------------------
      IF ( pb >= 0._wp ) THEN   ;   SIGN_SCALAR =  ABS(pa)
      ELSE                      ;   SIGN_SCALAR = -ABS(pa)
      ENDIF
   END FUNCTION SIGN_SCALAR
```

LonelyCat124:
Yeah - I put that in the tests and it works. PSyclone does just create a codeblock if I have `real elemental function` but I don't know why that is, I assumed it was a generic PSyclone thing and not related to this PR.

LonelyCat124:
Also - should I do something on xfailing tests that makes the test fail if the xfail condition isn't met? I feel like those tests should either xfail or fail - if they succeed we should have a proper test for them (I just realised while playing around with #2201 that if I make the xfail test work it doesn't report anything wrong).

arporter:
> Also - should I do something on xfailing tests that makes the test fail if the xfail condition isn't met? I feel like those tests should either xfail or fail - if they succeed we should have a proper test for them (I just realised while playing around with #2201 that if I make the xfail test work it doesn't report anything wrong) - adding an else in the test fails coverage.

There is a pytest flag to have x-passes reported as failures but the trouble is we have some tests that we know pass locally but fail on GHA for reasons we've never quite managed to get to grips with. You can protect the `else` with a pragma to say to ignore coverage for it.

LonelyCat124:
> > Also - should I do something on xfailing tests that makes the test fail if the xfail condition isn't met? I feel like those tests should either xfail or fail - if they succeed we should have a proper test for them (I just realised while playing around with #2201 that if I make the xfail test work it doesn't report anything wrong) - adding an else in the test fails coverage.
> 
> There is a pytest flag to have x-passes reported as failures but the trouble is we have some tests that we know pass locally but fail on GHA for reasons we've never quite managed to get to grips with. You can protect the `else` with a pragma to say to ignore coverage for it.
> 
> EDIT: we could probably have a pytest config file that's only used on GHA but that's starting to get quite complicated.

I found a better way to do this - I assert the fail condition then call pytest.xfail after that.

> Would you mind trying to fix this? It should work as we support the two things independently (i.e. real function x and elemental function x).

I'm running some tests for 2201 at the moment, once those complete I'll try to fix it.

LonelyCat124:
@arporter I'm not sure this was actually broken, but I think my syntax was the issue.

`real impure elemental function sub(...)` gives me a Routine object as expected. I'll add this into the other function tests as well.

src/psyclone/psyir/backend/fortran.py
arporter: I think I agree with that although it would be a very big change! What do @sergisiso and @hiker think?

arporter (commented):
All requested changes have been made (thanks). I'll fire off the integration tests once more.

Actually, I can't run the integration tests until #2571 is on.

src/psyclone/psyir/backend/fortran.py
sergisiso: The KeyError was originally to imitate the python dict interface. We could also add a default value like get has to avoid the KeyError.

But in this case I think we should use the one associated with the Routine which is: self.symbol_table.lookup_with_tag("own_routine_symbol", scope_limit=self)
and will be easily accessible as node.routine_symbol with @JulienRemy PR #2570
and if this it doesn't exist we can error.

(Unfortunately routines have 2 symbols and you will need to check that the elemental/pure properties have been properly set in both - this generates a few issues like if one symbol is updated the other is not. I think we should get rid of the "own_routine_symbol" and instead have a Routine.__symbol - and it's self.name be Routine._symbol.name but this is for a separate issue as it brings some other problems)

LonelyCat124: Is that what #2570 is about? I assumed that the symbol associated with the Routine was for `FUNCTION` and is the symbol associated with their return value and is not the `RoutineSymbol` stored in the scope of the `Container` that contains the `Routine` node. As far as I can work out from the PR/fparser that is how it works, and the only constructor of a `RoutineSymbol` I can see is in `_process_routine_symbols`, which is called from the container or module_handlers.

Also whenever a `RoutineSymbol` is found inside a `Routine` during fparser it is removed and replaced with a `DataSymbol` as far as I can work out - what have I missed here?

LonelyCat124: Is there a reason that a Routine has both its own copy of the RoutineSymbol while its parent also has one? Since searching for the symbol should search above it too?

LonelyCat124: From experimentation this is very difficult behaviour to rely on.

I was correct in part - if you have a function (e.g. `real function x`) then `lookup_with_tag("own_routine_symbol") gives you a `x: DataSymbol<Scalar<REAL, UNDEFINED>, Automatic>`

If you have a subroutine (e.g. `subroutine sub`) then instead you get `sub: RoutineSymbol<NoType, pure=unknown, elemental=unknown>` - this behaviour seems strange to me and I'm not clear why the routine stores a copy of its own routine symbol?

LonelyCat124: You are correct though Sergi - the `is_elemental` and `is_pure` properties are not making it to the "own_routine_symbol" at the moment.

LonelyCat124: Aha its setup in `.name` in `Routine`...

LonelyCat124: Ok if we rename a Routine using `name` all kinds of things could be wrong... This test doesn't pass
```
def test_dummy(fortran_reader):
    code = '''module test
    contains
    impure elemental subroutine sub()
    end subroutine sub
    end module test'''
    module = container.children[0]
    routine = container.children[0].children[0]
    routine.name = "RENAME"
    assert module.symbol_table.lookup(routine.name).name == routine.name
```

LonelyCat124: I can ensure the properties are copied downwards by changing `name` to:
```
            rsymbol = None
            if self._parent:
                try:
                    rsymbol = self.parent.symbol_table.lookup(new_name).copy()                                                                                   except KeyError:
                    pass
            if not rsymbol:
                rsymbol = RoutineSymbol(new_name, NoType())

            self._name = new_name
            # Since the constructor can not mark methods as functions directly
            # the symbol will always start being NoType and must be updated
            # if a return_value type is provided.
            self.symbol_table.add(rsymbol,
                                  tag='own_routine_symbol')
```

I'm not going to change the name change upwards stuff because thats a new issue not related to this I think

LonelyCat124: I created #2582 to solve the upwards change that we should probably have, for after the FileContainer changes also.

arporter: Thanks both. A lot of this trouble is due to (or summarised by) #1200. Should I therefore review @JulienRemy 's #2570 first do you think?

arporter: > The KeyError was originally to imitate the python dict interface. We could also add a default value like get has to avoid the KeyError.

This has the advantage that it preserves the existing API so we could make this change incrementally.

hiker: > I think I agree with that although it would be a very big change! What do @sergisiso and @hiker think?

I'd love this, I hate the constant `try` construct required.

LonelyCat124: > Thanks both. A lot of this trouble is due to (or summarised by) #1200. Should I therefore review @JulienRemy 's #2570 first do you think?

I think this and #2570 are essentially independent so there is no issues with both of these PR being submitted in any order. I think there are separate issues in his PR about whether he is always expecting to get a `RoutineSymbol` (as the #1200 issue highlights). If he always expects/wants a `RoutineSymbol` (and is happy to look in the `Container` to get it if there isn't one in the `Routine`), then his PR probably will need to wait on #2575 , which in turn is waiting on this PR since I know it will affect an xfailing test in this PR.


> This has the advantage that it preserves the existing API so we could make this change incrementally.
I think that probably that is a better initial implementation - but we should perhaps try to have a rule that if someone touches a file with the `KeyError` pattern then the reviewer asks them to change to the new version, and then we can remove it entirely later with less effort?

arporter: I was wondering why we allow a `Routine` to be renamed and have reminded myself that we do this for Kernels (in PSyKAl) when we transform them (psyGen.py::rename_and_write). Also, in gocean1p0.py::_read_from_device_routine we create a Routine by parsing some generic Fortran and then we rename it. There are probably other instances too. Basically, we can't just turn this off (as I was tempted to argue for).

arporter: Anyway, all that discussion aside, please could you add a comment explaining your new addition to the `Routine.name` method.

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -367,6 +367,11 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
     end module test'''
     container = fortran_reader.psyir_from_source(code)
     output = fortran_writer(container)
+    routine = container.children[0].children[0]
```
arporter, line 370: More robust to do `routine=container.walk(Routine)[0]`

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -378,6 +383,11 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
...
```
arporter, line 386: More robust to do `routine=container.walk(Routine)[0]`

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -387,8 +397,66 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
...
```
arporter, line 400: More robust to do `routine=container.walk(Routine)[0]`

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -387,8 +397,66 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
...
+    elemental function sub(val1, val2)
+        real :: sub
+        real :: val1
+        real :: val2
+        sub = val1 - val2
+    end function
+    end module
+    '''
+    container = fortran_reader.psyir_from_source(code)
+    routine = container.children[0].children[0]
```
arporter, line 419: More robust to do `routine=container.walk(Routine)[0]`

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -387,8 +397,66 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
...
+    impure elemental function sub(val1, val2)
+        real :: sub
+        real :: val1
+        real :: val2
+        sub = val1 - val2
+    end function
+    end module
+    '''
+    container = fortran_reader.psyir_from_source(code)
+    routine = container.children[0].children[0]
```
arporter, line 436: More robust to do `routine=container.walk(Routine)[0]`

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -387,8 +397,66 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
...
+    code = '''module test
+    contains
+    real impure elemental function sub(val1, val2)
+        real :: val1, val2
+        sub = val1 - val2
+    end function
+    end module
+    '''
+    container = fortran_reader.psyir_from_source(code)
+    routine = container.children[0].children[0]
```
arporter, line 452: More robust to do `routine=container.walk(Routine)[0]`

src/psyclone/psyir/nodes/routine.py
```diff
@@ -212,12 +212,20 @@ def name(self, new_name):
                     f"Can't assign '{new_name}' as the routine name because "
                     f"its symbol table contains a symbol ({existing_symbol}) "
                     f"already tagged as 'own_routine_symbol'.")
+            rsymbol = None
```
arporter, line 215: As mentioned in the discussion, please add a comment here.

src/psyclone/psyir/nodes/routine.py
```diff
@@ -212,12 +212,20 @@ def name(self, new_name):
...
+                    pass
+            if not rsymbol:
+                rsymbol = RoutineSymbol(new_name, NoType())
 
             self._name = new_name
             # Since the constructor can not mark methods as functions directly
             # the symbol will always start being NoType and must be updated
             # if a return_value type is provided.
-            self.symbol_table.add(RoutineSymbol(new_name, NoType()),
+            self.symbol_table.add(rsymbol,
```
arporter, line 228: Could be one line now?

src/psyclone/psyir/nodes/routine.py
arporter: This will also need a TODO for the new issue you've created about correcting the properties in the other direction.

arporter (changes requested):
Thanks Aidan.
A comment will really help me understand what your new addition to `routine.py` is doing. Aside from that there's just a bit of tidying to do.
I've fired off the integration tests as they are now working again for NEMO and compilation.
Ref. guide still builds fine.

arporter:
Just a note to record that the NEMO and compilation integration tests were fine for this PR. (LFRic tests are currently not working.)

src/psyclone/tests/psyir/backend/fortran_routine_test.py
LonelyCat124: Fixed all of these.

src/psyclone/psyir/nodes/routine.py
LonelyCat124: Both comments added.

src/psyclone/psyir/nodes/routine.py
LonelyCat124: Done

arporter (approved):
All requested changes have been made. Integration tests are all green.
Will proceed to merge.

PR #2570, opened 2024-05-03
(Closes #2568) Adds a routine_symbol property to Routine and tests.

### LINKED ISSUE #2568: [PSyIR] routine_symbol property
For convenience, I suggest adding a `routine_symbol` property in `Routine` to act as a getter of `Routine(...).symbol_table.lookup_with_tag("own_routine_symbol")`, which is a bit long.

I do not think a setter is needed, as: 1/ it's already dealt with by the `name` setter, 2/ `Routine.create` takes a `string` name and a not a `RoutineSymbol` instance as argument.

JulienRemy:


sergisiso:
> Is the non required -0.01% codecov/project drop an issue? It's been reporting this on my 3 PRs.

No, in general we require a 100% of the patch (new code). Codecov has been acting strangely lately, so unless it lists something under the Codecov "indirect changes" (which this PR does not), don't worry about it.

LonelyCat124:
Hi @JulienRemy - Can I request a quick test to check I'm not seeing some different behaviour with this - I think this is a quirk of how we initialise things in fparser/PSyIR/PSyclone.

The expected result from your call should be (one of ) the `RoutineSymbol`  corresponding to the Routine in question, but I think if you try:
```
def test_my_test(fortran_reader):
   code = '''
    module test
    contains
   real function x(a)
     real :: a
     x =a 
    end function
end module'''

assert isinstance(container.children[0].children[0].symbol_table.lookup_with_tag("own_routine_symbol"), RoutineSymbol)
```

this will fail, as the "own_routine_symbol" is changed and has become a DataSymbol of the type of the function, and the only place the RoutineSymbol will still exist is in the parent `Container` (and at the moment, only if the parent `Container` is not a `FileContainer`, though I have a fix for that in progress).

LonelyCat124:
> In my mind (as things stand right now) the symbol tagged "own_routine_symbol" should probably be a RoutineSymbol indeed?

I agree here, and think the `routine_symbol` property is useful and should return this.

My PRs don't plan to change the Routine's own RoutineSymbol - but 2566 simply ensures certain properties are correctly given the the symbol in the Routine that were previously getting lost, and 2582 should just add a way to find the Routines contained within a FileContainer (similar to how it currently works for Module).

I assume the `functions` symbols become a `DataSymbol` because when we do `my_func = vlaue` we create a `Reference` to it we usually expect a DataSymbol, but I don't see a requirement as to why it couldn't be a RoutineSymbol, but there may be sections of the code that break. This decision is done in fparser2.py in the frontend which I'm a bit unclear on some parts of.

@sergisiso @arporter Do either of you know why function's own symbols is a DataSymbol rather than a typed RoutineSymbol?

arporter:
Am I right in thinking that #2596 will make this PR redundant?

sergisiso:
Yes, that PR has a `routine.symbol` similar to the helper methos proposed here. Closing this PR

PR #2571, opened 2024-05-03
(Closes #2569) update integration workflows to use python 3.12.3 and nvfortran 24.3

### LINKED ISSUE #2569: Update integration tests to python 3.12.3
Due to an OS upgrade on the runner, we need to update the version of python used in the integration tests.

arporter:
This is required to get the integration tests working again. Just an update from 3.12.2 to 3.12.3.
Also update the version of nvfortran used in the compilation tests to 24.3 (from 24.1).


arporter:
It turns out (on at least the system running the integration tests), that somehow system libraries were overriding those we expected to be found in `LIBRARY_PATH`. This meant that we were getting *different* versions of libnetcdf and libnetcdff. I've therefore extended the relevant example makefiles to also make use of `nc-config --libs`.

nmnobre:
The tests for NEMO are failing because we are passing `MPI_INC_DIR` (which depends on `NVHPC_ROOT`) to PSyclone before we load the Nvidia HPC SDK. Should be a simple matter of moving the make call down, _after_ we load the modules.
It was probably working before due to a forgotten `NVHPC_ROOT` somewhere in the runner's environment settings.

arporter:
> The tests for NEMO are failing because we are passing `MPI_INC_DIR` (which depends on `NVHPC_ROOT`) to PSyclone before we load the Nvidia HPC SDK. Should be a simple matter of moving the make call down, _after_ we load the modules. It was probably working before due to a forgotten `NVHPC_ROOT` somewhere in the runner's environment settings.

Thanks Nuno. Previously I had accidentally started the runner while the NVIDIA module was loaded. This time I've made sure not to do that so there's no NVHPC_ROOT set. I've updated the workflow.

arporter:
The passthrough test for LFRic is now failing because there's a difference in the results:
image
Given that this build is using a Spack environment, I'm not sure what can have caused this change but perhaps @sergisiso has some ideas?

nmnobre:
But otherwise, it seems everything else is working. Long live Glados. :)

arporter:
I've added a `KGOs` directory to `examples/lfric/scripts` and put the namelist and expected output file in there.

arporter:
Unfortunately, now the *transformed* version of LFRic is failing the KGO test :*-(

arporter:
Given that the compilation tests are working, this seems unlikely to be a problem with the fact that Glados now has the latest Python.

arporter:
I can reproduce the diffs on my laptop. They appear before the first timestep:
image

arporter:
I've used the `global.py` script under `meto-spice` and gradually cut it back until it does nothing. This shows that it's the application of colouring that changes the answers!

arporter:
I've moved forward to a more recent revision (49565) of LFRic (just prior to the split) and with that version, the addition of colouring does not break the test. In fact, the full meto-spice/global.py doesn't break the test either. I need to check with the `everything_everywhere...` script.

arporter:
Although this doesn't fix the problems with the LFRic tests, it does fix the compilation and NEMO tests so this is worth having. Ready for review from @sergisiso or @hiker.

sergisiso (approved):
I agree with @arporter that this is worth having to restore part of the integration tests. All the changes look good. This is ready to merge.

Issue #1298, opened 2021-06-21
ContainerSymbols should sometimes have a LocalInterface


sergisiso:
Currently ContainerSymbol gives by default a FortranModuleInterface to its interface values. This expects it to be an external container that can be imported when needed. However, we don't currently account for locally declared containers (I assume we just give them the wrong Interface and never use the import function)

Also the "_reference" attribute needs to be properly populated for Local ContainerSymbols so that it points to the definition.

arporter:
Hi @sergisiso, is this fixed now that you've updated the way interfaces are handled?

sergisiso:
Yes. Closing the issue.

PR #2575, opened 2024-05-08
(Towards #2201) Implementation enabling FileContainer's to have information about routines outside of modules.

LonelyCat124:
Likely missing coverage and FAILED psyir/backend/fortran_test.py::test_fw_filecontainer_error1  test fails.

Also waiting on #2566 for fixing some tests before this will be ready.


LonelyCat124:
This is ready for a look now - I think its a pretty small change overall.

LonelyCat124:
Changed my mind, there's some missing coverage (which i need to decide if its reachable) and I think I will also fix #2582 here.

arporter:
> Changed my mind, there's some missing coverage (which i need to decide if its reachable) and I think I will also fix #2582 here.

I think we should fix #2582 by adding `Routine._symbol` (unless I'm missing something?).

LonelyCat124:
I think adding `Routine._symbol` is a bigger change but maybe thats a better place to solve #2582 than doing it effectively twice.

LonelyCat124:
Ok - I checked this now @sergisiso.
We can't pass a subroutine to either `psyir_from_expression` or `psyir_from_statement`, they aren't valid for those entry points so I think we will always have a `Container` above a subroutine, so I'm going to remove the try/except from here:
```
try:

routine_symbol = routine.symbol_table.lookup(routine.name)
routine_symbol.datatype = base_type
except KeyError:
 pass
```
as I think we will always have that routine.name in a container.

LonelyCat124:
Should now be ready for review.

arporter:
Are we best delaying this now until #2596 is done @LonelyCat124 ?

src/psyclone/tests/psyir/backend/fortran_routine_test.py
```diff
@@ -459,16 +459,12 @@ def test_fw_routine_prefixes(fortran_reader, fortran_writer):
...
-    # Handle xfail case when prefixed subroutine isn't in a module.
     code = '''impure elemental subroutine sub()
     end subroutine sub'''
     container = fortran_reader.psyir_from_source(code)
     output = fortran_writer(container)
-    assert "impure elemental" not in output
-    pytest.xfail(reason="Issues #2156 & #2201: Routine prefixes aren't "
-                        "handled by the fparser frontend when the Routine "
-                        "isn't contained inside a module.")
+    assert "impure elemental subroutine sub" in output
```
sergisiso, line 470: Thanks for fixing this!

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
```diff
@@ -77,3 +79,27 @@ def test_program_handler(parser):
...
+    ''' Test that the FileContainer contains the relevant RoutineSymbol
+    nodes when we have no containing module.'''
+    code = '''
+    subroutine a()
+    end subroutine
+    real function b(val)
+       real :: val
+       b = val
+    end function
+    '''
```
sergisiso, line 94: Can you add test for the following cases (no worries if some done work - they can be xfail, I am just interested to know where we are with this complex cases) :
- Combine different type of program units:
```
program test
end program test
module A
end module A
subroutine mysub
end subroutine mysub
```
- Have non-named programs:
```
subroutine mysub(a)
   integer, intent(inout) :: a
   a = a + 1
end subroutine
end
```
This is valid fortran, note the seconds end, here mysub is not directly in the FileContainer, but in a program with no name.
Similar to the previous case but with variable declaration in the program, this is also valid Fortran:
```
integer :: b, c
contains
subroutine mysub(a)
   integer, intent(inout) :: a
   a = a + 1
end subroutine
end
```

sergisiso (changes requested):
@LonelyCat124 I would be happy to merge this before finishing #2596 as this is easier and it fixes missing keywords (that nemo compilation has warnings about them) and will allow fixing the function statements in Socrates. See inline comments mostly regarding what happens when there is other things that subroutines in the FileContainer.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5286,6 +5286,12 @@ def _subroutine_handler(self, node, parent):
                 f"PSyclone does not support routines that contain one or more "
                 f"ENTRY statements but found '{entry_stmts[0]}'")
 
+        # If the parent of this subroutine is a FileContainer, then we need
+        # to create its symbol and store it there. No visibility information
+        # is available since we're not contained in module.
+        if isinstance(parent, FileContainer):
+            _process_routine_symbols(node, parent.symbol_table, {})
+
```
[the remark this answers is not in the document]
LonelyCat124: I hadn't thought about this at all for modules, since my assumption had been that modules in files currently just "work", but I can look into this as well - do you want this as part of this PR or can we leave it until later?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5386,11 +5392,8 @@ def _subroutine_handler(self, node, parent):
                 # declared with its own local name. Be wary that this
                 # function may not be referenced so there might not be
                 # a RoutineSymbol.
-                try:
-                    routine_symbol = routine.symbol_table.lookup(routine.name)
-                    routine_symbol.datatype = base_type
-                except KeyError:
-                    pass
```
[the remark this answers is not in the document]
LonelyCat124: Yes, I think the KeyError cannot happen any more because the Routine will always have a container which can contain the symbol - previously this would fail on `FileContainer`s but I don't think it will any more.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1055,16 +1055,18 @@ def filecontainer_node(self, node):
...
             with is_program set to True.
 
         '''
         if node.symbol_table.symbols:
-            raise VisitorError(
-                f"In the Fortran backend, a file container should not have "
-                f"any symbols associated with it, but found "
-                f"{len(node.symbol_table.symbols)}.")
+            for symbol in node.symbol_table.symbols:
+                if not isinstance(symbol, RoutineSymbol):
```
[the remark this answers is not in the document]
LonelyCat124: @sergisiso this currently doesn't create a container symbol as you expected. My concern from reading the container symbol docs is that the container symbol feels more like it's representing a 'use' statement than anything else.
It's possible we could create a new module interface to use for this type of module declaration (instead of FortranModuleInterface) but I'm still unclear that if `mysub` used the declared module here that PSyclone would work correctly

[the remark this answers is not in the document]
LonelyCat124: I'm a bit confused here as #1298 is closed however there is a TODO in Container symbol for 1298 @sergisiso

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
LonelyCat124: All of these are added as failing tests. The latter 2 produce Code blocks and the first produces the correct objects but no symbols for anything other than the subroutine

LonelyCat124: Xfailing

src/psyclone/psyir/backend/fortran.py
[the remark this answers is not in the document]
sergisiso: Ok, I will have a look, but in any case it seems this needs a bit of work, so no need to do it in this PR.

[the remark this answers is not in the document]
LonelyCat124: Ok - I changed title to "towards" 2201 as we still have some remaining issues.

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
LonelyCat124: Added these now.

LonelyCat124:
Ready for another look

sergisiso:
@LonelyCat124 I put this back with reviewed with actions to clarify if the last commit after your message that this was ready belongs here, because the same commit is also in your Routine.symbol PR

LonelyCat124:
@sergisiso Sorry i forgot to check this yesterday. I don't think that commit is on this branch, its just mentioned in the commit message.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1055,16 +1055,18 @@ def filecontainer_node(self, node):
         :rtype: str
 
         :raises VisitorError: if the attached symbol table contains \
-            any data symbols.
+            any non-routine symbols.
```
sergisiso, line 1058: Can you remove the backslashes in this docstring?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1055,16 +1055,18 @@ def filecontainer_node(self, node):
...
         :raises VisitorError: if more than one child is a Routine Node \
             with is_program set to True.
 
         '''
         if node.symbol_table.symbols:
```
sergisiso, line 1063: This if line is now unecessary

src/psyclone/psyir/backend/fortran.py
sergisiso, line 1065: Can you add a TODO here then (probably to the same issue 2201 since this a towards PR), saying that ContainerSymbols should also be accepted but currently this are stored to its containing scope.

src/psyclone/psyir/frontend/fparser2.py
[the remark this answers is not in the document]
sergisiso: Then you should remove the second part of the comment :)

(Anyway your other PR will maybe get rid of this altogether)

sergisiso (changes requested):
Sorry about the commit confusing @LonelyCat124 , there is just minor things left now.

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Done

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Added

sergisiso (approved):
@LonelyCat124 All issues have been resolved, and now NEMOv5 is tested again. This is ready to merge. The integration tests were successful but slow probably because the workstation was busy.

PR #2581, opened 2024-05-09
(Closes #2576) Truncated the VMware error message to avoid the broken link

### LINKED ISSUE #2576: VMware knowledge base has moved, resulting in markdown link check failing for every PR
This fails in `tutorial/practicals/README.md` at  ` https://kb.vmware.com/kb/76918 `

I think the new link is https://knowledge.broadcom.com/external/article?articleNumber=315420 but I'm not sure as I never saw the old one.

Feel free to delete if we already have an issue about this.

LonelyCat124:
Documentation fix to avoid the broken link in the practicals - removed that section of the error message and reworded the docs.


arporter (approved):
Small change that removes a now dead link from an example VMWare error message. I've taken the liberty of fixing a typo and apart from that, all is fine. Link tests now pass :-)

PR #2609, opened 2024-06-10
(Closes #2407) acc routine clauses

### LINKED ISSUE #2407: Support parallelism clause on `ACC routine` directive.
image
where
image
This issue will allow `gang`, `worker` or `vector` to be specified when creating the directive.

arporter:
This is a small fix that is required because more recent versions of the OpenACC standard mandate that an `ACC routine` have a clause specifying the level of parallelism. This PR just adds a suitable property to the `ACCRoutineDirective` class and defaults it to 'seq'. It also expands the `ACCRoutineTrans` class to allow the level of parallelism to be specified as an option to the `apply` method.

I did try moving `ACCRoutineTrans` to `psyclone.psyir.transformations` but that caused circular dependencies that probably require the `MarkRoutineForGPUMixin` class to be moved too.


arporter:
This is ready for a first look now from either @hiker or @sergisiso.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -144,20 +144,67 @@ class ACCStandaloneDirective(ACCDirective, StandaloneDirective,
...
+    '''
+    Class representing an "ACC routine" OpenACC directive in PSyIR.
+
+    :param str parallelism: the level of parallelism in the routine, one of
+        "gang", "worker", "vector", "seq". Defaults to "seq".
+
+    '''
+    SUPPORTED_PARALLELISM = ["seq", "vector", "worker", "gang"]
+
+    def __init__(self, parallelism=None, **kwargs):
```
sergisiso, line 156: Any reason not to use `parallelism="seq"` here for the default value?
- We could then remove the first if condition of the parallelism.setter
- the documented type will match without needing the Optional[]

src/psyclone/transformations.py
```diff
@@ -2563,7 +2562,10 @@ def apply(self, node, options=None):
         for child in routine.children:
             if isinstance(child, ACCRoutineDirective):
                 return  # The routine is already marked with ACCRoutine
-        routine.children.insert(0, ACCRoutineDirective())
+
+        para = options.get("parallelism", None) if options else None
```
sergisiso, line 2566: None -> "seq" if you do the change above

sergisiso (changes requested):
@arporter There is just a minor comment asking if we can have the default value explicit in the constructor signature. Everything else looks good. All tests and coverage pass and there is not need for new documentation.

src/psyclone/psyir/nodes/acc_directives.py
arporter: Oh yes, done.

src/psyclone/transformations.py
arporter: Done.

arporter:
CI-happiness permitting, this is ready for another look.

sergisiso (approved):
Thanks @arporter, all looks good now. I will merge it after #2625

PR #2611, opened 2024-06-10
Switch to HPC-X from OpenMPI w/ the Nvidia HPC SDK

nmnobre:


sergisiso (approved):
@nmnobre the integration test succeeded, I re-launched again because I am not sure it was before the latest commit, but if all goes well this is ready to merge.

PR #2613, opened 2024-06-10
Fix OpenMPI while testing HPC-X

nmnobre:
This needs to be merged asap to prevent CI tests from failing.


sergisiso:
@nmnobre Has there been a change to the nvhpc-23.7 module? Or why was this not a problem until now?

sergisiso:
@nmnobre Another question, the HPC-X PR integration test seems to be running fine. Then, do we need this PR or can we skip directly to merge that one?

nmnobre:
> @nmnobre Another question, the HPC-X PR integration test seems to be running fine. Then, do we need this PR or can we skip directly to merge that one?

Oh! Yes, this is unexpected! Then we can move to HPC-X today, let me just fix the module name, sorry for the disruption.

PR #2616, opened 2024-06-18
(Closes #2125) bugfix for InlineTrans when formal arg dimensioned by another formal arg.

### LINKED ISSUE #2125: Bug in InlineTrans when formal argument declaration references another formal argument
If we have:

    subroutine sub(var, start)
      real, dimension(start:) :: var
      var(start+1) = 5.0
    end subroutine

Then the inlined code still contains a reference to 'start' in the array-index expression that is generated to allow for non-default starting index values.

arporter:


arporter:
Current failure is in psyad eg1. On closer investigation, it turns out that the dimensioning symbol ('npts') in the PSyIR of the TL routine is not the same symbol as the routine argument 'npts'! Obviously, it should be.

arporter:
Closing this PR as I'll open a new one with a version of the branch where I haven't messed with ArrayBounds (I thought I could simplify things by having it support a `copy` method but that actually resulted in quite a lot more code).

PR #2625, opened 2024-06-20
(Closes #2619) Add an `OMPParallelDoDirective.create` static method.

### LINKED ISSUE #2619: Make OMPParallelDirective.create() a class method
Julien reports:

There's no OMPParallelDoDirective.create reimplementation, so it uses its parent OMPParallelDirective.create and the created instance is of the parent class. Making the first a classmethod and calling cls(...) rather than OMPParallelDirective() would fix it.

JulienRemy:
`OMPParallelDoDirective` currently inherits its parent (`OMPParallelDirective`) `create` static method, which leads to the creation of an instance of it.
After further inspection, making `OMPParallelDirective.create` a class method would not cover all attributes (inherited from `OMPDoDirective`) so this reimplements `create` in the derived class instead.

### DIFF
--- a/src/psyclone/psyir/nodes/omp_directives.py
+++ b/src/psyclone/psyir/nodes/omp_directives.py
@@ -2198,6 +2198,22 @@ class OMPParallelDoDirective(OMPParallelDirective, OMPDoDirective):
         OMPDoDirective.__init__(self, **kwargs)
         self.addchild(OMPDefaultClause(
             clause_type=OMPDefaultClause.DefaultClauseTypes.SHARED))
+        
+    @staticmethod
+    def create(children=None, **kwargs):
+        '''
+        Create an OMPParallelDoDirective.
+
+        :param children: The child nodes of the new directive.
+        :type children: List of :py:class:`psyclone.psyir.nodes.Node`
+        :param kwargs: additional keyword arguments provided to the PSyIR node.
+        :type kwargs: unwrapped dict.
+
+        :returns: A new OMPParallelDoDirective.
+        :rtype: :py:class:`psyclone.psyir.nodes.OMPParallelDoDirective`
+        '''
+
+        return OMPParallelDoDirective(children=children, **kwargs)
 
     @staticmethod
     def _validate_child(position, child):
--- a/src/psyclone/psyir/nodes/omp_directives.py
+++ b/src/psyclone/psyir/nodes/omp_directives.py
@@ -2198,7 +2198,7 @@ class OMPParallelDoDirective(OMPParallelDirective, OMPDoDirective):
         OMPDoDirective.__init__(self, **kwargs)
         self.addchild(OMPDefaultClause(
             clause_type=OMPDefaultClause.DefaultClauseTypes.SHARED))
-        
+
     @staticmethod
     def create(children=None, **kwargs):
         '''
--- a/src/psyclone/tests/psyir/nodes/omp_directives_test.py
+++ b/src/psyclone/tests/psyir/nodes/omp_directives_test.py
@@ -525,6 +525,23 @@ def test_omp_do_directive_validate_global_constraints(fortran_reader,
             in str(err.value))
 
 
+def test_omp_parallel_do_create():
+    ''' Test the OMPParallelDoDirective create method. '''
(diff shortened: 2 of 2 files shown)


JulienRemy:
@arporter @sergisiso This passed tests and codecov checks, what do you think?

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2199,6 +2199,22 @@ def __init__(self, **kwargs):
         self.addchild(OMPDefaultClause(
             clause_type=OMPDefaultClause.DefaultClauseTypes.SHARED))
 
+    @staticmethod
+    def create(children=None, **kwargs):
```
LonelyCat124, line 2203: Can we make this a class method? There is the `OMPTeamDistributeParallelDoDirective` that inherits from (only) `OMPParallelDoDirective` and this implementation will repeat this issue for that class. 
That class is identical to this directive apart from _directive_string so we shouldn't have the same issue.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2199,6 +2199,22 @@ def __init__(self, **kwargs):
...
+        '''
+        Create an OMPParallelDoDirective.
+
+        :param children: The child nodes of the new directive.
+        :type children: List of :py:class:`psyclone.psyir.nodes.Node`
+        :param kwargs: additional keyword arguments provided to the PSyIR node.
+        :type kwargs: unwrapped dict.
```
LonelyCat124, line 2210: Can you add documentation for the OMPDoDirective keyword arguments into the kwargs (omp_schedule, collapse and reprod).

Also could you help us and mark the omp_schedule argument in OMPDoDirective as optional as well.

LonelyCat124 (changes requested):
Hi @JulienRemy 

The implementation is mostly ok - I've asked if you can change the create on `OMPParallelDoDirective` to be a class method as it has a subclass. Otherwise, an expansion on documentation would be helpful, and please add yourself to the author list on both files.

src/psyclone/psyir/nodes/omp_directives.py
JulienRemy: Note that this reimplements `OMPParallelDirective.create` static method by a class method in the derived subclass which is maybe slightly unpythonic. Should I make the parent one a class method too?

src/psyclone/psyir/nodes/omp_directives.py
JulienRemy: I added these using a list-like format as I'm not sure if you have a specification for listing `kwargs` in docstrings, let me know if this is fine or not.

LonelyCat124: Apologies - please remove these after all.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: Yes - it doesn't look like that should have any other side effects so thats probably best.

src/psyclone/psyir/nodes/omp_directives.py
JulienRemy: No problem.

LonelyCat124: One other thing on docs I forgot to mention apologies. The `\` to signal a newline in the docs we no longer use - can you fix this here and in the OMPDoDirective docs that you fixed the indentation in? Then I'll submit this for integration tests.

LonelyCat124 (commented):
@JulienRemy One more tiny doc fix and we're good to go I think

src/psyclone/psyir/nodes/omp_directives.py
JulienRemy: I took the opportunity to get rid of them in all docstrings in this file and also to indent all of them correctly. Apologies for the longer review.

LonelyCat124:
I've reset the integration tests again since the NEMO 5 ones don't work so waiting on the results of those but should be fine.

LonelyCat124:
Integration tests were all fine so I think this is ready to merge now.

LonelyCat124 (approved):

PR #2645, opened 2024-07-03
(Closes #2585) Add an optional, 'default' argument to SymbolTable.lookup()

### LINKED ISSUE #2585: Extend SymbolTable.lookup() with a "default value" argument to avoid need for try...except.
Currently the `lookup` method of SymbolTable will raise a `KeyError` if a Symbol cannot be found (c.f. a Python dict). However, this leads to a lot of try..except blocks which could be avoided if we could just tell the method to return e.g. `None` if no Symbol is found. By adding an optional argument we can make this change backwards-compatible (which is good because it would otherwise touch a lot of code).

arporter:


arporter:
Just adds an optional, 'default' argument to the lookup() method. One for @sergisiso, @hiker or @LonelyCat124.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -56,6 +56,19 @@
 from psyclone.psyir.symbols.typed_symbol import TypedSymbol
 
 
+class _Default:
```
sergisiso, line 59: I don't think this class is necessary, the implementation below also passes the tests:
```
-class _Default:
-    '''
-    A mutable class used to provide a unique default value to methods within
-    the SymbolTable class. This enables us to determine when the user has
-    supplied a value of 'None' for arguments that have a default value.
-
-    :param Any value: a value associated with this instance.
-
-    '''
-    def __init__(self, value):
-        self.value = value
-
+DEFAULT_SENTINEL = object()
```

```
     def lookup(self, name, visibility=None, scope_limit=None,
-               default=_Default(1)):
+               default=DEFAULT_SENTINEL):
```
```
-            if default is SymbolTable.lookup.__defaults__[-1]:
+            if default is DEFAULT_SENTINEL:
```

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -933,32 +946,35 @@ def specify_argument_list(self, argument_symbols):
...
+        :param scope_limit: optional Node which limits the symbol
+            search space to the symbol tables of the nodes within the
+            given scope. If it is None (the default), the whole
+            scope (all symbol tables in ancestor nodes) is searched
+            otherwise ancestors of the scope_limit node are not
             searched.
-        :type scope_limit: :py:class:`psyclone.psyir.nodes.Node` or \
-            `NoneType`
+        :type scope_limit: Optional[:py:class:`psyclone.psyir.nodes.Node`]
+        :param Any default: an optional quantity to return if the named symbol
```
sergisiso, line 966: Maybe s/quantity/object/ or "value". An add "if not provided a KeyError is raised."?

sergisiso (changes requested):
Thanks for adding this @arporter. I think this could be implemented a bit simpler, see inline.

src/psyclone/psyir/symbols/symbol_table.py
arporter: Oh yes, updated.

src/psyclone/psyir/symbols/symbol_table.py
arporter: Embarrassing! The dangers of stack overflow :-)

arporter:
Dev. guide still builds fine. CI happiness permitting, this is ready for another look.

sergisiso (changes requested):
The implementation looks good now, I could have approved but I have a couple of questions to ask.

The docs build and the integration test is not necessary as the new feature is not yet used in scripts. This should be ready next time regardless of the responses.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -992,8 +1003,11 @@ def lookup(self, name, visibility=None, scope_limit=None):
                         f" match with the requested visibility: {vis_names}")
             return symbol
         except KeyError as err:
-            raise KeyError(f"Could not find '{name}' in the Symbol Table.") \
-                from err
+            if default is SymbolTable.lookup.__defaults__[-1]:
```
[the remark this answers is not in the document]
arporter: Oops, no, that was just an oversight.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -933,32 +939,37 @@ def specify_argument_list(self, argument_symbols):
         self._validate_arg_list(argument_symbols)
         self._argument_list = argument_symbols[:]
 
-    def lookup(self, name, visibility=None, scope_limit=None):
+    def lookup(self, name, visibility=None, scope_limit=None,
+               default=DEFAULT_SENTINEL):
```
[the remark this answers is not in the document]
arporter: Ah, I was under the mistaken impression that `default` was what it was called in `dict.get()`. I agree that `value` is a rubbish name. I've gone for `otherwise`.

arporter:
CI-happiness permitting, this is ready for another look.

sergisiso (approved):
@arporter All suggestions addressed and the CI is green, this is ready to merge.

PR #2649, opened 2024-07-05
(Towards #2586) Add support for object-oriented `var =[>] this[%...]%procedure(...)` assignment

### LINKED ISSUE #2586: Add support for Calls to StructureType members
It seems that Fortran functions are not supported by PsyIR.

Please see [test_a.zip](https://github.com/stfc/PSyclone/files/15292847/test_a.zip) which creates CodeBlocks for 
fparser.two.Fortran2003.Function_Reference
fparser.two.Fortran2003.Function_Reference
fparser.two.Fortran2003.Structure_Constructor

This output has been created with branch https://github.com/stfc/PSyclone/compare/2577_support_pointer_assignments

This might be depending on #2577.

JulienRemy:
Currently pointer assignments of the form `var =[>] this[%...]%procedure(...)` are `CodeBlock`s wrapping fparser `Function_Reference` objects.
This converts them to assignments with a call on the rhs.

Notes:
- a workaround was required to deal with some fparser `Intrinsic_Function_Reference`, wrongly parsed as `Function_Reference` [fparser/#447](https://github.com/stfc/fparser/issues/447) and associated TODO comments.
- this does not deal with the `Structure_Constructor` case mentioned [here](https://github.com/stfc/PSyclone/issues/2586#issuecomment-2109675980) so it does not close #2586.


arporter:
Hi @JulienRemy, I can't remember what you said about PRs on Monday. Was this the one that's ready for a review?

JulienRemy:
Hi @arporter, this one is indeed ready to review when you have time for this.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5143,6 +5144,13 @@ def _call_handler(self, node, parent):
 
         '''
         call = Call(parent=parent)
+        # TODO fparser/#447 For now `null()` is treated as a
+        # `Function_Reference` by fparser, instead of an
+        # `Intrinsic_Function_Reference` so we redirect to the correct
+        # handler for fparser2 tests to pass. This should be removed once
+        # fparser2 is fixed.
+        if str(node.items[0]).lower() == "null":
```
arporter, line 5152: Just to be on the safe side, please could you also check here that the call has no arguments. (I did some experiments and gfortran did let me define my own function called `null()`, albeit it hanged during execution.)

arporter (changes requested):
Looks good, thanks Julien. Just one small refinement and then this can go on.
No need to update any docs as this is a bug fix/workaround.
I'll fire off the integration tests now.

src/psyclone/psyir/frontend/fparser2.py
JulienRemy: This is done!

arporter (approved):
Requested change has been made (thanks).
Integration tests were all green (although I'll re-run them).
Will proceed to merge.

arporter:
Integration tests were green.

PR #2653, opened 2024-07-08
Experimental backend for xDSL Python-Native Compiler Toolkit

mcjamieson:
Initial experimental xDSL backend for PSyclone. The ExCALIBUR xDSL project provides a Python-Native Compiler Toolkit for MLIR.


sergisiso (approved):
@mcjamieson I am approving this for merging, note that since it goes under examples it does not need to pass our testing/codecoverage and styling, but also there won't be any guarantees that future psyclone modifications break it.

If at some point you want to bring it inside the backends it will need to follow https://psyclone-dev.readthedocs.io/en/latest/coding-style.html , have complete testing code coverage and the dependencies that it uses should be all in pypi.

PR #2070, opened 2023-03-14
Add ACCWaitDirective and async keyword on kernels, enter data, parallel (fixed #1664)

svalat:
While working on CROCO I had to make a try to reach performance of their manual ACC version by using the async/wait ACC semantic so I made the patch which by the way might solve issue #1664.

**Done**:

 * Add ACCWaitDirective
 * Patch ACCKernelDriective, ACCParallelDirective and AccEnterDataDir to add support for async keyword
 * Patch ACCKernelsTrans to support the async option.

**Questions**:

 * I made a commit for each, don't know if you prefer to split in a PR for each.
 * In the Trans, I currently forbid nesting async with different streams, don't know if it is too strict or not ?
 * We could also have introduced an intermediate class ACCAsync to inherit from for the ACCKernelDirective, ACCParallelDirective and ACCEnterDataDir which might help for latter tree analysis.

PR #2664, opened 2024-07-18
(Closes #1664) Add ACCWaitDirective and async keyword on kernels, enter data, parallel

### LINKED ISSUE #1664: Add support for the `async` clause for OpenACC PARALLEL and KERNELS directives.
Alex and Victoria have shown that there are performance benefits to be had by using asynchronous kernel launches, even if the kernels themselves have to be run in order. This is also bourne out by this Stackoverflow post: https://stackoverflow.com/questions/54355791/reducing-time-to-launch-kernels-in-time-stepping-loop-openacc.

An example of code using asynch is:

    !$acc kernels async(1) 
    !$acc loop independent 
    do ... 
    
    !$acc end kernels 

    !$acc parallel async(1) 
    !$acc loop independent 
    do ... 

    !$acc end parallel 

    !$acc wait 

where the `1` here means despatch all kernels to that command queue.

Presumably, OpenMP offload offers the same functionality?

arporter:
This is a replacement/takeover of #2070 as Sebastien has moved to a new job.


arporter:
Now that I look at this again, there don't seem to be any tests of the new directive and keyword for the Fortran psyir backend (it's all gen_code in LFRic).

arporter:
I see from https://github.com/stfc/PSyclone/pull/2070#discussion_r1319986879 that this PR was started before we had ACCClause (#2157). However, we do now have clauses and therefore this code needs updating (to remove the use of `Signature`) and tests for the Fortran backend are required.

schreiberx:
@arporter
1. I worked on the code coverage. Let's see what the next push brings.
2. There was something in the code that I didn't understand:
`async_queue` was never to `None`, but then checked against `None`.
But I found places where `async_queue` was set to `False` if the option didn't exist. I changed this throughout the code that `async_queue` is set per default to `None` and to an integer otherwise.
3. Regarding the options dictionary, I set the default argument value to {}. This avoids further checks below ala `option is None`.

schreiberx:
> @arporter
> 
>     1. I worked on the code coverage. Let's see what the next push brings.
> 
>     2. There was something in the code that I didn't understand:
>        `async_queue` was never to `None`, but then checked against `None`.
>        But I found places where `async_queue` was set to `False` if the option didn't exist. I changed this throughout the code that `async_queue` is set per default to `None` and to an integer otherwise.
> 
>     3. Regarding the options dictionary, I set the default argument value to {}. This avoids further checks below ala `option is None`.

I updated the async part in the previous push.
`async_queue` can be finally set to these values:
- `None` or `False` which doesn't make use of it
- `True` which uses it without specifying a concrete queue
- An integer or a Reference as the concrete queue to use.

src/psyclone/psyir/nodes/acc_mixins.py
```diff
@@ -0,0 +1,132 @@
...
+        :raises TypeError: if `wait_queue` is of the wrong type
+        '''
+        # check
+        if (async_queue is not None and
+           not isinstance(async_queue, (bool, Reference, int))):
+            raise TypeError(f"Invalid async_queue value, expect Reference or "
+                            f"integer or None or bool, got : {async_queue}")
+
+        # assign
+        self._async_queue = async_queue
```
sergisiso, line 94: Aha, I see that we store PSyIR in a private attribute, this could be problematic (e.g. the dependency analysis will not see it, the delete symbols will not check, ...). It is also not easy to resolve (by customising: copy/signal/update_symbols/ ...) because this is a Mixin. One alternative is to make it a Clause child instead of a Mixin, but that's a big change to the current implementation (another is make it a TODO with the limitations and come back to this later), what do you think @arporter ?

Also, should we convert int to Literal?

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -35,20 +35,26 @@
...
-    Reference, Return, Routine, Statement, WhileLoop)
+    ACCEnterDataDirective, ACCKernelsDirective, Assignment,
+    Call, CodeBlock, Loop, Node,
+    PSyDataNode, Reference, Return, Routine, Statement, WhileLoop)
 from psyclone.psyir.symbols import UnsupportedFortranType
 from psyclone.psyir.transformations.region_trans import RegionTrans
 from psyclone.psyir.transformations.transformation_error import (
     TransformationError)
 
+from typing import Any, Dict, List, Union
```
sergisiso, line 56: pylint wants this before third party libs

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -74,17 +80,18 @@ class ACCKernelsTrans(RegionTrans):
...
         Enclose the supplied list of PSyIR nodes within an OpenACC
         Kernels region.
 
         :param node: a node or list of nodes in the PSyIR to enclose.
-        :type node: :py:class:`psyclone.psyir.nodes.Node` |
-                    list[:py:class:`psyclone.psyir.nodes.Node`]
         :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
-        :param bool options["default_present"]: whether or not the kernels
+            "default_present": boolean whether or not the kernels
```
sergisiso, line 94: Should this still be `:param bool options["default_present"]:` as before?

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -99,32 +106,80 @@ def apply(self, node, options=None):
...
         Check that we can safely enclose the supplied node or list of nodes
         within OpenACC kernels ... end kernels directives.
 
         :param nodes: the proposed PSyIR node or nodes to enclose in the
                       kernels region.
-        :type nodes: (list of) :py:class:`psyclone.psyir.nodes.Node`
         :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
-        :param bool options["disable_loop_check"]: whether to disable the
+            "disable_loop_check": boolean whether to disable the
```
sergisiso, line 177: Same here and below, I am unsure which format is rendered more clearly in the docs.

src/psyclone/transformations.py
```diff
@@ -2496,6 +2498,11 @@ def apply(self, sched, options=None):
         :param sched: schedule to which to add an "enter data" directive.
         :type sched: sub-class of :py:class:`psyclone.psyir.nodes.Schedule`
         :param options: a dictionary with options for transformations.
+
+        The available options are :
+         - async_queue : Permit to force using the given async stream if
+                         not None.
```
sergisiso, line 2504: Use common format for option arguments.

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -35,6 +35,7 @@
 #         A. B. G. Chalk, STFC Daresbury Lab
 # Modified I. Kavcic, Met Office
 # Modified J. Henrichs, Bureau of Meteorology
+# Modified S. Valat, INRIA / LJK
```
sergisiso, line 38: Maybe delete since at this point there isn't modifications at this specific file (but there is a couple of actually modified files in this PR where the contributors' names could be added).

sergisiso (changes requested):
@arporter @schreiberx I don't know who is the owner of this PR but I added my review. My main concern is that it adds PSyIR nodes in a private attribute of a mixin, and this has been a source of issues in the past as these are easy to forget when doing tree modifications. I think we should consider making it a ACCClause so all the tree functionality works as expected.

src/psyclone/psyir/nodes/acc_mixins.py
arporter: Yes, I think this work was begun before we had `Clause`. I agree that we should move it over. I'll take this on.

src/psyclone/psyir/transformations/acc_kernels_trans.py
arporter: Done.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -202,42 +215,50 @@ def gen_code(self, parent):
...
+        variable Signature to say at runtime what stream to be
+        used.
+
     '''
-    def __init__(self, children=None, parent=None):
+    def __init__(
+                self,
+                children: List[Node] = None,
+                parent: Node = None,
+                async_queue: Union[bool, int, Reference, None] = None
```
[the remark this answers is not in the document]
arporter: I'm not sure they did but I've changed things now so that `True` means an `async` clause without an explicit queue while `False` means no `async` clause. I've removed the support for `None`.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -50,16 +52,23 @@
...
 from psyclone.psyir.nodes.routine import Routine
+from psyclone.psyir.nodes.reference import Reference
+from psyclone.psyir.nodes.acc_mixins import ACCAsyncMixin
 from psyclone.psyir.nodes.schedule import Schedule
 from psyclone.psyir.nodes.operation import BinaryOperation
 from psyclone.psyir.symbols import ScalarType
 
+from psyclone.f2pygen import BaseGen
+
+from typing import Dict, List, Optional, Set, Tuple, Union
```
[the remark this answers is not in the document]
arporter: Done, and f2pygen import merged with earlier instance.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -104,7 +114,12 @@ def validate_global_constraints(self):
                 f"region enclosed by an '{type(self).__name__}'")
 
     @property
-    def signatures(self):
+    def signatures(self) -> Union[
+                                Tuple[Set[Signature]],
+                                Tuple[
+                                    Set[Signature], Set[Signature]
+                                ]
```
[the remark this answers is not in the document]
arporter: Agreed.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -147,14 +159,15 @@ class ACCRoutineDirective(ACCStandaloneDirective):
...
+        "gang", "seq", "vector", "worker".
 
     '''
-    SUPPORTED_PARALLELISM = ["seq", "vector", "worker", "gang"]
+    SUPPORTED_PARALLELISM = ["gang", "seq", "vector", "worker"]
 
-    def __init__(self, parallelism="seq", **kwargs):
+    def __init__(self, parallelism: str = "seq", **kwargs):
         self.parallelism = parallelism
+        assert self.parallelism in self.SUPPORTED_PARALLELISM
```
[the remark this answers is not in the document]
arporter: Done.

src/psyclone/psyir/nodes/acc_directives.py
```diff
@@ -202,42 +215,50 @@ def gen_code(self, parent):
...
-    :type children: List[:py:class:`psyclone.psyir.nodes.Node`]
-    :param parent: the node in the InvokeSchedule to which to add this \
-                   directive as a child.
-    :type parent: :py:class:`psyclone.psyir.nodes.Node`
+    :param parent: the node in the InvokeSchedule to which to add this
+        directive as a child.
+    :param async_queue: Enable async support and attach it to the given queue.
+        Can use False to disable, True to enable on default
+        stream. Int to attach to the given stream ID or use a
+        variable Signature to say at runtime what stream to be
```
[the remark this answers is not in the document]
arporter: Done and moved to DataNode. Updated appropriate test to check this works.

src/psyclone/psyir/nodes/acc_mixins.py
```diff
@@ -0,0 +1,132 @@
...
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors S. Valat, Inria / Lab. Jean Kuntzmann
+# Modified M. Schreiber, Univ. Grenoble Alpes / Inria / Lab. Jean Kuntzmann
+# -----------------------------------------------------------------------------
+
+''' This module contains the mixins to apply some ACC features on many
+classes.'''
```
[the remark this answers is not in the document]
arporter: I'm fairly relaxed about having another file and as you say, the directives file is at 1.3K now. If we really care we could have an openacc submodule?

src/psyclone/psyir/nodes/acc_mixins.py
```diff
@@ -0,0 +1,132 @@
...
+
+import abc
+
+from psyclone.psyir.nodes.reference import Reference
+
+from typing import Union
```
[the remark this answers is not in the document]
arporter: Oh yes, done.

src/psyclone/psyir/nodes/acc_mixins.py
```diff
@@ -0,0 +1,132 @@
...
+                        Can use False to disable, True to enable on default
+                        stream. Use int to attach to the given stream ID or
+                        use a variable Reference to say at runtime what stream
+                        to be used.
+    '''
+    def __init__(
+                self,
+                async_queue: Union[bool, int, Reference, None] = None
+            ):
+        self.async_queue = async_queue
```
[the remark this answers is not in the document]
arporter: I don't believe we've disabled it and pylint still seems to support it so that's a mystery. In any case, I've got rid of this attribute now - there's an associated property that queries the tree instead.

src/psyclone/psyir/nodes/acc_mixins.py
arporter: This is now done.

src/psyclone/psyir/transformations/acc_kernels_trans.py
arporter: Yes it should. I've also put back the one for "async_queue".

src/psyclone/psyir/transformations/acc_kernels_trans.py
arporter: I've reverted to our original form.

src/psyclone/transformations.py
arporter: Done.

src/psyclone/tests/psyir/transformations/transformations_test.py
arporter: Gone now.

arporter:

sergisiso (approved):
@arporter Thanks for addressing all my comments. All tests are green and docstrings are good. This is ready to merge.

Issue #2670, opened 2024-07-23
Encapsulate TransformationError reporting functionality


sergisiso:
A couple of patterns have emerged to report TransformationErrors
- TransformationErrors often need the context of when they happen in the error message, the easy way to do it is with `node.debug_string()`, but since this invokes the backend it is a lot of work to create a string that will sometimes be just ignored. So we recommend using LazyString to delay the call to the visitor.
- TransformationErrors are often purposely ignored, but sometimes want to still keep track of the reason. In this cases we started adding it as a comment when the verbose option is provided.

So we end up with something like:
```
message = (f"My error message without view or debug_string")
if verbose:
    node.append_preceding_comment(message)
raise TransformationError(LazyString(
    lambda: f"{message}, but found:\n{node.debug_string()}"))
```
We could encapsulate this is a method so that it is a single-liner to call, but also because we could have the message syntax be more predictable, e.g. we may want to prefix all psyclone added comments with a specific identifier and in the future we may want to make the reporting configurable: e.g. comment/logfile/stderr, so it would be good to do in in a single place and not all over the code.

PR #2679, opened 2024-07-30
(closes #2676) Fix Nemo reproducibility issue

### LINKED ISSUE #2676: OpenMP version of NEMO is not run-to-run reproducible
As discovered in #2673: https://github.com/stfc/PSyclone/pull/2673#issuecomment-2253014754

Essentially, re-running the same binary on the same number of OMP threads can give different answers. (This is the 'ECMWF' version of NEMO in the integration tests.) We first saw this failure on 26/07/2024 at 10:08:13. Presumably, it is therefore due to (or exacerbated by) something that went onto master just before that.

sergisiso:


sergisiso:
@arporter This is ready to review. The integration tests all pass now.

src/psyclone/tests/psyir/transformations/parallel_loop_trans_test.py
```diff
@@ -103,9 +104,24 @@ def test_paralooptrans_validate_ignore_dependencies_for(fortran_reader,
...
+    # Test that the inner loop does not log again the same error message
+    with pytest.raises(TransformationError) as err:
+        trans.validate(loop.loop_body[0], options={"verbose": True})
+    assert ("PSyclone: Loop cannot be parallelised because the dependency"
+            not in loop.loop_body[0].preceding_comment)
+
+    # But if it's not already in an ancestor, it adds the comment there
+    loop.preceding_comment = ""
+    with pytest.raises(TransformationError) as err:
+        trans.validate(loop.loop_body[0], options={"verbose": True})
```
arporter, line 120: This has made me realise that we have slightly non-intuitive behaviour here - calling `validate` on a Node can actually modify it (in the sense that `preceding_comment` can be modified) if `verbose` is set. I don't think there's any easy way of avoiding this but I thought I would flag it.

examples/nemo/scripts/utils.py
```diff
@@ -300,6 +300,17 @@ def insert_explicit_loop_parallelism(
                 " 'nlay_i' or 'nlay_s'.")
             continue
 
+        # If we have a region_trans this is already tested
+        if not region_directive_trans:
+            not_pure = [call.routine.name for call in loop.walk(Call)
```
arporter, line 305: I'd be happier if we could put this check inside the appropriate transformation(s?) rather than in this utils.py. Is that possible?

arporter (changes requested):
Well done for tracking this down Sergi. I think it's fine but it would be better if we could put the functionality in the transformations themselves rather than `utils.py`, if that's possible?
Also, please could you tweak the `nemo_tests.yml` file so that it fails with a better error when the results don't match?
Integration tests were all green. I'll check the Ref. Guide next time.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -171,7 +171,15 @@ def validate(self, node, options=None):
...
                 if verbose:
-                    node.append_preceding_comment(f"PSyclone: {messages}")
+                    # This message can get quite long, we will skip it if an
+                    # ancestor loop already has the exact same message
+                    cursor = node.ancestor(Loop)
+                    while cursor:
+                        if messages in cursor.preceding_comment:
+                            break
+                        cursor = cursor.ancestor(Loop)
+                    if not cursor:
```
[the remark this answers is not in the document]
sergisiso: I didn't know while also has an else :) But I will leave it as it is.

src/psyclone/tests/psyir/transformations/parallel_loop_trans_test.py
sergisiso: This is the expected behaviour. Many verbose options now add comments during the validate. But I don't think this is a tree modification, just a logging mechanism using the preceding_comment mixing attribute.  I raised #2670 to formalise this pattern better and make it configurable: choose between comments/log/stderr.

examples/nemo/scripts/utils.py
sergisiso: Done, I didn't do it the first time because I wasn't sure if it would impact lfric, but in retrospective builtin and kern are not calls so it has no impact. Even if we change them to being calls, the pure attribute should be a requirement for them.

sergisiso:
@arporter CI and integration test-permitting (which I just triggered), this is ready for another review.

src/psyclone/tests/psyir/transformations/parallel_loop_trans_test.py
arporter: OK.

arporter (commented):
Changes look good. Thanks for improving the error handling in the integration tests.
Ref. guide builds fine.
Will proceed to merge.

examples/nemo/scripts/utils.py
arporter: That's a good idea (marking kernels as PURE). We should raise it with @TeranIvy and @christophermaynard once they're back.

arporter (approved):
Somehow I pressed the wrong button last time. This is ready to go!

arporter:
Just to say, I checked performance too and it was all fine.

Issue #1858, opened 2022-09-08
Support transforming reference to array to array reference for an array in a structure


rupertford:
The Reference2ArrayRangeTrans transformation replaces a reference to an array with an array reference. For example:

```
a = b
```
becomes

```
a(:) = b(:)
```

However this transformation raises an exception if we have a reference to an array within a structure i.e.

```
struct%a = b
```

sergisiso:
@arporter I started looking at #717 as I wanted to fix the "where" limitations next, but realised that for this we need a proper way to decide if a plain reference or structure is an scalar, an array (with implicit (:)) or it cannot be guaranteed. This is more than looking at the type because as in ArrayAssignment2Loop we can also infer what it is by looking at the syntax and knowing that it comes from a valid Fortran statement. And this is important because it means we can convert them to loops more often and provide more parallelism.

I was thinking where this logic should live as it can be complex and I want to reuse it, and I believe the Reference2ArrayReference is the right place (because we end up applying this transformation everytime we have this question as it resolves it in the syntax - after it Arrays are ArrayMixins and scalar are References), the only current problem is the "scalar" vs "cannot be guaranteed" is inconsistent.

But I think we can modify "validate" behaviour to let scalars pass the validation (and the apply unmodified), but guarantee that ALL cases that we cannot guarantee consistently fail the validation. So:

```
a = 1
struct%b = 1
```
If they are scalars the validate passes and the apply returns the references unmodified.
If they are arrays the validate pass and the apply returns:
```
a(:) = 1
struct(:)%b = 1 ! or "struct%b(:) = 1" depending of what is an array
```
If we don't have enough syntactic or type information to guarantee which one it is the validate produces a TransformationError:
so other transformations (ArrayAssignment2Loop, ...) can validate during their own validation and fail if necessary or the where produce a CodeBlock. But if the validation passes, we can continue this transformation/where logic knowing that all arrays are now ArrayMixins (which often already assume).

I will try to do this change as part of this issue and then go back to the "where" issues.

arporter:
I've just hit this again in #2685. Currently `_get_effective_shape` doesn't work correctly for e.g. `a%b%c(1)` if `b` is an array so I thought I'd try this transformation but it refuses. This is better than `_get_effective_shape` does - it simply says that the access is a scalar.

PR #2685, opened 2024-08-07
(Closes #2027) fix for ArrayMixin._effective_shape bug

### LINKED ISSUE #2027: Potential problem with array mixin _get_effective_shape
I was just adding tests, when I had crash. Cause:
```
    def _get_effective_shape(self):
...
       for idx_expr in self.indices:
            if isinstance(idx_expr, Range):
                shape.append(_num_elements(idx_expr))

            elif isinstance(idx_expr, Reference):
                dtype = idx_expr.datatype
                if dtype.shape:
   
```
This function is called from `array_reference.datatype`, with the content of `map_adspc1_field_a(:,cell)`.

When working on the 2nd index, `cell` is a reference, but it's not an array and has no shape, so the test `if dtype.shape` just crashes.

I can work around this problem (by using `array_reference.symbol.datatype` instead of `array_reference.datatype`, and that's ok for me - but I am still wondering if this should be fixed in the code to avoid the crash?

arporter:
We had a bug which would cause a crash if we had an array access like `a(b)` where `b` is actually an array.

In testing this, I also realised that the ArrayAssignment2LoopsTrans transformation would incorrectly convert something like:

    c(:) = a(b)

to

    do idx = 1, SIZE(c,1)
       c(idx) = a(b)
    end do

This should in fact be:

    do idx = 1, SIZE(c,1)
      c(idx) = a(b(idx))
    end do

This also means we have a bug in `examples/nemo/scripts/utils.py` because it should have converted `a(b)` to `a(b(:))`.


arporter:
The good news: the integration tests pass.
Bad news:
image

arporter:
I've diffed the generated OMP code with that from master and the only differences are in a comment where we don't sort the order of 'impure' routine names. Perhaps the machine just happened to be busy?

arporter:
I re-ran the NEMO tests and got:
image
so it may just be machine noise. I'll check the generated OMP-offload code against that from master too.

arporter:
The only diffs between master and this branch in the generated OMP-offoad version of NEMO4 are:
```
$ diff psycloned-openmp_gpu{,-2027}
diff psycloned-openmp_gpu/icbdyn.f90 psycloned-openmp_gpu-2027/icbdyn.f90
111c111
<       pt%lon = icb_utl_bilin_x(glamt,pt%xi,pt%yj)
---
>       pt%lon = icb_utl_bilin_x(glamt(:,:),pt%xi,pt%yj)
diff psycloned-openmp_gpu/icbrst.f90 psycloned-openmp_gpu-2027/icbrst.f90
365c365
<         nret = nf90_put_var(ncid,nsiceid,griddata,nstrt3,nlngth3)
---
>         nret = nf90_put_var(ncid,nsiceid,griddata,nstrt3(:),nlngth3(:))
```
Therefore I declare this ready for review. One for @sergisiso, @AidanChalk or @hiker.

sergisiso:
I brought the PR to master in order to run the integration tests again (it needed to use updated modules). With an empty system this PR shows no performance degradation 👍

src/psyclone/tests/psyir/nodes/array_mixin_test.py
```diff
@@ -691,6 +693,12 @@ def test_get_effective_shape(fortran_reader):
     with pytest.raises(NotImplementedError) as err:
         _ = routine.children[child_idx].lhs._get_effective_shape()
     assert "include a function call or expression" in str(err.value)
+    # Array access with indices given by another array that is not explicitly
+    # indexed.
+    child_idx += 1
+    shape = routine.children[child_idx].lhs._get_effective_shape()
+    assert len(shape) == 1
+    assert "SIZE(a)" in shape[0].debug_string()
```
sergisiso, line 701: Could you test if we do the right thing if we don't know the type of the symbols:
`b(scalarval, arrayval) = 1`

And I see this method is not overriden by ArrayMember nor ArrayOfStructuresMixing. So we could also add some tests with these constructs here.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
```diff
@@ -241,17 +241,16 @@ def validate(self, node, options=None):
...
         # All the ArrayMixins must have the same number of Ranges to expand
-        found = None
+        num_of_ranges = None
         for accessor in node.walk(ArrayMixin):
-            num_of_ranges = len([x for x in accessor.indices
-                                 if isinstance(x, Range)])
-            if num_of_ranges > 0:
-                if not found:
+            count = len([x for x in accessor.indices if isinstance(x, Range)])
+            if count:
```
sergisiso, line 247: I like the name changes, but for clarity I suggest still doing "if count > 0", even if technically it is the same.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
```diff
@@ -368,6 +367,34 @@ def validate(self, node, options=None):
                 # pylint: disable=cell-var-from-loop
                 raise TransformationError(LazyString(
                     lambda: f"{message} In:\n{node.debug_string()}"))
+            # We must understand any array accesses on the RHS. This means all
+            # array indices must be known (because otherwise they themselves
+            # might be arrays, e.g.
+            #      a(:) = b(c) where `c` can be a scalar or an array.
+            if isinstance(reference, ArrayMixin):
```
sergisiso, line 374: Again, do we still have a problem if the ArrayMixin is not in the top-level reference? e.g.
`a(:) = b%field(c)`
(maybe it should walk for inner arraymixins?)

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
```diff
@@ -368,6 +367,34 @@ def validate(self, node, options=None):
...
+                # Check the index expressions
+                range_count = len([x for x in accessor.indices
+                                   if isinstance(x, Range)])
+                if range_count != num_of_ranges:
+                    # If the number of ranges in this access is not the same as
+                    # on the LHS then we may or may not have a scalar.
```
sergisiso, line 380: Can you add: "To disambiguate we need to know that the Reference2ArrayRangeTrans in the apply will provide the remaining ranges, and for this it will need to know the type of each index expression"? (if this is the right reasoning here)

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
```diff
@@ -368,6 +367,34 @@ def validate(self, node, options=None):
...
+                    for exprn in reference.indices:
+                        if not isinstance(exprn, Reference):
+                            continue
```
sergisiso, line 383: I am not sure that's right, for example in the new "test_apply_indirect_indexing" if in
`ishtSi(5:8,jf) = ishtSi( iwewe,jf )`
if iwewe was not a Reference and instead a codeblock:
`ishtSi(5:8,jf) = ishtSi( (/ jpwe,jpea,jpwe,jpea /) , jf )`
it should not validate.

Similarly, if it was a function call that is not elemental but we don't know if it returns a scalar or an array.

By contrast, I think we will probably do the right thing for expressions, e.g.:
`ishtSi(5:8,jf) = ishtSi( iwewe + 1, jf )`
as expr.datatype with bubble up the UnresolvedType if any part of the expression is unresolved.

Please add tests for all these.

sergisiso (changes requested):
Thanks for fixing this @arporter, I suggested cases that may still break the logic of this functionality, but if fixing them is complicated feel free to defer them to follow up PRs.

sergisiso:
I am finding this same issue in lbclnk.f90 of NEMOv5 in my two open PRs, merging this branch into them fixes it.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
arporter: Done.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
arporter: I'm not sure. I've extended the comment - see what you think.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
arporter: Contrary to what you expected, we already refused to validate your first two examples. However, for the third one we haven't yet implemented support for finding the datatype when there are expressions in the array indices (#1799). I've added tests for all of these along with an associated TODO for the last one.

src/psyclone/tests/psyir/nodes/array_mixin_test.py
arporter: We didn't but we do now. Added tests.

arporter:
Unfortunately, my tightening-up of checks on the types of expressions appearing in array-indices now means that the WHERE handling is refusing some code that is/was OK, e.g.:

      WHERE( picefr(:,:) > 1.e-10 )
        zevap_ice(:,:,1) = snow(:,3,:) * frcv(jpr_ievp)%z3(:,:,1) / picefr(:,:)

Here, we don't know the type of `jpr_ievp` (because it is imported) and therefore we refuse to find the shape of `frcv(jpr_ievp)%z3(:,:,1)`. However, we can see that the whole expression already contains two ranges (`:`) and therefore we can conclude that `jpr_ievp` must be a scalar. This is going to be hard to deal with and also will conflict with @LonelyCat124 's work in #2687. I therefore don't really want to touch it until that PR is on (unless I can find a simple fix).

sergisiso:
@arporter The WHERE branch has been merged, so this PR can probably continue now. Also note that a `"lbclnk.f90",  # TODO #2685: effective shape bug` was added to examples/nemo/scripts/omp_cpu_trans.py as this was affecting NEMOv5. We want to make sure this PR resolves it and we can delete the exclusion.

arporter:
> @arporter The WHERE branch has been merged, so this PR can probably continue now. Also note that a `"lbclnk.f90", # TODO #2685: effective shape bug` was added to examples/nemo/scripts/omp_cpu_trans.py as this was affecting NEMOv5. We want to make sure this PR resolves it and we can delete the exclusion.

In what way did the bug manifest with this file? I've removed this exclusion and re-run the integration tests which were fine apart from the race condition affecting the OMP CPU results.

sergisiso:
It was a pyclone exception while processing the file. I knew this branch resolved it because I tried merging it even before the first review and I saw it was fixed here, that's why I added the TODO. It should be good now.

arporter:
I have some indirect coverage changes that I need to address.

examples/nemo/scripts/utils.py
```diff
@@ -246,6 +246,15 @@ def normalise_loops(
...
                     pass
+            if hasattr(reference, "indices"):
+                # Look at array-index expressions too.
+                for exprn in reference.indices:
+                    if (isinstance(exprn, Reference) and
+                            isinstance(exprn.symbol, DataSymbol)):
+                        try:
+                            Reference2ArrayRangeTrans().apply(exprn)
+                        except TransformationError:
+                            pass
```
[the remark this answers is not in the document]
arporter: I've tried removing the `stop_type` as you suggest and am re-running the integration tests...

arporter:
Indirect coverage changes all fixed. For one of them, I've taken the liberty of marking the bit of code with a no-cover pragma and a TODO for #1858 as that should allow the code to be covered again.

arporter:
Integration tests were fine. This is finally ready for another look now @sergisiso.

sergisiso (approved):
Thanks for addressing all the comments @arporter, this has a nice set of improvements related to datatype and shapes. The integration test pass with good results, all the formatting passes, and building the documentation doesn't introduce new warnings. This is ready to merge.

PR #2687, opened 2024-08-07
(closes #2684) Fix the failure case when we have a WHERE construct with no array index notation.

### LINKED ISSUE #2684: Fortran WHERE construct unsupported
Hello,
processing some legacy FORTRAN (UKCA glomap) that has several (many) WHERE statements; there are instances which PSyclone cannot support. i.e. when an array (1:nbox) is coded without the array index subscript... 
The example shown is declared as corrh(1:nbox), is occasionally written as corrh. I wrote a toy sample to demo the situation.
Also noted that the write_stmnt comment is interleaved with the WHERE unsupported report. Line 27 is the actual troublesome statement.
[corrh_where_f90.txt](https://github.com/user-attachments/files/16494639/corrh_where_f90.txt)

==== quoted  ===
21   ! PSyclone CodeBlock (unsupported code) reason:
 22   !  - Unsupported statement: Write_Stmt
 23   !  - Only WHERE constructs using explicit array notation (e.g. my_array(:,    :)) are supported but found 
 24 !& '[Level_4_Expr(Name('corrh'), '<', Real_Literal_Constant('1.0', None))]'.
 25   !  - Unsupported statement: Write_Stmt
 26   WRITE(6, '(20F5.2,2x)') corrh(1 : 20)
 27   WHERE (corrh < 1.0) corrh = 1.05

=== end quoted ===

LonelyCat124:


src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4188,23 +4117,38 @@ def _array_syntax_to_indexed(self, parent, loop_vars):
...
         arrays = parent.walk(ArrayMixin)
+        for ref in refs:
+            add_op = BinaryOperation.Operator.ADD
+            sub_op = BinaryOperation.Operator.SUB
+            if (type(ref) is Reference and type(ref.symbol) is not Symbol and
+                    isinstance(ref.symbol.datatype, ArrayType)):
+                # Skip if the shape doesn't match the loop_vars length.
+                if len(ref.datatype.shape) != len(loop_vars):
+                    continue
+                # Have an implicit full range reference to an array which
```
arporter, line 4136: Could you use `Reference2ArrayRangeTrans` here?

arporter: In fact, you might be better off running over the whole tree and applying this like in utils.py:
https://github.com/stfc/PSyclone/blob/d3f9715900c8cf655d0892fb4ddf4ec29f3f34ed/examples/nemo/scripts/utils.py#L186-L196

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4405,10 +4353,17 @@ def _contains_intrinsic_reduction(pnodes):
             # variable in the logical-array expression must be an array for
```
arporter, line 4353: You may be able to simplify this: just identify the child of `fake_parent` that represents "logical_expr" and ask for its `datatype`.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I probably can, I'll try to give that a go now I understand a bit how the WHERE thing is working.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I'm not sure - I need to replace any of the non-array syntax'd arrays in this logical_expr (since there can be multiples I think?) so just finding the datatype isn't sufficient, or am I misunderstanding.

arporter: Ah yes. I'd forgotten that you need the list of `arrays` in order to then update them. As with my other suggestion, it may be possible to apply `Reference2ArrayRangeTrans` to each of `fake_parent.walk(Reference)` and then you won't need your new bit of code (and the complexity is kept within that transformation).

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Its looking like the transformation can basically do everything we need for the non-structure cases here (which is what we want) so I am hoping I can just apply that generally before handling the where clauses.

LonelyCat124: There is a case where if we had `WHERE (arr1(1:20) < arr2)` this will currently fail and just generate us a codeblock still - but I'm not actually certain what the original Fortran means/if it makes sense.

LonelyCat124:
@sergisiso ready for a first look, I'm not super sure on the implementation details so its possible there's some things we're not testing for that i've missed.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -604,19 +525,36 @@ def test_where_mask_containing_sum_with_dim(fortran_reader):
...
 
 @pytest.mark.usefixtures("parser")
-@pytest.mark.parametrize("rhs", ["depth", "maxval(depth(:))"])
-@pytest.mark.xfail(reason="#717 need to distinguish scalar and array "
-                   "assignments")
-def test_where_with_scalar_assignment(rhs):
+def test_where_with_scalar_assignment():
     ''' Test that a WHERE containing a scalar assignment is handled correctly.
     Currently it is not as we do not distinguish between a scalar and an array
     reference that is missing its colons. This will be fixed in #717.
```
sergisiso, line 531: I am confused by this, do we handle this correctly now or not? If not, shouldn't it be a CodeBlock?

I think I would refactor this test to be: "test_where_with_plain_references" where we have 3 test cases with the same code and a plain reference in it.
In one case the reference is an scalar.
In the second the reference is an array without array syntax.
In the thirst is undecidable (e.g. comes from an import)

The first should keep it as scalar, the second should transform it to an indexed array and the third should put it in a codeblock. 

Also, its's up to you but I prefer to skip the process_where and use a full fortran_reader.psyir_from_source('code') snippet, where it is easier to see how each variable is declared, even if it is a bit longer.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -804,7 +742,9 @@ def test_where_ordering(parser):
      ("where (my_type%block(jl)%var(:) > epsi20)\n"
       "my_type%block%array(:,jl) = 3.0\n", "my_type%block(jl)%var"),
      ("where (my_type%block(jl)%var(:) > epsi20)\n"
-      "my_type%block(jl)%array(:,jl) = 3.0\n", "my_type%block(jl)%var")])
+      "my_type%block(jl)%array(:,jl) = 3.0\n", "my_type%block(jl)%var"),
+     ("where (my_type(:)%var > epsi20)\n"
+      "my_type%array(:,jl) = 3.0\n", "my_type")])
```
sergisiso, line 747: For this and the next test pylint says the "fortran_writer" argument is not used.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4397,18 +4324,21 @@ def _contains_intrinsic_reduction(pnodes):
...
         self.process_nodes(fake_parent, logical_expr)
+        # We want to convert all the non-explicit array syntax code to use
+        # explicit array syntax.
+        # TODO 1799 If we have two arrays where one uses a non-maximal range
+        # that is defined we will not convert this correctly yet.
+        # Convert References to arrays to use the array range notation unless
+        # they have an IntrinsicCall parent.
+        for ref in fake_parent.walk(Reference):
+            if (isinstance(ref.symbol, DataSymbol) and
+                    not ref.ancestor(IntrinsicCall)):
```
sergisiso, line 4335: Why it the IntrinsicCall check needed? If this is just related to "inquiry intrinsics" see next comment. But regardless, I would like to see a test with an inquiry, an elemental and a non-elemental intrinsic inside a where to make sure we do the right thing for each case. (if it's hard, placing them in codeblocks is good enough)

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4397,18 +4324,21 @@ def _contains_intrinsic_reduction(pnodes):
...
+                try:
+                    Reference2ArrayRangeTrans().apply(ref)
+                except TransformationError:
+                    pass
```
sergisiso, line 4339: Is it safe to ignore the TransformationErrors? I see one is related to intrinsics that may be why you need to ignore them before the try?

sergisiso (changes requested):
Thanks @LonelyCat124 I think it's a step in the right directions. See comments inline but essentially I would like to see some more testing about what happens with "references" where we can not figure out its type ow when we have different kinds of intrinsics.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4188,23 +4117,14 @@ def _array_syntax_to_indexed(self, parent, loop_vars):
...
-        assigns = parent.walk(Assignment)
-        # Check that the LHS of any assignment uses array notation.
-        # Note that this will prevent any WHERE blocks that contain scalar
-        # assignments from being handled but is a necessary limitation until
-        # #717 is done and we interrogate the type of each symbol.
-        for assign in assigns:
-            _ = self._array_notation_rank(assign.lhs)
-
-        # TODO #717 if the supplied code accidentally omits array
+        # If the supplied code accidentally omits array
```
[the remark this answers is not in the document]
LonelyCat124: I think this function handles something else - it takes all of the `x(:)` or similar and converts it to `x(lower_bound + array_var -1`.

I'll try adding to conversion in.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4309,6 +4233,9 @@ def _where_construct_handler(self, node, parent):
             not use array notation.
 
         '''
+        from psyclone.psyir.transformations import (
+                Reference2ArrayRangeTrans, TransformationError)
+
```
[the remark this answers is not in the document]
LonelyCat124: No - circular import if so.

src/psyclone/psyir/frontend/fparser2.py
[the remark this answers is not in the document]
LonelyCat124: That worked so I'll keep that change. I removed this comment as it doesn't make much sense now.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4397,18 +4324,21 @@ def _contains_intrinsic_reduction(pnodes):
         # parent for this logical expression we will repeat the processing.
         fake_parent = Assignment(parent=parent)
         self.process_nodes(fake_parent, logical_expr)
+        # We want to convert all the non-explicit array syntax code to use
+        # explicit array syntax.
```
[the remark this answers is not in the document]
LonelyCat124: Changed to that.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I saw something saying we don't allowed intrinsics (other than one specific one) inside `WHERE` statements, presumably because if you had something like `WHERE(a(:) > b(:)) c = MAX(a(:))` (not sure of exact syntax) then we'd need to do quite a lot more work to support this, so currently we disallow all but one specific one which already has a test i think. The others all just should result in a codeblock I think, but I'll try to put together a test.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
LonelyCat124: Yeah this comment was out of date.
I'll update the test to use fortran_reader too.

LonelyCat124: I also don't really understand if this test is valid fortran - can you check if it makes sense?

LonelyCat124: I think it does provided all the arrays are the same size in all dimensions (as i defined)

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
LonelyCat124: Removed them.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I think so - as far as I can see that all the TransformationErrors come without any modifications to the tree - if the transformation fails on an array, then in the worst case we get the current status (i.e. a CodeBlock).

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I'm not sure which intrinsics make sense from a Fortran perspective inside a WHERE clause, I'll try a few things out and see if it works.

LonelyCat124: maxval doesn't work for example.

LonelyCat124:
I think this is ready for another look @sergisiso . The outstanding thing I've not addressed is intrinsics (which were already "forbidden" in fparser2 before this change) - I'm not sure if/when it makes sense for them to exist.

src/psyclone/psyir/frontend/fparser2.py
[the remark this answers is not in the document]
sergisiso: Then add the pylint disable comment please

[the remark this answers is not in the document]
sergisiso: (in both added imports)

src/psyclone/psyir/frontend/fparser2.py
sergisiso: > I saw something saying we don't allowed intrinsics (other than one specific one)

If intrinsics are not supported, that would be fine, but here we skip the transformation if we fine a intrinsics, which is what I don't understand. I think the key is if they are elemental or not. For instance in this example:
```
    where(a > 1)
        b = abs(c) + sum(c)
    end where
```
we currently convert it to:
```
  do widx1 = 1, 10, 1
    if (a(widx1) > 1) then
      b(widx1) = ABS(c) + SUM(c)
    end if
  enddo
```
Which is wrong, we generate code that won't compile! It should be `b(widx1) = ABS(c(widx1)) + SUM(c)`

sergisiso: Technically, the issue is with function calls (which we need to know if they are elemental or not) not with intrinsics. If its not easy to solve we could fail if the expression has any call (and let it be a codeblock).

src/psyclone/psyir/frontend/fparser2.py
sergisiso: The issue may not come exactly from here, but if I do a test with:
```
    where(a(:) + b > 1)
        b = c + d
    end where
```
where all a, b, c, and d are improrted (so we don't know the types), it converts it to:
```
  do widx1 = 1, SIZE(a, dim=1), 1
    if (a(LBOUND(a, dim=1) + widx1 - 1) + b > 1) then
      b = c + d
    end if
  enddo
``` 
Which misses the indices in the condition 'b' and all the indices in the statement. The generated code also doesn't compile.

sergisiso:
@LonelyCat124 See the responses above, it is a bit difficult to pinpoint exactly where the issues come from, but I provided examples of where statements that psyclone produces invalid fortran when parsing them.

sergisiso:
Note that the problems that are causing issues, and the `frcv(jpr_ievp)%z3(:,:,1) / picefr(:,:)` style restriction that @arporter is mentioning in the other PR are already solved (or validation-refused) in the ArrayAssignment2LoopsTrans. WHERE's are just masked array assignments, so I am wondering if we could lean on that transformation to create a simpler implementation here that reuses that logic. It can not be applied directly but I was thinking we could convert:
```
WHERE(expr1 LOGICAL_OP expr2)
  expr3 = expr4
 ```
to:
```
expr1 = expr2 BINARY_OP expr3 BINARY_OP expr4 (... more BINARY_OP exprN if there are more statements)
```
This will create a
```
do bounds_expr1
   expr1 = expr2 BINARY_OP expr3 BINARY_OP expr4
end do
```
which then we can convert back to:
```
do bounds_expr1
   if expr1 LOGICAL_OP expr2 then
      expr3 = expr4
end do
```
I am aware that this is a totally different implementation, so maybe this PR should just identify and codeblock the issues for now. But I think this is a path to a possible solution.

File: src/psyclone/psyir/frontend/fparser2.py (hunk 5 above)
Comment (LonelyCat124) [reply src/psyclone/psyir/frontend/fparser2.py:4409]: Ah ok I see, if I find an intrinsic that isn't `SUM` then I'll raise a NotImplementedError for now.

File: src/psyclone/psyir/frontend/fparser2.py (hunk 6 above)
Comment (LonelyCat124) [reply src/psyclone/psyir/frontend/fparser2.py:4423]: Fixed this now I think.

File: src/psyclone/psyir/frontend/fparser2.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (LonelyCat124) [reply src/psyclone/psyir/frontend/fparser2.py:4318]: Done

File: src/psyclone/psyir/frontend/fparser2.py (hunk 5 above)
Comment (LonelyCat124) [reply src/psyclone/psyir/frontend/fparser2.py]: I'm actually not sure - I think intrinsics are fine except for the ones blocked by `_contains_intrinsic_reduction` so I'm going to remove this.

Comment (LonelyCat124) [reply src/psyclone/psyir/frontend/fparser2.py]: Added a test using SIN - should now work as expected.

**Comment by LonelyCat124** [comment]
@sergisiso I reviewed how we handle intrinsics, and now I think it mirrors the original and correct behaviour.
Arrays inside intrinsic are only expanded if they are elemental, inquiry ones we ignore (since we don't care) and others are acceptable (e.g. `DOT_PRODUCT(a, b)`) without being touched. The only disallowed ones are the ones that were previously disallowed (reductions with dimension arguments).

File: src/psyclone/psyir/frontend/fparser2.py [hunk 10]
```diff
@@ -4255,23 +4184,33 @@ def _array_syntax_to_indexed(self, parent, loop_vars):
...
+        from psyclone.psyir.transformations import (
+                Reference2ArrayRangeTrans, TransformationError)
+        # Convert References to arrays to use the array range notation unless
+        # they have an IntrinsicCall parent.
+        for ref in parent.walk(Reference):
+            if isinstance(ref.symbol.interface, ImportInterface):
+                raise NotImplementedError(
+                        "PSyclone doesn't yet support reference to imported "
+                        "symbols inside WHERE clauses.")
+            intrinsic_ancestor = ref.ancestor(IntrinsicCall)
```
sergisiso, line 4197: Not only Intrinsics, this will also need to be checked for any Call. If its elemental, its arguments must have array notation (for the next step to add the index), if it's not elemental, it shouldn't as it takes the whole array as argument.

For calls there is the additional case of when we don't know if it's elemental or not. In this case it should raise NotImplementedError.

Can you add tests for the 3 cases?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4255,23 +4184,33 @@ def _array_syntax_to_indexed(self, parent, loop_vars):
...
+            if (isinstance(ref.symbol, DataSymbol) and
+                    not intrinsic_ancestor):
+                try:
+                    Reference2ArrayRangeTrans().apply(ref)
+                except TransformationError:
+                    pass
+            elif (isinstance(ref.symbol, DataSymbol) and intrinsic_ancestor
```
sergisiso, line 4204: What happens if it's not a DataSymbol? Some are fine, like a RoutineSymbol, but some may need to raise a NotImplementedError, e.g. like a baseclass Symbol (which can happen with unqualified imports).

sergisiso (changes requested):
@LonelyCat124 see another couple of cases that I think we get wrong and then bring it to master as it is currently required to run the integration tests (NEMO has some WHERE so I want to see how it does for them).

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Do we need explicit tests for Call since we it will use the same code as IntrinsicCall for most of this, only the NotImplementedError case I should have a separate test.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I assumed it would already be handled by the frontend to result in a NotImplementedError if we weren't sure, e.g. `ImportInterface` is already blocked by line 4422 (`if isinstance(ref.symbol.interface, ImportInterface):`). I didn't feel like we needed to explicitly handle those during this function as they'd fail with a more accurate message later/earlier (depending on when we're calling this).

sergisiso:
@LonelyCat124 wrt the failing integration test, as we discussed last week, we need to do the following change:

```
         # For performance in lib_fortran, mark serial routines as GPU-enabled
         if psyir.name == "lib_fortran.f90":
-            if not subroutine.walk(Loop):
-                try:
-                    # We need the 'force' option.
-                    # SIGN_ARRAY_1D has a CodeBlock because of a WHERE without
-                    # array notation. (TODO #717)
-                    OMPDeclareTargetTrans().apply(subroutine,
-                                                  options={"force": True})
-                except TransformationError as err:
-                    print(err)
+            if subroutine.name.lower().startswith("sign_"):
+                OMPDeclareTargetTrans().apply(subroutine)
```
In `omp_gpu_trans.py`, `acc_loops_trans.py`, and `acc_kernels_trans.py`. I checked that it works and doesn't degrade the performance. The changes are:
- We don't identify the routine by having no loops anymore, we identify them by name.
- The force is gone because there is no Codeblock anymore as the WHERE is expanded. Well done.
- I removed the except because if this fails it will produce code that doesn't compile, so we better fail early

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Yes, if they use the same code you can only test the new NotImplementedError case

src/psyclone/psyir/frontend/fparser2.py
sergisiso: I see you added UnresovledInterface in the commit (which should be the same as symbols that use the baseclass as they could not be found)  and the performance of NEMO did not degrade. This is safer than before so I would keep this change. You will need to fix the new failing tests (probably you will need to convert some imported symbols to local?) and would be good to add a test with an unqualified import producing the NotImplementedError/Codeblock (even if the coverage says its the same line as the ImportInterface check).

LonelyCat124: I think some of the failing tests were testing "what happens if unqualified import" so I'll convert those into these tests probably.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: This should be covered then.

src/psyclone/psyir/frontend/fparser2.py

LonelyCat124:
I've set the integration tests running again now with those changes.

sergisiso (approved):
@LonelyCat124 I made changes in some comments and pointed out some additional issues in #2722. But since this already fixes some of the reported issues and it passes the integration tests, it is a good amount of progress to merge this and defer remaining issues to a future PR. This is ready to merge.

PR #2692, opened 2024-08-21
(closes #2691) Disable lowering copy when no more changes are expected and increase recursion limit when needed

### LINKED ISSUE #2691: PSyclone backend reaches recursion limit visiting a 500 depth nested if-else block.
This comes from a cannonicalised SELECT CASE with 500 cases in UKCA ukca_aero_ctl.F90 . The frontend, and exploring the tree works fine, but the backend (even when disabling the copy, is failing with `RecursionError: maximum recursion depth exceeded`  )

sergisiso:


sergisiso:
This is ready for a first review from @arporter, @hiker or @LonelyCat124

src/psyclone/generator.py
```diff
@@ -729,7 +729,7 @@ def code_transformation_mode(input_file, recipe_file, output_file,
             Profiler.add_profile_nodes(routine, Loop)
 
         # Generate Fortran
-        output = FortranWriter()(psyir)
+        output = FortranWriter(disable_copy=True)(psyir)
```
arporter, line 732: Please extend the comment to say why we 'disable_copy' i.e. this is the final step and so the tree cannot subsequently be modified.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -305,28 +305,16 @@ class FortranWriter(LanguageWriter):
...
-        global constraints when walking the tree. Defaults to True.
+    :param kwargs: additional keyword arguments provided to the super class.
+    :type kwargs: unwrapped dict.
 
     '''
     _COMMENT_PREFIX = "! "
 
-    def __init__(self, skip_nodes=False, indent_string="  ",
-                 initial_indent_depth=0, check_global_constraints=True):
+    def __init__(self, **kwargs):
```
arporter, line 314: Nice simplification.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -305,28 +305,16 @@ class FortranWriter(LanguageWriter):
...
         # Construct the base class using () as array parenthesis, and
```
arporter, line 315: This comment would be useful at L69 in c.py (updated appropriately of course).

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3965,6 +3966,14 @@ def _case_construct_handler(self, node, parent):
             if isinstance(child, Fortran2003.End_Select_Stmt):
                 clause_indices.append(idx)
 
+        # Deeply-nested ifblocks are problematic because they will exceed the
+        # python recursion limits for some psyir analysis. These are not
+        # typically found in input code, but select-case with a large number
+        # of cases are more probable. If we find one of such cases we
+        # preventively increase the python recursion limits.
+        if len(clause_indices) > 150:
```
arporter, line 3974: Please add a TODO for #11 as it would be good to log this.

src/psyclone/tests/psyir/frontend/fparser2_select_case_test.py
```diff
@@ -95,6 +96,37 @@ def test_handling_case_construct():
...
+    code += '''
+        END SELECT
+    end subroutine'''
+    _ = fortran_reader.psyir_from_source(code)
+
+    # Check that the recursion limit has been increased
+    assert sys.getrecursionlimit() > default_recursion_limit
+
+    # Restore default value
+    sys.setrecursionlimit(default_recursion_limit)
```
arporter, line 127: How safe is this? Should we perhaps do the body of the test within a `try` and then have this in a `finally`?

arporter: Alternatively, we could `monkeypatch` `sys.setrecursionlimit()` so that it just raises an error or something and check that the frontend causes it to be called?

arporter (changes requested):
More good stuff Sergi. Only thing worth mentioning is that I'm slightly worried about preventing the new test possibly modifying the environment of subsequent tests.

src/psyclone/psyir/backend/fortran.py
hiker: Might it be better to leave the comments in place? If I need to find out what parameters I can specify, I first need to find LanguageWriter, which tells me that I need to look at Visitor. I admit that this is not an idea solution either, since then we have to keep the comments across three classes in sync :( Or maybe just add a link to where the valid names are specified?

sergisiso: We had the same problem on the PSyIR nodes and Symbols kwargs, and the Transformations options which are passed to the super class. I don't like having them duplicated as we never check them when they are updated, or when new arguments added to the subclasses, and would be quite hard to review.

Adding a link to the super() is a good and simple solution (which some IDEs, static type tools may be able to follow). So I will use this here.

There is a more complex solution of programmatically modifying the `__doc__` to include the BASE.__doc__ or even spit __doc__ to separate the kwargs so that they can be included in the subclass docsting, but this may be over-engineering it, and probably not for this PR.

src/psyclone/generator.py
sergisiso: Done

src/psyclone/psyir/backend/fortran.py
sergisiso: I promoted this comment into an issue and will be looked at independently of this PR

src/psyclone/psyir/backend/fortran.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/tests/psyir/frontend/fparser2_select_case_test.py
sergisiso: Done (with the try-finally approach)

arporter (approved):
All looks good now. I've brought the branch up to master and fired off the integration tests one more time.

PR #2706, opened 2024-09-03
2483 support derived types in driver

hiker:
Adds support for the drivers to properly read in derived type (using non-derived Fortran variables based on the actual basic type of the members).


hiker:
To clarify: this does not support passing derived types to subroutines (upcoming PR for 2830_support_user_defined_type_arguments), it only handles the case that a subroutine somewhere uses a member

arporter (changes requested):
Thanks for this Joerg. Functionally it all looks OK but I do have questions about array accesses in structures and structure accesses of greater depths.
Aside from that, I'm not entirely comfortable about the Signature <-> PSyIR interface and think it should perhaps be the other way around. It's not a biggie for this PR though.
I'll launch the integration tests next time (unless you fancy doing it when you arrive at the office in the morning?).

PR #2708, opened 2024-09-09
(Closes #2663) add lfric integration tests on gpu

### LINKED ISSUE #2663: Extend integration tests to run 'LFRic' on GPU
Up to now, out testing of this locally has been ad-hoc. It would be good to capture what we have so far (optimisation script and compiler options) in the integration tests.

Currently however, I don't think that LFRic can be built with the latest NVIDIA compiler (24.5) without some source modifications to work around compiler bugs. Perhaps we could patch those in as part of the test?

arporter:


arporter:
re. the MongoDB @sergisiso, should I just cut-n-paste what you have for other job steps and edit it appropriately or is there more to it than that?

sergisiso:
@arporter Yes, that's all, it's an unstructured db so it will take whatever fields you want in your insert command.

Once the values are uploaded I do need to modify the chart to include the new series of values.

arporter:
The NEMOv4 integration tests are failing but that doesn't seem to be the fault of this PR (I suspect a race condition as it's the OMP test that fails) and the 'fix NEMO v5' issue #2692 seems to fix it.

arporter:
Ready for a first look. Probably has to be @sergisiso as it involves the integration tests and MongoDB.
Note that the transformation script is far from optimal but there are limitations that need to be addressed before we can do better.

.github/workflows/lfric_test.yml
```diff
@@ -75,6 +75,43 @@ jobs:
...
+        export PSYCLONE_LFRIC_DIR=${GITHUB_WORKSPACE}/examples/lfric/scripts
+        export PSYCLONE_CONFIG_FILE=${PSYCLONE_LFRIC_DIR}/KGOs/lfric_psyclone.cfg
+        export LFRIC_DIR=${HOME}/LFRic/lfric_apps_r2222
+        export OPT_DIR=${LFRIC_DIR}/applications/gungho_model/optimisation/psyclone-test
+        cd ${LFRIC_DIR}
+        # PSyclone scripts must now be under 'optimisation' and be called 'global.py'
+        mkdir -p ${OPT_DIR}
+        cp ${PSYCLONE_LFRIC_DIR}/acc_parallel.py ${OPT_DIR}/global.py
+        # Clean previous version and compile again
+        rm -rf appplications/gungho_model/working
```
sergisiso, line 94: s/ppp/pp/

.github/workflows/lfric_test.yml
```diff
@@ -75,6 +75,43 @@ jobs:
...
+    # PSyclone, compile and run MetOffice gungho_model on GPU
+    - name: LFRic GungHo with OpenACC offload
+      run: |
+        # Set up environment
+        source /apps/spack/psyclone-spack/spack-repo/share/spack/setup-env.sh
+        spack load lfric-build-environment%nvhpc
+        source .runner_venv/bin/activate
+        export PSYCLONE_LFRIC_DIR=${GITHUB_WORKSPACE}/examples/lfric/scripts
+        export PSYCLONE_CONFIG_FILE=${PSYCLONE_LFRIC_DIR}/KGOs/lfric_psyclone.cfg
+        export LFRIC_DIR=${HOME}/LFRic/lfric_apps_r2222
```
sergisiso, line 87: I see this doesn't use the LFRIC_APPS_REV revision, and it uses an older version. Was there any problem with the newer revision? If there was a problem, is it worth documenting it here with an inlined comment and TODO if an issue exists?

.github/workflows/lfric_test.yml
```diff
@@ -75,6 +75,43 @@ jobs:
...
+        cd ${LFRIC_DIR}
+        # PSyclone scripts must now be under 'optimisation' and be called 'global.py'
+        mkdir -p ${OPT_DIR}
+        cp ${PSYCLONE_LFRIC_DIR}/acc_parallel.py ${OPT_DIR}/global.py
+        # Clean previous version and compile again
+        rm -rf appplications/gungho_model/working
+        ./build/local_build.py -a gungho_model -p psyclone-test
+        # Run
+        cd applications/gungho_model/example
+        cp ${PSYCLONE_LFRIC_DIR}/KGOs/lfric_gungho_configuration_gpu_4its.nml configuration.nml
```
sergisiso, line 98: Also add a comment making explicit that we run a different configuration that the other tests. Also if the differences are because we use a different revision or because there are issues with the GPU.

.github/workflows/lfric_test.yml
```diff
@@ -75,6 +75,43 @@ jobs:
...
+        export VAR_TIME=$(grep "gungho_model" timer.txt | cut -d'|' -f5)
+        export VAR_HALOS=$(grep "gungho_model" halo_calls_counter.txt | cut -d'|' -f5)
+        echo $GITHUB_REF_NAME $GITHUB_SHA $VAR_TIME $VAR_HALOS >> ${HOME}/store_results/lfric_acc_performance_history
+        ${HOME}/mongosh-2.1.1-linux-x64/bin/mongosh \
+          "mongodb+srv://cluster0.x8ncpxi.mongodb.net/PerformanceMonitoring" \
+          --quiet --apiVersion 1 --username ${{ secrets.MONGODB_USERNAME }} \
+          --password ${{ secrets.MONGODB_PASSWORD }} \
+          --eval 'db.GitHub_CI.insertOne({branch_name: "'"$GITHUB_REF_NAME"'", commit: "'"$GITHUB_SHA"'",
+          github_job: "'"$GITHUB_RUN_ID"'"-"'"$GITHUB_RUN_ATTEMPT"'",
+          ci_test: "LFRic OpenACC", lfric_apps_version: '"$LFRIC_APPS_REV"', system: "GlaDos",
```
sergisiso, line 111: Replace LFRIC_APPS_REV with the right number

examples/lfric/scripts/acc_parallel.py
```diff
@@ -0,0 +1,149 @@
...
+from psyclone.psyir.transformations import (
+    ACCKernelsTrans, TransformationError)
+from psyclone.transformations import (
+    Dynamo0p3ColourTrans, Dynamo0p3OMPLoopTrans,
+    Dynamo0p3RedundantComputationTrans, OMPParallelTrans,
+    ACCParallelTrans, ACCLoopTrans, ACCRoutineTrans)
+
+
+ACC_EXCLUSIONS = [
+]
```
sergisiso, line 54: Do you want to keep this if it's empty?

sergisiso (changes requested):
Thanks @arporter it is great to see this finally on the Integration tests, also I see it automatically appeared in the charts which is nice. I am only suggesting some tweaks and clarifications on the workflow script, everything else looks good.

I agree the NEMOv4 issue is not related to this PR.

.github/workflows/lfric_test.yml
arporter: Oops.

.github/workflows/lfric_test.yml
arporter: It was more that I had a patch from Lukas and this was the version of the source he was working with. You're right though that we do need to upate this so I'll add a TODO.

.github/workflows/lfric_test.yml
arporter: Oh yes and TODO added.

examples/lfric/scripts/acc_parallel.py
arporter: Yes, I think I'll keep it for now. I'll add a comment.

.github/workflows/lfric_test.yml
arporter: Oh yes. I expect that we can remove this once the revision is updated.

.github/workflows/lfric_test.yml
arporter: I've asked Lukas if he can update his LFRic working copy and thus the patch.

arporter:
I've fired off the integration tests again. Should be ready for another look.

arporter:
Taking this back as Lukas has provided me with an updated patch so I'm going to update the GPU test to use the same revision as the CPU tests.

arporter:
Wow, that was a bit painful but I now have the integration tests working with the same (albeit patched) version of LFRic as the rest of the tests. I've also removed the gist upload. Ready for another look now.

.github/workflows/lfric_test.yml
arporter: Gone now.

.github/workflows/lfric_test.yml
arporter: Reverted now.

.github/workflows/lfric_test.yml
arporter: He already had a patch so I've used that and am working from the same LFRic apps revision now.

.github/workflows/lfric_test.yml
```diff
@@ -75,6 +75,58 @@ jobs:
...
+        cp -r ${HOME}/LFRic/lfric_apps_${LFRIC_APPS_REV} ${HOME}/LFRic/gpu_build/lfric_apps
+        cp -r ${HOME}/LFRic/lfric_core_50869 ${HOME}/LFRic/gpu_build/lfric
+        cd ${HOME}/LFRic/gpu_build
+        patch -p1 < ../lfric_${LFRIC_APPS_REV}_nvidia.patch
+        # Update the compiler definitions to build for GPU
+        cp ../nvfortran.mk lfric/infrastructure/build/fortran/.
+        cp ../nvc++.mk lfric/infrastructure/build/cxx/.
+        # Update the PSyclone commands to ensure transformed kernels are written
+        # to working directory.
+        cp ../psyclone.mk lfric/infrastructure/build/psyclone/.
```
sergisiso, line 100: Could you put `../lfric_${LFRIC_APPS_REV}_nvidia.patch` `../nvfortran.mk` `../nvc++.mk` and `../psyclone.mk` in somewhere in the examples/lfric/scripts?

I addition to being able to capture the whole test, I would also like to have all the files to be able to replicate it in my workspace easily.

sergisiso (changes requested):
@arporter Just one more request to bring the needed files into the repository.

.github/workflows/lfric_test.yml
arporter: Yes, good idea. I also realised that the transformation script adds OpenMP if it fails to add OpenACC but we weren't building with OMP enabled. Hopefully this will still work...

arporter:
It worked :-) Ready for another look now.

sergisiso (approved):
Thanks @arporter this is ready to merge

PR #2725, opened 2024-10-01
(closes #717) Remove unused code

### LINKED ISSUE #717: Where statement array syntax is limited

sergisiso:
#717 was the same than #2687 but there were some TODOs associated to it. Looking at them it turns out that they where related to the array shape querying ulility functionality in fparser2, but these are now unused (other than in its own tests) as we use the ArrayMixin utility methods.


sergisiso:
This is ready for review. Initially I just wanted to delete some unused test, but then realised that the WHERE fix also solves the last passthrough file in NEMOv4 🥳  and it has a TODO to this issue.

After removing it from the excluded file another issue hit that file: the fact that we cannot distinguish between imported calls with arguments and arrays. This was somewhat fixed in ACCKernels by listing all the NEMO FUNCTIONS, so I generalised that into utils so it can be used by all transformations.

It passed the integration tests with expected performance.

One for @arporter (since it touches the NEMO ACC Kernels) or @LonelyCat124

examples/nemo/scripts/utils.py
```diff
@@ -139,15 +146,18 @@ def enhance_tree_information(schedule):
...
-            reference.replace_with(call)
+        elif reference.symbol.name in NEMO_FUNCTIONS:
+            if reference.symbol.is_import or reference.symbol.is_unresolved:
+                # The parser gets these wrong, they are Calls not ArrayRefs
+                if not isinstance(reference.symbol, RoutineSymbol):
+                    # We haven't already specialised this Symbol.
+                    reference.symbol.specialise(RoutineSymbol)
+                if not (isinstance(reference.parent, Call) and
+                        reference.parent.routine is reference):
+                    call = Call.create(reference.symbol)
```
arporter, line 157: Comment please, e.g. "Replace the existing Reference with a Call"

examples/nemo/scripts/utils.py
```diff
@@ -226,9 +236,10 @@ def normalise_loops(
         statements out of the loop nest.
     '''
 
-    # TODO #1902: NEMO4 mpi_ini.f90 has a HoistLocalArraysTrans bug
-    if hoist_local_arrays and schedule.root.name != "mpp_ini.f90":
-        # Apply the HoistLocalArraysTrans when possible
+    if hoist_local_arrays and schedule.name not in CONTAINS_STMT_FUNCTIONS:
+        # Apply the HoistLocalArraysTrans when possible, it cannot be applied
+        # to files with statement functions because it will attempt to put the
+        # allocate avobe it, which is not valid Fortran.
```
arporter, line 242: "above"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -396,179 +396,6 @@ def _find_or_create_psyclone_internal_cmp(node):
         "not have an ancestor container in which to add it.")
```
arporter, line 398: Nice!

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -98,14 +89,14 @@ def trans(psyir):
...
 
-        # In the lib_fortran file we annotate each routine of the SIGN_*
-        # interface with the OpenMP Declare Target Directive
-        if psyir.name == "lib_fortran.f90":
-            if subroutine.name.lower().startswith("sign_"):
-                OMPDeclareTargetTrans().apply(subroutine)
-                # We continue parallelising inside the routine, but this could
-                # change if the parallelisation directive are not nestable, in
-                # which case we could add a 'continue' here
+        # This are functions that are called from inside parallel regions,
```
arporter, line 92: s/This/These/

examples/nemo/scripts/acc_loops_trans.py
```diff
@@ -108,12 +99,12 @@ def trans(psyir):
                 hoist_expressions=True
         )
 
-        # In the lib_fortran file we annotate each routine of the SIGN_*
-        # interface with the OpenACC Routine Directive
-        if psyir.name == "lib_fortran.f90":
-            if subroutine.name.lower().startswith("sign_"):
-                ACCRoutineTrans().apply(subroutine)
-                continue
+        # This are functions that are called from inside parallel regions,
```
arporter, line 102: s/This/These/

arporter (changes requested):
Nice work Sergi. There ought to be some prize for the PR which removes the most code. Tests all OK, integration tests look good. Reference guide unchanged.
Coverage unaffected.
Just a few minor things to tidy and then this can go on.

examples/nemo/scripts/utils.py
sergisiso: Fixed

examples/nemo/scripts/utils.py
sergisiso: Done

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: Done

examples/nemo/scripts/acc_loops_trans.py
sergisiso: Done

sergisiso:

arporter (approved):
All good now. I've re-triggered the integration tests as master has moved on since they were last run for this branch.

arporter:
Integration tests were green.

PR #2732, opened 2024-10-03
(Closes #2716) Add support for module-inlining calls to polymorphic kernels/routines

### LINKED ISSUE #2716: Bugs when transforming mixed-precision kernels
As found here: https://github.com/stfc/PSyclone/issues/2663#issuecomment-2332037300

The problem seems to be that we rename the interface rather than the actual subroutine that is being called.

arporter:


arporter:
As well as addressing the coverage, I need to extend the LFRic integration tests to use this new functionality.

arporter:
Added the KernelModuleInlineTrans back into the LFRic transformation script and get a crash when trying to add the interface symbol into the symbol table.

arporter:
Build and run with updated transformation script works:
image

sergisiso:
If you merge with master, I added multiple prints in the output, so you can grep and count the number of each cases that we could not get the kernel_schedule before due to polymorphic kernels, which was 331

arporter:
The OMP offload LFRic integration test failed with an ICE:
```
17:47:36 Pre-process and compile inventory/id_r32_field_array_pair_mod.F90
inventory/id_r32_field_array_pair_mod.F90:
NVFORTRAN-S-0000-Internal compiler error. flowgraph: node is zero       3  (lfric_xios_setup_mod_psy.f90: 85)
NVFORTRAN-F-0000-Internal compiler error. Invalid key for hash       0  (lfric_xios_setup_mod_psy.f90: 85)
NVFORTRAN/x86-64 Linux 24.5-1: compilation aborted
```
It's seems odd that there's a `_psy.f90` for xios setup??

arporter:
Do the build manually for OpenACC and things are looking better:
image
(note: I've marked MATMUL as being available on the GPU now.)

arporter:
I need to figure out why we get an inlined routine without an offload directive added to it. The build log says:
```
PSy name = 'lfric_xios_setup_mod_psy'
Transforming invoke 'invoke_0_nodal_xyz_coordinates_kernel_type' ...
Failed to annotate 'nodal_xyz_coordinates_code' with GPU-enabled directive due to:
Transformation Error: Kernel 'nodal_xyz_coordinates_code' accesses the symbol 'chi2xyz: RoutineSymbol<NoType, pure=unknown, elemental=unknown>' which is imported. If this symbol represents data then it must first be converted to a Kernel argument using the KernelImportsToArguments transformation.
Failed to annotate 'nodal_xyz_coordinates_code' with GPU-enabled directive due to:
Transformation Error: Kernel 'nodal_xyz_coordinates_code' accesses the symbol 'chi2xyz: RoutineSymbol<NoType, pure=unknown, elemental=unknown>' which is imported. If this symbol represents data then it must first be converted to a Kernel argument using the KernelImportsToArguments transformation.
Transforming invoke 'invoke_1_nodal_coordinates_kernel_type' ...
Failed to module-inline kernel 'nodal_coordinates_code' due to:
Transformation Error: Cannot inline subroutine 'nodal_coordinates_code' because another, different, subroutine with the same name already exists and versioning of module-inlined subroutines is not implemented yet.
Successfully offloaded loop with ['nodal_coordinates_code']
Successfully offloaded loop with ['nodal_coordinates_code']
Transforming invoke 'invoke_2_nodal_coordinates_kernel_type' ...
Failed to module-inline kernel 'nodal_coordinates_code' due to:
Transformation Error: Cannot inline subroutine 'nodal_coordinates_code' because another, different, subroutine with the same name already exists and versioning of module-inlined subroutines is not implemented yet.
Successfully offloaded loop with ['nodal_coordinates_code']
Successfully offloaded loop with ['nodal_coordinates_code']
```

arporter:
Part of the problem was that the optimisation script wasn't checking that it hadn't already transformed a given kernel for a given invoke. Fixing that removes the 'failed to inline' messages but we still end up with an inlined kernel that doesn't have a directive added to it.

arporter:
The problem was that, having successfully module-inlined the kernel routine, we proceed to apply the annotation transformation to the original Kern and that does not update the Routine that has been inlined. It probably should? For now, I can search for the newly inlined routine and apply the transformation to that and we get the expected code.

arporter:
Now get a compilation failure:
```
NVFORTRAN-S-0155-Ambiguous interfaces for generic procedure matrix_vector_code (
algorithm/norm_alg_mod_psy.f90: 338)
```

arporter:
`Kern.get_kernel_schedule()` (which is to be replaced/removed if/when we migrate Kern to subclass Call) currently caches the PSyIR of the kernel. `KernelModuleInlineTrans` creates a copy of this PSyIR and then inserts it into the PSy-layer. Therefore, `get_kernel_schedule()` should now return that copy of the PSyIR. In fact, that would happen automatically if I undid my changes to `KernelModuleInlineTrans` so that it *doesn't* copy the PSyIR. I can't remember why I made that change.

arporter:
Thinking about this a bit more, I belatedly realise that if a given PSy layer routine calls the same kernel more than once then either all of them must be module inlined or none of them (unless we attempt to rename the inlined version and that gets complicated). I think the reason that LFRic is still not working is that we fail to inline a second instance of the same kernel (because the first instance has been transformed) but then we do proceed to transform it. Therefore, `rename_and_write()` is called in order to generate a new version of that kernel on disk. I need a reproducer for this.

arporter:
I've realised that `Kern.module_inline` sets this flag for *all* Kernels with that name in the current `InvokeSchedule`. This seems sensible but is causing me problems because we end up with a Kernel marked as 'module inlined' but which actually isn't because the transformation on it failed and so it was written to file and renamed instead.

arporter:
The trouble with the LFRic example is that we have two, *separate* `invoke` calls, each with the same Kernel. The first Kernel gets module inlined but has to be modified to have some imports made local to it. That then means we refuse to inline the second Kernel because it is no longer identical. The question is, why does the search for the source of the second kernel not return the module-inlined source of the first one?

arporter:
With this branch and latest gpu script in lfric/examples:
image

arporter:
Adding RANDOM_NUMBER intrinsic to available_on_device - build still worked OK on GPU but profile didn't look much different. Need to update the Sankey diagram and see whether I've missed anything.

arporter:
One of the compilation tests fail with the nvidia compiler. This same test passes with gfortran:

image

arporter:
```
$ nvfortran -I../dynamo_wrapper0/utilities -I../dynamo_wrapper0/field -I../dynamo_wrapper0/operator -I../dynamo_wrapper0/function_space -I../dynamo_wrapper0/mesh -I../dynamo_wrapper0/scalar -I../dynamo_wrapper0/configuration -c psy.f90
NVFORTRAN-S-0155-Ambiguous interfaces for generic procedure mixed_code (psy.f90: 289)
```

arporter:
The problem is that the invoke subroutine is still importing the kernel interface, even though it has been module inlined:
```fortran
    SUBROUTINE invoke_1(scalar_r_phys, a, istp)
      USE mixed_kernel_mod, ONLY: mixed_code
```

arporter:
I'm beginning to think that this is a compiler bug. If I alter the second invoke so that it calls a specific implementation of the kernel (e.g. mixed_code_64) then it compiles fine.

arporter:
This is ready for a first review. It's quite big but can wait until 3.0 is out of the door. Probably needs to be reviewed by @sergisiso.

arporter:
Trying this updated branch with LFRic (and my enhanced global.py) I now get:
```
Failed to module-inline routine coordinate_jacobian:
Transformation Error: routine 'coordinate_jacobian' contains accesses to 'jacobian_abr2xyz' which is declared in the same module scope. Cannot inline such a routine.
```
where `jacobian_abr2xyz` is actually a routine in the same module so we could inline it...

Also get
```
File '../lfric/components/science/source/kernel/geometry/chi_transform_mod.F90' does contain module 'chi_transform_mod' but PSyclone is unable to create the PSyIR of it.
```
This is possibly because that module contains an unsupported form of interface definitions:
```fortran
  INTERFACE
    SUBROUTINE chi2xyz_interface(chi_1, chi_2, chi_3, panel_id, x, y, z)
      IMPORT :: r_def, i_def
      IMPLICIT NONE
      INTEGER(KIND = i_def), INTENT(IN) :: panel_id
      REAL(KIND = r_def), INTENT(IN) :: chi_1, chi_2, chi_3
      REAL(KIND = r_def), INTENT(OUT) :: x, y, z
    END SUBROUTINE chi2xyz_interface
  END INTERFACE
```

arporter:
Attempt to do recursive inlining:
```
Failed to inline call jacobian_abr2xyz(alpha, beta, radius, panel_id):
Transformation Error: Routine 'jacobian_abr2XYZ_r_single' cannot be inlined because it accesses variable 'r_single' from its parent container.
```
but `r_single` is actually named in an import into the parent container => should be easy to handle.

arporter:
Got a crash while processing `init_saturated_profile_alg_mod.x90`:
```
  File "/home//Projects/PSyclone/src/psyclone/domain/lfric/lfric_loop.py", line 895, in gen_code
    super().gen_code(parent)
    ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home//Projects/PSyclone/src/psyclone/psyir/nodes/loop.py", line 581, in gen_code
    kernel.rename_and_write()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home//Projects/PSyclone/src/psyclone/psyGen.py", line 1664, in rename_and_write
    self._rename_psyir(new_suffix)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home//Projects/PSyclone/src/psyclone/psyGen.py", line 1741, in _rename_psyir
    kern_schedule.name = new_kern_name[:]
    ^^^^^^^^^^^^^^^^^^
  File "/home//Projects/PSyclone/src/psyclone/psyir/nodes/routine.py", line 419, in name
    sym = self.symbol_table.lookup(symbol.name)
```
because the Routine instance doesn't yet have a symbol table.

arporter:
Fixed that and now get failure when doing:
```
LFRIC_OFFLOAD_DIRECTIVES=omp psyclone -api lfric -d /Projects/LFRic/spack_gfortran/lfric_apps_7536/applications/gungho_model/working/build_gungho_model -s applications/gungho_model/optimisation/psyclone-test/global.py /Projects/LFRic/spack_gfortran/lfric_apps_7536/applications/gungho_model/working/build_gungho_model/algorithm/initialisation/init_saturated_profile_alg_mod.x90
```
because the kernel schedule obtained in `CodedKern._rename_and_write()` has somehow been detached from its parent container.

arporter:
```
Failed to annotate 'initial_theta_code' with GPU-enabled directive due to:
Transformation Error: Kernel 'initial_theta_code' accesses the symbol 'analytic_temperature: RoutineSymbol<UnresolvedType, pure=unknown, elemental=unknown>' which is imported. If this symbol represents data then it must first be converted to a Kernel argument using the KernelImportsToArguments transformation.
Successfully offloaded loop with ['inc_x_times_y']
Added OMP threading to loop with ['initial_theta_code']
```

arporter:
Further investigation shows that it is when generating the code for `invoke_5` that we get the failure. When processing this invoke I see:
```
Transforming invoke 'invoke_5' ...
setting _parent None
setting _parent None
setting _parent None
setting _parent Container[init_saturated_profile_alg_mod_psy]

setting _parent None
setting _parent Container[init_saturated_profile_alg_mod_psy]

Failed to module-inline kernel 'theta_e_code' due to:
Transformation Error: Kernel 'theta_e_code' cannot be module inlined into Container 'init_saturated_profile_alg_mod_psy' because a *different* routine with that name already exists and versioning of module-inlined subroutines is not implemented yet.
Annotated kernel 'theta_e_code'
```

arporter:
LFRic with OMP offload now fails to compile:
```
NVFORTRAN-S-1058-Call to NVHPC runtime function not supported - pgf90_matmul_real4_i8 (algorithm/solver/pressure_diag_precon_alg_mod_psy.f90: 36)
  0 inform,   0 warnings,   1 severes, 0 fatal for dg_matrix_vector_code_r_single
```

arporter:
```
Failed to replace MATMUL in 'project_eliminated_theta_q32_code' because:
Transformation Error: Matmul2CodeTrans only supports the transformation of a MATMUL operation when it is the sole operation on the rhs of an assignment.
Annotated kernel 'project_eliminated_theta_q32_code'
```

arporter:
NEMOv4 OpenACC Kernels integration test failed with:
```
NVFORTRAN-S-0155-Ambiguous interfaces for generic procedure mpp_sum (/home/me/NEMO/cfgs/SPITZ12_openacc_kernels/WORK/bdyini.f90: 1742
```
This is with NVHPC 25.1. I've tried to reproduce this with 25.3 but no joy - it compiled fine.

arporter:
The difference seems to be that in the runner, we PSyclone lib_mpp.f90 whereas it is excluded (for some reason) when I attempt to replicate it locally.

arporter:
It looks as though we've inlined `mpp_sum` into `bdy_oce` but haven't made the interface symbol private:
```bash
$ grep -i 'interface mpp_sum' psycloned-openacc_kernels/*90
psycloned-openacc_kernels/bdy_oce.f90:  interface mpp_sum
psycloned-openacc_kernels/bdy_oce.f90:  end interface mpp_sum
psycloned-openacc_kernels/lib_mpp.f90:  interface mpp_sum
psycloned-openacc_kernels/lib_mpp.f90:  end interface mpp_sum
```

arporter:
I tried adding `lib_mpp` to the modules that the frontend will resolve but the resulting inlining of mpp_sum in icbrst.f90 caused another compilation failure with "ambiguous interface". Presumably because lib_mpp is imported in load of other modules and, of course, NEMO always does wildcard imports so there are lots of other ways the original `mpp_sum` could be brought into scope. The only solution to this is #2846 so that's probably what I'll do after this one.

arporter:
This is now ready for first review. Integration tests pass. I'm thinking we probably need to rename `Kern.get_kernel_schedule()` but I'll let you see what you think. Probably another one for @sergisiso I'm afraid (but it's a lot smaller than the last one).

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -278,20 +293,21 @@ def _prepare_code_to_inline(
...
                     :py:class:`psyclone.psyGen.CodedKern`
 
-        :returns: the name of the routine as seen by the caller and the
-                  PSyIR of the routine implementation.
-        :rtype: Tuple(str, :py:class:`psyclone.psyir.nodes.Call`)
+        :returns: the PSyIR of the routine implementation(s) and the associated
+            interface symbol if it is polymorphic.
+        :rtype: tuple[
+            list[:py:class:`psyclone.psyir.nodes.Routine`],
+            :py:class:`psyclone.psyir.symbols.GenericInterfaceSymbol`)]
```
sergisiso, line 309: I would also return a union one or the other, not a tuple of one plus maybe the other, so basically:
```
if isinstance(node, CodedKern):
    return node.get_kernel_schedule()
...
```

src/psyclone/psyir/nodes/call.py
```diff
@@ -506,14 +487,20 @@ def _location_txt(node):
...
-                                                      allow_private=True)
-                    if psyir:
+                    routines = []
+                    for name in cursor.resolve_routine(rsym.name):
+                        # Since we're looking in the local Container, the
+                        # target is permitted to be private.
+                        psyir = cursor.find_routine_psyir(rsym.name,
+                                                          allow_private=True)
+                        if psyir:
+                            routines.append(psyir)
```
sergisiso, line 497: I don understand why we resolve_routine but then not use the returned name. Should it be "name" what we pass to "find_routine_psyir"?

sergisiso (changes requested):
@arporter I made a first high-level review, I link the implementation but not so much the method return type that it leads to lots of ` = kernels[0]` over the code which can silently produce errors, so I suggested another interface which could allow to revert some of this changes. Let me know what you think.

src/psyclone/domain/lfric/lfric_kern.py
```diff
@@ -708,61 +708,68 @@ class creates the PSyIR schedule on first invocation which is
         Once issue #935 is implemented, this routine will return the
         PSyIR Schedule using LFRic-specific PSyIR where possible.
 
-        :returns: Schedule representing the kernel code.
-        :rtype: :py:class:`psyclone.psyGen.KernelSchedule`
+        :returns: the Symbol defining the interface to this kernel (if it is
+            polymorphic) and a list of the Schedule(s) representing the kernel
+            code.
+        :rtype: tuple[Optional[:py:class:`psyclone.psyir.symbols.Symbol`],
+                      list[:py:class:`psyclone.psyGen.KernelSchedule`]]
```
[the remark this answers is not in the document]
arporter: I started doing this but have hit a snag - currently `GenericInterfaceSymbol.routines` returns `List[tuple[:py:class:`psyclone.psyir.symbols.RoutineSymbol`, bool]]`. There's no way to get from that to the `Routine` 's themselves.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -178,6 +192,7 @@ def validate(self, node, options=None):
         # any existing Routine matches that required by the Call but for now
         # we live with the possibility of a false positive resulting in a
         # refusal to module inline.
+        parent_container = node.ancestor(Container)
         for routine in parent_container.walk(Routine, stop_type=Routine):
             if routine.name.lower() == kname.lower():
                 # Compare the routine to be inlined with the one that
```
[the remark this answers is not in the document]
arporter: Good spot, done.

src/psyclone/psyir/nodes/call.py
arporter: Ah! That seems to be an error. I need to make sure this is exercised by a test.

arporter: I've monkeypatched my way to a test for this.

arporter:
Finally got a profile with the same version of LFRic as used in the integration tests:
image
This is approximately one time-step. The big 'gap' at the end is for weighted_proj2thetav called from si_operators_alg_mod. I think this is where there are kernels with calls to relatively complex routines that we haven't handled yet.

arporter:
Ready for another look now. I haven't yet renamed `get_kernel_schedule` but will do this as a final step if you're happy with the approach.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -299,22 +311,10 @@ def _get_psyir_to_inline(node):
...
-            if len(routines) > 1:
-                raise TransformationError(
-                    f"The target of the call to '{caller_name}' cannot be "
-                    f"inserted because multiple implementations were found: "
-                    f"{[rout.name for rout in routines]}. TODO #924")
-        return (caller_name, routines[0])
+            return node.get_kernel_schedule()
+
+        # We have a generic routine call.
+        return node.get_callees()
```
sergisiso: I like the new return value of get_kernel_schedule(), it solves my main issue with the previous implementation that it was easy to miss. Now a kernesl[0] is required and it is the user responsibility to do this assumption.

Regarding the name, this method made me realise that we now return the same as `get_callees` which is great. We also plan to make codedkernels subclass of Call at some point, so it would make sense to rename `get_kernel_schedule` to `get_callees`? Otherwise, we will have 2 methods with the same intent when we change the hierarchy. Also, it hides the distinction from the transformations. For instance, this whole `_get_psyir_to_inline` can be removed and just use the `get_callees` in its place. This is good because this is a generic psyir transformation and is not ideal that it currently depends on CodedKern, which is a psykal component.

examples/lfric/scripts/gpu_offloading.py
```diff
@@ -59,9 +62,68 @@
...
 OFFLOAD_DIRECTIVES = os.getenv('LFRIC_OFFLOAD_DIRECTIVES', "none")
 
 
+def _inline_calls(kern):
+    '''
+    Recursively inline all calls within the supplied Kernel or Routine.
+
+    Currently only attempts to replace MATMUL intrinsic calls with inline
+    code.
+`
```
sergisiso: There is a leftover backtick in this line

examples/lfric/scripts/gpu_offloading.py
```diff
@@ -59,9 +62,68 @@
...
+    :param kern: the Kernel or Routine to inline any Calls into.
```
sergisiso: Kernel is 'kind of a Call' and routine is a Routine. Technically kernel is not necessarily a CodedKern, but in this case we only care about CodedKerns.

The different argument type make the implementation more complicated: e.g. get_kernel_schedule is called inside but get_callees, is called outside. The same with the recursion, get_kernel_schedule is called at the top while get_callees is called in the inner nest just before recursing.

I think it will be nicer to pick one or the other (call or routine) in both cases. Specially if you take the suggestion of using "get_callees" as a common method name.

examples/lfric/scripts/gpu_offloading.py
```diff
@@ -59,9 +62,68 @@
...
+            # the resulting TransformationError will signal (to the calling
+            # routine) that we are not to mark this kernel for offload.
+            if (isinstance(call, IntrinsicCall) and
+                    call.intrinsic == IntrinsicCall.Intrinsic.MATMUL):
+                matrans.apply(call)
+                continue
+
+            # For now we only look at MATMUL calls. In future we may want to
+            # remove this `continue` and attempt to inline more calls.
+            continue
```
sergisiso: I see this is stopped here, but the inlining transformations below are already in try/excepts. Is something failing that the try/except misses? I am asking just to know, this PR is already a big step if removing this continue needs other fixes it can be in a follow up.

examples/lfric/scripts/kernel_print.py
```diff
@@ -55,11 +55,12 @@ def trans(psyir):
     # Loop over all of the Kernels Calls
     for kernel in psyir.coded_kernels():
         try:
-            kernel_schedule = kernel.get_kernel_schedule()
-            if kernel_schedule not in already_printed:
-                kern = fortran_writer(kernel_schedule)
-                print(kern)
-                already_printed.append(kernel_schedule)
+            kernel_schedules = kernel.get_kernel_schedule()
+            for ksched in kernel_schedules:
```
sergisiso: Maybe merge these lines, if the expression is only used once we don't need a variable for it.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -436,28 +447,33 @@ def apply(self, node, options=None):
         # may already be in use, but the equality check below guarantees
         # that if it exists it is only valid when it references the exact same
         # implementation.
-        caller_name, code_to_inline = (
-            KernelModuleInlineTrans._get_psyir_to_inline(node))
-        callee_name = code_to_inline.name
-        codes_to_inline = [code_to_inline]
+        codes_to_inline = KernelModuleInlineTrans._get_psyir_to_inline(node)
+        interface_sym = codes_to_inline[0].symbol_table.lookup(
+            caller_name) if len(codes_to_inline) > 1 else None
```
sergisiso: I think this inlined if else is a bit hard to read, maybe:
```
interface_sym = None
if len(codes_to_inline) > 1:
   interface_sym = codes_to_inline[0].symbol_table.lookup(caller_name)
```

sergisiso: There is also a small caveat here, since we are using the caller name in the callee scope we want to use  "node.routine.interface.orig_name" and not "node.routine.symbol.name" to be resistant to renaming imports. It is not as simple as changing caller_name because I see below is used for callsite_table. So maybe we need a new "caller_external_name" which is "caller_name" unless a interface.orig_name is available?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -436,28 +447,33 @@ def apply(self, node, options=None):
...
+                called_sym = callsite_table.lookup(caller_name,
+                                                   otherwise=None)
+                if (not called_sym or called_sym is not routine.symbol or
+                        (called_sym.is_import or called_sym.is_unresolved)):
+                    # This routine is not module-inlined.
+                    break
+            else:
+                # All routines are module-inlined so there's nothing to do.
+                # TODO #11 - log this.
+                return
```
sergisiso: Why don't we do this in the validate and return a TransformationError?

src/psyclone/domain/lfric/lfric_kern.py
```diff
@@ -780,72 +781,83 @@ def gen_stub(self) -> Container:
 
         return stub_module
 
+    def get_interface_symbol(self) -> Union[GenericInterfaceSymbol, None]:
```
sergisiso: `Optional[GenericInterfaceSymbol]` ?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -807,6 +807,7 @@ def is_available_on_device(self):
             IntrinsicCall.Intrinsic.PRODUCT, IntrinsicCall.Intrinsic.SIZE,
             IntrinsicCall.Intrinsic.SUM, IntrinsicCall.Intrinsic.LBOUND,
             IntrinsicCall.Intrinsic.MAXVAL, IntrinsicCall.Intrinsic.MINVAL,
+            IntrinsicCall.Intrinsic.MATMUL,
```
sergisiso: Just checking if this is intentional or a leftover, in the transformation script you inline these saying that they sometimes cause issues with the nvidia compiler.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -497,6 +503,40 @@ def test_module_inline_apply_kernel_in_multiple_invokes(tmpdir):
...
+    schedule1 = psy.invokes.invoke_list[0].schedule
+    for coded_kern in schedule1.walk(CodedKern):
+        if coded_kern.name == "mixed_code":
+            inline_trans.apply(coded_kern)
+            # Transform that kernel. We have to use 'force' as it contains
+            # a CodeBlock.
+            artrans.apply(coded_kern, options={"force": True})
+    output = str(psy.gen).lower()
+    assert "subroutine mixed_code_32" in output
+    assert "!$acc routine seq" in output
```
sergisiso: could we do `assert output.count("!$acc routine seq") == 2` ?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -405,7 +411,7 @@ def _rename_import(table: SymbolTable,
                 csym, orig_name=name))
 
     def apply(self, node, options=None):
-        ''' Bring the kernel subroutine into this Container.
+        ''' Bring the kernel/subroutine into this Container.
```
sergisiso: Maybe "Bring the implementation of this kernel/call into this container." (To not confuse what it takes as node)

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -471,18 +479,16 @@ def test_module_inline_apply_kernel_in_multiple_invokes(tmpdir):
...
     for coded_kern in schedule1.walk(CodedKern):
         if coded_kern.name == "testkern_qr_code":
             inline_trans.apply(coded_kern)
+            artrans.apply(coded_kern)
     gen = str(psy.gen)
 
-    # After this, one invoke uses the inlined top-level subroutine
-    # and the other imports it (shadowing the top-level symbol)
-    assert gen.count("use testkern_qr_mod, only : testkern_qr_code") == 1
-    assert gen.count("end subroutine testkern_qr_code") == 1
```
sergisiso, line 483: Can we put this back? Are this 0 (or not in gent) and 1 now?

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -497,6 +503,40 @@ def test_module_inline_apply_kernel_in_multiple_invokes(tmpdir):
...
+        if coded_kern.name == "mixed_code":
+            inline_trans.apply(coded_kern)
+            # Transform that kernel. We have to use 'force' as it contains
+            # a CodeBlock.
+            artrans.apply(coded_kern, options={"force": True})
+    output = str(psy.gen).lower()
+    assert "subroutine mixed_code_32" in output
+    assert "!$acc routine seq" in output
+    assert "subroutine mixed_code_64" in output
+    # Since we don't currently rename module-inlined kernels, module-inlining
```
sergisiso: Mention the issue number that will resolve this

src/psyclone/transformations.py
```diff
@@ -412,89 +410,79 @@ def validate_it_can_run_on_gpu(self, node, options):
...
-                    raise TransformationError(
-                        f"Cannot apply {self.name} to kernel '{node.name}' as "
-                        f"it has multiple implementations - TODO #1946")
-
             k_or_r = "Kernel"
         else:
             # Supplied node is a PSyIR Routine which *is* a Schedule.
-            kernel_schedule = node
+            kernel_schedules = [node]
             k_or_r = "routine"
```
sergisiso, line 426: This also seems to take a kern (call) or a routine. Nemo needs routines as we search them by name since they may not be in the same file than the caller. Would it be possible to bring the `kernel_schedules = node.get_kernel_schedule()` outside in LFRic where this is called to be consistent? We can still do:
```
if isinstance(node, KernScedule):
```
if it helps to give different error messages.

(Can be a separate PR is this is getting to big)

sergisiso: I see below that this is also the case in OMPDeclareTargetTrans, so maybe a separate PR

sergisiso (changes requested):
This is getting close @arporter, it is just minor comments now.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes, that's a really good idea. I've renamed it and removed `_get_psyir_to_inline`. Happy days.

examples/lfric/scripts/gpu_offloading.py
arporter: Oops.

examples/lfric/scripts/gpu_offloading.py
arporter: I've simplified this now since there's a bit more to do to get full inlining working (currently the transformation rejects a Kernel as it wants a Routine).

examples/lfric/scripts/gpu_offloading.py
arporter: Yes, simplified for now.

examples/lfric/scripts/kernel_print.py
arporter: Done.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes, that was a very good spot. I've added two new tests and updated the code so that it works for renamed imports. It's a bit hot and it's complicated code so I may need to review what I've done.

(I also realised that I needed to forbid an attempt to module-inline a polymorphic routine if the call site didn't have a parent Container that I could declare the interface in.)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Good point, done.

src/psyclone/domain/lfric/lfric_kern.py
arporter: I thought you didn't like `Optional` on a return value (or perhaps that was Rupert)? I'm happy though so done.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: I wasn't sure about this - it only causes issues in LFRic where there's mixed precision. Otherwise it's fine. However, I've commented it out again now and left a comment.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
arporter: Yes!

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Done.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
arporter: I can't remember what happened there but it seems they are back already. Presumably as part of a change I did earlier.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
arporter: Done.

src/psyclone/transformations.py
arporter: Yes, I started doing this and it got out of hand. As things stand, it would become the responsibility of the person writing the transformation script to do `node.get_callees()`, allow for cases where the PSyIR can't be obtained, and then call the transformation on each result. That doesn't feel very nice.

arporter:
Ready for another look now. CodeCov is doing weird things but perhaps it will get sorted out if master moves on and then we bring this branch up to date.

sergisiso:
@arporter since you are on AL I fixed the merge conflicts and a leftover get_kernel_schedule myself as I want to proceed with the integration tests for this PR.

sergisiso (approved):
@arporter All my previous comments have been addressed, the coverage was complete and the integration tests all passed (the system was noisy but there seems to be a slowdown by doing more kernels on the gpu, this is acceptable as the coverage is significantly increased and we don't do the code-inline and atomics yet).

Good job @arporter this has been a long PR but I like how it has end up. This opens the door to do the code-inlining of the mixed kernels and the deprecation of okern.

PR #2472, opened 2024-01-12
2471 update copyright

hiker:
As discussed. The script updated all:

Updating python scripts
Updating documentation
Updating shell scripts
Updating Makefiles
Updating Fortran programs  <--  [FfXx]90
Updating READMEs

If I should have missed any types, it's easy to add ;)

PR #2743, opened 2024-10-10
(Closes #2742) Sort the ordering of function spaces in LFRic algorithms

### LINKED ISSUE #2742: [LFRic] [PSyAD] Function space pointers declared and assigned in random orders in generated adjoint test
This is a critical issue that prevents the development of further patches for LFRic adjoint test algorithms.

Essentially if you have a test which generates the following function space code

```
function_space_1_ptr => get_fs(mesh, element_order, fs_1_id)
function_space_2_ptr => get_fs(mesh, element_order, fs_2_id)
```

you can write a patch that will expect the generated ordering as `fs_1` followed by `fs_2`. Making a patch with these in the hunks will result in patching errors 50% of the time because PSyAD will eventually set the ordering to

```
function_space_2_ptr => get_fs(mesh, element_order, fs_2_id)
function_space_1_ptr => get_fs(mesh, element_order, fs_1_id)
```

There are developments in LFRic that are going to require the patching of these function space pointers, whereas previously it was avoided. These function space pointers ought to be sorted in some way so that they consistently produce a correct ordering.

DrTVockerodtMO:
Closes [#2472](https://github.com/stfc/PSyclone/issues/2742), which has a description of the problem.

### Reason of the problem
When the function spaces are added to the adjoint test algorithms, duplicates are removed by using the `set` python function. This has the side effect of not respecting the ordering of the original list, so between different calls of `PSyAD` this results in the function space declarations and assignments to have a random ordering.

### Solution
The solution is to sort the `fspaces` variable as @arporter suggested in [#2472](https://github.com/stfc/PSyclone/issues/2742).

### Testing
I modified the `src/psyclone/tests/domain/lfric/algorithm/lfric_alg_test.py` file to do a few things. Firstly, I modified the input to use a `set` version of the original `list`, which will ensure that it will shuffle. I modified the function space assertions to check if the `w1` pointer is declared/assigned before the `w3` pointer in the generated code.

### Test results
Without the sorting of the `fspaces` variable, the modified test will occassionally fail because of the randomised ordering, which reflects the current situation when trying to patch the adjoint tests.

With the sorting of the `fspaces` variable, the modified test now passes consistently.


arporter (commented):
Thanks very much for doing this @DrTVockerodtMO. Everything looks good. I'll run the integration tests just to be on the safe side and then I'll merge it.

arporter (approved):
Integration tests were green.
Will proceed to merge.

PR #2746, opened 2024-10-14
Update CI software versions

nmnobre:


nmnobre:
Hi Andy, this should now be ready to go :)

Note that the following is now a bit outdated, but I'll leave it to you to decide if it's worth the change:

https://github.com/stfc/PSyclone/blob/a11a3899cf387eb93dfeefb5983be36ace478785/doc/psyad/user_guide/psyad_command.rst?plain=1#L64-L86

arporter (approved):
Thanks Nuno, that all looks fine. Since no KGOs have changed there's not much to check :-)
I've set your latest integration test battery running in case that's important.
So long as that's green then this can go on.

arporter:
(Oh, and I don't feel the need to update the user guide for the change in formatting of the help message. I couldn't see what had changed until I read your comment in the code :-) )

arporter:
Integration tests were green (modulo the race condition in NEMOv4).

arporter:
@nmnobre unfortunately this has broken the ReadTheDocs builds. They don't accept Python 3.13:
image
Would you mind fixing this?

nmnobre:
> Integration tests were green (modulo the race condition in NEMOv4).

Wait? Race condition? So this is/was known?

arporter:
> > Integration tests were green (modulo the race condition in NEMOv4).
> 
> Wait, race condition? So this is/was known?

Yep. Sergi is working on tracking it down (well, when he's not on leave).

Issue #2244, opened 2023-08-03
[LFRic] Add loop tiling transformation


tinyendian:
A new `mesh_tiling` module will soon be available in LFRic, which adds methods to `mesh` objects that PSyclone can use for generating OpenMP-parallelised tiled loops, such as
```
if (mesh%is_tiled()) then
  tmap => mesh%get_coloured_tiling_map()
  do colour = 1, mesh%get_ntilecolours()
    !$omp parallel default(shared), private(tile, cell)
    !$omp do schedule(static)
    do tile = 1, mesh%get_last_halo_tile_per_colour(colour, 1)
      do cell = 1, mesh%get_last_halo_cell_per_colour_and_tile(colour, tile, 1)
        call kernel(tmap(colour, tile, cell), ...)
      end do
    end do
    !$omp end do
    !$omp end parallel
  end do
else
  ! Use coloured loop to ensure thread safety
end if
```
The API is very similar to the existing colouring API, with tiles added as an extra dimension. Tiling is not guaranteed to be available for each mesh (e.g., some multigrid levels may be too coarse for tiling, or a user decides not to configure it), so colouring needs to be kept as a fall-back option to ensure thread-safety. Tiling is also unlikely to be beneficial for LFRic kernels where no function space shares DOFs between neighbour cells, so it can probably be restricted to kernels that need colouring.

Documentation can be found here: https://code.metoffice.gov.uk/trac/lfric/wiki/LFRicInfrastructure/MeshTiling
The implementation ticket is here: https://code.metoffice.gov.uk/trac/lfric/ticket/3572
An example of a tiled loop implementation is here: https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/branches/dev/wolfganghayek/r41055_colour_tiling/gungho/source/psy/psykal_lite_mod.F90?rev=43701#L1395

hiker:
I am not sure if we would want to combine tiling with colouring, since tiling might(??) be useful for single threaded setups as well, or? Do you have any idea? If it makes sense, could you give a code example of usage of tiling without colouring (e.g. I would assume that `do tile = 1, mesh%get_last_halo_tile_per_colour(colour, 1)` will be different).

What is the difference between `mesh%get_ntilecolours()` and `mesh%get_ncolours()`? I assumed that the colour loop would be identical, but apparently that's not the case.

In general, I would expect that there will be four transformation applied in the following order (based on what LFRic does atm):
```
for loop in all_loops:
    if ...
[...]

PR #2750, opened 2024-10-22
(related to #2244) LFRic tiled colouring

sergisiso:
Currently this PR is just to monitor the changes, tests and coverage or the LFRic tiled colouring implementations. It is unclear if we should merge this now or wait until the colouring code is refactored so that it can be encapsulated inside its associated transformation instead of spread it around the code.


sergisiso:
I am preparing a new implementation of tiled colouring branching from the refactored LFRic PSyIR backend. Closing this one

Issue #2717, opened 2024-09-18
Merge PSyACC functionality into PSyclone


jwallwork23:
The https://github.com/MetOffice/psyacc repository provides tools for automating OpenACC GPU porting efforts using PSyclone. Rather than having a separate repository, we should merge this functionality into PSyclone.

I should have some time to work on this late October.

Issue #2661, opened 2024-07-15
FFSL transport in LFRic depends on a kernel that loops over halo cells


tommbendall:
One crucial part of the FFSL transport scheme in LFRic involves a kernel which loops over only cells in the halos. This is obviously not currently supported by PSyclone, so is currently done through a psykal_lite implementation: see the `invoke_ffsl_panel_swap_kernel_type` here: https://code.metoffice.gov.uk/trac/lfric_apps/browser/main/trunk/science/gungho/source/psy/psykal_lite_transport_mod.F90?rev=755#L786

To summarise what needs to happen here: we need to perform a halo exchange to the full halo depth, and loop over all halo cells (and not owned cells)

This loop over the halo cells doesn't really fit in the current design of LFRic kernels, but to avoid a psykal_lite implementation, I think we need new metadata. Two first suggestions I have thought of are:
- having a `GH_HALOWRITE` access permission
- introducing a `CELL_HALOCOLUMN` value for the `operators_on` variable

Mea culpa: I should've opened this issue several months ago when the code was added to trunk. I overlooked it and recently remembered, so sorry about that!

tommbendall:
I thought I'd just add another comment here, as there is a whole category of kernels that currently have psykal lite implementations because they loop to some depth in the halo regions, before setting the output field to be clean (to ensure that halo swaps don't happen subsequently on these fields).

It might be helpful to tackle this other problem together, with whatever solution is best for this issue

tommbendall:
Thanks Andy, I agree with that summary and that it makes most sense to add this through the `operates_on` variable. I suppose that the invoke would also require a `depth` argument.

arporter:
Currently, the docs for `operates_on` say:
![image](https://github.com/user-attachments/assets/34007bd4-cec2-4644-9182-d8f47c8d448f)
We need to extend this to support the two further cases given above. Really, `CELL_COLUMN` is effectively `OWNED_CELL_COLUMN` and then we would naturally have `HALO_CELL_COLUMN` and `OWNED_PLUS_HALO_CELL_COLUMN`. For backwards compatibility we can make `OWNED_CELL_COLUMN` a synonym for the existing `CELL_COLUMN`.

As Tom says, both of the new values of 'operates-on' will require a halo depth to be supplied. If this is only known at run-time then it will need to be passed to the kernel in the invoke, e.g.:

    call invoke(a_new_kern(halo_depth, field1, field2))

[...]

PR #2753, opened 2024-10-23
Add `is_character` method for `DataNode`

jwallwork23:
Partially addresses #2717.

This PR merges one of two functions from the `types` module of [PSyACC](https://github.com/MetOffice/psyacc) into PSyclone.

### DIFF
--- a/src/psyclone/psyir/nodes/datanode.py
+++ b/src/psyclone/psyir/nodes/datanode.py
@@ -67,12 +67,12 @@ class DataNode(Node):
         :returns: True if this DataNode is a character, otherwise False.
         :rtype: bool
 
-        :raises Exception: if the intrinsic type cannot be determined.
+        :raises ValueError: if the intrinsic type cannot be determined.
 
         '''
         if not hasattr(self.datatype, "intrinsic"):
             if unknown_as is None:
-                raise Exception(
+                raise ValueError(
                     "is_character could not resolve whether the expression"
                     f" '{self.debug_string()}' operates on characters."
                 )
--- a/src/psyclone/psyir/nodes/datanode.py
+++ b/src/psyclone/psyir/nodes/datanode.py
@@ -57,16 +57,22 @@ class DataNode(Node):
         '''
         return UnresolvedType()
 
-    def is_character(self, unknown_as=False):
+    def is_character(self, unknown_as=None):
         '''
         :param unknown_as: Determines behaviour in the case where it cannot be
-            determined whether the DataNode is a character.
+            determined whether the DataNode is a character. Defaults to None,
+            in which case an exception is raised.
         :type unknown_as: bool
         :returns: True if this DataNode is a character, otherwise False.
         :rtype: bool
 
         '''
         if not hasattr(self.datatype, "intrinsic"):
+            if unknown_as is None:
+                raise Exception(
+                    "is_character could not resolve whether the expression"
+                    f" '{self.debug_string()}' operates on characters."
+                )
             return unknown_as
         return (
             self.datatype.intrinsic == ScalarType.Intrinsic.CHARACTER
--- a/src/psyclone/psyir/nodes/datanode.py
+++ b/src/psyclone/psyir/nodes/datanode.py
@@ -32,12 +32,13 @@
 # POSSIBILITY OF SUCH DAMAGE.
 # -----------------------------------------------------------------------------
 # Authors: A. R. Porter and S. Siso, STFC Daresbury Lab
+# Modified J. G. Wallwork, University of Cambridge
 # -----------------------------------------------------------------------------
 
 ''' This module contains the DataNode abstract node implementation.'''
 
 from psyclone.psyir.nodes.node import Node
-from psyclone.psyir.symbols.datatypes import UnresolvedType
+from psyclone.psyir.symbols.datatypes import ScalarType, UnresolvedType
 
 
 class DataNode(Node):
@@ -55,3 +56,14 @@ class DataNode(Node):
         :rtype: :py:class:`psyclone.psyir.symbols.UnresolvedType`
         '''
         return UnresolvedType()
+
+    @property
+    def is_character(self):
+        '''
+        :returns: True if this reference is a character, otherwise False.
+        :rtype: bool
+
+        '''
+        return (
+            self.datatype._intrinsic == ScalarType.Intrinsic.CHARACTER
+        )
--- a/src/psyclone/psyir/nodes/datanode.py
+++ b/src/psyclone/psyir/nodes/datanode.py
@@ -62,9 +62,12 @@ class DataNode(Node):
         :param unknown_as: Determines behaviour in the case where it cannot be
             determined whether the DataNode is a character. Defaults to None,
             in which case an exception is raised.
-        :type unknown_as: bool
+        :type unknown_as: Optional[bool]
+
         :returns: True if this DataNode is a character, otherwise False.
         :rtype: bool
+
+        :raises Exception: if the intrinsic type cannot be determined.
 
         '''
         if not hasattr(self.datatype, "intrinsic"):
--- a/src/psyclone/psyir/nodes/datanode.py
+++ b/src/psyclone/psyir/nodes/datanode.py
@@ -57,13 +57,17 @@ class DataNode(Node):
         '''
         return UnresolvedType()
 
-    @property
-    def is_character(self):
+    def is_character(self, unknown_as=False):
         '''
-        :returns: True if this reference is a character, otherwise False.
+        :param unknown_as: Determines behaviour in the case where it cannot be
+            determined whether the DataNode is a character.
+        :type unknown_as: bool
+        :returns: True if this DataNode is a character, otherwise False.
         :rtype: bool
 
         '''
+        if not hasattr(self.datatype, "intrinsic"):
+            return unknown_as
         return (
-            self.datatype._intrinsic == ScalarType.Intrinsic.CHARACTER
+            self.datatype.intrinsic == ScalarType.Intrinsic.CHARACTER
         )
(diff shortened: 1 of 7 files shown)


src/psyclone/psyir/nodes/reference.py
```diff
@@ -94,6 +95,17 @@ def is_array(self):
...
 
+    @property
+    def is_character(self):
+        '''
+        :returns: True if this reference is a character, otherwise False.
+        :rtype: bool
+
+        '''
+        return (
+            self.symbol.datatype._intrinsic == ScalarType.Intrinsic.CHARACTER
```
sergisiso, line 106: Note that you would also need to use: `self.datatype` for References anyway because not all symbols have a datatype attribute (alternatively you would need the proper checks here).

Similarly, not all datatypes have "intrinsic" (you can check it with `hasattr(self.datatype, "intrinsic")`).

Which raises the question about what should it return if "we don't know if its character or not". For example for:
```
subroutine test
    use other_module, only: a, b, c
    a = b + c
end subroutine
```

- One solution would be to return True or False and rename the method as could_be_character / guaranteed_to_be_character . This would be fine but it is unclear to me what is the best as I can see scenarios for both.

- Another solution would be to return something different or raise an Error. But this leads to code that always needs the "checks" or "try excepts" around it, which is easy to forget.

- A third solution is to add an argument to chose what to return. e.g. `reference.is_character(unknowns_as=True)`. The drawback is that it is not a "property" anymore, but the argument can be omitted if we agree in a good default.

What do you think?

src/psyclone/tests/psyir/nodes/reference_test.py
```diff
@@ -149,6 +155,13 @@ def test_reference_is_array():
     assert ref.is_array is True
 
 
+def test_reference_is_character():
+    '''Test that a character reference is marked correctly.
+    '''
+    reference = Reference(DataSymbol("test", CHARACTER_TYPE))
+    assert reference.is_character
```
sergisiso, line 162: As mentioned before, I would like to see the test extended for a symbol without datatype:
`reference = Reference(Symbol("test")`
and a datatype without intrinsic:
`reference = Reference(DataSymbol("test", UnresolvedType()))`

sergisiso (changes requested):
Thanks for bringing PSyACC functionality @jwallwork23 , my main worry with this PR is what should it return when it is unknown if the reference is a character or not while being clear of it's behaviour for the caller. I added some alternatives to the comments which we could discuss.

src/psyclone/psyir/nodes/reference.py
sergisiso: Thinking a bit more about this I don't like the "return something different" because this "something" needs to be imported or be None. While none/null would be the prefered solution in other languages, in python None evaluates to False in conditions, so `if reference.is_character:` will behave as returning False for Unknowns which could be surprising if not familiar with the API. So I am leaning towards the third option because `if reference.is_character():` will work as expected for users until it finds a complex case, in that case it would raise and error that can inform about the unkonws_as optional argument to the user, and there is nothing unexpected in its behaviour.

This is consistent with the "otherwise" parameter that we introduced in the symbol table and maybe could be a rule/style for similar methods. What do @arporter @hiker think?

arporter:
Could I be a bit cheeky and request a fix for a docstring formatting error from a previous PR? Building the Ref. Guide now gives:
```
PSyclone/src/psyclone/psyir/nodes/assignment.py:docstring of psyclone.psyir.nodes.Assignment.is_literal_assignment:3: WARNING: Definition list ends without a blank line; unexpected unindent.
```
which is new!

src/psyclone/psyir/nodes/reference.py
arporter: I like the third option too. In #2661 I have also hit the problem of wanting to query something that can be True, False or Unknown. I've temporarily introduced a `Fuzzy` enum that contains values for TRUE, FALSE and MAYBE but I'm unsure yet whether this is a good solution.

jwallwork23:
> Could I be a bit cheeky and request a fix for a docstring formatting error from a previous PR? Building the Ref. Guide now gives:
> 
> ```
> PSyclone/src/psyclone/psyir/nodes/assignment.py:docstring of psyclone.psyir.nodes.Assignment.is_literal_assignment:3: WARNING: Definition list ends without a blank line; unexpected unindent.
> ```
> 
> which is new!

Ah good spot. Apologies, should be fixed in f2952478fc01c8e8d78cad0544dc43efd562054a.

src/psyclone/tests/psyir/nodes/reference_test.py
jwallwork23: Extended the testing of this functionality in 772d3ff1726e96db078b7d9fbc4e7c0a98e2f3a8.

src/psyclone/psyir/nodes/reference.py
jwallwork23: Thanks. I implemented what I think you mean in 772d3ff1726e96db078b7d9fbc4e7c0a98e2f3a8, where `unknown_as` is a `bool` that gives the return value if it can't be determined whether the `DataNode` is a character. However, you mention "error by default" so should the flag instead be about whether to raise an error if the value is unknown?

src/psyclone/psyir/nodes/reference.py
```diff
@@ -94,6 +95,17 @@ def is_array(self):
...
+        )
+
```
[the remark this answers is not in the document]
jwallwork23: Addressed in 88f592eb0457783cc16d8b8ae09a748db2cf1933.

src/psyclone/psyir/nodes/reference.py
sergisiso: What I meant is for the `unknown_as` to be of type `Optional[bool]` and default to `=None`

In the implementation:

- if it is none and an unresolvable case is found, raise and error saying something like "is_character could not resolve whether the expression {self.debug_string()} operates on characters. Note that 'is_character' has the optional argument 'unknown_as' to provide a fallback value to return in this case."

- If it is not none, just return "unknown_as" as you do now.

sergisiso:
> Ah good spot. Apologies, should be fixed in https://github.com/stfc/PSyclone/commit/f2952478fc01c8e8d78cad0544dc43efd562054a.

Ups, I missed this on my review. I think the issue comes from missing the starting colon in the `:returns:` and `:rtype:` tags. It is also wrong in the method above.

With this fixed I don't think the backslash is necessary.

src/psyclone/psyir/nodes/reference.py
jwallwork23: Ah I see, thanks. Addressed in ae82fd765c5c5a3b1bde00c660b617ef701a83ec.

jwallwork23:
> Ups, I missed this on my review. I think the issue comes from missing the starting colon in the `:returns:` and `:rtype:` tags. It is also wrong in the method above.
> 
> With this fixed I don't think the backslash is necessary.

Oh okay, I had just copied the syntax from the method above it. Fixed for both methods in 2c783a787f0b8daacf14f385b75e1d74290e0f65.

src/psyclone/psyir/nodes/datanode.py
```diff
@@ -55,3 +56,24 @@ def datatype(self):
...
+
+    def is_character(self, unknown_as=None):
+        '''
+        :param unknown_as: Determines behaviour in the case where it cannot be
+            determined whether the DataNode is a character. Defaults to None,
+            in which case an exception is raised.
+        :type unknown_as: bool
+        :returns: True if this DataNode is a character, otherwise False.
+        :rtype: bool
+
```
sergisiso, line 68: In the docstring we typically add a blank line between the arguments and the return tags, then another blank line and the "raises" tags. It also misses the Optional type (since it accepts None as value). In this case it should be:
```
...
:type unknown_as: Optional[bool]

:returns: True if this DataNode is a character, otherwise False.
:rtype: bool

:raises Exception:  if the intrinsic type cannot be determined.
```

src/psyclone/tests/psyir/nodes/reference_test.py
```diff
@@ -149,6 +150,28 @@ def test_reference_is_array():
...
+    assert not reference.is_character()
+
+    reference = Reference(DataSymbol("unknown", UnresolvedType()))
+    with pytest.raises(Exception) as excinfo:
+        _ = reference.is_character()
+    assert ("is_character could not resolve whether the expression 'unknown'"
+            " operates on characters." in str(excinfo.value))
+    reference = Reference(DataSymbol("unknown", UnresolvedType()))
+    assert not reference.is_character(unknown_as=False)
+    assert reference.is_character(unknown_as=True)
```
sergisiso, line 172: We have a rule to put whenever possible to test in a matching path to its implementation, so for a method in
`src/psyclone/psyir/nodes/datanode.py`
we expect the test to be located in:
`src/psyclone/tests/psyir/nodes/datanode_test.py`

Which make them easier to find in the future.

Also s/reference/datanode/ in the test name.

sergisiso (changes requested):
Thanks for addressing the comments @jwallwork23. I like how this is implemented now. I just added a couple more comments related with our codebase style guide. But it should be ready after this.

src/psyclone/psyir/nodes/datanode.py
jwallwork23: Ah thanks. Applied in c31018b60.

src/psyclone/tests/psyir/nodes/reference_test.py
jwallwork23: Moved in 5c17b4a3c23acb823bee6e667f92b5027292e6df.

jwallwork23:
> This is ready to merge now @jwallwork23 . Note that I ended up changing the "raise Exception" to a "raise ValueError" to avoid the pylint "broad-exception-raised" issue.
> 
> I also built the multiple documentations to verify the docstring warnings are gone.

Great, thanks @sergisiso!

PR #2754, opened 2024-10-23
Port over loop handling utils from PSyACC

jwallwork23:
Partially addresses #2717.

This PR merges functions from the `loop` module of [PSyACC](https://github.com/MetOffice/psyacc) into PSyclone:
* `is_outermost` checks that the Loop doesn't have any ancestors which are Loops.
* `is_perfectly_nested` checks that the Loop nest starting at a given Loop is "loops all the way down" until the last.
* `is_simple` checks that a Loop nest is perfect and that the deepest level only contains literal assignments (i.e., the loop nest is trivially parallelisable).
* `nest_variable_names` extracts the tuple of variable names for the Loop nest starting at a given Loop.


sergisiso:
@jwallwork23 Isn't this equivalent to the Loop.independent_iterations method? (I think we considered this name for that method but decided the independent iterations could be more generic)

jwallwork23:
> @jwallwork23 Isn't this equivalent to the Loop.independent_iterations method? (I think we considered this name for that method but decided the independent iterations could be more generic)

Oh good, if there is already an equivalent method then that's one less thing to merge in :) Thanks!

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +583,109 @@ def gen_code(self, parent):
...
+            :param list1: the first list to intersect
+            :type list1: list
+            :param list2: the second list to intersect
+            :type list2: list
+            :returns: the intersection of the two lists
+            :rtype: list
+            '''
+            return [item for item in list1 if item in list2]
+
+        # Check whether the subnest is perfect by checking each level in turn
```
arporter, line 631: This looks pretty complicated. I think you'd be better off using the `loop_body` property of `Loop` rather than doing a `walk`. Could you use/adapt the implementation at:
https://github.com/stfc/PSyclone/blob/8367de1fbce19bbd0fa64e2dc2ad2bfcd0ed6e82/src/psyclone/psyir/transformations/parallel_loop_trans.py#L271-L325
(and then replace/update that code to use this new method)?

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +583,109 @@ def gen_code(self, parent):
...
+            isinstance(child, Assignment) and child.walk(Literal)
+            for child in children
+        )
+
+    @property
+    def nest_variable_names(self):
+        '''
+        :returns: Names of all variables within the Loop nest of descendents
+            of this Loop, inclusive.
+        :rtype: tuple
```
arporter, line 689: `tuple[str]` please.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -43,6 +44,10 @@
 from psyclone.psyir.nodes.statement import Statement
 from psyclone.psyir.nodes.routine import Routine
 from psyclone.psyir.nodes import Schedule
+from psyclone.psyir.nodes.assignment import Assignment
+from psyclone.psyir.nodes.intrinsic_call import IntrinsicCall
+from psyclone.psyir.nodes.literal import Literal
+from psyclone.psyir.nodes.reference import Reference
 from psyclone.psyir.symbols import ScalarType, DataSymbol
 from psyclone.core import AccessType, Signature
```
arporter, line 52: Not yours but would you mind putting all of these in alpha order in fact?

arporter (changes requested):
Thanks @jwallwork23. I have a couple of questions about two of the new methods so I haven't gone on to look at the tests yet.
Ref. guide builds cleanly though.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -43,6 +44,10 @@
 from psyclone.psyir.nodes.statement import Statement
 from psyclone.psyir.nodes.routine import Routine
 from psyclone.psyir.nodes import Schedule
+from psyclone.psyir.nodes.assignment import Assignment
```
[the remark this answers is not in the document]
jwallwork23: Addressed in f703b5df0.

src/psyclone/psyir/nodes/loop.py
jwallwork23: Thanks @arporter, what do you think of 97d1143ce?

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +583,109 @@ def gen_code(self, parent):
...
+                        break
+                else:
+                    continue
+
+            # Otherwise, the nest level is not okay
+            return False
+        return True
+
+    @property
+    def is_simple(self):
```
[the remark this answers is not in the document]
jwallwork23: In the UKCA GPU port we collapsed and applied gang and vector clauses to any "simple" loop that just initialised to zero. For other loop nests, we'd analyse whether they contained loops over a certain variable (denoting looping over the spatial domain/chunk). Don't mind dropping this here, though - see 4a23ed5418a608550b2e9fb806accab6c177429f.

src/psyclone/psyir/nodes/loop.py
jwallwork23: Addressed in 2dffe5ee2.

src/psyclone/psyir/nodes/loop.py
jwallwork23: Addressed in f703b5df0.

arporter: Thanks.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +579,53 @@ def gen_code(self, parent):
...
+    @property
+    def is_outermost(self):
+        '''
+        :returns: True if the loop has no ancestors, otherwise False
+        :rtype: bool
+        '''
+        return self.ancestor(Loop) is None
+
+    @property
+    def is_perfectly_nested(self):
```
arporter, line 592: On discussion with @sergisiso I realised I should have asked what the use case for this method is? We've recently improved the 'collapse' option on ParallelLoopTrans so that it is a fair bit smarter:
image
Will that cover what you need or is this new method used in some other way?

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +579,53 @@ def gen_code(self, parent):
...
+                        isinstance(next_loop.loop_body[0], Loop)):
+                    return False
+                return len(self.walk(Loop)) == loop_count
+
+            # If there are no more loops then check all the expected loops
+            # have been counted.
+            next_loop = next_loop.loop_body[0]
+            if not isinstance(next_loop, Loop):
+                return len(self.walk(Loop)) == loop_count
+        return True
```
arporter, line 622: If you still need this method then I suggest a little bit of re-factoring to address the missed coverage: could you make the loop `while True:` and instead of the two `return len(self.walk(Loop)) == loop_count` statements have `break`. Then L622 would do `return len(self.walk......` instead.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +579,53 @@ def gen_code(self, parent):
...
+
+    @property
+    def nest_variable_names(self):
```
arporter, line 625: I think this needs a better name as I interpreted it to mean "the names of all variables accessed within a loop nest". Perhaps "nest_loop_variables" and return Symbols rather than strings (or is that going to make using the results difficult?)?
Failing that, how about "nest_loop_variable_names"?

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +579,53 @@ def gen_code(self, parent):
...
+        '''
+        :returns: Names of all variables within the Loop nest of descendents
```
arporter, line 627: "...all loop variables within..."

src/psyclone/psyir/nodes/loop.py
```diff
@@ -578,3 +579,53 @@ def gen_code(self, parent):
...
+            of this Loop, inclusive.
+        :rtype: tuple[str]
```
arporter, line 629: Just to emphasise, I think it would be more in-keeping with the PSyIR to return Symbols here. However, if you can show me how you use this then that will inform my opinion :-)

arporter (changes requested):
Thanks Joe. Discussing with Sergi made me realise that you might not need the 'perfectly nested' method at all now that we've improved the 'collapse' option on the transformations - let me know what you think.

src/psyclone/psyir/nodes/loop.py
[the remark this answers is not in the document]
arporter: Thanks.

src/psyclone/psyir/nodes/loop.py
jwallwork23: Oh nice, I wasn't aware of this update - sounds really useful to pass `True`. I'll experiment with this on the UKCA GPU port work and get back to you.

jwallwork23: I notice that this will insert `collapse(1)` in cases where a collapse isn't possible. I'd thought that was invalid syntax but it seems to compile fine.

src/psyclone/psyir/nodes/loop.py
jwallwork23: I realised we don't actually use this functionality in the transformation script for the UKCA GPU port so can probably drop it.

PR #2763, opened 2024-10-30
Handle inconsistent OpenACC loop clauses

jwallwork23:
Partially addresses #2717.

In PSyACC, we disallow sequential clauses to be used in conjuction with gang and/or vector clauses because the combination doesn't make sense. This PR handles such clause combinations in a similar way.

I also noticed that the private `ACCLoopDirective._gang` (etc.) members are often accessed rather than the public properties `ACCLoopDirective.gang`, so addressed this. The new `ACCLoopDirective._check_clauses_consistent` method is used in a few places that seem sensible.


arporter (changes requested):
Thanks very much for this @jwallwork23. All looks fine apart from the fact that you've consistently misspelt "conjunction" as "conjuction" :-) Once that's fixed this will be good to go providing the integration tests are OK (I've set them running).

jwallwork23:
> Thanks very much for this @jwallwork23. All looks fine apart from the fact that you've consistently misspelt "conjunction" as "conjuction" :-) Once that's fixed this will be good to go providing the integration tests are OK (I've set them running).

Thanks for the review, @arporter. Ah, well at least I was consistent! Fixed in de3ba19c07736418ffb6d4a5359917b15360ded1.

arporter (approved):
All good now thanks.
Reference guide builds without new errors.
Integration tests were fine.
Will proceed to merge.

PR #2767, opened 2024-11-05
(Closes #2755) Bug fix for psyad failing to transform array assignment

### LINKED ISSUE #2755: [LFRic] [PSyAD] PSyAD no longer computes adjoint of vector assigments properly in PSyclone 3.0-dev
In testing PSyclone 3.0, we find that forward code that looks like

```
vector(:) = vector(:) + constant(:) * scalar
```

where `vector` and `scalar` are active no longer adjoints to

```
do idx = vector_ubound, vector_lbound, -1
  scalar = scalar + constant(idx) * vector(idx)
end do
```

but instead does

```
scalar = scalar + constant(:) * vector(:)
```

which fails to compile. The former behaviour is expected.

arporter:
A small change that (pleasingly) only involved me updating some code to remove an old restriction. @DrTVockerodtMO please could you try this branch and let me know whether it solves your problem?


arporter:
This needs a 'science' review by @DrTVockerodtMO and then I'll put it up for proper code review.

DrTVockerodtMO:
> This needs a 'science' review by @DrTVockerodtMO and then I'll put it up for proper code review.

Will be happy to, just need to figure out how to test specific branches on our system and I'll have a look!

arporter:
@DrTVockerodtMO and I have jointly test this branch against some kernels and everything looks OK. This is therefore ready for review. It's not a big change. Could be one for @hiker, @sergisiso, @LonelyCat124

arporter:
I've just set the integration tests going.

src/psyclone/psyad/transformations/preprocess.py
```diff
@@ -98,6 +88,20 @@ def preprocess_trans(kernel_psyir, active_variable_names):
...
 
+    # Replace array-ranges with explicit loops
+    for assignment in kernel_psyir.walk(Assignment):
+        if node_is_passive(assignment, active_variable_names):
+            # No need to modify passive assignments
+            continue
+        try:
+            arrayassign2loops_trans.apply(assignment)
+        except TransformationError:
+            # Double-check that the transformation succeeded in
```
hiker, line 99: That feels very hard to understand. After some digging around, I assume this happens if the assignment does not contain a range? Either it needs better documentation - atm it sounds like the transformation succeeded, but raised an exception(??) - while I think it should mean "if the transformation couldn't be applied because it wasn't an array range"?
Or perhaps a slight code change: do the walk first to verify if there is a range, then call the transformation?

src/psyclone/tests/psyir/nodes/array_mixin_test.py
```diff
@@ -688,12 +689,17 @@ def test_get_effective_shape(fortran_reader):
...
+    assert "include a function call or unsupported feature" in str(err.value)
+    # Array access with simple expression in indices.
     child_idx += 1
-    with pytest.raises(NotImplementedError) as err:
-        _ = routine.children[child_idx].lhs._get_effective_shape()
-    assert "include a function call or expression" in str(err.value)
+    shape = routine.children[child_idx].lhs._get_effective_shape()
+    assert shape == []
+    # Array access with expression involving indirect access in indices.
+    child_idx += 1
```
hiker, line 698: Not all your fault, but I would find it much easier to understand if there would be a comment after each `child_idx+=1` to  indicate exactly what the statement in question is. E.g.:
```
    # Array access with expression involving indirect access in indices.
    # b(idx, 1+indices(1,1):) = 1
    child_idx += 1
```

hiker (changes requested):
Some minor comment issues, and the need to bring it up to master.

src/psyclone/psyad/transformations/preprocess.py
arporter: Oh yes, that's better. I did think about doing that but then decided not to for some reason. Anyway, it works nicely and is much clearer.

src/psyclone/tests/psyir/nodes/array_mixin_test.py
arporter: Good idea. (I think it was all my fault, just not in this PR :-) ). Done.

src/psyclone/psyad/transformations/preprocess.py
arporter: I've tested this with a couple of the LFRic kernels that first flagged the problem and it seems OK. @DrTVockerodtMO are you in a position now to test this branch with the full adjoint test app?

DrTVockerodtMO: I still do not know how to set that up yet, but if it produced the same KGO as before then it should be okay.

arporter:
Somehow, some warnings related to the Dev Guide had made it onto master and were causing the link check to fail. I've fixed those here so that the CI is green.

arporter:
This should be ready for another look now @hiker.

arporter:
The doc build is still failing but that's because the link to the OpenCL standard is returning a 403. If I visit the URL I get a check that I am a human so presumably this is what the problem is.

src/psyclone/tests/psyir/nodes/array_mixin_test.py
hiker: Great, thanks!

src/psyclone/psyad/transformations/preprocess.py
hiker: Much better, thanks.

hiker (approved):
All good. I've triggered the CI (because of changes to array_mixin), and will continue to merge if this is successful.

arporter:
Hi @hiker, not that the integration-test failure for NEMO5 is expected (@sergisiso says he's broken something).

hiker:
Assuming that the NEMO failures are known (as indicated by Andy), and the link check is also unrelated, I progress with merging.

PR #2769, opened 2024-11-07
(Closes #2768) Improve robustness of Sign2Code and Abs2Code transformations

### LINKED ISSUE #2768: Sign2CodeTrans generates invalid code for an array argument
Simon has reported that `Sign2CodeTrans` produces invalid code when applied to a SIGN that has an array as argument. One of the examples of this is from src/OCE/SBC/sbcblk_algo_andreas.F90 in NEMO from which I've constructed the following:
```
    program test_prog
      integer, parameter :: wp = kind(1.0d0)
      integer, parameter, dimension(0:4) :: A2D = (/1, 2, 3, 4, 5/)
      REAL(wp), DIMENSION(A2D(0)) :: ztmp1
      ztmp1 = 0.0
      ztmp1 = SIGN( MAX(ABS(ztmp1),1.E-6_wp), ztmp1 )
    end program test_prog'''
```

arporter:
Ensures that the target of the transformation has a scalar argument of known type (previously we were very naughty and assumed it was a REAL scalar). In order to allow us to seamlessly support arguments of integer type, I've extended the Fortran backend for Literals slightly so that it now ensures that reals end in ".0" and doubles in ".0d0". This revealed a bug in PSyAD where we were declaring array indices as REAL rather than INTEGER.


arporter:
This is almost ready now. I've set the integration tests going.

arporter:
NEMO4 OpenMP CPU failed with the usual race condition but was otherwise OK. NEMO5 is currently broken anyway. Apart from that, the integration tests passed.
This is therefore ready for review. It's not a biggie but could do with going on for the release.
Probably one for @hiker or @LonelyCat124.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1350,12 +1350,27 @@ def literal_node(self, node):
...
+            if precision == ScalarType.Precision.DOUBLE:
+                # The PSyIR stores real scalar values using the standard 'e'
+                # notation. If the scalar is in fact double precision then this
+                # 'e' must be replaced by 'd' for Fortran.
+                result = node.value.replace("e", "d", 1)
+                # Ensure it ends with ".0" if it isn't already explicitly
+                # formatted as a real.
+                try:
+                    _ = int(result)
+                    result = result + ".0d0"
```
LonelyCat124, line 1363: Is there a reason we force the specification of the real output in the backend? If its just for code appearance thats ok - but then I think this should also be updated in the other language writers as well (i think mainly `c.py` should also have this change). Alternatively we could move this logic into Literal itself

LonelyCat124 (changes requested):
I've had a look through the changes and they mostly look fine, it seems like a small fix- I need to run tests and linting locally to check them though still.

I've just got one discussion that needs fixing - once that's done I'll run through the rest of the things I need to do before merging.

src/psyclone/psyir/backend/fortran.py
arporter: That made me think but then I realised that this particular line is Fortran specific (the "0d0") so it does belong here rather than in the Node. I'll take a look at `c.py` as that will probably inform my opinion.

arporter: `c.py` had a redundant 'replace "d" with "e"' (redundant because the PSyIR always uses 'e' internally) so I've tidied it up.

arporter:
Thanks for the review @LonelyCat124. Note you don't need to worry about linting locally anymore as that's enforced as part of the CI. Similarly, as long as the integration tests are OK, you don't need to run tests locally either.

arporter:
Ready for another look now, assuming CI happiness.

LonelyCat124:
I've set the integration tests running and I'll check the changes with those results tomorrow.

LonelyCat124:
@arporter I'm not sure if the failures in the integration tests are normal?
Nemo 4 failed with:
```
cd /home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/EXP00; ./nemo
Warning: ieee_invalid is signaling
Warning: ieee_inexact is signaling
FORTRAN STOP
cd /home/gh_runner/NEMO/cfgs/SPITZ12_openmp_gpu/EXP00; cat timing.output; cat run.stat
 
       CNRS - NERC - Met OFFICE - MERCATOR-ocean - CMCC - INGV
                              NEMO team
                   Ocean General Circulation Model
                         version 4.0  (2019) 
 
                         Timing Informations 
 
 
 it :       1    |ssh|_max:  0.0000000000000000D+00 |U|_max:  0.0000000000000000D+00 S_min:  0.1797693134862316+309 S_max: -0.1797693134862316+309
Error: 'it :      10' not found!

arporter:
I broke the NEMOv4 integration tests yesterday :-( (While doing some housekeeping.) We're going to put the input files in a central location to fix this and prevent it happening again.

LonelyCat124:
Ok @arporter - I'm running the tests locally (still on the old laptop so it'll be a while) but I'm otherwise happy with this being ready to merge (so we can wait for integration tests fix or merge, up to you).

I can't actually build the user-guide pdf - could I ask you to do that for me and then I'll update the changelog and merge once the tests are complete?

LonelyCat124 (approved):

PR #2775, opened 2024-11-13
Added get_callee() which determines the right routine of the return values of get_callees()

schreiberx:
It does what's in the title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ cloning this repository and using:
 
     $ pip install .
 
+or in developer mode using
+
+    $ pip install -e .
+
 For more information about the installation process see
 [this section of the User Guide](https://psyclone.readthedocs.io/en/latest/system_specific_setup.html).
 
--- a/src/psyclone/psyir/nodes/call.py
+++ b/src/psyclone/psyir/nodes/call.py
@@ -47,8 +47,8 @@ from psyclone.psyir.nodes.datanode import DataNode
 from psyclone.psyir.nodes.reference import Reference
 from psyclone.psyir.nodes.routine import Routine
 from psyclone.psyir.symbols import (
-    RoutineSymbol, Symbol, SymbolError, UnsupportedFortranType)
-
+    RoutineSymbol, Symbol, SymbolError, UnsupportedFortranType, DataSymbol)
+from typing import List
 
 class Call(Statement, DataNode):
     ''' Node representing a Call. This can be found as a standalone statement
@@ -334,7 +334,7 @@ class Call(Statement, DataNode):
         return None
 
     @property
-    def arguments(self):
+    def arguments(self) -> DataNode:
         '''
         :returns: the children of this node that represent its arguments.
         :rtype: list[py:class:`psyclone.psyir.nodes.DataNode`]
@@ -418,6 +418,7 @@ class Call(Statement, DataNode):
     def __str__(self):
         return self.node_str(False)
 
+
     def copy(self):
         '''Return a copy of this node. This is a bespoke implementation for
         a Call node that ensures that any internal id's are
@@ -425,7 +426,6 @@ class Call(Statement, DataNode):
 
         :returns: a copy of this node and its children.
         :rtype: :py:class:`psyclone.psyir.node.Node`
-
         '''
         # ensure _argument_names is consistent with actual arguments
         # before copying.
@@ -442,16 +442,17 @@ class Call(Statement, DataNode):
 
         return new_copy
 
+
     def get_callees(self):
         '''
-        Searches for the implementation(s) of the target routine for this Call.
+        Searches for the implementation(s) of all potential target routines
+        for this Call without any arguments check.
 
         :returns: the Routine(s) that this call targets.
         :rtype: list[:py:class:`psyclone.psyir.nodes.Routine`]
 
         :raises NotImplementedError: if the routine is not local and not found
             in any containers in scope at the call site.
-
         '''
         def _location_txt(node):
             '''
@@ -593,3 +594,148 @@ class Call(Statement, DataNode):
             f"Failed to find a Routine named '{rsym.name}' in "
             f"{_location_txt(root_node)}. This is normally because the routine"
             f" is within a CodeBlock.")
+
+    class MatchingArgumentsNotFound(BaseException): ...
+
+    def get_argument_routine_match(self, routine: Routine):
+        """Return a list of integers giving for each argument of the call
+        the index of the argument in argument_list (typically of a routine)
+
+        :param argument_list: List of arguments
+        :type argument_list: Arguments
+        :return: None if no match was found, otherwise list of integers
+        referring to matching arguments.
+        :rtype: None|List[int]
+        """
+
+        # Create a copy of the list
+        # Once an argument has been successfully matched, set it to 'None'
+        routine_argument_list: List[DataNode] = routine.symbol_table.argument_list[:]
+
+        # Find matching argument list
+        #if len(self.arguments) != len(routine_argument_list):
+        #    return None
+
+        if len(self.arguments) > len(routine.symbol_table.argument_list):
+            raise self.MatchingArgumentsNotFound(
+                f"More arguments in callee  (call '{self.routine.name}') than caller (routine '{routine.name}')"
+            )
+
+        assert len(self.arguments) == len(self.argument_names)
+
+        # Iterate over all arguments
+        ret_arg_idx_list = []
+        for call_arg_idx, call_arg in enumerate(self.arguments):
+            call_arg_idx: int
+            call_arg: DataSymbol
+
+            # If None, it's a positional argument => Just return the index if the types match
+            if self.argument_names[call_arg_idx] is None:
+                routine_arg = routine_argument_list[call_arg_idx]
+                routine_arg: DataSymbol
+
+                # Do the types of arguments match?
+                #
+                # TODO #759: If optional is used, it's an unsupported Fortran type and we need to use the following workaround
+                # Once this issue is resolved, simply remove this if branch
+                if not isinstance(routine_arg.datatype, UnsupportedFortranType):
+                    if call_arg.datatype != routine_arg.datatype:
+                        raise self.MatchingArgumentsNotFound(
+                            f"Argument type mismatch of call argument '{call_arg}' and routine argument '{routine_arg}'"
+                        )
+                
+                ret_arg_idx_list.append(call_arg_idx)
+                routine_argument_list[call_arg_idx] = None
+                continue
+
+            #
+            # Next, we handle all named arguments
+            #
+            arg_name = self.argument_names[call_arg_idx]
+            named_arg_found = False
+            for routine_arg_idx, routine_arg in enumerate(routine_argument_list):
+                routine_arg: DataSymbol
+
+                # Check if argument was already processed
+                if routine_arg is None:
+                    continue
+
+                if arg_name == routine_arg.name:
+                    # TODO #759: If optional is used, it's an unsupported Fortran type and we need to use the following workaround
+                    # Once this issue is resolved, simply remove this if branch
+                    if not isinstance(routine_arg.datatype, UnsupportedFortranType):
+                        if call_arg.datatype != routine_arg.datatype:
+                            raise self.MatchingArgumentsNotFound(
+                                f"Argument type mismatch of call argument '{call_arg}' and routine argument '{routine_arg}'"
+                            )
+                    
+                    ret_arg_idx_list.append(routine_arg_idx)
+                    named_arg_found = True
+                    break
+
+            if not named_arg_found:
+                # It doesn't match => Raise exception
+                raise self.MatchingArgumentsNotFound
+            
+            routine_argument_list[routine_arg_idx] = None
+
+
+
+        #
+        # Finally, we check if all left-over arguments are optional arguments
+        #
+        for routine_arg in routine_argument_list:
+            routine_arg: DataNode
+
+            if routine_arg is None:
+                continue
+
+            # TODO #759: Optional keyword is not yet supported in psyir.
+            # Hence, we use a simple string match.
+            if ", OPTIONAL" in str(routine_arg.datatype):
+                continue
+
+            raise self.MatchingArgumentsNotFound(
+                f"Argument '{routine_arg}' in subroutine '{routine.name}' not handled"
+            )
+
+        return ret_arg_idx_list
+
+
+    def get_callee(self, ret_arg_match_list:List[int] = None, check_matching_arguments: bool=True):
+        '''
+        Searches for the implementation(s) of the target routine for this Call
+        including argument checks.
+
+        :param ret_arg_match_list: List in which the matching argument indices will be returned
+
+        :returns: the Routine(s) that this call targets.
+        :rtype: list[:py:class:`psyclone.psyir.nodes.Routine`]
+
+        :raises NotImplementedError: if the routine is not local and not found
+            in any containers in scope at the call site.
+        '''
+
+        routine_list = self.get_callees()
+
+        # Search for the routine matching the right arguments
+        for routine in routine_list:
+            routine: Routine
+
+            try:
+                arg_match_list = self.get_argument_routine_match(routine)
+            except self.MatchingArgumentsNotFound as e:
+                continue
+
+            if ret_arg_match_list is not None:
+                ret_arg_match_list[:] = arg_match_list
+            
+            return routine
+        
+        # If we didn't find any routine, return some routine if no matching arguments have been found.
+        # This is handy for the transition phase until optional argument matching is supported.
+        if not check_matching_arguments:
+            return routine_list[0]
+        
+
+        raise NotImplementedError(f"No matching routine for call '{self.routine.name}' found")
--- a/src/psyclone/psyir/nodes/call.py
+++ b/src/psyclone/psyir/nodes/call.py
@@ -595,7 +595,10 @@ class Call(Statement, DataNode):
             f"{_location_txt(root_node)}. This is normally because the routine"
             f" is within a CodeBlock.")
 
-    class MatchingArgumentsNotFound(BaseException): ...
+    class MatchingArgumentsNotFound(BaseException):
+        """Excepction to signal that matching arguments have not been found 
+        for this routine
+        """
 
     def get_argument_routine_match(self, routine: Routine):
         """Return a list of integers giving for each argument of the call
@@ -653,6 +656,7 @@ class Call(Statement, DataNode):
             #
             arg_name = self.argument_names[call_arg_idx]
             named_arg_found = False
+            routine_arg_idx = None
             for routine_arg_idx, routine_arg in enumerate(routine_argument_list):
                 routine_arg: DataSymbol
 
@@ -676,6 +680,9 @@ class Call(Statement, DataNode):
             if not named_arg_found:
                 # It doesn't match => Raise exception
                 raise self.MatchingArgumentsNotFound
+            
+            if routine_arg_idx is None:
+                raise IndexError("Internal error")
             
             routine_argument_list[routine_arg_idx] = None
 
@@ -724,7 +731,7 @@ class Call(Statement, DataNode):
 
             try:
                 arg_match_list = self.get_argument_routine_match(routine)
-            except self.MatchingArgumentsNotFound as e:
+            except self.MatchingArgumentsNotFound:
                 continue
 
             if ret_arg_match_list is not None:
--- a/src/psyclone/tests/psyir/nodes/call_test.py
+++ b/src/psyclone/tests/psyir/nodes/call_test.py
@@ -255,7 +255,7 @@ def test_call_add_args_error2():
         _ = Call._add_args(routine, [(1, 2)])
     assert ("If a child of the children argument in create method of Call "
             "class is a tuple, its first argument should be a str, but "
-            " - Found int." in str(info.value))
+            "found int." in str(info.value))
 
 
 def test_call_appendnamedarg():
--- a/src/psyclone/tests/psyir/nodes/call_test.py
+++ b/src/psyclone/tests/psyir/nodes/call_test.py
@@ -43,7 +43,7 @@ from psyclone.core import Signature, VariablesAccessInfo
 from psyclone.parse import ModuleManager
 from psyclone.psyir.nodes import (
     ArrayReference, Assignment, BinaryOperation, Call, CodeBlock, Literal,
-    Reference, Routine, Schedule)
+    Reference, Routine, Schedule, Node)
 from psyclone.psyir.nodes.node import colored
 from psyclone.psyir.symbols import (
     ArrayType, INTEGER_TYPE, DataSymbol, NoType, RoutineSymbol, REAL_TYPE,
@@ -255,7 +255,7 @@ def test_call_add_args_error2():
         _ = Call._add_args(routine, [(1, 2)])
     assert ("If a child of the children argument in create method of Call "
             "class is a tuple, its first argument should be a str, but "
-            "found int." in str(info.value))
+            " - Found int." in str(info.value))
 
 
 def test_call_appendnamedarg():
@@ -626,6 +626,447 @@ end module some_mod'''
     call = psyir.walk(Call)[0]
     result = call.get_callees()
     assert result == [psyir.walk(Routine)[1]]
+
+
+
+def test_call_get_callee_1_simple_match(fortran_reader):
+    '''
+    Check that right routine has been found.
+    '''
+    code = '''
+module some_mod
+  implicit none
+contains
+
+  subroutine main()
+    integer :: e, f, g
+    call foo(e, f, g)
+  end subroutine
+
+  ! Matching routine
+  subroutine foo(a, b, c)
+    integer :: a, b, c
+  end subroutine
+
+end module some_mod'''
+
+    psyir = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = psyir.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+
+    result = call_foo.get_callee()
+    
+    routine_match: Routine = psyir.walk(Routine)[1]
+    assert result is routine_match
+
+
+def test_call_get_callee_2_optional_args(fortran_reader):
+    '''
+    Check that optional arguments have been correlated correctly.
+    '''
+    code = '''
+module some_mod
+  implicit none
+contains
+
+  subroutine main()
+    integer :: e, f, g
+    call foo(e, f)
+  end subroutine
+
+  ! Matching routine
+  subroutine foo(a, b, c)
+    integer :: a, b
+    integer, optional :: c
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    routine_match: Routine = root_node.walk(Routine)[1]
+    assert routine_match.name == "foo"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+    assert call_foo.routine.name == "foo"
+
+    arg_idx_list = []
+    result: Routine = call_foo.get_callee(ret_arg_match_list=arg_idx_list)
+
+    assert len(arg_idx_list) == 2
+    assert arg_idx_list[0] == 0
+    assert arg_idx_list[1] == 1
+    
+    assert result is routine_match
+
+
+def test_call_get_callee_3_trigger_error(fortran_reader):
+    '''
+    Test which is supposed to trigger an error.
+    '''
+    code = '''
+module some_mod
+  implicit none
+contains
+
+  subroutine main()
+    integer :: e, f, g
+    call foo(e, f, g)
+  end subroutine
+
+  ! Matching routine
+  subroutine foo(a, b)
+    integer :: a, b
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    routine_match: Routine = root_node.walk(Routine)[1]
+    assert routine_match.name == "foo"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+    assert call_foo.routine.name == "foo"
+
+    arg_idx_list = []
+
+    try:
+        result: Routine = call_foo.get_callee(ret_arg_match_list=arg_idx_list)
+    except:
+        print("Success! Exception triggered (as expected)")
+        return
+
+    assert False, "This should have triggered an error since there are more arguments in the call than in the routine"
+
+
+def test_call_get_callee_4_named_arguments(fortran_reader):
+    '''
+    Check that named arguments have been correlated correctly
+    '''
+    code = '''
+module some_mod
+  implicit none
+contains
+
+  subroutine main()
+    integer :: e, f, g
+    call foo(c=e, a=f, b=g)
+  end subroutine
+
+  ! Matching routine
+  subroutine foo(a, b, c)
+    integer :: a, b, c
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    routine_match: Routine = root_node.walk(Routine)[1]
+    assert routine_match.name == "foo"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+    assert call_foo.routine.name == "foo"
+
+    arg_idx_list = []
+    result: Routine = call_foo.get_callee(ret_arg_match_list=arg_idx_list)
+
+    assert len(arg_idx_list) == 3
+    assert arg_idx_list[0] == 2
+    assert arg_idx_list[1] == 0
+    assert arg_idx_list[2] == 1
+    
+    assert result is routine_match
+
+
+def test_call_get_callee_5_optional_and_named_arguments(fortran_reader):
+    '''
+    Check that optional and named arguments have been correlated correctly
+    '''
+    code = '''
+module some_mod
+  implicit none
+contains
+
+  subroutine main()
+    integer :: e, f, g
+    call foo(b=e, a=f)
+  end subroutine
+
+  ! Matching routine
+  subroutine foo(a, b, c)
+    integer :: a, b
+    integer, optional :: c
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    routine_match: Routine = root_node.walk(Routine)[1]
+    assert routine_match.name == "foo"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+    assert call_foo.routine.name == "foo"
+
+    arg_idx_list = []
+    result: Routine = call_foo.get_callee(ret_arg_match_list=arg_idx_list)
+
+    assert len(arg_idx_list) == 2
+    assert arg_idx_list[0] == 1
+    assert arg_idx_list[1] == 0
+    
+    assert result is routine_match
+
+
+def test_call_get_callee_6_interfaces(fortran_reader):
+    '''
+    Check that optional and named arguments have been correlated correctly
+    '''
+    code = '''
+module some_mod
+  implicit none
+
+  interface foo
+    procedure foo_a, foo_b, foo_c
+  end interface
+contains
+
+  subroutine main()
+    integer :: e_int, f_int, g_int
+    real :: e_real, f_real, g_real
+
+    ! Should match foo_a
+    call foo(e_int, f_int)
+
+    ! Should match foo_a
+    call foo(e_int, f_int, g_int)
+
+    
+    ! Should match foo_b
+    call foo(e_real, f_int)
+
+    ! Should match foo_b
+    call foo(e_real, f_int, g_int)
+
+    ! Should match foo_b
+    call foo(e_real, c=f_int, b=g_int)
+
+    
+    ! Should match foo_c
+    call foo(e_int, f_real, g_int)
+
+    ! Should match foo_c
+    call foo(b=e_real, a=f_int)
+
+    ! Should match foo_c
+    call foo(b=e_real, a=f_int, g_int)
+  end subroutine
+
+  subroutine foo_a(a, b, c)
+    integer :: a, b
+    integer, optional :: c
+  end subroutine
+
+  subroutine foo_b(a, b, c)
+    real :: a
+    integer :: b
+    integer, optional :: c
+  end subroutine
+
+  subroutine foo_c(a, b, c)
+    integer :: a
+    real :: b
+    integer, optional :: c
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    if 1:
+        routine_foo_a: Routine = root_node.walk(Routine)[1]
+        assert routine_foo_a.name == "foo_a"
+
+        if 1:
+            print("Subtest foo_a[0]:")
+
+            call_foo_a: Call = routine_main.walk(Call)[0]
+            assert call_foo_a.routine.name == "foo"
+
+            arg_idx_list = []
+            result: Routine = call_foo_a.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 2
+            assert arg_idx_list[0] == 0
+            assert arg_idx_list[1] == 1
+
+            assert result is routine_foo_a
+            print(" - Passed subtest foo_a[0]")
+
+        if 1:
+            print("Subtest foo_a[1]:")
+
+            call_foo_a: Call = routine_main.walk(Call)[1]
+            assert call_foo_a.routine.name == "foo"
+
+            arg_idx_list = []
+            result: Routine = call_foo_a.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 3
+            assert arg_idx_list[0] == 0
+            assert arg_idx_list[1] == 1
+            assert arg_idx_list[2] == 2
+
+            assert result is routine_foo_a
+            print(" - Passed subtest foo_a[1]")
+    
+    if 1:
+        routine_foo_b: Routine = root_node.walk(Routine)[2]
+        assert routine_foo_b.name == "foo_b"
+
+        if 1:
+            print("Subtest foo_b[0]:")
+
+            call_foo_b: Call = routine_main.walk(Call)[2]
+            assert call_foo_b.routine.name == "foo"
+          
+            arg_idx_list = []
+            result: Routine = call_foo_b.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 2
+            assert arg_idx_list[0] == 0
+            assert arg_idx_list[1] == 1
+
+            assert result is routine_foo_b
+            print(" - Passed subtest foo_b[0]")
+
+        if 1:
+            print("Subtest foo_b[1]:")
+
+            call_foo_b: Call = routine_main.walk(Call)[3]
+            assert call_foo_b.routine.name == "foo"
+          
+            arg_idx_list = []
+            result: Routine = call_foo_b.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 3
+            assert arg_idx_list[0] == 0
+            assert arg_idx_list[1] == 1
+            assert arg_idx_list[2] == 2
+
+            assert result is routine_foo_b
+            print(" - Passed subtest foo_b[1]")
+
+        if 1:
+            print("Subtest foo_b[2]:")
+
+            call_foo_b: Call = routine_main.walk(Call)[4]
+            assert call_foo_b.routine.name == "foo"
+          
+            arg_idx_list = []
+            result: Routine = call_foo_b.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 3
+            assert arg_idx_list[0] == 0
+            assert arg_idx_list[1] == 2
+            assert arg_idx_list[2] == 1
+
+            assert result is routine_foo_b
+            print(" - Passed subtest foo_b[2]")
+
+
+
+    if 1:
+        routine_foo_c: Routine = root_node.walk(Routine)[3]
+        assert routine_foo_c.name == "foo_c"
+
+        if 1:
+            print("Subtest foo_c[0]:")
+
+            call_foo_c: Call = routine_main.walk(Call)[5]
+            assert call_foo_c.routine.name == "foo"
+          
+            arg_idx_list = []
+            result: Routine = call_foo_c.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 3
+            assert arg_idx_list[0] == 0
+            assert arg_idx_list[1] == 1
+            assert arg_idx_list[2] == 2
+
+            assert result is routine_foo_c
+            print(" - Passed subtest foo_c[0]")
+
+        if 1:
+            print("Subtest foo_c[1]:")
+
+            call_foo_c: Call = routine_main.walk(Call)[6]
+            assert call_foo_c.routine.name == "foo"
+          
+            arg_idx_list = []
+            result: Routine = call_foo_c.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 2
+            assert arg_idx_list[0] == 1
+            assert arg_idx_list[1] == 0
+
+            assert result is routine_foo_c
+            print(" - Passed subtest foo_c[1]")
+
+        if 1:
+            print("Subtest foo_c[2]:")
+
+            call_foo_c: Call = routine_main.walk(Call)[7]
+            assert call_foo_c.routine.name == "foo"
+          
+            arg_idx_list = []
+            result: Routine = call_foo_c.get_callee(ret_arg_match_list=arg_idx_list)
+
+            print(f" - Found matching argument list: {arg_idx_list}")
+
+            assert len(arg_idx_list) == 3
+            assert arg_idx_list[0] == 1
+            assert arg_idx_list[1] == 0
+            assert arg_idx_list[2] == 2
+
+            assert result is routine_foo_c
+            print(" - Passed subtest foo_c[2]")
+
+
 
 
 @pytest.mark.usefixtures("clear_module_manager_instance")


arporter:
Thanks @schreiberx, unfortunately, linting is still failing.

In case it helps, I have the following pre-push hook setup:
```
$ cat PSyclone/.git/hooks/pre-push
#!/usr/bin/sh

# An example hook script to verify what is about to be pushed.  Called by "git
# push" after it has checked the remote status, but before anything has been
# pushed.  If this script exits with a non-zero status nothing will be pushed.
#
# This hook is called with the following parameters:
#
# $1 -- Name of the remote to which the push is being done
# $2 -- URL to which the push is being done
#
# If pushing without using a named remote those arguments will be equal.
#
# Information about the commits which are being pushed is supplied as lines to
# the standard input in the form:
#
#   <local ref> <local oid> <remote ref> <remote oid>
#
# This script ensures that the whole of PSyclone is linted successfully
# before the push is executed.

remote="$1"
url="$2"

if ! command -v flake8; then
    echo "WARNING: source not linted because flake8 unavailable"
    exit 0
fi

flake8=$(flake8 src/psyclone)
if (( flake8 == 1 )); then
    echo "Linting failed"
else
    echo "Linting succeeded"
fi
exit $flake8
```

schreiberx:
Hi Andy, Thanks for this script.
I think that all these helper scripts should be part of psyclone to quickly run them over the source code. I added this script in the newest pushed version in `bin_git/run_flake8.sh`.
Cheers,
Martin

arporter:
@schreiberx please could you hold off making further changes until I pass it back to you? It's hard to review if code changes underneath me while I'm doing it :-)

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+                                UnsupportedFortranType):
+                        if call_arg.datatype != routine_arg.datatype:
+                            raise CallMatchingArgumentsNotFound(
+                                f"Argument type mismatch of call argument "
+                                f"'{call_arg}' and routine argument "
+                                f"'{routine_arg}'"
+                            )
+
+                    ret_arg_idx_list.append(routine_arg_idx)
+                    named_arg_found = True
```
arporter, line 688: This pattern can be achieved without the bool:
```python
    for something in my_search:
        if found_it:
            break
    else:
        # Didn't find it.
```

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+        #
+        for routine_arg in routine_argument_list:
+            routine_arg: DataNode
+
+            if routine_arg is None:
+                continue
+
+            # TODO #759: Optional keyword is not yet supported in psyir.
+            # Hence, we use a simple string match.
+            if ", OPTIONAL" in str(routine_arg.datatype):
```
arporter, line 710: I think we can do a bit better here with not much additional work. The PSyIR type system has the concept of a `partial_datatype` which we can use to capture all the parts of an UnsupportedFortranType that the PSyIR *does* support. It should be a small change to the frontend (`fparser2.py`) to ensure this is populated for OPTIONAL arguments. I can take a look at this if you like?

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
             f"Failed to find a Routine named '{rsym.name}' in "
             f"{_location_txt(root_node)}. This is normally because the routine"
             f" is within a CodeBlock.")
+
+    def get_argument_routine_match(self, routine: Routine):
+        """Return a list of integers giving for each argument of the call
```
arporter, line 605: The passing back of a list of integers seems a bit clunky. Could you explain why you need this? Could we pass back the Symbols themselves perhaps?

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+
+    def get_callee(
+                self,
+                ret_arg_match_list: List[int] = None,
+                check_matching_arguments: bool = True):
+        '''
+        Searches for the implementation(s) of the target routine for this Call
+        including argument checks.
+
+        :param ret_arg_match_list: List in which the matching argument
```
arporter, line 728: As I comment above, I'd like to understand the need for this. If we decide to keep it, you need a `:type ret_arg_match_list:` in the docstring too.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+            indices will be returned
+
+        :returns: the Routine(s) that this call targets.
```
arporter, line 731: Just one Routine for this variant.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+    def get_callee(
+                self,
+                ret_arg_match_list: List[int] = None,
+                check_matching_arguments: bool = True):
+        '''
+        Searches for the implementation(s) of the target routine for this Call
+        including argument checks.
+
+        :param ret_arg_match_list: List in which the matching argument
+            indices will be returned
```
arporter, line 729: Please document the `check_matching_arguments` argument.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+
+        # If we didn't find any routine, return some routine if no matching
+        # arguments have been found.
+        # This is handy for the transition phase until optional argument
+        # matching is supported.
+        if not check_matching_arguments:
+            return routine_list[0]
+
+        if error is not None:
+            raise error
```
arporter, line 767: Perhaps `raise NotImplementedError(blah) from error` here might be nicer?

arporter (changes requested):
Thanks very much for this Martin. I like the new routine making use of the existing functionality, that seems sensible. I did struggle to get my head around it so you'll see I've requested that some comments be tweaked or added.
As I say in the review, I think a little work on the frontend for those situations where we hit OPTIONAL could really help here.
The only thing I'm really not sure of is the passing around of the list of integers. That seems a bit odd so I'd like to understand why you do that.
I've not looked at the test suite yet (but thanks for addressing the missed lines).

arporter:
Back to @schreiberx :-)

README.md
```diff
@@ -50,6 +50,10 @@ cloning this repository and using:
 
     $ pip install .
 
+or in developer mode using
```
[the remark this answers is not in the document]
schreiberx: I never worked on such comments on github and hope I'm doing things right:
I'll do this in the branch and push the updates.
Agreed.

bin_git/run_flake8.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
schreiberx: For me, that's not directly git related - although you used it as git hooks.
It's about good coding styles and scripts that help one ensure this.
In addition, this can also be used by git and the CI system.
Therefore, I'd just put this directly into the ./utils folder. I also added a `run_pytest_cov.sh` script.

bin_git/run_flake8.sh
```diff
@@ -0,0 +1,43 @@
...
+# $2 -- URL to which the push is being done
+#
+# If pushing without using a named remote those arguments will be equal.
+#
+# Information about the commits which are being pushed is supplied as lines to
+# the standard input in the form:
+#
+#   <local ref> <local oid> <remote ref> <remote oid>
+#
+# This script ensures that the whole of PSyclone is linted successfully
```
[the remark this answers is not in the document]
schreiberx: I moved up the sentence "This ..." and rewrote this slightly.

src/psyclone/psyir/nodes/call.py
```diff
@@ -334,7 +335,7 @@ def routine(self):
         return None
 
     @property
-    def arguments(self):
+    def arguments(self) -> DataNode:
```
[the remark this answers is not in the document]
schreiberx: Yes!

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +593,173 @@ def _location_txt(node):
...
+        """Excepction to signal that matching arguments have not been found
+        for this routine
+        """
+
+    def get_argument_routine_match(self, routine: Routine):
+        """Return a list of integers giving for each argument of the call
+        the index of the argument in argument_list (typically of a routine)
+
+        :param argument_list: List of arguments
+        :type argument_list: Arguments
```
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
```diff
@@ -47,7 +47,14 @@
 from psyclone.psyir.nodes.reference import Reference
 from psyclone.psyir.nodes.routine import Routine
 from psyclone.psyir.symbols import (
-    RoutineSymbol, Symbol, SymbolError, UnsupportedFortranType)
+    RoutineSymbol, Symbol, SymbolError, UnsupportedFortranType, DataSymbol)
+from typing import List
+
+
+class CallMatchingArgumentsNotFound(BaseException):
+    """Excepction to signal that matching arguments have not been found
```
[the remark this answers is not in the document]
schreiberx: I'll install a spellchecker :-)

src/psyclone/psyir/nodes/call.py
```diff
@@ -425,7 +433,6 @@ def copy(self):
 
         :returns: a copy of this node and its children.
         :rtype: :py:class:`psyclone.psyir.node.Node`
```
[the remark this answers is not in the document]
schreiberx: OK. If there's a flake8 option, we should add it also to the script in `utils`

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
             f" is within a CodeBlock.")
+
+    def get_argument_routine_match(self, routine: Routine):
+        """Return a list of integers giving for each argument of the call
+        the index of the argument in argument_list (typically of a routine)
+
+        :param argument_list: List of arguments
+        :type argument_list: Arguments
+        :return: None if no match was found, otherwise list of integers
+        referring to matching arguments.
```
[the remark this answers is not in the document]
schreiberx: OK. One Python "tab" it is.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+        """
+
+        # Create a copy of the list
+        # Once an argument has been successfully matched, set it to 'None'
+        routine_argument_list: List[DataNode] = \
+            routine.symbol_table.argument_list[:]
+
+        if len(self.arguments) > len(routine.symbol_table.argument_list):
+            raise CallMatchingArgumentsNotFound(
+                f"More arguments in callee  (call '{self.routine.name}')"
```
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+        the index of the argument in argument_list (typically of a routine)
+
+        :param argument_list: List of arguments
+        :type argument_list: Arguments
+        :return: None if no match was found, otherwise list of integers
+        referring to matching arguments.
+        :rtype: None|List[int]
+        """
+
+        # Create a copy of the list
```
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+        routine_argument_list: List[DataNode] = \
+            routine.symbol_table.argument_list[:]
+
+        if len(self.arguments) > len(routine.symbol_table.argument_list):
+            raise CallMatchingArgumentsNotFound(
+                f"More arguments in callee  (call '{self.routine.name}')"
+                f" than caller (routine '{routine.name}')"
+            )
+
+        assert len(self.arguments) == len(self.argument_names)
```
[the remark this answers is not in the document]
schreiberx: I don't think so. I just added this to catch some potential internal psyir error.
If you ask me to do it like that, I have to introduce errors in psyir which I don't want to do :-).

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+
+        # Iterate over all arguments
```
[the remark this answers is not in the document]
schreiberx: OK

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+
+        # Create a copy of the list
+        # Once an argument has been successfully matched, set it to 'None'
+        routine_argument_list: List[DataNode] = \
+            routine.symbol_table.argument_list[:]
+
+        if len(self.arguments) > len(routine.symbol_table.argument_list):
+            raise CallMatchingArgumentsNotFound(
+                f"More arguments in callee  (call '{self.routine.name}')"
+                f" than caller (routine '{routine.name}')"
```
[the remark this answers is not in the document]
schreiberx: OK. Thanks for spotting this.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+            if self.argument_names[call_arg_idx] is None:
+                routine_arg = routine_argument_list[call_arg_idx]
+                routine_arg: DataSymbol
+
+                # Do the types of arguments match?
+                #
+                # TODO #759: If optional is used, it's an unsupported Fortran
+                # type and we need to use the following workaround
+                # Once this issue is resolved, simply remove this if branch
+                if not isinstance(
```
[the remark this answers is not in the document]
schreiberx: OK.

.gitignore
```diff
@@ -7,6 +7,7 @@ gnu_opt_report.txt
 *.MOD
 .cache
 .coverage
+**/*.swp
```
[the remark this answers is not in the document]
schreiberx: OK.

README.md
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+                routine_arg: DataSymbol
+
+                # Check if argument was already processed
+                if routine_arg is None:
+                    continue
+
+                if arg_name == routine_arg.name:
+                    # TODO #759: If optional is used, it's an unsupported
+                    # Fortran type and we need to use the following workaround
+                    # Once this issue is resolved, simply remove this if branch
```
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: Nice solution!

src/psyclone/psyir/nodes/call.py
schreiberx: Sure, if you would like to do this, I'd make the other changes again in call.py.

src/psyclone/psyir/nodes/call.py
schreiberx: There's an excellent purpose of returning the numbers:
This can be used by the inlining functionality.
Finding the right routine also means that you need to check whether the arguments match.
Once you know this, for inlining, you need to know which call argument is related to which subroutine argument.
This is given by these numbers.
Alternatively, a list "ret_arg_idx:List[int]=None" could be handed over as an argument to this function which is only updated if it's provided:

```
if ret_arg_idx is not None:
    ret_arg_idx[:] = ... # The list with the indices
```

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: I changed it.

schreiberx:
Back to @arporter :-)

README.md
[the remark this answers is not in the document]
arporter: Thanks Martin. You're doing the right thing. Our working practise is that it's up to the reviewer to 'resolve' conversations once they're sure that things have been done - see https://github.com/stfc/PSyclone/wiki/CodeReview.

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
arporter: I don't understand what you mean by "I have to introduce errors in psyir which I don't want to do"?

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +600,172 @@ def _location_txt(node):
...
+
+        assert len(self.arguments) == len(self.argument_names)
+
+        # Iterate over all arguments
+        ret_arg_idx_list = []
+        for call_arg_idx, call_arg in enumerate(self.arguments):
+            call_arg_idx: int
+            call_arg: DataSymbol
+
+            # If None, it's a positional argument => Just return the index if
```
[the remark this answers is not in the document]
arporter: This one is still outstanding.

bin_git/run_flake8.sh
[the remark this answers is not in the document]
schreiberx: I'll add information for run_flake8.sh and run_pytest_cov.sh to the documentation `working_practises`

src/psyclone/psyir/nodes/call.py
```diff
@@ -51,6 +51,12 @@
 from typing import List
 
 
+class CallMatchingArgumentsNotFound(BaseException):
```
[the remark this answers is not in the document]
schreiberx: OK

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
schreiberx: OK. An empty line it is. Let's see if I understood this right.

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
schreiberx: To make the if-branch to be executed, I would have needed to get Psyclone into an invalid internal representation where exactly this assertion would be triggered.

By changing it to this simple assertion, the coverage test doesn't complain anymore. I hope this resolves this. We can also remove this line :-)

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
schreiberx: OK

schreiberx (commented):
Back to @arporter :-)

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
arporter: If removing this line is a valid thing to do then I'm happy. I agree that getting coverage for error conditions is a pain but it's a vital part of our development process and, as I say, it makes future extension of the code much easier and more robust.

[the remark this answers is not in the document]
schreiberx: Then I'll remove it.
I already saw and agree with the value of coverage test cases.
It might take hours, but making it robust is worth it.  :-)

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +604,176 @@ def _location_txt(node):
...
+
+        return ret_arg_idx_list
+
+    def get_callee(
+                self,
+                ret_arg_match_list: List[int] = None,
+                check_matching_arguments: bool = True):
+        """
+        Searches for the implementation(s) of the target routine for this Call
+        including argument checks.
```
arporter, line 730: Please also explain that `check_matching_arguments` may be set to False and in this case the very first implementation of the matching routine will be returned. Please emphasise that the arguments this routine expects may then not be what is provided by this call.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -43,11 +43,12 @@
 from psyclone.parse import ModuleManager
 from psyclone.psyir.nodes import (
     ArrayReference, Assignment, BinaryOperation, Call, CodeBlock, Literal,
-    Reference, Routine, Schedule)
+    Reference, Routine, Schedule, Node)
 from psyclone.psyir.nodes.node import colored
 from psyclone.psyir.symbols import (
     ArrayType, INTEGER_TYPE, DataSymbol, NoType, RoutineSymbol, REAL_TYPE,
     SymbolError, UnsupportedFortranType)
```
arporter, line 50: We aspire to have all imports in alphabetical order, at least by module name. Therefore, please could you move this one up to L47.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
     assert result == [psyir.walk(Routine)[1]]
 
 
+def test_call_get_callee_1_simple_match(fortran_reader):
+    '''
+    Check that right routine has been found.
```
arporter, line 634: "...found for a single routine implementation."

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+    assert routine_main.name == "main"
+
+    routine_match: Routine = root_node.walk(Routine)[1]
+    assert routine_match.name == "foo"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+    assert call_foo.routine.name == "foo"
+
+    arg_idx_list = []
+    result: Routine = call_foo.get_callee(ret_arg_match_list=arg_idx_list)
```
arporter, line 700: As I say in the comments earlier, it's more usual for us to return a tuple in situations where we want more than one quantity back from a routine. So this would be `result, arg_list = call_foo.get_callee(...)`

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+
+    routine_match: Routine = root_node.walk(Routine)[1]
+    assert routine_match.name == "foo"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+    assert call_foo.routine.name == "foo"
+
+    arg_idx_list = []
+
+    try:
```
arporter, line 743: You want `with pytest.raises(CallMatchingArgumentsNotFound) as err:` here.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+        call_foo.get_callee(ret_arg_match_list=arg_idx_list)
+
+    except CallMatchingArgumentsNotFound as err:
+        assert "No matching routine found for" in str(err)
+        print("Success! Exception triggered (as expected)")
```
arporter, line 748: No need for `print` or `return`. pytest will fail if the exception isn't raised and you just need to check that the error contains the expected message.

src/psyclone/psyir/nodes/call.py
arporter: As always, looking at the tests helped me think this through a bit more and I realise that my list idea won't work. I think we do want a `dict` but the keys should be the routine's formal arguments (either their names or the Symbols) and the values should be the actual arguments provided to the Call. I think this then mirrors what happens when calling a Python function which is encouraging.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+    assert len(arg_idx_list) == 2
+    assert arg_idx_list[0] == 1
+    assert arg_idx_list[1] == 0
+
+    assert result is routine_match
+
+
+def test_call_get_callee_6_interfaces(fortran_reader):
+    '''
+    Check that optional and named arguments have been correlated correctly
```
arporter, line 843: "...correctly when the call is to a generic interface."

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    if 1:
```
arporter, line 909: Please remove all these `if`s and `print` statements (but do add comments describing what specific case each part of the test is for).

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+end module some_mod"""
+
+    psyir = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = psyir.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+
+    try:
```
arporter, line 1108: Again, you want `with pytest.raises(...) as err` here...

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+    Trigger error that matching arguments were not found
+    """
+    code = """
+module some_mod
+  implicit none
+contains
+
+  subroutine main()
+    integer :: e, f
+    ! Use name 'd' which doesn't exist
```
arporter, line 1133: I don't understand this comment. Perhaps 'omit the required third argument'?

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,539 @@ def test_call_get_callees_local(fortran_reader):
...
+end module some_mod"""
+
+    psyir = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = psyir.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+
+    try:
```
arporter, line 1151: `with pytest.raises(...)` please.

arporter (changes requested):
Thanks for the updates Martin. I think the main thing to agree upon now is how best to return the information on the argument mapping. Apart from that it's a matter of tidying a few things. You also still have a couple of lines missing from coverage (but best to leave that until we've hammered-out the implementation).
I have the one-line fix to improve our OPTIONAL support as #2785 which is just waiting on review. Since it's so simple, you could merge that branch into this one.

doc/developer_guide/system_specific_setup.rst
```diff
@@ -64,11 +64,11 @@ Installing Documentation Tools
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Install Sphinx along with bibtex support for creating PSyclone documentation::
 
-   > sudo pip install sphinx sphinxcontrib.bibtex
+   > pip install sphinx sphinxcontrib.bibtex sphinx-rtd-theme
```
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK. I assume you're talking about "Node". The new push will also be reformatted to have only one line per object.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: Changed to "Check that the right routine has been found for a single routine implementation."

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: Thanks. I'll also try to use this for other tests.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: I extended the comment:

    ! Use named argument 'd', which doesn't exist
    ! to trigger an error when searching for the matching routine.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: These if statements are very handy for debugging potential future problems. Setting some of them to '0' allows you to skip this subtest.

Also, the print statements provide helpful output again for debugging.
I don't understand why there should be no output. Is this a convention from PSyclone?

schreiberx: PS: The indentation with the "if 1" statements also structures such subtests nicely, hence, improving readability :-)

schreiberx:
Back to @arporter

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK

src/psyclone/psyir/nodes/call.py
```diff
@@ -47,11 +47,27 @@
...
+    for this routine
+    """
+    def __init__(self, value):
+        PSycloneError.__init__(self, value)
+        self.value = "CallMatchingArgumentsNotFound: " + str(value)
 
 
 class Call(Statement, DataNode):
-    ''' Node representing a Call. This can be found as a standalone statement
+    """Node representing a Call. This can be found as a standalone statement
```
arporter, line 70: Please could you revert these formatting changes (the change from ''' to """). The smaller a PR is, the fewer conflicts we get.

src/psyclone/psyir/nodes/call.py
```diff
@@ -111,17 +129,19 @@ def create(cls, routine, arguments=()):
...
         :raises GenerationError: if the arguments argument is not an Iterable.
 
-        '''
+        """
         if not isinstance(routine, (Reference, RoutineSymbol)):
             raise TypeError(
                 f"The Call routine argument should be a Reference to a "
                 f"RoutineSymbol or a RoutineSymbol, but "
-                f"found '{type(routine).__name__}'.")
+                f"found '{type(routine).__name__}'."
```
arporter, line 137: Please revert these formatting changes too. I don't mind if you're already changing that bit of code but otherwise it could well cause pain for other people.

src/psyclone/psyir/nodes/call.py
arporter: This is still outstanding?

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
arporter: Now that OPTIONAL will have a partial_datatype (this is on master), you can refactor this to compare `partial_datatype` if the datatype is of `UnsupportedFortranType`.

src/psyclone/psyir/nodes/call.py
```diff
@@ -592,4 +634,189 @@ def _location_txt(node):
...
+                if routine_arg is None:
+                    continue
+
+                if arg_name == routine_arg.name:
+                    # TODO #759: If optional is used, it's an unsupported
+                    # Fortran type and we need to use the following workaround
+                    # Once this issue is resolved, simply remove this if
+                    # branch.
+                    # Optional arguments are processed further down.
+                    if isinstance(
```
arporter, line 714: Same here, you should be able to use `partial_datatype` and drop the check for 'OPTIONAL' later.

src/psyclone/psyir/nodes/call.py
arporter: Actually, coming back to this and reading what you've written, I'm wondering whether this is a good idea - having a routine that, on the face of it, ensures the thing it returns matches the argument list, only to allow that to be disabled is a bit odd. Could you just use `get_callees` and take the first routine it returns instead of using this routine with this argument?

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -34,32 +34,49 @@
 # Authors: R. W. Ford and A. R. Porter, STFC Daresbury Lab
 # -----------------------------------------------------------------------------
 
-''' Performs py.test tests on the Call PSyIR node. '''
+""" Performs py.test tests on the Call PSyIR node. """
```
arporter, line 37: Please revert these formatting changes.

src/psyclone/tests/psyir/nodes/call_test.py
arporter: Yes, not having output from tests is a convention we have. It keeps things tidy. If a test is functioning correctly then it isn't needed and if a test fails then you're going to have to do some debugging anyway.

Regarding readability, I agree that is an issue with a test as long as this but I don't think these `if`s are the way to go. A better solution is to break the test up, possibly by using `@pytest.mark.parameterize` to provide the different `call` statements one at a time. Have a grep through the various test files to see how we use this. Breaking it up into smaller chunks also helps with debugging, as you say.

arporter (changes requested):
Thanks Martin. You'll see I'm getting a bit concerned that you're getting carried away with reformatting :-) I know it's hard to resist but it's important to keep PRs as small as possible to avoid conflicts for other people (or even yourself).
OPTIONAL arguments should now have a `partial_datatype` so I'm asking that you improve that bit of the code. I also think we can be more pythonic about passing back the matching argument list.

Just to re-iterate, please leave conversations unresolved as that really helps me see where we're up to as we progress.

src/psyclone/psyir/nodes/call.py
schreiberx: OK

schreiberx: It would be nice to have an agreement on what to you. It's again and again different across PSyclone.

schreiberx: Oh. Sorry, I see it's all over this file.
I assume it was done automatically by some reformatting tool. I'll undo it.

src/psyclone/psyir/nodes/call.py
schreiberx: OK

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: I think that a list of integers is sufficient. The optional arguments of a subroutine can be handled as follows (future work):

For inlining with optional arguments:
1. create a copy of the routine.
2. Search for optional arguments which are not in the list of indices.
3. For all of the optional arguments, replace PRESENT(...) with booleans.
4. Eliminate branches which are always true/false.
5. Check that there's no unused access to an optional argument left.

That's it. It's an algorithm which might not work in all possible cases, but it should with Nemo from what I saw so far.

src/psyclone/psyir/nodes/call.py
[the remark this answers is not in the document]
schreiberx: OK. I wrote this into a separate method since this is then reused also at other places of this code.

src/psyclone/psyir/nodes/call.py
schreiberx: OK. Replaced by `_check_argument_type_matches` method.

src/psyclone/psyir/nodes/call.py
schreiberx: Done.

schreiberx:
Back to @arporter

src/psyclone/psyir/nodes/call.py
schreiberx: This is an example how easily this list of indices can be used (it's not yet finished):

https://github.com/stfc/PSyclone/blob/martin_nemo_call_trace/src/psyclone/psyir/transformations/inline_trans.py#L990

schreiberx:
> Thanks Martin. You'll see I'm getting a bit concerned that you're getting carried away with reformatting :-) I know it's hard to resist but it's important to keep PRs as small as possible to avoid conflicts for other people (or even yourself). OPTIONAL arguments should now have a `partial_datatype` so I'm asking that you improve that bit of the code. I also think we can be more pythonic about passing back the matching argument list.
> 
> Just to re-iterate, please leave conversations unresolved as that really helps me see where we're up to as we progress.

OK. It's all done.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,188 @@ def _location_txt(node):
...
+            ) -> bool:
+        """Return information whether argument types are matching.
+        This also supports 'optional' arguments by using
+        partial types.
+
+        :param call_arg: _description_
+        :type call_arg: DataSymbol
+        :param routine_arg: _description_
+        :type routine_arg: DataSymbol
+        :raises CallMatchingArgumentsNotFound: _description_
```
arporter, line 627: Blank line before :raises: please and please populate the description. You could just have one :raises: as they both essentially mean 'no match'.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,188 @@ def _location_txt(node):
...
+                )
+        else:
+            if call_arg.datatype != routine_arg.datatype:
+                raise CallMatchingArgumentsNotFound(
+                    f"Argument type mismatch of call argument "
+                    f"'{call_arg}' and routine argument "
+                    f"'{routine_arg}'"
+                )
+
+        return True
```
arporter, line 652: No need to return anything.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,188 @@ def _location_txt(node):
             f"Failed to find a Routine named '{rsym.name}' in "
             f"{_location_txt(root_node)}. This is normally because the routine"
             f" is within a CodeBlock.")
+
+    def _check_argument_type_matches(
+                self,
+                call_arg: DataSymbol,
+                routine_arg: DataSymbol,
+            ) -> bool:
```
arporter, line 618: I don't think this routine needs a return value. (Also please update the docstring accordingly.)

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,188 @@ def _location_txt(node):
...
+                    f"Argument type mismatch of call argument "
+                    f"'{call_arg}' and routine argument "
+                    f"'{routine_arg}'"
+                )
+
+        return True
+
+    def _get_argument_routine_match(self, routine: Routine):
+        '''Return a list of integers giving for each argument of the call
+        the index of the argument in argument_list (typically of a routine)
```
arporter, line 656: "...of the corresponding entry in the argument list of the supplied routine."

src/psyclone/psyir/nodes/call.py
arporter: This is still outstanding.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +630,560 @@ def test_call_get_callees_local(fortran_reader):
...
+    assert len(arg_idx_list) == 2
+    assert arg_idx_list[0] == 0
+    assert arg_idx_list[1] == 1
+
+    assert result is routine_match
+
+
+def test_call_get_callee_3_trigger_error(fortran_reader):
+    '''
+    Test which is supposed to trigger an error.
```
arporter, line 713: Please specify the error - e.g. "error when no matching routine is found" or something?

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +630,560 @@ def test_call_get_callees_local(fortran_reader):
...
+  ! Matching routine
+  subroutine foo(a, b)
+    integer :: a, b
+  end subroutine
+
+end module some_mod'''
+
+    root_node: Node = fortran_reader.psyir_from_source(code)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
```
arporter, line 734: I don't think you need L734-L738.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +630,560 @@ def test_call_get_callees_local(fortran_reader):
...
+  subroutine main()
+    integer :: e_int, f_int, g_int
+    real :: e_real, f_real, g_real
+
+    ! Should match foo_a
+    call foo(e_int, f_int)
+
+    ! Should match foo_a
+    call foo(e_int, f_int, g_int)
+
```
arporter, line 854: Please rm this extra blank line and at L865.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +630,560 @@ def test_call_get_callees_local(fortran_reader):
...
+    real :: b
+    integer, optional :: c
+  end subroutine
+
+end module some_mod'''
+
+
+def test_call_get_callee_6_interfaces_0_0(fortran_reader):
+    '''
+    Check that optional and named arguments have been correlated correctly
```
arporter, line 898: Please extend each of the following docstrings to briefly say precisely which aspect of the correlation is being tested.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +630,560 @@ def test_call_get_callees_local(fortran_reader):
...
+    routine_main: Routine = psyir.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    call_foo: Call = routine_main.walk(Call)[0]
+
+    with pytest.raises(CallMatchingArgumentsNotFound) as err:
+        call_foo.get_callee()
+
+    assert "No matching routine found for 'call foo(e, f)" in str(err.value)
+
```
arporter, line 1185: Please could you also add a test to demonstrate that we don't match a routine if the *types* of the arguments don't match. You'll need to do this for both a named and a positional argument.

arporter (changes requested):
Thanks Martin, looking good. Main remaining things are to reach an agreement on how we pass around the list of matching arguments and to document the new tests. There's also some lines not covered by the tests. Please also check though the outstanding 'conversations'.

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/psyir/nodes/call.py
schreiberx: Done.
This will become obsolete here with my other PR where this is put into a class which has the set_option(check_matching_arguments).

src/psyclone/psyir/nodes/call.py
schreiberx: I changed it to "-> None".
The :rtype: should be covered by the "-> None".
Do we still need to write a Docstring for return types?

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: Understood: The entire import/form should be alphabetically ordered.
Done.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK. I did the same with all following ones as well and added also to the code which call is related to which test.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK. Seems that this got lost.

schreiberx:
> Thanks Martin, looking good. Main remaining things are to reach an agreement on how we pass around the list of matching arguments and to document the new tests. There's also some lines not covered by the tests. Please also check though the outstanding 'conversations'.

I wrote it with the list of integer numbers because it works two ways:
- For the call, you know which arguments are mapped to which routine argument.
- For the routine, you can check with the index of the routine (!) argument whether the argument is actually used.
Et voila, we're done. That's really all we need for this. That's how I realized it in the other PR where inlining now works with this.

If you like to only have the actual arguments, that doesn't work for both sides (call and routine). And if you still like to have them for a reason that is now known to me, just use this index list as a lookup in the argument list itself with a single comprehension list.

schreiberx:
PS: About the coverage issue: I also have 100% coverage in the future PR.
https://github.com/stfc/PSyclone/pull/2801
Can we do the coverage for this in the next PR?

src/psyclone/psyir/nodes/call.py
arporter: Not if it hasn't got one, no.

arporter: (You can remove the `return` statement altogether too)

arporter:
> Can we do the coverage for this in the next PR?

I'm afraid not. Especially since we might well release 3.0 in the interim.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: There's no difference between named and positional. Once the pairs or arguments (from call + routine) have been determined, the types are matched separately to this.

src/psyclone/psyir/nodes/call.py
schreiberx: OK

schreiberx:
"codecov/patch — 100.00% of diff hit" => back to @arporter  :-)

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,189 @@ def _location_txt(node):
...
+        """Return information whether argument types are matching.
+        This also supports 'optional' arguments by using
+        partial types.
+
+        :param call_arg: One argument of the call
+        :param routine_arg: One argument of the routine
+
+        :raises CallMatchingArgumentsNotFound: Raised if no matching argument
+            was found.
+        :raises CallMatchingArgumentsNotFound: _description_
```
arporter, line 628: Please rm this line.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,189 @@ def _location_txt(node):
...
+            routine_arg: DataSymbol
+
+            if routine_arg is None:
+                continue
+
+            # TODO #759: Optional keyword is not yet supported in psyir.
+            # Hence, we use a simple string match.
+            if ", OPTIONAL" not in str(routine_arg.datatype):
+                raise CallMatchingArgumentsNotFound(
+                    f"Argument '{routine_arg.name}' in subroutine"
```
arporter, line 733: This error message and the comment above need a bit of rewording now. e.g. the comment should say something like "Any arguments that haven't been matched must be OPTIONAL otherwise we have an error."
And the error message:
"Argument .... in subroutine .... does not match any in the call '{self.debug_string()}' and is not OPTIONAL.

src/psyclone/psyir/nodes/call.py
arporter: OK, I'll let this through for now then.

src/psyclone/psyir/nodes/call.py
arporter: You're argument about being able to use the list of integers to map in both directions seems good, thanks.

src/psyclone/tests/psyir/nodes/call_test.py
arporter: Thanks :-)

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,649 @@ def test_call_get_callees_local(fortran_reader):
...
+    root_node: Node = fortran_reader.psyir_from_source(_code_test_get_callee_6)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    routine_foo_c: Routine = root_node.walk(Routine)[3]
+    assert routine_foo_c.name == "foo_c"
+
+    print("Subtest foo_c[0]:")
```
arporter, line 1096: Please rm the print statements in this test.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -628,6 +629,649 @@ def test_call_get_callees_local(fortran_reader):
...
+    Check that matching a partial data type can also go wrong.
+    '''
+
+    root_node: Node = fortran_reader.psyir_from_source(_code_test_get_callee_6)
+
+    routine_main: Routine = root_node.walk(Routine)[0]
+    assert routine_main.name == "main"
+
+    routine_foo_optional: Routine = root_node.walk(Routine)[4]
+    print(routine_foo_optional.name)
```
arporter, line 1187: Please rm.

src/psyclone/psyir/nodes/call.py
```diff
@@ -593,3 +610,189 @@ def _location_txt(node):
...
+
+                if arg_name == routine_arg.name:
+                    self._check_argument_type_matches(call_arg, routine_arg)
+                    ret_arg_idx_list.append(routine_arg_idx)
+                    break
+
+            else:
+                # It doesn't match => Raise exception
+                raise CallMatchingArgumentsNotFound(
+                    f"Named argument '{arg_name}' not found"
```
arporter, line 715: "...found for routine '{routine.name}' in call '{self.debug_string()}'"

src/psyclone/tests/psyir/nodes/call_test.py
arporter: Sorry Martin, I might be being dense here but L1250 still talks about 'd' but neither the call nor the routine definition have a 'd'?

src/psyclone/tests/psyir/nodes/call_test.py
arporter: Generally speaking, it's best to attempt to test as though you don't know how it is implemented. However, I think the range of tests you have now covers this so I'm happy.

arporter (changes requested):
Thanks very much Martin. Pretty much there now bar some tidying.
Also, please could you extend the developer guide to mention the new `run_flake8.sh` script:
Please edit doc/developer_guide/working_practises.rst and add a new section called "Linting" immediately after the one on "Continuous Integration". This just needs to say that all code is linted by flake8 and must be free from errors to be merged with master. This check is performed automatically by the CI and it can be useful to setup a git hook to do this check before pushing to the repository. An example script is provided in ....

Please could you then refer to this section from the second bullet point in Continuous Integration.

I'll check the integration tests once #2697 is on (as that fixes the NEMO ones).

src/psyclone/psyir/nodes/call.py
schreiberx: OK

src/psyclone/psyir/nodes/call.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: OK.

src/psyclone/tests/psyir/nodes/call_test.py
schreiberx: Indeed, 'd' didn't make sense here.
I hope I treated this now correctly.

schreiberx:
PS: I wasn't able to reply to "Use name 'd' which doesn't exist", but did the changes in the comments. Thanks for spotting this error.

schreiberx:
I don't know what's causing this build problem of build-documentation:
```
/home/runner/work/PSyclone/PSyclone/doc/user_guide/index.rst:115: WARNING: toctree contains reference to document 'zz_bibliography' that doesn't have a title: no link will be generated [toc.no_title]
```
I'm not able to reproduce it on my computer.

src/psyclone/psyir/nodes/call.py
```diff
@@ -712,7 +712,8 @@ def _get_argument_routine_match(self, routine: Routine):
             else:
                 # It doesn't match => Raise exception
                 raise CallMatchingArgumentsNotFound(
-                    f"Named argument '{arg_name}' not found"
+                    f"Named argument '{arg_name}' found for routine"
```
arporter, line 715: This is missing a "not" but I can do this if that's all that's left.

src/psyclone/psyir/nodes/call.py
arporter: You've missed the comment but again, I can do this.

arporter (approved):
Thanks very much Martin.
This is now ready to be merged :-)
I'll bring it up to master and tweak the two bits of text that I've mentioned in the review. Then I'll run the integration tests. Then I shall merge it...

src/psyclone/psyir/nodes/call.py
arporter: Done.

src/psyclone/psyir/nodes/call.py
arporter: Actually, looking at it in context, this is already covered by an earlier comment so it's fine.

arporter:
Integration tests have finally finished and were green. Will proceed to merge.

PR #2776, opened 2024-11-13
(Closes #1280) fix ref guide warnings

### LINKED ISSUE #1280: Warnings from Sphinx autoapi when building reference guide
While doing #1194 it was found that adding an `__all__` for `TransformationError` to `psyclone/psyir/transformations/transformation_error.py` caused over 130 'more than one target for cross-reference' warnings:

    Projects/PSyclone/src/psyclone/transformations.py:docstring of psyclone.transformations.KernelGlobalsToArguments.validate:: WARNING: more than one target found for cross-reference 'TransformationError': psyclone.psyir.transformations.TransformationError, psyclone.psyir.transformations.transformation_error.TransformationError

Without that `__all__` we do not get detailed documentation generated for the `TransformationError` class. This problem has been raised on the sphinx mailing list: https://groups.google.com/g/sphinx-users/c/vuW6OOb96Yo

We also get errors about duplicate documentation of classes that are themselves defined within other classes (such as psyclone.domain.lfric.kern_call_arg_list.NdfInfo).

arporter:
Fixes all warnings when building the Ref. Guide and enables the check for warnings in the CI. A lot of the warnings have been fixed by switching from named tuples to DataClasses. This also enables better documentation.
Some of the warnings have been fixed by removing unnecessary setting of `__all__` in some locations.
Finally, the only way I found to resolve warnings about `Operation.Operator` (and subclasses) was to register a handler with autodoc. This handler tells autodoc not to generate documentation for things called Operator that are an Enum.

### DIFF
--- a/changelog
+++ b/changelog
@@ -279,6 +279,8 @@
 
 	106) PR #2770 towards #2543. Improve UG introduction.
 
+	107) PR #2776 for #1280. Fixes ref_guide warnings.
+
 release 2.5.0 14th of February 2024
 
 	1) PR #2199 for #2189. Fix bugs with missing maps in enter data


arporter:
In fact, the CI doesn't test the Jupyter notebooks anymore as they are out of date. We could probably therefore relax our dependencies?

arporter:
Another update. The CI installs jupyter itself (i.e. it's not in setup.py) and then does run the notebooks in examples/gocean/eg1

.github/workflows/python-package.yml
```diff
@@ -76,7 +76,7 @@ jobs:
     # Now we can check for warnings and broken links
     - run: cd doc/user_guide; make html SPHINXOPTS="-W --keep-going"
     - run: cd doc/developer_guide; make html SPHINXOPTS="-W --keep-going"
-    # TODO #1280: - run: cd doc/reference_guide; make html SPHINXOPTS="-W --keep-going"
+    - run: cd doc/reference_guide; make html SPHINXOPTS="-W --keep-going"
```
sergisiso, line 79: Well done! That will make the reviews a step easier 🥳

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -870,10 +897,21 @@ class StructureType(DataType):
...
+        Represents a member of a StructureType.
+
+        :param name: the name of the member.
+        :param datatype: the type of the member.
+        :param visibility: whether this member is public or private.
+        :param initial_value: the initial value of this member (if any).
+        :type initial_value: Optional[:py:class:`psyclone.psyir.nodes.Node`]
+        '''
+        name: str
+        datatype: DataType | DataTypeSymbol
```
sergisiso, line 912: This make CI fail:
```
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/psyclone/psyir/symbols/datatypes.py", line 912, in ComponentType
    datatype: DataType | DataTypeSymbol
TypeError: unsupported operand type(s) for |: 'ABCMeta' and 'type'
```
I think you will need to use `Union` importing it from `typing`.

sergisiso (changes requested):
Thanks for doing this @arporter. I don't have much to say other than I really like that the doc testing is now automatic and that the dataclasses are much better documented than the previous namedtuples.

The only problem is that the CI failed (see proposed fix inline), also the NEMOv5 OpenMP offloading integration failed, but this second issue is probably my fault (sorry), as I upgraded the runner folder to a more recent version of NEMOv5 and it is working with my upcoming branch. So feel free to ignore this integration test for now.

Also this PR has some good comparisons of using type hinting vs docstring:
image
The important thing is in both the types are "clickable" to go to their definition.

src/psyclone/psyir/symbols/datatypes.py
arporter: Thanks yes. It turns out I hit 'ready for review' a bit too soon. There was another, similar problem in the new handler in conf.py that I've also had to fix.

arporter:
> The important thing is in both the types are "clickable" to go to their definition.

Yes, it's good. Ideally I'd have done the whole thing with type hinting but having `Node` types within the `datatypes` module caused circular dependencies.

.github/workflows/python-package.yml
arporter: Yes, I found we'd already broken the Ref Guide with the FFSL PR!

arporter:
NEMO4 integration test failed but it's the usual race condition problem.

arporter:
OK, 3.8 works again now. Ready for another look.

sergisiso:
@arporter The compilation tests also fail due to an issue in my last PR, it tries to do:
`make -C ../../../lib/extract/standalone/nemo`
but the last 'nemo' should be 'generic' now. Could you fix this and launch the integration tests again.

arporter:
Oh yes. Fixed. I'll re-launch the integration tests.

arporter:
Compilation tests finally OK. Ready for another look.

sergisiso (approved):
Thanks @arporter All comments have been addressed, integration test pass (at least the ones that should pass). This is ready to merge.

PR #2795, opened 2024-11-20
(Closes #2774) Refactor matmul2code to allow for more permutations of full ranges

### LINKED ISSUE #2774: [LFRic] [PSyAD] Change in operator index ordering does not match PSyAD's matmul standard
Previously, an valid operator from PSyAD's perspective would be of the form `operator( : , : , idx )`, i.e.: the first two indices are full ranges. LFRic is now changing the index accessing to be like `operator( idx, : , : )` which does not fit the standard imposed by PSyAD. This causes the following failure:

```
psyclone.psyir.transformations.transformation_error.TransformationError: Transformation Error: To use matmul2code_trans on matmul, the first two indices of the 1st argument 'operator' must be full ranges.
```

DrTVockerodtMO:
Fixes #2774.

There was a hardcoded standard in PSyclone where for matrix multiplications, the first one or two indices must be full ranges (i.e.: the Fortran would be written as `matrix(:,:,...)` or `vector(:,...)`). However, Fortran can perform matrix multiplications with the full ranges in different orderings, as long as the shape conforms.

When recent LFRic changes re-ordered some of the operator index accesses, this shifted the full ranges into forbidden permutations.

This change addresses this issue by allowing more permutations that produce valid Fortran matmul code.

**Description of changes**
Added a new private method `_get_full_range_split` to matmul2code_trans.py, which splits information about full range and non full range nodes as well as performs some validation. For a given `array`, the method iterates over the children, and checks whether or not they are `Range` nodes. If the `Range` node is a full range, we append its position in the array to a list. If not, we throw an error. If it is not a `Range` node then its position and the node itself are appended to two different lists. 

After the children are iterated over, if we have more than two full ranges then this `array` is not a valid input for matmul, so we throw an error. Otherwise, return the two lists containing the positions of the full range and non full range nodes in the array, as well as the non full range node list.

Some further validation is performed to ensure that the first matrix involved in the matmul has exactly two full ranges, and that the second matrix has one or two (since it can be a vector or matrix).

The orderings and list of non full range nodes are fed to the modified `_create_matrix_ref` to make the matrix reference.

**Testing**
A new test `test_apply_matmat_reordered` has been added which tests the expected behaviour directly, by feeding the PSyIR some Fortran code with spatially conforming but re-ordered (with respect to the old standard) matrices. Other tests will be modified since a lot of them tested aspects of the old standard that no longer apply.

### DIFF
--- a/src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
+++ b/src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
@@ -52,7 +52,8 @@ from psyclone.psyir.transformations.intrinsics.intrinsic2code_trans import (
     Intrinsic2CodeTrans)
 
 
-def _create_matrix_ref(matrix_symbol, loop_idx_symbols, other_dims, order):
+def _create_matrix_ref(matrix_symbol, loop_idx_symbols, other_dims,
+                       loop_idx_order, other_dims_order):
     '''
     Utility function to create a reference to a matrix element being
     accessed using one or more loop indices followed by zero or more
@@ -74,16 +75,18 @@ def _create_matrix_ref(matrix_symbol, loop_idx_symbols, other_dims, order):
     :rtype: :py:class:`psyclone.psyir.nodes.ArrayReference`
 
     '''
-    n_indices = len(loop_idx_symbols) + len(other_dims)
-    indices = [-1]*n_indices
-    for it, order_idx in enumerate(order):
-        indices[order_idx] = other_dims[it].copy()
-    # Fill the remaining indices with loop index symbols
-    for it, sym in enumerate(loop_idx_symbols):
-        for i in range(n_indices):
-            if indices[i] == -1:
-                indices[i] = Reference(sym)
-                break
+    # If there are no other dims, then we simply create
+    # a matrix with the loop_idx_symbols in order.
+    if len(other_dims) == 0:
+        indices = [Reference(sym) for sym in loop_idx_symbols]
+    else:
+        n_indices = len(loop_idx_symbols) + len(other_dims)
+        indices = [-1]*n_indices
+        for it, order_idx in enumerate(other_dims_order):
+            indices[order_idx] = other_dims[it].copy()
+        # Fill the remaining indices with loop index symbols
+        for it, order_idx in enumerate(loop_idx_order):
+            indices[order_idx] = Reference(loop_idx_symbols[it])
     return ArrayReference.create(matrix_symbol, indices)
 
 
@@ -159,13 +162,13 @@ def _get_full_range_split(array):
         if there are more than two full range nodes.
 
     '''
-    n_full_ranges = 0
     non_full_ranges = []
-    order = []
+    full_range_order = []
+    non_full_range_order = []
     for it, idx in enumerate(array.children):
         if isinstance(idx, Range):
             if array.is_full_range(it):
-                n_full_ranges += 1
+                full_range_order.append(it)
             else:
                 from psyclone.psyir.transformations import TransformationError
                 raise TransformationError(
@@ -174,36 +177,15 @@ def _get_full_range_split(array):
                     f"but found non full range at position {it}.")
         else:
             non_full_ranges.append(idx)
-            order.append(it)
+            non_full_range_order.append(it)
         # Early error raising if we go above 2 full ranges
-        if n_full_ranges > 2:
+        if len(full_range_order) > 2:
             from psyclone.psyir.transformations import TransformationError
             raise TransformationError(
                 f"To use matmul2code_trans on matmul, no more than "
                 f"two indices of the argument '{array.name}' "
                 f"must be full ranges but found {n_full_ranges}.")
-    return (n_full_ranges, non_full_ranges, order)
-
-
-def _get_first_full_range_idx(array, reverse = False):
-
-    # Find last index if asked to
-    child_enumerate = enumerate(array.children)
-    if reverse:
-        child_enumerate = reversed(list(child_enumerate))
-
-    for it, idx in child_enumerate:
-        if isinstance(idx, Range):
-            if array.is_full_range(it):
-                return it
-            else:
-                from psyclone.psyir.transformations import TransformationError
-                raise TransformationError(
-                    f"To use matmul2code_trans on matmul, each Range index "
-                    f"of the argument '{array.name}' must be a full range "
-                    f"but found non full range at position {it}.")
-    # Default behaviour
-    return -int(reverse == True)
+    return (full_range_order, non_full_range_order, non_full_ranges)
 
 
 class Matmul2CodeTrans(Intrinsic2CodeTrans):
@@ -331,7 +313,8 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
             # There should be one index per dimension. This is enforced
             # by the array create method so is not tested here.
             # For matrix1, we must have exactly 2 full ranges.
-            n_full_ranges, _, _ = _get_full_range_split(matrix1)
+            full_range_order, _, _ = _get_full_range_split(matrix1)
+            n_full_ranges = len(full_range_order)
             if n_full_ranges != 2:
                 raise TransformationError(
                     f"To use matmul2code_trans on matmul, exactly two "
@@ -354,7 +337,8 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
             # There should be one index per dimension. This is enforced
             # by the array create method so is not tested here.
             # For matrix2, we can have 1 or 2 full ranges.
-            n_full_ranges, _, _ = _get_full_range_split(matrix2)
+            full_range_order, _, _ = _get_full_range_split(matrix2)
+            n_full_ranges = len(full_range_order)
             if n_full_ranges not in [1,2]:
                 raise TransformationError(
                     f"To use matmul2code_trans on matmul, one or two "
@@ -442,12 +426,13 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
                 vector_dims.append(child.copy())
         vector_array_reference = ArrayReference.create(
             vector.symbol, vector_dims)
-        _, ref_non_full_ranges, order = _get_full_range_split(matrix)
+        fr_order, nfr_order, non_full_ranges = _get_full_range_split(matrix)
         # Create "matrix(i,j)"
         matrix_array_reference = _create_matrix_ref(matrix.symbol,
                                                     [i_loop_sym, j_loop_sym],
-                                                    ref_non_full_ranges,
-                                                    order)
+                                                    non_full_ranges,
+                                                    fr_order,
+                                                    nfr_order)
         # Create "matrix(i,j) * vector(j)"
         multiply = BinaryOperation.create(
             BinaryOperation.Operator.MUL, matrix_array_reference,
@@ -459,7 +444,13 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
         assign = Assignment.create(result_ref.copy(), rhs)
         # Create j loop and add the above code as a child
         # Work out the bounds
-        first_pos = _get_first_full_range_idx(vector)
+        vec_full_range_order, _, _ = _get_full_range_split(vector)
+        if len(vec_full_range_order) == 0:
+            # If no full ranges, then the vector was specified
+            # without them i.e.: only has one dimension.
+            first_pos = 0
+        else:
+            first_pos = vec_full_range_order[0]
         lower_bound, upper_bound, step = _get_array_bound(vector, first_pos)
         jloop = Loop.create(j_loop_sym, lower_bound, upper_bound, step,
                             [assign])
@@ -467,7 +458,12 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
         assign = Assignment.create(result_ref.copy(),
                                    Literal("0.0", REAL_TYPE))
         # Create i loop and add assignment and j loop as children
-        first_pos = _get_first_full_range_idx(matrix)
+        if len(fr_order) == 0:
+            # If no full ranges, then matrix was specified
+            # without them i.e.: only has two dimensions.
+            first_pos = 0
+        else:
+            first_pos = fr_order[0]
         lower_bound, upper_bound, step = _get_array_bound(matrix, first_pos)
         iloop = Loop.create(i_loop_sym, lower_bound, upper_bound, step,
                             [assign, jloop])
@@ -499,23 +495,26 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
         ii_loop_sym = symbol_table.new_symbol("ii", symbol_type=DataSymbol,
                                               datatype=INTEGER_TYPE)
         # Create "result(i,j)"
-        _, res_non_full_ranges, order = _get_full_range_split(result)
+        fr_order, nfr_order, non_full_ranges = _get_full_range_split(result)
         result_ref = _create_matrix_ref(result.symbol,
                                         [i_loop_sym, j_loop_sym],
-                                        res_non_full_ranges,
-                                        order)
+                                        non_full_ranges,
+                                        fr_order,
+                                        nfr_order)
         # Create "matrix2(ii,j)"
-        _, m2_non_full_ranges, order = _get_full_range_split(matrix2)
+        m2_fr_order, m2_nfr_order, m2_nfr = _get_full_range_split(matrix2)
         m2_array_reference = _create_matrix_ref(matrix2.symbol,
                                                 [ii_loop_sym, j_loop_sym],
-                                                m2_non_full_ranges,
-                                                order)
+                                                m2_nfr,
+                                                m2_fr_order,
+                                                m2_nfr_order)
         # Create "matrix1(i,ii)"
-        _, m1_non_full_ranges, order = _get_full_range_split(matrix1)
+        m1_fr_order, m1_nfr_order, m1_nfr = _get_full_range_split(matrix1)
         m1_array_reference = _create_matrix_ref(matrix1.symbol,
                                                 [i_loop_sym, ii_loop_sym],
-                                                m1_non_full_ranges,
-                                                order)
+                                                m1_nfr,
+                                                m1_fr_order,
+                                                m1_nfr_order)
         # Create "matrix1(i,ii) * matrix2(ii,j)"
         multiply = BinaryOperation.create(
             BinaryOperation.Operator.MUL, m1_array_reference,
@@ -527,7 +526,12 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
         assign = Assignment.create(result_ref.copy(), rhs)
         # Create ii loop and add the above code as a child
         # Work out the bounds
-        pos_last = _get_first_full_range_idx(matrix1, reverse=True)
+        if len(m1_fr_order) == 0:
+            # If no full ranges, then matrix was specified
+            # without them i.e.: only has two dimensions.
+            pos_last = 1
+        else:
+            pos_last = m1_fr_order[-1]
         lower_bound, upper_bound, step = _get_array_bound(matrix1, pos_last)
         # Must be the same as _get_array_bound(matrix2, pos_first)
         iiloop = Loop.create(ii_loop_sym, lower_bound, upper_bound, step,
@@ -536,12 +540,22 @@ class Matmul2CodeTrans(Intrinsic2CodeTrans):
         assign = Assignment.create(result_ref.copy(),
                                    Literal("0.0", REAL_TYPE))
         # Create i loop and add assignment and ii loop as children.
-        pos_first = _get_first_full_range_idx(matrix1)
+        if len(m1_fr_order) == 0:
+            # If no full ranges, then matrix was specified
+            # without them i.e.: only has two dimensions.
+            pos_first = 0
+        else:
+            pos_first = m1_fr_order[0]
         lower_bound, upper_bound, step = _get_array_bound(matrix1, pos_first)
         iloop = Loop.create(i_loop_sym, lower_bound, upper_bound, step,
                             [assign, iiloop])
         # Create j loop and add i loop as child.
-        pos_last = _get_first_full_range_idx(matrix2, reverse=True)
+        if len(m2_fr_order) == 0:
+            # If no full ranges, then matrix was specified
+            # without them i.e.: only has two dimensions.
+            pos_last = 1
+        else:
+            pos_last = m2_fr_order[-1]
         lower_bound, upper_bound, step = _get_array_bound(matrix2, pos_last)
         jloop = Loop.create(j_loop_sym, lower_bound, upper_bound, step,
                             [iloop])
--- a/src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
+++ b/src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
@@ -67,9 +67,12 @@ def _create_matrix_ref(matrix_symbol, loop_idx_symbols, other_dims,
                        not being looped over. (If there are none then this \
                        must be an empty list.)
     :type other_dims: List[:py:class:`psyclone.psyir.nodes.ExpressionNode`]
-    :param order:      list of indices in the original matrix where the
-                       other_dims are.
-    :type order:      List[int]
+    :param loop_idx_order: list of indices in the original matrix where the
+                           loop_idx_symbols are.
+    :type loop_idx_order: List[int]
+    :param other_dims_order: list of indices in the original matrix where the
+                             other_dims are.
+    :type other_dims_order: List[int]
 
     :returns: the new reference to a matrix element.
     :rtype: :py:class:`psyclone.psyir.nodes.ArrayReference`
@@ -154,9 +157,10 @@ def _get_full_range_split(array):
 
     :param array: the reference that we are interested in.
     :type array: :py:class:`psyir.nodes.Reference`
-    :returns: tuple with number of full ranges and
-              the list of non full range indices for this array.
-    :rtype: (int, List[psyclone.psyir.nodes.DataNode])
+    :returns: tuple with list of full range indices,
+              the list of non full range indices, and
+              list of non full range nodes for this array.
+    :rtype: (List[int], List[int], List[psyclone.psyir.nodes.DataNode])
 
    :raises TransformationError: if any Range nodes are not full, or
         if there are more than two full range nodes.
@@ -180,7 +184,8 @@ def _get_full_range_split(array):
             non_full_range_order.append(it)
 
     # Error raising if we go above 2 full ranges
-    if len(full_range_order) > 2:
+    n_full_ranges = len(full_range_order)
+    if n_full_ranges > 2:
         from psyclone.psyir.transformations import TransformationError
         raise TransformationError(
             f"To use matmul2code_trans on matmul, no more than "
--- a/src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
+++ b/src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
@@ -33,6 +33,7 @@
 # -----------------------------------------------------------------------------
 # Author: R. W. Ford, STFC Daresbury Laboratory
 # Modified: S. Siso, A. R. Porter and N. Nobre, STFC Daresbury Lab
+#           T. Vockerodt, Met Office
 
 '''Module providing a transformation from a PSyIR MATMUL operator to
 PSyIR code. This could be useful if the MATMUL operator is not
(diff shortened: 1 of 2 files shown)


src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -257,28 +318,14 @@ def validate(self, node, options=None):
...
-            if len(matrix1.children) > 2:
-                # The 3rd index and onwards must not be ranges.
-                for (count, index) in enumerate(matrix1.children[2:]):
-                    if isinstance(index, Range):
-                        raise TransformationError(
-                            f"To use matmul2code_trans on matmul, only the "
-                            f"first two indices of the 1st argument are "
-                            f"permitted to be Ranges but found "
-                            f"{type(index).__name__} at index {count+2}.")
+                    f"To use matmul2code_trans on matmul, exactly two "
```
arporter, line 326: Not yours but while we're here would you mind replacing "matmul2code_trans" with "{self.name}".

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -257,28 +318,14 @@ def validate(self, node, options=None):
...
+                    f"indices of the 1st argument '{matrix1.name}' "
```
arporter, line 327: We can do a bit better now - please replace 'matrix1.name' with 'matrix1.debug_string()'

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -295,32 +342,14 @@ def validate(self, node, options=None):
...
-            if len(matrix2.children) > 2:
-                # The 3rd index and onwards must not be ranges.
-                for (count, index) in enumerate(matrix2.children[2:]):
-                    if isinstance(index, Range):
-                        raise TransformationError(
-                            f"To use matmul2code_trans on matmul, only the "
-                            f"first two indices of the 2nd argument are "
-                            f"permitted to be a Range but found "
-                            f"{type(index).__name__} at index {count+2}.")
+                    f"To use matmul2code_trans on matmul, one or two "
```
arporter, line 350: self.name and matrix2.debug_string() please.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -478,19 +532,37 @@ def _apply_matrix_matrix(node):
...
-        lower_bound, upper_bound, step = _get_array_bound(matrix1, 1)
-        # Must be the same as _get_array_bound(matrix2, 0)
+        if len(m1_fr_order) == 0:
+            # If no full ranges, then matrix was specified
+            # without them i.e.: only has two dimensions.
+            pos_last = 1
+        else:
+            pos_last = m1_fr_order[-1]
+        lower_bound, upper_bound, step = _get_array_bound(matrix1, pos_last)
+        # Must be the same as _get_array_bound(matrix2, pos_first)
```
arporter, line 542: I don't understand this comment as there's no use of this fact in the code? (Probably a good thing should we ever generalise this to allow partial ranges.)

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -478,19 +532,37 @@ def _apply_matrix_matrix(node):
...
+            pos_first = 0
+        else:
+            pos_first = m1_fr_order[0]
+        lower_bound, upper_bound, step = _get_array_bound(matrix1, pos_first)
         iloop = Loop.create(i_loop_sym, lower_bound, upper_bound, step,
                             [assign, iiloop])
         # Create j loop and add i loop as child.
-        lower_bound, upper_bound, step = _get_array_bound(matrix2, 1)
+        if len(m2_fr_order) == 0:
+            # If no full ranges, then matrix was specified
```
arporter, line 560: We could get rid of all these "if no full ranges" branches by applying `Reference2ArrayRangeTrans` as part of this transformation but that's a bigger change and this is urgent.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -507,64 +531,70 @@ def test_validate11():
     matrix.children[2] = my_range
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("To use matmul2code_trans on matmul, only the first two indices "
-            "of the 1st argument are permitted to be Ranges but "
-            "found Range at index 2." in str(excinfo.value))
+    assert ("Transformation Error: To use matmul2code_trans on matmul, "
+            "each Range index of the argument 'x' must be a full range "
+            "but found non full range at position 2." in str(excinfo.value))
```
arporter, line 536: Again, we can use `debug_string()` to improve this error message now.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -507,64 +531,70 @@ def test_validate11():
...
+    assert ("Transformation Error: To use matmul2code_trans on matmul, "
+            "each Range index of the argument 'y' must be a full range "
+            "but found non full range at position 2." in str(excinfo.value))
 
 
-def test_validate_2nd_dim_2nd_arg():
-    ''' Check that the Matmul2Code validate method raises the expected
-    exception when the second dimension of the second argument to MATMUL
-    is not a full range. '''
+def test_validate14():
```
arporter, line 557: You don't have to do this now but I would stay away from numbering tests like this. Rupert used to like doing it this way but I found that it was a pain when you came back and wanted to add new tests. It's also nice to have informative names.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -507,64 +531,70 @@ def test_validate11():
...
+    matrix = matmul.arguments[0]
+    matrix.children[2] = Range.create(Literal("1", INTEGER_TYPE),
+                                      Literal("15", INTEGER_TYPE))
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("To use matmul2code_trans on matmul for a matrix-matrix "
-            "multiplication, the second index of the 2nd argument 'y' must "
-            "be a full range." in str(excinfo.value))
+    assert ("Transformation Error: To use matmul2code_trans on matmul, "
+            "no more than two indices of the argument 'x' "
```
arporter, line 572: `debug_string()` again.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -925,3 +955,42 @@ def test_apply_matmat_name_clashes(tmpdir, fortran_reader, fortran_writer):
...
+        "  real, dimension(2,4,6) :: jac\n"
+        "  real, dimension(4,6,3) :: jac_inv\n"
+        "  real, dimension(2,3) :: result\n"
+        "  result(:,:) = matmul(jac(:,1,:), jac_inv(2,:,:))\n"
+        "end subroutine my_sub\n")
+    trans = Matmul2CodeTrans()
+    assign = psyir.walk(Assignment)[0]
+    trans.apply(assign.rhs)
+    out = fortran_writer(psyir)
+    print(out)
```
arporter, line 979: Please rm this print.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -925,3 +955,42 @@ def test_apply_matmat_name_clashes(tmpdir, fortran_reader, fortran_writer):
...
+        "  real, dimension(2,3) :: result\n"
+        "  integer :: i\n"
+        "  integer :: j\n"
+        "  integer :: ii\n"
+        "\n"
+        "  do j = 1, 3, 1\n"
+        "    do i = 1, 2, 1\n"
+        "      result(i,j) = 0.0\n"
+        "      do ii = 1, 6, 1\n"
+        "        result(i,j) = result(i,j) + jac(i,1,ii) * jac_inv(2,ii,j)\n"
```
arporter, line 992: This is great, well done. Please could you add a similar test for a matrix-vector (e.g., when you have `jac_inv(2,:,arg+3)` as the second argument to matmul. It would also be valuable, as I indicate there (the `arg+3`), to ensure that the indices that aren't full ranges can be anything we like.

arporter (changes requested):
Good stuff @DrTVockerodtMO.
I think the fix/extension is good. Just a few minor things to tidy up and I've requested an extra test.
Although not strictly your fault, I note that the associated test file does not give 100% coverage - lines 424-425 are missed.  You can see this by doing `pytest --cov psyclone.psyir.transformations.intrinsics.matmul2code_trans --
cov-report term-missing psyir/transformations/intrinsics/matmul2code_trans_test.py`. Could you take a look?
Please could you also update the class docstring of `apply()` to mention the restriction to full ranges and simple arrays (i.e. structures of arrays etc.).
I've fired off the integration tests so these can run in the meantime.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -65,13 +67,29 @@ def _create_matrix_ref(matrix_symbol, loop_idx_symbols, other_dims):
                        not being looped over. (If there are none then this \
                        must be an empty list.)
     :type other_dims: List[:py:class:`psyclone.psyir.nodes.ExpressionNode`]
+    :param loop_idx_order: list of indices in the original matrix where the
```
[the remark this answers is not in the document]
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227. The comment was previously there without influencing the functionality so I initially just changed the notation, but now I have removed it.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227.

arporter:
Just a note to say integration tests were OK (barring the expected failures).

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227. I tried to rename all the numbered validation tests into something that their docstring describes.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in 10d5abbfb758dbaa45439829f974a99504e27227. This was a really good shout because there actually wasn't a coverage test for vectors without the first index being a full range.

To get this to work I changed the condition in the `apply` method to look at the number of full ranges, to determine whether it's a matrix-matrix or matrix-vector application. This also required changes in the matrix-vector application to properly pack the vector to create the array reference.

It's the same code as the old `_create_matrix_ref` so I simply renamed this to `_create_array_ref`.

DrTVockerodtMO:
> Good stuff @DrTVockerodtMO. I think the fix/extension is good. Just a few minor things to tidy up and I've requested an extra test. Although not strictly your fault, I note that the associated test file does not give 100% coverage - lines 424-425 are missed. You can see this by doing `pytest --cov psyclone.psyir.transformations.intrinsics.matmul2code_trans -- cov-report term-missing psyir/transformations/intrinsics/matmul2code_trans_test.py`. Could you take a look? Please could you also update the class docstring of `apply()` to mention the restriction to full ranges and simple arrays (i.e. structures of arrays etc.). I've fired off the integration tests so these can run in the meantime.

Changes should all be addressed. Regarding not covering lines 424-425, these lines have been replaced in favour of the new method of generating array references, which are tested by the new test that was requested.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -494,12 +494,12 @@ def test_validate10():
     matrix.children[0] = Literal("1", INTEGER_TYPE)
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("To use matmul2code_trans on matmul, exactly two indices of the "
-            "1st argument 'x' must be full ranges "
-            "but found 1." in str(excinfo.value))
+    assert (f"To use {trans.name} on matmul, exactly two indices of the "
```
arporter, line 497: Please could you change this so that it checks that the output string is what we expect, rather than what we coded. i.e. this doesn't need to be an f-string - just hardwire in the values that we expect.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -513,12 +513,13 @@ def test_validate11():
     vector.children[0] = my_index
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("Transformation Error: To use matmul2code_trans on matmul, "
-            "one or two indices of the 2nd argument 'y' "
-            "must be full ranges but found 0." in str(excinfo.value))
+    assert (f"Transformation Error: To use {trans.name} on matmul, "
```
arporter, line 516: Again, no f-string here please.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -531,12 +532,13 @@ def test_validate12():
     matrix.children[2] = my_range
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("Transformation Error: To use matmul2code_trans on matmul, "
-            "each Range index of the argument 'x' must be a full range "
-            "but found non full range at position 2." in str(excinfo.value))
+    assert (f"Transformation Error: To use {trans.name} on matmul, "
```
arporter, line 535: No f-string...

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -549,12 +551,13 @@ def test_validate13():
     vector.children[2] = my_range
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("Transformation Error: To use matmul2code_trans on matmul, "
-            "each Range index of the argument 'y' must be a full range "
-            "but found non full range at position 2." in str(excinfo.value))
+    assert (f"Transformation Error: To use {trans.name} on matmul, "
```
arporter, line 554: No f-string...

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -568,12 +571,13 @@ def test_validate14():
                                       Literal("15", INTEGER_TYPE))
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("Transformation Error: To use matmul2code_trans on matmul, "
-            "no more than two indices of the argument 'x' "
-            "must be full ranges but found 3." in str(excinfo.value))
+    assert (f"Transformation Error: To use {trans.name} on matmul, "
```
arporter, line 574: No f-string...

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -589,12 +593,13 @@ def test_validate15():
                                       Literal("10", INTEGER_TYPE))
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(matmul)
-    assert ("Transformation Error: To use matmul2code_trans on matmul, "
-            "no more than two indices of the argument 'y' "
-            "must be full ranges but found 3." in str(excinfo.value))
+    assert (f"Transformation Error: To use {trans.name} on matmul, "
```
arporter, line 596: No f-string...

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -622,9 +627,9 @@ def test_validate_matmat_with_slices_on_rhs(fortran_reader):
     assign = psyir.walk(Assignment)[0]
     with pytest.raises(TransformationError) as excinfo:
         trans.validate(assign.rhs)
-    assert ("To use matmul2code_trans on matmul, each range on the result "
-            "variable 'result' must be a full range but found "
-            "result(2:4,2:5)" in str(excinfo.value))
+    assert (f"To use {trans.name} on matmul, each range on the result "
```
arporter, line 630: "No f-string"

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
arporter: Thanks for all that renaming. Above and beyond :-)

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
arporter: Ah, I see now that I might have confused you here - apologies. I meant 'use `debug_string()` when creating the error message', not when testing it!

arporter:
> Changes should all be addressed. Regarding not covering lines 424-425, these lines have been replaced in favour of the new method of generating array references, which are tested by the new test that was requested.

That's good but those lines are still there and are still not covered? On closer examination it looks as though this code will only work for something of the form `result(:,:,any,other,indices) = matmul(....)` but this isn't checked in validate. If it's not too hard, do you think you could generalise this bit too? Alternatively, we should check for this situation in `validate` and reject it.

arporter (changes requested):
Thanks Terry. Apologies for the confusion over the strings (KGOs) in the tests.
Apart from that, looking at the missed lines made me realise that we still seem to have an implicit assumption in the code about the form of the `result` array that we support.
We can either generalise this too or we can make the assumption explicit by checking for it in `validate`.
The good news is that the integration tests were all fine :-)

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in c4d60bc3ef52a6d7b2853dd5b3c3b01ce82996dc.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -52,34 +52,33 @@
...
-                           loop_idx_symbols are.
-    :type loop_idx_order: List[int]
-    :param other_dims_order: list of indices in the original matrix where the
-                             other_dims are.
-    :type other_dims_order: List[int]
-
-    :returns: the new reference to a matrix element.
+    :param List[int] loop_idx_order: list of indices in the original array \
+                                     where the loop_idx_symbols are.
+
```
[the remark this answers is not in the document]
DrTVockerodtMO: Addressed in 4629275d73a0373c1f3f3f2ced3aadb35f198d49.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -994,3 +998,40 @@ def test_apply_matmat_reordered(tmpdir, fortran_reader, fortran_writer):
         "    enddo\n"
         "  enddo\n" in out)
     assert Compile(tmpdir).string_compiles(out)
+
+
+def test_apply_matmat_varexpr_index(tmpdir, fortran_reader, fortran_writer):
```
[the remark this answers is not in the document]
DrTVockerodtMO: Addressed in 4629275d73a0373c1f3f3f2ced3aadb35f198d49.

DrTVockerodtMO:
> Thanks Terry. Apologies for the confusion over the strings (KGOs) in the tests. Apart from that, looking at the missed lines made me realise that we still seem to have an implicit assumption in the code about the form of the `result` array that we support. We can either generalise this too or we can make the assumption explicit by checking for it in `validate`. The good news is that the integration tests were all fine :-)

Addressed in 4629275d73a0373c1f3f3f2ced3aadb35f198d49 and 703c3e170f6a376c6f445f8ad69f0d9553ea1718 which now has a different permutation of the `result` matrix in the calculation.

examples/psyad/lfric/tangent_linear_tweaked/tl_kinetic_energy_gradient_kernel_mod_tweaked.F90
```diff
@@ -201,8 +201,8 @@ subroutine tl_kinetic_energy_gradient_code(nlayers,       &
         end do
 
         ! Calculation
-        mul1(3) = matmul( jac(:,:,qp1,qp2), ls_u_at_quad )
-        mul2(3) = matmul( jac(:,:,qp1,qp2), u_at_quad )
+        mul1 = matmul( jac(:,:,qp1,qp2), ls_u_at_quad )
```
arporter, line 204: Good spot. Presumably we don't compiler these kernels which is how this got through? Probably not something for this PR though.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -350,13 +349,22 @@ def validate(self, node, options=None):
...
 
         # Make sure the result has as many full range as needed
-        if result.children:
-            for idx, child in enumerate(result.children):
-                if isinstance(child, Range) and not result.is_full_range(idx):
-                    raise TransformationError(
-                        f"To use {self.name} on matmul, each range on "
-                        f"the result variable '{result.name}' must be a full "
-                        f"range but found {result.debug_string()}")
+        if len(result.symbol.shape) in [1, 2] and not result.children:
```
arporter, line 352: We now have this check three times (for 'result', 'matrix1' and 'matrix2'). Please could you refactor so we just have one utility routine that does this check - you can pass in the permitted number of ranges and the text for which the check is for.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -417,12 +424,12 @@ def _apply_matrix_vector(self, node):
                                              datatype=INTEGER_TYPE)
 
         # Create "result(i)"
-        result_dims = [Reference(i_loop_sym)]
-        if len(result.children) > 1:
-            # Add any additional dimensions (in case of an array slice)
-            for child in result.children[1:]:
-                result_dims.append(child.copy())
-        result_ref = ArrayReference.create(result_symbol, result_dims)
+        r_fr_order, r_nfr_order, r_nfr = self._get_full_range_split(result)
```
arporter, line 427: Great! Glad this was straightforward.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
```diff
@@ -627,9 +626,54 @@ def test_validate_matmat_with_slices_on_rhs(fortran_reader):
...
+        "  real, dimension(6,3) :: jac_inv\n"
+        "  real, dimension(10,10,10) :: result\n"
+        "  result(:,:,:) = matmul(jac(:,:), jac_inv(:,:))\n"
+        "end subroutine my_sub\n")
+    trans = Matmul2CodeTrans()
+    assign = psyir.walk(Assignment)[0]
+    with pytest.raises(TransformationError) as excinfo:
+        trans.validate(assign.rhs)
+    assert ("Transformation Error: To use Matmul2CodeTrans on matmul, "
+            "no more than two indices of the argument 'result(:,:,:)' "
```
arporter, line 653: s/argument/result variable/ perhaps?

arporter (changes requested):
Thanks Terry. Looking very good now.
I think the only thing left is to refactor the validation as we have quite a bit of duplication there now (the part that checks the number of ranges).
Once that's done this I'll proceed to merge!

arporter:
(Just a note to say that the test file now gives 100% coverage of the transformation file :-) )

examples/psyad/lfric/tangent_linear_tweaked/tl_kinetic_energy_gradient_kernel_mod_tweaked.F90
DrTVockerodtMO: This is what I think yes.

src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
DrTVockerodtMO: Addressed in 6fb1e5a1ffda53c4a956266e070b3a161dda58d0. The text was fairly simple to homogenise so I opted not to pass it through.

src/psyclone/tests/psyir/transformations/intrinsics/matmul2code_trans_test.py
DrTVockerodtMO: Addressed in 6fb1e5a1ffda53c4a956266e070b3a161dda58d0, I just call them all arrays now in that part of the validation.

arporter (approved):
All good now - thanks for removing the duplication.
There are no unaccounted-for coverage changes.
CI is green. Integration tests were previously green.
Will proceed to merge.

PR #2800, opened 2024-11-23
Update CI software versions

nmnobre:


sergisiso (approved):
Thanks for updating all outdated PGI compiler comments @nmnobre . This PR updates the compilation workflow to the lastest version of installed modules (The nemo4 workflow is still using older modules due to an issue already tracked somewhere else). All tests are green.

Issue #2525, opened 2024-03-22
InlineTrans (and PSyIR) does not support routine arguments with the `optional` attribute.


arporter:
InlineTrans will refuse to countenance a routine if it has optional arguments since they end up as DataSymbols of `UnknownFortranType`. This is an issue for the NEMO sea-ice code.
We could probably add support for the optional attribute @sergisiso?

schreiberx:
This pull request (currently under review by Andy)
https://github.com/stfc/PSyclone/pull/2775
will provide the subroutine
`get_argument_routine_match(...)`
which returns a list of the indices of the matching arguments from a call to a subroutine with support for optional as well as named arguments.
Once this is accepted, updating the inlining with respect should be easy with this list.

hiker:
@arporter , can this be closed now?

arporter:
I think so. The fact that PSyIR doesn't directly support the `optional` attribute can be tackled separately if anyone cares :-)

Issue #2668, opened 2024-07-22
Transformation options questions


sergisiso:
When we validate the transformation options argument we raise TransformationErrors (as with any other errors that happen in the validation). But in several cases we let TransformationError be ignored in the caller, e.g.:

```
try:
    mytrans.apply(node, options={"my_option": True})
except TransformationError:
    continue
```
However, if "my_option" was expecting an "integer" and not a "bool" this is also ignored in the code above, this seems like something that should be reported regardless. @arporter @hiker Should we make option validation produce TypeError/ValueError instead?

I was also trying to remember why we use the options dict. I think we didn't use arguments because we wanted to be able to pass pre-populated objects around. But did we consider kwargs and dictionary unpacking? For instance:
```
mytrans.apply(node, my_option=3, my_option2=True)
mytrans.apply(node, **options)
```
would allow for both formats.

hiker:
I agree with TypeError/ValueError.

Re dict: yes, the idea was to support 'meta transformations', i.e. a transformation that calls several other transformations, and might need to set options everywhere. At some stage I wanted to replace this dict with an object (so we can detect if an option is provided that is not supported anywhere, which could indicate a typo).

It's more or less the same as `**kwargs`, I preferred a 'normal' dict, sometimes you need the `**` operators, which I find confusing :) But I am happy either way (though, unless there is a real advantage, I wouldn't want to spend time on a refactor ;) )

sergisiso:
The advantage I was thinking is in the more strightforward docstring and that IDEs may show/autocomplete the kwargs available. (although I may not sure if they can chase them down when the kwargs is just passed). But I agree it's not a pressing issue.

I will start using the TypeError/ValueError on these from now on.

schreiberx:
**kwargs don't trigger the error at the place that caused it.
The apply() function is called, but the arguments are then forwarded to the validate function where the error would be triggered. Is this really what should happen?

arporter:
I'm comfortable with having `validate` raise the error - I don't think it causes any confusion.

schreiberx:
I realized this in a PR as a dedicated `set_option` call:
https://github.com/stfc/PSyclone/blob/martin_2_issue_inline_optional_args_2525/src/psyclone/psyir/transformations/inline_trans.py#L164

This provides a clear **separation** between **setting the options and applying the transformation**.
These options are then **stored locally as member variables** of the class.

You might wonder why and the answer is quite simple:
[...]

PR #2801, opened 2024-11-23
Support optional and named arguments in inlining

schreiberx:
This PR would solve issue #2525


schreiberx:
This PR depends on PR #2775

schreiberx:
@arporter @sergisiso @hiker I thought it would be good to have you all informed about this PR because of some restructuring of the callee detection and I guess this is getting more and more important for other use cases of psyclone.

We have the following two different situations:
- We like to create a call trace. However, we don't want to use the inline transformation for a call trace. What we would like to have is to find the callee.
- If we like to inline, we need to find the callee as well.
Since finding the matching callee for interfaced calls/subroutines is part of these two situations, and since adding the feature of supporting optional and named arguments added quite a lot of lines of codes, I put the functionality to find the matching calls/subroutine into the file in
`psyir/tools/call_routine_matcher.py`
To do the coverage tests in the Andy style (Each test has to cover 100% of the corresponding file), I moved over various test cases which have been formerly part of the `call_tests.py` and `inline_trans_tests.py` to the new test file `call_routine_matcher_tests.py` .
All three test files now result in 100% coverage of the respective python file.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,47 +134,196 @@ class InlineTrans(Transformation):
...
+
+        self._option_check_diff_container_clashes: bool = True
+        self._option_check_diff_container_clashes_unres_types: bool = True
+        self._option_check_resolve_imports: bool = True
+        self._option_check_static_interface: bool = True
+        self._option_check_array_type: bool = True
+        self._option_check_unsupported_type: bool = True
+        self._option_check_unresolved_symbols: bool = True
+
+    def set_option(
```
arporter, line 164: 1. This list of options is massive - can it be refined/reduced?
2. As mentioned elsewhere, we haven't yet agreed on how to handle options to transformations (#2668) so please stick with the current method for now.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,47 +134,196 @@ class InlineTrans(Transformation):
         Some of these restrictions will be lifted by #924.
 
     '''
-    def apply(self, node, options=None):
-        '''
+
+    def __init__(self):
+        # List of call-to-subroutine argument indices
+        self._ret_arg_match_list: List[int] = None
```
arporter, line 140: Why is this Transformation class now holding state? I wouldn't expect this to be necessary...

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,47 +134,196 @@ class InlineTrans(Transformation):
...
+    ):
+        """
         Takes the body of the routine that is the target of the supplied
         call and replaces the call with it.
 
-        :param node: target PSyIR node.
-        :type node: :py:class:`psyclone.psyir.nodes.Routine`
+        :param call_node: target PSyIR node.
+        :type call_node: :py:class:`psyclone.psyir.nodes.Call`
+        :param routine: PSyIR subroutine to be inlined.
```
arporter, line 262: What's the reason for adding this argument?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,47 +134,196 @@ class InlineTrans(Transformation):
...
         :type options: Optional[Dict[str, Any]]
         :param bool options["force"]: whether or not to permit the inlining
             of Routines containing CodeBlocks. Default is False.
-        '''
-        self.validate(node, options)
+        """
+
+        # Validate that the inlining can also be accomplish.
+        # This routine will also update
+        # self.node_routine and self._ret_arg_match_list
```
arporter, line 273: Ah, so that's why there's state? Please update the comments on the declarations of these member variables.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,47 +134,196 @@ class InlineTrans(Transformation):
...
+        # with the routine to be inlined and the relation between the
+        # arguments and to which routine arguments they are matched to.
```
arporter, line 275: "...and which routine arguments..."

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,47 +134,196 @@ class InlineTrans(Transformation):
...
+        #   - Replace with `True` or `False`, depending on whether
+        #     it's provided or not.
+        self._optional_arg_resolve_present_intrinsics()
+
+        # Step 2)
+        # - For all If-Statements, handle constant conditions:
+        #   - `True`: Replace If-Block with If-Body
+        #   - `False`: Replace If-Block with Else-Body. If it doesn't exist
+        #     just delete the if statement.
+        self._optional_arg_eliminate_ifblock_if_const_condition()
```
arporter, line 308: Nice. However, I think this should be a separate transformation. That way, it can be used after @hbrunie 's new `ReplaceReferenceByLiteralTrans` (#2828) to eliminate branching. For preference I think this should be done as a separate PR to avoid us ending up with a massive review again.

arporter (changes requested):
Thanks very much Martin, this is exciting functionality. However, there are a couple of biggish issues that I want to tackle before I go any further with the review. First, you have implemented a different way of handling options for InlineTrans. We need project-wide agreement on how we do this so that we can adopt a uniform approach (#2668 ). We should use #2668 to come to some agreement and then you can implement that here. Second, you have cool new functionality to remove branching when the condition is a literal and I think that should be broken out as a separate transformation with an associated PR.
Finally, there are some files with non-functional changes that you can just revert to master to shrink this PR.

doc/Makefile
```diff
@@ -0,0 +1,18 @@
+
+all: developer_guide reference_guide user_guide
+.PHONY: developer_guide reference_guide user_guide
+
+developer_guide:
+	make -C developer_guide html SPHINXOPTS="-W --keep-going"
+
+reference_guide:
+	make -C reference_guide html SPHINXOPTS="-W --keep-going"
```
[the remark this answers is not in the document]
schreiberx: I'm working on the proposed changes quite late. There have been huge updates on this Makefile. I'll remove my suggested changes since the new version from `master` is much better.

doc/Makefile
```diff
@@ -0,0 +1,18 @@
+
```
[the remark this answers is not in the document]
schreiberx: I'm working on the proposed changes quite late. There have been huge updates on this Makefile. I'll remove my suggested changes since the new version from master is much better.

src/psyclone/tests/psyir/transformations/omp_task_transformations_test.py
```diff
@@ -178,7 +178,8 @@ def test_omptask_apply_kern(fortran_reader, fortran_writer):
     new_container.addchild(my_test)
     sym = my_test.symbol_table.lookup("test_kernel")
     sym.interface.container_symbol._reference = test_kernel_mod
-    trans = OMPTaskTrans()
+    trans: OMPTaskTrans = OMPTaskTrans()
+    trans.set_option(check_matching_arguments_of_callee=False)
```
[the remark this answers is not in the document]
schreiberx: OK

src/psyclone/psyir/transformations/inline_trans.py
schreiberx: OK. I'll open first an issue and then we'll see. Not sure if I can do this in the near future.

src/psyclone/psyir/nodes/__init__.py
```diff
@@ -195,5 +195,5 @@
         'OMPScheduleClause',
         'OMPFirstprivateClause',
         'OMPSharedClause',
-        'OMPDependClause'
-        ]
+        'OMPDependClause',
```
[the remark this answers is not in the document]
schreiberx: OK.

src/psyclone/psyir/symbols/containersymbol.py
```diff
@@ -135,6 +135,10 @@ def find_container_psyir(self, local_node=None):
                            the container.
         :type local_node: Optional[:py:class:`psyclone.psyir.nodes.Node`]
 
+        :param ignore_missing_modules: If 'True', no ModuleNotFound exception=
```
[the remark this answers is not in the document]
schreiberx: OK.

schreiberx:
@arporter I'll close this one and reopen another one.

PR #2803, opened 2024-11-26
2796 fix profile nodes inside of lfric loops

### LINKED ISSUE #2796: [LFRic] bug when profiling applied within a Loop
Jaffery has found that applying ProfileTrans to nodes within a loop generates incorrect code. The `profile_psydatatype` is imported correctly but no variable with that type is ever declared. If I alter the script to make sure the profiling is put *outside* the loop then the associated variable is declared correctly. I suspect this is because we currently handle everything within a loop with PSyIR and everything outside with the old `gen_code` method.

If it isn't too hard to fix then it would be good to get this done before 3.0 is released.

hiker:
Fixes #2796.


hiker:
My apologies, I didn't see the branch you had created that had a reproducer.  The patch is small, and a little bit hacky (I didn't go through the effort of adding a fix_gen_code to node as etc, so we could recurse over all nodes ), but given that this seems to be the only use case, and it will go away soon anyway, I think that's ok.

Unfortunately, the CI failed with both nemo tests: 
nemo_v5:
https://github.com/stfc/PSyclone-mirror/actions/runs/12022570624/job/33515010510#logs
is even caused by profile_mod not being found ... I can't see how this is related, after all, this is lfric only code?? 

nemo_v4 (https://github.com/stfc/PSyclone-mirror/actions/runs/12022570632/job/33515010575) ... not sure, somehow 'it 10' not found??

arporter:
> Unfortunately, the CI failed with both nemo tests: nemo_v5: https://github.com/stfc/PSyclone-mirror/actions/runs/12022570624/job/33515010510#logs is even caused by profile_mod not being found ... I can't see how this is related, after all, this is lfric only code??
> 
> nemo_v4 (https://github.com/stfc/PSyclone-mirror/actions/runs/12022570632/job/33515010575) ... not sure, somehow 'it 10' not found??

Both of those are expected and are very shortly to be fixed by #2697.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -576,5 +577,8 @@ def gen_code(self, parent):
         # - Add the kernel module import statements
         for kernel in self.walk(CodedKern):
             if not kernel.module_inline:
-                parent.add(UseGen(parent, name=kernel._module_name, only=True,
-                                  funcnames=[kernel._name]))
+                parent.add(UseGen(parent, name=kernel.module_name, only=True,
+                                  funcnames=[kernel.name]))
+
+        for psydata_node in self.walk(PSyDataNode):
```
arporter, line 583: I think this fix should be in `LFRicLoop.gen_code()`. It's LFRic-specific so belongs there. (Presumably GOcean doesn't have this bug but it would be worth adding a test for this.)

src/psyclone/psyir/nodes/psy_data_node.py
```diff
@@ -694,6 +694,26 @@ def gen_code(self, parent, options=None):
 
         self._add_call("PostEnd", parent)
 
+    def fix_gen_code(self, parent):
+        '''This function might be called from loop.gen_code if a PSyData node
```
arporter, line 698: This will need updating if you can move your fix to LFRicLoop.

src/psyclone/tests/psyir/nodes/psy_data_node_test.py
```diff
@@ -610,3 +610,28 @@ def test_psy_data_node_name_clash(fortran_writer):
             'module_with_name_clash_mod", f2_data_1)' in code)
     assert ('CALL extract_psy_data % ProvideVariable("f2_data@'
             'module_with_name_clash_mod_post", f2_data_1)' in code)
+
+
+# ----------------------------------------------------------------------------
+def test_psy_data_node_lfric_inside_of_loop():
+    '''Test that if a PSyData node is inside a loop (which means the code will
+    already be generated by PSyIR), the required psydata variable is declared.
+    ATM (TODO #1010) Loop.gen_code calls a fix_gen_code function in the
```
arporter, line 619: Hopefully `LFRicLoop.gen_code()`.

arporter (changes requested):
Thanks for fixing this Joerg. I think the fix itself is fine but should be in `LFRicLoop.gen_code()` rather than `Loop.gen_code()`. (Assuming that this bug doesn't manifest for GOcean - I've requested a test for that.)

src/psyclone/psyir/nodes/loop.py
hiker: OK, but I have only moved the code that I have added, not the already existing, LFRic-specific code (I believe) that's just before that section.

src/psyclone/psyir/nodes/psy_data_node.py
hiker: Done.

arporter:
Sorry Joerg, I've just brought this branch up to master but then realised you might not have finished your changes. Apologies...

src/psyclone/tests/psyir/nodes/psy_data_node_test.py
hiker: Updated.

src/psyclone/psyir/nodes/loop.py
hiker: I've also added a test with gocean, and used a regex (in both tests) to actually make sure the psydata calls are in the expected place ;)

hiker:
Ah, that's why my git pull reported on a merge, which confused me a bit, but decided to trust in git :)

Now it should be ready.

src/psyclone/psyir/nodes/loop.py
arporter: Whoop!

arporter:
Changes all good now, thanks Joerg. I've brought the branch up to master so hopefully all of the integration tests will pass now - I've set these running.

arporter (approved):
Integration tests were all fine. The GOcean-related web site appears to have been fixed so all of the link checks should hopefully be green this time.
Will merge once the CI reports back.

PR #2808, opened 2024-11-28
Avoid Singleton of module manager

schreiberx:
Does what's in the title.

- ModuleInfo
    - Can be created with `None` module_manager.
    - As soon as it's registered with a ModuleManager, the module manager updates the information in ModuleInfo about this.
    - ModuleInfo can then also update the psyir root tree.

- Some test cases don't use Containers as the root node
  => ModuleManager needs to be attached to "Node"
    - Tests `reset` module manager by using ..._instance = None.
    - => Now moved into `ModuleManager._test_helper_reset()`` which also resets the newly introduced usage counter
- Different ways to reset ModuleManager in tests
    - Fixture with monkeypatch in `psyclone.tests.psyir.conftest` :: `modmanager_fixture`
    - AND in `psyclone.tests.conftest` :: `clear_module_manager_instance`
    `clear_module_manager_instance` is now removed
- *Everytime some code is created*, module_manager has/should be provided.
    - `get_invoke()`: New module_manager argument
    - Also at various other places
- *Symbol lookup* is relative to some module/container.
  => requires providing a `Node` to know the context.


schreiberx:
Maybe there's some misunderstanding: I'm seeing this PR not as something ready to be merged but as a good discussion basis on how a ready PR could look like (including running through all CI tests).

I created this PR mainly to have a very concrete discussion based on
- things that would need to change
- issues with the current design
- impact on other parts (e.g., examples).
You can now clearly see this from the difference.

hiker:
While I still need an update from the meeting, I believe there are two things or three being mixed together (maybe because of me misunderstanding or misinterpreting something??) that imho have nothing to do with each other, which is what confuses me.

1. Your application requires to read all files, and act on a subset of the files, or you want to do this in two (or more??) stages - first reading all files, storing state somewhere on the file system, then in a second phase read only a subset of the files. 
    As far as I can see, this will just work fine with the module manager the way it is. Accesses are using a `dict`, so it doesn't really matter if you have all files or only a subset. I can't see anything existing in the code to support this usage pattern.

2. You want to explicitly list the files (instead of just pointing to directories). That should be a small change (check if what is currently a search path in the module manager is a file, and if so, only handle this one single file). 
     While I still like the convenience of just specifying the root path, we also have an edge case: dl_esm_inf has a MPI and non-MPI version of a module, and it's basically the build system that picks one version or another. So, in PSyclone we need to have a way to specify exactly which version to use. This is easy to support. I'd suggest we support a more general form: recursive directory, non-recursive directory, file, and an exclude list. I think that should cover all cases.

3. That then leaves the other, and imho unrelated issue: you want to refactor the code to avoid a Singleton. My understanding was that your use case would not work with the current implementation, but that might have been my misunderstanding, which confused me quite a bit. I'll add more about singleton later

schreiberx:
> While I still need an update from the meeting, I believe there are two things or three being mixed together (maybe because of me misunderstanding or misinterpreting something??) that imho have nothing to do with each other, which is what confuses me.
> 
>     1. Your application requires to read all files, and act on a subset of the files, or you want to do this in two (or more??) stages - first reading all files, storing state somewhere on the file system, then in a second phase read only a subset of the files.
>        As far as I can see, this will just work fine with the module manager the way it is. Accesses are using a `dict`, so it doesn't really matter if you have all files or only a subset. I can't see anything existing in the code to support this usage pattern.
> 
>     2. You want to explicitly list the files (instead of just pointing to directories). That should be a small change (check if what is currently a search path in the module manager is a file, and if so, only handle this one single file).
>        While I still like the convenience of just specifying the root path, we also have an edge case: dl_esm_inf has a MPI and non-MPI version of a module, and it's basically the build system that picks one version or another. So, in PSyclone we need to have a way to specify exactly which version to use. This is easy to support. I'd suggest we support a more general form: recursive directory, non-recursive directory, file, and an exclude list. I think that should cover all cases.
> 
>     3. That then leaves the other, and imho unrelated issue: you want to refactor the code to avoid a Singleton. My understanding was that your use case would not work with the current implementation, but that might have been my misunderstanding, which confused me quite a bit. I'll add more about singleton later

[For others reading this discussion: We had a video call on this]

Misunderstandings/interpretations are common with that complexity, and I need to familiarize myself with what you discussed in the last couple of years.

1. Yes, that’s what I had in mind by saying that I can develop a workaround if you still want it to be a singleton.

2. Agreed. That’s what I meant by mentioning that I changed my suggestion to allow different module managers. We agree that the existing one can be used for that, but it may be cleaned up slightly and enriched with certain features. But what should not happen IMHO is to have features very specific to just one development, which is, by default, active for all use cases. There should be an option to activate it (e.g., the recursive search).

3. As pointed out, my use case would work with the workaround from (1).

To summarize the main points from our meeting, why we should avoid the Singleton:

- This allows multiple ModuleManagers to be used in one Python execution. E.g., we could do individual optimizations on the barotropic and then baroclinic solver of NEMO/Croco (future plan).
A workaround with a singleton is that the ModuleManager (and hence Psyclone) can only be used sequentially and needs to be reset/reinitialized. That’s all not elegant, but I could live with it as a workaround.

- I also like to pick up Sergi’s argument that psyir is used as a library by other developments, which is what we do in the Poseidon project. Here, you never know what others want to do. If the requirement to avoid the Singleton shows up in a few years without a possible workaround, it’s getting harder to split it.

- **my main argument** is that avoiding singletons **ALWAYS** leads to a clearer software design. I had various experiences in avoiding these singletons, and every time, rewriting code to avoid this revealed the real dependencies and made the overall software architecture far clearer.

To give two examples of what showed up during the process to avoid the Singleton:
- From our meeting, someone claimed that the ModuleManager shouldn’t be used everywhere, but indeed, it’s already used in the `Config` itself by writing data into the Singleton of the ModuleManager. From the PR, you also see various other places. As a user of the psyclone module, you basically never know where it could be used. You must always trace down the call stack to see whether it shows up somewhere.
- It becomes much more apparent from the software design where symbols are looked up.
E.g., `self._interface.get_container` now has module_manager as an argument (see PR), making it clear that it also looks up information further up the module manager.
- Another example: To find a container in which a symbol exists, you must now provide a local_node argument: `csym.find_container_psyir(local_node=local_node)`. Without it, you’re not aware of the context of this existing symbol.

To summarize my main point (which you didn't mention):
**These are all consequences of avoiding the module manager being a singleton**, enforcing a **clearer software design**, or showing design problems as it did in my previous examples.

This PR is just the first step of many steps in this direction. If you're open to this PR, I could also clean it up (e.g., undoing renamings of some arguments which would reduce the number of changed lines).

Issue #2737, opened 2024-10-07
Release 3.0


arporter:
As discussed, we need to make a 3.0 release. Mainly to facilitate the new users of the 'code transformation' functionality and to emphasise that the API has changed since the 2.5.0 release.

 - [x] https://github.com/stfc/fparser/pull/451
 - [x] https://github.com/stfc/fparser/pull/450
 - [ ] #2701
 - [x] FFSL support is required (#2661)
 - [x] Document the OpenMP offload functionality (under https://psyclone.readthedocs.io/en/latest/transformations.html#openmp) and add/extend an LFRic example to show it.
 - [x] #2739 
 - [x] #2742 (PSyAD test-harness reproducibility)
 - [x] #2755 
 - [x] #2697 Would be nice to have as it also fixes NEMO integration tests.
 - [ ] Update version in Spack (and update Wiki to remind us to do this).
 - [x] #2766 Change index order for LFRic operators.
 - [x] #2795 Allow Fortran matmuls to take matrices with non-contiguous ranges.

We want this before SC24.

hiker:
- [ ] #2694 
- [ ] #257

I might also consider #2483, that might be useful for work with the UM, but I need to check where this PR is exactly at

LonelyCat124:
@arporter Should we rename the "nemo" tutorials to Code transformation before this release as well?

arporter:
> @arporter Should we rename the "nemo" tutorials to Code transformation before this release as well?

Ideally, yes.

LonelyCat124:
> > @arporter Should we rename the "nemo" tutorials to Code transformation before this release as well?
> 
> Ideally, yes.

I'll see if I can do this then

hiker:
> > > @arporter Should we rename the "nemo" tutorials to Code transformation before this release as well?
> > 
> > 
> > Ideally, yes.
> 
> I'll see if I can do this then

#2759 (which I assume should go into 3.0 as well) addresses a similar issue with library files. And there are the examples as well.

We should agree on a name - maybe just `transformation`? Or `no_psykal`?

sergisiso:
For the folder names: `generic` or `std_fortran` are other options.

arporter:
Thanks @LonelyCat124 and @hiker. I like "generic" but am happy to go with the consensus.

LonelyCat124:
I had a chat with Sergi about the tutorials and we decided to leave that until we have time to redo them entirely, as they're very out of date.

[...]

PR #2809, opened 2024-11-29
(towards #2737) update OMP section of docs

arporter:
I took a look at the existing OMP documentation and decided that, rather than go into great detail, it would be best just to point people to the existing examples. Therefore this is a small PR that does just that.


arporter:
Should be good to look at now @TeranIvy (or @mo-lottieturner or @oakleybrunt).

oakleybrunt (approved):
Simple change, fine with me.

Issue #2724, opened 2024-09-30
Automatically update version number by appending git hash


arporter:
At the telco today, we agreed that the PSyclone version number (stored in `src/psyclone/version.py` and also in `doc/reference_guide/doxygen.config`) should have the git hash of the last commit (to master) appended to it. This will enable people to see when they are working with a version checked-out of GitHub, as opposed to a released version.

We could do this manually as part of the end-of-the-review process but it would be much better if we could automate it using a GitHub Action.

hiker:
I've found https://stackoverflow.com/questions/3442874/in-git-how-can-i-write-the-current-commit-hash-to-a-file-in-the-same-commit. In short, we can't add the hash before writing the file (since adding the hash will change the hash). One solution mentioned was to use the previous hash. Or we could use the post-checkout hook? Not sure how this works with updates, merge, cherry-pick etc.
git attributes are mentioned, but they apparently need to be installed each time manually(?)

I am getting the feeling that just adding "-alpha" to the version number after a release might be the easiest option (and would work for me, all I need is to distinguish current master from the last release). It would also follow the semantic versioning standard (https://semver.org/).

One other idea: `version.py` tries to run and interpret `git branch` (we can't use `git branch --show-current`, it provides no output in case of a tag or hash based checkout). Possible outcomes (not sure if this list is really complete)
- `(HEAD detached at 2.5.0)`  --> checked out tags/2.5.0 - identified by a regular expression of integers and dots
- `(HEAD detached at 2f4162581)`  --> checked out a hash - identified by not a version number as above, and a valid hex number
- Any other non-error: we are on a branch - use the existing version number and append `-` with branch-name (2.5.0-257_verify_loop_fusion)
- Error - not a git checkout, just use the existing version number.

arporter:
Thanks for thoroughly investigating Joerg. I like the idea of sticking to SemVer but I'm worried we're going to have e.g. 'alpha' versions of releases that will never actually exist. e.g. at the moment we should probably be at 3.0-alpha which is fine as we will definitely release 3.0. As soon as that is out of the door, we bump the 'working' version to 3.1-alpha. Say we then discover a bug in 3.0 that requires an immediate fix so we need to release 3.0.1. Do we backport the bug fix to 3.0 and then release that or does it go into 3.1? I guess the answer will depend on how urgently the bug fix is needed?

hiker:
[...]

PR #2818, opened 2024-12-06
#2724 Updated version number to be 3.0.1-dev.

hiker:
This doesn't fully fix 2724. MY understanding is that we first use `-dev`, then wait for feedback from Iva if we can switch to the automated way using setuptools_scm (which caused problems on some UK systems I Believe).

So, this only adds `-dev` for now

### DIFF
--- a/src/psyclone/version.py
+++ b/src/psyclone/version.py
@@ -33,6 +33,7 @@
 # -----------------------------------------------------------------------------
 # Author: A. R. Porter, STFC Daresbury Laboratory
 # Modified by R. W. Ford and N. Nobre, STFC Daresbury Lab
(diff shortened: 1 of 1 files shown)


hiker:
As discussed, update the version number to 3.0.1-dev (while we investigate if we can automate this). Note that both Sergi and I agreed that there is no need to call it `3.0.1-develop` in order to be supported by spack, 'develop' is just the spack name that can be given to any branch.

sergisiso (approved):
@hiker This introduces the "-dev" change that we agreed and has passed all CI tests.

@hiker @arporter breaking convention I will merge without the changelog update (as this seems a repo managing change (not functionality) and it will be reversed anyway for the next release. But let me know if you disagree  and will add it later.

Issue #2577, opened 2024-05-08
Add pointer/target/pointer_assignment support


sergisiso:
Due to popular demand, I think we should reconsider adding support for pointers. I have usually been reticent because they add many complications to properly validating transformations and do correct dependency analysis and I was hoping that the CodeBlocks and UnsupportedTypes provided some barrier/protection for these cases. But in practice:
- We often forget validate/test about these cases. And with the increase in number of processed applications we need to improve the correctness for the general case.
- It's the ugliest part when programmatically creating PSyIR trees (which is a common use case) because we need to import fparser, call it and put the result in the CodeBlock/Type.

The assignment is nothing special, it could be an attribute to Assignment:
`Assignment.create(lhs=expr, rhs=expr, is_pointer=True)`

This will remove the CodeBlocks and defer this specific validity checks to analysing the symbols which we need anyway (fewer special cases).

The problem with pointer and target attributes is that is possible to alias memory. This is also true for UnsupportedType and UnresolvedType (and some interfaces?) - we need to assume the worst case.
```
real, dimension(:), target :: a
real, dimension(:), pointer :: b => a
```

But it would be cumbersome to add all these type checks all over the place. So, we also need a new property in Symbol (or Type?) that does all this checking:
```
def is_potentially_aliased():
    return symbol.is_pointer or symbol.is_target or symbol.is_unsupported or symbol.is_unresolved
```
and I think it will also make it less likely to forget because the name is more relevant to the problem.

Then we need to modify the DependencyTools, reference_accesses, forward_reference with this information (which now we get wrong it?).

In:
```
do i = 1, 10
    a(i) = b(i-1)
end do
```
the `is_loop_independent` should return false (also if a and be are simply imported symbols)

In:
```
a = 3
c = b
```
the `stmt[0].lhs.next_access` / `reference_access` should return `stmt[1].rhs`

This is what we need when deciding if we need a data_movement directive, a halo_exchange, a omp variable not being private, ...
(if there use cases for just caring about the symbol, and not dependencies, we could add an attribute or similar alternative properties but I am usure there are)

This will impact lfric, where all fields are pointers (but we still not use the general psyir transformations and when we do it, we can specialise the Proxys and other symbols to `def is_potentially_aliased(): return False` if that is guaranteed by LFRic).

[...]

PR #2822, opened 2024-12-09
(Towards #2577) Add `pointer` and `target` attributes in datatypes, frontend and backend

JulienRemy:
This starts implementing point 2 of Sergi's list in #2577 by adding new attributes to datatypes, as well as frontend and backend support for them.

Note that `is_potentially_aliased` is not used anywhere, except to fix an existing `Matmul2CodeTrans` test.
So this *does not check for aliasing anywhere else for now*. Stated differently, this might be dangerous to merge as it effectively supports pointers in PSyIR but does not yet restrict transformations to non-aliased variables.


src/psyclone/psyir/transformations/intrinsics/matmul2code_trans.py
```diff
@@ -331,6 +333,15 @@ def validate(self, node, options=None):
...
                 f"'{matrix1.symbol}' and '{matrix2.symbol}'.")
 
+        # The arrays must not be potentially aliased
+        if any(var.symbol.is_potentially_aliased for var in
+               [matrix1, matrix2, result]):
+            raise TransformationError(
+                f"Expected result and operands of MATMUL IntrinsicCall to "
+                f"be references to arrays that are not potentially aliased "
+                f"but found '{result.symbol}', '{matrix1.symbol}' and "
+                f"'{matrix2.symbol}'.")
```
arporter, line 343: probably worth adding "...symbol}' in '{node.debug_string()}'." to give more information.

arporter (changes requested):
Very nicely done, thanks Julien. Implementation is good, tests good (just need a bit more coverage). As you say in a previous comment, I am concerned about putting this onto master as is as it opens the door to potential problems. Therefore, I think we do have to tackle making sure that transformations that care about this do the right thing (at this stage, that can simply be refuse to do anything). I suggest going through the list in the Reference Guide:
image
and checking whether or not they should care about aliasing.

sergisiso (changes requested):
I see @arporter already reviewed it, so I will let him continue being the reviewer for this, but I had some comments already stored in github so I will push them anyway (even though some are the same as @arporter ). The reason I never finished this review is that it was unclear to me if the "is a pointer" attribute had to be part of the symbol itself or the datatype of the symbol. And I think it would be good to discuss this before we commit to a solution.

PR #2827, opened 2024-12-16
Update Zenodo reference

sergisiso:
I made a version artifact for PSyclone 3.0 in our Zenodo DOI (https://zenodo.org/records/14500240), but it has given it a new number/reference.

There is also a DOI that cite all versions of PSyclone and will always resolve to the latest one, so I updated the README to it.

@arporter This is a small one, I didn't create a related issue, let me know if you prefer it. This is ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
   <img width="460" src="doc/logo/psyclone_v1.0.png">
 </p>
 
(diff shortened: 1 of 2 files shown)


arporter (approved):
Thanks for doing this Sergi. Looks good. I've grep'd and the README is the only place we refer to Zenodo. I've also checked the link and it's good.
This only affects the README so no need to do integration tests.
Will proceed to merge.

PR #2828, opened 2024-12-17
Replace ref by literal

hbrunie:
Replacing PsyIR Reference by PsyIR Literal when it is easy to do so.
Same file, same subroutine, or subroutine contained in module.
Build a param_table with symbol name and Literal, then go through psyir references and symbol_table array type shapes to look for ref with name contained in param_table. 

Issues: 
 * it seems that initial value will always be a literal, is this expected behavior? If so I can remove a test raise error not triggered yet.
 * I might have found a bug: `character(len=4), parameter :: x = "toto" is not parsed a a is_constant.


arporter:
> * it seems that initial value will always be a literal, is this expected behavior? If so I can remove a test raise error not triggered yet.

Thanks Hugo. I'm pretty sure we allow any PSyIR expression as an initial value?

hbrunie:
> > * it seems that initial value will always be a literal, is this expected behavior? If so I can remove a test raise error not triggered yet.
> 
> Thanks Hugo. I'm pretty sure we allow any PSyIR expression as an initial value?

Thanks. I made a test showing the current limit of this MR with respect to non-literal initial value (for example a binary operation).

But I don't understand why 
`character (len=4), parameter :: x ="toto" ` does not make a constant symbol with initial_value being

arporter:
> But I don't understand why
> `character (len=4), parameter :: x ="toto" ` does not make a constant symbol with initial_value being

I think that's because we don't support the `len=xxx` specification on a character declaration.

hbrunie:
> > But I don't understand why
> > `character (len=4), parameter :: x ="toto" ` does not make a constant symbol with initial_value being
> 
> I think that's because we don't support the `len=xxx` specification on a character declaration.

That's ok. I don't deal with "complex" right hand side for now. If this is not a Literal, the parameter is just ignored.

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+
+                integer i, invariant, ic1, ic2, ic3
+                real, dimension(10) :: a
+                invariant = 1
+                do i = 1, 10
+                    t1%a = z
+                    a(ic1) = u1+(ic1+x)*ic1
+                    a(ic2) = u2+(ic2+y)*ic2
+                    a(ic3) = u3+(ic3+z)*ic3
+                    a(t1%a) = u4+(t1%a+u4*z)*t1%a
```
arporter, line 97: I'm sure it will be fine but please could you extend this test to cover more edge cases:
  * constant var in a loop bound, e.g. `do i = u1, u4, u1`
  * constant var as index in derived-type access, e.g. `t1%b(u4)`
  * constant var as argument to subroutine call, e.g. `call my_sub(ic1+u3)`

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+                end do
+                end program test"""
+    psyir = fortran_reader.psyir_from_source(source)
+    # The first child is the assignment to 'invariant'
```
arporter, line 101: Outdated comment?

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+    rrbl.apply(routine)
+    written_code = fortran_writer(routine)
+
+    assert "a(ic1) = 1 + (ic1 + 3) * ic1" in written_code
+    assert "a(ic2) = 2 + (ic2 + 12) * ic2" in written_code
+    assert "a(ic3) = 3 + (ic3 + 13) * ic3" in written_code
+    assert "a(t1%a) = 4 + (t1%a + 4 * 13) * t1%a" in written_code
+
+
+# ----------------------------------------------------------------------------
```
arporter, line 114: Please rm the duplicated separator line.

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+# ----------------------------------------------------------------------------
+def test_rrbl_module_defined_parameter(fortran_reader, fortran_writer):
+    source = """module test
```
arporter, line 117: Docstring please.

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+                real, dimension(size) :: a
+                    do i = 1, size
+                        if (y) then
+                            a(i) = 2+(i+u)
+                        endif
+                    end do
+                end subroutine
+                """
+    psyir = fortran_reader.psyir_from_source(source)
+    # The first child is the assignment to 'invariant'
```
arporter, line 168: Outdated comment?

arporter (changes requested):
Good stuff Hugo, thanks for this.
Mostly my comments are about docstrings :-)
As you hint in one of your comments, I'm wondering whether we should always consider all constant symbols that are in scope, rather than explicitly looking at the Routine and Container symbol tables?
I also think it would be better if we quietly skipped constant symbols with a non-Literal value rather than raise an error. That would enable the transformation to do as much as it can rather than aborting early.
Coverage is good and the new test file fully covers the new source file.
If you feel brave, it would be good to try the transformation 'for real' by adding it to the transformation script that we use for NEMO in the integration tests. You could add it immediately after the call to `enhance_tree_information`:
https://github.com/stfc/PSyclone/blob/9b8a0ba039b8eb46c60042b19a0bfa6950b15256/examples/nemo/scripts/omp_gpu_trans.py#L85-L94

hbrunie:
> Good stuff Hugo, thanks for this. Mostly my comments are about docstrings :-) As you hint in one of your comments, I'm wondering whether we should always consider all constant symbols that are in scope, rather than explicitly looking at the Routine and Container symbol tables? I also think it would be better if we quietly skipped constant symbols with a non-Literal value rather than raise an error. That would enable the transformation to do as much as it can rather than aborting early. Coverage is good and the new test file fully covers the new source file. If you feel brave, it would be good to try the transformation 'for real' by adding it to the transformation script that we use for NEMO in the integration tests. You could add it immediately after the call to `enhance_tree_information`:
> 
> https://github.com/stfc/PSyclone/blob/9b8a0ba039b8eb46c60042b19a0bfa6950b15256/examples/nemo/scripts/omp_gpu_trans.py#L85-L94

I am not sure Andrew about what you meant by : " should we always consider all constant symbols that are in scope, rather than explicitly looking at the Routine and Container symbol tables?"
Could you explain?

Besides, when you say "quietly skipping" does it mean that if the transformation finds a non-literal symbol it should not even throw a warning?
I thought that this could be handled by the transformation caller with exception catchers.
But I will do the change, and skip it quietly.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,188 @@
...
+    Routine,
+    Reference,
+    Literal,
+    Container,
+)
+from psyclone.psyir.transformations.transformation_error import (
+    TransformationError,
+)
+
+from typing import Dict, List, Union
```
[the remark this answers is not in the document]
hbrunie: I did it inside the same package in  'from PACKAGE import MODULE '.

And I also sorted the packages in from PACKAGE import MODULE

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+    foo: Routine = psyir.walk(Routine)[0]
+    assert foo.symbol_table is not None
+
+    rbbl = ReplaceReferenceByLiteralTrans()
+    error_str = ""
+    try:
+        rbbl.apply(foo)
+    except TransformationError as e:
+        error_str = e.__str__()
+    assert "initial value is not a Literal" in error_str
```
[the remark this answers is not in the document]
hbrunie: I have written comment in code in the form `! Psyclone(REplaceRefereceByLiteraltrans): message`

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,188 @@
...
+    # ------------------------------------------------------------------------
+    def apply(self, node: Routine, options=None):
+        self.validate(node, options)
+        param_table: Dict[str, Literal] = {}
+        if isinstance(node.parent, Container):
+            param_table = self._update_param_table(
+                param_table, node.parent.symbol_table
+            )
+
+        param_table = self._update_param_table(param_table, node.symbol_table)
```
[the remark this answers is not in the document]
hbrunie: Yes indeed I have written a small test for this
`test_rrbl_conflict_between_module_and_routine_scopes`
I decided to remove the symbol from `param_table` and so not to replace it if there is such conflict.
Also for now on I just go through one level, but this could be easily extended if needed as you mentioned.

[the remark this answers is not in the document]
hbrunie: Can I mark this as `resolved` @arporter ?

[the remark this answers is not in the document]
arporter: It's OK, Hugo - the reviewer does that (that way they can see what is still outstanding).

hbrunie:
@arporter I think I addressed all your concerns. Please let me know. Best, Hugo.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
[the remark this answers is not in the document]
arporter: Perfect, thanks.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,188 @@
...
+)
+
+from typing import Dict, List, Union
+
+
+class ReplaceReferenceByLiteralTrans(Transformation):
+    """Replace Reference by Literal if the corresponding symbol from
+    the symbol table is constant. That is to say this is a Fortran parameter.
+    For example:
+
```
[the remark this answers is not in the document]
arporter: Nice, thanks.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,294 @@
...
+        * Goes through all datasymbols in the symbol_table.
+            * if symbol already in param_table or initial_value is not Literal:
+                * annotate code with warning.
+            * copy and detach the symbol.initial_value (Literal)
+            * update the param_table with this copy.
+        * Returns the updated param_table
+
+        :param param_table: To be updated
+        :param symbol_table: scope symbol table to look for the symbols.
+        :return: the updated param_table (same reference as the entry one)
```
arporter, line 158: Blank line before :return: please.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,294 @@
...
+        :rtype: Dict[str, Literal]
```
arporter, line 159: No need for rtype as you've added a type hint :-)

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        for sym in symbol_table.datasymbols:
+            sym: DataSymbol
+            if sym.is_constant:
+                sym_name = sym.name
+                print(sym_name)
```
arporter, line 165: Please remove these prints (and at L186).

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -0,0 +1,188 @@
...
+    ) -> Dict[str, Literal]:
+        for sym in symbol_table.symbols:
+            sym: Symbol
+            if not isinstance(sym, DataSymbol):
+                continue
+            if sym.is_constant:
+                sym_name = sym.name
+                if param_table.get(sym_name):
+                    raise TransformationError(
+                        f"Symbol already found {sym_name}."
```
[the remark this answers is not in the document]
arporter: Thanks, that's better but remember, a user isn't going to know what 'param_table' means so how about "To avoid replacing it with the wrong value, we skip this symbol."?

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -182,101 +178,179 @@ def test_rrbl_array_shape(fortran_reader, fortran_writer):
...
+        end module
+        """
+    psyir = fortran_reader.psyir_from_source(source)
+    routine_foo: Routine = psyir.walk(Routine)[1]
+    rbbl = ReplaceReferenceByLiteralTrans()
+    rbbl.apply(routine_foo)
+    written_code = fortran_writer(routine_foo.ancestor(Container))
+    print(written_code)
+    assert (
+        "! Psyclone(ReplaceReferenceByLiteralTrans): Symbol already found a"
```
arporter, line 222: Please update the generation of this message to use single quotes around the variable name (so we get e.g. 'a' here).

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -182,101 +178,179 @@ def test_rrbl_array_shape(fortran_reader, fortran_writer):
...
+        integer, parameter :: y=a
+        end subroutine
+        end module
+        """
+    psyir = fortran_reader.psyir_from_source(source)
+    routine_foo: Routine = psyir.walk(Routine)[1]
+    rbbl = ReplaceReferenceByLiteralTrans()
+    rbbl.apply(routine_foo)
+    written_code = fortran_writer(routine_foo.ancestor(Container))
+    print(written_code)
```
arporter, line 220: Please rm the `print`

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -182,101 +178,179 @@ def test_rrbl_array_shape(fortran_reader, fortran_writer):
...
+        call foo(u2+ u3)
+    enddo
+    end subroutine
+    subroutine foo()
+        integer, dimension(10, a) :: array
+        end subroutine
+        end module
+        """
+    psyir = fortran_reader.psyir_from_source(source)
+    routine_foo: Routine = psyir.walk(Routine)[0]
```
arporter, line 251: Isn't this getting the 'tata' routine? If so, please rename the variable to avoid confusion.

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -182,101 +178,179 @@ def test_rrbl_array_shape(fortran_reader, fortran_writer):
...
+    rbbl.apply(routine_foo)
+    written_code = fortran_writer(routine_foo.ancestor(Container))
+    assert "integer, dimension(10,a) :: array" in written_code
+    assert "call foo(2 + 3)" in written_code
+
+
+def test_rrbl_same_constant_data_symbol_twice(fortran_reader, fortran_writer):
+    """test fortran code annotation with transformation warning and
+    value is coming from the closest scope.
+    * constant var as argument to subroutine call, e.g. call my_sub(ic1+u3)
```
arporter, line 262: Please rm this line. Also, you could say that we could extend to handle this situation in future.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -33,74 +33,159 @@
...
 
-    becomes
-    module toto
-    integer, parameter x= 3
-    integer, parameter z = 3*2
+    '''
 
-    """
+    _ERROR_MSG = (
+        "Psyclone(ReplaceReferenceByLiteral): only "
```
arporter, line 131: For consistency with other transformations, please change this to "ReplaceReferenceByLiteralTrans: only..."

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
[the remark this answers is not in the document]
arporter: That's a good idea, thanks.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -132,35 +225,52 @@ def _replace_bounds(
...
+        """Applies the transformation to a Routine node:
+        * First update a dictionary (param_table) with the Literal of constant
+        (parameter) symbol from node.parent symbol_table, and from
+        node.symbol_table.
+        * Second, use this updated param_table to replace reference in node
+        psyir_tree with the corresponsing Literal.
+        * Third, use this updated param_table to replace reference in node
+        symbol_table DataSymbol array's dimensions with the corresponsing
+        Literal.
+
```
arporter, line 237: Please add a note here to say that if an update cannot be performed for any reason then the substitution is skipped and a comment added to the generated code.

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -182,101 +178,179 @@ def test_rrbl_array_shape(fortran_reader, fortran_writer):
...
-        error_str = e.__str__()
-    assert "initial value is not a Literal" in error_str
+    rbbl.apply(routine_foo)
+    written_code = fortran_writer(routine_foo.ancestor(Container))
+    assert ReplaceReferenceByLiteralTrans._ERROR_MSG in written_code
+    assert "x = b" in written_code
 
 
-def test_raise_transformation_error_initial_value(
+def test_rrbl_raise_transformation_error_initial_value(
```
arporter, line 332: Please rename now that it doesn't raise an error.

src/psyclone/tests/psyir/transformations/replace_reference_by_literal_trans_test.py
```diff
@@ -0,0 +1,282 @@
...
+        error_str = e.__str__()
+    assert "initial value is not a Literal" in error_str
+
+
+def test_raise_transformation_error_initial_value(
+    fortran_reader, fortran_writer
+):
+    source = """subroutine foo()
+    character(len=4), parameter ::  a = "toto"
+    integer, parameter ::  b = 3+2
```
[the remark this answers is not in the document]
arporter: Still there?

arporter (changes requested):
Looking good, thanks Hugo. I think it's a good idea to add the warnings to the generated code. In my comment about outer scopes, I was thinking we could just create the mapping for *all* symbols in scope (by recursing up through parent symbol tables), while ensuring that we don't conform to scoping rules (i.e. the value associated with the innermost scope takes precedence). However, it's a small point and this PR is probably already sufficient as it is.
Just a few minor things to tidy up.
I'll fire off the integration tests now.
As a reminder - please leave it to me to 'resolve' converstations as that makes it much easier to check everything :-)

arporter:
I've created #2879 to run the integration tests (because you can't do it for a PR from a fork).

arporter:
Integration tests were green :-)

arporter:
Hi @hbrunie, I can't remember: did you say this was ready for review now? (If you could set the label on the PR that really helps me see what I need to do next.)

hbrunie:
> Hi @hbrunie, I can't remember: did you say this was ready for review now? (If you could set the label on the PR that really helps me see what I need to do next.)

Yes sorry, I do it now :-).

arporter (changes requested):
Thanks Hugo. Very, very nearly there now. Just a little bit more tidying and then this is done.

src/psyclone/psyir/transformations/replace_reference_by_literal_trans.py
```diff
@@ -155,38 +156,45 @@ def _update_param_table(
...
                     )
                     sym.preceding_comment += message
                     param_table.pop(sym_name)
                     continue
                 if not isinstance(sym.initial_value, Literal):
+
                     message = (
-                        ReplaceReferenceByLiteralTrans._ERROR_MSG
-                        + f"{sym_name} is assigned "
+                        ReplaceReferenceByLiteralTrans._ERROR_MSG_ONLY_INITVAL
```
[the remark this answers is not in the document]
hbrunie: If I use these static constant it was to facilitate string comparison in unittest.
This way I can just reuse these constant in the test. If we want to change the msg string, we
thus don't have to update the test.

arporter (approved):
All good now, thanks very much @hbrunie.
I'll proceed to merge.

Issue #2831, opened 2024-12-19
Update PSyclone argument definitions to be consistent


mo-alistairp:
In PSyclone the definitions for the function arguments contain inconsistencies in the ordering of abbreviated and full names of arguments (in `PSyclone/src/psyclone/generator.py`). 

The suggested best practice is for the abbreviated flag to come before the full name in the arguments to the parser. For example, `parser.add_argument('-v', '--version',...)` as opposed to `parser.add_argument('--version', '-v',...)`.

PR #2832, opened 2024-12-19
#2831 Reorder arguments to parser for consistency

mo-alistairp:


src/psyclone/generator.py
```diff
@@ -444,7 +444,7 @@ def main(arguments):
         'output Fortran. Use \'output\' to apply line-length limit to output '
```
LonelyCat124, line 444: I know this isn't your modiciations, but can you update this string to use double quotes so we can remove the backslashes from the string (i.e. `"output Fortran. Use 'output'....") (This is just PEP8 compliance update).

LonelyCat124 (changes requested):
Hi @mo-alistairp 

Happy with the changes - I've added one comment to change since I wanted to send it back as you need to add yourself to the Author list in the `generator.py` file. Can you also update the copyright date in that file to include 2025?

I'll submit integration tests in the mean time.

src/psyclone/generator.py
mo-alistairp: Made the requested changes.

LonelyCat124 (approved):
No changes to generated code so no need to run Ready to merge.

PR #2836, opened 2025-01-07
(Closes #2835) Test reverting to 22.04 for tests

### LINKED ISSUE #2835: Python 3.7 tests fail in CI
It seems like 3.7 tests are now failing with error: 
```
  Version 3.7 was not found in the local cache
  Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

@arporter I think we still need to support 3.7 right? Updating the tests to 3.7.1 (or another subnumbered version) might be a solution but I'm not sure.

LonelyCat124:
Tests seem to be failing with 3.7 now, testing if reverting to ubuntu 22.04 fixes this for now.


arporter:
Thanks very much for doing this and top marks for adding the TODO! Please could you do the same trick for the link-checking action as that too is failing with an error that looks like it's caused by using a more recent Ubuntu. Alternatively, maybe there's a better fix? There may be a new release of the linkcheck action?

PR #2840, opened 2025-01-08
Raise error if `ignore_directives=False` but `ignore_comments=True`

### LINKED ISSUE #2839: `ignore_directives` can be `False` if `ignore_comments` is `True`
I've just been testing out the new `ignore_comments` and `ignore_directives` functionality and think it'll be really useful.

However, I noticed that it's possible to set `ignore_directives=False` with `ignore_comments=True`, which negates the `ignore_directives` option choice. Personally, I think this combination should raise an error.

jwallwork23:
Closes #2839.

> I've just been testing out the new `ignore_comments` and `ignore_directives` functionality and think it'll be really useful.
> 
> However, I noticed that it's possible to set `ignore_directives=False` with `ignore_comments=True`, which negates the `ignore_directives` option choice. Personally, I think this combination should raise an error.


src/psyclone/psyir/frontend/fortran.py
```diff
@@ -83,6 +83,11 @@ def __init__(self, free_form: bool = True, ignore_comments: bool = True,
         if not self._parser:
             self._parser = ParserFactory().create(std="f2008")
         self._free_form = free_form
+        if ignore_comments and not ignore_directives:
+            raise ValueError(
```
arporter, line 87: Please document this with a new ":raises ValueError:" in the docstring.

src/psyclone/psyir/frontend/fortran.py
```diff
@@ -83,6 +83,11 @@ def __init__(self, free_form: bool = True, ignore_comments: bool = True,
...
+                "Setting ignore_directives to False will only have an effect"
```
arporter, line 88: "...False in the FortranReader will only..."

arporter (changes requested):
Thanks for tidying this up Joe. Just a couple of small things to fix up. I'll run the integration tests in the meantime (although they should be unaffected).

src/psyclone/psyir/frontend/fortran.py
jwallwork23: Ah yes, good point. Addressed in 347af2ec2.

src/psyclone/psyir/frontend/fortran.py
jwallwork23: Yes, this improves the clarity, thanks. Addressed in 347af2ec2.

arporter (approved):
All good now, thanks very much.
Will just have to wait on the integration tests so will be about an hour.

jwallwork23:
> All good now, thanks very much. Will just have to wait on the integration tests so will be about an hour.

Okay great. Thanks for the review @arporter!

arporter:
The integration tests were (eventually) all green.

PR #2842, opened 2025-01-08
2841 update copyright

hiker:
There are two small updates to the script: 
1. Ignore case
2. Also handle .cfg files.

I did fix up a few files manually (files with no date range given, and no "," after the year) - I couldn't see an easy fix to automate this, and once the file were modified, they won't be a problem again :)

I've asked Sergi, and he's ok with this update.


arporter (approved):
Changes are all good, thanks.
I'll run the integration tests, just to be on the safe side.

arporter:
Integration tests were green.

Issue #2853, opened 2025-01-15
Add support for contiguous attribute


hiker:
We have seen some performance advantage of using contiguous for kernel fields (in gocean using ifort). While fparser supports this :
```
$ fparser2 --std=f2008 ../../gol-lib/combine_mod.f90 
File: '../../gol-lib/combine_mod.f90'
...
    DOUBLE PRECISION, DIMENSION(:, :), CONTIGUOUS, INTENT(OUT) :: current
```
(f2008 is required), PSyclone atm does not (based on a `grep`), which makes it hard to test this more extensively. 

Surprisingly, the error PSyclone raises is an fparser no-match error, the same I get when starting fparser without f2008:
```
  File "/home/joerg/.local/lib/python3.13/site-packages/fparser/two/Fortran2003.py", line 271, in __new__
    return Base.__new__(cls, string, _deepcopy=_deepcopy)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/joerg/.local/lib/python3.13/site-packages/fparser/two/utils.py", line 511, in __new__
    raise NoMatchError(errmsg)
fparser.two.utils.NoMatchError: at line 39
>>>        double precision, dimension(:,:), contiguous, intent(out) :: current
```
As far as I can tell, the parser is initialised with f2008:
```
        if not self._parser:
            self._parser = ParserFactory().create(std="f2008")
```
And I've verified that `self._parser` is indeed None initially.

- [ ] Investigate why contiguous is not accepted by fparser
- [ ] Add support for contiguous to PSyclone

hiker:
While fparser is initially constructed to use f2008, in `create_from_fortran_string` it (re-)initialises fparser as f2003. So from then on the parser instance stored in `fortran._parser` is also changed to be f2003 :( And from then on files are parsed with f2003.

Grepping shows that there is also:
```
./domain/lfric/kernel/common_metadata.py:        _ = ParserFactory().create(std="f2003")
```
Changing `create_from_fortran_string` to also use f2003 fixes the issue, and allows us to handle kernels that use contiguous. 

I suggest that it would make sense to consistently define the Fortrsan standard version in the configuration and consistently use it everywhere.

The additional question: Why are there so many calls like:
```
        # Ensure the Fortran2003 parser is initialised.
        _ = ParserFactory().create(std="f2008")
```
[...]

PR #2855, opened 2025-01-16
#2853 Allow setting of the Fortran standard in the config file.

hiker:
Allows setting of the Fortran standard in the config file, and makes sure that it is consistent everywhere in PSyclone.


hiker:
CI was green as well. Ready for review.

src/psyclone/tests/config_test.py
```diff
@@ -33,7 +33,7 @@
 # -----------------------------------------------------------------------------
 # Author: A. R. Porter, STFC Daresbury Lab
```
arporter, line 34: Not yours, but as a final step, please could you rename this file to `configuration_test.py` so that it conforms to our naming scheme.

arporter (changes requested):
Thanks Joerg, all good apart from just a couple of comments that need updating.
Integration was green before so I won't re-run it.
Updated User Guide renders well.
I also realised that the associated test file (`config_test.py`) doesn't follow our naming convention so am requesting you renaming it to `configuration_test.py`.

src/psyclone/domain/gocean/kernel/psyir.py
```diff
@@ -252,7 +253,8 @@ def create_from_fortran_string(fortran_string):
         kernel_metadata = GOceanKernelMetadata()
 
         # Ensure the Fortran2003 parser is initialised.
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/psyGen.py
```diff
@@ -1577,7 +1577,8 @@ def ast(self):
         # Use the fparser1 AST to generate Fortran source
         fortran = self._module_code.tofortran()
         # Create an fparser2 Fortran2008 parser
-        my_parser = parser.ParserFactory().create(std="f2008")
+        std = Config.get().fortran_standard
```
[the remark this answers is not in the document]
hiker: Done.

src/psyclone/tests/config_test.py
hiker: OK, to run this past you - there are unfortunately two options:
1. The class is actually called `Config`, so from a coding style point of view the file should be called `config.py` (not `configuration.py`), and the test is actually named correctly :(
2. Alternatively, we could call the class `Configuration` instead (and rename the test)

Unfortunately, both these correct solutions will make for e huge patch (since all imports or `Config` uses needs to be updated everywhere :( ).

So I assume you prefer just renaming the test? I am happy to do any of the above (I believe editors can rename a class everywhere), but am aware of the potential impact (I could do this after Sergi's patch goes in?). Just let me know what you prefer.

hiker:
@arporter - I need one clarification re the renaming, as asked above. Just quickly let me know which of the three solutions you prefer.

src/psyclone/tests/config_test.py
arporter: Oh dear. Let's just rename the test then :-)

hiker:
I thought you would opt for renaming the test only :) Done.

arporter (approved):
Thanks Joerg. Since master has moved on quite a bit I shall re-run the integration tests but I expect everything will be fine.

arporter (approved):
Integration tests were green.
Will proceed to merge.

PR #2857, opened 2025-01-17
(Closes #2854) exclude char assignments from ACC KERNELS regions

### LINKED ISSUE #2854: ACCKernelsTrans should exclude character assignments
Daley has found that if we include character assignments within a KERNELS region, the nvfortran compiler (24.5) errors with:
```
Character local variable unsupported on device
```

Unfortunately, a simple example does not replicate this problem which indicates it's probably a compiler bug and depends on the precise context.

NEMO source files that exercise this problem: `obs_group_def.f90`, `sbccpl.f90`

arporter:
Small PR that picks up the functionality already in ArrayAssignment2LoopsTrans and copies it into ACCKernelsTrans.


arporter:
I'm not entirely happy with the code duplication here but this PR is for OpenACC functionality and the existing check is in the conversion of array assignments to loops so there's not much common ground between them.

arporter:
I've solved the duplication by making the validation a separate classmethod in ArrayAssignment2LoopsTrans. Coverage is 100%. I'll set the integration tests running.

arporter:
Integration test for NEMOv4 failed but only because of a small bug where we failed to lower-case the name of a structure component when getting its datatype. I'll re-run that job.

arporter:
Integration tests now green. Ready for review from anyone :-)

arporter:
I changed my mind - it turns out my 'fix' revealed general case-sensitivity problems in our storage of components of a StructureType. I've fixed these now and extended the tests slightly.

arporter:
Ready for review now. Integration tests are finally happy :-)

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -148,6 +161,10 @@ def validate(self, nodes, options=None):
                 "GOcean InvokeSchedules")
         super().validate(node_list, options)
 
+        # Any errors below this point will optionally log the resason, which
```
LonelyCat124: s/resason/reason/

src/psyclone/tests/psyir/nodes/structure_reference_test.py
```diff
@@ -241,8 +241,8 @@ def test_struc_ref_datatype():
     grid_type = symbols.StructureType.create([
         ("nx", symbols.INTEGER_TYPE, symbols.Symbol.Visibility.PUBLIC, None),
         ("data", atype, symbols.Symbol.Visibility.PRIVATE, None),
-        # A single member of structure type.
-        ("roger", rtype, symbols.Symbol.Visibility.PUBLIC, None),
+        # A single member of structure type with a mixed-case name.
+        ("roGEr", rtype, symbols.Symbol.Visibility.PUBLIC, None),
         # An array of structure type.
         ("titty", artype, symbols.Symbol.Visibility.PUBLIC, None)])
```
LonelyCat124, line 247: I'll leave this up to you - but maybe some renaming of arguments/variables in this file.

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -77,12 +77,12 @@ class ACCKernelsTrans(RegionTrans):
     excluded_node_types = (CodeBlock, Return, PSyDataNode,
                            psyGen.HaloExchange, WhileLoop)
 
-    def apply(self, node, options=None):
+    def apply(self, node: Union[Node, List[Node]], options: dict = None):
```
LonelyCat124, line 80: I'm not sure if typing this is intentional or not - I've not seen this anywhere in PSyclone previously - if it is a new plan and I've missed it then let me know.

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -133,8 +141,13 @@ def validate(self, nodes, options=None):
             a routine that is not available on the accelerator.
         :raises TransformationError: if there are no Loops within the
             proposed region and options["disable_loop_check"] is not True.
+        :raises TransformationError: if any assignments in the region contain a
+            character type child and the allow_strings option is not set.
```
LonelyCat124, line 151: the option is 'allow_string', I think maybe `'allow_string'` inside quotes or something here to make it clear. Alternatively match the previous Error docstring for this and just use `options["allow_string"] is not True.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
```diff
@@ -372,6 +359,41 @@ def validate(self, node, options=None):
                 raise TransformationError(LazyString(
                     lambda: f"{message} In:\n{node.debug_string()}"))
 
+    @staticmethod
+    def validate_no_char(node: Node, message: str, options: dict) -> None:
```
LonelyCat124, line 363: Typed argument list - see my other query on this.

LonelyCat124 (changes requested):
@arporter Couple of small changes in docs, and a couple of queries I'm just unsure about.

arporter:
Imaran has found that this PR doesn't fix e.g. `obs_surf_def.f90` because that has a character string as a member of a derived type and we haven't resolved the derived type so don't know that it's a character string. There are a few options:

 * always attempt to resolve the type in such a case
 * refuse to allow unresolved types inside ACC regions
 * exclude the file from PSyclone processing

src/psyclone/psyir/transformations/acc_kernels_trans.py
arporter: This has gone following changes on master.

src/psyclone/tests/psyir/nodes/structure_reference_test.py
arporter: Yes, fair point. It was a Swallows and Amazons naming scheme but that is probably too obscure :-)

src/psyclone/psyir/transformations/acc_kernels_trans.py
arporter: Good idea, done.

arporter:
Although this doesn't fix all our (OpenACC) problems, it is a step forward so I've opted to keep this PR small and self-contained and not attempt to fix anything else. Ready for another look now. I'll fire off the integration tests as master has moved on a lot since I last did that.

src/psyclone/tests/psyir/nodes/structure_reference_test.py
LonelyCat124: Ah ok, I did not know the reference so was a little confused!

LonelyCat124 (approved):

LonelyCat124:
Integration tests work - am going to have one more local pass before marking this ready to merge.

PR #2869, opened 2025-01-23
Remove psyir if branch

hbrunie:
This PR adds a new PsyIR transformation.

What this transformation should do exactly should be discussed between @schreiberx @arporter and other interested.

Currently there is no test, it is only reusing the code present in Martin's PR #2801.


src/psyclone/psyir/transformations/remove_ifblock_trans.py
```diff
@@ -0,0 +1,145 @@
...
+from typing import Dict, Optional
+
+
+class RemoveIfBlockTrans(Transformation):
+
+    def __init__(self, json_file_abspath: Optional[str] = None) -> None:
+        super().__init__()
+        self._known_reference_bool: Dict[str, bool] = {}
+        self._known_reference_int: Dict[str, bool] = {}
+        if json_file_abspath is not None:
```
arporter, line 55: This part would need to either be removed entirely (and the transformation purely on PSyIR information for determining compile-time constants) or replaced with functionality that enables a list of names/symbols to be provided to the `apply` method (to support run-time constants).

arporter:
Thanks for this Hugo. Obviously this will need full testing etc. Also, the reading of the json file needs to be replaced with one or more parameters to `apply`. One of us could take this on if need be?

PR #2877, opened 2025-01-29
(Towards #2668) Implementation of keyword arguments to transformations instead of options dict.

LonelyCat124:
This is still an early draft - documentation not yet handled.
This doesn't replace options at current, merely provides an alternative, more pythonic approach. Removing options would be a future issue for a large release of PSyclone, unlikely to happen in the immediate future.


LonelyCat124:
Docs still fail here, but I can't duplicate it locally so I can't diagnose it - for me the docs build and look ok. There's some warning to do with line indentation/breaks but I can't work out what it means.

LonelyCat124:
Documentation is now building correctly it seems, I need to check the changes I've made haven't broken the docs.

This is still TODO.

> It doesn't have type checking in docstrings correct yet. E.g. if we have something like :param bool myname: it can't yet determine that the name of the variable is myname with type bool at this point (or even worse if its like bool | int).

LonelyCat124:
Documentation now builds, essentially it seems like generating `:type ...` wasn't working correctly, I'm not sure why but presumably I was doing it wrong, but since its reading from the type I can just add it to the `:param`.

Improvements and detailed testing are up next - I want this to be able to support metatransformations with inheriting docstrings as well, so I want to do that as part of the testing and improvements.

LonelyCat124:
@arporter This is ready for a look now - once we're happy with the implementation then I'll start propagating the changes around more transformations.

setup.py
```diff
@@ -163,7 +163,7 @@ def get_files(directory, install_path, valid_suffixes):
         package_dir={"": "src"},
         install_requires=['pyparsing', 'fparser>=0.2.0', 'configparser',
                           'jsonschema', 'sympy', "Jinja2", 'termcolor',
-                          'graphviz'],
+                          'graphviz', 'sphinx'],
```
sergisiso, line 166: I am not too keen on making 'shpinx' a required dependency, it is a huge dependency, see spack below.

It seems we use it to "pretty-print" types? Could we modify the "from sphinx.util.typing import stringify_annotation" so that if sphinx is not found we define a stringify_annotation method with: `retrun str(v.annotation)` or something simple. Development environments and when we generate the documentation will have the shpinx libs and will generate the nicer text? or do we use it for something else?

image

LonelyCat124: From what I remember I had issues with 3.7 and 3.8 without the sphinx import and it not being able to turn it into a string, but I can take another look.

LonelyCat124: Just pushed a change without sphinx being a required dependency, ended up on pragma no cover for the lines importing a local version of stringify_annotation when i couldn't import one from sphinx.

src/psyclone/psyGen.py
```diff
@@ -2784,6 +2796,108 @@ def validate(self, node, options=None):
...
+                continue
+            signature = inspect.signature(base_cls_apply)
+            # Loop over the arguments to the apply call.
+            for k, v in signature.parameters.items():
+                if k == "options":
+                    continue
+                # If the argument is a keyword argument, i.e. it has a default
+                # value then we add it to the list of options.
+                if v.default is not inspect.Parameter.empty:
+                    valid_options[k] = {}
```
arporter, line 2831: Having `valid_options` be a dict of `dataclass`'s might be nicer.

src/psyclone/psyGen.py
```diff
@@ -2784,6 +2796,108 @@ def validate(self, node, options=None):
...
+        :raises ValueError: If an invalid option was provided.
+        :raises TypeError: If an option was provided with a value of the wrong
+                           type.
+        '''
+        valid_options = type(self).get_valid_options()
+        invalid_options = []
+        wrong_types = {}
+        for option in kwargs:
+            if option == "options":
+                continue
```
arporter, line 2863: Presumably we'll also look into the `options` dict eventually?

src/psyclone/psyir/transformations/loop_trans.py
```diff
@@ -99,16 +115,19 @@ def validate(self, node, options=None):
...
         if not options:
-            options = {}
-        if not isinstance(options, dict):
-            raise TransformationError(
-                f"Transformation validate method 'options' argument must be a "
-                f"dictionary but found '{type(options).__name__}'.")
-
-        verbose = options.get("verbose", False)
+            self.validate_options(**kwargs)
+            verbose = self.get_option("verbose", **kwargs)
```
arporter, line 119: Could/should we merge together `options` with **kwargs and then we could do away with this `if not options....else...`?

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -228,7 +228,9 @@ def _directive(self, children, collapse=None):
             node.reprod = self._reprod
         return node
 
-    def apply(self, node, options=None):
+    def apply(self, node, options=None,
+              reprod: bool = Config.get().reproducible_reductions,
```
arporter, line 232: This use of Config() may have unintended consequences as it will instantiate it when this module is imported and that *may* be before any custom config file has been parsed? @hiker is probably best placed to comment on this.

LonelyCat124: Thats fine - I can change this to be None and do an `if reprod is None` check after to initialise it when this file is called.

src/psyclone/psyGen.py
LonelyCat124: Not as planned for now, this is replacement for the `options` dict so at some point in the future i'd deprecate and remove the options dict in favour of the new strategy (thus removing the need for this check).

Can reevaluate if you want to keep both strategies, but I think its unnecessary when for most current options you can just pass `**my_dict` instead of `options=mydict` with no other changes.

src/psyclone/psyir/transformations/omp_task_trans.py
```diff
@@ -69,7 +69,7 @@ def name(self):
         '''
         return "OMPTaskTrans"
 
-    def validate(self, node, options=None):
+    def validate(self, node, options=None, **kwargs):
```
arporter, line 72: I wanted to examine the new docs produced for this class but when I look at the Sphinx AutoAPI docs for this method I just see:
image
which isn't very helpful :-( The doc for LoopTrans looks OK:
image
Do you or @sergisiso have any idea why this might be?

arporter (changes requested):
This is just a first pass. I think it looks good and I really like the fact you've managed to get type checking in there too. As mentioned inline, I don't seem to be able to see the results of this change in the generated ReferenceGuide though.

.github/workflows/python-package.yml
```diff
@@ -132,6 +132,8 @@ jobs:
         # Some of the examples use Jupyter. We need to keep IPython to
         # below 8.0 as we still run the test suite with Python 3.7.
         pip install "ipython>=7.0,<8.0" jupyter
+        # We need to install sphinx to get correct doc testing
+        pip install sphinx
```
arporter, line 136: Could we just change L140 below to be `pip install .[test,doc]`?

src/psyclone/psyir/transformations/omp_task_trans.py
LonelyCat124: Thats extra weird because the OMPTaskTrans hasn't had the new stuff applied to it (so won't yet inherit its docs) and printing the classes docs makes them appear correctly.

Do the docs for ParallelLoopTrans appear correctly?

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -243,9 +245,13 @@ def apply(self, node, options=None):
 
         '''
         if not options:
```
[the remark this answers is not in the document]
LonelyCat124: One challenge here (other than the medium term removal of the options dict) is that we have already specified option names on some transformations that are not supported as arguments to functions (e.g. `"node-type-check"` on RegionTrans). I don't think we should rename those arguments so this gets a bit complex.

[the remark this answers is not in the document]
sergisiso: Could we rename them with s/-/_/ when passing them to kwargs?

I would also be in favour to "print" a deprecation warning everytime somebody uses the "options" even if we now support both.

src/psyclone/psyir/transformations/loop_trans.py
```diff
@@ -62,7 +62,23 @@ class LoopTrans(Transformation, metaclass=abc.ABCMeta):
...
+        :type node: subclass of :py:class:`psyclone.psyir.nodes.Node`
+        :param options: a dictionary with options for transformations.
+        :type options: Optional[Dict[str, Any]]
+        :param node_type_check: If the type of nodes enclosed in the loop
+                                should be tested to avoid including
+                                unsupported nodes in the transformation.
+        :param verbose: whether to log the reason the validation failed, at
+                        the moment with a comment in the provided PSyIR node.
+        '''
+        super().apply(node, options=options, node_type_check=node_type_check,
```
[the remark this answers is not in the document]
LonelyCat124: Added

src/psyclone/utils.py
```diff
@@ -67,3 +70,186 @@ def a_or_an(string):
...
+            )
+        else:
+            added_parameters = ""
+        update_func_docstring(cls.apply, added_parameters)
+        # Update the validate docstring
+        add_parameters = get_inherited_parameters(cls, cls.validate, cls.apply)
+        update_func_docstring(cls.validate, add_parameters)
+
+        return cls
+    return wrapper(*args, **kwargs)
```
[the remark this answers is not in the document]
LonelyCat124: Updates the docstrings for apply and validate for the provided class (cls) according to the arguments provided to the decorator (`transformation_documentation_wrapper`). I'll add documentation.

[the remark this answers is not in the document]
LonelyCat124: Docstrings added.

src/psyclone/psyir/transformations/omp_loop_trans.py
LonelyCat124: Changed.

src/psyclone/psyir/transformations/omp_loop_trans.py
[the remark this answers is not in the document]
LonelyCat124: I guess something like:
```
if options:
    print("Deprecation warning: Use of the options dict is now deprecated, see etc.....")
    for option in options:
        # Replace all - with _ in option.
        kwargs[underscored_option] = options[option]
```

Then just do the base code. The bit where I'm not sure how to do this is for the non-keyword arguments, e.g.:

```
        if not options:
            if reprod is None:
                reprod = Config.get().reproducible_reductions
            self.validate_options(
                    reprod=reprod, **kwargs
            )
            self._reprod = reprod
        else:
            self._reprod = options.get("reprod",
```

I think this code is unavoidable to have `if not options` code?

.github/workflows/python-package.yml
LonelyCat124: I'll try it

LonelyCat124: On the same line it was failing for some reason, have moved it to be on 2 lines.

LonelyCat124:
Ok better parser done and tested.
I realised that having the ability to merge the `DocstringData` objects together will be helpful. I'll try to convert the other code to use this at some point soon, not sure exactly when.

LonelyCat124:
@arporter This is ready for another look, the code in utils.py has changed significantly - if the documentation is still not in the reference guide for things let me know - in particular for ParallelLoopTrans (as that is the one function right now using the wrapper to generate inherited docstrings).

src/psyclone/utils.py
```diff
@@ -67,3 +70,186 @@ def a_or_an(string):
...
+        docs = func.__doc__
+        parent_docstring = inheriting_func.__doc__
+        if parent_docstring is not None:
+            parent_lines = parent_docstring.splitlines()
+        else:
+            parent_lines = []
+        added_docs = ""
+        for x, line in enumerate(parent_lines):
+            if ":param" in line:
+                param_str = re.search(":[a-zA-Z0-9_\\s]*:", line)
```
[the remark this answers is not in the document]
sergisiso: @arporter @LonelyCat124 Why do we need to pull what we need and not just use it, it is on spack with reasonable dependencies: https://packages.spack.io/package.html?name=py-docstring-parser and if the results are good, why can not we replace fully replace sphinx stringify_annotation?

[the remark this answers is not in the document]
LonelyCat124: It doesn't do everything we need, it doesn't grab docstrings from type hints (which is what we need stringify_Annotation for also) which is important for inheriting docstring I think (since sphinx can't work it out from just the type hint of the argument).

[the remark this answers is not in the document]
arporter: I guess we could potentially contribute back to that package and, if our changes are accepted, migrate to using it? That's perhaps (?) more work than we want to do right now though.

src/psyclone/psyir/transformations/omp_task_trans.py
LonelyCat124: It might be - I'll try to check the output on Monday since the validate method docstring will be generated by the docstring generator now.

sergisiso: Isn't the apply above still missing the arguments/return/raises sections? for other interfaces it looks like:
image
I was expecting to see the same, now with each inherited "option" parameter also listed and documented?

src/psyclone/psyGen.py
```diff
@@ -2781,6 +2803,120 @@ def validate(self, node, options=None):
...
+
+        if len(invalid_options) > 0:
+            invalid_options_detail = []
+            for invalid in invalid_options:
+                invalid_options_detail.append(f"'{invalid}'")
+            invalid_options_list = ", ".join(invalid_options_detail)
+            raise ValueError(f"'{type(self).__name__}' received invalid "
+                             f"options [{invalid_options_list}]. Please see "
+                             f"the documentation and check the available "
+                             f"options.")
```
arporter, line 2906: we could just print the list of valid options here too.

src/psyclone/psyir/transformations/loop_trans.py
arporter: Since we're going to deprecate `options`, please add a TODO here. I think @sergisiso was in favour of printing a deprecation warning too but presumably we can do that in the base class?

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -228,13 +228,15 @@ def _directive(self, children, collapse=None):
...
+    def apply(self, node, options=None,
+              reprod: bool = None,
+              **kwargs):
         '''Apply the OMPLoopTrans transformation to the specified PSyIR Loop.
 
         :param node: the supplied node to which we will apply the \
                      OMPLoopTrans transformation
         :type node: :py:class:`psyclone.psyir.nodes.Node`
-        :param options: a dictionary with options for transformations\
+        :param options: a dictionary with options for transformations
```
arporter, line 239: We need a description of the `reprod` argument now.

src/psyclone/psyir/transformations/omp_loop_trans.py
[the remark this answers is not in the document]
arporter: I'm a bit confused here.  What do you mean by the "non-keyword arguments"?

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -266,4 +274,4 @@ def apply(self, node, options=None):
                     "nthreads", tag="omp_num_threads",
                     symbol_type=DataSymbol, datatype=INTEGER_TYPE)
 
-        super().apply(node, options)
+        super().apply(node, options, **kwargs)
```
arporter, line 277: Can we merge `options` into `**kwargs` now (allowing for the `-` -> `_` replacement)?

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -311,14 +331,31 @@ def apply(self, node, options=None):
...
-        ignore_dep_analysis = options.get("force", False)
-        list_of_names = options.get("ignore_dependencies_for", [])
-        privatise_arrays = options.get("privatise_arrays", False)
+            self.validate_options(
+                    verbose=verbose, collapse=collapse,
+                    ignore_dependencies_for=ignore_dependencies_for,
+                    privatise_arrays=privatise_arrays,
+                    sequential=sequential, **kwargs
+            )
+            # Rename the input options that are renamed in this apply method.
```
arporter, line 340: We should probably rename them to make them consistent. Please add a TODO.

arporter (changes requested):
Thanks Aidan. I like the new dataclasses but have some suggestions that might improve things a bit. I'm in two minds as to whether we should include the license of the parser code. Now that I write that, I think we should - probably something like:

# our normal license info....
#
# This code is based on the xxxx package by yyyyy. This is subject to the following
# license:
#
# Their MIT license....

src/psyclone/psyir/transformations/omp_task_trans.py
LonelyCat124: > Isn't the apply above still missing the arguments/return/raises sections? for other interfaces it looks like: image I was expecting to see the same, now with each inherited "option" parameter also listed and documented?

The OpenMPLoopTrans doesn't have the new thing applied yet - only the ParallelLoopTrans for now, let me see if I can build the docs locally and chekc.

src/psyclone/psyir/transformations/omp_loop_trans.py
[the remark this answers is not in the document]
LonelyCat124: I think here what I mean is we can't remove the `if not options` for reprod because its an argument to this apply, so we can't just do
```
if not options:
    options = kwargs
self._reprod = options.get("reprod", ....)
```
because if we're not using the options dictionary we instead need to do `self._reprod = reprod` etc.

src/psyclone/psyir/transformations/omp_task_trans.py
LonelyCat124: Ok - it looks like there is a bug in the docstring writer where its not generating correct documentation.

Separately I'm not sure if the docs are generating the docstring from the updated `__doc__` or not or if thats related, but for now I need to check that docdata <=> doc is a fully reversible thing.

LonelyCat124: Ok, I fixed this, there were some indentation issues, I'll now check the sphinx generated docs.

LonelyCat124: Still some (probable) indentation issues. I think I should also switch the "\n" in outputs to `os.linesep`

LonelyCat124: image
It seems to work correctly now for me.

src/psyclone/psyir/transformations/loop_trans.py
LonelyCat124: I think it'll have to be on a class-by-class basis, those that call super().apply() can use the inherited call to their parent for the deprecation print. I'll add this into some of these methods for now but spreading it around the code more will be later in this PR.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    :param obj: The object that is having its docstring analysed.
+
+    :raises InternalError: If an unsupported docstring input is found.
+
+    :returns: An object representing the input arg and desc.
+    '''
+    # If its a param then we can create an ArgumentData for this.
+    if args[0] == "param":
+        if len(args) == 2:
+            inline_type = False
```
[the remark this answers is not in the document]
LonelyCat124: Added something similar

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+            for k, v in signature.parameters.items():
+                if k != name:
+                    continue
+                if v.annotation is not inspect.Parameter.empty:
+                    datatype = stringify_annotation(
+                        v.annotation
+                    )
+                    inline_type = True
+        elif len(args) == 3:
+            inline_type = True
```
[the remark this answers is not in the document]
LonelyCat124: Added something similar

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    # If its a param then we can create an ArgumentData for this.
+    if args[0] == "param":
+        if len(args) == 2:
+            inline_type = False
+            datatype = None
+            name = args[1]
+            # Check if there is a datatype on the argument itself and store
+            # it.
+            signature = inspect.signature(obj)
+            for k, v in signature.parameters.items():
```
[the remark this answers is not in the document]
LonelyCat124: Yeah that's much neater

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+                f"Found parameter docstring of unsupported type, expected "
+                f":param arg: or :param type arg: but found "
+                f":{' '.join(args)}:"
+            )
+        return ArgumentData(
+            name=name, desc=desc, datatype=datatype, inline_type=inline_type
+        )
+    elif args[0] == "raises":
+        if len(args) != 2:
+            raise InternalError(
```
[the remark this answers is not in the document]
LonelyCat124: Yeah, DocParseError seems like a good idea.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+                f"Found raises docstring of unsupported type, expected "
+                f":raises Error: but found :{' '.join(args)}:"
+            )
+        return RaisesData(exception=args[1], desc=desc)
+    elif args[0] == "returns":
+        if len(args) == 1:
```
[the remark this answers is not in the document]
LonelyCat124: Added

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    :param docdata: The docdata object to convert to docstring.
+    :param indentation: The indentation to use. Default is "    ".
+    :param obj: The object who the generated docstring will be for. Default
+                option is None, which means to generate the full docstring
+                including all type lines.
+
+    :returns: The docstring for the input docdata.
+    '''
+    if docdata.desc is not None:
+        description = indentation + docdata.desc + "\n"
```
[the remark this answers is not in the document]
LonelyCat124: I think I fixed these all now to be consistent and indent, but don't check lines are less than 80 chars. Do we need to worry about that since it doesn't change the code as viewed, only the docstrings when printed or generated by sphinx?

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+        obj: Any
+) -> Union[DocstringData, None]:
+    '''
+    Converts the docstring of obj into a DocstringData object. Only supports
+    docstring used in PSyclone, so some valid docstring may result in failure.
+
+    :param Any obj: The object whose docstring will be parsed.
+
+    :raises ValueError: if a docstring element cannot be parsed.
+    :raises InternalError: if a type docstring is found with no correspdongin
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    '''
+    text = obj.__doc__
+
+    if text is None:
+        return None
+
+    # Remove indentation from the documentation
+    text = inspect.cleandoc(text)
+
+    # Find the first instance of ^: as the start of the meta information.
```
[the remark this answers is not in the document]
LonelyCat124: Yeah - I swapped to an explicit regex to handle a line starting with a : inside the description.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    types = {}
+    rtype = None
+    docstring_data = DocstringData(
+        desc="", arguments=OrderedDict(), raises=[],
+        returns=None
+    )
+
+    docstring_data.add_data(desc_chunk)
+
+    # Loop through the meta chunk for each line that section that has
```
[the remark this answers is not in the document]
LonelyCat124: Added an in so it makes sense now.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+    # Find the first instance of ^: as the start of the meta information.
+    match = re.search("^:", text, flags=re.M)
+    if match:
+        desc_chunk = text[:match.start()]
+        meta_chunk = text[match.start():]
+    else:
+        desc_chunk = text
+        meta_chunk = ""
+
+    # Can break the ValueError with something like this
```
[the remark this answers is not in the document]
LonelyCat124: I moved this comment to the correct part of the code now. Its the reason for the ValueError in the loop.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+        if len(args) == 2 and args[0] == "type":
+            types[args[1]] = desc
+        elif len(args) in [1, 2] and args[0] == "rtype":
+            rtype = desc
+        else:
+            # Get the information.
+            docdata = create_docstring_data(args, desc, obj)
+            docstring_data.add_data(docdata)
+
+    for param in types:
```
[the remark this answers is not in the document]
LonelyCat124: Done.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,455 @@
...
+        if param not in docstring_data.arguments.keys():
+            raise InternalError(
+                f"Found a type string with no corresponding parameter: "
+                f"'{param}' type found with no parameter docstring."
+            )
+        docstring_data.arguments[param].datatype = types[param]
+        docstring_data.arguments[param].inline_type = False
+
+    if rtype:
```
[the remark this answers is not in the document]
LonelyCat124: Done

src/psyclone/psyGen.py
LonelyCat124: Added

src/psyclone/psyir/transformations/omp_loop_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/omp_loop_trans.py
LonelyCat124: I tried it here and its possible I think - I think some function to do this would be good so it can be used in more places, I suspect as a class function of Transformation?

src/psyclone/psyir/transformations/parallel_loop_trans.py
LonelyCat124: Added a TODO (put this issue since this is all about options).

LonelyCat124:
@arporter Ready for another look - the only outstanding thing I've not managed to work out is the options/kwargs merging - I tried it and it didn't pass options upstream for now but maybe there's something I missed.

src/psyclone/psyir/transformations/omp_loop_trans.py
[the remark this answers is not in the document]
arporter: Ah, I think I understand now - we're keeping `options` for backwards compatibility (for now) so we have to support both routes. That's fine.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+    name: str
+    datatype: str
+    desc: str
+    inline_type: bool
+
+    def gen_docstring(self, obj: Union[None, Any] = None) -> str:
+        '''
+        :param obj: The object who the generated docstring will be for. Default
+                    option is None, which means to generate the full docstring
+                    including all type lines.
```
arporter, line 103: I'm a bit confused here. You say "all type lines" but, even if `obj` is None, there's only one type line generated. Is it more accurate to say something like: "If no object is supplied, there can be no type annotation and so the type information is included inline in the :param or in a separate :type entry."?

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+                f":raises Error: but found :{' '.join(args)}:"
+            )
+        return RaisesData(exception=args[1], desc=desc)
+    elif args[0] == "returns":
+        if len(args) == 1:
+            # Return of form :returns: desc
+            inline_type = False
+            datatype = None
+        elif len(args) == 2:
+            # Return of form :returns datatype: desc
```
arporter, line 520: As mentioned earlier - are you sure this is valid Sphinx syntax?

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+    :param str datatype: The string representation of the parameter's
+                         datatype.
+    :param str desc: The description of the parameter.
+    :param bool inline_type: whether the type of this parameter should be
+                             inline or a separate :type param: docstring line.
+    '''
+    name: str
+    datatype: str
+    desc: str
+    inline_type: bool
```
arporter, line 97: Since this class represents the description of an argument, could it also store the argument object to save you having to pass it to `gen_docstring()`?

src/psyclone/psyGen.py
```diff
@@ -2711,6 +2725,14 @@ def _find_subclasses(self, module, base_class):
                 issubclass(cls, base_class) and cls is not base_class]
 
 
+@dataclass
+class ValidOption:
+    '''Class used to specify the valid options dict for a Transformation.'''
+    default: object
```
arporter, line 2731: It's 'early' and I've not had much coffee but I'm struggling to remember what these different elements correspond to. Please could you document them (using the normal :param syntax in the docstring).

src/psyclone/psyGen.py
```diff
@@ -2750,8 +2772,14 @@ def apply(self, node, options=None):
         :type options: Optional[Dict[str, Any]]
 
         '''
+        # TODO 2668: options are now deprecated:
+        if options is not None:
+            print("Deprecation Warning: The options parameter to "
```
arporter, line 2777: Probably best to prefix this, e.g. "PSyclone deprecation warning: ..."

src/psyclone/psyGen.py
```diff
@@ -2780,6 +2808,126 @@ def validate(self, node, options=None):
         :type options: Optional[Dict[str, Any]]
         '''
         # pylint: disable=unused-argument
+        # TODO 2668: options are now deprecated:
+        if options is not None:
+            print("Deprecation Warning: The options parameter to "
```
arporter, line 2813: "PSyclone deprecation warning" again please. You could store this message text as a class variable perhaps.

src/psyclone/psyGen.py
```diff
@@ -2781,6 +2803,120 @@ def validate(self, node, options=None):
...
+        Pulls the valid options from the apply method. It also recurses
+        upwards to the superclasses of this transformation and pulls
+        their valid options as well if they exist.
+
+        :returns: A dict of the valid option name and corresponding
+                  ValidOption dataclass.
+        '''
+        valid_options = OrderedDict()
+        # Loop through the inherited classes of this class, starting with the
+        # highest superclass. For some reason super isn't working here.
```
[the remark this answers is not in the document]
arporter: Yeah, let's leave it.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -272,7 +289,11 @@ def validate(self, node, options=None):
                         node.append_preceding_comment(f"PSyclone: {messages}")
                 raise TransformationError(messages)
 
-    def apply(self, node, options=None):
+    def apply(self, node, options=None, verbose: bool = False,
```
arporter, line 292: Presumably we need to document each of these arguments now or are you planning to leave that until we remove the `options` argument?

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -310,15 +331,39 @@ def apply(self, node, options=None):
...
+            # Rename the input options that are renamed in this apply method.
+            # TODO 2668, rename options to be consistent.
+            ignore_dep_analysis = force
+            if ignore_dependencies_for is None:
+                list_of_names = []
+            else:
+                list_of_names = ignore_dependencies_for
+        else:
+            if len(options.keys()) > 0:
+                print("Deprecation Warning: The options parameter to "
```
arporter, line 351: Use message from a class variable perhaps?

src/psyclone/tests/psyGen_test.py
```diff
@@ -183,6 +183,83 @@ def apply(self, _1):
     assert trans.name == "TestTrans"
 
 
+def test_transformation_get_options():
+    ''' Test that the get_option method behaves in the
+    expected way.'''
+    class TestTrans(Transformation):
```
arporter, line 189: You could move this class definition out of this test and into the parent module. Then it's available to be reused in each test, instead of having to define it multiple times.

src/psyclone/tests/psyGen_test.py
```diff
@@ -183,6 +183,83 @@ def apply(self, _1):
...
+            pass  # pragma: no cover
+
+    options = TestTrans.get_valid_options()
+    assert options['valid'].default
+    assert options['valid'].type is bool
+    assert options['valid'].typename == "bool"
+    assert options['untyped'].default is False
+    assert options['untyped'].type is None
+    assert options['untyped'].typename is None
+
```
arporter, line 236: Please add a test showing that valid options are inherited correctly from parent transformation classes.

src/psyclone/tests/psyad/main_test.py
```diff
@@ -624,7 +624,17 @@ def test_main_otest_lfric(tmpdir, capsys):
...
-    assert output == ""
+    # TODO 2668: This deprecation of options causes this test to see the
+    # deprecation message.
+    correct = '''Deprecation Warning: The options parameter to \
+Transformation.validate is now deprecated. Please use the individual \
+arguments, or unpack the options with **options. \
+See the documentation for more details.
+Deprecation Warning: The options parameter to Transformation.validate \
+is now deprecated. Please use the individual arguments, or unpack the \
+options with **options. See the documentation for more details.
```
arporter, line 635: This makes me think we should write a short paragraph (probably in the user guide section on Transformations) about the move from the `options` dict to having named arguments. We could give an example of the old and the new syntax.

src/psyclone/tests/psyir/transformations/loop_trans_test.py
```diff
@@ -183,3 +183,23 @@ def fake_validate(_1, _2, options=None):
...
+           integer :: i
+           integer, target :: a
+           integer, pointer :: b
+           do i = 1, 10
+               b => a
+           enddo
+       end subroutine test
+    ''')
+    loop = psyir_test.walk(Loop)[0]
+    fake_class().apply(loop)
```
arporter, line 205: I guess this should do nothing as it's an empty method? Please could you check therefore that the transformation has left the code unchanged (with a comment explaining).

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -538,6 +539,52 @@ def test_omplooptrans_apply(sample_psyir, fortran_writer):
...
+    # Check we get the relevant error message when submitting multiple
+    # options with the wrong type
+    with pytest.raises(TypeError) as excinfo:
+        omplooptrans.apply(tree.walk(Loop)[0], verbose=3, force="a")
+    assert ("'OMPLoopTrans' received options with the wrong types:\n"
+            "'verbose' option expects type 'bool' but received '3' "
+            "of type 'int'.\n"
+            "'force' option expects type 'bool' but received 'a' "
+            "of type 'str'.\n"
+            "Please see the documentation and check the provided types."
```
arporter, line 567: Nice :-)

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -538,6 +539,52 @@ def test_omplooptrans_apply(sample_psyir, fortran_writer):
...
+            in str(excinfo.value))
+
+    # Check python version, as this tests have different behaviour for
+    # new python versions vs 3.8 or 3.7.
```
arporter, line 571: Please add a TODO here to the appropriate issue if we can remove this bit once we drop support for 3.8.

arporter (changes requested):
Looking good now Aidan. I think the code is almost there - let's just see if it can be tidied a little more and then I'll finish reviewing the tests.
As I've said in the comments, I don't think you should expand this PR to cover any more transformations. That can be done by several of us in subsequent PRs.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: A. B. G. Chalk, STFC Daresbury Lab
+# -----------------------------------------------------------------------------
+#
+# This code is influenced by the docstring_parser packaged created by Marcin
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+    desc: str
+    datatype: object
+    inline_type: bool
+
+    def gen_docstring(self) -> str:
+        '''
+        :returns: The docstring for this ReturnsData object.
+        '''
+        if self.inline_type:
+            return f":returns {self.datatype}: {self.desc}"
```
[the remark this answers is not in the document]
LonelyCat124: Ah ok, I'll remove the inline_type part of ReturnsData and the corresponding code.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+
+@dataclass
+class DocstringData():
+    '''
+    Dataclass object representing a full docstring.
+
+    :param str desc: The description of the docstring.
+    :param OrderedDict[str, ArgumentData] arguments: An OrderedDict containing
+        the ArgumentData that respresents the parameter section of the
+        docstring.
```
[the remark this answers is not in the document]
LonelyCat124: Added

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+        '''
+        Adds the docstring_element to the DocstringData object.
+
+        :docstring_element: The docstring_element object to be added to the
+                            DocstringData.
+
+        :raises DocParseError: If a docstring element not supported by
+                               PSyclone is found.
+        '''
+        # FIXME We can do some checking that its not already present? But
```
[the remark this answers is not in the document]
LonelyCat124: Removed and docstring updated.

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+
+        if self.returns is not None:
+            returnstring = self.returns.gen_docstring()
+            if "\n" in returnstring:
+                lines = returnstring.split("\n")
+                returnstring = ""
+                for line in lines:
+                    returnstring += indentation + line + os.linesep
+            else:
+                returnstring = indentation + returnstring + "\n"
```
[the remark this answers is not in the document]
LonelyCat124: I removed all traces of \n from the code (not comments) and replaced with os.linesep

src/psyclone/docstring_parser.py
```diff
@@ -0,0 +1,539 @@
...
+        text = obj.__doc__
+
+        if text is None:
+            return None
+
+        # Remove indentation from the documentation
+        text = inspect.cleandoc(text)
+
+        # Find the first instance of ^: as the start of the meta information.
+        match = re.search("^:(param|type|raise|returns|rtype)", text,
```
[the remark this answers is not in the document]
LonelyCat124: Changed the code to support both now,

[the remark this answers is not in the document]
LonelyCat124: Also one of the tests now uses `:return:`

src/psyclone/docstring_parser.py
LonelyCat124: Yeah - I meant all types docstrings will be included for all parameters but your explanation is better.

LonelyCat124: Updated in both locations.

src/psyclone/docstring_parser.py
LonelyCat124: I had made an assumption but I'm not sure it is so I've removed it. If we don't use it thats good enough for us.

src/psyclone/docstring_parser.py
LonelyCat124: I think no - the `ArgumentData` can be computed from `superclass.apply` and later applied to `class.apply` when inheriting docstrings.

src/psyclone/psyGen.py
LonelyCat124: Added short explanations - if you think they need expanding let me know but I think they're fairly straightforward.

src/psyclone/psyGen.py
LonelyCat124: Added to both.

src/psyclone/psyGen.py
LonelyCat124: I added it to the class now - not sure if its a class or instance variable though but it seems to work.

src/psyclone/psyir/transformations/parallel_loop_trans.py
LonelyCat124: I'm not sure, what would your preference be? Having both `options["verbose"]` and `verbose` would be a bit too much - I'm happy to remove the `options["verbose"]` if you think just documenting the new options would be better (and would fit with deprecation maybe?).

src/psyclone/psyir/transformations/parallel_loop_trans.py
LonelyCat124: Using the one inherited from Transformation.

src/psyclone/tests/psyGen_test.py
LonelyCat124: Each TestTrans is different in the arguments it has I think to test different sections of the code.

src/psyclone/tests/psyir/transformations/loop_trans_test.py
LonelyCat124: Added.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: Added - i think its the right issue (the only one I know of regarding 3.7/3.8 removal).

src/psyclone/tests/psyad/main_test.py
LonelyCat124: Added documentation here, which should guide people towards the new solution as Transformations support it.

src/psyclone/tests/psyGen_test.py
LonelyCat124: Added

src/psyclone/docstring_parser.py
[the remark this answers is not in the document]
arporter: Yes, I think you're right about the line length.

LonelyCat124:
@arporter Just pushed a fix that resolved some documentation being displayed poorly/having too many `\n` inside.

src/psyclone/tests/psyGen_test.py
```diff
@@ -196,9 +196,26 @@ def apply(self, node, valid: bool = True):
...
+            super().apply(node, options=options)
+
+    instance = TestTrans()
+    instance.apply(options={"dict": True})
+    out, err = capsys.readouterr()
+    assert ("PSyclone Deprecation Warning: The options parameters to "
+            "Transformation.apply and Transformation.validate are now "
+            "deprecated. Please use "
+            "the individual arguments, or unpack the options with "
+            "**options. See the documentation for more details." in out)
```
arporter, line 218: s/documentation/Transformations section of the User Guide/

src/psyclone/tests/psyad/main_test.py
```diff
@@ -624,7 +624,20 @@ def test_main_otest_lfric(tmpdir, capsys):
...
     assert error == ""
-    assert output == ""
+    # TODO 2668: This deprecation of options causes this test to see the
+    # deprecation message.
+    correct = '''PSyclone Deprecation Warning: The options parameters to \
+Transformation.apply and Transformation.validate are \
+now deprecated. Please use the individual \
+arguments, or unpack the options with **options. \
+See the documentation for more details.
+PSyclone Deprecation Warning: The options parameters to \
```
arporter, line 634: Seeing this message twice makes me wonder - is it because we're generating it from both `apply` and `validate`? In which case we should probably only do it in `validate` as `apply` should always call that. The other possibility is that psyad uses >1 transformation (which I'm pretty sure it does).

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -583,6 +587,141 @@ def test_omploop_trans_new_options(sample_psyir):
                 in str(excinfo.value))
 
 
+def test_omplooptrans_apply_nowait(fortran_reader, fortran_writer):
```
arporter, line 590: Any idea why this is showing as a new test or is it something you really have just added?

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -302,14 +302,17 @@ def apply(self, node, options=None,
...
         :type node: :py:class:`psyclone.psyir.nodes.Node`
+        :param bool reprod: indicating whether reproducible reductions should
+            be used. By default the value from the config file will be used.
         :param options: a dictionary with options for transformations
                         and validation.
         :type options: Optional[Dict[str, Any]]
-        :param bool options["reprod"]:
-                indicating whether reproducible reductions should be used. \
+        :param bool reprod:
+                whether reproducible reductions should be used. \
```
arporter, line 311: No need for line-continuation char here.

src/psyclone/psyGen.py
```diff
@@ -2727,7 +2727,11 @@ def _find_subclasses(self, module, base_class):
 
 @dataclass
 class ValidOption:
-    '''Class used to specify the valid options dict for a Transformation.'''
+    '''Class used to specify the valid options dict for a Transformation.
+    :param default: The default value for this option.
```
arporter, line 2731: Blank line before first :param please.

arporter (changes requested):
Just minor things now although I have asked to see if we can have LoopTrans.validate call Transformation.validate and that may or may not go well. If it doesn't we can always back out of it.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: I think I had a merge conflict in this file so I suspect its just due to that.

src/psyclone/tests/psyad/main_test.py
LonelyCat124: I think this was coming from validate twice - but yeah validate only is probably sufficient.

LonelyCat124: Apparently it was coming from apply twice - I guess because something is calling Transformation.apply but not Transformation.validate - not sure where this comes from.

doc/user_guide/transformations.rst
```diff
@@ -181,6 +181,46 @@ described in detail in the
...
+
+would be the new possible ways to call the ``apply`` method we defined above,
+as opposed to the old strategy of::
+
+    options = {"option1": 2, "option2: True}
+    mytransformation.apply(node, options=options)
+
+Options dictionaries will continue to remain in the code until they have been
+deprecated for all Transformations, and providing an options dict will
+make the Transformation ignore any other arguments.
```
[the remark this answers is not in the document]
LonelyCat124: Added

src/psyclone/docstring_parser.py
```diff
@@ -150,6 +248,210 @@ def merge(self, other_data, replace_desc: bool = False,
...
             self.returns = other_data.returns
 
+    def gen_docstring(
+            self, indentation: str = "    ", obj: Union[None, Any] = None
+    ) -> str:
+        '''
+        Generates the docstring from the input docdata. The indentation of the
+        docstring can be specified and by default is 4 spaces.
+
+        :param docdata: The docdata object to convert to docstring.
```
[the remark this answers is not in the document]
LonelyCat124: Removed

src/psyclone/docstring_parser.py
```diff
@@ -150,6 +248,210 @@ def merge(self, other_data, replace_desc: bool = False,
                 and other_data.returns is not None):
             self.returns = other_data.returns
 
+    def gen_docstring(
+            self, indentation: str = "    ", obj: Union[None, Any] = None
```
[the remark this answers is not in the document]
LonelyCat124: I've narrowed it down to `Callable` - its not entirely correct as in theory this could be applied to anything that can have a docstring, but its not really designed for that at the moment and I think non-Callable docstringable things odn't need  passing in, and can instead have None passed in with how its currently designed.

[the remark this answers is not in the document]
LonelyCat124: Actually since there are :param: etc. clauses here and obj is used only w.r.t to those I think Callable is fine. I think I will rename obj to function.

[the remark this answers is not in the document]
LonelyCat124: I didn't change it for `create_from_object` because any input there is fine.

src/psyclone/docstring_parser.py
```diff
@@ -150,6 +248,210 @@ def merge(self, other_data, replace_desc: bool = False,
...
+    ) -> str:
+        '''
+        Generates the docstring from the input docdata. The indentation of the
+        docstring can be specified and by default is 4 spaces.
+
+        :param docdata: The docdata object to convert to docstring.
+        :param indentation: The indentation to use. Default is "    ".
+        :param obj: The object who the generated docstring will be for. Default
```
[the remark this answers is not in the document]
LonelyCat124: Added (also refers to function now).

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -327,49 +328,56 @@ def apply(self, node, options=None, verbose: bool = False,
...
             )
             # Rename the input options that are renamed in this apply method.
+            # TODO 2668, rename options to be consistent.
             ignore_dep_analysis = force
             if ignore_dependencies_for is None:
                 list_of_names = []
             else:
                 list_of_names = ignore_dependencies_for
         else:
+            if len(options.keys()) > 0:
```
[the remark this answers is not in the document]
LonelyCat124: Yeah, I've changed that to all go up to Transformation.validate and it causes no issues.

src/psyclone/tests/docstring_parser_test.py
```diff
@@ -473,10 +477,84 @@ def internalerror_docstring():
...
+    basedata2 = DocstringData.create_from_object(docstring_object)
+    assert len(basedata2.arguments) == 3
+    assert list(basedata2.arguments.keys())[0] == "param1"
+    assert list(basedata2.arguments.keys())[1] == "param2"
+    assert list(basedata2.arguments.keys())[2] == "param3"
+    assert len(basedata2.raises) == 1
+    assert basedata2.returns is not None
+    assert basedata2.desc is not None
+
+    # Get the docstring object 2 docdata
```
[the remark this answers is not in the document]
LonelyCat124: Made this clearer.

src/psyclone/tests/docstring_parser_test.py
```diff
@@ -473,10 +477,84 @@ def internalerror_docstring():
...
+    doc2data = DocstringData.create_from_object(docstring_object2)
+
+    # Merge it into basedata
+    basedata.merge(doc2data)
+    assert "param4" in basedata.arguments.keys()
+
+    processed_doc = basedata.gen_docstring()
+
+    # Generate a new basedata from the meregd docstring
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/tests/psyGen_test.py
LonelyCat124: Updated.

src/psyclone/psyir/transformations/omp_loop_trans.py
LonelyCat124: Removed.

LonelyCat124: Removed throughout the file.

src/psyclone/psyGen.py
LonelyCat124: Added

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -294,24 +294,35 @@ def _directive(self, children, collapse=None):
...
+              reprod: bool = None,
+              **kwargs):
         '''Apply the OMPLoopTrans transformation to the specified PSyIR Loop.
 
-        :param node: the supplied node to which we will apply the \
+        :param node: the supplied node to which we will apply the
                      OMPLoopTrans transformation
         :type node: :py:class:`psyclone.psyir.nodes.Node`
-        :param options: a dictionary with options for transformations\
+        :param bool reprod: indicating whether reproducible reductions should
```
arporter, line 305: `reprod` is now documented twice in this docstring :-)

arporter (changes requested):
Thanks Aidan. Very, very nearly there now. Just a few small things to tidy up.

src/psyclone/docstring_parser.py
```diff
@@ -249,18 +251,20 @@ def merge(self, other_data, replace_desc: bool = False,
             self.returns = other_data.returns
 
     def gen_docstring(
-            self, indentation: str = "    ", obj: Union[None, Any] = None
+            self, indentation: str = "    ",
+            function: Union[None, Callable[..., Any]] = None
```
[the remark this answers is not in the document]
LonelyCat124: From what I understood of the type hint docs, the `Callable` type hint needs to have this `[inputs, outputs]` section attached, with `...` being supported for any list of inputs.

src/psyclone/docstring_parser.py
```diff
@@ -324,13 +328,18 @@ def gen_docstring(
         if len(argstrings) > 0:
             docstring += os.linesep
 
-        # Add an empty line between param and raises
-        if len(argstrings) > 0 and len(raisestrings) > 0:
+        # Add an empty line beturn params and returns.
```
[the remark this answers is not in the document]
LonelyCat124: Yea

src/psyclone/docstring_parser.py
```diff
@@ -453,7 +462,8 @@ def create_from_object(cls, obj: Any):
         return docstring_data
 
 
-def create_docstring_data(args: List[str], desc: str, obj: Any) ->\
+def create_docstring_data(args: List[str], desc: str,
+                          function: Callable[..., Any]) ->\
```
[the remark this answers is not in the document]
LonelyCat124: My understanding might also be wrong, but from: https://docs.python.org/3/library/typing.html

> The subscription syntax must always be used with exactly two values: the argument list and the return type. The argument list must be a list of types, a [ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec), [Concatenate](https://docs.python.org/3/library/typing.html#typing.Concatenate), or an ellipsis. The return type must be a single type.

> If a literal ellipsis ... is given as the argument list, it indicates that a callable with any arbitrary parameter list would be acceptable:

Its possible I can just do `Callable` by itself but I wasn't sure from reading it and they only show examples with the type subscription.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -142,19 +142,24 @@ def validate(self, node, options=None, **kwargs):
...
             some symbol names from the dependence analysis checks.
-        :param bool options["sequential"]: whether this is a sequential loop.
-        :param bool options["verbose"]: whether to report the reasons the
+        :param bool sequential: whether this is a sequential loop.
+        :param bool verbose: whether to report the reasons the
             validate and collapse steps have failed.
+        :param bool nowait: whether to add a nowait clause and a
+            corresponding barrier (or equivalent) to enable asynchronous
+            execution.
+        :param bool privatise_arrays: whether to make the write after write
```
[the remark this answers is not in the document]
LonelyCat124: Fixed

src/psyclone/psyir/transformations/omp_loop_trans.py
LonelyCat124: My mistake - I think I was testing something and didn't remove it.

LonelyCat124:
I found an issue with the docstring parser when I added it to another class (in that it failed to understand `:param int | bool name`) so I'll have to try to fix that.

LonelyCat124:
@arporter The examples seem to run locally correctly, so don't hit any of the deprecation warnings.

I changed the docstring parser slightly to allow any number of length of param string, as it needs to handle things like `:param int | bool name` or `:param Union[int, bool] name` as having one type.

arporter (changes requested):
I was all set to give this the green light but the integration tests have shown a problem. It would be good if we could get that import line covered by a standard test.

LonelyCat124:
@arporter I fixed the coverage and tests and integration tests are green now.

src/psyclone/tests/psyGen_test.py
```diff
@@ -253,6 +256,32 @@ def apply(self, node, valid2: int = 1):
...
 
+def test_transformation_get_valid_options_no_sphinx():
+    '''Test that the get_valid_options method behaves in the expected
+    way when sphinx isn't available.'''
+    # Test that importing stringify_annotations works without sphinx.
+    # Trick the import into thinking sphinx.util.typing is unavailable
+    with patch.dict(sys.modules, {'sphinx.util.typing': None}):
+        # Unload the previously imported Transformation class
+        # pylint: disable=import-outside-toplevel
+        del sys.modules['psyclone.psyGen']
```
arporter, line 267: Aha! I've often had this problem in the past (wanting to patch a class that has already been imported) so it's really good to know how it can be done.

arporter (approved):
All good now, thanks Aidan. I've taken the liberty of adding a missing docstring in a test but this is now ready to merge.

Issue #2168, opened 2023-06-09
Array Ranges and SymbolicMaths functions


hiker:
When #1537 is merged. we can reason about array ranges, but it might be required to change how certain functions behave. It might also not be obvious what the right way is.
Example: when are two array ranges equal?

- If they have the same start, stop, step values (and the same signature)? 
- Or maybe (in the context of dependency analysis) it might be more useful to think in terms of 'overlapping' - do these two array ranges overlap (i.e. access the same memory)? That would be `r1 overlap with r2` is `not (r1.stop < r2.start || r1.start > r2.stop)` - assuming that the step value is 1.

Maybe some functions need to be re-implemented (and renamed) to properly describe their behaviour independent of whether the arguments are ranges or not (or a mixture).

PR #2880, opened 2025-02-03
#2723 Add some support for dependency analysis involving array ranges.

### LINKED ISSUE #2723: can_loop_be_parallelised fails if underlying SymbolicMath comparison produce a TypeError
NEMOv5 (-r ORCA2_ICE_PISCES) [mppini.f90](https://forge.nemo-ocean.eu/nemo/nemo/-/blob/main/src/OCE/LBC/mppini.F90)

fails in the `loop_directive_trans.apply(loop)` because it calls the dependency tools `dtools.can_loop_be_parallelised` and this does a symbolic comparison that fails with:
```
  File "psyclone/core/symbolic_maths.py", line 203, in _subtract
    return simplify(sympy_expressions[0] - sympy_expressions[1])
                    ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for -: 'One' and 'tuple'
```

### LINKED ISSUE #2878: [PSyData] Minor extraction library updates
With the kernel extraction started to be used more (e.g. in NG-ARCH), I am in the process of integration this with our Baf build system (which is the lfric Fab build system). That will involve copying the PSyclone NetCDF LFRic extraction files into the build system (so they can be injected into the Fab build, so no need for creating libraries etc). A few changes required:
- [ ] Minor updates to debug output (using `trim` to remove white spaces)
- [ ] LFRic's real field types have been renamed (`field_r{{prec}}_mod` to `field_real{{prec}}_mod` etc). We can't simply switch this, since atm our infrastructure is outdated (see #2623 for details, and also #2239 - best solution is to switch to the now stand-alone lfric infrastructure files, which should support non-MPI builds now). As a quick solution till this is properly fixed, I'll add a second copy of the jinja files with the new names.

hiker:
Fixes #2723. I am not sure if more support is needed or would be useful (e.g. I could check if two ranges have the same stride, but accesses different elements, ...).

### DIFF
--- a/changelog
+++ b/changelog
@@ -50,6 +50,9 @@
 	17) PR #2829 for #2826. Adds a --config command-line argument to the
 	PSyAD script (to replicate that for psyclone).
 
+	18) PR #2280 for #2723. Add support for depencency analysis involving
+	array ranges.
+
 release 3.0.0 6th of December 2024
 
 	1) PR #2477 for #2463. Add support for Fortran Namelist statements.
--- a/src/psyclone/psyir/tools/dependency_tools.py
+++ b/src/psyclone/psyir/tools/dependency_tools.py
@@ -48,7 +48,7 @@ from psyclone.core import (AccessType, Signature, SymbolicMaths,
 from psyclone.errors import InternalError, LazyString
 from psyclone.psyir.backend.sympy_writer import SymPyWriter
 from psyclone.psyir.backend.visitor import VisitorError
-from psyclone.psyir.nodes import Loop
+from psyclone.psyir.nodes import Loop, Node, Range
 
 
 # pylint: disable=too-many-lines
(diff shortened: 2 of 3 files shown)


hiker:
CI was green. Note that I have some typos in the issue number - I mixed up 2878 and 2723, so e.g. the CI shows 2878, not 2723 :(

sergisiso (changes requested):
Thanks to tackling this @hiker, the implementation, docstring and tests all look great. The only comment I have is that when looking for TODOs in the same file I found a reference to #2168 (L401) which I am wondering if it's the same issue and can now also be fixed/closed. But I will approve regardless. Let me know what you think.

hiker:
> Thanks to tackling this @hiker, the implementation, docstring and tests all look great. The only comment I have is that when looking for TODOs in the same file I found a reference to #2168 (L401) which I am wondering if it's the same issue and can now also be fixed/closed. But I will approve regardless. Let me know what you think.

I had a look. That issue is concerned about array-ranges that involve the loop variable that is to be parallelised, e.g.:
```
do i=1, n
   a(2*i:2*i+1)  = 0
enddo
```
(stupid example, but you get the idea). The moment the loop variable is involved, this new range comparison is not called at all. I looked if we could simply call this new function here:

Now we could do a similar test to the rest of 'one index that depends on the loop` part of the dependency analysis: if there is an overlap between the array ranges for two loop iterations `i1` and `i2` the loop cannot be parallelised. Following the current implementation, that is done by computing `d=i2-i1`, the distance between the loop iterations for which the indices would overlap. If this has only 0 as solution for `d`, the loops can be parallelised, because there exist no other loop iteration that accesses the same range.

That leads to a systems of inequalities (I hope that's the right word), which simplifies to:
``` -0.5 <=d <= 0.5```
So, no integer solution except 0 -> the loop can be parallelised. But you can see that this is not at all trivial, and does require more work than what we now have (atm any array range with the loop index will just be rejected).

So, #2168 is not fixed here. But if you agree that array ranges depending on loop indices do not happen in any realistic code we need to handle (or will in general be not parallelisable anyway, so the current behaviour of rejecting it is good enough), we can close #2168, and replace the #TODO with just a comment for 'the future in case we need to look at it again'?

sergisiso (approved):
@hiker I am ok with keeping the issue, I just wanted to check if it was something easy to solve with the new functionality but I agree is not. This was my only comment, this is ready to merge.

PR #2891, opened 2025-02-10
(Closes #898) add general support for sym table remove

### LINKED ISSUE #898: SymbolTable.remove() method should check for references to target symbol
The SymbolTable has the remove() method to remove a symbol. This is also used by the swap() method introduced in #887. With the exception of ContainerSymbols, this routine currently does not check that the target symbol is referred to by other Symbols or PSyIR. e.g. by an array declaration or an initialisation or by a `Reference` .

arporter:


arporter:
Unfortunately, `remove` is called when updating a Routine and this happens if we `add` a new Routine node into the tree. This then results in the DA being run and if that fails, we get a CodeBlock. This means that the *whole* of:
```
program test_prog
             integer :: i, h(10)
             real :: a, g(10)
             g(h(i)) = a
end program test_prog
```
ends up in a CodeBlock whereas previously, we had 0 CodeBlocks.

PR #2893, opened 2025-02-11
(Closes #2561) Add support for allocatables in the HoistLocalArraysTrans

### LINKED ISSUE #2561: Add support for hoisting allocate statements
Allocates within compute code are OK on the CPU but when adding directives to accelerate the code on a GPU, these allocates are left on the CPU and then (with managed memory) copied over to the GPU leading to a lot of unnecessary data traffic. Ideally we should be able to transform these allocates in a similar manner to what we do for `automatic` arrays. (i.e. make them module scope if they aren't already and then only reallocate them if their size changes).

sergisiso:


sergisiso:
I merged #3037 in this branch, so this PR is blocked until that one is merged

sergisiso:
@arporter This is ready for review

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -223,26 +273,46 @@ def apply(self, node, options=None):
...
-        # Finally, remove the hoisted symbols (and any associated tags)
-        # from the routine scope.
-        for sym in automatic_arrays:
+                            [Reference(sym)])]),
+                    index=0
+                )
+
+            if already_allocatable:
+                # Find and remove the deallocate statements
+                for ref in node.walk(Reference):
```
arporter: Might be more efficient to search for IntrinsicCall's first and then check their arguments? Probably not much in it though now that I've written that.

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -154,17 +153,73 @@ def apply(self, node, options=None):
...
-            sym.visibility = Symbol.Visibility.PRIVATE
+            # Check if the original is already an ALLOCATABLE variable
+            already_allocatable = any(dim == ArrayType.Extent.DEFERRED
+                                      for dim in sym.shape)
+
+            # Find or Create the array reference that will be the argument to
+            # the new memory allocation statement.
+            if already_allocatable:
+                # If it was already an allocatable, we should be able to find
+                # it
```
arporter: s/it/the allocate statement./

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -223,26 +273,46 @@ def apply(self, node, options=None):
...
-
-        # Finally, remove the hoisted symbols (and any associated tags)
-        # from the routine scope.
-        for sym in automatic_arrays:
+                            [Reference(sym)])]),
+                    index=0
+                )
+
+            if already_allocatable:
+                # Find and remove the deallocate statements
```
arporter: s/the/any/

src/psyclone/psyir/transformations/hoist_trans.py
```diff
@@ -217,6 +217,9 @@ def _validate_dependencies(self, statement, parent_loop):
             else:
                 read_only_sigs.append(sig)
 
+        # Make sure abs_positions are caches, otherwise this will be extremely
```
arporter: s/caches/cached/

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
```diff
@@ -602,6 +604,173 @@ def test_validate_tagged_symbol_clash(fortran_reader):
             str(err.value))
 
 
+def test_apply_with_hoist_with_dependent_symbols(fortran_reader,
+                                                 fortran_writer):
+    ''' Check that we host declarations with dependent symbols succeeds
```
arporter: s/we host/the hoist of/

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
```diff
@@ -602,6 +604,173 @@ def test_validate_tagged_symbol_clash(fortran_reader):
...
+        "  do i=1,10\n"
+        "    a(i) = 1.0\n"
+        "  end do\n"
+        "end subroutine test\n"
+        "end module my_mod\n")
+    psyir = fortran_reader.psyir_from_source(code)
+    routine = psyir.walk(Routine)[0]
+    hoist_trans = HoistLocalArraysTrans()
+    hoist_trans.apply(routine)
+    assert "real(kind=wp), dimension(10) :: a" in fortran_writer(psyir)
```
arporter: Do we/could we attach comments to the declarations of variables that we can't hoist? Would that be too verbose perhaps?

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
```diff
@@ -602,6 +604,173 @@ def test_validate_tagged_symbol_clash(fortran_reader):
...
+    assert "type(my_type), dimension(N) :: b" in fortran_writer(psyir)
+
+
+def test_apply_with_allocatables(fortran_reader, fortran_writer, tmpdir):
+    ''' Test the apply method correctly handles an automatic arrays with
```
arporter: Strictly speaking, they're not 'automatic' if they're allocatable.

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
```diff
@@ -602,6 +604,173 @@ def test_validate_tagged_symbol_clash(fortran_reader):
...
+  public
+
+  contains
+  subroutine test(arg, var)
+    real, allocatable, dimension(:), intent(in) :: arg
+    integer, intent(in) :: var
+    integer :: i
+    real, allocatable, dimension(:) :: d
+    real, allocatable, dimension(:) :: e
+
```
arporter, line 741: What happens if someone has made a mistake and there's still a local, allocatable array declared but which is never used or allocated. Do we cope with that OK?

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/hoist_trans.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
sergisiso: Agreed

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
sergisiso: In that case we leave it alone, I added an explicit example now.

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
sergisiso: Done for this "dependent symbol" cases

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -350,9 +347,19 @@ def _get_local_arrays(node):
             # statement, not the hoisted declaration
             if isinstance(sym.datatype.intrinsic, DataTypeSymbol):
```
arporter, line 348: Here, it would be better to use `sym.reference_accesses()`, just in case e.g. the precision corresponds to an expression (e.g. I think it's valid to have `2*wp`). This would also avoid having to consider the precision and the intrinsic type separately. Should the comment at L344 say "which are not global" rather than "local"?

sergisiso: Good catch but:

1) This would return the dimension symbols as well, which are not a problem because they stay inisde the scope (in the allocate statement). But we could do `sym.precision.reference_accesses`.

2) Precision sometiems returns an enumerator so it needs to be guarded by a similar conditon, 

3) Even solving the two points above, "2*wp" is lost. I started solving this but also precision currently can only be a Symbol and the fix was becoming too big, I so I added a TODO to fix this.

sergisiso:
@arporter Trying to address your last comment exposed a passthrough bug that we need to solve, so for now I just added a TODO. The linkcheck was failing for password protected links - don't ask me why we didn't fail on these before. Ready for another look.

arporter (commented):
Thanks Sergi, all looks good now.
I'll fire off the integration tests to be on the safe side.

arporter (approved):
Will proceed to merge.

Issue #2892, opened 2025-02-10
[LFRic] User-defined DoF kernels should work with field vectors


TeranIvy:
Support for user-defined kernels that operate on DoFs is available in PSyclone 3.0.0. However, there is a limitation on them to not accept field vectors. That is an issues as a lot of PSyKAl-lite code that refers to this functionality uses coordinate fields which are field vectors. 

Historically, this limitation may have arosen from inability of PSyclone built-ins to deal with field vectors, but it is not required for user-defined kernels in LFRic. This issue removes the limitation and adds the relevant code generation.

arporter:
#2897 is merged.

Issue #1010, opened 2020-11-25
Generate the PSy-layer with PSyIR Fortran backend


sergisiso:
Currently the workflow of a psyclone (generate.py) is:
```
1) create PSy (using API factory to choose __init___)
      |-> create (__init__) Invokes
             |->  create (__init__) Invoke
                       |-> create part of the Schedule (using nodes.__init__'s and Call/BuiltInFactories)

2) execute script (psy -> psy):
      def trans(psy):
             s = psy.invokes[0].invoke.schedule
             modify(s)
             return psy

3) Generate f2pygen 
f2pygen = psy.gen() # Generates remaining Schedule (e.g. Loop bounds) and Symbols, lowers PSyIR and DSL concepts to f2pygen

4) Generate code
code = str(f2pygen)
```

There are two main issues here:

1) One issue is that `psy.gen()` still generates part of the PSy-layer after the PSyclone script has been ran, this means:
  - psy.gen() generation is coupled to the f2pygen generation and the logic can not be re-used by PSyIR back-ends
  - There are no opportunity to modify the code/symbols produced by psy.gen()

2) The second issue is the the generation of Fortran is hardcoded into generate.py. There should be a way to choose the PSy-layer target (by command-line, config, transformation, ...)

I think the solution to the second issue is easier, we just have to agree in the way to choose the language and do:
```
if f2pygen:
     code = str(psy.gen())
elif fortranbackend:
    generic_psyir = psy.lower()
    code = FortranWriter(psy)
elif cppbackend:
    generic_psyir = psy.lower()
    bindings_layer, psy_layer  = generic_psyir.create_bindings_layer()
    code1 = FortranWriter(bindings_layer)
    code2 = CppWriter(psy_layer)
endif
```

The first issue is more complex because we have to decouple what is generation of PSy-layer (which imo should be all done before the script) from the lowering of DSL to generic concepts (which should be done after `trans(psy)` (note we can have an optional trans_generic(psy)) step if we need)(see #851 ). I think resolving this issue can be done as we go along and hit the issues in each node.

sergisiso:
Note that this changes may be disruptive (which makes me uncomfortable as there are a few open questions yet).  So I think it would be sensible to start creating the proposed interface (if we agree) but keep the changes inside a transformation until we are more confident with the solution. Something like:

```
1) create PSy (using API factory to choose __init___)
      |-> create (__init__) Invokes
             |->  create (__init__) Invoke
[...]

PR #2897, opened 2025-02-13
#2892 Allow user-defined DoF kernels to act on field vectors

mo-alistairp:


TeranIvy:
Changing the label as there is another PR for the release and Andy may agree to do this one.

src/psyclone/domain/lfric/lfric_kern_metadata.py
```diff
@@ -614,20 +614,19 @@ def _validate_operates_on_dof(self, need_evaluator):
...
-                    f" This is not permitted in the LFRic API.")
-
         # Check function spaces are the same
         # list out all function spaces
         arg_fs_names = set()
         for arg in self._arg_descriptors:
             for fs_name in arg.function_spaces:
                 arg_fs_names.add(fs_name)
+
+        # Previously the field vector test fails before this
```
arporter, line 624: I'm afraid I don't understand this bit. We should still be able to check that there's only one distinct function space, even for a field vector.

src/psyclone/tests/test_files/dynamo0p3/testkern_dofs_mod.f90
```diff
@@ -42,26 +42,30 @@ module testkern_dofs_mod
...
-     type(arg_type), dimension(5) :: meta_args =          &
+     type(arg_type), dimension(6) :: meta_args =          &
           (/ arg_type(gh_field, gh_real, gh_write, w1),   &
              arg_type(gh_field, gh_real, gh_read,  w1),   &
              arg_type(gh_field, gh_real, gh_read,  w1),   &
              arg_type(gh_field, gh_real, gh_read,  w1),   &
-             arg_type(gh_scalar, gh_real, gh_read)        &
+             arg_type(gh_scalar, gh_real, gh_read),       &
+             arg_type(gh_field*3, gh_real, gh_read,  &
+              ANY_SPACE_1)                                &
```
arporter, line 52: I think this additional field vector argument *must* be on the same space as the other fields, i.e. `w1` in this case.

src/psyclone/tests/test_files/dynamo0p3/testkern_dofs_mod.f90
```diff
@@ -42,26 +42,30 @@ module testkern_dofs_mod
...
 contains
 
-  subroutine testkern_dofs_code(a, b, c, d, scalar_arg)
+  subroutine testkern_dofs_code(a, b, c, d, scalar_arg, field_vec)
     implicit none
 
     real(kind=r_def),    intent(inout) :: a
     real(kind=r_def),    intent(in)    :: b, c, d
     real(kind=r_def),    intent(in)    :: scalar_arg
+    type(field_type),    intent(in)    :: field_vec(3)
```
arporter, line 68: A kernel must contain essentially Fortran77 code - i.e. plain arrays rather than objects. Just as `a`, `b`, `c` and `d` are real scalars, this field vector argument should also be three scalars.

src/psyclone/tests/domain/lfric/dofkern_test.py
```diff
@@ -341,7 +321,8 @@ def test_multi_invoke_cell_dof_builtin(tmpdir, monkeypatch, annexed, dist_mem):
             output = (
                 "      DO df = loop0_start, loop0_stop, 1\n"
                 "        CALL testkern_dofs_code(f1_data(df), f2_data(df), "
-                "f3_data(df), f4_data(df), scalar_arg)\n"
+                "f3_data(df), f4_data(df), scalar_arg, field_vec_1_data, "
```
arporter, line 324: The three field-vector components must also be indexed by dof, e.g. `field_vec_1_data(df)`.

src/psyclone/tests/domain/lfric/dofkern_test.py
```diff
@@ -341,7 +321,8 @@ def test_multi_invoke_cell_dof_builtin(tmpdir, monkeypatch, annexed, dist_mem):
...
+                "field_vec_2_data, field_vec_3_data)\n"
                 "      END DO\n"
                 "      !\n"
                 "      ! Set halos dirty/clean for fields modified in the "
```
arporter, line 328: I would guess this test does not pass when compilation is enabled (`pytest --compile domain/lfric/dofkern_test.py`)?

arporter (changes requested):
Thanks for this Alistair. I'm afraid it needs some more work as the code being generated is not correct currently - please see inline.

src/psyclone/tests/test_files/dynamo0p3/testkern_dofs_mod.f90
TeranIvy: @mo-alistairp, please have a look at this science kernels that has an example of field vector for coordinate fields (`chi`): https://code.metoffice.gov.uk/trac/lfric/browser/LFRic/trunk/components/science/source/kernel/geometry/sci_calc_detj_at_w2_kernel_mod.F90

src/psyclone/domain/lfric/lfric_kern_metadata.py
mo-alistairp: You're right - I must have had a bug at the time which failed this test and I misinterpretted as being the tests fault. I've reintroduced the test.

src/psyclone/tests/test_files/dynamo0p3/testkern_dofs_mod.f90
mo-alistairp: Corrected this

src/psyclone/tests/domain/lfric/dofkern_test.py
mo-alistairp: Corrected the code generation to index the field vector components.

mo-alistairp:
Addressed all comments and added both the code generation and documentation.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -358,14 +358,29 @@ def field_vector(self, argvect, var_accesses=None):
...
-            self.psyir_append(Reference(cmpt_sym))
-            text = cmpt_sym.name
+            if self._kern.iterates_over == "dof":
+                # If dof kernel, add access to the field by dof ref
+                dof_sym = self._symtab.find_or_create_integer_symbol(
+                    "df", tag="dof_loop_idx")
+
+                self.append_array_reference(cmpt_sym.name,
+                                            [Reference(dof_sym)],
+                                            ScalarType.Intrinsic.INTEGER,
```
arporter, line 377: I've realised that this isn't correct (a field will typically be REAL but we do support other types too) but I can see that it isn't correct in the `field` handler either. However, in Sergi's big PR #2834 he's removed the need to pass this type information entirely so I think we leave it as is for now. If you could add a TODO for #1010 both here and at the equivalent call in the field handler (L413) that will cover it.

arporter (changes requested):
Thanks @mo-alistairp, looking good now. There's just a couple of TODO's required to flag that there's an issue that @sergisiso will fix in his PR for #1010. I've fired off the integration tests.
Have you tried this in anger in a version of LFRic modified to make use of this new functionality?

src/psyclone/domain/lfric/kern_call_arg_list.py
mo-alistairp: I've added the TODOs you requested. I haven't had an opportunity to test on a modified LFRic yet.

arporter (approved):
All good now, thanks.
I'll proceed to merge but note that this will need testing against a suitably modified LFRic prior to us making a release.

PR #2899, opened 2025-02-14
2889 extract non alloc arrays

hiker:
Adds support for non-allocatable arrays in the driver extraction. I had to add a new generic interface with `NonAlloc` appended to the name. It adds support for `is_allocatable` for some types (as much as I needed), but only for arrays.


hiker:
I just brought it up to recent master, assuming that the tests passes, it's ready for review.

lib/extract/standalone/lfric/read_kernel_data_mod.f90
```diff
@@ -243,6 +284,43 @@ subroutine ReadArray1dChar(this, name, value)
 
     end subroutine ReadArray1dChar
 
+    ! -------------------------------------------------------------------------
+    !> @brief This subroutine reads the values of a 1D array of character(*)
+    !! that is not allocatable (e.g. a fixed size array).
+    !! @param[in,out] this The instance of the extract_PsyDataType.
+    !! @param[in] name The name of the variable (string).
+    !! @param[out] value An allocatable, unallocated 2d-double precision array
```
arporter, line 292: This also needs updating.

lib/extract/standalone/read_kernel_data_mod.jinja
```diff
@@ -258,6 +263,61 @@ contains
 
     end subroutine ReadArray{{dim}}d{{name}}
 
+    ! -------------------------------------------------------------------------
+    !> @brief This subroutine reads the values of a {{dim}}D array of {{type}}
+    !! that is not allocatable (e.g. a fixed size array).
+    !! @param[in,out] this The instance of the extract_PsyDataType.
+    !! @param[in] name The name of the variable (string).
+    !! @param[out] value An allocatable, unallocated 2d-double precision array
```
arporter, line 271: This is the bit that needs updating.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -291,6 +291,21 @@ def intrinsic(self):
             return self.partial_datatype.intrinsic
         return None
 
+    @property
+    def is_allocatable(self) -> bool:
+        '''If we have enough information in the partial_datatype,
+        determines if this data type is allocatable, and False otherwise.
```
arporter, line 297: I think we have to be careful here to distinguish the case where we don't know from that where we do know that it's not allocatable. In other places we return `None` to indicate that we don't know something.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -291,6 +291,21 @@ def intrinsic(self):
...
+        :returns: whether this UnsupportedFortranType is known to be
+            allocatable. Note that atm PSyclone only supports allocatable
+            arrays.'''
+        if (self.partial_datatype and
+                isinstance(self.partial_datatype, ArrayType)):
+            return self.partial_datatype.is_allocatable
+        return False
+
 
 class ScalarType(DataType):
```
arporter, line 310: Should we put a default implementation in `DataType` that just returns `False`? Otherwise, a caller will first have to check that they have either an ArrayType or an UnsupportedFortranType.

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
```diff
@@ -492,8 +492,18 @@ def _create_output_var_code(name, program, is_input, read_var,
...
                 post_tag = f"{name}{postfix}"
         name_lit = Literal(post_tag, CHARACTER_TYPE)
-        BaseDriverCreator.add_call(program, read_var,
-                                   [name_lit, Reference(post_sym)])
+
+        if sym.is_array and not sym.datatype.is_allocatable:
+            # In case of a non-allocatable array (e.g. a constant
+            # size array from a module), call the ReadVariable
+            # function that does not require an allocatable field
+            BaseDriverCreator.add_call(program, read_var+"NonAlloc",
```
arporter, line 500: I should know this but confess I can't remember: for arrays that are declared in modules within LFRic, does the driver just populate their values?

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
```diff
@@ -620,8 +630,20 @@ def _sym_is_field(sym):
             else:
                 sym = symbol_table.lookup_with_tag(str(signature))
                 name_lit = Literal(str(signature), CHARACTER_TYPE)
-            self.add_call(program, read_var,
-                          [name_lit, Reference(sym)])
+
+            # TODO #2898: the test for array can be removed if
+            # `is_allocatable` is supported for non-arrays.
```
arporter, line 635: We could add a default implementation that just returns False but that would have the downside that all the Scalar bindings would then have to have "NonAlloc" added to their names. (Or, we retain the check that it is an array too.)

src/psyclone/tests/psyir/symbols/datatype_test.py
```diff
@@ -415,6 +417,8 @@ def test_arraytype_unsupportedtype():
     # Since no partial datatype is provided, these return None
     assert utype.partial_datatype is None
     assert utype.intrinsic is None
+    # Test the allocatable flag
+    assert not utype.is_allocatable
```
arporter, line 421: As suggested in the implementation, I think this should return None in this case (c.f. the `intrinsic` property).

src/psyclone/tests/psyir/symbols/datatype_test.py
```diff
@@ -816,6 +820,30 @@ def test_unsupported_fortran_type_eq():
             UnsupportedFortranType("save :: blue_blood"))
 
 
+def test_unsupported_fortran_type_is_allocatable(fortran_reader):
+    '''Test the copy() method of UnsupportedFortranType.'''
```
arporter, line 824: :-)

src/psyclone/tests/psyir/symbols/datatype_test.py
```diff
@@ -816,6 +820,30 @@ def test_unsupported_fortran_type_eq():
...
+      type(some_type), pointer :: var(nelem), var2(start:stop)
+      type(some_type), target, allocatable :: var_alloc(:)
+    end subroutine
+    '''
+    psyir = fortran_reader.psyir_from_source(code)
+    routine = psyir.walk(Routine)[0]
+    vsym = routine.symbol_table.lookup("var")
+    # Make sure we do indeed test the UnsupportedFortranType
+    assert isinstance(vsym.datatype, UnsupportedFortranType)
+    assert not vsym.datatype.is_allocatable
```
arporter, line 838: I suggest we want `is False` here.

arporter (changes requested):
Nice job Joerg, pleasingly small change. It's up for debate but I suggest we could add an `is_allocatable` to the base `DataType` class and have it return False. I also think that for UnsupportedFortranType it should return `None` if we don't know.

lib/extract/netcdf/read_kernel_data_mod.f90
```diff
@@ -285,6 +334,49 @@ subroutine ReadArray1dChar(this, name, value)
 
     end subroutine ReadArray1dChar
 
+    ! -------------------------------------------------------------------------
+    !> @brief This subroutine reads the values of a 1D array of character(*)
+    !! that is not allocatable (e.g. a fixed size array)
+    !! @param[in,out] this The instance of the extract_PsyDataType.
+    !! @param[in] name The name of the variable (string).
+    !! @param[out] value An allocatable, unallocated 2d-double precision array
```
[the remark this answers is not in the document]
hiker: Oops - that error has been in the jinja templates for some time :) Fixed.

lib/extract/netcdf/read_kernel_data_mod.jinja
```diff
@@ -323,6 +328,81 @@ contains
 
     end subroutine ReadArray{{dim}}d{{name}}
 
+    ! -------------------------------------------------------------------------
+    !> @brief This subroutine reads the values of a {{dim}}D array of {{type}}
+    !! that is not allocatable (e.g. a fixed size array).
+    !! @param[in,out] this The instance of the extract_PsyDataType.
+    !! @param[in] name The name of the variable (string).
+    !! @param[out] value An allocatable, unallocated 2d-double precision array
```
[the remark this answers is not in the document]
hiker: Fixed.

lib/extract/standalone/lfric/read_kernel_data_mod.f90
hiker: Fixed.

lib/extract/standalone/read_kernel_data_mod.jinja
hiker: Fixed.

src/psyclone/psyir/symbols/datatypes.py
hiker: Fair point, I return None now.

src/psyclone/psyir/symbols/datatypes.py
hiker: I thought we agreed not to do this for now, but I've done it now.

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
hiker: Yes, their values will be stored in the kernel data file and so the arrays will be initialised.

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
hiker: I left the tests the way there are, mainly otherwise this patch would become quite large, since all tests would be affected.

src/psyclone/tests/psyir/symbols/datatype_test.py
hiker: Fixed.

src/psyclone/tests/psyir/symbols/datatype_test.py
hiker: Aeh ... we love copy&paste.

src/psyclone/tests/psyir/symbols/datatype_test.py
hiker: Done.

hiker:
I've triggered the CI as well. Ready for next review.

lib/extract/netcdf/read_kernel_data_mod.f90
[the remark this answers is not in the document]
arporter: Ah, I hadn't noticed that error. The one I was talking about was the fact that this says that it is an 'allocatable' when in fact it isn't.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -294,17 +300,18 @@ def intrinsic(self):
     @property
     def is_allocatable(self) -> bool:
         '''If we have enough information in the partial_datatype,
-        determines if this data type is allocatable, and False otherwise.
+        determines whether this data type is allocatable or not.
+        If it is unknown, it will return None. Note that atm PSyclone
+        only supports allocatable arrays.
```
arporter, line 305: Took me a while to get to grips with this statement (because I thought "but PSyclone supports static arrays too"). Perhaps add emphasis to "arrays"?

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -294,17 +300,18 @@ def intrinsic(self):
     @property
     def is_allocatable(self) -> bool:
```
arporter, line 301: `Union[bool, None]` please.

src/psyclone/tests/psyir/symbols/datatype_test.py
```diff
@@ -79,6 +79,12 @@ def test_unresolvedtype_str():
     assert str(data_type) == "UnresolvedType"
 
 
+def test_unresolvedtype_is_allocatable():
+    '''Test that the UnresolvedType class' is_allocatable property works.'''
+    data_type = UnresolvedType()
+    assert data_type.is_allocatable is False
```
arporter, line 85: If it's unresolved, we don't know whether it's allocatable so this should be `None`?

src/psyclone/tests/psyir/symbols/datatype_test.py
```diff
@@ -835,7 +850,7 @@ def test_unsupported_fortran_type_is_allocatable(fortran_reader):
     vsym = routine.symbol_table.lookup("var")
     # Make sure we do indeed test the UnsupportedFortranType
     assert isinstance(vsym.datatype, UnsupportedFortranType)
-    assert not vsym.datatype.is_allocatable
+    assert vsym.datatype.is_allocatable is False
```
arporter, line 853: Please add a check that this returns `None` if `partial_datatype` is None.

lib/extract/standalone/read_kernel_data_mod.jinja
arporter: You need to fix the 'allocatable' bit too :-)

arporter (changes requested):
Thanks Joerg. Unfortunately the error you fixed was not the error I spotted so that needs addressing :-) Also, the docs are flagging a warning within indentation in the new `is_allocatable` method.
Once those things are fixed this is ready to go.

lib/extract/netcdf/read_kernel_data_mod.f90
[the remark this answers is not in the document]
hiker: Ha, that's so funny :) Fixed.

lib/extract/standalone/read_kernel_data_mod.jinja
hiker: Fixed ... again

src/psyclone/psyir/symbols/datatypes.py
hiker: I reworded it a bit, and added emphasis.

src/psyclone/psyir/symbols/datatypes.py
hiker: Oops.

src/psyclone/tests/psyir/symbols/datatype_test.py
hiker: Good point. I was originally thinking of even making the default (in `Datatype`) to return None, but then we get all scalars returned as unknown, unless I'll fix that as well. I left that with a TODO :) So, UnresolvedType now returns None.

src/psyclone/tests/psyir/symbols/datatype_test.py
hiker: Done.

hiker:
CI also triggered. Assuming that it passes, ready for next review.

src/psyclone/psyir/symbols/datatypes.py
arporter: Oh yes, that helps. Thanks.

arporter (approved):
All good now, thanks Joerg.
Integration tests were all happy.
Will proceed to merge.

PR #2907, opened 2025-02-25
2878 extraction update

### LINKED ISSUE #2878: [PSyData] Minor extraction library updates

hiker:
Fixed #2878 (and other minor improvements to output format to make it easier to read).


hiker:
This makes the extraction comparison more readable (using ints instead of floating point format), fixes a but or two, and also includes the a compiled and executed test of the comparison library.

.github/workflows/compilation.yml
```diff
@@ -109,6 +109,18 @@ jobs:
         module load hdf5/${HDF5_VERSION} netcdf_c/${NETCDF_C_VERSION} netcdf_fortran/${NETCDF_FORTRAN_VERSION}
         module load cuda/${CUDA_VERSION}
         F90=gfortran make -C examples compile
+    - name: Libraries with compilation - gfortran
```
sergisiso: Remove "with" here and below?

doc/user_guide/psyke.rst
```diff
@@ -606,28 +606,33 @@ The columns from left to right are:
 .. only:: latex or has_dvipng
```
sergisiso, line 607: Can you make the distinction between latex and html only for the last two bullet points, and remove the duplication on the remaining ones?

lib/extract/Makefile
```diff
@@ -52,16 +52,29 @@ PSYDATA_LIB_DIR ?= ./..
 # the listed data types and scalar and 2-dimensional arrays. Note that
 # domain-specific implementation will create their own version with different
 # number of dimensions if required.
```
sergisiso: 2-dimensional to n-dimensional.

There is no more "domain-specific implementation" of the extraction driver right? Could this be reworded as "The type and n-arity of the comparable type may need to be adjusted to match the capabilities of the target application using the arguments below."? If that's the case, can we also remove the sentence at the top that mentions domain-specific libraries.

Also, since this can potentially be changes, can we make it a `?=` assingment.

lib/extract/Makefile
```diff
@@ -52,16 +52,29 @@ PSYDATA_LIB_DIR ?= ./..
...
+.PHONY: all compare test
 	
 compare: compare_variables_mod.o
 
+test_compare: compare_variables_mod.o test_compare.o
+	$(F90) $(F90FLAGS) $^ -o $@
+
+make test: test_compare
+	# Check all the counts only, to avoid that floating point
+	# differences invalidate this test.
```
sergisiso: What do you mean by "invalidate this test"? Isn't that supported by the asterisks in the grep?

lib/extract/netcdf/lfric/Makefile
```diff
@@ -71,7 +71,7 @@ LFRIC_INCLUDE_FLAGS += $$(nf-config --fflags)
 PSYDATA_LIB_NAME = _extract
 PSYDATA_LIB = lib$(PSYDATA_LIB_NAME).a
 
-PROCESS_ARGS = -prefix=extract_ -types=char,int,logical,real,double \
```
sergisiso, line 74: Is this whole folder needed anymore? The type and dims are the same as the parent Makefile, the jinja files have the implementation with `%get_proxy() ` but the lowering version may not need this? and for the prefix we could be consistent with the generic. Is there something else?

If it can be removed, maybe for a separate PR.

lib/extract/test_compare.f90
```diff
@@ -0,0 +1,44 @@
...
+    a_int( 7:10) = 2001000000
+    a_int(11:15) = 2100000000
+
+    call compare_init(3)
+    call compare("a_dbl", a_dbl, a_dbl_correct)
+    call compare("a_single", a_single, a_single_correct)
+    call compare("a_int", a_int, a_int_correct)
+    call compare_summary()
+
+end program test_compare
```
sergisiso: Add a linebreak at the end of this line

.github/workflows/compilation.yml
sergisiso: Also can you move these steps before the "Examples with compilation" steps. Since some of the tests in the later depend on the these, it will be easier to make it fail before so the examples only fail if the generated code is wrong.

sergisiso (changes requested):
@hiker Just some minor comments, I like the new extraction stats table, it's easier to quickly see if there is important differences imo.

lib/extract/Makefile
```diff
@@ -52,16 +52,29 @@ PSYDATA_LIB_DIR ?= ./..
...
-PROCESS_ARGS = -types=char,int,long,logical,real,double -dims=2
+PROCESS_ARGS = -types=char,int,long,logical,real,double -dims=1,2,3,4
 PROCESS = $$($(PSYDATA_LIB_DIR)/get_python.sh) $(PSYDATA_LIB_DIR)/process.py
 
-.PHONY: all compare
+.PHONY: all compare test
 	
 compare: compare_variables_mod.o
 
+test_compare: compare_variables_mod.o test_compare.o
```
sergisiso, line 63: Oh and gitignore the generated test_compare binary

doc/user_guide/psyke.rst
hiker: No. I tried originally (and there is a comment), and I just tried again (who knows ... software updates??), but if I move them out I get warnings for either latex or html builds about:
```
exception while evaluating only directive expression: chunk after expression
```

lib/extract/Makefile
hiker: Done.

lib/extract/netcdf/lfric/Makefile
hiker: I've opened #3044 for that. It affects quite a few files and documentation.

.github/workflows/compilation.yml
hiker: Good point, done.

lib/extract/Makefile
hiker: Both done.

lib/extract/Makefile
hiker: I've reworded the comment, I hope it's better. I meant: we don't compare the full output (which includes floating point numbers for max absolute and relative errors, l2 norm etc), only the count, which should make the test more robust when running with different compilers and/or different hardware.

lib/extract/test_compare.f90
hiker: Done.

hiker:
Looks like I forgot to mark this as ready. I've brought it up to latest master, and triggered the CI (which should not really be affected by the changes here).

sergisiso (approved):
@hiker All comments have been addressed, this is ready to merge if the integration tests come green.

sergisiso:
@hiker Unfortunately the libraries compilation failed with:
```
./test_compare | grep "a_dbl    *15 *1 *2 *3 *4 *5" > /dev/null
At line 147 of file compare_variables_mod.F90 (unit = 6, file = 'stdout')
Fortran runtime error: Missing comma between descriptors
(A8' ',6(I12, ' '),4(E12.7,' '))                                                
     ^
```

hiker:
I think I got it. I'll trigger the full CI tomorrow.

hiker:
I've fixed the conflict in the changelog, I don't think there's a need to rerun the CI because of this.

PR #2910, opened 2025-02-28
(closes #2896) Adds IncreaseRankLoopArraysTrans

### LINKED ISSUE #2896: Increase array rank to eliminate outer loop dependencies for temporary arrays
This is similar to #2671, there are cases that in order to parallelise a outer loop we need to make a temporary array that all the iterations of a loops uses private so that no dependencies are reported between the usage in each iteration. #2671 focused on doing it using the "private" clause of the directive, but another alternative is increasing the rank of the work array so that each iteration uses a different location.

This is useful in both NEMO and LFRic.

In NEMO the work arrays and loops are in the same file (e.g. src/OCE/ZDF/zdftke.F90)

```fortran
real(kind=wp), dimension(nlay_i) :: ztmp
! We want to parallelise this loop
do ji = 1, npti, 1
  do jk = 1, nlay_i, 1
     ztmp(jk) = 3
  enddo
  ! In the code this loops cannot be fused, otherwise we could just do it and scalarise ztmp
  do jk = 1, nlay_i, 1
     field(jk,jj) = ztmp(jk)
  enddo
end do
```
The proposed transformation will take a symbol and a loop and increase the rank of the symbol declaration with the bounds of the loop and each reference inside the loop with the loop iteration variable. No references can be found outside the loop. The resulting code will be:

```fortran
real(kind=wp), dimension(nlay_i, 1:npti) :: ztmp
! We want to parallelise this loop
do ji = 1, npti, 1
  do jk = 1, nlay_i, 1
     ztmp(jk, ji) = 3
  enddo
  ! In the code this loops can not be fused, otherwise we could just do it and scalarise ztmp
  do jk = 1, nlay_i, 1
     field(jk, ji) = ztmp(jk, ji)
  enddo
end do
```

In LFRic is a bit more complicated because there is a call between the loop and the loop body with the temporaries are local arrays inside the kernel:

```fortran
do cell = 1, num_cells
  call kernel(..., map_field1(:,cell), map_field2(:,cell))
enddo

subroutine kernel(...)
  real, dimension(ndf1) :: x_e
	
   do k = 0, nlayers-1
     do df = 1, ndf2
       x_e(df) = x(map2(df)+k)
     end do
     ... uses x_e and cannot be fused&scalarized ...
   enddo
   
end subroutine kernel
```
So array (x_e) and the loop (over cells) are in different locations. We have 2 alternatives:
- Move the local work array to the module scope, pass the cell as an argument, apply the transformation with the loop and the references to update in different scopes (seems dangerous).
- convert the local array to an argument. Apply the rank-increasing transformation just at the call site.

The second alternative seems easier/safer, the resulting code should look:
```fortran
real, dimension(1:num_cells, ndf1) :: x_e

do cell = 1, num_cells
    call kernel(..., map_field1(:,cell), map_field2(:,cell), x_e(:, cell) )
enddo

subroutine kernel(...)
   real, dimension(ndf1), intent(in) :: x_e

   do k = 0, nlayers-1
     do df = 1, ndf2
       x_e(df) = x(map2(df)+k)
     end do
     ... uses x_e and can not be fused&scalarized ...
   enddo
   
end subroutine kernel
```

sergisiso:


src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Authors: S. Siso, STFC Daresbury Lab
+
+'''This module contains the IncreaseRankLoopArrays transformation. This
+transformation increases the dimensionality of the selected arrays with
+the length of the loop iteration space, when it updates each array access
```
arporter: I don't quite follow what "when it updates each array access..." means?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+        enddo
+      enddo
+    <BLANKLINE>
+    end program test
+    <BLANKLINE>
+
+    '''
+    def __str__(self):
+        return ("Increases the Rank of the supplied arrays by the iteration "
+                "space of the given loop, and update all its references")
```
arporter: "...update all references to those arrays"

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def validate(self, node, options=None):
```
arporter: Please convert this to use keyword arguments instead of options.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+        ''' Checks that the supplied node is a valid target.
+
+        :param node: target PSyIR node.
+        :type node: :py:class:`psyclone.psyir.nodes.Loop`
+        :param options: a dictionary with options for transformations.
+        :type options: Optional[Dict[str, Any]]
+
+        :raises TransformationError: if the node is not a Loop.
+        :raises TransformationError: if the node is not inside a Routine.
+        :raises TransformationError: the array option has not been provided.
```
arporter: "if no array names are provided."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+        :raises TransformationError: the given array name  or the
+            symbol is not local or not an array.
+        :raises TransformationError: the array is references inside a
```
arporter: "if any of the arrays are referenced inside a CodeBlock."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+                raise TransformationError(
+                    f"{self.name} provided 'arrays' must be a local array "
+                    f"symbols, but '{array}' is not")
+
+            if array.name.lower() in codeblock_symbols:
+                raise TransformationError(
+                    f"{self.name} does not support arrays that are referenced "
+                    f"inside a Codeblock, but '{array.name}' is inside one.")
+
+            # TODO: What should happen with references outside the loop?
```
arporter: Yes, I was going to ask the same question :-) Presumably we should reject these, at least as a first cut?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+
+        if not isinstance(node, Loop):
+            raise TransformationError(
+                f"The target of the {self.name} transformation should be a "
+                f"Loop, but found '{type(node).__name__}'.")
+
+        routine = node.ancestor(Routine)
+        if routine is None:
+            raise TransformationError(
+                f"The target of the {self.name} transformation should be a "
```
arporter: To make it clearer, how about "The target Loop of the ....must be inside a Routine."?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+                f"Loop inside a Routine.")
+
+        # Capture all symbols used inside codeblocks, this are not permited
```
arporter: "these" and "permitted"

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+        array_list = options.get("arrays", [])
+        if len(array_list) == 0:
+            raise TransformationError(
+                f"{self.name} has a mandatory 'arrays' option that "
+                f"needs to be provided to inform what arrays needs their "
+                f"rank increased.")
+        for array in array_list:
+            if isinstance(array, str):
+                array = node.scope.symbol_table.lookup(array, otherwise=None)
+            if array is None:
```
arporter: Is silently ignoring symbols that we can't find the right way to go? I guess it makes it easier in the transformation script so perhaps it's OK. I'm just worried it might make script debugging difficult if someone misspells an array name.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+                continue
+
+            if not array.is_automatic or not array.is_array:
+                raise TransformationError(
+                    f"{self.name} provided 'arrays' must be a local array "
```
arporter: "...must be local.."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
arporter: The alternative is to apply this transformation to a Routine/Schedule rather than a Loop and have it add suitable loops around the outside of every access to the array that has had its rank increased?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+                    f"symbols, but '{array}' is not")
+
+            if array.name.lower() in codeblock_symbols:
+                raise TransformationError(
+                    f"{self.name} does not support arrays that are referenced "
+                    f"inside a Codeblock, but '{array.name}' is inside one.")
+
+            # TODO: What should happen with references outside the loop?
+
+    def apply(self, node, options=None):
```
arporter: Please update to use **kwargs.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,197 @@
...
+
+        array_list = options.get("arrays", [])
+
+        for array in array_list:
+            if isinstance(array, str):
+                array = node.scope.symbol_table.lookup(array, otherwise=None)
+            if array is None:
+                continue
+
+            # TODO: check that the bound expressions are valid as static
```
arporter: Should this be a 'towards' PR? Alternatively, please add checks, e.g. that any Symbols referenced in the start/stop expressions are not written before the target loop.

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -184,6 +190,24 @@ def validate(self, node, options=None):
...
+                    if hasattr(dtype, "intrinsic"):
+                        if dtype.intrinsic == ScalarType.Intrinsic.CHARACTER:
+                            message = (
+                                f"OpenMP Target cannot enclose a region that "
+                                f"uses characters, but found: "
+                                f"{datanode.debug_string()}"
+                            )
+                            if verbose:
+                                node.preceding_comment = message
+                            raise TransformationError(message)
```
arporter, line 206: Please update the method docstring with this new :raises:

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -161,18 +162,23 @@ def validate(self, node, options=None):
             attempts to enclose the assingment setting the return value.
         '''
         device_string = options.get("device_string", "") if options else ""
+        strings = options.get("allow_strings", False) if options else False
```
arporter, line 163: Please document these new options. I'll let you off the **kwargs for this one.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -0,0 +1,130 @@
...
+import pytest
+from psyclone.psyir.nodes import Loop
+from psyclone.psyir.transformations import IncreaseRankLoopArraysTrans
+from psyclone.transformations import TransformationError
+
+
+def test_trans_name_and_str():
+    ''' Check that the name and str method works as expected. '''
+    trans = IncreaseRankLoopArraysTrans()
+    assert trans.name == "IncreaseRankLoopArraysTrans"
```
arporter: You don't need this as `.name` just calls the base-class implementation.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+def test_irla_validate(fortran_reader):
+    ''' Check that the validate method works as expected. '''
+    trans = IncreaseRankLoopArraysTrans()
+
+    psyir = fortran_reader.psyir_from_source("""
+        program test
+            real, dimension(10) :: a
+            integer :: i = 3
+            a(i) = 3
```
arporter, line 66: `validate` should always fail for this code as we won't touch the reference of `a` in this assignment?

arporter: Please add tests for accesses before and after the target loop. Alternatively, as mentioned previously, perhaps this transformation should target a whole scoping region in which the array is local?

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -0,0 +1,130 @@
...
+    assert ("IncreaseRankLoopArraysTrans provided 'arrays' must be a local "
+            "array symbols, but 'i: " in str(err.value))
+
+    with pytest.raises(TransformationError) as err:
+        trans.apply(routine.children[1], options={'arrays': ['a']})
+    assert ("IncreaseRankLoopArraysTrans does not support arrays that are "
+            "referenced inside a Codeblock, but 'a' is inside one."
+            in str(err.value))
+
+    # Referencing non-existan arrays is fine, they are just ignored, this
```
arporter: Please add this information to the docstring of the transformation.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -0,0 +1,130 @@
...
+         integer :: N=10, M=10
+         integer :: i, j
+         real, dimension(N) :: ztmp
+         do i = -5, M+3
+             do j = 1, N
+                 ztmp(j) = 1
+             end do
+             do j = 1, N
+                 ztmp(j) = ztmp(j) + 1
+             end do
```
arporter, line 208: What would happen if we had e.g. `ztmp(:) = 0` somewhere inside the `i` loop?

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -0,0 +1,130 @@
...
+         end do
+     end program
+    """)
+
+    trans.apply(psyir.walk(Loop)[0], options={'arrays': ['ztmp']})
+    code = fortran_writer(psyir)
+
+    # The declaration and references have been updated
+    assert "real, dimension(n,-5:m + 3) :: ztmp" in code
```
arporter, line 217: We need a separate test for a case where e.g. `m` is written to before the `i` loop.

arporter (changes requested):
Thanks Sergi. I have some concerns about the new transformation as it stands - it will be very easy to make it generate code that does not compile. We perhaps need to discuss whether it should target a whole scoping region rather than a single loop? Either way, it needs some extra validation checks and associated tests.

We will also need to update the User Guide with information on this transformation.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: The "when" was a "then", but actually, I prefer to document this on the class docstring, not the module. As that is the part that ends up in the documentation. So I deleted most of this.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Applying it to a loop is very convinient because it also provide the iteration space (size of the new rank) and iteration variable (to use as indexing value inside the loop). I will instead add a new loop around accesses outside the selected loop (the NEMO this are mostly init statement e.g `ztmp(:) = 0`, which now need to become `ztmp(:,:) = 0` - but with explicit loops for adding directives).

This initialistion statements are a very common pattern in NEMO and something I need to fix anyway to apply this transformation to more places.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Fixed

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: I agree with this. If a name is given it must exist.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done (I don't think we can currently guarantee this, but I tried a best-effort check). I think this PR should close the issue.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: I provided a different solution, let me know what you think. In this case this example should pass the validate.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: I think one of the new tests covers this

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: That should be alright, I added this statement in one test.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: This is not the case anymore

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: Removed

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
             raise TransformationError(
-                f"The target of the {self.name} transformation should be a "
-                f"Loop inside a Routine.")
-
-        # Capture all symbols used inside codeblocks, this are not permited
+                f"The target Loop of the {self.name} transformation must be "
+                f"inside a Routine.")
+
+        # Check if the loop bound expressions are static
+        values_to_check = set()
```
arporter, line 153: Comment pls, e.g. "First, find all Symbols accessed in the loop bounds."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
+                    node.stop_expr.walk(Reference)):
+            if ref.symbol.is_static:
+                continue
+            if isinstance(ref.parent, IntrinsicCall):
+                # The routine name of a intrinsic, or the arguments of inquiry
+                # functions don't need to be check
+                if ref.parent.is_inquiry or ref.position == 0:
+                    continue
+            values_to_check.add(ref.symbol)
+        for assignment in routine.walk(Assignment):
```
arporter, line 166: "# Second, check that none of these Symbols are assigned to within the Routine."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
+            if isinstance(assignment.lhs, Reference):
+                if assignment.lhs.symbol in values_to_check:
+                    raise TransformationError(
+                        f"{self.name} can only be applied to loops with static"
+                        f" loop bound expressions, but in has been attempted "
```
arporter: s/in/it/

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
+                        f"in a loop with the variable "
+                        f"'{assignment.lhs.symbol.name}'."
```
arporter: ...' which is assigned to: {assignment.debug_string().strip()}'."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
             for name in cb.get_symbol_names():
                 codeblock_symbols.add(name.lower())
 
         # Each item listed in the array list must be a local Array Symbol or a
         # string that resolves to it
-        array_list = options.get("arrays", [])
-        if len(array_list) == 0:
+        if not isinstance(arrays, list) or len(arrays) == 0:
             raise TransformationError(
                 f"{self.name} has a mandatory 'arrays' option that "
```
arporter: "...that is required to specify which arrays are to have their rank increased."

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
                 raise TransformationError(
-                    f"{self.name} provided 'arrays' must be a local array "
+                    f"{self.name} provided 'arrays' must be local array "
                     f"symbols, but '{array}' is not")
 
             if array.name.lower() in codeblock_symbols:
                 raise TransformationError(
                     f"{self.name} does not support arrays that are referenced "
                     f"inside a Codeblock, but '{array.name}' is inside one.")
+            # TODO: check that the bound expressions are valid as static
```
arporter: This one is done now?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -99,28 +98,30 @@ class IncreaseRankLoopArraysTrans(Transformation):
...
-        :raises TransformationError: the array is references inside a
-            CodeBlock.
+        :raises TransformationError: if any of the arrays are referenced inside
+            a CodeBlock.
 
         '''
-        if options is None:
-            options = {}
-
         if not isinstance(node, Loop):
```
arporter, line 140: I'm still getting to grips with this myself but I *think* validate should just have the **kwargs argument and then call `get_option` and `validate_options` - see https://psyclone.readthedocs.io/en/latest/developer_guide/transformations.html#moving-to-the-new-transformations-options

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
-        :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
+        :param node: target Loop node.
+        :param arrays: list of arrays that will have the rank increased.
 
         '''
-        self.validate(node, options)
-
-        array_list = options.get("arrays", [])
+        self.validate(node, arrays, **kwargs)
```
arporter: Probably this needs to be `arrays=arrays`?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
-        for array in array_list:
+        for array in arrays:
             if isinstance(array, str):
                 array = node.scope.symbol_table.lookup(array, otherwise=None)
-            if array is None:
-                continue
 
-            # TODO: check that the bound expressions are valid as static
-            # expressions in the declarations.
             array.shape.append(
```
arporter, line 255: Comment pls - "Add an additional dimension to this array with the same bounds as the target Loop"

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -33,26 +33,25 @@
...
     import TransformationError
 
 
 class IncreaseRankLoopArraysTrans(Transformation):
     ''' This transformation takes a loop and a list of arrays accessed inside
     the loop, and increases those arrays with an additional dimension with the
-    size of the interation space.
+    size of the interation space. Then it indexes all accesses with the loop
+    variable, so that each iteration accesses a unique location. Effectively
+    making the sub-array private for each iteration of the loop.
```
arporter, line 58: Please also document the fact that it modifies all accesses outside the target Loop too. Please could you extend the example in the docstring to show this.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
             array.shape.append(
                 ArrayType.ArrayBounds(node.start_expr, node.stop_expr)
             )
 
-            for ref in node.walk(ArrayReference):
+            for ref in array.find_symbol_table(node).node.walk(ArrayReference):
                 if ref.symbol is array:
-                    ref.addchild(Reference(node.variable))
+                    if ref.is_descendent_of(node):
+                        # Inside the target loop index the reference to the
```
arporter: s/to/using/

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -129,69 +130,101 @@ def validate(self, node, options=None):
...
+                        # loop variable
+                        ref.addchild(Reference(node.variable))
+                    else:
+                        # Outside the target loop index the whole range
```
arporter, line 266: What about if the scoping region contains two explicit loops that both use the same array (before its rank is increased)? I think this will end up transforming the accesses inside a second loop to use full slices. We should probably catch and reject this case in `validate`.

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -147,11 +147,17 @@ def validate(self, node, options=None):
         :type options: Optional[Dict[str, Any]]
         :param str options["device_string"]: provide a compiler-platform
             identifier.
+        :param str options["allow_strings"]: prermit OMP target regions
```
arporter: "permit"

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
arporter: I quite like the new solution but, as commented earlier, I think it needs to be made a bit more robust.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -80,25 +79,63 @@ def test_irla_validate(fortran_reader):
...
+    """)
+    routine = psyir.children[0]
+    with pytest.raises(TransformationError) as err:
+        trans.apply(routine.walk(Loop)[0], arrays=['a'])
+    assert ("IncreaseRankLoopArraysTrans can only be applied to loops with "
+            "static loop bound expressions, but in has been attempted "
+            "in a loop with the variable 'variable'." in str(err.value))
+
+    # This is fine
+    trans.apply(routine.walk(Loop)[2], arrays=['a'])
```
arporter: Actually, I suspect it isn't - please check the output.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -121,10 +158,47 @@ def test_irla_apply(fortran_reader, fortran_writer):
...
+    trans = IncreaseRankLoopArraysTrans()
+    trans.apply(psyir.walk(Loop)[1], arrays=['ztmp'])
+    code = fortran_writer(psyir)
+    # Check the ztmp accesses outside the target loop
+    assert "ztmp = 1" in code  # This already indexes the whole array
+    assert "ztmp(:,:) = ztmp(:,:) + 2" in code
+    assert "ztmp(j,:) = ztmp(j,:) / 3" in code
+    # And the two inside the loop
+    assert "ztmp(j,i) = ztmp(j,i) + 4" in code
+    assert "ztmp(:,i) = 5" in code
```
arporter, line 257: Please add a compilation check.

arporter (changes requested):
Thanks Sergi, I'm a lot happier now. I like the new direction although I think some of the validation needs a bit more tightening. I think the use of `**kwargs` also needs a bit of work.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Yes (I forgot to delete it)

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: The output looks good to me, what is the concern?

sergisiso: Oh did you mean because the ArrayBound missed the dangling_parent to avoid the "linebreak" in the declaration? (I now moved the dangling parent one step down to make sure ArrayBound always have them regardless of how they are created)

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: After exploring a bit I think I prefer the current error messages which are more clear about arrays being a mandatory argument and what it does. Maybe validate_options should be the bare minimum for type checking but could be replaces by more context-aware custom errors?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Explicit is good but I think then the arguent name is the same as the argument variable it feels overly verbose, so I have not added this. Unless you really disagree.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: > I think this will end up transforming the accesses inside a second loop to use full slices.

Yes it will. But I think this is the right behaviour, see last example with the initialisation test. We essentially trade more parallelism (independent elements), by duplicating this elements outside the loop.

sergisiso:
@arporter This is also ready for another look

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -645,12 +666,15 @@ def __post_init__(self):
                     f"ArrayBounds must be either ArrayType.Extent.ATTRIBUTE or"
                     f" an instance of DataNode but got "
                     f"'{type(self.upper).__name__}'")
+            # setattr necessary to bypass frozen dataclass restrictions
+            object.__setattr__(self, 'lower', _dangling_parent(self.lower))
```
arporter, line 670: Good idea moving this inside the class.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
arporter: I'm happy to keep the specific checks that you already have but I think we still then need a following call to `validate_options` since that checks for any unsupported options supplied by the caller. Is that correct @LonelyCat124 ? Also, we need to change the signature of `validate` so that it just has **kwargs. So, I think we'd call `get_option` to get the "arrays" option, then do our specific checks with explicit error messages, then call `validate_options`.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
arporter: I think it's safer in case other arguments get added in future (unlikely in this case). I see you've made the change anyway so thanks :-)

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
arporter: I think what I meant (but didn't say) was a case where we'd want to do the *same* thing to more than one explicit loop in the scoping region. As written, I think we can only apply this transformation to a single loop and all other accesses will get changed to use full slices. Specifically, if the example code in the transformation docstring had a *second* loop nest over `i`:
```
    >>> code = ("""
    ... program test
    ...     integer :: N=10, M=10
    ...     integer :: i, j
    ...     real, dimension(N) :: ztmp
    ...     ztmp(:) = 0
    ...     do i = -5, M + 3
    ...         do j = 1, N
    ...             ztmp(j) = 1
    ...         end do
    ...         do j = 1, N
    ...             ztmp(j) = ztmp(j) + 1
    ...         end do
    ...     end do
    ...     do i = ...
    ...       do j = 1, N
    ...           ztmp(j) = 5
    ...           etc.
    ... end program
    ... """)

File: src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py (hunk 33 above)
Comment (arporter) [reply src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py]: No, I didn't mean that (although thanks for fixing). The output is:
```fortran
  variable = 3
  do j = 1, variable, 1
    do i = 1, 10, 1
      a(i,:) = 3
    enddo
  enddo
  do j = SIZE(variable), constant + imported, 1
    do i = 1, 10, 1
      a(i,j) = 3
    enddo
  enddo
```
So the first loop now writes to `a(i,:)` every trip which isn't great. My concern (from the earlier discussion) is that we can't currently make that a write to `a(i,j)`, even if we wanted to.

arporter: N.B. the `fortran_writer` fixture is currently unused in this test (but it might be good to extend the test so that it explicitly checks for the expected output, once we've agreed what that should be ;-)).

arporter (changes requested):
Thanks Sergi. I think we need to agree what we want this transformation to do if there's more than one loop containing assignments to an array that we wish to increase the rank of.
Apart from that, we also need to agree with @LonelyCat124 what we should be doing with `validate`.
Finally (probably after we've reached agreement), the User Guide needs updating so that it mentions this transformation.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: > So the first loop now writes to a(i,:) every trip which isn't great.

@arporter The example may be a bit unfortunate but the original first loop already writes to `a(i)` multiple time:
```
            do j = 1, variable
                do i = 1, 10
                   a(i) = 3
                end do
            enddo
```
so the transformed code just does the same but for the whole slice `a(i,:)`

arporter:
The only potential improvement I can think of is to perhaps create a new, temporary array rather than extend the size of the existing one. That way, we could copy (in parallel) into it before the loop we are transforming and leave all other uses alone. This assumes the array doesn't have a DUC that extends outside of the target loop but perhaps that's already a condition of the transformation?

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Ok, I think I added the right steps now.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Ok, I partially came around on this since our discussion and now I am much more restrictive in that the validate only allows simple initialisation expressions "ztmp(..) = ..." outside the loop. As evaluating the correctness of other cases was very complex. I think with this I don't need the warning in the docstring, as now I am more confident that the semantics do not change. And the previous condition already blocks applying it two times, as it will find the array usage in the other loop unless they are simple write-only loops, in which case is also semantically equivalent. Let me know what you think.

sergisiso:
@arporter This is ready for another look, instead of the proposed scattering to a new temporary, I made the current implementation much more restrictive and only allow simple 'write-only' expressions outside the loop, which effectively makes it not possible to apply to two loops that "use" the arrays, as each will complain about the other one. It also prevents rank-changes of expressions outside the loop which would have created invalid code. Let me know what you think of this.

examples/nemo/scripts/utils.py
```diff
@@ -332,10 +334,12 @@ def normalise_loops(
         operate on arrays to explicit loops (currently only maxval).
     :param bool convert_range_loops: whether to convert ranges to explicit
         loops.
-    :param bool hoist_expressions: whether to hoist bounds and loop invariant
-        statements out of the loop nest.
     :param scalarise_loops: whether to attempt to convert arrays to scalars
         where possible, default is False.
+    :param increase_array_ranks: whether to increase the rank of selected
+        arrays.
+    :param bool hoist_expressions: whether to hoist bounds and loop invariant
```
arporter: You don't need the "bool" here.

examples/nemo/scripts/utils.py
```diff
@@ -410,6 +417,62 @@ def normalise_loops(
     # top level. This would allow the collapse clause to be applied.
 
 
+def increase_rank_and_reorder_nemov5_loops(routine: Routine):
+    ''' This method increases the rank of temporary loops used inside selected
```
arporter: s/loops/arrays/

examples/nemo/scripts/utils.py
```diff
@@ -410,6 +417,62 @@ def normalise_loops(
...
+    loops (in order to parallelise the outer loop without overlapping them)
+    and the rearranges the outer loop next to the inner ones (in order to
```
arporter: "and then..."

examples/nemo/scripts/utils.py
```diff
@@ -410,6 +417,62 @@ def normalise_loops(
...
+    collapse them), so that more parallelism can be leverage. This is useful
+    in GPU contexts, but it increases the memory footprint and may not be
+    benefitial for caching-architectures.
```
arporter: "beneficial"

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,266 @@
...
+            if isinstance(check, CodeBlock):
+                for name in check.get_symbol_names():
+                    codeblock_names.add(name.lower())
+            if isinstance(check, Reference):
+                if check.is_descendent_of(node):
+                    # These are fine because inside the loop we will add an
+                    # index and the resulting expression rank will be the same
+                    continue
+                if (isinstance(check.parent, Assignment) and
+                        check.position == 0):
```
arporter: I found this quite hard to understand. Perhaps `... and check is check.parent.lhs` would be clearer? i.e. we're saying that if the array appears anywhere other than on the LHS of an assignment outside of the target loop then we forbid it? Unless I'm missing something, this doesn't seem any/much stricter than what we had before? I guess I might understand more when I get to the tests.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,266 @@
...
+                    # Initialisation statements are fine, because the apply
+                    # we will extend the initialisation to the whole new rank
+                    continue
+                # Everything else is currently forbiden
```
arporter: "forbidden"

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -0,0 +1,266 @@
...
+    def apply(
+        self,
+        node: Loop,
+        arrays: Optional[list[Union[Symbol, str]]] = None,
+        **kwargs
+    ):
+        '''Applies the transformation.
+
+        :param node: target Loop node.
+        :param arrays: list of arrays that will have the rank increased.
```
arporter: s/the/their/

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
arporter: I may well be missing something here but I don't think the new implementation conforms to what you say here?

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
```diff
@@ -0,0 +1,223 @@
...
+    assert ("IncreaseRankLoopArraysTrans can only be applied to loops with "
+            "static loop bound expressions, but it has been attempted "
+            "in a loop with the variable 'variable' which is assigned to: "
+            "'variable = 3'." in str(err.value))
+
+    # This is fine, it can safely add a new rank to keep each a(i) inside the
+    # second loop independent (note that each a(i) in the first loop will still
+    # be overwritted as in the original code)
+    trans.validate(routine.walk(Loop)[2], arrays=['a'])
+
```
arporter, line 156: Please could you add another test showing cases where there are unsupported accesses to the target array outside the target loop and that `validate` rejects these.

arporter:
Another question, is the small performance improvement:
image
as expected? (I thought it would be larger.)

sergisiso:
> Another question, is the small performance improvement: as expected? (I thought it would be larger.)

To comment on this, the loops that this optimisation targets were already significanly improved by the array_privatisation option I added some time ago. That transformation already allowed the outer loop to be offloaded and avoided the data traffic which was the biggest performance penalty.

The IncreaseRankLoopArraysTrans by itself just does the same, with approx the same perf results. The advantage of this new transformation is that it can be followed up by bringing the outer-loop next to the inner loop (done in utils.py), which allows to collapse them and have more parallelism to distribute the loop. This is the benefit that you see in the chart above but is smaller than avoiding the data copies that we already had.

In the nemo sct it was a big win because they got both benefits at once.

There is a second benefit (again not for the transformation itself but by enabling the follow up loop reordering), which is that moving outer-loop down makes the condition which sometimes have characters end up outside the offloaded region. For some reason the compiler was fine with this string comparisons on the GPU, but I am more confident with them being outside.

examples/nemo/scripts/utils.py
sergisiso: Deleted

examples/nemo/scripts/utils.py
sergisiso: Done

examples/nemo/scripts/utils.py
sergisiso: Done

examples/nemo/scripts/utils.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
sergisiso: I updated to the suggested code.

> we're saying that if the array appears anywhere other than on the LHS of an assignment outside of the target loop then we forbid it? Unless I'm missing something, this doesn't seem any/much stricter than what we had before?

Anywhere other than "the top expression of the lhs of an array assignment". This is significanly more restrictive because it means that the only expressions outside the target loop that are supported are array assignments, and by the semantics of the language this can only write (repeatedly with an implicit loop is we miss some index), the same value in all possitions. This bypasses all problems I could think regarding reading from the wrong dimension or passing to another function with an array with a different rank, etc. Also prevent applying it to two loops that "use" the array, which I think was your previous concern.

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
sergisiso: I already had this in the test above with `variable = sum(map(int(a + 3)))`, but maybe you were looking for the case with two explicit loops, which I have now also added. Let me know if there is something in particular that you want to see.

sergisiso:

> When you say "simple initialisation expressions" did you mean "the target array is only assigned to and not read from"? If you could add a couple of tests for cases that validate() should reject that would help me. Probably also good to add this to (perhaps a second) example in the docstring of the class.

I see that using "initialisation" was confusing, I have updated the text regarding this and added a new validation test. I also modified the docstring example to be more explicit about this, however it is difficult to add a second (essentially different) docsting example because this is pretty much all that is premitted now.

src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
arporter: OK, I've got it now. Thanks :-)

src/psyclone/tests/psyir/transformations/increase_rank_loop_arrays_test.py
arporter: Oh yes, so you did (and I was getting confused by your use of 'array assignment' which I took to mean 'assignment to the whole of an array') but I understand now and the new test helps.

Issue #2909, opened 2025-02-26
[LFRic] setval_random and setop_random do not work with redundant computation


arporter:
As discussed in #2805, `setval_random` cannot correctly perform redundant computation because there's no way of computing what pseudo-random values would be assigned to dof locations by neighbouring processors. This limitation applies both to annexed and halo dofs and therefore means that we cannot have `COMPUTE_ANNEXED_DOFS=True` in the configuration file.

I propose to put a check for this into PSyclone which will cause it to stop if it encounters either of the named kernels and `COMPUTE_ANNEXED_DOFS` is `True`. It will also refuse to apply the RedundantComputation transformation to loops containing these kernels.

@TeranIvy and @DrTVockerodtMO, I presume these kernels are *only* used in the adjoint test harness? (They were never intended for rigorous scientific use.)

DrTVockerodtMO:
By kernels do you mean whenever `setval_random` is invoked? In which case yes I believe this is only used by the `adjoint_tests` app.

I like the idea of refusing the redundant transformation calculation on `setval_random` but am less sure about the check, I'll explain why. Essentially the check will cause the exporting of `adjoint` itself to fail when used outside of the `adjoint_tests` app. This is how the current building behaviour works:

1. The `adjoint` model is imported, and its `import.mk` script is ran.
2. The `import.mk` script runs the PSyAd command, which generates both the adjoint kernels and the adjoint test algorithms (even if the latter is not used).
3. The project working directory is PSycloned since otherwise the generated adjoint tests would not be PSycloned.

If we want to proceed in the way you suggest, then we would need to be careful as any project that imports the `adjoint` model will fail with the otherwise valid global `psyclone.cfg` script, merely because the adjoint tests were generated along with exporting the `adjoint` model (even if they won't be used).

I could create a dual behaviour in which the `adjoint` import optionally builds the adjoint tests (with the default set to false), and then the PSyAd command can check the config option only if the adjoint test is slated to be generated. I opted not to do this initially so that the adjoint build scripts were less complicated for when they eventually get ported to Fab.

arporter:
Yes, I mean whenever it is invoked. If it is invoked and COMPUTE_ANNEXED_DOFS is True then we will get wrong answers. The only safe way to handle this is to abort - if someone were to (ill advisedly) start using `setval_random` somewhere other than in the Adjoint testing, then that too would silently not work correctly. This wouldn't be picked up by a check in the `psyad` driver since psyad would not be used.
[...]

PR #2919, opened 2025-03-12
(Towards #2909) forbid redundant computation setval/op_random and rename Dynamo0p3RedundantComputationTrans

arporter:
A  small PR that does precisely that. I've also renamed the associated transformation. However, for reviewing simplicity and to avoid clashes with other, big PRs, I've not done any moving of the implementation or the tests.

### DIFF
--- a/doc/developer_guide/APIs.rst
+++ b/doc/developer_guide/APIs.rst
@@ -929,7 +929,7 @@ Modifying the Schedule
 ----------------------
 
 Transformations modify the schedule. At the moment only one of these
-transformations - the ``Dynamo0p3RedundantComputationTrans`` class in
+transformations - the ``LFRicRedundantComputationTrans`` class in
 ``transformations.py`` - affects halo exchanges. This transformation can
 mean there is a requirement for new halo exchanges, it can mean
 existing halo exchanges are no longer required and it can mean that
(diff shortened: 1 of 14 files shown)


arporter:
This is ready for a first look now. In the interests of load-balancing the reviews, it could be one for @hiker or @TeranIvy perhaps?

arporter:
@TeranIvy might have a view on how I've done this - I rely on kernel names. Since one is a Builtin I think that is ~OK but the other is an LFRic user-supplied kernel. The alternative would be to have some extra metadata to flag that a kernel cannot perform redundant computation (including on annexed dofs).

TeranIvy:
@arporter, I'm happy to take this for review. The only major remark that I have is that this change will make the next release non-backward compatible in LFRic as the optimisation scripts will change. I we're going down this way then I would suggest we make a concentrated effort to update all class names, file paths etc from dynamo to LFRic and be done with it.

arporter:
That's a good point about the transformation renaming. I'm happy to revert that (but will hold off for now). We should have a concerted effort to fix all of these before the next non-backwards compatible release.

arporter:
In fact, shall we pause this one while we address the metadata change required?

TeranIvy:
> In fact, shall we pause this one while we address the metadata change required?

I think this would be a good idea, @arporter :)

Issue #2534, opened 2024-04-02
PDFs missing on readthedocs


hiker:
There are no more pdf versions of our manuals on readthedocs.
@arporter mentioned:
```
I think it's because we've switched to using `.readthedocs.yaml` configuration files and have missed adding that option.
```

sergisiso:
If we make readthedocs build the pdf documentation again, should we point the README reference to it instead of having the psyclone.pdf inside the repository? It's an output file which git doesn't track that well and we often don't keep it up to date with the last merge.

sergisiso:
We generate a pdf in readthedocs:

<img width="515" height="557" alt="Image" src="https://github.com/user-attachments/assets/4b3f0488-2262-4253-a8e5-d011f24b4e34" />

PR #2920, opened 2025-03-13
#2534 Try to trigger creation of the PDF version on readthedocs again.

hiker:
@arporter , @sergisiso - I've created a PR for #2534, but I don't know if this can be tested without merging it to master. I don't think I have access to the readthedocs account (nor do I know what would be possible if I had :) ).

If we get the PDFs to work again, I would agree (as Sergi suggested) that pointing to the readthedocs version of the user guide would be better, esp. since it saves us the trouble of having to manually rebuild it all the time

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,5 +1,8 @@
 version: 2
 
+formats:
+    - pdf
+
 python:
   version: "3.8"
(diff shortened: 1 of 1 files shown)


hiker:
Sure, thanks!!!

PR #2944, opened 2025-04-01
(Closes #2909) Add OWNED_DOF and OWNED_CELL_COLUMN operates_on and forbid redundant computation

### LINKED ISSUE #2909: [LFRic] setval_random and setop_random do not work with redundant computation

arporter:


arporter:
This is ready for a first review now. As it is LFRic-specific functionality it would be good if @TeranIvy or @hiker did the review.

arporter:
This is needed for the release so would @hiker or @sergisiso mind picking it up? Thanks :-)

doc/user_guide/lfric.rst
```diff
@@ -1961,15 +1961,27 @@ operates_on                    Data passed for each field/operator argument
 ``cell_column``                Single column of cells from the owned region (except
                                when performing an INC operation on continuous fields
 			       when it will include one level of halo cells).
+``owned_cell_column``          As ``cell_column`` but the iteration space must *only*
+                               cover owned columns (i.e. excluding halo columns).
```
sergisiso: As discussed, I find this hard to understand, probably because the "cell_column" is outdated but there are other problems. I feel the "owned_cell_column" description just repeats what I already know from its name and doesn't give me useful information. But also I think it is trying to descrive two things at the same time, the data passed to the kernel, and the looping used in the psy-layer for this kernel. Maybe it will be good to separate them?

For example:
```
operates_on: "cell_column"
Data provided for each field/operator argument: An array representing a single column of cells.
Iteration space: All owned columns plus the halo columns that psyclone deems necessary to minimise the halo exchanges. This will depend on the access pattern and the optimisations applied (e.g. redundant computations).

operates_on: "owned_cell_column"
Data provided for each field/operator argument: An array representing a single column of cells.
Iteration space: Only provide owned columns, so all halo column will be dirty after this kernel, requiring a halo exchange when accessing them and preventing any redundant computation.
```

Another example is the 'dof', it describes what is provided to the kernel, but doesn't says anything about the iteration space, I don't know if it means for the iteration space that psyclone deems optimal, or all the dofs from all the halos.

doc/user_guide/lfric.rst
```diff
@@ -1961,15 +1961,27 @@ operates_on                    Data passed for each field/operator argument
...
+``dof``                        Single DoF.
+``owned_dof``                  Single DoF but the iteration space must *only* include
+                               owned DoFs (i.e. not annexed or halo DoFs).
 ``domain``                     All columns of cells in the (sub-)domain.
 ============================== =======================================================
 
-(For a description of the concepts of 'owned' and 'halo' cells please see the
-:ref:`lfric-developers`.)
+(For a description of the concepts of 'owned' and 'halo' cells and 'annexed' DoFs
+please see the :ref:`lfric-developers`.)
```
sergisiso: Aha, here are the pictures I was looking for. This may be a separate PR, but I think this needs reorganising.

I already think that this operates_on section is too deep in this page, some preceding text already assumes I know about these. But the LFRic concepts explained inside "Developer Guide > Generic Code" should definitely be introduced earlier in this page.

Unfortunately it is not as simple as copy paste because that section also eventually goes too deep into internal details, but the basic concepts at the top would be useful.

doc/user_guide/lfric.rst
```diff
@@ -1961,15 +1961,27 @@ operates_on                    Data passed for each field/operator argument
...
+
+The ``owned_cell_column`` and ``owned_dof`` values of ``OPERATES_ON`` are intended for
+use only with special cases where the kernel concerned cannot be used to perform
+redundant computation (e.g. when filling a field with pseudo-random numbers without
+regard to cell location). Processing an application that makes use of such a kernel
+requires that the ``COMPUTE_ANNEXED_DOFS`` configuration option (see
+:ref:`lfric-annexed_dofs`) be set to ``False`` as PSyclone can no longer guarantee that
+annexed DoFs are always clean between different ``invoke`` calls.
```
sergisiso, line 2009: I don't understand this, shouldn't using the new "owned_dofs" operate in a similar way regarding as compute_anexed_dofs  as "owned_cell_columns" does to redundant computation? E.g. preventing computing annex dofs (requiring a halo exchange) for such kernels.

doc/user_guide/lfric.rst
```diff
@@ -3242,6 +3254,9 @@ pseudo-random numbers in the interval ``0 <= x < 1``::
 where ``RAND()`` is some function that returns a new pseudo-random number
 each time it is called.
 
+Due to its simplistic implementation, this kernel does *not* support
+redundant computation and hence has ``OPERATES_ON=owned_dof``.
```
sergisiso: Maybe:
Due to different parallel elements using independent random number generator streams, this built-in has ``OPERATES_ON=owned_dof``, which will prevent optimisations such as redundant computation and computed annexed dofs involving this built-in.

doc/user_guide/lfric.rst
```diff
@@ -879,7 +879,8 @@ types.
 Rules specific to General-Purpose Kernels without CMA Operators
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-1) General-purpose kernels with ``operates_on = CELL_COLUMN`` accept
+1) General-purpose kernels with ``operates_on = CELL_COLUMN`` or
+   ``OWNED_CELL_COLUMN`` accept
```
sergisiso: `General-purpose kernels that operate in any kind of CELL_COLUMN`?(or does this not apply to owned_and_halo_cell_colum and halo_cell_colum)

sergisiso (changes requested):
@arporter See initial comments, I am afraid I didn't make it further than the documentation :) But I want to understand first what this is trying to do, before I see if the implementation matches it.

doc/user_guide/lfric.rst
arporter: Good point. I've re-structured the table so that it now has a separate, 'iteration space' column. I think it's much clearer now.

doc/user_guide/lfric.rst
arporter: The COMPUTE_ANNEXED_DOFS option is a global setting that tells PSyclone to always perform redundant computation on annexed dofs. When this is in effect, PSyclone always assumes that annexed dofs are clean and places halo exchanges accordingly. The trouble is that this assumption affects *all* potential halo exchanges, not just those in the immediate invoke of a kernel using the new `owned_dofs` operates_on. Therefore, we have to turn it off if any kernel cannot redundantly compute annexed dofs.

doc/user_guide/lfric.rst
arporter: Thanks. I've updated along those lines.

doc/user_guide/lfric.rst
arporter: Oh yes, I missed that.

doc/user_guide/lfric.rst
arporter: I agree. It requires work. It would perhaps be better if we could link to LFRic documentation rather than document LFRic concepts here. For now, I've added some cross-references to where `operates_on` is described.

doc/user_guide/lfric.rst
arporter: I've also added a note to the introduction to the whole LFRic section signposting the Dev Guide for those unfamiliar with general LFRic concepts. It's a bit of a sticking plaster but it's better than nothing.

doc/user_guide/lfric.rst
```diff
@@ -1955,21 +1964,38 @@ is supplied with the specified data for each field/operator argument.
...
+============================== ============================================ ==========================================
+operates_on                    Data passed for each field/operator argument Iteration space
+============================== ============================================ ==========================================
+``cell_column``                Single column of cells.                      Defaults to columns from the owned region
+                                                                            unless the kernel performs an INC
+									    operation on continuous fields when it
+									    will include one level of halo cells.
+									    The iteration space may be extended into
+									    the halo in order to perform redundant
+									    computation.
```
sergisiso: Different indentation (convert the tabs to spaces)

doc/user_guide/lfric.rst
```diff
@@ -1955,21 +1964,38 @@ is supplied with the specified data for each field/operator argument.
...
-``dof``                        Single DoF .
-``domain``                     All columns of cells in the (sub-)domain.
-============================== =======================================================
-
-(For a description of the concepts of 'owned' and 'halo' cells please see the
-:ref:`lfric-developers`.)
+============================== ============================================ ==========================================
+operates_on                    Data passed for each field/operator argument Iteration space
+============================== ============================================ ==========================================
+``cell_column``                Single column of cells.                      Defaults to columns from the owned region
```
sergisiso: I find the INC comment a bit out of place here, it seems more a description for a psyclone developer than a user (writing psykal kernels). What do you think of:

All columns from global view. For each DM process this will operate on all owned columns and may be extended into the halo in order to perform redundant computation.

doc/user_guide/lfric.rst
```diff
@@ -1955,21 +1964,38 @@ is supplied with the specified data for each field/operator argument.
...
+                                                                            unless the kernel performs an INC
+									    operation on continuous fields when it
+									    will include one level of halo cells.
+									    The iteration space may be extended into
+									    the halo in order to perform redundant
+									    computation.
+``owned_cell_column``          Single column of cells.                      Restricted to owned columns (i.e. excludes
+                                                                            halo columns).
+``halo_cell_column``           Single column of cells.                      Columns exclusively from the halo region
+                                                                            (to a specified depth).
```
sergisiso: This look good but maybe "restricted to" and "exclusively from" can be made consistent, to prevent users to thing there is a deeper meaning form the different expressions.

Also, in "owned_cell_column" I would add "Prevents extending into the halos to perform redundant computation" (the negative of the sentence used in "cell_columns")

src/psyclone/domain/lfric/lfric_arg_descriptor.py
```diff
@@ -442,7 +442,7 @@ def _init_field(self, arg_type, operates_on):
                            const.VALID_ANY_SPACE_NAMES)
 
         # Check accesses for kernels that operate on DoFs
-        if operates_on == "dof":
+        if operates_on in ["dof", "owned_dof"]:
```
sergisiso: Maybe we are done adding OPERATES_ON, but here and on line 545, is there a LFRicConstant that represents this set of values? If not, should there be?

src/psyclone/domain/lfric/lfric_loop.py
```diff
@@ -268,11 +268,11 @@ def load(self, kern):
         # Loop bounds
         self.set_lower_bound("start")
         const = LFRicConstants()
-        if kern.iterates_over == "dof":
+        if kern.iterates_over in ["dof", "owned_dof"]:
```
sergisiso: Const parameter?

sergisiso (changes requested):
@arporter This is almost ready now, I added some more comments about the documentation (some are opinionated to feel free to push back). And I agree with the code changes in the PR.

What it is a bit harder to check is the code not changes regarding operates_on comparisons because they come in different patterns. I have grep all "operates_on"/ "iterates_over"/"iterations_space" followed by "=="/"in". (Is there any other combination we should check?)

And I think some of them may be wrong. Particularly in the examples and tutorials there are many
`.iteration_space == "cell_column"`  and `iteration_space == "dof"` which I think need to be updated. For example in the ones in "examples/lfric/scripts/gpu_offloading.py"

doc/user_guide/lfric.rst
arporter: Dang. That's emacs for you. Hopefully I've fixed that now.

doc/user_guide/lfric.rst
arporter: Oh yes. I've checked and we do cover this in the Dev Guide so I will update the text as you suggest.

doc/user_guide/lfric.rst
arporter: Good ideas, done.

src/psyclone/domain/lfric/lfric_arg_descriptor.py
arporter: Oh yes. I've updated them.

src/psyclone/domain/lfric/lfric_loop.py
arporter: Oh yes.

arporter:
Ready for another look now. I've fired off the integration tests too.

PR #2953, opened 2025-04-11
(closes #2952) Gracefully fail resolving transitive type

### LINKED ISSUE #2952: symbol.resolve_type fails for transitive imports
```
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/centos/workspace/psyclone/PSyclone/src/psyclone/transformations.py", line 597, in validate
    self.validate_it_can_run_on_gpu(node, options)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/centos/workspace/psyclone/PSyclone/src/psyclone/transformations.py", line 459, in validate_it_can_run_on_gpu
    symbol.resolve_type()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol.py", line 263, in resolve_type
    if extern_symbol.initial_value:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Symbol' object has no attribute 'initial_value'
```

sergisiso:


sergisiso:
@arporter This is a small fix to gracefully fail in case of calling resolve_type on a transitive import

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -248,10 +248,18 @@ def resolve_type(self):
                   examining the Container from which it is imported.
         :rtype: subclass of :py:class:`psyclone.psyir.symbols.Symbol`
 
+        :raises SymbolError: if the type could not be resolved.
         '''
         if self.is_import:
             extern_symbol = self.get_external_symbol()
+            # pylint: disable=import-outside-toplevel
```
arporter, line 255: Just to be precise, could you do `disable-next=...` here and below.

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -248,10 +248,18 @@ def resolve_type(self):
...
             from psyclone.psyir.symbols import RoutineSymbol
+            # pylint: disable=unidiomatic-typecheck
+            if type(extern_symbol) is Symbol:
+                raise SymbolError(
+                    f"The external symbol '{extern_symbol.name}' was found "
+                    f"but it does not have a type. Maybe it is a transitive "
+                    f"import which is currently not resolvable.")
```
arporter, line 262: Please could you add "...transitive (indirect) import which..." as, until you started using the term "transitive", I'd never heard of it.

arporter (changes requested):
Thanks Sergi. Just a couple of little tweaks and then this can go on.

Separately, I'm wondering whether we should have a `Symbol.is_typed` property that could return `False` for `Symbol` and `True` otherwise? Although I'm not sure what we'd do with Routine symbols if we did that?

src/psyclone/psyir/symbols/symbol.py
sergisiso: Done

src/psyclone/psyir/symbols/symbol.py
sergisiso: but I removed one of them as I don't need the type equality check anymore

sergisiso:
@arporter Ready for another review

> Separately, I'm wondering whether we should have a Symbol.is_typed property that could return False for Symbol and True otherwise? Although I'm not sure what we'd do with Routine symbols if we did that?

Routine is typed (as in has `.datatype`) but it returns the NoType. I am unsure about `is_typed` we already have TypedSymbol (which I now use in the conditional)

arporter (commented):
All good now, thanks Sergi.
I'll run the integration tests, just to be on the safe side.

arporter (approved):
Will proceed to merge.

PR #2958, opened 2025-04-17
nemo acc transformation

imaranresearch:
Changes in acc_kernels_trans.py and utils.py for OpenACC transformation.


examples/nemo/scripts/utils.py
```diff
@@ -81,68 +83,65 @@
...
     'arr_lbnd_5d_dp', 'arr_lbnd_5d_i', 'arr_lbnd_5d_sp', 'atg',
     'bdy_oce_alloc', 'bdy_segs_surf', 'Cd_from_z0', 'CdN10_f_LU12',
     'CdN10_f_LU13', 'cd_n10_ncar', 'cd_neutral_10m', 'CdN_f_LG15',
     'CdN_f_LG15_light', 'CdN_f_LU12_eq36', 'ce_n10_ncar', 'charn_coare3p0',
     'charn_coare3p6', 'charn_coare3p6_wave', 'check_hdom', 'ch_n10_ncar',
-    'cp_air', 'cp_air_sclr', 'cp_air_vctr', 'cpl_freq', 'crs_dom_alloc',
+    'cp_air', 'cpl_freq', 'crs_dom_alloc',
     'crs_dom_alloc2', 'dayjul', 'def_newlink', 'delta_skin_layer',
-    'depth', 'dep_to_p', 'de_sat_dt_ice_sclr', 'de_sat_dt_ice_vctr',
+    'depth', 'dep_to_p', 'de_sat_dt_ice',
```
sergisiso, line 96: That's a good point, I can't believe we missed marking the interfaces as calls for so long, adding them also solves some compiler segmentation faults I seen!

PR #2969, opened 2025-05-01
2902 add driver tests to integration tests

hiker:
This is ready to go in

### DIFF
--- a/.github/workflows/extraction_test.yml
+++ b/.github/workflows/extraction_test.yml
@@ -96,8 +96,8 @@ jobs:
 
         #TOO SCARY ATM# rm -rf ${RUNDIR}
         mkdir -p ${LFRIC_CORE} ${LFRIC_APPS}
-        cp -r ${HOME}/LFRic/lfric_core_${LFRIC_CORE_REV} ${LFRIC_CORE}
-        cp -r ${HOME}/LFRic/lfric_apps_${LFRIC_APPS_REV} ${LFRIC_APPS}
+        cp -r ${HOME}/LFRic/lfric_core_${LFRIC_CORE_REV}/ ${LFRIC_CORE}
+        cp -r ${HOME}/LFRic/lfric_apps_${LFRIC_APPS_REV}/ ${LFRIC_APPS}
         # Install the BAF scripts into the lfric directories:
         cd lfric-baf
         ./install.sh ${LFRIC_CORE} ${LFRIC_APPS}
@@ -105,5 +105,21 @@ jobs:
         mkdir -p $PWD/fab_workspace
         export FAB_WORKSPACE=$PWD/fab_workspace
         # For debugging:
-        env
-        ${LFRIC_CORE}/build.sh ./fab_gungho_model_extract.py -fc mpif90-gfortran -ld linker-mpif90-gfortran
+        export PATH="$(dirname $CC):$(dirname $LDMPI):$PATH"
+        unset CC FC LD
+        ${LFRIC_CORE}/build.sh ./fab_gungho_model_extract.py   \
+          -fc mpif90-gfortran -ld linker-mpif90-gfortran  -cc gcc -rp v2.0
+        # Run
+        # Modify number of timesteps
+        cd example
+        sed -i "s/timestep_end='10'/timestep_end='1'/" configuration.nml
+        
+        mpirun -n 6 ../fab_workspace/gungho_model_extract-mpif90-gfortran/gungho_model configuration.nml
+
+        cd ..
+
+        FILE=sci_fem_constants_mod_psy-invoke_compute_w1_mass_matrix-compute_mass_matrix_w1_code-r0
+
+        mpif90 driver-$(FILE).F90 $FFLAGS $LDFLAGS -l yaxt_c -o driver-$FILE
+
+        ./driver-$FILE ./example/$(FILE)-0.nc
--- a/.github/workflows/extraction_test.yml
+++ b/.github/workflows/extraction_test.yml
@@ -81,7 +81,7 @@ jobs:
         pip install .
 
     # Compile GungHo with extraction
-    -name GungHo Extraction Compilation    
+    -name: GungHo Extraction Compilation    
         # Set up environment
         source /apps/spack/psyclone-spack/spack-repo/share/spack/setup-env.sh
         spack load lfric-build-environment%gcc@14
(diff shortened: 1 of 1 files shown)


sergisiso:
We agreed that #1010 will temporally break the extraction until the follow up PR, but this means that this integration test will be broken during the time in-between. We can proceed with the merge but expect to see it red until the fix is merged or wait on this PR. What do you think @hiker @arporter ?

.github/workflows/extraction_test.yml
```diff
@@ -0,0 +1,151 @@
...
+        cd ${LFRIC_APPS}/applications/gungho_model
+        mkdir -p $PWD/fab_workspace
+        export FAB_WORKSPACE=$PWD/fab_workspace
+        # Fab expects that all compilers are in PATH, and
+        # does not support CC/FC etc with a full path. Since
+        # the spack environment does not add all required paths,
+        # we need to add the required paths manuall, and then
+        # unset the environment variables (which would otherwise
+        # be used as default)
+        export PATH="$(dirname $CC):$(dirname $LDMPI):/apps/spack/psyclone-spack/spack-repo/opt/spack/linux-centos9-skylake_avx512/gcc-14.1.0/rose-picker-2-mmt2v3hqjwoqgkyf4f2ijlzlsy7nubnp/bin:$PATH"
```
sergisiso: Can the hardcoded rose-picker path be replaced with a "spack location -i rose-picker" command?

.github/workflows/extraction_test.yml
```diff
@@ -0,0 +1,151 @@
...
+        # the kernel data files are called $FILE-$RANK.nc
+        FILE=sci_fem_constants_mod_psy-invoke_compute_w1_mass_matrix-compute_mass_matrix_w1_code-r0
+
+        # We link with yaxt, since a few drivers inline a module that contains yaxt
+        # functions (though these functions and therefore yaxt are not actually used)
+        mpif90 driver-${FILE}.F90 $FFLAGS $LDFLAGS -l yaxt_c -o driver-${FILE}
+
+        # Execute the driver, which will print results to the screen
+        # -----------------------------------------------------------
+        ./driver-${FILE} ./example/${FILE}-0.nc
```
sergisiso: At the end, can we also print out some statistics so it is easy to compare PRs? I am thinking `ls` the number of created drivers and number of created nc files for now?

sergisiso (changes requested):
@hiker Now that #1010 is on master this will fail until we get #2948 (as we expected), so I suggest waiting until that one merged to avoid consuming integration test time until then. Other than that there is just minor comments.

hiker:
I'll agree with holding off - I had no idea that 1010 would go on master that quickly :)

Additionally, I am just adding support for using an absolute path when specifying a compiler in Fab. This would make two of the three modifications to `PATH` unnecessary. It will be a bit before this is merged in all the way through (fab  -> BAF -> our LFRic CI),  but I'd hope that we then can have a better PR :)

sergisiso:
@hiker I created a merge of this branch and #2948 and it created the extraction gungho_model but failed to execute it because the path was wrong, see https://github.com/stfc/PSyclone-mirror/actions/runs/14881133082/job/41789313604

It created `fab_workspace/gungho_model_extract-full-debug-mpif90-gfortran/gungho_model`
it tries to run `fab_workspace/gungho_model_extract-mpif90-gfortran/gungho_model`

hiker:
Yes, that branch has a new PR merged in (supporting compilation profiles like "fast-debug", or "optimsed"), which changes the name. Even then, I don't think you can use the compiler, due to additional consistency checks in the mpif90-wrapper: it will check that the wrapped compiler (gfortran) has the same version as mpif90 reports. And the setup in this PR fixes that mpif90 has a full path, but not gfortran (unless we add this to the script, or give this machine a site-specific setup, all of which I want to avoid to make this use case as simple as possible).

I will change the compiler wrapper to not abort, as long as mpif90 reports a valid version number in the form of the wrapped compiler.

sergisiso:
@hiker As mentioned, we can unblock this now, but also, you will need to replace the spack installation dir to the new one:
`source /apps/spack/psyclone-spack-25.3/spack-repo/share/spack/setup-env.sh`

sergisiso:
@hiker I made a merge of my extraction branch and this one, and launched the integration test, see https://github.com/stfc/PSyclone-mirror/actions/runs/15445409188/job/43473560245

It failed with:
```
mpirun was unable to launch the specified application as it could not access
or execute an executable:

Executable: ../fab_workspace/gungho_model_extract-mpif90-gfortran/gungho_model
```
because the binary is in called `fab_workspace/gungho_model_extract-fast-debug-mpif90-gfortran/gungho_model` (the "fast-debug" bit is missing in the mpirun)

.github/workflows/extraction_test.yml
hiker: Done.

.github/workflows/extraction_test.yml
hiker: Done. Note that the number of extracted kernels will likely change from one gungho version to the next.

hiker:
Also note that we are now pulling a separate branch, so the PSyclone CI won't be affected by changed to our lfric builds (or fab).

Additional comment: our lfric-fab build system includes a copy of the PSyclone extraction library (since they are not installed by a standard PSyclone install), and they contain a newer version of the compare functions (i.e. the one in #2907), which has nicer statistics.

examples/lfric/scripts/extract_script.py
```diff
@@ -0,0 +1,55 @@
...
+                if loop.kernels()[0].name in SETVAL_BUILTINS:
+                    rtrans.apply(loop, options={"depth": 1})
+
+    for invoke in psy.invokes.invoke_list:
+        schedule = invoke.schedule
+        for kern in schedule.walk(LFRicLoop):
+            try:
+                extract.apply(kern, {"create_driver": True})
+            except NotImplementedError:
+                pass
```
sergisiso: Minor comment but can you print the kern name and the str(err), so at least is kept in the log of the action if we want to explore more.

sergisiso (changes requested):
@hiker I was about to approve but I have one question regarding:
> our lfric-fab build system includes a copy of the PSyclone extraction library (since they are not installed by a standard PSyclone install)

This means that changes in the extraction library are not being tested here? Is there a way to point fab to the extraction library coming with the commit that we are testing? or could there be in the future? If not, maybe it should be mentioned prominently as a comment in the workflow to prevent getting somebody confused by unexpected behaviour in the future.

examples/lfric/scripts/extract_script.py
hiker: Done.

PR #2980, opened 2025-05-08
(closes #2976) Remove jsonschema dependency

### LINKED ISSUE #2976: Remove jsonschema dependency
jsonshcema is a huge dependency, on Spack it ends up installing the rust toolchain plus other non trivial dependencies.

For what I can see it was only used for the notebooks which are now disabled. So it should be deleted from the dependencies (or at least not be a mandatory one).

Also we need to update the Spack recipe. If it is really unused we could modify the upstream spack package now (we don't need to wait for a release), but we need to test that this is true first.

sergisiso:


sergisiso:
@arporter @LonelyCat124 @hiker This is ready to review

arporter (approved):
Change looks good. I've grepped and the only mentions for json-schema are all related to Jupyter notebooks. The readme for those examples says to install jupyter using pip so I think we're fine.

PR #2986, opened 2025-05-09
#2984 Fix python version test to support python 3.10

### LINKED ISSUE #2984: Test failure for python 3.10
Simple typo:
```
a/src/psyclone/tests/psyir/transformations/transformations_test.py b/src/psyclone/tests/psyir/transformations/transformations_test.py
index 3051fc0375..c6294a997d 100644
--- a/src/psyclone/tests/psyir/transformations/transformations_test.py
+++ b/src/psyclone/tests/psyir/transformations/transformations_test.py
@@ -571,7 +571,7 @@ def test_omploop_trans_new_options(sample_psyir):
     # Check python version, as this tests have different behaviour for
     # new python versions vs 3.8 or 3.7.
     # TODO #2837: This can be removed when Python 3.7 and 3.8 are retired.
-    if sys.version_info[1] < 11:
+    if sys.version_info[1] < 10:
         with pytest.raises(TypeError) as excinfo:
             omplooptrans.apply(tree.walk(Loop)[0], collapse="x")
         assert ("The 'collapse' argument
```
3.10 is 'new' python for this test.

hiker:
Fixes #2984 (and a minor pylint warning)

### DIFF
--- a/src/psyclone/tests/psyir/transformations/transformations_test.py
+++ b/src/psyclone/tests/psyir/transformations/transformations_test.py
@@ -41,8 +41,10 @@ API-agnostic tests for various transformation classes.
 '''
 
 import os
-import pytest
 import sys
+
+import pytest
+
 from fparser.common.readfortran import FortranStringReader
 from psyclone.psyir.nodes import CodeBlock, IfBlock, Literal, Loop, Node, \
     Reference, Schedule, Statement, ACCLoopDirective, OMPMasterDirective, \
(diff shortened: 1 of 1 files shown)


hiker:
Ready for review - the doc failure is an external link to https://gtr.ukri.org/projects?ref=NE%2FL01209X%2F1  timing out.

LonelyCat124:
Hi @hiker - thanks for fixing this. I've checked that its all working correctly with the changes for 3.9 and 3.10 (provided sphinx is installed).
While we're modifying this - could I get you to add a check for if sphinx is installed (e.g. `sys.modules.get("sphinx", None)` to skip the type checking tests in this section of tests (i.e. lines 560 to 585), or to do the alternate output (i.e. similar to what we have for pre-3.10 for the second test). These tests currently break without sphinx installed.

I'm personally happy with the former as our test suite runs with sphinx installed now - @sergisiso or @arporter any preference?

arporter:
I think we've accepted that `sphinx` is now a dependence for the test suite so I'm happy to stick with that.

arporter:
My apologies but I've belatedly realised that my #3034 makes this one unnecessary. Does any of the discussion about Sphinx need to be taken up in #3034?

LonelyCat124:
> My apologies but I've belatedly realised that my #3034 makes this one unnecessary. Does any of the discussion about Sphinx need to be taken up in #3034?

Probably yes - though I think there are some tests in `psyGen_test` and `docstring_parser_test` that pretend sphinx isn't available. so it depends if those are sufficient for what we had discussed here.

PR #2989, opened 2025-05-12
Update CI dependencies

nmnobre:


nmnobre:
Marked for review, I want to know what people think of this... Can use a preprocessor conditional to target the workaround more precisely, but that implies a bigger changelog and changing compiler flags in multiple places (`pytest` + `makefile`).

sergisiso:
Hey @nmnobre , the problem modifying `src/psyclone/tests/test_files/dynamo0p3/infrastructure` is that this code is essentially copied from the lfric core repository. We do not want to diverge from them because from time to time we want to update it by copying a recent version (what @hiker is doing in his branch). Ultimately we may replace this with a submodule when lfric moves to git.

We asked yesterday the LFRic team and they are aware of the gcc15 issues (this one and others), I will send them your fix. But it will be better to pause the gcc update and give some time to the MetOffice to do it on their end, so we just replicate their infrastructure. Otherwise we need to start maintaining fixes in this code. The rest of package updates can go in.

nmnobre:
That makes a lot of sense, I've tried rolling back gfortran just for the problematic step and requested a run on the self-hosted runner. :)

nmnobre:
This is done now. Let CI do the complete test suite, ready to be merged when all green.

arporter (approved):
Thanks very much for digging into the problem Nuno and finding the compiler-flag solution.
All looks good. Integration tests were green.

Issue #1604, opened 2022-02-04
Allow GH_READWRITE/WRITE access for continuous function spaces


TeranIvy:
There are instances in LFRic Physics schemes where halo exchanges are not required despite a field being on a continuous function space. E.g. winds are in `W2` space, however the way certain Physics schemes are computed means there does not need to be communication between different MPI processes to update the halos for continuous quantities.  `W2` is then more about marking the placement of DoFs (data points) for values than the continuity in the horizontal.

Currently PSyclone does not allow  `GH_READWRITE` and `GH_WRITE` accesses for continouous fields for correctness. After some discussions, we agreed that we will allow `GH_READWRITE` and `GH_WRITE` access for continouous fields for cases like these. It is then up to the developers who wish to use this feature to ensure that they will get correct results without halo exchanges.

arporter:
We already have an issue for this @TeranIvy (and I've started work there).

TeranIvy:
Thanks @arporter! Could you please refer it here so I can pass it to Ian?

TeranIvy:
Ok, I see it is #1542. I'll add the mention of GH_READWRITE there.

PR #2994, opened 2025-05-13
(Closes #2992) fix bug in LFRicKern.all_updates_are_writes

### LINKED ISSUE #2992: Turning `COMPUTE_ANNEXED_DOFS=false` results in dirty annexed dofs being used in calculations
## Description

Found during the adjoint development - setting `COMPUTE_ANNEXED_DOFS=false` results in annexed dofs becoming dirty. The problem here is that kernels unknowingly use the dirty annexed dofs and this can cause error in calculations (shown through an adjoint test).

## Further information

In particular, it was noticed in a matrix operation between W2/W3 functionspaces. The psy-layer does not check the "dirtiness" of annexed dofs if it does not access the halo (there is no distinction).

```fortran
call invoke( X_plus_Y(w2, some_other_w2_a, some_other_w2_b), &
             ! W2 annexed dofs out-of-date/dirty at this point
             X_plus_Y(w3, some_other_w3_a, some_other_w3_b), &
             ! Dirty annexed dofs used by this kernel. Psy-layer does not check `dirty(depth=1)`
             dg_matrix_vector_kernel(w2, w3, matrix) )
```

arporter:


arporter:
If I repeat this with `master` (and `COMPUTE_ANNEXED_DOFS=False`) then GungHo runs but the checksums are wrong.

arporter:
This is a small but significant bug fix to the functionality that identifies special LFRic kernels that have GH_WRITE on continuous fields. This impacts the halo-exchange logic and is a cause of problems identified when COMPUTE_ANNEXED_DOFS is set to False.

hiker:
> With this fix, I can run GungHo and get the same answers (checksums) with and without COMPUTE_ANNEXED_DOFS set. @hiker / @christophermaynard does that mean the original problem that we had is fixed or is there some other test I need to do?

Not sure what exactly the problem is either, i.e. if it was a crash, invalid results (different checksums). I am not even sure how much it affected gungho, I assume the original test was lfric_atm.

mo-joshuacolclough:
Thanks for all the work on this. I'll be able to test this soon with the adjoint - currently asking around about how to do so!

> if it was a crash, invalid results (different checksums). I am not even sure how much it affected gungho, I assume the original test was lfric_atm.

For some context: This was found when developing the LFRic adjoint model. For those tests we use `setval_random` to initialise fields, and in this case with `COMUTE_ANNEXED_DOFS=false`. When also running with a debug build, unset field values are set as `nan` values, which is useful for seeing when `dirty` dofs are being accessed. This is quite useful for the adjoint, as instances of this happening may result in an invalid adjoint & a hard-to-find bug!

In this particular case, dirty annexed dofs are accessed in the linear code call. If I manually extend the `setval_random` calls to populate annexed dofs (i.e, dirty but at least set to a value), then the nan issues go away, but we see an adjoint test failure. And so a halo exchange is required.

> I am not even sure how much it affected gungho

It's possible this has a small impact on results from MPI runs, around the partition boundaries, when kernels operate between functionspaces with annexed dofs in particular... However I have not measured the impact myself; it perhaps should be checked by someone in-the-know.

On the other hand the adjoint model has to be an exact transpose of the linear model, which is why this small discrepancy has been picked up here and must be fixed in this case.

src/psyclone/domain/lfric/lfric_kern.py
```diff
@@ -601,13 +601,20 @@ def mesh(self):
         return self._mesh_properties
 
     @property
-    def all_updates_are_writes(self):
+    def all_continuous_updates_are_writes(self) -> bool:
```
sergisiso, line 604: I am confused, the name does not match the implementation: it is called 'all_continuous_updates_are_writes' but it returns True only if all updates are write-only, including the discontinuous, as far as and the argument we use to dereference the iteration space is not discontinuous.

So my question is why is the argument we use to dereference the iteration space special? Should it be if any argument is discontinuous it returns False? Or is it because if the iteration space is discontinuous, we iterate over the halo which is what we are testing here. If its the second, the name will be something or even the caller should be responsible for the check. Do the discontinuous check be off if the distributed memory is off?

Also, wouldn't READS need the same halo exchanges, isn't the caller already checking the same conditions?

I have the feeling that this is trying to fix a problem of the caller making this internal functionality very specific.

Also, I do not understand the docstring, again it seems to talk about a very specific condition for a method with a very generic name, and I can't figure out the context.

sergisiso (changes requested):
@arporter See may comments about the updated method

arporter:
I've significantly tightened-up the function but am now getting an integration test failure with the OMP offload version of LFRic. Passthrough with DM works fine. Failure is in the solver:
```
20250519164016.214+0100:P0:S1:INFO : BLOCK_GCR starting ... ||b|| =  0.10021990E
+04: 0.10266128E+04
20250519164017.246+0100:P0:S1:INFO : BLOCK_GCR solver_algorithm: [ 1, 4], iters,
 init=  0.1027E+04 final=            NaN abs=            NaN
```
weirdly, this is with `COMPUTE_ANNEXED_DOFS = true`.

arporter:
I can reproduce this failure on CPU with `everything_everywhere_all_at_once.py` and ANNEXED_DOFS set to false. My latest changes have substantially changed the behaviour of the method - it excludes those kernels that update fields on discontinuous spaces and, separately, those that update an operator. It didn't do that previously.

arporter:
Having slept on it, I decided that what I needed to do was effectively revert my changes and just include the check that the field with GH_WRITE access is on a continuous space. I can build and run GungHo with `everything_everywhere_all_at_once.py` with this change and get the correct results (with COMPUTE_ANNEXED_DOFS=false).

However, now that I understand what we're checking for, it seems incredibly fragile. A user could easily make the mistake of specifying GH_WRITE on a continuous field when they really meant GH_INC. As I've found to my cost, debugging such a mistake is hard. I would therefore argue that, for this special case, we should have a different form of access, e.g. `GH_CONTINUOUS_WRITE` to make it clear this is something special. Obviously this would be a non-backwards compatible change but I think it's important. What do @mo-joshuacolclough, @TeranIvy and @tommbendall think?

tommbendall:
> Having slept on it, I decided that what I needed to do was effectively revert my changes and just include the check that the field with GH_WRITE access is on a continuous space. I can build and run GungHo with `everything_everywhere_all_at_once.py` with this change and get the correct results (with COMPUTE_ANNEXED_DOFS=false).
> 
> However, now that I understand what we're checking for, it seems incredibly fragile. A user could easily make the mistake of specifying GH_WRITE on a continuous field when they really meant GH_INC. As I've found to my cost, debugging such a mistake is hard. I would therefore argue that, for this special case, we should have a different form of access, e.g. `GH_CONTINUOUS_WRITE` to make it clear this is something special. Obviously this would be a non-backwards compatible change but I think it's important. What do @mo-joshuacolclough, @TeranIvy and @tommbendall think?

Hi Andy, just to check I've understood, are you proposing that we again ban `GH_WRITE` metadata for continuous fields (it was banned before #1604), but instead have a new `GH_CONTINUOUS_WRITE` metadata? What exactly would that do differently?

arporter:
> Hi Andy, just to check I've understood, are you proposing that we again ban `GH_WRITE` metadata for continuous fields (it was banned before #1604), but instead have a new `GH_CONTINUOUS_WRITE` metadata? What exactly would that do differently?

Hi Tom, functionally it would be identical. However, it would make it clearer that the kernel does something special. I'm worried about users just cut-n-pasting some metadata and modifying it until it appears to work, without realising that GH_WRITE on a continuous space is a special thing. I'm aware that changing the metadata will have a large cost associated with it so perhaps there are other solutions? Ideally we would extend PSyclone to check whether the kernel implementation is conformant but I suspect that will be very difficult or impossible. Perhaps we could introduce support for PSyKE to generate tests for kernels based on their metadata @hiker ? Actually, it might not be a job for PSyKE - it would be similar to what PSyAD does in generating a test harness for adjointed kernels.

tommbendall:
> > Hi Andy, just to check I've understood, are you proposing that we again ban `GH_WRITE` metadata for continuous fields (it was banned before #1604), but instead have a new `GH_CONTINUOUS_WRITE` metadata? What exactly would that do differently?
> 
> Hi Tom, functionally it would be identical. However, it would make it clearer that the kernel does something special. I'm worried about users just cut-n-pasting some metadata and modifying it until it appears to work, without realising that GH_WRITE on a continuous space is a special thing. I'm aware that changing the metadata will have a large cost associated with it so perhaps there are other solutions? Ideally we would extend PSyclone to check whether the kernel implementation is conformant but I suspect that will be very difficult or impossible. Perhaps we could introduce support for PSyKE to generate tests for kernels based on their metadata @hiker ? Actually, it might not be a job for PSyKE - it would be similar to what PSyAD does in generating a test harness for adjointed kernels.

Hi Andy, actually I think this is a good idea. I'm not sure how many instances of `GH_WRITE` on a continuous field that there are in LFRic, but it shouldn't be too much work to change those.

I don't know if this is relevant, but there are also situations in LFRic where there are continuous fields but the metadata is specified as `GH_WRITE` and `ANY_DISCONTINUOUS_SPACE`, because only one column will write to the shared DoFs.

src/psyclone/domain/lfric/lfric_kern.py
arporter: I take your point(s). It was poor. I've reverted my changes to this method and have put the additional logic (checking that the field being updated is on a continuous space) in the appropriate places. This has simplified things quite a lot.

arporter:
I've modified the passthrough integration test to have COMPUTE_ANNEXED_DOFS=false. However, I think the default transformation script used by LFRic will still be adding redundant computation to `setval_?` builtins. We could perhaps undo that as well?

If the CI and integration tests are green then this is ready for another look.

arporter:
I've added a `does_nothing.py` script and have tested it with `COMPUTE_ANNEXED_DOFS=false` and the answers matched. I've also updated the LFRic 'passthrough' integration test to use this script. I'll fire that off to check it works.

arporter:
Integration test was green :-) I'll just put Alistair's big PR on (#2979) and then bring this up-to-date.

arporter:

sergisiso (approved):
All good now @arporter and I see the integration tests that you launched are green. This is ready to merge.

PR #2996, opened 2025-05-15
Fixes acc ker

imaranresearch:
Fixes for openACC integration with nemo_v5


examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -197,24 +234,33 @@ def valid_acc_kernel(node):
...
+        if isinstance(enode, IntrinsicCall):
+            if "dim" in enode.argument_names:
+                return False
 
         if isinstance(enode, IfBlock):
             # We permit IF blocks originating from WHERE constructs and
             # single-statement IF blocks containing a Loop in KERNELS regions
-            if "was_where" in enode.annotations or \
-               "was_single_stmt" in enode.annotations and enode.walk(Loop):
+            if (
```
arporter, line 252: I think this check can probably be removed now.

examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -197,24 +234,33 @@ def valid_acc_kernel(node):
 
     # Rather than walk the tree multiple times, look for both excluded node
     # types and possibly problematic operations
-    excluded_types = (IfBlock, Loop)
+    excluded_types = (IfBlock, Loop, ArrayReference, IntrinsicCall)
     excluded_nodes = node.walk(excluded_types)
 
     for enode in excluded_nodes:
+        if isinstance(enode, IntrinsicCall):
```
arporter, line 245: The check for this is now on master so you can remove this bit.

examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -236,30 +284,35 @@ def valid_acc_kernel(node):
...
             # model configuration will contain.
             child = enode.loop_body[0] if enode.loop_body.children else None
-            if isinstance(child, Loop) and child.loop_type == "levels":
-                # We have a loop around a loop over levels
-                log_msg(routine_name, "Loop is around a loop over levels",
-                        enode)
-                return False
-            if enode.loop_type == "levels" and \
-               len(enode.loop_body.children) > 1:
+            # if isinstance(child, Loop) and child.loop_type == "levels":
```
arporter, line 295: Aha! I see we'd already disabled this heuristic. Once you've tested without it, we may be able to permanently delete this part.

examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -364,14 +419,16 @@ def try_kernels_trans(nodes):
...
+    """Applies OpenACC 'kernels' directives to NEMO code. Data movement can be
     handled manually or through CUDA's managed-memory functionality.
 
     :param psyir: the PSyIR of the provided file.
     :type psyir: :py:class:`psyclone.psyir.nodes.FileContainer`
-    '''
-    logging.basicConfig(filename='psyclone.log', filemode='w',
-                        level=logging.INFO)
+    """
+    logging.basicConfig(
```
arporter, line 435: I can't remember but I think we had to delete this part to get logging to work correctly (because PSyclone itself now sets-up the logging)?

examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -236,30 +284,35 @@ def valid_acc_kernel(node):
...
                 # How many distinct loop nests are there?
                 loop_count = 0
                 for child in enode.loop_body.children:
                     if child.walk(Loop):
                         loop_count += 1
                         if loop_count > 1:
-                            log_msg(routine_name,
-                                    "Loop over levels contains several "
-                                    "other loops", enode)
+                            log_msg(
```
arporter: Talking with you and Sergi today reminded me that it's better if these messages are put into the generated code rather than logged separately.
Therefore, please could you update all places where we call `log_msg` with some code that appends a comment instead. e.g. something like:
```python
msg = "Loop over levels contains several other loops"
if msg not in enode.preceding_comment:
    enode.append_preceding_comment(msg)
```
(The `if msg not in ...` avoids us generating duplicate comments as the recursive implementation means we may end up performing the same check several times.)

examples/nemo/scripts/acc_kernels_trans.py
```diff
@@ -270,7 +323,7 @@ def add_kernels(children):
     :returns: True if any KERNELS regions are successfully added.
     :rtype: bool
 
-    '''
+    """
     added_kernels = False
     if not children:
         return added_kernels
```
arporter, line 337: I've done a bit of experimenting and realised we can improve things a bit by altering this code so it doesn't bother to look at the children of Calls and Assignments (since they can't be put in a KERNELS region without their parent). Therefore, could you add:
```python
            elif isinstance(child, (Assignment, Call)):
                # There's no point looking at the children of a Call
               pass
```
just before the final `else` (at L348).

examples/nemo/scripts/utils.py
```diff
@@ -33,165 +33,438 @@
...
 ]
 
+
 # Currently fparser has no way of distinguishing array accesses from statement
 # functions, the following subroutines contains known statement functions
 CONTAINS_STMT_FUNCTIONS = ["sbc_dcy"]
 
 # These files change the results from the baseline when psyclone adds
 # parallelisation dirctives
-PARALLELISATION_ISSUES = [
```
arporter, line 177: We will need to make sure that these changes don't break the other uses of `utils.py` (since it is common to all of the NEMO transformation scripts). Running the integration tests will do that but I won't try that yet.

PR #2997, opened 2025-05-16
2923 support boolean in sympy

hiker:
Adds support for logical operators in the Sympy interface (i.e. reading and writing)

### DIFF
--- a/src/psyclone/psyir/backend/sympy_writer.py
+++ b/src/psyclone/psyir/backend/sympy_writer.py
@@ -39,15 +39,19 @@
 '''
 
 import keyword
+from typing import Dict, List, Optional, Set, Union
 
 import sympy
 from sympy.parsing.sympy_parser import parse_expr
 
 from psyclone.psyir.backend.fortran import FortranWriter
 from psyclone.psyir.backend.visitor import VisitorError
+from psyclone.core import Signature
 from psyclone.psyir.frontend.sympy_reader import SymPyReader
 from psyclone.psyir.nodes import (
-    DataNode, Range, Reference, IntrinsicCall, Call)
+    ArrayOfStructuresReference, ArrayReference, BinaryOperation, Call,
+    DataNode, IntrinsicCall, Literal, Node,
+    Range, Reference, StructureReference)
 from psyclone.psyir.symbols import (ArrayType, ScalarType, SymbolTable)
 
 
@@ -96,7 +100,7 @@ class SymPyWriter(FortranWriter):
     # A list of all reserved Python keywords (Fortran variables that are the
     # same as a reserved name must be renamed, otherwise parsing will fail).
     # This class attribute will get initialised in __init__:
-    _RESERVED_NAMES = set()
+    _RESERVED_NAMES: Set[str] = set()
 
     def __init__(self):
         super().__init__()
@@ -197,7 +201,11 @@ class SymPyWriter(FortranWriter):
                                   "never be called.")
 
     # -------------------------------------------------------------------------
-    def _create_sympy_array_function(self, name, sig=None, num_dims=None):
+    def _create_sympy_array_function(
+            self,
+            name: str,
+            sig: Optional[Signature] = None,
+            num_dims: Optional[List[int]] = None) -> sympy.Function:
         '''Creates a Function class with the given name to be used for SymPy
         parsing. This Function overwrites the conversion to string, and will
         replace the triplicated array indices back to the normal Fortran
@@ -206,7 +214,7 @@ class SymPyWriter(FortranWriter):
         to the object, so that the SymPyReader can recreate the proper
         access to a user-defined type.
 
-        :param str name: name of the function class to create.
+        :param name: name of the function class to create.
         :param sig: the signature of the variable, which is required
             to convert user defined types back properly. Only defined for
             user-defined types.
@@ -246,8 +254,10 @@ class SymPyWriter(FortranWriter):
         return new_func
 
     # -------------------------------------------------------------------------
-    def _create_type_map(self, list_of_expressions, identical_variables=None,
-                         all_variables_positive=None):
+    def _create_type_map(self,
+                         list_of_expressions: List[Node],
+                         identical_variables: Optional[Dict[str, str]] = None,
+                         all_variables_positive: Optional[bool] = None):
         '''This function creates a dictionary mapping each Reference in any
         of the expressions to either a SymPy Function (if the reference
         is an array reference) or a Symbol (if the reference is not an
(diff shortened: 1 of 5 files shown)


src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -39,7 +39,7 @@
 '''
 
 import keyword
-from typing import Dict, List, Optional
+from typing import Dict, Iterable, List, Optional, Set, Union
```
sergisiso: Now that we drop support for python < 3.9, we can use dict, set, list with lowercase which don't require importing.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -514,24 +529,23 @@ def _to_str(self, list_of_expressions, identical_variables=None,
...
         '''
-        is_list = isinstance(list_of_expressions, (tuple, list))
+        is_list = isinstance(list_of_expressions, (Iterable))
         if not is_list:
+            # Make mypy happy:
+            assert isinstance(list_of_expressions, Node)
             list_of_expressions = [list_of_expressions]
 
+        # Make mypy happy:
+        assert isinstance(list_of_expressions, Iterable)
```
sergisiso: I should start using mypy, but I was wondering if needing the asserts is a hint that the surrounding code is not very clear. Would:
```
if isinstance(list_of_expressions, Node):
    is_list = False
    new_list = [list_of_expressions]
elif isinstance(list_of_expressions, Iterable):
    is_list = True
    new_list = list_of_expressions
else:
   TypeError()
```
or simply requiring the argument to always be a list of Nodes and update the callers, solve the mypy issues without asserts.

Don't worry if not.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -597,7 +611,9 @@ def __call__(self, list_of_expressions, identical_variables=None,
 
         is_list = isinstance(list_of_expressions, (tuple, list))
         if not is_list:
+            assert isinstance(list_of_expressions, Node)
```
sergisiso: Same here?

src/psyclone/psyir/frontend/sympy_reader.py
```diff
@@ -37,10 +37,37 @@
...
 from psyclone.psyir.frontend.fortran import FortranReader
 
 
+# pylint: disable=invalid-name
+class FortranPrinter(Printer):
+    '''A helper class that converts Fortran logical operators back
+    to Fortran format (while SymPy has a Fortran printer (fcode) does this
+    as well, it does not handle e.g. Fortran Array expressions (a(2:5)),
+    so we need to use the not-Fortran-aware output to a normal string,
+    but handle logical operators separately.
```
sergisiso: Maybe:
'''Specialise the SymPy Printer to convert logical operators back to Fortran format. While SymPy has a Fortran printer (fcode), it does not handle e.g. Fortran Array expressions (a(2:5)), so we specialise the generic SymPy Printer and handle the necessary conversions.'''

src/psyclone/psyir/frontend/sympy_reader.py
```diff
@@ -123,7 +150,9 @@ def psyir_from_expression(self, sympy_expr, symbol_table):
         '''
         # Convert the new sympy expression to PSyIR
         reader = FortranReader()
-        return reader.psyir_from_expression(str(sympy_expr), symbol_table)
+        fp = FortranPrinter()
+        return reader.psyir_from_expression(fp.doprint(sympy_expr),
```
sergisiso, line 153: The Sympy printer code has this section:
```
    The printer looks for a specific method in every object. The name of that method
    depends on the specific printer and is defined under ``Printer.printmethod``.
    For example, StrPrinter calls ``_sympystr`` and LatexPrinter calls ``_latex``.
    Look at the documentation of the printer that you want to use.
    The name of the method is specified there.
```
I am wondering if the `print_fortran_array` below can be implemented as a `printmethod` in the FortranWriter instead. But don't worry if it is not a simple change.

sergisiso (changes requested):
The PR looks good @hiker, I have some minor questions about typing, mypy and sympy printing. The integration test should be passing but zdfsh2.f90 has given issues before, you could exclude this file  in OFFLOADING_ISSUES if it is just this.

src/psyclone/psyir/frontend/sympy_reader.py
hiker: I tried (and I actually liked your suggestion ;) ). But there are two problems:
1. If the `FortranPrinter` class is derived from `Printer` (as it is atm), then SymPy will switch to use the `emptyPrinter` (since `Printer` does not implement any `_print_XYZ` methods), which then replaces the uses `sstr` - so none of the `_print_XYZ` methods are called at all (since the methods are defined in `FortranPrinter`, not SymPy's string printer )
2. If I derive `FortranPrinter` from `StrPrinter`, and implement `_print_AppliedUndef`, the triggering works fine. BUT: we do have other `AppliedUndef`, like `LBOUND` ... which then all trigger the special handling code, which expects the number of parameter to be a multiple of 3 and does all kind of incorrect things for an LBOUND - so tests fail.

So tbh, I can't see a way of getting this to work without a deep dive into SymPy, or trying to define all these `AppliedUndef` functions (not sure if it's only intrinsic), and handle them separately in the `_print_AppliedUndef` method - both of which are kind of counter-productive.

The way of letting an object 'print' itself is well document, so I would be in favour of leaving it.

hiker: Should we perhaps move the `print_fortran_array` method into SymPyWriter? Since that is where the function is set for the newly created symbols.

src/psyclone/psyir/backend/sympy_writer.py
hiker: Good point (I have to admit I am never certain what the correct, future-proof way is - so thanks :) ). Done

src/psyclone/psyir/backend/sympy_writer.py
hiker: Good idea, I did something quite similar, without having to introduce a new variable

src/psyclone/psyir/backend/sympy_writer.py
hiker: Again, a small change simplified that.

src/psyclone/psyir/frontend/sympy_reader.py
hiker: Done

src/psyclone/psyir/frontend/sympy_reader.py
sergisiso: Thanks for trying, I am ok with the method living in the Reader as it is part of the SymPy->PSyIR logig. We can continue exploring this refactoring outside the PR as it is not part of the issue/PR description.

sergisiso (approved):
All comments are resolved and the integration test was green, this is ready to merge.

PR #3002, opened 2025-05-20
(Closes #2838) Linkspector should now work with ubuntu-latest

### LINKED ISSUE #2838: Linkspector doesn't currently support ubuntu-latest
See https://github.com/UmbrellaDocs/action-linkspector/issues/32

Linkspector fails with ubuntu-latest - no update currently available. We should keep an eye out on a fix for this, but in the mean time I'll roll this back to use ubuntu-22.04 as well.

LonelyCat124:


LonelyCat124:
@sergisiso @arporter @hiker ready for review from anyone. Just reverting a previous change now that this works with ubuntu-latest again.

sergisiso (approved):
@LonelyCat124 @nmnobre This is approved for merging. It moves up the version of the OS and compiler used and the multiple CIs are still all green. Note that from now on we need to use `-lnvtx3interop -cuda` when inserting the nvidia profiling hooks.

PR #3004, opened 2025-05-21
(Closes #2981) disable exit profiling

### LINKED ISSUE #2981: Presence of EXIT within a loop can break profiling
@imaranresearch has found that, when we add profiling to `p4zpoc` in function `gamain`, we end up enclosing an `EXIT`. This causes the closing caliper statement to be jumped over and we end up with profiling regions that appear to last the whole of the run.

The reason for this is that the Profiling transformation permits CodeBlock nodes to be enclosed. We should check such CodeBlock nodes for the presence of EXIT statements.

LonelyCat124:
First commit and implementation.

I think this will need potentially review from both Andy/Sergi and Joerg as we'll need to check that I haven't broken anything while updating the Transformations/PsyDataNode options.

### DIFF
--- a/src/psyclone/domain/gocean/transformations/gocean_extract_trans.py
+++ b/src/psyclone/domain/gocean/transformations/gocean_extract_trans.py
@@ -38,14 +38,17 @@
 '''This module contains the GOcean-specific extract transformation.
 '''
 
+from typing import Tuple
+
 from psyclone.gocean1p0 import GOLoop
 from psyclone.psyir.nodes import ExtractNode
 from psyclone.psyir.symbols import REAL8_TYPE, INTEGER_TYPE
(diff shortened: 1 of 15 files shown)


LonelyCat124:
Outside of coverage issues, I need to still check if the documentation is up-to-date

LonelyCat124:
Ready for a look now @arporter @sergisiso @hiker . Before merging at least it would be good to have it run with some production extraction use-cases to ensure I've not broken anything (the tests all pass but user usage might be different).

sergisiso:
@LonelyCat124 As we discussed, test regarding extraction are temporarily marked xfail in master. So we won't be able to test this properly until #2948 is merged, I will mark this PR as blocked until then.

PR #3008, opened 2025-05-27
Fix missing -p flag in passthrough in lfric_test.yml

arporter:
Fixes the LFRic passthrough integration test.


arporter:
Somehow the `-p psyclone-test` flag got missed off the LFRic build line for the passthrough part of the integration tests. This then means that the build system attempts to use `minimum/global.py` which a) it doesn't need to and b) causes it to fail because of all the recent name changes to the transformations.

PR #3009, opened 2025-05-29
(#Towards 2981) Fixes for profile trans without the updates

LonelyCat124:
This is the small PR with just the functionality changes from #3004 .

### DIFF
--- a/src/psyclone/psyir/transformations/profile_trans.py
+++ b/src/psyclone/psyir/transformations/profile_trans.py
@@ -33,12 +33,15 @@
 # -----------------------------------------------------------------------------
 # Author J. Henrichs, Bureau of Meteorology
 # Modified by R. W. Ford, A. R. Porter and S. Siso, STFC Daresbury Lab
+# Modified by A. B. G. Chalk, STFC Daresbury Lab
 # -----------------------------------------------------------------------------
 
 '''This module provides the Profile transformation.
 '''
 
-from psyclone.psyir.nodes import Return, ProfileNode
+from fparser.two.Fortran2003 import Exit_Stmt
+from psyclone.psyir.transformations import TransformationError
+from psyclone.psyir.nodes import CodeBlock, Return, ProfileNode
 from psyclone.psyir.transformations.psy_data_trans import PSyDataTrans
 
 
@@ -75,3 +78,27 @@ class ProfileTrans(PSyDataTrans):
 
     def __init__(self):
         super().__init__(ProfileNode)
+
+    def validate(self, nodes, options=None):
+        """
+        Checks that the supplied list of nodes is valid for profiling
+        callipers.
+        :param nodes: a node or list of nodes to be instrumented with
+                      profiling.
+        :type nodes: :py:class:`psyclone.psyir.nodes.Node` or
+                     list[:py:class:`psyclone.psyir.nodes.Node`]
+        :raises TransformationError: if the supplied region contains an
+                                     EXIT statement.
+        """
+        # Find all the codeblocks.
+        node_list = self.get_node_list(nodes)
+        for node in node_list:
+            codeblocks = node.walk(CodeBlock)
+            for block in codeblocks:
+                if isinstance(block._fp2_nodes[0], Exit_Stmt):
+                    raise TransformationError(
+                        "Cannot apply the ProfileTrans to a code region "
+                        "containing a Fortran EXIT statement."
+                    )
+
+        super().validate(nodes, options)
--- a/src/psyclone/tests/psyir/transformations/profile_test.py
+++ b/src/psyclone/tests/psyir/transformations/profile_test.py
@@ -36,6 +36,7 @@
 # Modified by A. R. Porter, STFC Daresbury Lab
 # Modified by I. Kavcic, Met Office
 # Modified by S. Siso, STFC Daresbury Lab
+# Modified by A. B. G. Chalk, STFC Daresbury Lab
 
 ''' Module containing tests for generating monitoring hooks'''
 
@@ -815,3 +816,23 @@ def test_auto_invoke_empty_schedule(capsys):
     _, err = capsys.readouterr()
     assert ("Not adding profiling to routine 'work1' because it does not "
             "contain any statements." in err)
+
+
+def test_profiling_exit_statement(fortran_reader):
+    ''' Check the profiling transformation validation fails if there is an
+    EXIT block in the region.'''
+
+    code = """subroutine a()
+        integer :: i
+        do i = 1, 100
+          EXIT
+        end do
+    end subroutine a
+    """
+    psyir = fortran_reader.psyir_from_source(code)
+
+    ptrans = ProfileTrans()
+    with pytest.raises(TransformationError) as excinfo:
+        ptrans.validate(psyir.children[0].children[0])
+    assert ("Cannot apply the ProfileTrans to a code region containing a "
+            "Fortran EXIT statement." in str(excinfo.value))


src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +78,27 @@ class ProfileTrans(PSyDataTrans):
...
+                     list[:py:class:`psyclone.psyir.nodes.Node`]
+        :raises TransformationError: if the supplied region contains an
+                                     EXIT statement.
+        """
+        # Find all the codeblocks.
+        node_list = self.get_node_list(nodes)
+        for node in node_list:
+            codeblocks = node.walk(CodeBlock)
+            for block in codeblocks:
+                if isinstance(block._fp2_nodes[0], Exit_Stmt):
```
sergisiso, line 98: Would it be safer to just exclude any region with a Codeblock with:
```
excluded_node_types = (Return, Codeblock)
```
Or is this too strict?

In addition to EXIT can think of at least GOTO, ENTRY and any labelled statement that means control flow jumps into the middle or to the outside of this region and therefore miss one of the surrounding hooks.

sergisiso (changes requested):
@LonelyCat124 See question below

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Update to handle GOTO and labelled though. Routines containing an ENTRY already just result in a codeblock so can't be profiled anyway.

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -33,12 +33,16 @@
...
 # Modified by R. W. Ford, A. R. Porter and S. Siso, STFC Daresbury Lab
+# Modified by A. B. G. Chalk, STFC Daresbury Lab
 # -----------------------------------------------------------------------------
 
 '''This module provides the Profile transformation.
 '''
 
-from psyclone.psyir.nodes import Return, ProfileNode
+from fparser.two.Fortran2003 import (Comment, Exit_Stmt,
+                                     Goto_Stmt, BlockBase, StmtBase)
```
sergisiso: I am going to ask for a bit more refactoring.

The idea of the frontend/backend is that we isolate the PSyIR and transformations from the parser itself, so the users are not exposed to the specifics of the parser, and this component can be changed/updated easily.

However, I agree that sometimes with CodeBlock it make sense to look deeper. But I would like to keep this encapsulated inside CodeBlock. We already have the "get_symbols_names" that explore the fparser tree. We could add a "potential_control_flow_jump" property or similar where we check for this fparser items.

This has some advantages:
- Is encapsulated inside CodeBlock, so if we add another parser we just need inheritance: fparserCodeBlock and otherCodeBlock, each implementing their logic but not change the rest of the code.
- It may be useful in other places. For instance, I think for the DefUseChains we also were quite strict about CodeBlocks, but in reality what we want to know is if it affects the control-flow or not. I can image that other transformations have the same requirement in their validation but we have been lax about it.

I don't mind if we lose ProfilingTrans error message granularity for this. Maybe a single TransformationError that prints the fparser line that is problematic would be enough.

What do you think?

sergisiso (changes requested):
@LonelyCat124 See comment below

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Yeah I think I'm happy with this.

LonelyCat124: @sergisiso I currently am printing the `.debug_string` for the offending codeblock, but I'd like something more fine grained (or at least without the CodeBlock's comment) - is there a way to do this easily?

LonelyCat124:
@sergisiso ready for another look - only thing I'm not sure about is using `debug_string` as opposed to something neater.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -42,6 +42,8 @@
 from typing import List
 
 from fparser.two import Fortran2003
+from fparser.two.Fortran2003 import (BlockBase, Comment, Exit_Stmt,
+                                     Goto_Stmt, StmtBase)
```
sergisiso: Not a big deal but I would prefer that fparser is used in a consistent style in the two methods, I slightly prefer the "from fparser.two import Fortran2003; ... Fortran2003.Node ... " as out-of-context some of the names could be confused to be PSyIR node names. So can you remove this import and use the Fortran2003.* style.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -238,3 +240,31 @@ def reference_accesses(self) -> VariablesAccessMap:
 
     def __str__(self):
         return f"CodeBlock[{len(self._fp2_nodes)} nodes]"
+
+    def is_potential_control_flow_jump(self) -> bool:
+        """
```
sergisiso: For now we use single quotes for docstrings (that may change is we adopt a formatter)

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -238,3 +240,31 @@ def reference_accesses(self) -> VariablesAccessMap:
 
     def __str__(self):
         return f"CodeBlock[{len(self._fp2_nodes)} nodes]"
+
+    def is_potential_control_flow_jump(self) -> bool:
```
sergisiso: Maybe s/is/has/ as we look for anything inside?

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -238,3 +240,31 @@ def reference_accesses(self) -> VariablesAccessMap:
...
+        """
+        :returns: whether this CodeBlock contains a potential control flow
+                  jump, e.g. GOTO, EXIT or a labeled statement.
+        """
+        # Loop over the fp2_nodes and check if any are GOTO, EXIT or
+        # labelled statements
+        for child in self._fp2_nodes:
```
sergisiso: This would only check the top-level nodes I believe, we want "walk" all nodes inside _fp2_nodes.

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,31 @@ class ProfileTrans(PSyDataTrans):
 
     def __init__(self):
         super().__init__(ProfileNode)
+
+    def validate(self, nodes, options=None):
+        """
```
sergisiso: Single quotes to keep consistent style

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,31 @@ class ProfileTrans(PSyDataTrans):
...
+        # flow jump.
+        node_list = self.get_node_list(nodes)
+        for node in node_list:
+            codeblocks = node.walk(CodeBlock)
+            for block in codeblocks:
+                if block.is_potential_control_flow_jump():
+                    raise TransformationError(
+                        f"Cannot apply the ProfileTrans to a code region "
+                        f"containing a potential control flow jump. "
+                        f"Found '{block.debug_string()}'")
```
sergisiso: I don't know how to easily be more accurate about the specific node without breaking the fparser encapsulation inside CodeBlock, so I think this is fine. But I would add a `\n` before the code, as itself can contain linebreaks and it feels odd to leave the first line inlined.

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,31 @@ class ProfileTrans(PSyDataTrans):
...
+                                     potential control flow jump, e.g.
+                                     EXIT or GOTO.
+        """
+        # Find all the codeblocks and check if they contain a control
+        # flow jump.
+        node_list = self.get_node_list(nodes)
+        for node in node_list:
+            codeblocks = node.walk(CodeBlock)
+            for block in codeblocks:
+                if block.is_potential_control_flow_jump():
```
sergisiso: I suppose this can have false negatives (e.g. the exit statement breaks from one loop, but the profiling hooks are outside the loop). I was going to request to add a "force" option. But then I realise that I have been using the profiling with for all NEMO code without an issue, and I know it contains many EXITS.

This change means that `psyclone --profile routine ...` will now have many false negatives, essentially any subroutine with a control flow jump inside it. But these are not a problem for the typical use of adding hooks at the begging and the end of a routine (multi-entry and multi-exit are covered by return and full routine codeblock, fortran gotos are to locally-scoped labels, ...).

In fact I am wondering if the original reported issue was adding the hooks in the wrong place. But in any case, the current implementation could be too overzealous, could we exclude this check then the transformation is provided a whole routine (which I think can be when we provide a routine node, or a list of nodes that contains all children of a routine). Additionally add the 'force' option.

sergisiso (changes requested):
Thanks @LonelyCat124 . This is getting closer, see inline comments. Is the PR still "towards" and not "closes"?

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Done

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Done

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Yeah thats better.

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Yeah I wasn't sure about that. fixed now.

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Done.

LonelyCat124:
@sergisiso Ready for another look.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -238,3 +238,36 @@ def reference_accesses(self) -> VariablesAccessMap:
...
+                    # An instance of BlockBase describes a block of code
+                    # so we have to examine the first statement within it.
+                    # We must allow for the case where the block is empty
+                    # though.
+                    if (child.content and child.content[0] and
+                        (not isinstance(child.content[0],
+                                        Fortran2003.Comment)) and
+                        child.content[0].item and
+                            child.content[0].item.label):
+                        return True
```
sergisiso: I am not sure how this part of fparser work but is the Fortran2003.BlockBase condition necessary? Now that we do the walk wouldn't the "so we have to examine the first statement" already be done by the walk to Fortran2003.StmtBase?

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,54 @@ class ProfileTrans(PSyDataTrans):
...
+        Checks that the supplied list of nodes is valid for profiling
+        callipers.
+
+        :param nodes: a node or list of nodes to be instrumented with
+                      profiling.
+        :type nodes: :py:class:`psyclone.psyir.nodes.Node` or
+                     list[:py:class:`psyclone.psyir.nodes.Node`]
+        :param bool options["force"]: whether to ignore potential control
+                                      flow jumps when applying this
+                                      transformation. Default is False.
```
sergisiso, line 92: Shouldn't this be a keyword argument now :)

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,54 @@ class ProfileTrans(PSyDataTrans):
...
+
+        :raises TransformationError: if the supplied region contains a
+                                     potential control flow jump, e.g.
+                                     EXIT or GOTO.
```
sergisiso: ... that could leave unclosed hooks.

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,54 @@ class ProfileTrans(PSyDataTrans):
...
+            options = {}
+        forced = options.get("force", False)
+        if forced:
+            super().validate(nodes, options)
+            return
+        # Find all the codeblocks and check if they contain a control
+        # flow jump.
+        node_list = self.get_node_list(nodes)
+        # If the node_list is the same as a whole routine then we allow
+        # this transformation.
```
sergisiso: s/we allow this transformation/we skip the check for internal control flow jumps/

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,54 @@ class ProfileTrans(PSyDataTrans):
...
+                                     EXIT or GOTO.
+        '''
+        if not options:
+            options = {}
+        forced = options.get("force", False)
+        if forced:
+            super().validate(nodes, options)
+            return
+        # Find all the codeblocks and check if they contain a control
+        # flow jump.
```
sergisiso: This is in the wrong place now, it should be in the for loop below

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,54 @@ class ProfileTrans(PSyDataTrans):
...
+        node_list = self.get_node_list(nodes)
+        # If the node_list is the same as a whole routine then we allow
+        # this transformation.
+        parent = node_list[0].parent
+        if (isinstance(parent, Routine) and
+                len(parent.children) == len(node_list)):
+            for i, child in enumerate(parent.children):
+                if child is not node_list[i]:
+                    break
```
sergisiso: CodeCov says this is not being tested.

src/psyclone/psyir/transformations/profile_trans.py
```diff
@@ -75,3 +77,54 @@ class ProfileTrans(PSyDataTrans):
...
+                super().validate(nodes, options)
+                return
+
+        for node in node_list:
+            codeblocks = node.walk(CodeBlock)
+            for block in codeblocks:
+                if block.has_potential_control_flow_jump():
+                    raise TransformationError(
+                        f"Cannot apply the ProfileTrans to a code region "
+                        f"containing a potential control flow jump. "
```
sergisiso: Maybe "containing potential internal control flow jumps as this could leave unclosed hooks."
(if you think that 'unclosed hooks' is a concept clear to understand)

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: No - that will be the job of #3004 - changing this to a keyword argument has too many knock on effects.

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: I'm not sure to be honest, I took this out of the fparser2.py code, I'll try without it and see.

LonelyCat124: The tests still pass without this so I'm going to remove it. I think the walk is sufficient.

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: I added something to make it clearer.

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Moved.

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: This code is removed, and a comment explaining why the if condition is sufficient.

src/psyclone/psyir/transformations/profile_trans.py
LonelyCat124: Added more to explain this better.

arporter:
A quick note of encouragement - this PR fixes the issues I saw when attempting to profile NEMOv5 with Chris' sea-ice refactoring :-)

sergisiso (approved):
Thanks for doing the changes @LonelyCat124, I like this implementation with the encapsulated "has_potential_control_flow_jump" much more, this is ready to merge.

I will close the issue and rename the other PR, as this already fully fixes the issue.

PR #3011, opened 2025-06-02
(Closes #2968) operation operands property

### LINKED ISSUE #2968: Add an `operands` accessor (or similar) to `Operation` subclasses
A lot of the PSyIR now has ways to access children explicitly, instead of relying on `.children[...]`, e.g. `dir_body`, `schedule`, or `lhs/rhs`.

The Operation classes don't have this which feels a bit outdated, so we should add it at some point.

LonelyCat124:
Small PR to add an explicit way to access the chidlren of an operation, rather than having to access the children list.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   32) PR #3011 for #2968. Add semantic navigation methods to Operations and
+   update indices/arguments/clauses methods to return tuples.
+
    31) PR #2998 for #2620. Refactor reference_accesses method.
 
    30) PR #2959. Update spack and enable NEMO multi-gpu.


src/psyclone/psyir/nodes/operation.py
```diff
@@ -203,6 +204,13 @@ def datatype(self):
         '''
         return self.children[0].datatype
 
+    @property
+    def operands(self) -> DataNode:
```
sergisiso: Should this return a tuple of 1-element to match the other "operands"? otherwise we would need to do an isinstance before calling it for each generic operation.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -488,6 +496,13 @@ def datatype(self):
                  argtypes[1].shape)
         return ArrayType(base_type, shape=shape)
 
+    @property
+    def operands(self) -> Tuple[DataNode, DataNode]:
```
sergisiso, line 500: I notice that this is different than other semantic navigation properties where we return a list, e.g.:
```
    @property
    def arguments(self) -> List[DataNode]:
        '''
        :returns: the children of this node that represent its arguments.
        :rtype: list[py:class:`psyclone.psyir.nodes.DataNode`]
        '''
        if len(self._children) >= 2:
            return self.children[1:]
        return []

```
But I like the Tuple that you are returning much more because if doesn't allow re-assignment
```
>>> a = (3, 4)
>>> a[1] = 2
  File "<stdin>", line 1, in <module>
TypeError: 'tuple' object does not support item assignment
>>> del a[1]
  File "<stdin>", line 1, in <module>
TypeError: 'tuple' object doesn't support item deletion
```
Which is useful as a common mistake I do is things like:
`call.arguments[3] = Reference(symbol)`
which ends up doing nothing to the tree but not failing, since it only modifies a temporary list slice.

With a tuple it would fail and make it easier to see that `calls.arguments[3].replace_with(Rerefence(symbo)) or .detach()` is what we need.

Should be update all the semantic navigation methods that return list of nodes to return tuples?

sergisiso (changes requested):
@LonelyCat124 I like it, in fact I think that other semantic navigation methods should also look like this, see inline comments.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: I'm happy to for this case since its a shared class (though you'd probably need to check anyway since they take different numbers of operands).

sergisiso: The other option is to call it operand (singular) so Unary and Binary have different sematic navigation methods, and return different things. Would that be more useful in practice?

I think this methods are quite useful with destructuring:
`lhs, rhs = bop.operands`

But I am not sure this is nice for the 1-element tuple case I was asking.

sergisiso: @arporter opinions?

arporter: From a quickish grep of the code, the code that handles UnaryOperations seems to be distinct from that for BinaryOperations and therefore it's probably best to have `operand` and `operands` as that will avoid the tuple-unpacking error that could result if we try to do the `lhs, rhs = bop.operands` when bop is actually unary. I think we probably have to try it and see.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: This returns a "tuple" because it just returns 2 elements (which python packs to be a tuple by default).

I've changed a few other methods now that returned list of nodes to return tuples, and updated the codes and tests accordingly.

LonelyCat124:
@sergisiso ready for another look.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: Ah I missed this converation. I'll switch it to operand now.

LonelyCat124: Done now.

src/psyclone/psyir/nodes/operation.py
```diff
@@ -488,6 +496,13 @@ def datatype(self):
...
+        '''
+        :returns: the operand of this UnaryOperation.
```
sergisiso: "operands" and "BinaryOperation"

src/psyclone/tests/psyad/domain/lfric/test_lfric_adjoint_harness.py
```diff
@@ -269,14 +269,14 @@ def test_init_fields_random_vector(type_map):
         assert isinstance(kernels[kidx], LFRicBuiltinFunctor)
         assert kernels[kidx].symbol.name == "setval_random"
         assert kernels[kidx].children[0].symbol.name == "field1"
-        assert kernels[kidx].children[0].indices == [lit]
+        assert list(kernels[kidx].children[0].indices) == [lit]
```
sergisiso: It would be better to update the other side of the assert to the expected expression `(lit, )` , also below

src/psyclone/tests/psyir/nodes/array_reference_test.py
```diff
@@ -467,7 +467,7 @@ def test_array_indices():
     array = ArrayReference.create(DataSymbol("test",
                                              ArrayType(REAL_TYPE, [10])),
                                   [one])
-    assert array.indices == [one]
+    assert array.indices[0] is one
```
sergisiso: Same here and in the rest of the test changes, we should be updating the other side of the assert to tuples (e.g. ` == (one,)` instead of the return value

sergisiso (changes requested):
Looks good @LonelyCat124, there are just minor modification needed now and in the documentation we have a list all the semantic navigation methods at `user_guide/psyir.rst::189`, we need to add the new methods there. The Call, routine/arguments methods are also missing.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: done

src/psyclone/tests/psyad/domain/lfric/test_lfric_adjoint_harness.py
LonelyCat124: Argh thats annoying, I tried that initially but just did `(lit)` which obviouslu wasn't a tuple. I'll adjust

src/psyclone/tests/psyir/nodes/array_reference_test.py
LonelyCat124: Done.

LonelyCat124:
@sergisiso ready for another look.

sergisiso (approved):
All looks good now, thanks @LonelyCat124 (the linkcheck broken links are not this PRs fault)

PR #3013, opened 2025-06-03
(closes #2856) Configurable intrinsics available on device

### LINKED ISSUE #2856: Add support for excluding maths functions from GPU offload regions
During the NG-Arch NEMO hackathon we concluded that some numerical differences (between the results from NEMO on CPU vs GPU) result from differences in implementation of certain maths functions on those architectures. However, in order to test for correctness of PSyclone transformations, we really want to be able to be confident that any numerical differences are arising from errors. Therefore, it would be good to have support in PSyclone for an option to exclude certain maths functions from being offloaded to the GPU. This would make *testing* much easier. Note, this is not intended to be something that would be used in production.

sergisiso:
It turns out that the performance degradation of NEMO BENCH with the new compiler is not due to the new excluded files, but due to the new excluded intrinsic REAL. This is only necessary for the reproducible run, so I introduced a "device-id" option to some of the transformation that mark regions of code for the GPU to decide what intrinsics to permit using this device string. 

At the moment there is only a nvfortran-all (what we should use for performance results) and a nvfortran-repr that we should use for numerically reproducible results. But this now can be easily extensible for other compiler/architectures/feature-sets, e.g. amd/cray/gnu.

Sine I am building NEMO BENCH twice, I thought it would also be useful to keep the profiling hooks in one of them.

The final result for NEMO bench is faster than it use to be even before excluding REAL, but this probably because now I use optimising flags for the build to get the performance timings.


sergisiso:
@arporter This is ready for a first review

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -97,6 +100,7 @@
     "trczdf.f90",
     "trcice_pisces.f90",
     "dtatsd.f90",
+    "trcatf.f90",
```
arporter, line 103: Just to check, is this new one also a signal 11 issue?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -770,44 +770,23 @@ def intrinsic(self):
...
-    # Currently it is inspired from: https://docs.nvidia.com/hpc-sdk/
-    # compilers/hpc-compilers-user-guide/#acc-fort-intrin-sum
-    # But that list is incomplete (e.g. SUM is supported and not listed)
-    def is_available_on_device(self):
+    def is_available_on_device(self, device_string: str = "") -> bool:
         '''
+        :param device_string: optional string to identify the offloading
+            device (or its compiler-platform family).
         :returns: whether this intrinsic is available on an accelerated device.
-        :rtype: bool
```
arporter, line 778: Please doc the ValueError.

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -158,11 +158,12 @@ def validate(self, node, options=None):
...
+        device_string = options.get("device_string", "") if options else ""
         node_list = self.get_node_list(node)
         super().validate(node, options)
         for node in node_list:
             for call in node.walk(Call):
-                if not call.is_available_on_device():
+                if not call.is_available_on_device(device_string):
                     raise TransformationError(
                         f"'{call.routine.name}' is not available on the "
                         f"accelerator device, and therefore it cannot "
```
arporter: Include `device_string` in this message, e.g. "with device 'nvidia-repr'"? We should perhaps s/nvidia-repr/nvidia-reproducible/ in the interests of readability?

src/psyclone/transformations.py
```diff
@@ -378,6 +378,8 @@ def validate_it_can_run_on_gpu(self, node, options):
         :type options: Optional[Dict[str, Any]]
         :param bool options["force"]: whether to allow routines with
             CodeBlocks to run on the GPU.
+        :param bool options["device_string"]: provide a compiler-platform
+            identifier strign.
```
arporter: "identifier." and s/bool/str/

src/psyclone/transformations.py
```diff
@@ -488,7 +491,7 @@ def validate_it_can_run_on_gpu(self, node, options):
 
         calls = kernel_schedule.walk(Call)
         for call in calls:
-            if not call.is_available_on_device():
+            if not call.is_available_on_device(device_string):
                 call_str = call.debug_string().rstrip("\n")
                 raise TransformationError(
                     f"{k_or_r} '{node.name}' calls another routine "
```
arporter, line 509: Please extend the error message both here and at L1776 to mention the value of 'device_str' if it is set.

src/psyclone/transformations.py
arporter: Please could you also describe this new option in the class doc string in order to publicise it.

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -158,11 +158,12 @@ def validate(self, node, options=None):
         :raises TransformationError: if its a function and the target region
             attempts to enclose the assingment setting the return value.
         '''
+        device_string = options.get("device_string", "") if options else ""
```
arporter, line 163: Please document this new options argument and mention it in the apply docstring too. How much work would it be to convert this transformation to use the new kwargs form for options?

arporter (changes requested):
Looks good Sergi. I did debate whether the device identifier should be a configuration option but then I thought that it's probably best to have it in the same place as the transformation logic. It's a similar concept to the 'reproducible OMP reductions' option that we have already although that isn't compiler/device specific.

Aside from that, I'm wondering how much work it would be to update one or both of the transformations affected by this to use the new kwargs form for options?

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: Yes, there is definitely something we do wrong in BENCH, this file gave me a signal 11. It could be because this parallelises more code as with LOG10 and REAL we now do more outer loops which the REPRODUCIBLE run excludes.

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: Done

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: I agree with the readability, but I wasn't happy about calling it reproducible because it is not quite right (nvfortran-all is reproducible between GPU runs, the nvfortran-repr was not reproducible unless using math_uniform flag, it doesn't guarantee reproducibility if there is atomics/reductions). So, I changed it to "nvfortran-uniform" to use the same semantic as the compiler and hopefully makes clearer that it needs to be combined with math_uniform.

I have not changed the NEMO script envvar because in this case we do target for reproducibility, using the uniform intrinsics is just one aspect of it, but it also should prevent the atomics/reduction which we don't have yet.

src/psyclone/transformations.py
sergisiso: Good point.

I attempted moving the options to kwargs but being in a mixin and also the transformation that use it are in a hierarchy with options in other classes it propagates to many places, and I think it will be better to do it in a more focused PR.

src/psyclone/transformations.py
sergisiso: Done

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Added the docstring. Converting to kwargs got quite complicated so I would prefer to do it separately.

src/psyclone/psyir/transformations/omp_target_trans.py
arporter: Thanks, I like the new naming. Please could you update the error message to include it.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -161,17 +161,17 @@ def test_intrinsiccall_is_available_on_device(intrinsic, result):
 
 def test_intrinsiccall_is_available_on_device_with_device_string():
     '''Tests that the is_available_on_device() method with a device_string
-    argument provides different results with the 'nvfortran-reprod'
+    argument provides different results with the 'nvfortran-uniformod'
```
arporter: "uniform"

src/psyclone/transformations.py
arporter: OK. I thought that might be the case.

src/psyclone/transformations.py
```diff
@@ -1773,6 +1786,14 @@ def validate(self, node_list, options=None):
         for node in node_list:
             for call in node.walk(Call):
                 if not call.is_available_on_device(device_string):
+                    if isinstance(call, IntrinsicCall):
+                        dstr = (f"in '{device_string}'" if device_string else
```
arporter: Please fix the coverage of L1790 and 1792

src/psyclone/tests/psyir/transformations/kernel_transformation_test.py
```diff
@@ -429,10 +429,9 @@ def test_gpumixin_validate_no_call():
         IntrinsicCall.create(IntrinsicCall.Intrinsic.GET_COMMAND))
     with pytest.raises(TransformationError) as err:
         rtrans.validate(kernel)
-    assert ("Kernel 'testkern_with_call_code' calls another routine "
-            "'GET_COMMAND()' which is not available on the accelerator device "
-            "and therefore cannot have ACCRoutineTrans applied to it "
-            "(TODO #342)."
+    assert ("Kernel 'testkern_with_call_code' calls intrinsic 'GET_COMMAND' "
+            "which is not available by default. Use the 'device_string' "
```
arporter: I'm a bit worried that we've lost "accelerator" from this message. What about "...which is not available on the 'default' accelerator device. Use ..."?

src/psyclone/tests/psyir/transformations/kernel_transformation_test.py
```diff
@@ -475,15 +474,15 @@ def test_kernel_gpu_annotation_device_id(rtrans, fortran_reader):
     # But not if we are targeting "nvidia-repr" or an invalid device
     with pytest.raises(TransformationError) as err:
         rtrans.validate(routine, options={'device_string':
-                                          'nvfortran-repr'})
-    assert ("routine 'myfunc' calls another routine 'REAL(a)' which is not "
-            "available on the accelerator device"
-            in str(err.value))
+                                          'nvfortran-uniform'})
+    assert ("routine 'myfunc' calls intrinsic 'REAL' which is not available in"
+            " 'nvfortran-uniform'. Use the 'device_string' option to specify a"
```
arporter: Same here?

arporter (changes requested):
I like the new names. There's just a bit more tidying to do and then this can go on.
Documentation builds OK. The new concept of "device_str" feels important but I think we can decide how we want to document it in a future PR.

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Done

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
sergisiso: ups, fixed

src/psyclone/tests/psyir/transformations/kernel_transformation_test.py
sergisiso: Good point, done

src/psyclone/tests/psyir/transformations/kernel_transformation_test.py
sergisiso: Done

sergisiso:

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -168,8 +168,10 @@ def validate(self, node, options=None):
                 if not call.is_available_on_device(device_string):
                     raise TransformationError(
                         f"'{call.routine.name}' is not available on the "
-                        f"accelerator device, and therefore it cannot "
-                        f"be called from within an OMP Target region.")
+                        f"default accelerator device, and therefore it cannot "
```
arporter: text = device_string if device_string else "default"
and then s/default/'{text}'/

src/psyclone/tests/psyir/transformations/kernel_transformation_test.py
```diff
@@ -475,9 +475,9 @@ def test_kernel_gpu_annotation_device_id(rtrans, fortran_reader):
     with pytest.raises(TransformationError) as err:
         rtrans.validate(routine, options={'device_string':
                                           'nvfortran-uniform'})
-    assert ("routine 'myfunc' calls intrinsic 'REAL' which is not available in"
-            " 'nvfortran-uniform'. Use the 'device_string' option to specify a"
-            " different device." in str(err.value))
+    assert ("routine 'myfunc' calls intrinsic 'REAL' which is not available on"
+            " the 'nvfortran-uniform' device. Use the 'device_string' option "
```
arporter: Better to have "...uniform' accelerator device." I think.

src/psyclone/psyir/transformations/omp_target_trans.py
arporter: i.e. the same as you do in transformations.py

src/psyclone/transformations.py
```diff
@@ -493,8 +493,10 @@ def validate_it_can_run_on_gpu(self, node, options):
         for call in calls:
             if not call.is_available_on_device(device_string):
                 if isinstance(call, IntrinsicCall):
-                    device_str = (f"in '{device_string}'" if device_string else
-                                  "by default")
+                    if device_string:
+                        device_str = f"on the '{device_string}' device"
```
arporter: "accelerator device" I think.

src/psyclone/transformations.py
```diff
@@ -1787,11 +1789,13 @@ def validate(self, node_list, options=None):
             for call in node.walk(Call):
                 if not call.is_available_on_device(device_string):
                     if isinstance(call, IntrinsicCall):
-                        dstr = (f"in '{device_string}'" if device_string else
-                                "by default")
+                        if device_string:
+                            device_str = f"on the '{device_string}' device"
```
arporter: "accelerator device" again.

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/kernel_transformation_test.py
sergisiso: Done

src/psyclone/transformations.py

src/psyclone/transformations.py

arporter:
Sorry Sergi but the NEMOv5 integration test failed with a Signal 11 :-(

sergisiso:
@arporter I click re-run and the test succeeded, I think it was a pre-existing issue that does not manifest every time. Also the file that failed does not have any of the intrinsics configured by this PR.

arporter (approved):
Thanks Sergi. I'll proceed to merge.
Doc test failure was a link-check problem with the GOcean page so nothing to do with this PR.

PR #3018, opened 2025-06-05
(Closes #3017) fix bug when getting ubound of ArrayMixin

### LINKED ISSUE #3017: Bug in ArrayMixin._get_bound_expression()
It turns out that it is possible for this method to sometimes return an `ArrayType.Extent` which it's not supposed to:
https://github.com/stfc/PSyclone/blob/9328e6bc4cea8c1985f7636b3ddc5175ffd8a3dc/src/psyclone/psyir/nodes/array_mixin.py#L256-L262
In this code, we also need to check that xxx.upper is not of type `ArrayType.Extent`. If it is, we should not return but carry on to the rest of the routine.

arporter:


arporter:
Integration tests were all green. Ready for review from @LonelyCat124 @sergisiso or @hiker :-)

src/psyclone/tests/psyir/nodes/array_mixin_test.py
```diff
@@ -431,6 +436,14 @@ def test_get_bound_expression():
     assert ub2 is not ubnd_ref
     assert ub2 == ubnd_ref
 
+    # An ArrayType where the upper bound is unknown.
+    symbol2 = DataSymbol("other", ArrayType(INTEGER_TYPE,
+                                            [(1, ArrayType.Extent.ATTRIBUTE)]))
```
sergisiso, line 441: Reading the ArrayType docstring this is not a valid shape, it should be just `[ArrayType.Extent.ATTRIBUTE]`

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -259,7 +259,11 @@ def _get_bound_expression(self, pos, bound):
             # We have the full type information and the bound is known.
             if bound == "lower":
                 return cursor_type.shape[pos].lower.copy()
-            return cursor_type.shape[pos].upper.copy()
+            # If the upper bound is required and is of ArrayType.Extent type
+            # then we'll have to proceed to construct a call to the UBOUND
+            # intrinsic.
+            if not isinstance(cursor_type.shape[pos].upper, ArrayType.Extent):
```
sergisiso, line 265: Reading the ArrayBounds dataclass docstring "upper" can only return a "DataNode"

arporter:
I don't think it will affect the generated code - the crash was in PSyclone itself.

arporter:
I'm thinking that it might be best to keep the implementation the way it is and update the docstrings to say that `ArrayBounds.upper` is permitted to be `ArrayType.Extent.ATTRIBUTE`. This will be less error prone than having `None` mean the same thing. What do you think @sergisiso?

arporter:
> And with the static information we have here it should. Maybe the kttrd and the nis0 - 0/njs0 - 0 match internally, but psyclone cannot know.

Actually, I'm not sure I follow this logic. The original code has, e.g.:
```fortran
trdt(ntsi-(0):ntei+(0),ntsj-(0):ntej+(0),:) = ptrd(ntsi-(0):ntei+(0),ntsj-(0):ntej+(0),:) * tmask(ntsi-(0):ntei+(0),ntsj-(0):ntej+(0),:)
```
and we've used the limits in the slice expressions to set the start and stop values of the loops. That looks correct as all we've done is copy the implementation?

src/psyclone/psyir/nodes/array_mixin.py
arporter: I've updated the docstring now.

src/psyclone/tests/psyir/nodes/array_mixin_test.py
arporter: I've updated the docstring.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -203,19 +203,19 @@ def is_lower_bound(self, index):
         '''
         return self._is_bound(index, "lower")
 
-    def _get_bound_expression(self, pos, bound):
+    def _get_bound_expression(self, pos: int, bound: str):
```
sergisiso, line 206: There are these similar methods that will need the same fix:
- reference2arrayrange_trans.py::_get_array_bound
- dotproduct2code_trans.py::_get_array_bound
- matmul2code_trans.py::_get_array_bound

But also, could we bring them into a single implementation? Note that those are for the symbol bounds (the type), and this one here for array expression. So, its two different methods, but the later can call the former for the `isinstance(cursor_type, ArrayType)` condition, which is the modified part. 

Maybe the _get_array bound can be inside the Array type itself `def shape_to_bounds(for_symbol)`, so the logic is near the shape definition?

Regardless of what you want to do, could you also sprinkle some test cases with "lower-bound-defined, upper-bound empty" in the tree transformations and any other place where it could matter. I admit that I have not considered these before.

sergisiso (changes requested):
I have not look at the previous changes yet, but I found similar methods with the same issue, could you have a look.

src/psyclone/psyir/nodes/array_mixin.py
arporter: Oh, good spot. I've added `DataSymbol.get_bounds()` as this worked a bit better than adding it to the ArrayType.

arporter: I've also added a few new tests.

doc/user_guide/psyir.rst
```diff
@@ -300,27 +300,31 @@ about its extent. It is necessary to distinguish between four cases:
 |                                            | values                         |
 +--------------------------------------------+--------------------------------+
 |An array has an extent defined by another   | ``ArrayType.ArrayBounds``      |
-|symbol or (constant) PSyIR expression.      | containing ``Reference`` or    |
-|                                            | ``Operation`` nodes            |
+|symbol or (constant) PSyIR expression.      | containing ``DataNode``\s (i.e.|
+|                                            | ``Reference`` or ``Operation``)|
```
sergisiso: Why do we distinguish these from the previous row, they could both have compile-time known parameters and the sympy equality should allow the same transformation/validations that with Literals.

Maybe we could merge them and just say that this are explicit:

"An array that has explicit bound expressions | ``ArrayType.ArrayBounds`` containing two PSyIR DataNodes"

which also matches the Fortran explicit-shape concept.

doc/user_guide/psyir.rst
```diff
@@ -300,27 +300,31 @@ about its extent. It is necessary to distinguish between four cases:
...
+|at runtime.                                 | with the lower bound given by a|
+|                                            | ``DataNode`` and the upper     |
+|                                            | bound  by                      |
+|                                            | ``ArrayType.Extent.ATTRIBUTE`` |
 +--------------------------------------------+--------------------------------+
 |It is not known whether an array has memory | ``ArrayType.Extent.DEFERRED``  |
 |allocated to it in the current scoping unit.|                                |
 +--------------------------------------------+--------------------------------+
 
 where ``ArrayType.ArrayBounds`` is a ``namedtuple`` with ``lower`` and
```
sergisiso: It's a dataclass now

src/psyclone/tests/psyir/symbols/datasymbol_test.py
```diff
@@ -582,3 +584,71 @@ def test_datasymbol_reference_accesses():
...
+    atype = ArrayType(INTEGER_SINGLE_TYPE,
+                      [ArrayType.Extent.ATTRIBUTE])
+    utype = UnsupportedFortranType("integer, dimension(:), pointer :: var",
+                                   partial_datatype=atype)
+    sym = DataSymbol("c", utype)
+    lbnd, ubnd = sym.get_bounds(0)
+    assert isinstance(lbnd, IntrinsicCall)
+    assert lbnd.intrinsic == IntrinsicCall.Intrinsic.LBOUND
+    assert isinstance(ubnd, IntrinsicCall)
+    assert ubnd.intrinsic == IntrinsicCall.Intrinsic.UBOUND
```
sergisiso, line 654: F2008std says L/UBOUND first argument "Shall be an array, of any type." which in this case is not.

Maybe it should also return IndexError, since it does not have any?

src/psyclone/tests/psyir/transformations/reference2arrayrange_trans_test.py
```diff
@@ -193,6 +145,19 @@ def test_multid(fortran_reader, fortran_writer):
...
+    '''Test when the underlying array is of assumed shape.'''
+    code = ('''\
+    subroutine sub(var, istart)
+      integer, intent(in) :: istart
+      integer, dimension(istart:) :: var
+      var = 1
+    end subroutine sub
+    ''')
+    result = apply_trans(fortran_reader, fortran_writer, code)
+    assert "var(:) = 1" in result
```
sergisiso, line 158: Can you move or continue this example at the arrayrange2loop, and also with a rhs with a different lower bound symbol?

doc/user_guide/psyir.rst
```diff
@@ -300,27 +300,31 @@ about its extent. It is necessary to distinguish between four cases:
...
+``upper`` member set to ``ArrayType.Extent.ATTRIBUTE``.
 
 The distinction between the last two cases is that in the former the
 extents are known but are kept internally with the array (for example
-an assumed shape array in Fortran) and in the latter the array has not
+an assumed shape array in Fortran) while in the latter, the array has not
 yet been allocated any memory (for example the declaration of an
-allocatable array in Fortran) so the extents may have not been defined
-yet.
+allocatable array in Fortran) so the extents may have not yet been defined.
```
sergisiso, line 322: I am right in thinking that "assumed-size" arrays are not supported? Is it worth adding a sentence after the example saying so and that this will currently be DataSymbols with UnsupportedFortranType?

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
```diff
@@ -311,11 +261,20 @@ def apply(self, node, options=None):
...
+        score2 = sum(1 for bnd in bounds2 if isinstance(bnd, IntrinsicCall))
+        if score1 <= score2:
+            lower_bound, upper_bound = bounds1[0], bounds1[1]
+        else:
+            lower_bound, upper_bound = bounds2[0], bounds2[1]
         # Create i loop and add the above code as a child
         iloop = Loop.create(i_loop_symbol, lower_bound.copy(),
-                            upper_bound.copy(), step.copy(), [assign])
+                            upper_bound.copy(), Literal("1", INTEGER_TYPE),
+                            [assign])
```
sergisiso: You may not need the copy since get_bounds already does it?

sergisiso (changes requested):
@arporter only a few more minor comments to address. It is good to see duplicated code being removed.

doc/user_guide/psyir.rst
arporter: Agreed and done.

doc/user_guide/psyir.rst
arporter: Good spot.

src/psyclone/tests/psyir/symbols/datasymbol_test.py
arporter: I'm afraid I don't understand - `c` is of UnsupportedType but with a partial_datatype that is an ArrayType?

src/psyclone/tests/psyir/transformations/reference2arrayrange_trans_test.py
arporter: Good idea - I've added a test to arrayassignment2loops_trans_test.py.

doc/user_guide/psyir.rst
arporter: Yes, done.

src/psyclone/psyir/transformations/intrinsics/dotproduct2code_trans.py
arporter: Oh yes. Removed.

arporter:
There was one comment I didn't understand but I've now done everything else. Ready for another look/clarification of that comment.

PR #3020, opened 2025-06-06
#3019 Added deprecated classes for old-style Dynamo0p3 transformation names

### LINKED ISSUE #3019: Create dynamop3 compatibility layer
Recently all lfric transformations have been renamed from using `Dynamo0p3` to using `LFRic` as prefix. Since this breaks all LFRic builds, we need to create a (temporary) compatibility layer that provides the old names with a deprecation warning.

hiker:
Fixes #3019.

Note that I uses `exec`. While the classes could be created using `type`, this would (as far as I can see) require to define a dedicated ``__new__`` function (with any name) to be assigned to the `__new__` attribute, which seems messy. Given the limited time that we need these classes, I stayed with `exec` (which is used elsewhere as well :) ).

For the tests, I do all the imports in the test function, so it's easy to remove this test later.


arporter (commented):
Thanks Joerg. All looks good.
I've tried it with a modified version of examples/lfric/eg3 and the deprecation warning was printed as expected.
Just to be on the safe side I'll run the compilation and LFRic integration tests. I can't see how they could fail but experience shows that's no guarantee of anything :-)

arporter (approved):
Integration tests were green.

PR #3021, opened 2025-06-06
[PSyIR Transformation] Make loop-bounds hoisting optional in chunking/tiling

mn416:
[Minor improvement]

This PR adds an option to `ChunkLoopTrans` and `LoopTiling2DTrans` that allows control over whether loop bounds are hoisted or not. For backwards compatibility, the option defaults to `True`. However, it is sometimes desirable to disable hoisting, or apply it in a separate pass using `HoistLoopBoundExprTrans`. For example, in the case of 2D tiling, it is not possible to collapse (in the OpenMP sense) the outer two loops because there is an intervening statement introduced by bounds hoisting. Similarly, reordering loops using `LoopSwapTrans` is also hindered by intervening statements between loops. It also seems reasonable to leave loop bounds hoisting to the compiler.

Background: To learn about PSyclone, I've been transforming the classic triple-nested matrix multiplication loop using `ChunkLoopTrans`, `LoopTiling2DTrans`, `LoopSwapTrans`, and `OMPLoopTrans`. These transformations play much more nicely together when loop bounds hoisting is disabled, hence the PR.


sergisiso:
Thanks @mn416 , I agree that hoisting the expressions makes the resulting code more difficult for follow up operations. And, as you pointed, we already have `HoistLoopBoundExprTrans`, I am wondering if we should remove the hoisting from chunking/tiling altogether. I need to find out why we implemented it in the first place, it may be just that we implemented this first when we still didn't have the other capabilities.

mn416:
@sergisiso Ready for review (I don't seem to have permission to add labels). Let me know if you'd like the commits squashed.

src/psyclone/psyir/transformations/chunk_loop_trans.py
```diff
@@ -40,7 +40,7 @@
 
 from psyclone.core import Signature, AccessType
 from psyclone.psyir import nodes
-from psyclone.psyir.nodes import Assignment, BinaryOperation, Reference, \
+from psyclone.psyir.nodes import BinaryOperation, Reference, \
```
sergisiso, line 44: Feel free to add you name in authors list above, L33-L36.

Also in the other files modified by this PR

src/psyclone/tests/psyir/transformations/loop_tiling_2d_trans_test.py
```diff
@@ -197,7 +195,7 @@ def test_loop_tiling_2d_trans_apply(fortran_reader, fortran_writer):
     assert expected in result
 
 
-def test_loop_tiling_2d_trans_apply_options(fortran_reader, fortran_writer):
+def test_loop_tiling_2d_trans_apply_options1(fortran_reader, fortran_writer):
```
sergisiso: Did you meant to append the '1' in the name here or is it a leftover?

sergisiso (changes requested):
Thanks for removing the hoisted references @mn416 . All tests and formatting checks succeeded now. There is only a couple of minor comments, then this can be merged.

I like your new `examples/psyir/transformations/` feel free to contribute them in a future PR.

src/psyclone/tests/psyir/transformations/loop_tiling_2d_trans_test.py
mn416: Well spotted. Removed.

src/psyclone/psyir/transformations/chunk_loop_trans.py
mn416: Done!

PR #3023, opened 2025-06-10
(closes #3012) Initial version of psyclonefc command

### LINKED ISSUE #3012: PSyclone compiler wrapper
Splitting the pre-processor from the compilation is a typical paint point of using psyclone. "cpp" can be used for custom preprocessor defines but for doing it properly (psyclone workflow equivalent to using only the compiler) is quite tricky because there is compiler-specific defines .e.g the  `__NVCOMPILER` causing the issue here, but also feature-specific `_OPENMP` or architecture-specific `__SSE3__`. Not doing the exact same thing on the cpp than in the compiler leads to very subtle differences.

When injecting psyclone to other projects we always have to go through this initial step of modifying the build systems, which is complex because each build-system is different and the previously mentioned subtleties  make it is a complex logic to get right. So for some time I thought that psyclone should have an easier way to be injected to build systems that take responsibility of this preprocessor/compiler splitting step with a compiler wrapper tool `psyclonefc` (similar to mpifort - btw mpi also adds defines that cpp misses) where the wrapper:
1) does the preprocessor with the same compiler and feature-flags that the compiler would take
2) applies psyclone
3) compiles.
This will make sure that no compiler/architecture define is lost and the logic of dealing with each compiler preprocessor are encoded into psyclone instead of repeated in each project.

Using the tool would look like this:

`FC=psyclonefc PSYCLONE_COMPILER=mpifort make`

Which inside the build system, when it find the `$FC <flags> <input.f90>` it does steps 1 and 2 for each *.f90 file and finally step 3 with all preprocessed and psycloned sources.

sergisiso:


.github/workflows/nemo_v5_tests.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         # than the latest release from pypi.
         # pip install external/fparser
         pip install .
+        git clone https://forge.nemo-ocean.eu/nemo/nemo.git nemo_5.0 --branch 5.0 --single-branch
+        git clone https://forge.nemo-ocean.eu/nemo/nemo.git nemo_main --branch main --single-branch
```
sergisiso: @arporter The tool is not complete but it is enough to psyclone both nemov5 and nemo_main without any modification to nemo (other than copying our arch file)

.github/workflows/nemo_v5_tests.yml
```diff
@@ -174,18 +176,20 @@ jobs:
...
         export FCFLAGS="-fdefault-real-8 -O2 -fcray-pointer -ffree-line-length-none -g -fopenmp"
+        export PSYCLONE_COMPILER=$MPIF90
+        export MPIF90=psyclonefc
+        export PSYCLONE_OPTS="-l output -s ${PSYCLONE_NEMO_DIR}/omp_cpu_trans.py"
 
         # Clean up and compile
         rm -rf tests/${TEST_DIR}
-        ./makenemo -r BENCH -m linux_spack -n ${TEST_DIR} -p ${PSYCLONE_NEMO_DIR}/omp_cpu_trans.py \
-          add_key "key_nosignedzero" -j 4 -v 1
+        ./makenemo -r BENCH -m linux_spack -n ${TEST_DIR} add_key "key_nosignedzero" -j 4 -v 1
```
sergisiso: And here are the instructions it needs

sergisiso:
@arporter I have scaled this PR back, make it just a towards, and mark it ready for review. The reason is that I don't think we should advertise/document it until we figure out what to do with the intercepted -J/-I and we apply the preprocessor step (these are its main value proposition). However, it can already provide value to Imaran/Chris and other internal NEMO workflows, but since these are also happening in branches it makes it hard to combine them. So, it would be convenient to introduce this barebones psyclonefc to master for internal use and finish it later.

bin/psyclonefc
```diff
@@ -0,0 +1,43 @@
...
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+
+''' Entry point for the psyclonefc command. '''
```
arporter: "...command (PSyclone's compiler wrapper script)."

examples/nemo/scripts/utils.py
```diff
@@ -175,6 +175,7 @@
 PARALLELISATION_ISSUES = [
     "ldfc1d_c2d.f90",
     "tramle.f90",
+    "traqsr.f90",
```
arporter, line 178: If we know precisely what the problem is for this file it would be good to add a comment here.

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def compiler_wrapper(arguments):
+    '''
+    PSyclone compiler wrapper.
+
+    This utility intercepts compilation commands and applies the psyclone
+    transformation tool to any encountered Fortran file before passing them
+    to the target compiler. The goal is to provide an easy way to inject
+    psyclone to build systems without needed to modify them.
```
arporter: s/to/into/

arporter: s/needed/needing/

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+
+    '''
+    fortran_compiler = os.getenv("PSYCLONE_COMPILER", default=None)
+    psyclone_options = os.getenv("PSYCLONE_OPTS", default="").split(' ')
+
+    # Validate mandatory PSYCLONE_COMPILER
+    if fortran_compiler is None:
+        sys.exit(
+            'psyclonefc error: PSYCLONE_COMPILER environment variable not '
+            'found!')
```
arporter: Could we make this a bit more helpful - perhaps "...variable must be set to the Fortran compiler to use."

arporter: While we're validating, should we also check that we can find the specified compiler so that we can fail early and provide a useful message?

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+    # Remove empty strings from the list (caused by the default empty envvar or
+    # multi-spaces gaps)
+    if "" in psyclone_options:
+        psyclone_options.remove("")
```
arporter, line 94: It turns out that `remove` only removes the *first* instance of the value you specify. If it's possible that there may be more than one empty string then this code needs extending.

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+            # TODO #3012: preprocessing is currently ignored, this is not a
+            # problem for NEMO because the build system does the proprocessor
+            # in a separate step
+
+            # 2) Run psyclone
+            stem = Path(argument).stem
+            suffix = Path(argument).suffix
+            output = f"{stem}.psycloned{suffix}"
+            psyclone_args = psyclone_options + ['-o', output, argument]
+            print("psyclone " + ' '.join(psyclone_args))
```
arporter, line 127: This could be a logger.debug(...) instead?

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+            main(psyclone_options + ['-o', output, argument])
```
arporter: just `main(psyclone_args)`?

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+
+            # 3) Replace the fortran file in the compiler command with the
+            # fortran file produced by psyclone
+            new_compiler_command.append(output)
+        else:
+            new_compiler_command.append(argument)
```
arporter, line 134: We could also use logging here to record that we've chosen not to run PSyclone since the file extension isn't one we recognise.

src/psyclone/psyclonefc_cli.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Run the new compiler command and propagate error code
+    print(' '.join(new_compiler_command))
```
arporter, line 137: Again, replace with logging?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1896,15 +1897,12 @@ def resolve_imports(self, container_symbols=None, symbol_target=None):
...
-                # been found in the given include_path or any issue has arisen
-                # during parsing.
-                # TODO #11: It would be useful to log this.
-                continue
+                external_container = None
 
             if not external_container:
-                # Failed to get a Container (possibly due to parsing or raising
-                # errors).
+                message = f"Module '{c_symbol.name}' not found"
```
arporter, line 1903: Whoop! :-)

src/psyclone/tests/psyclonefc_cli_test.py
```diff
@@ -0,0 +1,94 @@
...
+def test_psyclonefc(monkeypatch, capsys):
+    ''' Test the cli with different arguments provided. '''
+
+    # Use the true command as compiler, this will always return a successful
+    # errcode 0 and accepts any flag. Also monkeypatch the psyclone main to
+    # don't do anything. The tests use the companion print statements
+    monkeypatch.setattr(os, 'environ', {'PSYCLONE_COMPILER': 'true'})
+    monkeypatch.setattr('psyclone.psyclonefc_cli.main', lambda x: None)
+
+    # If we provide not argument it goes through without failing
```
arporter: s/not/no/

src/psyclone/tests/psyclonefc_cli_test.py
```diff
@@ -0,0 +1,94 @@
...
+    # This is a successful exit
+    assert err.value.code == 0
+
+    # Now provide more typical compilation arguments
+    with pytest.raises(SystemExit) as err:
+        compiler_wrapper(['source.f90', '-c', '-o', 'source.o'])
+    assert err.value.code == 0
+    stdout, _ = capsys.readouterr()
+    # This will execute:
+    assert "psyclone -o source.psycloned.f90 source.f90" in stdout
```
arporter, line 79: As I commented in the code, I think it will be nicer to use logging here.

src/psyclone/tests/psyclonefc_cli_test.py
```diff
@@ -0,0 +1,94 @@
...
+import os
+from psyclone.psyclonefc_cli import compiler_wrapper
+
+
+def test_psyclonefc_errors(monkeypatch):
+    ''' Test the cli error exits. '''
+    with pytest.raises(SystemExit) as err:
+        compiler_wrapper([])
+    assert ("psyclonefc error: PSYCLONE_COMPILER environment variable not "
+            "found!" in str(err.value))
```
arporter: Please also add a test for when the specified compiler command cannot be found.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -2990,8 +2991,10 @@ def test_resolve_imports(fortran_reader, tmpdir, monkeypatch):
     assert not isinstance(b_1, symbols.DataSymbol)
 
     # Now resolve all found containers (this will not fail for the
-    # unavailable c_mod)
-    subroutine.symbol_table.resolve_imports()
+    # unavailable c_mod, but it will be loged)
```
arporter: "logged"

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -3018,18 +3021,21 @@ def test_resolve_imports(fortran_reader, tmpdir, monkeypatch):
...
     # Monkeypatch the find_container_psyir() method of this ContainerSymbol
     # so that it returns None.
     monkeypatch.setattr(csym, "find_container_psyir", lambda local_node: None)
     table.add(csym)
-    # Resolving imports should run without problems.
-    table.resolve_imports()
+    # Resolving imports should run without problems, but log a Warning.
+    with caplog.at_level(logging.WARNING):
+        table.resolve_imports()
+    assert "Module 'a_mod' not found" in caplog.text
```
arporter, line 3038: Nice :-)

bin/psyclonefc
sergisiso: Done

src/psyclone/psyclonefc_cli.py
sergisiso: Done

src/psyclone/psyclonefc_cli.py
sergisiso: I improved the error, but for now I think checking that the given command is valid is not our responsibility, its the build system's. Also I don't want to be restrictive here, we should accept aliases, bash builtin commands like 'echo' which in my system is a binary but I am not sure that is always the case, compound commands `PSYCLONE_COMPILER="time $(which mpif90 -fc=$(which gfortran))`

src/psyclone/tests/psyclonefc_cli_test.py
sergisiso: Done

src/psyclone/tests/psyclonefc_cli_test.py
sergisiso: I argued against this in a comment above

src/psyclone/tests/psyir/symbols/symbol_table_test.py
sergisiso: Fixed

src/psyclone/psyclonefc_cli.py
sergisiso: Good catch, fixed

src/psyclone/psyclonefc_cli.py
sergisiso: Ups, of course

src/psyclone/psyclonefc_cli.py
sergisiso: I am reticent to this, it is true that I started with a small set of extensions and I am sure we will find a reason to increase them, but the log messages will be mostly false positives creating noise. Note that this checks all arguments, so:

`psyclonefc input.f90 dependency1.o dependency2.o -fopt-info-inline-optimized-missed=inline.txt -c -o output.o`

if fine, but it will create 5 false positive log messages.

examples/nemo/scripts/utils.py
sergisiso: Other than what the message above says (diverging results), I don't know much more, it doesn't happen with the metoffice patch branch that we use in the integration test. But I saw it somewhere, i can't remember if tag 5.0 or main, and since it doesn't slow down any of our tests I decided to put it here anyway.

src/psyclone/psyclonefc_cli.py
sergisiso: I am not sure about this. The "--log-level" and "--log-file" are not parsed/set-up yet here. I don't like too much the idea of pre-argparsing the PSYCLONE_OTPS nor the idea of ending in a separate log file. Also, the prints will be removed or hidden in the future, and I was thinking that if psyclonefc needs to have some flags it would be good to match [mpif90](https://www.mpich.org/static/docs/v3.1/www1/mpif90.html) -show/-echo (as HPC users may already familiar with these) and hey print, not log, the commands. But I am not ready to make a decision about this. So if you agree I added this in the checkboxes of remaining things to do for this issue.

sergisiso:
@arporter This is ready for review. I pushed back on some of the comments as at this stage I am unsure how it should work but

>  add a checklist item to the original issue that we need to write a section on this functionality in the user guide.

I agree with this, and I listed all the missing aspects

> Would it be possible to extend one or more of the examples to use this functionality so that it gets exercised "for real"?

I converted two steps of the integration test into using psyclonefc (should I convert also the other actions?)

examples/nemo/scripts/utils.py
arporter: OK.

src/psyclone/psyclonefc_cli.py
arporter: That's a good point, agreed.

src/psyclone/psyclonefc_cli.py
arporter: Oh yes, I'd forgotten that the logging flags wouldn't have been parsed. Thanks for adding it to the Issue.

src/psyclone/psyclonefc_cli.py
arporter: Oh I see. Yes, this wouldn't be very good.

PR #3029, opened 2025-06-11
(Closes #3026) bug with RegionTrans.get_node_list

### LINKED ISSUE #3026: Bug in RegionTrans.get_node_list() when passed `[Schedule]`
RegionTrans.get_node_list is supposed to ensure that we always have a valid list of Nodes to transform. However, @imaranresearch has discovered that it is possible to defeat it by passing it a list containing just a Schedule. In this case, we need to do the same as we do when passed a single Schedule node - i.e. return a new list containing its children.

arporter:


src/psyclone/tests/psyir/transformations/region_trans_test.py
```diff
@@ -38,7 +38,9 @@
 import pytest
 
 from psyclone.psyir.transformations import TransformationError
-from psyclone.psyir.nodes import CodeBlock, Node, Schedule
+from psyclone.psyir.nodes import (CodeBlock, IfBlock, IntrinsicCall,
```
LonelyCat124: Update author list for this file please.

LonelyCat124 (changes requested):
I'm happy with all the new functionality here. I have one tiny comment and then asking you to add the `apply` method to `RegionTrans` and use it to specify keyword arguments and use the new docstring inheritance system for this Transformation.

I've not built the docs yet as I suspect we'll see those change a reasonable amount.

src/psyclone/tests/psyir/transformations/region_trans_test.py
arporter: Oh yes, done.

LonelyCat124 (approved):
Happy with the options changes, the documentation builds and shows the inherited arguments and the implementation is good.

Ready to merge this now.

Issue #3031, opened 2025-06-13
[Bug] Adding profiling to a 'PURE' routine should remove the 'PURE' attribute


arporter:
When adding profiling to ORCA2_ICE_PISCES, there are numerous warnings about pure routines that now have calls to profiling calipers. The code still builds but we should fix this - adding profiling should remove the 'PURE' attribute on a routine.

arporter:
True. This is what I used to do (and maybe still do) in the kernels transformation script. I can extend the PR I've just opened (#3033) to refuse to do functions if you like?

sergisiso:
Yeah it would make your PR be about any function, I had this problem a couple of times in the past which I always had to go and modify that method.

arporter:
In testing #3032 I've found that gfortran refuses to compile an ELEMENTAL routine that has had profiling calipers added to it.
I think probably we have to refuse to add profiling calipers to such a routine as otherwise we might have to modify the call site (if we removed the elemental attribute)?

PR #3033, opened 2025-06-13
(Closes #3032) small update to improve NEMO profiling

### LINKED ISSUE #3032: Should not add profiling inside PSyclone-generated comparison functions
Currently the profiling functionality in nemo/scripts/utils.py puts calipers everywhere it can. Unfortunately this includes the Fortran functions that PSyclone generates to do comparisons of variables of unknown type. This results in a very large profiling overhead.

arporter:


arporter:
Having improved the script, I immediately ran into more issues relating to #3031 so I'm going to fix those in this PR too.

arporter:
Ready for a first look now. I've fired off the integration tests "just in case".

sergisiso (approved):
@arporter I agree with the changes: not annotating with profiling wrappers elemental functions and removing the pure attribute of annotated routines. Also not annotating any function for NEMO as this have a big overhead. The integration tests are green, this is approved for merging.

PR #3039, opened 2025-06-26
Generalised loop tiling with examples

mn416:
This PR generalises `LoopTiling2DTrans` to `LoopTilingTrans` by lifting restrictions on the number of dimensions of the tile and the sizes of those dimensions (they no longer have to be uniform).  For backwards compatibility, `LoopTiling2DTrans` remains but is now defined in terms of `LoopTilingTrans`.

This work was motivated by a simple example I was exploring for learning purposes: accelerating the matrix multiplication loop, which benefits from 3D tiling. This example is included in the PR, as well as a simpler example for matrix transposition, which only involves 2D tiling.


examples/psyir/matmul/Makefile
```diff
@@ -0,0 +1,58 @@
...
+	psyclone -s tile.py matmul.F90 -o matmul_tiled.F90
+
+matmul_omp.F90: matmul.F90 omp.py
+	psyclone -s omp.py matmul.F90 -o matmul_omp.F90
+
+matmul_omp_tiled.F90: matmul.F90 omp-tile.py
+	psyclone -s omp-tile.py matmul.F90 -o matmul_omp_tiled.F90
+
+%: %.F90
+	gfortran -O2 -fopenmp $< -o $@
```
hiker: Could you please rewrite the Makefile slighty to support using `$(F90)` and `($F90FLAGS)`? You can check other examples for details (e.g. examples/gocean/eg2, but these two variables are basically supported everywhere) Feel free to set `gfortran` etc as default if not specifed (`F90 ?= ...`).

examples/psyir/matmul/omp-tile.py
```diff
@@ -0,0 +1,46 @@
...
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+
+from psyclone.psyir.transformations import LoopTilingTrans
+from psyclone.psyir.transformations import OMPLoopTrans
+from psyclone.psyir.nodes import Loop, Routine
+
+
+def trans(psyir):
+    for routine in psyir.walk(Routine):
```
hiker, line 45: Can you please add a comment here of what this example script does?

examples/psyir/matmul/omp.py
```diff
@@ -0,0 +1,44 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+
+from psyclone.psyir.transformations import OMPLoopTrans
+from psyclone.psyir.nodes import Loop, Routine
+
+
+def trans(psyir):
+    for routine in psyir.walk(Routine):
```
hiker, line 43: Same, a comment what this does.

examples/psyir/matmul/tile.py
```diff
@@ -0,0 +1,44 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+
+from psyclone.psyir.transformations import LoopTilingTrans
+from psyclone.psyir.nodes import Loop, Routine
+
+
+def trans(psyir):
+    for routine in psyir.walk(Routine):
```
hiker, line 43: And another one :)

examples/psyir/transpose/Makefile
```diff
@@ -0,0 +1,58 @@
...
+	psyclone -s tile.py trans.F90 -o trans_tiled.F90
+
+trans_omp.F90: trans.F90 omp.py
+	psyclone -s omp.py trans.F90 -o trans_omp.F90
+
+trans_omp_tiled.F90: trans.F90 omp-tile.py
+	psyclone -s omp-tile.py trans.F90 -o trans_omp_tiled.F90
+
+%: %.F90
+	gfortran -O2 -fopenmp $< -o $@
```
hiker: Same comment as above, could you add support for `$F90` and `$F90FLAGS`

examples/psyir/transpose/omp-tile.py
```diff
@@ -0,0 +1,46 @@
...
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+
+from psyclone.psyir.transformations import LoopTiling2DTrans
+from psyclone.psyir.transformations import OMPLoopTrans
+from psyclone.psyir.nodes import Loop, Routine
+
+
+def trans(psyir):
+    for routine in psyir.walk(Routine):
```
hiker, line 44: A quick comment please.

examples/psyir/transpose/omp.py
```diff
@@ -0,0 +1,44 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2019-2025, Science and Technology Facilities Council
```
hiker: I just realise, that should only be 2025 (and would apply to the other examples as well, I won't comment that everywhere :) )

examples/psyir/transpose/omp.py
```diff
@@ -0,0 +1,44 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+
+from psyclone.psyir.transformations import OMPLoopTrans
+from psyclone.psyir.nodes import Loop, Routine
+
+
+def trans(psyir):
+    for routine in psyir.walk(Routine):
```
hiker, line 43: Comment.

examples/psyir/transpose/tile.py
```diff
@@ -0,0 +1,44 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+
+from psyclone.psyir.transformations import LoopTiling2DTrans
+from psyclone.psyir.nodes import Loop, Routine
+
+
+def trans(psyir):
+    for routine in psyir.walk(Routine):
```
hiker, line 43: Comment.

examples/psyir/transpose/trans.F90
```diff
@@ -0,0 +1,103 @@
...
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author: M. Naylor, University of Cambridge, UK
+
```
hiker, line 34: Please add a comment describing briefly what this program does.

examples/psyir/matmul/matmul.F90
```diff
@@ -0,0 +1,120 @@
...
+! COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+! -----------------------------------------------------------------------------
+! Author: M. Naylor, University of Cambridge, UK
```
hiker, line 34: A brief comment on what this example does.

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
```diff
@@ -32,29 +32,29 @@
...
+    Apply a 2D loop tiling transformation to a loop.  This is a special
+    case of LoopTilingTrans for 2D square tiles. For example:
 
     >>> from psyclone.psyir.frontend.fortran import FortranReader
     >>> from psyclone.psyir.nodes import Loop
     >>> from psyclone.psyir.transformations import LoopTiling2DTrans
     >>> psyir = FortranReader().psyir_from_source("""
     ... subroutine sub()
-    ...     integer :: ji, tmp(100)
+    ...     integer :: i, j, tmp(100)
```
hiker, line 58: Nice catch, thanks!

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
```diff
@@ -32,29 +32,29 @@
...
-from psyclone.psyir.nodes import Loop
-from psyclone.psyir.transformations.chunk_loop_trans import ChunkLoopTrans
-from psyclone.psyir.transformations.loop_swap_trans import LoopSwapTrans
+from psyclone.psyir.transformations.loop_tiling_trans import LoopTilingTrans
 from psyclone.psyir.transformations.loop_trans import LoopTrans
 from psyclone.psyir.transformations.transformation_error import \
     TransformationError
 
 
 class LoopTiling2DTrans(LoopTrans):
```
hiker, line 48: Wouldn't it be easier if `LoopTiling2DTrans` would be based on `LoopTilingTrans`? Esp. since `validate` calls a method from `LoopTilingTrans`.

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
```diff
@@ -135,19 +135,7 @@ def validate(self, node, options=None):
...
-        # guarantees that we have a 2 loop construct with only one loop
-        # statement inside the outer loop.
-        LoopSwapTrans().validate(node)
-
-        # Check that we can chunk both loops
-        outer_loop = node
-        inner_loop = node.loop_body.children[0]
-        ChunkLoopTrans().validate(outer_loop, options={'chunksize': tilesize})
-        ChunkLoopTrans().validate(inner_loop, options={'chunksize': tilesize})
+        LoopTilingTrans().validate(node, {"tiledims": [tilesize, tilesize]})
```
hiker: If the base class is changed, this would just become `super().validate(...)`, which is much easier to understand.

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
```diff
@@ -167,13 +155,4 @@ def apply(self, node, options=None):
...
-        position = node.position
-        outer_loop = node
-        inner_loop = node.loop_body.children[0]
-
-        ChunkLoopTrans().apply(outer_loop, options={'chunksize': tilesize})
-        ChunkLoopTrans().apply(inner_loop, options={'chunksize': tilesize})
-
-        loops = parent[position].walk(Loop)[1]
-        LoopSwapTrans().apply(loops)
+        LoopTilingTrans().apply(node, {"tiledims": [tilesize, tilesize]})
```
hiker: The same, with a different base class, you would be able to just use `super().apply(...)`.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2021-2025, Science and Technology Facilities Council.
```
hiker, line 4: 2025 as well here

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+    def __str__(self):
+        return "Tile the loop construct"
+
+    def _sink_validate(self, node, num_levels):
+        '''
+        Check that we can sink the outermost loop of a loop nest
+        downwards by the given number of levels.
+
+        :param node: the Loop that we want to sink.
+        :type node: :py:class:`psyclone.psyir.nodes.Loop`
```
hiker: We recently started to use typing, so the declaration should be changed to:
```
def _sink_validate(self, node: Loop, num_levels: int):
```
And then remove the `:type...` declarations (but leave the `:param`).

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+            loop by the requested number of levels.
+        '''
+
+        # Try to sink a loop by repeated swapping.
+        # Do this on a copy of the loop as we are only validating here.
+        loop = node.copy()
+        # Make sure the copy has a parent, as required by LoopSwapTrans.
+        Schedule().addchild(loop)
+        for i in range(0, num_levels):
+            swap = LoopSwapTrans()
```
hiker: The creation of `swap` can be moved outside of the loop.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+            swap.validate(loop)
```
hiker: You don't need to call validate explicitly if you need to apply anyway (unless you would want to catch the error and do something) - `apply` will first call `validate` anyway (so your code will validate twice)

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+                enddo
+              enddo
+            enddo
+        end subroutine sub
+
+    '''
+    def __str__(self):
+        return "Tile the loop construct"
+
+    def _sink_validate(self, node, num_levels):
```
hiker: The parameter `node` is a `Loop`, so I would suggest to rename `node` to be `loop`.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+        # Do this on a copy of the loop as we are only validating here.
+        loop = node.copy()
+        # Make sure the copy has a parent, as required by LoopSwapTrans.
+        Schedule().addchild(loop)
+        for i in range(0, num_levels):
+            swap = LoopSwapTrans()
+            swap.validate(loop)
+            swap.apply(loop)
+
+    def _sink_apply(self, node, num_levels):
```
hiker: Same as above about using Python typing.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+            loop by the requested number of levels.
+        '''
+
+        self._sink_validate(node, num_levels)
+
+        # Sink a loop by repeated swapping.
+        for i in range(0, num_levels):
+            LoopSwapTrans().apply(node)
+
+    def validate(self, node, options=None):
```
hiker: Please add Python typing.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+            assumed, i.e. a 2D 32x32 tile. \
+
+        :raises TransformationError: if an unsupported option has been \
+            provided.
+        :raises TransformationError: if the provided tilesize is not a \
+            integer.
+        '''
+        if options is None:
+            options = {}
+        super(LoopTilingTrans, self).validate(node, options=options)
```
hiker: You can use just `super().validate(...)` here.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+    def validate(self, node, options=None):
+        '''
+        Validates that the given Loop node can have a LoopTilingTrans
+        applied.
+
+        :param node: the loop to validate.
+        :type node: :py:class:`psyclone.psyir.nodes.Loop`
+        :param options: a dict with options for transformation.
+        :type options: Optional[Dict[str, Any]]
+        :param list[int] options["tiledims"]: The dimensions of the
```
hiker: Could you please check with https://psyclone.readthedocs.io/en/latest/developer_guide/transformations.html#moving-to-the-new-transformations-options
We are moving to a new way of handling optional parameters. I have to admit that I haven't kept up-to-date with the new way, but I believe you should be using the new way now. @sergisiso , can you quickly confirm this?

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+        if options is None:
+            options = {}
+        super(LoopTilingTrans, self).validate(node, options=options)
+
+        # Validate options map
+        # TODO #613: Hardcoding the valid_options does not allow for
+        # subclassing this transformation and adding new options, this
+        # should be fixed.
+        valid_options = ['tiledims']
+        for key, value in options.items():
```
hiker: Using the new `kwargs` way for optional arguments should also simplify this part.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+        for loop in node.walk(Loop)[0:numdims]:
+            for i in range(1, sink_upto):
+                self._sink_validate(node, i)
+            sink_upto = sink_upto - 1
+
+        # Check that we can chunk each loop
+        for (dim, loop) in zip(tiledims, node.walk(Loop)):
+            ChunkLoopTrans().validate(loop, options={'chunksize': dim})
+
+    def apply(self, node, options=None):
```
hiker: As above, use Python typing, and `**kwargs`.

src/psyclone/tests/psyir/transformations/loop_tiling_trans_test.py
```diff
@@ -0,0 +1,201 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2021-2025, Science and Technology Facilities Council.
```
hiker, line 4: 2025

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+        :raises TransformationError: if it is not possible to sink the \
+            loop by the requested number of levels.
+        '''
+
+        # Try to sink a loop by repeated swapping.
+        # Do this on a copy of the loop as we are only validating here.
+        loop = node.copy()
+        # Make sure the copy has a parent, as required by LoopSwapTrans.
+        Schedule().addchild(loop)
+        for i in range(0, num_levels):
```
hiker: pycode style reports that `i` is not used. Just use `_` if the variable is not required.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+
+        :raises TransformationError: if it is not possible to sink the \
+            loop by the requested number of levels.
+        '''
+
+        self._sink_validate(node, num_levels)
+
+        # Sink a loop by repeated swapping.
+        for i in range(0, num_levels):
+            LoopSwapTrans().apply(node)
```
hiker: Again `i` not used, use `_` instead.

src/psyclone/psyir/transformations/loop_tiling_trans.py
```diff
@@ -0,0 +1,231 @@
...
+                    f"transformation option '{key}', the supported options "
+                    f"are: {valid_options}.")
+
+        tiledims = options.get("tiledims", [32, 32])
+        numdims = len(tiledims)
+
+        # Even though the loops that ultimately will be sunk are the ones
+        # resulting from ChunkLoopTrans, sinking these has the same
+        # validation constraints as sinking the original loops. The
+        # following validations also guarantee that we have a nested loop
```
hiker, line 157: That is very hard to see that this tests for 'perfect' nesting (each loop except innermost only contains one loop). I assume that's because of LoopSwapTrans used in `_sink_validate`? If that is correct, could you add a comment to explain this either here, or document this in `_sink_validate`.

src/psyclone/tests/psyir/transformations/loop_tiling_trans_test.py
```diff
@@ -0,0 +1,201 @@
...
+        do j = j_out_var, MIN(j_out_var + (16 - 1), 100), 1
+          do k = k_out_var, MIN(k_out_var + (32 - 1), 100), 1
+            tmp(i,j) = 2 * tmp(i,j) + k
+          enddo
+        enddo
+      enddo
+    enddo
+  enddo
+enddo'''
+    assert expected in result
```
hiker, line 205: Could you add two more tests:
1. Have three nested loop, and tile the outer two only. Just to make sure all loops indices are indeed derived from `tiledims` specified (and not from the number of nested loops or so)
2. While you have a test that triggers when you have too many `tiledims`, but add a test where one loop is not perfectly nested, e.g.:
```
do i
   do j1
      do k
      ...
      enddo
   enddo
   do j2
      ...
   enddo
enddo
```

src/psyclone/psyir/transformations/loop_tiling_trans.py
sergisiso: Yes please, if you could move to using keyword arguments it would be great. But if there is some complication we can skip it for this PR.

hiker:
> Thank you very much @hiker for the review. I am AFK for the rest of this week, but will respond and make the fixes next week, all being well.

No rush, I have a few days off myself as well :)

examples/psyir/matmul/Makefile
mn416: Done

examples/psyir/matmul/omp-tile.py
mn416: Done

examples/psyir/matmul/omp.py
mn416: Done

examples/psyir/matmul/tile.py
mn416: Done

examples/psyir/transpose/Makefile
mn416: Done

examples/psyir/transpose/omp-tile.py
mn416: Done

examples/psyir/transpose/omp.py
mn416: Done (except in some cases where the code in the file originated from an existing file, in which case I've kept the dates from the original).

examples/psyir/transpose/omp.py
mn416: Done

examples/psyir/transpose/tile.py
mn416: Done

examples/psyir/transpose/trans.F90
mn416: Done

examples/psyir/matmul/matmul.F90
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
mn416: Originally, I defined it as a subclass but later hit the problem described in the comment:

```
# TODO #613: Hardcoding the valid_options does not allow for
# subclassing this transformation and adding new options, this
# should be fixed.
```

Now, LoopTilingTrans uses the new way of handling transformation options, which is different to the backwards-compatible LoopTiling2DTrans. So I think it's better to not use subclassing here.

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
mn416: (See previous response)

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
mn416: (See previous response)

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: I've kept 2021-2025 here as some code in this file originated from `loop_tiling_2d_trans.py`.

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: In theory, `LoopSwapTrans` could maintain state between calls to `validate` or `apply`, so I prefer to use a new instance for each independent transformation.

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Good point, I've removed the extra call to `validate`

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done.

I've done a similar renaming in other methods too.

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Thanks @hiker for pointing me to these docs. I've moved to the new way of handling options. With the removal of `loop_tiling_2d_trans.py`, I also decided to make the `tiledims` argument mandatory (it seems a bit arbitrary to assume the user wants a 2D tiling of a specific size).

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/tests/psyir/transformations/loop_tiling_trans_test.py
mn416: I've kept 2021-2025 here as some code in this file originated from loop_tiling_2d_trans_test.py.

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done

src/psyclone/psyir/transformations/loop_tiling_trans.py
mn416: Done (extra comment added to `_sink_validate` description)

src/psyclone/tests/psyir/transformations/loop_tiling_trans_test.py
mn416: I've done both of these:

(1) is checked by `test_loop_tiling_trans_apply2`.

(2) is checked by `test_loop_tiling_trans_validation4`.

sergisiso:
We just said in the previous comment that we would delete it :) I am fairly sure nobody uses it, and the update is very simple. I think we are just adding unnecessary complexity to keeping it around.

src/psyclone/tests/psyir/transformations/loop_tiling_trans_test.py
hiker: Thanks!

src/psyclone/psyir/transformations/loop_tiling_2d_trans.py
hiker: I still don't like it (mainly a change in LoopTilingTrans could break LoopTiling2DTrans ... I think, e.g. if the constructor would add an attribute, which is used in validate. Since LoopTiling2DTrans never calls the constructor, that attribute would be undefined).
BUT: there is the issue that I had this misunderstanding with Andy/Sergi about removing or deprecating `LoopTiling2DTrans`. So, given that this PR is good now, I'll proceed with merging it, and open a ticket to then remove LoopTiling2DTrans, which removes unfortunate code :)

hiker:
I'm bringing this up-to-master, and will then trigger the CI.

src/psyclone/psyir/transformations/loop_tiling_trans.py
hiker: That would actually be the reason to move it outside imho - maybe it starts caching something :) But overall it's not important.

hiker (approved):
CI passed, all issues addressed. Proceeding to merge.

Issue #1235, opened 2021-05-05
caplog does not capture output in github actions


rupertford:
pytest uses the caplog fixture to capture any logging output. This works fine when running it on my laptop but gives not output at all when run using github actions.

sergisiso:
I closed this without knowing there are TODOs to fix mentioning this issue, reopening.

PR #3040, opened 2025-06-26
(Closes #3038) catch parse error in frontend and test

### LINKED ISSUE #3038: Preprocessor directives causing syntax error
When trying to parse [`ukca_mode_tracer_maps_mod`](https://code.metoffice.gov.uk/trac/ukca/browser/main/trunk/src/science/core/aerosols/glomap/ukca_mode_tracer_maps_mod.F90), I get the following error.

```
  File "/home/jelsey/NG-ARCH/transformer/psyclone_ve/bin/psyclone", line 42, in <module>
    main(sys.argv[1:])
  File "/home/jelsey/NG-ARCH/cemac_ngarch_tools/psyclone/src/psyclone/generator.py", line 592, in main
    code_transformation_mode(input_file=args.filename,
  File "/home/jelsey/NG-ARCH/cemac_ngarch_tools/psyclone/src/psyclone/generator.py", line 772, in code_transformation_mode
    .psyir_from_file(input_file)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jelsey/NG-ARCH/cemac_ngarch_tools/psyclone/src/psyclone/psyir/frontend/fortran.py", line 255, in psyir_from_file
    parse_tree = self._parser(reader)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/home/jelsey/NG-ARCH/transformer/psyclone_ve/lib/python3.12/site-packages/fparser/two/Fortran2003.py", line 275, in __new__
    raise FortranSyntaxError(string, "")
fparser.two.utils.FortranSyntaxError: at line 175
>>>END DO\n
```

The code that it is tripping up on looks like:

```
! Construct the number and component mass index arrays
! by looping over all entries in the all_tracers array
DO i=1,n_tracers
! (Slow) Workaround for NAG Fortran vn7.0 internal compiler error.
! Fixed in vn7.1
#if defined(NAG_FORTRAN) && (NAG_FORTRAN == 7000000)
tracer_belongs = .FALSE.
DO j=1,nmodes
IF ( mode_names(j) == all_tracers_names(i)(1:7) ) THEN
tracer_belongs = .TRUE.
END IF
END DO   
IF ( tracer_belongs ) THEN
#else
IF ( ANY(mode_names(:) == all_tracers_names(i)(1:7)) ) THEN
#endif

<code> 

END DO    ! this is the line it detects the syntax error on
```
The code compiles fine, but PSyclone passthrough doesn't work.
I've set up a [branch](https://code.metoffice.gov.uk/trac/ukca/browser/main/trunk/src/science/core/aerosols/glomap/ukca_mode_tracer_maps_mod.F90) that refactors this bit of code (which now works fine), but I just wanted to flag it as something that PSyclone didn't know how to handle.

arporter:
A small PR to improve the Fortran frontend - we were falling over with a nasty stack trace when given some invalid Fortran.


arporter:
Ready for a look from either @LonelyCat124 or @sergisiso.

sergisiso (changes requested):
@arporter I agree in catching and reraising the fparser errors, as the fparser encapsulation should not escape the frontend files, but this change alone doesn't solve the error reported in the issue because the psyclone command uses the psyir_from_file, not the psyir_from_source method:
```
psyclone ukca_mode_tracer_maps_mod.F90
  File "/home/centos/workspace/psyclone/.venv/lib/python3.13/site-packages/fparser/two/Fortran2003.py", line 271, in __new__
    return Base.__new__(cls, string, _deepcopy=_deepcopy)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/centos/workspace/psyclone/.venv/lib/python3.13/site-packages/fparser/two/utils.py", line 511, in __new__
    raise NoMatchError(errmsg)
fparser.two.utils.NoMatchError: at line 177
>>>END DO


During handling of the above exception, another exception occurred:

  File "/home/centos/workspace/psyclone/.venv/bin/psyclone", line 42, in <module>
    main(sys.argv[1:])
    ~~~~^^^^^^^^^^^^^^
  File "/home/centos/workspace/psyclone/PSyclone/src/psyclone/generator.py", line 592, in main
    code_transformation_mode(input_file=args.filename,
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
                             recipe_file=args.script,
                             ^^^^^^^^^^^^^^^^^^^^^^^^
                             output_file=args.o,
                             ^^^^^^^^^^^^^^^^^^^
                             line_length=args.limit)
                             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/centos/workspace/psyclone/PSyclone/src/psyclone/generator.py", line 772, in code_transformation_mode
    .psyir_from_file(input_file)
     ~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/frontend/fortran.py", line 263, in psyir_from_file
    parse_tree = self._parser(reader)
  File "/home/centos/workspace/psyclone/.venv/lib/python3.13/site-packages/fparser/two/Fortran2003.py", line 275, in __new__
    raise FortranSyntaxError(string, "")
fparser.two.utils.FortranSyntaxError: at line 177
>>>END DO
```

Also, I know some year ago I pushed for the removal of capturing errors in generate.py because then they were reported without the backtrace context. But I think that now that the front-end is clearly separated, I would accept surrounding the "psyir_from_file"s with a try, and just print the error without the backtrace and sys.exit(-1).

arporter:
The test failure was the link-checker on the GOcean web page.

arporter:
Ready for another look/question to answer.

sergisiso:
> Oops! Embarrassing. Fixed now.
> 
> > Also, I know some year ago I pushed for the removal of capturing errors in generate.py because then they were reported without the backtrace context. But I think that now that the front-end is clearly separated, I would accept surrounding the "psyir_from_file"s with a try, and just print the error without the backtrace in stderr and sys.exit(-1).
> 
> Do you want me to do this in this PR? i.e. find all the `psyir_from_file()` calls in `generate.py` and put them inside try..excepts?

Yes please. But instead of discarding the backtrace as I said above, we can put the full error in the logfile. And the just the error message in stderr.

arporter:
Found it. psyclone.parser.kernel.get_kernel_parse_tree() attempts to disable logging in fparser:
https://github.com/stfc/PSyclone/blob/dc66e55c00740c39580c76b5b353ab9f0825ae4c/src/psyclone/parse/kernel.py#L161-L164
but this seems to disable it everywhere.

arporter:
I've been a bit naughty but, since I've now got to grips with logging and testing, I've fixed a few of the xfails in the adjoint tests. There are still some that refer to #1235 but I thought I'd leave it as is for now.

arporter:
Having removed the line that effectively disabled logging, I do see a lot more log messages when running the tests (since pytest reports these by default). Hopefully there aren't any other unexpected changes but I'll fire off the integration tests to check.

src/psyclone/generator.py
```diff
@@ -542,7 +555,8 @@ def main(arguments):
     else:
         logging.basicConfig(level=loglevel)
     logger = logging.getLogger(__name__)
-    logger.debug("Logging system initialised.")
+    logger.debug("Logging system initialised. Default level is %s.",
+                 args.log_level)
```
sergisiso: Do you mean: "Logging level is %s" (it is not always the default, sometimes it is the given by the argument)

src/psyclone/generator.py
```diff
@@ -799,10 +815,18 @@ def code_transformation_mode(input_file, recipe_file, output_file,
...
+                               ignore_directives=not keep_directives)
+        try:
+            psyir = reader.psyir_from_file(input_file)
+        except (InternalError, ValueError) as err:
+            print(f"Failed to create PSyIR from file '{input_file}'",
+                  file=sys.stderr)
+            logger.error(
+                "Failed to create PSyIR from file '%s'. Error was:\n%s",
+                input_file, str(err))
+            sys.exit(1)
```
sergisiso: This is not what I understood we are going for, in my opinion this misses too much information. It does not show the error message in non-logging mode, nor the backtrace in the logging mode, e.g.:
```
$ echo "syntax error" > a.f90
$ psyclone a.f90
Failed to create PSyIR from file 'a.f90'
```
I think it should be: ` print(f"Failed to create PSyIR from file '{input_file}' due to:\n{str(err)}")` so the output is 

```
$ psyclone a.f90
Failed to create PSyIR from file 'a.f90' due to:
Failed to parse source in file 'a.f90'.
Error was: at line 1
>>>syntax error

Is the input valid Fortran (note that CPP directives must be handled by a pre-processor)?
```

If we do that, IOError can also be added, because it is not important for the user where in the psyclone code this happens:
```
$ psyclone b.f90
Failed to create PSyIR from file 'b.f90' due to:
[Errno 2] No such file or directory: 'b.f90'
```

However, I think the log should provide all information, including the traceback. In fact logging has a method for this usecase:
```
-            logger.error(
-                "Failed to create PSyIR from file '%s'. Error was:\n%s",
-                input_file, str(err))
+            logger.exception("Following exception raised:")
```

Same in the other places.

What do you think?

sergisiso (changes requested):
@arporter Thanks for figuring out the pytest log issues. This is getting close but the output is not what I expected, see discussion inline.

src/psyclone/generator.py
arporter: Oops. Yes, I did mean that. Fixed.

src/psyclone/generator.py
arporter: Ah, I misunderstood, sorry. If I do:
```python
logger.error(err, stack_info=True, exc_info=True)
```
and extend it to handle IOError too, then, for `psyclone --log-level ERROR --log-file ./log.txt ./missing` we do get all the gory detail:
```
ERROR:psyclone.generator:[Errno 2] No such file or directory: './missing'
  File "/home/me/Projects/PSyclone/src/psyclone/generator.py", line 821, in code_transformation_mode
    psyir = reader.psyir_from_file(input_file)
  File "/home/me/Projects/PSyclone/src/psyclone/psyir/frontend/fortran.py", line 261, in psyir_from_file
    reader = FortranFileReader(file_path,
                               include_dirs=Config.get().include_paths,
                               ignore_comments=self._ignore_comments)
  File "/home/me/Projects/PSyclone/external/fparser/src/fparser/common/readfortran.py", line 1670, in __init__
    self.file = open(
                ~~~~^
        file_candidate, "r", encoding="UTF-8", errors="fparser-logging"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
FileNotFoundError: [Errno 2] No such file or directory: './missing'
Stack (most recent call last):
  File "/home/me/.pyenv/versions/psyclone13/bin/psyclone", line 42, in <module>
    main(sys.argv[1:])
  File "/home/me/Projects/PSyclone/src/psyclone/generator.py", line 628, in main
    code_transformation_mode(input_file=args.filename,
  File "/home/me/Projects/PSyclone/src/psyclone/generator.py", line 825, in code_transformation_mode
    logger.error(err, stack_info=True, exc_info=True)
```
Is that what you had in mind?

sergisiso: yes

sergisiso: Well, just `logger.error(err, exc_info=True)`, the stack info of the log is not necessary, just the error

arporter:
Hopefully ready for another look now.

sergisiso (approved):
Thanks @arporter , I like how this is handled now. This is approved for merging after #2967

Issue #2935, opened 2025-03-25
[PSyAd] [LFRic] Errors in use of LBOUND and looping in generated adjoint kernel


tom-j-h:
Found when using PSyAD to generate adjoint kernel for `science/gungho/source/kernel/solver/apply_mixed_operator_kernel_mod.F90` for LFRic Apps ticket 669.

The relevant bits of the required patch file for the generated adjoint kernel are:
```patch
-        u_e(idx_12 - iw3 + LBOUND(u_e(:,:), 1),df) = u_e(idx_12 - iw3 + LBOUND(u_e(:,:), 1),df) + q32(idx_12 + ij - iw3,1,df) * &
+        u_e(idx_12 - iw3 + LBOUND(u_e, dim=1),df) = u_e(idx_12 - iw3 + LBOUND(u_e, dim=1),df) + q32(idx_12 + ij - iw3,1,df) * &

-          u_e(idx_10 - iw2v + LBOUND(u_e(:,:), 1),df2) = u_e(idx_10 - iw2v + LBOUND(u_e(:,:), 1),df2) + mu_cd(idx_10 + ij - &
+          u_e(idx_10 - iw2v + LBOUND(u_e, dim=1),df2) = u_e(idx_10 - iw2v + LBOUND(u_e, dim=1),df2) + mu_cd(idx_10 + ij - &

-    do idx_7 = UBOUND(t_col, dim=1), LBOUND(t_col, dim=1), -1
-      t_col(:) = t_col(:) + mt_lumped_inv(idx_7 + iwt - LBOUND(t_col(idx_7), 1)) * t_col(idx_7)
-      t_col(idx_7) = 0.0
-    enddo
+    t_col(:) = t_col(:) * mt_lumped_inv(iwt:iwt+1+nm1)

-        u_e(idx_6 + LBOUND(u_e(:,:), 1) - 1,df) = u_e(idx_6 + LBOUND(u_e(:,:), 1) - 1,df) + (-pt2(idx_6 + ij - 1,2,df) * &
+        u_e(idx_6 + LBOUND(u_e, dim=1) - 1,df) = u_e(idx_6 + LBOUND(u_e, dim=1) - 1,df) + (-pt2(idx_6 + ij - 1,2,df) * &

-          u_e(idx_3 - iw2h + LBOUND(u_e(:,:), 1),df2) = u_e(idx_3 - iw2h + LBOUND(u_e(:,:), 1),df2) + mu_cd(idx_3 + ij - &
[...]

PR #3041, opened 2025-06-26
(closes #2823) Parse unresolved symbols with parenthesis as calls

### LINKED ISSUE #2823: Function Calls vs ArrayAccessors vs Type Constructors
When we don't have the type of a symbol, it is ambiguous what the expression below is:
```
subroutine test()
   use other_module
   a = myimportedysymbol(1)
end subrotuine
```

Currently we often default to ArrayAccessor, but this is unsafe for GPU offloading as array accessors are pure/available on the device, but this is not the case for the other 2 cases where this needs to be checked. For this reason, we have a NEMO_FUNCTIONS in utils.py which is often outdated or broken for untested branches.

In such cases we should make "myimportedysymbol" a CodeBlock, or at least a Call. This will have a significant impact on the GPU offloading performance of NEMO but I think it is necessary for getting people going with correct results and we can get it back the performance by more precise use of the "resolve_imports" functionality.

There is a similar issue in https://github.com/stfc/fparser/issues/384 , but it can be resolved there or in the Fparser2Reader.

sergisiso:


src/psyclone/tests/psyir/backend/sympy_writer_test.py
```diff
@@ -193,7 +193,7 @@ def test_sym_writer_functions(fortran_reader, expressions):
                           ("f(:)", {'f': Function('f')}),
                           ("a%b", {'a_b': Symbol('a_b')}),
                           ("a%b(1)", {'a_b': Function('a_b')}),
-                          ("c + c(1)", {'c': Function('c')}),
+                          # ("c + c(1)", {'c': Function('c')}),
```
sergisiso: This is ready for review, maybe one for @arporter 

You will notice that I commented out a couple of SympyWriter test cases that combine the same symbol used differently. I do not understand the SympyWriter enough to fix it, when I tried it I end up with huge modifications to that class (maybe @arporter or @hiker could suggest if I am missing a simpler fix). But then the change is quite inconsequential because when this is used, e.g. for symbolic comparison, if the sympy fails, it evaluates to False anyway, so I don't think is worth adding a lot of extra complexity for this. So I would propose deleting or postponing solving this case.

examples/nemo/scripts/utils.py
```diff
@@ -486,6 +401,9 @@ def insert_explicit_loop_parallelism(
                 " 'nlay_i' or 'nlay_s'.")
             continue
 
+        if routine_name == "tra_zdf_imp":
+            opts['allow_strings'] = True
```
arporter: Comment please.

src/psyclone/core/variables_access_map.py
```diff
@@ -86,7 +86,7 @@ def __str__(self):
                     mode = "WRITE"
                 else:
                     # The data associated with this signature is not accessed.
-                    mode = "NO_DATA_ACCESS"
+                    mode = "UNKNOWN"
```
arporter: The doc string was already out of date but please could you now make sure it mentions "UNKNOWN".

src/psyclone/domain/common/transformations/raise_psyir_2_alg_trans.py
```diff
@@ -247,10 +246,11 @@ def apply(self, call, index, options=None):
...
                 continue
-            elif isinstance(call_arg, ArrayReference):
-                # kernel misrepresented as ArrayReference
-                args = call_arg.pop_all_children()
-                type_symbol = call_arg.symbol
+            elif isinstance(call_arg, Call):
+                # Get the symbols and args to reconstruct it as a
+                # higer-abstraction AlgorithmInvokeCall node
+                type_symbol = call_arg.routine.symbol
+                args = call_arg.pop_all_children()[1:]
```
arporter, line 257: Better to use `call_arg.arguments` here?

src/psyclone/domain/lfric/transformations/raise_psyir_2_lfric_alg_trans.py
```diff
@@ -74,16 +74,15 @@ def apply(self, call, index, options=None):
 
             if call.argument_names[idx]:
                 call_name = f"{call_arg.value}"
-            elif isinstance(call_arg, ArrayReference):
-                # kernel or builtin misrepresented as ArrayReference
-                args = call_arg.pop_all_children()
+            elif isinstance(call_arg, Call):
+                symbol = call_arg.routine.symbol
+                args = call_arg.pop_all_children()[1:]
```
arporter, line 79: `call_arg.arguments` again?

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -426,7 +426,10 @@ def _create_type_map(self, list_of_expressions: Iterable[Node],
                 else:
                     orig_sym = None
 
-            is_fn_call = isinstance(orig_sym, RoutineSymbol)
+            is_fn_call = (
+                isinstance(orig_sym, RoutineSymbol) or
+                any(x.access_type in [AccessType.CALL, AccessType.UNKNOWN]
```
arporter, line 406: Comment please.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,44 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
         name, interface=UnresolvedInterface(), **kargs)
 
 
+def _refine_symbols_with_usage_location(
+    location: Node,
+    execution_part: Fortran2003.Execution_Part
+):
+    ''' Refine the symbol infomration that we obtained from parsing
+    the declarations sections by knowledge infered by looking at the
```
arporter: "inferred"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,44 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
         name, interface=UnresolvedInterface(), **kargs)
 
 
+def _refine_symbols_with_usage_location(
+    location: Node,
+    execution_part: Fortran2003.Execution_Part
+):
+    ''' Refine the symbol infomration that we obtained from parsing
```
arporter: "information"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,44 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+    the declarations sections by knowledge infered by looking at the
+    usage location of the symbols in the execution_part
+
+    :param symtab: SymbolTable to enhance information for.
+    :param execution_part: fparser nodes to analyse for symbol usage.
+
+    '''
+    # The top-reference of the assingment lhs is guaranteed to be a DataSymbol
```
arporter: "assignment"

arporter: Probably should also mention that we don't support statement functions as this statement isn't true for them.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,44 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+    # a(i,:,k) -> 'a' must be a DataSymbol, for now of UnresolvedType because
+    # we don't have enough information to know the shape of the ArrayType
+    for part_ref in walk(execution_part, Fortran2003.Part_Ref):
+        for child in part_ref.items:
+            if isinstance(child, Fortran2003.Section_Subscript_List):
+                if not any(isinstance(subchild, Fortran2003.Subscript_Triplet)
+                           for subchild in child.items):
+                    continue
+                if isinstance(part_ref.parent, Fortran2003.Data_Ref):
+                    continue
```
arporter: Comment please.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5033,9 +5071,16 @@ def _parenthesis_handler(self, node, parent):
 
     def _part_ref_handler(self, node, parent):
         '''
-        Transforms an fparser2 Part_Ref to the PSyIR representation. If the
-        node is connected to a SymbolTable, it checks the reference has been
-        previously declared.
+        Transforms an fparser2 Part_Ref to the PSyIR representation.
+
+        fparser2 cannot always disambiguate between Array Accessors, Calls and
+        DerivedType constuctors, and it fallbacks to Part_Ref when unknown.
```
arporter: "constructors" and "falls back"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5033,9 +5071,16 @@ def _parenthesis_handler(self, node, parent):
...
+        PSyclone has a better chance of properly categorising them because we
+        can follow 'use' statements to retrieve symbol information. If
+        psyclone does not find the definition it fallbacks to a Call. The
```
arporter: "falls back" again

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5033,9 +5071,16 @@ def _parenthesis_handler(self, node, parent):
...
+        reason for this is that it is the more safest option. Constructors and
```
arporter: s/more//

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5033,9 +5071,16 @@ def _parenthesis_handler(self, node, parent):
...
+        accessors can be cosidered calls (of unknown purity and elemental attr)
```
arporter: "considered"

src/psyclone/psyir/nodes/datanode.py
```diff
@@ -55,6 +56,12 @@ def datatype(self):
             better then it must override this method.
         :rtype: :py:class:`psyclone.psyir.symbols.UnresolvedType`
         '''
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.loop import Loop
+        from psyclone.psyir.nodes.ranges import Range
+        # If it is a direct child of Loop or Range, it can only be an Integer
+        if self.parent and isinstance(self.parent, (Loop, Range)):
+            return INTEGER_TYPE
```
arporter, line 64: While this is true, we still don't know what precision the integer is. We could potentially (in a separate PR) introduce the concept of unknown precision?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -918,6 +918,21 @@ def reference_accesses(self) -> VariablesAccessMap:
...
+        '''
+        :returns: the datatype of the result of this IntrinsicCall.
+        :rtype: :py:class:`psyclone.psyir.symbols.DataType`
+
+        '''
+        if self.intrinsic in (
+            IntrinsicCall.Intrinsic.LBOUND,
+            IntrinsicCall.Intrinsic.UBOUND,
+        ):
+            if "dim" in self.argument_names:
```
arporter, line 3538: Worth a comment to say these intrinsics return a 1D array unless `idim` is specified. Again, we should have a TODO here for the precision of the integer type.

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -161,18 +162,23 @@ def validate(self, node, options=None):
             attempts to enclose the assingment setting the return value.
         '''
         device_string = options.get("device_string", "") if options else ""
+        strings = options.get("allow_strings", False) if options else False
```
arporter, line 163: Please document these options in the doc string. Although we have introduced the "you touch it, you fix it" rule for options on transformations, this PR is already quite big so I suggest we we leave that for now.

arporter: Also, in ACCKernelsTrans, this option is called "allow_string". We should be consistent (I don't mind which).

src/psyclone/psyir/transformations/omp_target_trans.py
```diff
@@ -184,6 +190,28 @@ def validate(self, node, options=None):
...
+                        if dtype.intrinsic == ScalarType.Intrinsic.CHARACTER:
+                            message = (
+                                f"OpenMP Target cannot enclose a region that "
+                                f"uses characters, but found: "
+                                f"{datanode.debug_string()}"
+                            )
+                            if verbose:
+                                node.preceding_comment = message
+                            raise TransformationError(message)
+                    # TODO #3054: Deal with UnresolvedType
```
arporter, line 207: Picky but, to reduce the amount of commented-out code we have, how about:
```python
# TODO #3054 - deal with isinstance(dtype, UnresolvedType).
```
and remove the rest.

src/psyclone/tests/core/variables_access_map_test.py
```diff
@@ -361,10 +361,10 @@ def test_variables_access_map_domain_loop():
     vam = invoke.schedule.reference_accesses()
     assert str(vam) == (
         "a: READ, b: READ, f1_data: READWRITE, f2_data: "
-        "READWRITE, field_type: NO_DATA_ACCESS, i_def: NO_DATA_ACCESS, "
+        "READWRITE, field_type: UNKNOWN, i_def: UNKNOWN, "
         "map_w3: READ, ncell_2d_no_halos: "
         "READ, ndf_w3: READ, nlayers_f1: READ, nlayers_f2: READ, "
-        "r_def: NO_DATA_ACCESS, undf_w3: READ")
+        "r_def: UNKNOWN, undf_w3: READ")
```
arporter: I'm unsure about this - we should know that r_def is a kind parameter so its access is not UNKNOWN?

src/psyclone/tests/core/variables_access_map_test.py
```diff
@@ -384,12 +384,12 @@ def test_lfric_access_map():
...
-        "READ+WRITE, f2_data: READ, field_type: NO_DATA_ACCESS, i_def: "
-        "NO_DATA_ACCESS, m1_data: READ, "
+        "READ+WRITE, f2_data: READ, field_type: UNKNOWN, i_def: "
+        "UNKNOWN, m1_data: READ, "
         "m2_data: READ, map_w1: READ, map_w2: READ, map_w3: READ, ndf_w1: "
         "READ, ndf_w2: READ, ndf_w3: READ, nlayers_f1: READ, np_xy_qr: READ, "
-        "np_z_qr: READ, quadrature_xyoz_type: NO_DATA_ACCESS, "
-        "r_def: NO_DATA_ACCESS, undf_w1: READ, undf_w2: READ, "
+        "np_z_qr: READ, quadrature_xyoz_type: UNKNOWN, "
+        "r_def: UNKNOWN, undf_w1: READ, undf_w2: READ, "
```
arporter: Same here.

src/psyclone/tests/dependency_test.py
```diff
@@ -85,8 +85,9 @@ def test_assignment(fortran_reader):
     array_assignment = schedule.children[1]
     assert isinstance(array_assignment, Assignment)
     var_accesses = array_assignment.reference_accesses()
-    assert (str(var_accesses) == "c: WRITE, d: READ, e: READ, f: READ, "
-                                 "i: READ, j: READ, x: READ, y: READ")
+    assert (str(var_accesses) == "c: WRITE, d: READ, e: READ, f: UNKNOWN, "
+                                 "i: READ, j: READ, x: READWRITE, "
```
arporter: Has this changed because `x` and `y` are now potentially arguments to a call to `f`?

src/psyclone/tests/dependency_test.py
```diff
@@ -263,10 +264,10 @@ def test_lfric():
     var_accesses = schedule.reference_accesses()
     assert str(var_accesses) == (
         "a: READ, cell: READ+WRITE, f1_data: READ+WRITE, f2_data: READ, "
-        "field_type: NO_DATA_ACCESS, i_def: NO_DATA_ACCESS, m1_data: READ, "
+        "field_type: UNKNOWN, i_def: UNKNOWN, m1_data: READ, "
```
arporter: Presumably we also know that `field_type` is a DataTypeSymbol?

src/psyclone/tests/domain/common/transformations/raise_psyir_2_alg_trans_test.py
```diff
@@ -342,18 +342,18 @@ def test_arg_error(fortran_reader, arg):
         invoke_trans.validate(psyir.children[0][0])
     if arg == "alg(field)":
```
arporter, line 343: This bit needs to be removed because `arg` is never set to "alg(field)" now.

src/psyclone/tests/generator_test.py
```diff
@@ -1160,13 +1160,8 @@ def test_generate_trans_error(tmpdir, capsys, monkeypatch):
     _, output = capsys.readouterr()
     # The output is split as the location of the algorithm file varies
```
arporter: You can remove this comment now.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
arporter: I spent ages getting this working :-) I could take a look.

src/psyclone/tests/psyir/nodes/datanode_test.py
```diff
@@ -40,19 +40,34 @@
...
+                    DataSymbol("a", INTEGER_SINGLE_TYPE),
+                    Call.create(RoutineSymbol("get_start")),
+                    Call.create(RoutineSymbol("get_stop")),
+                    Call.create(RoutineSymbol("get_step")),
+                    []
+    )
+
+    assert loop.start_expr.datatype == INTEGER_TYPE
+    assert loop.stop_expr.datatype == INTEGER_TYPE
+    assert loop.step_expr.datatype == INTEGER_TYPE
```
arporter, line 70: Please include a check when the access is an index into an array too.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -451,10 +451,10 @@ def test_reference_accesses_bounds(operator, fortran_reader):
     psyir = fortran_reader.psyir_from_source(code)
     schedule = psyir.walk(Assignment)[0]
 
-    # The access to 'a' should be reported as 'NO_DATA_ACCESS' as its
+    # The access to 'a' should be reported as 'UNKNOWN' as its
```
arporter: As mentioned before, I think/feel that 'UNKNOWN' is different to 'NO_DATA_ACCESS'.

src/psyclone/psyir/nodes/call.py
```diff
@@ -319,7 +319,11 @@ def reference_accesses(self) -> VariablesAccessMap:
 
         # The RoutineSymbol has a CALL access.
         sig, indices_list = self.routine.get_signature_and_indices()
-        var_accesses.add_access(sig, AccessType.CALL, self.routine)
+        # pylint: disable=unidiomatic-typecheck
+        if type(self.symbol) is Symbol:
+            var_accesses.add_access(sig, AccessType.UNKNOWN, self.routine)
```
arporter, line 326: Comment please - e.g. If the type of the Symbol is unknown then this may be a Call or an array access so the access is UNKNOWN.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
arporter: The problem seems to be that we no longer identify an UNKNOWN access as being an array access and therefore we don't specialise the associated Symbol. This then means that the Sympy generation fails because the type of the Symbol (just unknown) is not consistent with the fact that we've stored a function (that expects a certain number of arguments) in the type-map.

Obviously, we no longer want to specialise an unknown Symbol as being of ArrayType because the point of this PR is that such a Symbol is now considered a Call. However, unlike the ArrayType, a RoutineSymbol has no information on the number of arguments and therefore we can't use it to transfer information across (from the SympyWriter to the SympyReader). I'm a bit hazy on this so will tag @hiker in case he has any bright ideas.

arporter (changes requested):
Thanks Sergi. It's a big improvement that's been a long time coming (probably since we first started working with NEMO). You'll see I have some concerns about describing certain accesses as UNKNOWN when I would argue that they're not.

I've had a dig into the Sympy issue and can at least understand what's going wrong. I'm not sure how to fix it so hope inspiration will strike over the weekend.

src/psyclone/tests/core/variables_access_map_test.py
sergisiso: Just to clarify this because there are multiple comments asking about this, the AccessType of r_def or the field_type that you asked in another comment has not changed. What changed is the `__str__` method of the map that contains them, which was:
```
  if self.is_read(signature):
      if self.is_written(signature):
          mode = "READ+WRITE"
      else:
          mode = "READ"
  elif self.is_written(signature):
      mode = "WRITE"
  else:
      # The data associated with this signature is not accessed.
      mode = "NO_DATA_ACCESS"
```
And I changed the last line from `mode = "NO_DATA_ACCESS"` to `mode = "UNKNOWN"` because it is not true that all the `else` represent no data access.

In retrospective I think we should just get rid of this logic and let it print the value of the `AccessType` enumerator directly, it is confusing that the string representation changes the underlying names. The only question is that there are multiple accesses per symbol in a region, and what we do if they have a different ACCESS_TYPE, for example:
`a = 1_x + x`

Maybe the map should just print the +-separated set of access modes?
"a: WRITE, x: TYPE_INFO+READ"

What do you think @arporter?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
sergisiso: > therefore we don't specialise the associated Symbol

This writer sets the "_DISABLE_LOWERING = True", which means that it operates on the main tree and not a copy. So changes persist after the visitors. It is unexpected that an equality operation ends up modifying the tree. If there is some guaranteed ACCESS_TYPE change it should be done in the frontend.

sergisiso: > However, unlike the ArrayType, a RoutineSymbol has no information on the number of arguments

Btw, this PR only affects ArrayReferences that had generic Symbols or UnresolvedTypes, if it had an ArrayType and knew the shape, it is still an ArrayReference with the same type.

src/psyclone/tests/core/variables_access_map_test.py
sergisiso: I decided to attempt this in a separate (earlier) PR #3068, so then the behaviour change of the Unresolved->Calls is more clear by the diff on this PR.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
hiker: I am not entirely certain what the expected behaviour should be. My understanding is that an unresolved expressions `c(i)` should be considered a call (and not an array access). From that point of view, `c + c(i)` is actually an invalid expression.

Or is the expectation that the writer should then try to 'reconsider' `c` to be an array, and converting it this way? To some degree, on a sympy level (where arrays get converted to function), this kind of shouldn't matter, but of course Fortran can have named parameter (`c(i=i)`), which I would guess will not work.

I agree that it just shouldn't crash (which should be easy todo by just catching the exception).

We can discuss tonight.

hiker: OK, I found the root cause: the variable access information for `c(1)` returns an 'unknown' access, call.py line 323:
```
        if type(self.symbol) is Symbol:
            var_accesses.add_access(sig, AccessType.UNKNOWN, self.routine)
```
The result of this is that variable access info then does not detect the accesses to `c+c(1)` to include an array index (`1`), and as such the symbol for `c` is not specialised to become an array, and then doesn't get the `c(sympy!no_bounds, sympy_nobounds,1)` replacement done, which then causes a crash.

sergisiso:
@arporter Thanks for the first review but I decided to delay this a little bit. I realise that this will impact a couple of NG-Arch partners because they don't have the correct import statements and they probably won't find out until the release which will be very close to the end of the project. So to avoid last minute disruption I plan to retake this PR after the next psyclone release.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
sergisiso: This was fixed by the previous PR that changed the how access_sequences are stringified, but I missed this comment, this is not a more accurate "INQUIRY" access.

examples/nemo/scripts/utils.py
sergisiso: Removed, this is now better solved by the loop reordering enabled by RankIncrease and the string falls outside the loop.

src/psyclone/core/variables_access_map.py
sergisiso: Code removed in a previous PR.

src/psyclone/domain/common/transformations/raise_psyir_2_alg_trans.py
sergisiso: I still need the pop_all_children as the arguments returns a slice which is not where I want to remove the items from.
But if you find it more readable I can convert it to two different statements, the .argument and then the .pop_all_children

src/psyclone/domain/lfric/transformations/raise_psyir_2_lfric_alg_trans.py
sergisiso: Same

src/psyclone/psyir/backend/sympy_writer.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Fixed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Fixed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Fixed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/nodes/datanode.py
sergisiso: I am unsure if it is unknown or they are implicitly casted to the default size. But as you say I would leave this for a separate discussion/PR.

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: Done

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: Done in a previous PR

src/psyclone/tests/core/variables_access_map_test.py
sergisiso: As mention I did this is a separate PR to show that it was a printing artifact, and not a real change.

src/psyclone/tests/core/variables_access_map_test.py
sergisiso: Same response

src/psyclone/tests/dependency_test.py
sergisiso: yes, we don't know what f is, so we need to assume the worst access conditions, otherwise the multiple dependency analysis that we have could miss a real dependency (we can resolve the module to get more accurate accesses)

src/psyclone/tests/dependency_test.py
sergisiso: This was fixed in a previous PR

src/psyclone/psyir/nodes/call.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/tests/generator_test.py
sergisiso: Done

src/psyclone/tests/psyir/backend/sympy_writer_test.py
sergisiso: I have, for now, moved this to a TODO and removed the specialise_symbol method inside the sympy_writer (partially because that visitor does not operate on a copy, and I think it is unexpected that running a backend modifies the original tree). I think we can recover the functionality by letting the new frontend refine_symbols_by_usage_location do the same specialisation if we can guarantee that this pattern 'c + c(1)' always imply that c is an array an never a function.

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -77,6 +77,7 @@
 FILES_TO_SKIP = [
     "vremap.f90",  # TODO #2772
     "icefrm.f90",  # Has unsupportet implicit symbol declaration
+    "fldread.f90",
```
arporter: Please add a comment giving the reason we skip it.

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -109,12 +110,10 @@
     "trczdf.f90",
     "trcice_pisces.f90",
     "dtatsd.f90",
-    # Runtime Error: Illegal address during kernel execution with
```
arporter, line 111: Ah, in fact this comment is the one should be at L80 perhaps?

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -203,8 +202,7 @@ def trans(psyir):
         if (
             subroutine.name.lower().startswith("sign_")
             or subroutine.name.lower() == "solfrac"
-            # Important for performance but causes SIGNAL 11 in some cases
-            # or (psyir.name == "sbc_phy.f90" and not subroutine.walk(Loop))
+            or (psyir.name == "sbc_phy.f90" and not subroutine.walk(Loop))
```
arporter, line 212: Has the signal 11 gone away with a more recent version of the compiler?

examples/nemo/scripts/utils.py
```diff
@@ -182,6 +93,7 @@
 
 PRIVATISATION_ISSUES = [
     "ldftra.f90",  # Wrong runtime results
+    "zdftke.f90",
```
arporter: Comment please.

src/psyclone/psyclonefc_cli.py
```diff
@@ -123,7 +124,13 @@ def compiler_wrapper(arguments):
             stem = Path(argument).stem
             suffix = Path(argument).suffix
             output = f"{stem}.psycloned{suffix}"
-            psyclone_args = psyclone_options + ['-o', output, argument]
+            # Always add an include to the current directory, because even if
+            # it is the default, psyclone removes it when adding another -I.
+            # Also, having the absolute path instead of '.' is convinient to
```
arporter: "convenient"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,48 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+    # References that have Subscript in direct children are DataSymbol, e.g.:
+    # a(i,:,k) -> 'a' must be a DataSymbol, for now of UnresolvedType because
+    # we don't have enough information to know the shape of the ArrayType
+    for part_ref in walk(execution_part, Fortran2003.Part_Ref):
+        for child in part_ref.items:
+            if isinstance(child, Fortran2003.Section_Subscript_List):
+                if not any(isinstance(subchild, Fortran2003.Subscript_Triplet)
+                           for subchild in child.items):
+                    continue
+                # The same considereation applies if this is an derived type
```
arporter: "consideration" and s/an/a/

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,48 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+                symbol_type=DataSymbol,
+                datatype=UnresolvedType())
+    # References that have Subscript in direct children are DataSymbol, e.g.:
+    # a(i,:,k) -> 'a' must be a DataSymbol, for now of UnresolvedType because
+    # we don't have enough information to know the shape of the ArrayType
+    for part_ref in walk(execution_part, Fortran2003.Part_Ref):
+        for child in part_ref.items:
+            if isinstance(child, Fortran2003.Section_Subscript_List):
+                if not any(isinstance(subchild, Fortran2003.Subscript_Triplet)
+                           for subchild in child.items):
```
arporter: I'm finding this quite hard to follow. So far I have:
 * `child` is a Part_Ref and may be an array access or a call.
 * It can't be a call if any of its index expressions is a slice...

I can't remember enough about the fparser2 parse tree. Please could you add comments saying what Section_Subscript_List and Subscript_Triplet are. Perhaps with example Fortran fragments of the things we are excluding from being calls.

src/psyclone/psyir/frontend/fparser2.py
arporter: Thanks. Again, I can't remember what `Data_Ref` corresponds to so please elaborate on why we skip if the parent is one.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -643,14 +641,8 @@ def _get_effective_shape(self):
...
                         f"indirect access) cannot be determined.")
-            elif isinstance(idx_expr, (Call, CodeBlock)):
-                # We can't yet straightforwardly query the type of a function
-                # call - TODO #1799.
-                raise NotImplementedError(
-                    f"The array index expressions for access "
-                    f"'{self.debug_string()}' include a function call or "
-                    f"unsupported feature. Querying the return type of "
-                    f"such things is yet to be implemented.")
+            # The validate only allows [Range | DataNode] children, so there is
```
arporter: While this is true, having code that will silently ignore having the wrong type of `idx_expr` is a bit dangerous. Could we raise an InternalError here? It should be relatively straightforward to test as this is a standalone routine.

src/psyclone/psyir/nodes/call.py
```diff
@@ -358,6 +365,16 @@ def reference_accesses(self) -> VariablesAccessMap:
                 var_accesses.update(arg.reference_accesses())
         return var_accesses
 
+    @property
+    def symbol(self):
+        '''
+        :returns: the routine symbol that this call calls.
+        :rtype: Optional[py:class:`psyclone.psyir.symbol.Symbol`]
```
arporter: Type hint instead of rtype?

src/psyclone/psyir/nodes/call.py
```diff
@@ -358,6 +365,16 @@ def reference_accesses(self) -> VariablesAccessMap:
...
+        '''
+        if self.routine and self.routine.symbol:
```
arporter, line 380: Is it possible for us to have a Call without an associated routine? Pls add a comment.

src/psyclone/psyir/nodes/call.py
```diff
@@ -387,8 +404,8 @@ def is_elemental(self):
             information is not known then it returns None.
         :rtype: NoneType | bool
         '''
-        if self.routine and self.routine.symbol:
-            return self.routine.symbol.is_elemental
+        if (self.symbol and isinstance(self.symbol, RoutineSymbol)):
```
arporter: Same here although I see we use `self.symbol` here. Could we use that at L374 too?

src/psyclone/tests/core/symbolic_maths_test.py
```diff
@@ -474,13 +474,15 @@ def test_symbolic_math_use_range(fortran_reader, expressions):
...
-    ("a(i)*((b(i,j)+c(j))/d)",
-     "a(i) * b(i,j) / d + a(i) * c(j) / d"),
+    ("a(i)*((b(i, j)+c(j))/d)",
+     "a(i) * b(i, j) / d + a(i) * c(j) / d"),
     # 'a' is unresolved so we don't know from the first occurrence whether or
     # not it is a scalar.
-    ("a / a(i)", "a / a(i)"),
-    ("norm_u(idx+iw2) * u_e(idx + (LBOUND(u_e,dim=1)-iw2v), df2)",
-     "norm_u(idx + iw2) * u_e(idx - iw2v + LBOUND(u_e, 1),df2)")])
+    # TODO #3175: Re-enable sympywriter support for mix array usage expressions
```
arporter: I'm worried that this means we'll have broken PSyAD as I added these tests in specifically for #2935. I see that I've neglected to add the kernel that exhibited the problem (in #2935) to our testing, mea culpa. I can possibly help to solve #3175.

src/psyclone/tests/dependency_test.py
```diff
@@ -85,9 +85,12 @@ def test_assignment(fortran_reader):
     array_assignment = schedule.children[1]
     assert isinstance(array_assignment, Assignment)
     var_accesses = array_assignment.reference_accesses()
+    # We don't know if 'f' is a function or an array (CALLED or READ), so
+    # it is catergorised as UNKNOWN. It's arguments take the worst case
```
arporter: "Its"

src/psyclone/tests/domain/common/transformations/raise_psyir_2_alg_trans_test.py
arporter: This is still here. Although I now wonder why "alg(field)" has been removed as one of the test options? Is it because it's now impossible to have it as both the subroutine name and an array reference?

src/psyclone/tests/psyir/backend/sympy_writer_test.py
arporter: As discussed, I think we *can* guarantee that `c + c(1)` does always imply that `c` is an array. The F2008 standard has:
image
so the parentheses are optional on a call to a subroutine that has no arguments but they are not optional for a call to a function.
Could we extend the new frontend functionality to apply this logic?

Either way, I tried PSyAD with this branch with the kernel named in #2935 and it seemed to be fine (although that kernel may have changed a lot in the interim as I get nowhere near as many PSyclone-generated loop variables).

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -1771,6 +1771,55 @@ def test_process_use_stmts_resolving_external_imports(
     assert isinstance(stmt_rhs.children[1], Call)
 
 
+def test_process_resolving_modules_give_correct_types(
+        parser, tmpdir, monkeypatch):
+    ''' Test that if the Fparser2Reader if provided with a list of
```
arporter: "is" and s/this are/this is/

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -1771,6 +1771,55 @@ def test_process_use_stmts_resolving_external_imports(
...
+            integer :: pure_func
+            integer, intent(in) :: i
+            pure_func = 3
+        end function
+    end module
+    ''')
+    # Add the path to the include_path and set up a frontend instance
+    # witth the module_to_resolve names
+    monkeypatch.setattr(Config.get(), '_include_paths', [tmpdir])
+    processor = Fparser2Reader(resolve_modules=["other"])
```
arporter: For completeness, it might be worth parameterising this test so that it either does or does not supply `resolve_modules`. We can then check that, for the case we're not resolving modules, all of the RHS are listed as Calls.

arporter (changes requested):
Looks good Sergi, thanks very much. I got a bit worried about the fact that we might be breaking PSyAD but, when I tried, it seemed fine. I agree that the frontend is the right place to do the reasoning about the type of unresolved symbols and I like that change. As discussed, I think we can safely extend it to say that a symbol that is accessed with and without parentheses must be an array.

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: Done

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: It was not true for that file, which is an identified bug, but I think it is true for the files below it. So I restored the comment.

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: We have not updated the compiler in quite some time. The signal 11 is quite elusive and doesn't necessarely happen in the file that is excluded, sometimes excluding one file makes another not have a singnal 11. We added more exclusions in the top list, which makes this line ok. This is better as this one is important for performance.

examples/nemo/scripts/utils.py
sergisiso: Done

src/psyclone/psyclonefc_cli.py
sergisiso: Fixed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/tests/dependency_test.py
sergisiso: Fixed

src/psyclone/tests/psyir/frontend/fparser2_test.py
sergisiso: Fixed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: I have restructured this loop and comments to make it more readable, let me know what you think.

src/psyclone/psyir/nodes/array_mixin.py
sergisiso: I don't think I agree with this. It won't be silently ignored because such a Node cannot be created. Note that this is not what I deleted. There was no else before. What I changed is that Call and CodeBlock should still go thought he previous condition as they may have a idx_expr.datatype. And if they don't, it will end up in the bottom NotImplementedError, the code behaviour is equivalent of what it was before.

src/psyclone/psyir/nodes/call.py
sergisiso: Done

src/psyclone/psyir/nodes/call.py
sergisiso: Just in the case of incomplete nodes without the mandatory children. I added a comment.

src/psyclone/psyir/nodes/call.py
sergisiso: Done (the lines have moved since the review, I assumed you meant the is_pure method)

src/psyclone/psyir/transformations/omp_target_trans.py
sergisiso: I was going to say that it was done on a previous PR, but actually your last comment wasn't, so I modified it here.

src/psyclone/tests/domain/common/transformations/raise_psyir_2_alg_trans_test.py
sergisiso: Yes, this is needed for coverage, the test that any kernel_functor name must not be the same as the algorithm subroutine name is still there.

src/psyclone/tests/psyir/frontend/fparser2_test.py
sergisiso: Done, but I considered it was better as an extension to the test with a comment than a parameterise

src/psyclone/tests/psyir/backend/sympy_writer_test.py
sergisiso: I implemented the refinement and now "c + c(1)" is a "Reference + ArrayReference". However, the sympy evaluation still didn't like it be because it doesn't matter if it is an ArrayReference or a Call, both end up being SympyFunctions.

The real issue/fix was in sympy_writer.py::reference_node.

But since I already implemented the previously discussed Node type refinement, I also left it in this PR.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -805,19 +778,29 @@ def reference_node(self, node: Reference) -> str:
...
+                # This is represented by a sympy.Function, since this is just a
+                # Reference, it must be an array expression without
+                # parentheses. For consistency, we still treat it as a Sympy
+                # function call and therefore add the triple array indices to
+                # represent `lower:upper:1` for each dimension:
+                if node.symbol.is_array:
+                    shape = node.symbol.shape
+                else:
+                    # If we don't know the dimension we make it look like a
+                    # function call without argumetns, this will make it not
```
arporter: "arguments"

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -805,19 +778,29 @@ def reference_node(self, node: Reference) -> str:
...
-        # Now this must be an array expression without parentheses. For
-        # consistency, we still treat it as a Sympy function call and therefore
-        # add the triple array indices to represent `lower:upper:1` for each
-        # dimension:
-        shape = node.symbol.shape
-        result = [f"{self.no_bounds},{self.no_bounds},1"]*len(shape)
-
-        return (f"{name}{self.array_parenthesis[0]}"
-                f"{','.join(result)}{self.array_parenthesis[1]}")
+        if name in self.type_map:
```
arporter, line 804: Good solution! Thanks.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,64 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+                # we know its a DataSymbol, for now of UnresolvedType, as we
+                # don't know enough to infer the whole shape.
+                if any(isinstance(subchild, Fortran2003.Subscript_Triplet)
+                       for subchild in child.items):
+                    _find_or_create_unresolved_symbol(
+                        location, name,
+                        symbol_type=DataSymbol,
+                        datatype=UnresolvedType())
+        if name in direct_refnames_in_exprs:
+            # If this any other expression has the same reference name without
```
arporter: s/this//

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -359,6 +359,64 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+            # parenthesis, e.g.: a + a(3), we know a is an array and not a
+            # function call, as the later have mandatory parenthesis.
```
arporter: "latter"

src/psyclone/tests/dependency_test.py
```diff
@@ -85,9 +85,12 @@ def test_assignment(fortran_reader):
     array_assignment = schedule.children[1]
     assert isinstance(array_assignment, Assignment)
     var_accesses = array_assignment.reference_accesses()
+    # We don't know if 'f' is a function or an array (CALLED or READ), so
```
arporter, line 88: Good comment, thanks.

src/psyclone/tests/psyir/frontend/fparser2_part_ref_test.py
```diff
@@ -99,6 +94,48 @@ def test_handling_part_ref_expression():
     assert len(new_node.children) == 3  # Array dimensions
 
 
+def test_handling_part_ref_without_symbol_type(fortran_reader):
+    '''Test that fparser2 Part_Ref is converted to a Call if the reference
+    symbol if of unknown kind. Unless it is found on a specific location that
```
arporter: "is" and best to remove the full stop before "Unless"

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -1771,6 +1771,67 @@ def test_process_use_stmts_resolving_external_imports(
...
+            a = unsupported_array(3)
+            a = pure_func(3)
+        end subroutine
+    end module
+    ''')
+    parse_tree = parser(reader)
+    module = parse_tree.children[0]
+
+    # By default this will all be parsed as Calls with unknown
+    # is_lememental/is_pre attributes
```
arporter: "is_elemental/is_pure"

src/psyclone/tests/psyir/nodes/datanode_test.py
arporter: This one is still outstanding. Possibly your new 'refine_symbols_with_usage` method in the frontend could be extended for this. Similarly for precision symbols (e.g 1.0_wp means `wp` is an integer).

src/psyclone/psyir/frontend/fparser2.py
arporter: It's a lot better now - thanks :-)

src/psyclone/psyir/nodes/array_mixin.py
arporter: Oh yes, I can see that it was this way before. However, I think it's useful to defend against these cases as it saves time later - if we come to modify e.g. `_validate_child` in future and forget to modify this routine at the same time it could cause some pain.

src/psyclone/tests/psyir/frontend/fparser2_test.py
arporter: Great, thanks.

src/psyclone/tests/psyir/backend/sympy_writer_test.py
arporter: Well done!

src/psyclone/psyir/backend/sympy_writer.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Removed

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Fixed

src/psyclone/tests/psyir/frontend/fparser2_part_ref_test.py
sergisiso: Done both

src/psyclone/tests/psyir/frontend/fparser2_test.py
sergisiso: Done

src/psyclone/tests/psyir/nodes/datanode_test.py
sergisiso: We cannot do it for indices, since this can be scalar of arrays of integers. But we can do it for children of Ranges.

I added asserts for both

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -76,6 +75,7 @@
 # List of all files that psyclone will skip processing
 FILES_TO_SKIP = [
     "icefrm.f90",  # Has an unsupported implicit symbol declaration
+    "icerst.f90",
```
arporter: Can you give a reason for this in a comment?

src/psyclone/psyclonefc_cli.py
```diff
@@ -88,6 +89,11 @@ def compiler_wrapper(arguments):
             'psyclonefc error: PSYCLONE_COMPILER environment variable not '
             'found! This environment variable must be set to the Fortran '
             'compiler to use.')
+    if fortran_compiler.endswith("psyclonefc"):
+        sys.exit(
```
arporter, line 93: Great :-)

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -360,6 +360,64 @@ def _find_or_create_unresolved_symbol(location, name, scope_limit=None,
...
+):
+    ''' Refine the symbol information that we obtained from parsing
+    the declarations sections by knowledge inferred by looking at the
+    usage location of the symbols in the execution_part
+
+    :param location: scope to enhance information for.
+    :param execution_part: fparser nodes to analyse for symbol usage.
+
+    '''
+    # The top-reference of the assignment lhs is guaranteed to be a DataSymbol
```
arporter: s/the assignment/an assignment/

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -2387,6 +2387,25 @@ def reference_accesses(self) -> VariablesAccessMap:
             var_accesses.update(child.reference_accesses())
         return var_accesses
 
+    @property
```
arporter, line 3527: This will presumably clash with #3150 which is almost ready to go on.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -279,9 +279,10 @@ def validate(self, node, options=None, **kwargs):
                     if not call.is_pure]
         if not_pure:
             message = (
-                f"Loop cannot be parallelised because it cannot "
-                f"guarantee that the following calls are pure: "
-                f"{sorted(set(not_pure))}")
+                f"Loop cannot be parallelised because psyclone cannot "
+                f"guarantee that the accesses to {sorted(set(not_pure))} are "
+                f"arrays or pure calls. If they are but the symbol is "
+                f"imported, try adding the module in RESOLVE_IMPORTS.")
```
arporter: "...module name to RESOLVE_..."

src/psyclone/tests/psyir/frontend/fparser2_part_ref_test.py
```diff
@@ -99,6 +94,48 @@ def test_handling_part_ref_expression():
...
+        """
+        module test_mod
+           use other
+        contains
+          subroutine test
+             ! On the lhs of the assignment is an array
+             array1(i + unknown(i)) = 3 + 2 + unknown(i)
+
+             ! If it has any triplet construct is an array
+             tmp =  array2(i,i:j,j) + unknown(array3(:))
```
arporter, line 114: We have to be careful to allow for character substrings:
image
In this case, `array3` could be a character sub-string.

arporter: i.e. although fparser has created a triplet construct for `array3(:)`, it doesn't actually know that it is!

arporter (changes requested):
Thanks Sergi, it's great to see the simplifications to the NEMO scripts.
I was all set to say "just minor things" but then realised that character substrings are still likely to be a problem - see inline. This is partly because fparser always seems to create a `Subscript_Triplet` whenever it sees a colon.

src/psyclone/tests/psyir/frontend/fparser2_part_ref_test.py
sergisiso: Good point, seeing your comment my first thought was that maybe this was ok to still represent them as ArrayReferences since they suport indexing/slicing and we know they work (this is also what they are now in the PSyIR).

But when I check the snippet below, all indexing/slicing works (first 5 statements) but the 3 first lbounds fail, while the 4th succeeds (is a character array). 

```
program string_test
    ! Character substrings
    character(8) :: a
    character(len=8) :: b
    character*8 :: c
    ! Character array
    character, dimension(8) :: d

    a = "a"
    b(:) = "b"
    c(2:3) = "c"
    d(:) = "d"
    write(*,*) a(2:4), b(2:4), c(2:4), d(2:4)
    write(*,*) lbound(a), lbound(b), lbound(c)
end program
```

So that means:

1) clearly we cannot infer that a "<unresolved_symbol>(Triplet)" is an ArrayType, because some of our array operations/transfomrations create lbound/ubound/size, ....

2) Maybe that's why we needed to make the "allow_strings=False"? In theory strings operations are supported on the GPU, but maybe we added some unsupported intrinsics arround some of them? Tightening (1) may allow as to enable allow_strings and bring more things on the gpu (needs testing).

3) I am not sure about the node type. We could convert those to be CodeBlocks, this may have false positives, specially in LFRic_atm where we don't follow imports yet. Or we could let them be ArrayReferences(as now), making sure transformations that modify them, e.g. array expansion, do check the datatype and not the nodetype. Agai I need some testing with the integration tests to have a better opinion.

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/transformations/parallel_loop_trans.py
sergisiso: Done

src/psyclone/tests/psyir/frontend/fparser2_part_ref_test.py
sergisiso: And after all this text, it turns out we already do the right thing for (1) which is the one as I worried about. We only say they are datasymbols, but not the type:
```
DataSymbol:
  array1: DataSymbol<UnresolvedType, Unresolved>
  array2: DataSymbol<UnresolvedType, Unresolved>
  array3: DataSymbol<UnresolvedType, Unresolved>
  array4: DataSymbol<UnresolvedType, Unresolved>
```

There is still the question of (2) and (3) but these is not impacted by this PR, so I suggest we look at them in an independent issue/PR.

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: There was no clash here, that PR only adds names for now, not datatypes.

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: The CI succeeded without it, so I removed it.

src/psyclone/tests/psyir/frontend/fparser2_part_ref_test.py
arporter: OK. So I think you're saying that this PR doesn't alter the fact that we get an ArrayReference representing an access of a character sub-string. The important thing (for this PR) is that the symbol itself is left as UnresolvedType rather than ArrayType?

PR #3043, opened 2025-07-02
Bring back python 3.9 in the integration tests

sergisiso:
python3.9 is not EOLd yet and Archer2 which has LFRic users defaults to it. So it would be good to test against it.


sergisiso:
@LonelyCat124 This is ready for review. I am bringing back python 3.9 as EPCC and other people from NG-Arch are using it on Archer2. I am tempted to suggest that we update the previous changelog instead of creating a new issue for this as it was only a couple of PRs ago and the issue didn't specify 3.9.

LonelyCat124:
Hi @sergisiso - I'll review this when I'm back from leave next week - can you set the integration running in the mean time?

schreiberx:
@sergisiso  Thanks. I had some issues with Python 3.9 with the most recent Psyclone version.

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -590,11 +590,13 @@ def test_omploop_trans_new_options(sample_psyir):
 
     with pytest.raises(TypeError) as excinfo:
         omplooptrans.apply(tree.walk(Loop)[0], collapse="x")
-    assert ("'OMPLoopTrans' received options with the wrong types:\n"
-            "'collapse' option expects type 'int | bool' but "
-            "received 'x' of type 'str'.\n"
-            "Please see the documentation and check the provided types."
-            in str(excinfo.value))
+    # The exact error message depends on the python version, so we only
```
arporter: Please could you add a new issue (to remove 3.9 when EOL) and add a TODO here.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -43,7 +43,7 @@
 from collections import OrderedDict
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any
+from typing import Any, Optional, Union
```
arporter, line 46: Please add a TODO to an Issue to remove 3.9.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -44,7 +44,7 @@
 from dataclasses import dataclass, field
 import os
 import sys
-from typing import Iterable
+from typing import Iterable, Optional
```
arporter, line 47: Please add a TODO to an Issue to remove 3.9

arporter (changes requested):
All looks fine @sergisiso, thanks for fixing this (and apologies for being too hasty in removing it). Just a request to make a new issue to drop 3.9 when it goes EOL (in October 2025) so that we can put back the commented-out test.

sergisiso:
@arporter I don't think I agree with the comments, we don't need a TODO for things that will continue to be valid, there are many Optional, Unions through the codebase and I don't think it is a goal to remove those. The commented out assert may be a better location but even then, I don't think we should deprecate 3.9 anytime soon (October), since it is the default on Archer2 and probably means it is the default in similar RedHat/Cray machines.

arporter:
> The commented out assert may be a better location but even then, I don't think we should deprecate 3.9 anytime soon (October), since it is the default on Archer2 and probably means it is the default in similar RedHat/Cray machines.

Good point - we'll probably need 3.9 for longer as you say. In that case, please could you change the test back to something that works for all versions so that we don't have commented-out code.

sergisiso:
@arporter Ready for another review

arporter (approved):
Thanks Sergi. This can be merged now.

PR #3055, opened 2025-07-14
Duplicated async statements in OpenACC

schreiberx:
The async ACC clause results in duplicated async statements, which are fixed with this PR.

```
!$acc kernels default(present) async(1) async(1)
!$acc end kernels

.!$acc parallel default(present) async(1) async(1)
!$acc end parallel

.!$acc update if_present host(x) async(1) async(1)

.!$acc enter data copyin(x) async(1)
```


sergisiso:
@schreiberx Good spot, but we can actually go further and completely remove the `_build_async_string`. Since the async is in a PSyIR Clause it should be the PSyIR backend that produces the right code when visiting the node and not the the begin_string method. The method was probably a leftover of gen_code.

I took the liberty to do the change in this branch, but this means that the review will need to be finished by @arporter or @LonelyCat124

arporter (commented):
Thanks very much Martin and Sergi.
This looks good.
I've set the integration tests running. If they are green then this is good to go.

arporter (approved):
Will proceed to merge.

PR #3057, opened 2025-07-15
(closes #3056) Demote OpenMP lowering error to warning

### LINKED ISSUE #3056: Confusing (and sometimes incorrect) OMPParallelDirective lowering error message
We need to address the
```
Note that some nodes need to be lowered from an ancestor in order to properly apply their in-tree modifications. Original error was 'Generation Error: Lowering 'OMPParallelDirective' does not support symbols that need synchronisation unless they are in a depend clause, but found: 'x' which is not in a depend clause.'.
```

This is confusing and not possible to disable with the `--backend disable-validation`

In a discussion @arporter @LonelyCat124 suggested to keep it but making it resilient to "force" and "ignore_dependencies_for" transformations.

My opinion is still that the lowering should not attempt to do this validation as there is not way to distinguish false negative, and the validation should have already been done by the transformation. To add one more argument, with @LonelyCat124 new parsing of existing directives, this condition means that valid input Fortran code may not go trough passthrough when triggering this condition.

But as last time, I can compromise if at least the error is improved and is resilient to force/ignore. So, what do you want to do @arporter @LonelyCat124 ?

sergisiso:


LonelyCat124 (commented):
Happy with this, will sort out changelog and merge.

Issue #3061, opened 2025-07-17
DefUseChain does not consider explicitly_private_symbols


sergisiso:
```
subroutine my_sub()
  integer :: ji
  integer :: jj
  integer :: i
  real, dimension(10,10) :: var1
  real, dimension(10) :: ztmp
  real, dimension(10) :: ztmp2

  do i = 1, 2, 1
    var1 = 1.0
    ztmp2 = 1.0
    if (i == 1) then
      do ji = 1, 10, 1
        do jj = 1, 10, 1
          ztmp(jj) = 3
        enddo
        do jj = 1, 10, 1
          var1(ji,jj) = ztmp(jj) * 2
        enddo
      enddo
    else
      do ji = 1, 10, 1
        do jj = 1, 10, 1
          var1(ji,jj) = ztmp(jj) * 2
        enddo
      enddo
    end if
  enddo

end subroutine my_sub
```

@LonelyCat124 if I select the first ztmp read in the first branch and ask for ref.next_accesses() it does not show the ztmp from the second branch, but it should as the control flow may loop back and take the other branch. Do you know what could it be?

sergisiso:
I have impremented it on the callee as a workaround on my branch, but I think it should be the DUC, it is a property of PSyIR loop, so PSyIR tools should follow it. It seems it should just be a check somewhere before we append more nodes to basic_blocks/control_flow_nodes? But maybe these are in multiple places in DUC?

LonelyCat124:
@sergisiso I might fix this next, but I have one slight concern. If its a property of a PSyIR loop, is there enforcement of the symbols privatisation if nothing else is done to the loop? I know it would again be handling poor user input to have to worry about that, but its my one concern that a variable could be marked private for a loop, but if nothing ensures that it is before codegen then the output from DUC would be wrong. Maybe we just don't worry though.

sergisiso:
>  any of the loops above is given a loop.explicitly_private_sybmol(ztmp_symbol) (which is a loop attribute) the DefUseChain still looks at ancestors. It shouldn't, it should just loop back to the same loop.

I believe I wrote this the other way around, it should still look upwards (it could be a firstprivate), but it shouldn't look at the same loop as the variable is private to the iteration.

I am not sure what to do with explicit_private without OpenMP/ACC.. I would say still follow what the loop properties say

LonelyCat124:
Yeah that might be best for now.

PR #3062, opened 2025-07-17
(closes #2671) Improve array privatisation

### LINKED ISSUE #2671: Support Parallel Array Privatisation
Currently we always consider arrays shared on parallel loops, and never privatise them. For example in: 

```
real(kind=wp), dimension(nlay_i) :: ztmp
do ji = 1, npti, 1
  ...
  do jk = 1, nlay_i, 1
     ztmp(jk) = ...
  enddo
  ... uses ztmp ...
end do
```

We could privatise ztmp in the outer loop. We don't but this is not a problem because the dependency analysis also correctly reports:
`The write access to 'ztmp' in 'ztmp(jk)' causes a write-write race condition.`
So we skip the outer loop and only parallelise the inner one.

However, it would be more efficient to parallelise the outer loop and mark ztmp as private.

There are three solutions to explore:
- Mark them as private manually in the transformation (this this needs a new attribute to carry the information until the backend):
    `OMPLoopTrans.apply(loop, options={"private": ["ztmp"]})`

- Mark them as private manually in a different preceding transformation. Since PSyIR has nested scopes we could move it to the loop_body scope (after validating that it is not used outside). Then infer_sharing_attributes could use this case to assume private (kind of how it works in C). This could have uses outside loop parallelisation.

- Or let the validate accept this kind of dependency and make the infer_sharing_attributes identify and mark them. This would be better because it won't require manually adding them one by one, but I am afraid of false positives.  We should also look at compiler literature for this, unfortunately the dependency analysis book that we have been following only have scalar privatisation.

@arporter @hiker @LonelyCat124  Any ideas for this? This seems a pattern that will be important for NEMO5

sergisiso:


src/psyclone/psyir/nodes/reference.py
```diff
@@ -257,6 +257,37 @@ def next_accesses(self):
         chain = DefinitionUseChain(self)
         return chain.find_forward_accesses()
 
+    def escapes_scope(self, scope, visited_nodes=None) -> bool:
```
sergisiso, line 260: I also add @LonelyCat124 as a reviewer, particularly because I added this method that uses DefUseChain to decide if a reference can be private (by checking if it is ever used after a given scope, .e.g. a loop). I am also planning a symmetric method looping backwards to replace the logic about when the private should be firstprivate.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -780,6 +780,10 @@ def is_available_on_device(self, device_string: str = "") -> bool:
             supported values.
 
         '''
+        if self.intrinsic in REDUCTION_INTRINSICS:
+            if len(self.arguments) > 1:
```
arporter, line 786: Please add a comment here.

src/psyclone/psyir/nodes/reference.py
```diff
@@ -257,6 +257,37 @@ def next_accesses(self):
...
+        '''
+        :param scope: the given scope that we evaluate.
+        :param visited_nodes: a set of nodes already visited, this is necessary
+            because the dependency chains may contain cycles. Defaults to an
+            empty set.
+        :returns: whether the symbol lifetime expands after the given scope.
```
arporter: s/expands/continues/

src/psyclone/psyir/nodes/reference.py
```diff
@@ -257,6 +257,37 @@ def next_accesses(self):
         chain = DefinitionUseChain(self)
         return chain.find_forward_accesses()
 
+    def escapes_scope(self, scope, visited_nodes=None) -> bool:
+        '''
+        :param scope: the given scope that we evaluate.
+        :param visited_nodes: a set of nodes already visited, this is necessary
+            because the dependency chains may contain cycles. Defaults to an
+            empty set.
```
arporter, line 265: It defaults to None?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -630,6 +630,13 @@ def _find_basic_blocks(self, nodelist):
                 # No control for the condition - we always check that.
                 control_flow_nodes.append(None)
                 basic_blocks.append([node.condition])
+                if node.ancestor((Loop, WhileLoop)):
+                    control_flow_nodes.append(node)
```
arporter, line 641: I have to admit that I've not looked at this code before and I don't really understand it. I would say (of course) that this new bit needs at least a comment. It also appears odd to me that at both L634 and L637 you append the same `node` to `control_flow_nodes` - is that right? Apart from that I defer to @LonelyCat124 .

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -630,6 +630,13 @@ def _find_basic_blocks(self, nodelist):
...
+                    basic_blocks.append(node.if_body.children[:])
+                    if node.else_body:
+                        control_flow_nodes.append(node)
+                        basic_blocks.append(node.else_body.children[:])
+                    continue
```
LonelyCat124, line 646: @arporter I'm happy with this change - it is similar to my suggestion in #3061 but correctly handles whether there is an else_body.

@sergisiso There should probably be a unit test for this section added in `definition_use_chains_forward_dependence_test.py::test_definition_use_chain_find_basic_blocks` though, as the code inside that doesn't currently consider this case.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -159,6 +159,18 @@ def test_intrinsiccall_is_available_on_device(intrinsic, result):
...
 
 
+def test_intrinsiccall_reductions_is_available_on_device():
+    '''Tests that the is_available_on_device() refuses reduction intrinsics
+    with optional arguments'''
+    intrinsic_call = IntrinsicCall(IntrinsicCall.Intrinsic.SUM)
+    intrinsic_call.addchild(Reference(DataSymbol("result", REAL_TYPE)))
+    # This is avaliabe on the device
+    assert intrinsic_call.is_available_on_device()
+    # But not when it has more arguments
```
arporter: As in the code itself, please explain the reason that it's not supported with arguments.

src/psyclone/tests/psyir/transformations/parallel_loop_trans_test.py
```diff
@@ -662,6 +663,73 @@ def test_paralooptrans_with_array_privatisation(fortran_reader,
...
+    trans = ParaTrans()
+
+    # By default this can not be parallelised because 'ztmp' has a write-write
+    # race condition in the outer loops
+    with pytest.raises(TransformationError) as err:
+        trans.validate(loop)
+    assert ("ztmp(jj)\' causes a write-write race condition."
+            in str(err.value))
+
+    # If should succeed if we enable array privatisation
```
arporter: "It"

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -98,40 +97,40 @@ def _directive(self, children, collapse=None):
...
-               if isinstance(ref, Reference)):
+        refs_in_loop = filter(lambda x: x.symbol is sym, loop.walk(Reference))
+        last_access = list(refs_in_loop)[-1]
+        loop.compute_cached_abs_positions()
+        if last_access.escapes_scope(loop):
             return False
 
         if not dry_run:
-            node.explicitly_private_symbols.add(sym)
+            loop.explicitly_private_symbols.add(sym)
```
arporter, line 133: I'm wondering whether `Loop.node_str` should include this list of private variables? It might help with debugging.

src/psyclone/psyir/tools/definition_use_chains.py
sergisiso: I added a comment, control_flow_nodes has sometimes repeated nodes, as it must match the size of basic_blocks

src/psyclone/psyir/tools/definition_use_chains.py
sergisiso: Added a new test

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: Done

src/psyclone/psyir/nodes/reference.py
sergisiso: Done

src/psyclone/psyir/nodes/reference.py
sergisiso: But None means an empty set. I think is clearer this way.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
sergisiso: Done

src/psyclone/tests/psyir/transformations/parallel_loop_trans_test.py
sergisiso: Done

src/psyclone/psyir/transformations/parallel_loop_trans.py
sergisiso: Yes it should. Added

src/psyclone/psyir/tools/definition_use_chains.py
arporter: Thanks.

sergisiso:
@LonelyCat124 The async in both nemov4 and nemov5 fail with "Transformation Error: Found a nowait with no barrier satisfying its dependency which is unsupported behaviour for OMPMinimiseSyncTrans.". I excluded the first file that failed with this but I got another relatively soon in the build so it is probably something to consider. Do you now if adding the loop to itself dependency in the DUC is causing these and what we can do about it?

LonelyCat124:
> @LonelyCat124 The async in both nemov4 and nemov5 fail with "Transformation Error: Found a nowait with no barrier satisfying its dependency which is unsupported behaviour for OMPMinimiseSyncTrans.". I excluded the first file that failed with this but I got another relatively soon in the build so it is probably something to consider. Do you now if adding the loop to itself dependency in the DUC is causing these and what we can do about it?

Its possible - can you find out what loop/dependency resulted in this? Since the previous implementation saw no issues then I suspect the changes to the DUC must be causing false self-dependencies which would cause this.

PR #3067, opened 2025-07-18
Updates to InlineTrans to support optional arguments

schreiberx:
This PR would solve the issue https://github.com/stfc/PSyclone/issues/2525


src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -914,8 +931,8 @@ def merge(self, other_table, symbols_to_skip=()):
         :type other_table: :py:class:`psyclone.psyir.symbols.SymbolTable`
         :param symbols_to_skip: an optional list of Symbols to exclude from
                                 the merge.
-        :type symbols_to_skip: Iterable[
-            :py:class:`psyclone.psyir.symbols.Symbol`]
+        :param check_unresolved_symbols: If `True`, also check unresolved
```
arporter: Please update the method description in this docstring to say what happens if this parameter is False and there is a clash (apparent or otherwise) between two unresolved symbols. What's the worst case scenario?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
+        :param options: A dictionary with options for transformations.
+        :param bool options["force"]:
+            whether or not to permit the inlining
             of Routines containing CodeBlocks. Default is False.
+        :param bool options["check_argument_strict_array_datatype"]:
+            If `True`, make strict checks for matching arguments of
+            array data types.
+            If `False`, it's sufficient that both arguments are of
+            ArrayType. Then, no further checks are performed, defaults
+            to None
```
arporter: Should this be True or False rather than None?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
+        :param bool options["check_matching_arguments"]:
+            If `True`, check for all arguments
+            to match. If `False`, if no matching argument was found, take
```
arporter: the "1st one in list" needs to be a bit clearer. Perhaps "the 1st of the potential callees"

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
+            1st one in list. Defaults to `True`
+        :param bool options["check_codeblocks"]:
+            If `True`, raise Exception
+            if encountering code blocks, defaults to None
```
arporter: 'None' should probably be one of True or False. Probably True since that's the safe option.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
+        :param bool options["check_unsupported_type"]:
+            If `True`, also perform checks (fail inlining) on arguments of
+            unsupported type, defaults to None
```
arporter: Again, probably should be True.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
         :raises InternalError: if the merge of the symbol tables fails.
             In theory this should never happen because validate() should
             catch such a situation.
 
         '''
-        self.validate(node, options)
+
+        options = {} if options is None else options
+        option_use_first_callee_and_no_arg_check = options.get(
+            "use_first_callee_and_no_arg_check", False
```
arporter: This appears to be an agglomeration of two separate options mentioned in the docstring.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
         '''
         Takes the body of the routine that is the target of the supplied
         call and replaces the call with it.
 
-        :param node: target PSyIR node.
-        :type node: :py:class:`psyclone.psyir.nodes.Routine`
-        :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
-        :param bool options["force"]: whether or not to permit the inlining
+        :param call_node: Target PSyIR node.
```
arporter: s/call_node/node/

src/psyclone/psyir/transformations/omp_task_trans.py
```diff
@@ -112,17 +112,18 @@ def validate(self, node, options=None, **kwargs):
         kintrans = KernelModuleInlineTrans()
         cond_trans = FoldConditionalReturnExpressionsTrans()
         intrans = InlineTrans()
+        options = {"check_matching_arguments": False}
```
arporter: Is there a good reason for having to turn this off? If the answer is "to make the tests pass" we should probably expose this as an option on this transformation itself and then we can set it in the test(s) explicitly with an appropriate comment. I don't want to go to town here since this transformation is some way off being ready for production but I don't want to make it worse :-)

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -1749,3 +1751,177 @@ def test_get_callees_import_private_clash(fortran_reader, tmpdir, monkeypatch):
             "but that Container defines a private Symbol of the same name. "
             "Searching for the Container that defines a public Routine with "
             "that name is not yet supported - TODO #924" in str(err.value))
+
+
+def test_unresolved_types(fortran_reader):
```
arporter: This needs renaming and a new docstring - it's a test of Call.get_callees and there aren't any unresolved symbols.

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -1749,3 +1751,177 @@ def test_get_callees_import_private_clash(fortran_reader, tmpdir, monkeypatch):
...
+        "  real :: opt\n"
+        "  x = x + 1.0\n"
+        "end subroutine sub\n"
+        "end module test_mod\n"
+    )
+
+    psyir = fortran_reader.psyir_from_source(code)
+    call: Call = psyir.walk(Call)[0]
+
+    call.get_callees(options={"ignore_unresolved_types": True})
```
arporter: Please check the result of get_callees. Also, I don't understand why you have `ignore_unresolved_types` here because, AFAICS, there aren't any unresolved types in the example?

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -300,7 +303,13 @@ def test_apply_struct_arg(fortran_reader, fortran_writer, tmpdir):
     psyir = fortran_reader.psyir_from_source(code)
     inline_trans = InlineTrans()
     for routine in psyir.walk(Routine)[0].walk(Call, stop_type=Call):
```
arporter: Not yours but better to s/routine/call/ as that's what it is.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -300,7 +303,13 @@ def test_apply_struct_arg(fortran_reader, fortran_writer, tmpdir):
...
-        inline_trans.apply(routine)
+        inline_trans.apply(
+            routine,
+            options={
+                "check_matching_arguments": False,
```
arporter: I would have thought that not checking for matching arguments would be the same as just taking the first callee without an arg check?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
         Some of these restrictions will be lifted by #924.
 
     '''
-    def apply(self, node, options=None):
+    def apply(
+        self,
+        node: Call,
+        orig_routine: Routine = None,
```
arporter: Should be `Optional[Routine]`

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
         Takes the body of the routine that is the target of the supplied
         call and replaces the call with it.
 
-        :param node: target PSyIR node.
-        :type node: :py:class:`psyclone.psyir.nodes.Routine`
-        :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
-        :param bool options["force"]: whether or not to permit the inlining
+        :param call_node: Target PSyIR node.
+        :param routine_node: Routine to be inlined.
```
arporter: Please extend to say that if it is not provided (the default) then the Callee with an interface matching that of the Call will be inlined.

src/psyclone/psyir/nodes/call.py
```diff
@@ -498,7 +502,6 @@ def get_callees(self):
                         if psyir:
                             routines.append(psyir)
                     if routines:
-                        rsym.interface = DefaultModuleInterface()
```
arporter: Why have you changed this?

src/psyclone/psyir/nodes/call.py
```diff
@@ -656,27 +684,38 @@ def _check_argument_type_matches(
                 call_arg.datatype
                 != routine_arg.datatype.partial_datatype
             ):
+                call_arg_str = str(call_arg).replace("\n", "")
```
arporter: I think you want `call_arg.debug_string()` and same for `routine_arg`.

src/psyclone/psyir/nodes/call.py
```diff
@@ -656,27 +684,38 @@ def _check_argument_type_matches(
...
                 raise CallMatchingArgumentsNotFound(
                     f"Argument partial type mismatch of call "
-                    f"argument '{call_arg}' and routine argument "
-                    f"'{routine_arg}'"
+                    f"argument '{call_arg_str}' and routine argument "
+                    f"'{routine_arg_str}'"
                 )
         else:
             if call_arg.datatype != routine_arg.datatype:
+                call_arg_str = str(call_arg).replace("\n", "")
```
arporter: `.debug_string()` again please.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Also, could you explain the purpose of this argument.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,28 +135,56 @@ class InlineTrans(Transformation):
...
 
         '''
-        self.validate(node, options)
+
+        options = {} if options is None else options
+        option_use_first_callee_and_no_arg_check = options.get(
+            "use_first_callee_and_no_arg_check", False
+        )
+
+        (orig_routine, arg_index_list) = self.validate(
```
arporter: The `validate` method does not usually return anything so, for consistency, please refactor this one so it doesn't either.

src/psyclone/psyir/nodes/call.py
```diff
@@ -656,27 +684,38 @@ def _check_argument_type_matches(
...
-                    f"'{call_arg}' and routine argument "
-                    f"'{routine_arg}'"
+                    f"'{call_arg_str}' and routine argument "
+                    f"'{routine_arg_str}'"
                 )
 
-    def _get_argument_routine_match(self, routine: Routine):
+    def _get_argument_routine_match(
+        self,
+        routine: Call,
```
arporter: This is a `Routine`, not a `Call`?

src/psyclone/psyir/nodes/call.py
```diff
@@ -686,13 +725,14 @@ def _get_argument_routine_match(self, routine: Routine):
...
+                f"More arguments in call ('{call_str}')"
                 f" than callee (routine '{routine.name}')"
             )
 
-        # Iterate over all arguments to the call
         ret_arg_idx_list = []
+        # Iterate over all arguments to the call
         for call_arg_idx, call_arg in enumerate(self.arguments):
             call_arg_idx: int
             call_arg: DataSymbol
```
arporter: Not from this PR but isn't `call_arg` a `DataNode` rather than a symbol?

src/psyclone/psyir/nodes/call.py
```diff
@@ -750,33 +796,35 @@ def _get_argument_routine_match(self, routine: Routine):
...
-        :param check_matching_arguments: Also check argument types to match.
+        :param options: a dictionary of options.
+        :param bool options["check_matching_arguments"]:
+            Also check argument types to match.
             If set to `False` and in case it doesn't find matching arguments,
             the very first implementation of the matching routine will be
             returned (even if the argument type check failed). The argument
             types and number of arguments might therefore mismatch!
-        :type ret_arg_match_list: bool
+            This argument exists for backwards compatibility.
```
arporter: I think I may have misunderstood one of your earlier questions. I think you probably are the only people using this functionality so we can drop backwards compatibility here.

src/psyclone/psyir/nodes/call.py
```diff
@@ -807,13 +862,14 @@ def get_callee(
         # arguments have been found.
         # This is handy for the transition phase until optional argument
```
arporter: Optional argument matching is now implemented though?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+            routine_node.symbol_table.datasymbols
+        ):
+            if not isinstance(datasymbol.datatype, UnsupportedFortranType):
+                continue
+
+            if ", OPTIONAL" not in str(datasymbol.datatype):
+                continue
+
+            sym_name = datasymbol.name.lower()
+
```
arporter: This next bit can just be:
```
optional_sym_present_dict[sym_name] = optional_arg_idx in arg_match_list
```

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+                optional_sym_present_dict[sym_name] = True
+
+        # Check if we have any optional arguments at all and if not, return
+        if len(optional_sym_present_dict) == 0:
+            return
+
+        # Find all "PRESENT()" calls
+        for intrinsic_call in routine_node.walk(IntrinsicCall):
+            intrinsic_call: IntrinsicCall
+            if intrinsic_call.routine.name.lower() == "present":
```
arporter: `instrinsic_call.intrinsic is IntrinsicCall.Intrinsic.PRESENT`

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+
+                # The argument is in the 2nd child
+                present_arg: Reference = intrinsic_call.children[1]
```
arporter: s/children[0]/arguments[0]/ I think.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+                present_arg_name = present_arg.name.lower()
+
+                assert present_arg_name in optional_sym_present_dict
```
arporter: No bare assert please.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+                    intrinsic_call.replace_with(Literal("true", BOOLEAN_TYPE))
+                else:
+                    intrinsic_call.replace_with(Literal("false", BOOLEAN_TYPE))
+
+    def _optional_arg_eliminate_ifblock_if_const_condition(
+        self, routine_node: Routine
+    ):
+        """Eliminate if-block if conditions are constant booleans.
+
+        TODO: This also requires support of conditions containing logical
```
arporter: This needs an issue number. Possibly our Sympy support can already do this?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+        expressions such as `(.true. .or. .false.)`
+        """
+
+        def if_else_replace(main_schedule, if_block, if_body_schedule):
+            """Little helper routine to eliminate one branch of an IfBlock
+
+            :param main_schedule: Schedule where if-branch is used
+            :type main_schedule: Schedule
```
arporter: type-hints please.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+            :param if_block: If-else block itself
+            :type if_block: IfBlock
+            :param if_body_schedule: The body of the if or else block
+            :type if_body_schedule: Schedule
+            """
+
+            from psyclone.psyir.nodes import Schedule
+
+            assert isinstance(main_schedule, Schedule)
```
arporter: No bare asserts please. Since this is very much an internal, helper method we can drop these checks.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+            assert isinstance(if_body_schedule, Schedule)
+
+            # Obtain index in main schedule
+            idx = main_schedule.children.index(if_block)
+
+            # Detach it
+            if_block.detach()
+
+            # Insert childreen of if-body schedule
+            for child in if_body_schedule.children:
```
arporter: Might be nicer to use `pop_all_children` here.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+
+            :param main_schedule: Schedule where if-branch is used
+            :type main_schedule: Schedule
+            :param if_block: If-else block itself
+            :type if_block: IfBlock
+            :param if_body_schedule: The body of the if or else block
+            :type if_body_schedule: Schedule
+            """
+
+            from psyclone.psyir.nodes import Schedule
```
arporter: Imports at top level if poss.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+
+            # Detach it
+            if_block.detach()
+
+            # Insert childreen of if-body schedule
+            for child in if_body_schedule.children:
+                main_schedule.addchild(child.copy(), idx)
+                idx += 1
+
+        from psyclone.psyir.nodes import IfBlock
```
arporter: Imports at top level pls.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +353,132 @@ def apply(self, node, options=None):
...
+
+            condition = if_block.condition
+
+            # Make sure we only handle a BooleanLiteral as a condition
+            # TODO #2802
+            if not isinstance(condition, Literal):
+                continue
+
+            # Check that it's a boolean Literal
+            assert (
```
arporter: No bare assert please. Possibly the PSyIR does (or should) guarantee that a single Literal in this location is always boolean.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -280,18 +497,56 @@ def _replace_formal_arg(self, ref, call_node, formal_args):
...
             # The supplied reference is not to a formal argument.
             return ref
 
-        # Lookup the actual argument that corresponds to this formal argument.
-        actual_arg = call_node.arguments[formal_args.index(ref.symbol)]
+        # Lookup index in routine argument
+        routine_arg_idx = formal_args.index(ref.symbol)
+
+        if option_use_first_callee_and_no_arg_check:
+            actual_arg = call_node.arguments[formal_args.index(ref.symbol)]
```
arporter: Comment please.

arporter: Can use `routine_arg_idx` here rather than re-computing it.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -327,17 +582,31 @@ def _replace_formal_arg(self, ref, call_node, formal_args):
...
+        call_node,
+        formal_args,
+        local_idx,
+        decln_start,
+        actual_start,
+        routine_node: Routine,
+        options,
+    ):
         '''
         Utility that creates the PSyIR for an inlined array-index access
```
arporter, line 567: Pls update docstring to include new, `routine_node` argument

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -374,30 +643,66 @@ def _create_inlined_idx(self, call_node, formal_args,
...
+    def _update_actual_indices(
+        self,
+        actual_arg,
+        local_ref,
+        call_node,
+        formal_args,
+        routine_node: Routine,
+        options,
+    ):
         '''
```
arporter, line 663: pls document the new `routine_node` and `options` arguments.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -609,7 +957,180 @@ def _replace_formal_struc_arg(self, actual_arg, ref, call_node,
...
 
-    def validate(self, node, options=None):
+    def _validate_inline_of_call_and_routine_argument_pairs(
+        self,
+        call_node: Call,
+        call_arg: DataSymbol,
+        routine_node: Routine,
+        routine_arg: DataSymbol
+    ):
+        """This function performs tests to see whether the
```
arporter: tests of what?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -609,7 +957,180 @@ def _replace_formal_struc_arg(self, actual_arg, ref, call_node,
...
+        inlining can cope with it.
+
+        :param call_node: The call used for inlining
+        :param call_arg: The argument of a call
+        :param routine: The routine to be inlined
+        :param routine_arg: The argument of a routine
+
+        :raises TransformationError: Raised if transformation can't be done
```
arporter: Please be a bit more specific about the possible reasons.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -609,7 +957,180 @@ def _replace_formal_struc_arg(self, actual_arg, ref, call_node,
...
+
+        # If the formal argument is an array with non-default bounds then
+        # we also need to know the bounds of that array at the call site.
+        if not isinstance(routine_arg.datatype, ArrayType):
+            # Formal argument is not an array so we don't need to do any
+            # further checks.
+            return
+
+        if not isinstance(call_arg, (Reference, Literal)):
+            # TODO #1799 this really needs the `datatype` method to be
```
arporter: The `datatype` method *has* been extended now. (Albeit, not to function calls yet.)

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -609,7 +957,180 @@ def _replace_formal_struc_arg(self, actual_arg, ref, call_node,
...
+            # of any general expression is or is not an array.
+            # pylint: disable=cell-var-from-loop
+            raise TransformationError(LazyString(
+                lambda: f"The call '{call_node.debug_string()}' cannot be "
+                        f"inlined because actual argument "
+                        f"'{call_arg.debug_string()}' corresponds to a "
+                        f"formal argument with array type but is not a "
+                        f"Reference or a Literal."))
+
+        # We have an array argument. We are only able to check that the
```
arporter: Although github is marking this as new code, ~it doesn't look like code/comments you have written? Is it old code accidentally brought back in when merging?~ I see this is because this method has moved. If you put it back in its original location in this file then this will hopefully improve the diff.

src/psyclone/psyir/symbols/symbol_table.py
arporter: I've reverted this change as it seemed to be unused and didn't fully work.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -609,7 +957,180 @@ def _replace_formal_struc_arg(self, actual_arg, ref, call_node,
...
+                    # TODO #1646. We could resolve this problem by making
+                    # a new array and copying the necessary values into it.
+                    # pylint: disable=cell-var-from-loop
+                    raise TransformationError(LazyString(
+                        lambda: f"Cannot inline routine '{routine_node.name}' "
+                        f"because one of its arguments is an array slice "
+                        f"with a non-unit stride: "
+                        f"'{call_arg.debug_string()}' (TODO #1646)"))
+
+    def get_callee(
```
arporter: This method should not be here. We already have `Call.get_callee()`

src/psyclone/psyir/nodes/call.py
arporter: I've put this back but it seems not to have affected any tests. Perhaps it is not required or perhaps we are missing a test?

src/psyclone/psyir/transformations/omp_task_trans.py
arporter: Removed and all `transformations/omp*_test.py` tests pass OK.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Gone.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: I've moved this check into the `IfBlock.condition` getter and extended the appropriate test.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Gone now.

src/psyclone/tests/psyir/nodes/call_test.py
arporter: Gone now (or renamed).

src/psyclone/tests/psyir/nodes/call_test.py
arporter: Gone now.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
arporter: Done.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
arporter: Gone now.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

src/psyclone/psyir/nodes/call.py
arporter: Done.

src/psyclone/psyir/nodes/call.py
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Done.

src/psyclone/psyir/transformations/inline_trans.py
arporter: I've checked and core.SymbolicMaths can already do this. Therefore, I've added a line to ensure that the `condition` is simplified if possible and removed this TODO.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Updated the wording.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Added.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Updated.

arporter:
Note that some of the functionality in this PR is also in #2869. Probably it should be a separate transformation but it might then also need to be more general than the implementation here.

src/psyclone/psyir/nodes/call.py
```diff
@@ -633,69 +637,78 @@ def get_callees(self):
...
     def _check_argument_type_matches(
                 self,
-                call_arg: DataSymbol,
-                routine_arg: DataSymbol,
-            ) -> None:
-        """Return information whether argument types are matching.
-        This also supports 'optional' arguments by using
+                call_arg: DataNode,
+                routine_arg: DataSymbol
+            ):
```
hiker: The `-> None` has disappeared ;)

src/psyclone/psyir/nodes/call.py
```diff
@@ -750,70 +766,63 @@ def _get_argument_routine_match(self, routine: Routine):
             # TODO #759: Optional keyword is not yet supported in psyir.
             # Hence, we use a simple string match.
             if ", OPTIONAL" not in str(routine_arg.datatype):
+                routine_arg_name = routine_arg.name.replace("\n", "")
```
hiker: Are the first two `replace`s actually necessary? I would assume that the name attribute is always a string without newlines.

src/psyclone/psyir/nodes/call.py
```diff
@@ -750,70 +766,63 @@ def _get_argument_routine_match(self, routine: Routine):
...
-        match was found. Then, the arguments of the call and routine
-        might not match each other.
-
-        :param check_matching_arguments: Also check argument types to match.
-            If set to `False` and in case it doesn't find matching arguments,
-            the very first implementation of the matching routine will be
-            returned (even if the argument type check failed). The argument
-            types and number of arguments might therefore mismatch!
-        :type ret_arg_match_list: bool
+        :param use_first_callee_and_no_arg_check: whether or not (the default)
```
hiker, line 842: Could you leave the original content of the doc string, I think it might be useful to highlight the potential risks:

> if `use_first_callee_and_no_arg_check` is set to ... might not match each other

src/psyclone/psyir/nodes/call.py
```diff
@@ -633,69 +637,78 @@ def get_callees(self):
...
-        if isinstance(
-            routine_arg.datatype, UnsupportedFortranType
-        ):
-            # This could be an 'optional' argument.
-            # This has at least a partial data type
-            if (
-                call_arg.datatype
-                != routine_arg.datatype.partial_datatype
-            ):
+        if isinstance(call_arg.datatype, ArrayType) and isinstance(
```
hiker: I don't understand this part - any arrays are just accepted (e.g. not even comparing the number of dimensions)? If that is a temporary work around, please add a #TODO. If I am missing something, I would appreciate a comment here.

src/psyclone/psyir/nodes/if_block.py
```diff
@@ -89,20 +91,31 @@ def _validate_child(position, child):
...
+
         '''
         if len(self.children) < 2:
             raise InternalError(
                 f"IfBlock malformed or incomplete. It should have at least 2 "
                 f"children, but found {len(self.children)}.")
-        return self._children[0]
+
+        cond = self._children[0]
+        if isinstance(cond, Literal) and (cond.datatype.intrinsic
```
hiker: This seems to be unrelated to this PR?? Not sure if I am missing something.
I am also not quite certain if that is the right place to test this, shouldn't that be when creating the node? After all, it might be that a user script just uses `if_node.children[0]` (yes, it's bad of course ;) ), but I would try to prevent this from happening in the first place.

src/psyclone/psyir/tools/__init__.py
```diff
@@ -37,9 +37,9 @@
 '''
 
 from psyclone.psyir.tools.call_tree_utils import CallTreeUtils
+from psyclone.psyir.tools.definition_use_chains import DefinitionUseChain
```
hiker, line 41: Thanks for fixing this

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -668,7 +667,7 @@ def test_call_get_callees_local(fortran_reader):
   end subroutine bottom
 end module some_mod'''
     psyir = fortran_reader.psyir_from_source(code)
-    call = psyir.walk(Call)[0]
+    call: Call = psyir.walk(Call)[0]
```
hiker: I admit I learned something. I thought that mypy would flag this (if I add typing to node.walk), since it assigns a `Node` to a variable that only stores `Call`. I used to add either `cast` or `assert isinstance(call, Call)` to get mypy, but it looks like this works, and is much neater - thanks

hiker: Ah, I need to take that back, I had a typo when I added the type info to walk, which mypy apparently happily ignored?? Once done properly (with '"'):
```
def walk(self, my_type, stop_type=None, depth=None) -> List["Node"]:
```
I now get:
```
 error: Incompatible types in assignment (expression has type "Node", variable has type "Call") 
```
That works for me:
```
    call: Call = cast(Call, psyir.walk(Call)[0])
```

src/psyclone/tests/psyir/nodes/call_test.py
```diff
@@ -1749,3 +1780,91 @@ def test_get_callees_import_private_clash(fortran_reader, tmpdir, monkeypatch):
...
+    call._check_argument_type_matches(call.arguments[0],
+                                      DataSymbol("dummy1", REAL_TYPE))
+    # Integer instead of real.
+    with pytest.raises(CallMatchingArgumentsNotFound) as err:
+        call._check_argument_type_matches(call.arguments[0],
+                                          DataSymbol("dummy1", INTEGER_TYPE))
+    assert "Argument type mismatch of call argument 'var'" in str(err.value)
+    # For an array argument.
+    call2 = calls[1]
+    call2._check_argument_type_matches(
```
hiker, line 1894: Since I didn't understand the array handling above, does this need a #TODO, or better a x-failing test?

src/psyclone/tests/psyir/nodes/if_block_test.py
```diff
@@ -109,6 +109,15 @@ def test_ifblock_properties():
     assert ("IfBlock malformed or incomplete. It should have "
             "at least 2 children, but found 0." in str(err.value))
 
+    # Condition must be a boolean if it is a Literal.
+    ifblock.addchild(Literal("0.0", REAL_SINGLE_TYPE))
```
hiker: Again, not really related to this PR (but since it is apparently small, I am happy to leave that in), BUT: the test should be added to the `_validate` method, so it should fail when `addchild`.

src/psyclone/tests/psyir/transformations/omp_task_transformations_test.py
```diff
@@ -183,24 +183,31 @@ def test_omptask_apply_kern(fortran_reader, fortran_writer):
...
     taskt = OMPTaskTrans()
     schedule = invoke.schedule
+
     # Currently the InlineTrans validation will reject the GOcean kernel call
-    # because it can't determine the type of `fld%data` being passed in. We
-    # therefore monkeypatch the validate() method to get round this.
-    monkeypatch.setattr(InlineTrans, "validate", lambda _1, _2, _3: None)
-    taskt._inline_kernels(schedule.children[0])
-    assert not schedule.walk(Kern)
+    # because it can't determine the type of `fld%data` being passed in.
```
hiker, line 198: While I agree with this additional test, I think the original test should still be in place (i.e. that it works if we could determine the right type). I can't easily check the coverage change (in theory we should have one test file for each PSyclone source files that covers it), but this file is one of the existing exceptions, and not really your problem :)
So I'd prefer to play it safe, and add the original test with monkey patch (in the same function), just to test the inlining.

utils/run_pytest_cov.sh
```diff
@@ -23,10 +23,13 @@ COV_REPORT="xml:cov.xml"
 
 # Additional options
 # Also write to Terminal
-#OPTS=" --cov-report term"
+OPTS=" --cov-report term"
+
+if [[ -e cov.xml ]]; then
+	echo "Removing previous reporting file 'cov.xml'"
+	rm -rf cov.xml
```
hiker: I admit I wasn't aware that this existed ... nice :)
But can we keep a backup of the previous xml file? e.g.
```
rm -f cov.xml.backup
mv cov.xml cov.xml.backup
```
That might be handy if you want to check for changes in the coverage (while developing code)

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,42 +139,90 @@ class InlineTrans(Transformation):
...
             return
 
         # Ensure we don't modify the original Routine by working with a
         # copy of it.
-        container = orig_routine.ancestor(Container).copy()
-        routine = container.find_routine_psyir(orig_routine.name,
-                                               allow_private=True)
+        container: Container = routine.ancestor(Container).copy()
+        routine: Routine = container.find_routine_psyir(
+            routine.name, allow_private=True
```
hiker: As previously, I don't think mypy would accept this (if find_routine_psyir was typed). Importing `cast` and then `routine = cast(Routine, ...)` should work.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,42 +139,90 @@ class InlineTrans(Transformation):
...
+        )
         routine_table = routine.symbol_table
 
+        if not use_first_callee_and_no_arg_check:
```
hiker: Why is this not done if the user specifies a routine? As far as I can see, the same changes should be done in that case.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -122,42 +139,90 @@ class InlineTrans(Transformation):
...
+
+            # Step 2)
+            # - For all If-Statements, handle constant conditions:
+            #   - `True`: Replace If-Block with If-Body
+            #   - `False`: Replace If-Block with Else-Body. If it doesn't exist
+            #     just delete the if statement.
+            self._optional_arg_eliminate_ifblock_if_const_condition(routine)
+
+        else:
+            arg_match_list = list(
```
hiker: That variable is already set above (if a routine is specified).

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -208,10 +277,22 @@ def apply(self, node, options=None):
                     if isinstance(dim, ArrayType.Extent):
```
hiker: Not a requirement, but a suggestion: if you change the if statement in line 274 (sorry, I can't add a comment at the right place) to:
```
if not isinstance(...):
    continue
```
then you can de-dent the rest, giving you more space, and you reduce nesting level

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -208,10 +277,22 @@ def apply(self, node, options=None):
...
                         new_shape.append(dim)
                     else:
-                        lower = self._replace_formal_arg(dim.lower, node,
-                                                         formal_args)
-                        upper = self._replace_formal_arg(dim.upper, node,
-                                                         formal_args)
+                        lower = self._replace_formal_arg(
```
hiker: Can you put more arguments in one line, there should be plenty of space. While I totally favour using one-argument-per-line when declaring the arguments of a function (so there is proper space for typing and it's easier to see what is an argument and what is typing), in this case these two calls take up nearly all my screen space on my laptop

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -222,10 +303,22 @@ def apply(self, node, options=None):
                 if isinstance(ctype.datatype, ArrayType):
                     new_shape = []
                     for dim in ctype.datatype.shape:
-                        lower = self._replace_formal_arg(dim.lower, node,
-                                                         formal_args)
-                        upper = self._replace_formal_arg(dim.upper, node,
-                                                         formal_args)
+                        lower = self._replace_formal_arg(
+                            dim.lower,
```
hiker: Same, more arguments in one line to make better use of screen space

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +354,117 @@ def apply(self, node, options=None):
                 idx += 1
                 parent.addchild(child, idx)
 
-    def _replace_formal_arg(self, ref, call_node, formal_args):
+    def _optional_arg_resolve_present_intrinsics(
+        self, routine_node: Routine, arg_match_list: List = []
```
hiker: Here I would actually suggest:
```
 def _optional_arg_resolve_present_intrinsics(
            self,
            routine_node: Routine,
            arg_match_list: List = []
```
to make it easier to see what is an argument.

src/psyclone/psyir/transformations/inline_trans.py
hiker: @arporter , I am a but confused, since I still see the assert. What would you prefer? If we raise an exception, that line will be difficult (and quite hacky) to test for  :(

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -261,7 +354,117 @@ def apply(self, node, options=None):
...
+
+                assert present_arg_name in optional_sym_present_dict
+
+                if optional_sym_present_dict[present_arg_name]:
+                    # The argument is present.
+                    intrinsic_call.replace_with(Literal("true", BOOLEAN_TYPE))
+                else:
+                    intrinsic_call.replace_with(Literal("false", BOOLEAN_TYPE))
+
+    def _optional_arg_eliminate_ifblock_if_const_condition(
```
hiker, line 389: Nice stuff!

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -270,28 +473,56 @@ def _replace_formal_arg(self, ref, call_node, formal_args):
...
         :returns: the replacement reference.
-        :rtype: :py:class:`psyclone.psyir.nodes.Reference`
 
         '''
         if not isinstance(ref, Reference):
             # Recurse down in case this is e.g. an Operation or Range.
             for child in ref.children[:]:
-                self._replace_formal_arg(child, call_node, formal_args)
+                self._replace_formal_arg(
+                    child,
```
hiker: Again, I would prefer less lines here.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -327,17 +558,32 @@ def _replace_formal_arg(self, ref, call_node, formal_args):
 
         # Neither the actual or local references are simple, i.e. they
         # include array accesses and/or structure accesses.
-        new_ref = self._replace_formal_struc_arg(actual_arg, ref, call_node,
-                                                 formal_args)
+        new_ref = self._replace_formal_struc_arg(
```
hiker, line 543: Less lines?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -609,24 +937,23 @@ def _replace_formal_struc_arg(self, actual_arg, ref, call_node,
         # Just an array reference.
         return ArrayReference.create(actual_arg.symbol, members[0][1])
 
-    def validate(self, node, options=None):
+    def validate(
+                self,
+                node: Call,
+                **kwargs
+            ):
```
hiker: Typing: `-> None`

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -677,18 +1008,30 @@ def validate(self, node, options=None):
...
+                    f" cannot be found:\n{str(err)}"
+                ) from err
+
+        else:
+            arg_match_list = node.get_argument_map(routine)
 
         if not routine.children or isinstance(routine.children[0], Return):
             # An empty routine is fine.
-            return
+            return (routine, range(len(routine.symbol_table.argument_list)))
```
hiker: I feel lost. Isn't this the validate function, which should not return a value?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -794,100 +1134,25 @@ def validate(self, node, options=None):
...
-                            f"'{actual_arg.debug_string()}' (TODO #1646)"))
+        # Create a list of routine arguments that is actually used
+        routine_arg_list = [
+            routine_table.argument_list[i] for i in arg_match_list
+        ]
+
+        for routine_arg, actual_arg in zip(
+            routine_arg_list, node.arguments
+        ):
+            self._validate_inline_of_call_and_routine_argument_pairs(
```
hiker, line 1129: Thanks for refactoring.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -720,9 +1063,9 @@ def validate(self, node, options=None):
...
 
-        if routine.walk(CodeBlock) and not forced:
-            # N.B. we permit the user to specify the "force" option to allow
-            # CodeBlocks to be included.
+        if not permit_codeblocks and routine.walk(CodeBlock):
+            # N.B. we allow the user to specify the "permit_codeblocks" option
+            # to allow CodeBlocks to be included.
             raise TransformationError(
                 f"Routine '{name}' contains one or more CodeBlocks and "
                 f"therefore cannot be inlined. (If you are confident that "
```
hiker: In (the new) line 1073, should that be `options={'permit_codeblocks': True}`?

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2277,8 +2341,8 @@ def test_validate_non_unit_stride_slice(fortran_reader):
 
 
 def test_validate_named_arg(fortran_reader):
-    '''Test that the validate method rejects an attempt to inline a routine
-    that has a named argument.'''
+    '''Test that the validate method is successful in an attempt to inline
+    a routine that has a named argument.'''
     # In reality, the routine with a named argument would almost certainly
     # use the 'present' intrinsic but, since that gives a CodeBlock that itself
```
hiker: I believe that needs to be updated.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2292,20 +2356,243 @@ def test_validate_named_arg(fortran_reader):
...
         "end module test_mod\n"
     )
     psyir = fortran_reader.psyir_from_source(code)
     call = psyir.walk(Call)[0]
     inline_trans = InlineTrans()
-    with pytest.raises(TransformationError) as err:
-        inline_trans.validate(call)
-    assert ("Routine 'sub' cannot be inlined because it has a named argument "
-            "'opt' (TODO #924)" in str(err.value))
+    inline_trans.validate(call)
```
hiker: Can you check that you have the expected output (e.g. convert to string, then check if the `present` was replaced as expected.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2292,20 +2356,243 @@ def test_validate_named_arg(fortran_reader):
...
+        "  x = x + 1.0\n"
+        "end subroutine sub\n"
+        "end module test_mod\n"
+    )
+    psyir = fortran_reader.psyir_from_source(code)
+    calls = psyir.walk(Call)
+    inline_trans = InlineTrans()
+    inline_trans.apply(calls[0])
+    output = fortran_writer(psyir)
+    assert ('''\
```
hiker, line 2428: Can you also check that the first `opt` argument is indeed not used? I.e. as a `if(.false.)`?

hiker (changes requested):
Hi, great work! And I learned to appreciate the effort for inlining a lot more.

I have a few questions, but it seems mostly ok.

We might have a slightly different approach when using arguments in calls, you tend to put each argument in a new line, which imho makes lines way to long. @arporter, what is your preference. I am happy to accept the long lines (but given that in one case two calls now take around 20 to 25 lines is a bit annoying.

src/psyclone/psyir/transformations/inline_trans.py
arporter: Oh yes, don't know what I did there :-( I've changed the code to do a `dict.get()` so that, if the key isn't present, we get None and won't crash.

src/psyclone/psyir/nodes/call.py
arporter: Fixed.

src/psyclone/psyir/nodes/call.py
arporter: Agreed and simplified.

src/psyclone/psyir/nodes/call.py
arporter: Done.

src/psyclone/psyir/nodes/call.py
arporter: No, that wasn't very good. I've updated it now with proper checks.

arporter:
Thanks for the review @hiker, I think it's looking a lot better now. Ready for another look :-)

src/psyclone/psyir/nodes/call.py
hiker: Thanks

src/psyclone/psyir/nodes/call.py
hiker: Thanks, now I understand that :)

src/psyclone/tests/psyir/nodes/call_test.py
hiker: Resolving as not relevant ;)

src/psyclone/tests/psyir/nodes/if_block_test.py
hiker: Thanks for taking this out

utils/run_pytest_cov.sh
hiker: Uh - nifty move statement (`mv -f cov.xml{,.backup}`).

src/psyclone/psyir/transformations/inline_trans.py
hiker: Thanks

src/psyclone/psyir/transformations/inline_trans.py
hiker: Thanks!

src/psyclone/psyir/transformations/inline_trans.py
hiker: Thanks

src/psyclone/psyir/transformations/inline_trans.py
hiker: Thanks

src/psyclone/tests/psyir/transformations/inline_trans_test.py
hiker: Thanks!

src/psyclone/psyir/nodes/call.py
hiker: These lines have been changed when merging in master, so I am closing this comment as outdated.

PR #3086, opened 2025-08-04
(Closes #3071) fix transinfo deprecation warning

### LINKED ISSUE #3071: Fix deprecation warnings printed when using TransInfo
Unfortunately, using TransInfo causes deprecation errors about the Dynamo0p3 naming of Transformations. @hiker said this patch will fix it:
```
git diff psyGen.py
diff --git a/src/psyclone/psyGen.py b/src/psyclone/psyGen.py
index 3bf18fb849..748eff21ef 100644
--- a/src/psyclone/psyGen.py
+++ b/src/psyclone/psyGen.py
@@ -2701,7 +2701,8 @@ class TransInfo():
         import inspect
         return [cls for name, cls in inspect.getmembers(module)
                 if inspect.isclass(cls) and not inspect.isabstract(cls) and
-                issubclass(cls, base_class) and cls is not base_class]
+                issubclass(cls, base_class) and cls is not base_class
+                and name[:9] != "Dynamo0p3"]


@dataclass
```

arporter:
Fixes a minor bug whereby *any* use of `TransInfo` would cause a deprecation warning to be printed about the use of "Dynamo0p3..." transformation classes.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,5 @@
+   72) PR #3086 for #3071. Improves TransInfo-related deprecation warnings.
+
    71) PR #3062 for #2671. Improves the support for allowing arrays to be
    declared thread private.
 


arporter:
Integration tests were green.

src/psyclone/psyGen.py
```diff
@@ -2696,13 +2696,24 @@ def get_trans_name(self, name):
                                   f"but expected one of "
                                   f"{self._obj_map.keys()}")
 
-    def _find_subclasses(self, module, base_class):
-        ''' return a list of classes defined within the specified module that
-            are a subclass of the specified baseclass. '''
-        import inspect
+    def _find_subclasses(self, module: type, base_class: type) -> List[type]:
+        '''
+        Return a list of classes defined within the specified module that
```
sergisiso, line 2711: But this function is what TransInfo uses to find the available transformations, so won't this break backwards compatibility for: `TransInfo.get_trans_name("Dynamo03p...` ? which I believe is what LFRic script uses.

hiker:
Good point (though tbh, afaik no lfric transformation uses TransInfo :) ). 

But the following patch might solve it by creating all non-Dynamo0p3 transformations, then pointing the Dynamo0p3 named transformations to the new names?

```
diff --git a/src/psyclone/psyGen.py b/src/psyclone/psyGen.py
index aff4abbc5..f2beb0083 100644
--- a/src/psyclone/psyGen.py
+++ b/src/psyclone/psyGen.py
@@ -2641,10 +2641,20 @@ class TransInfo():
         # create our transformations
         self._objects = []
         self._obj_map = {}
+        # First only import the transformation with new names (not Dynamo0p3):
         for my_class in self._classes:
+            if my_class.__name__[:9] == "Dynamo0p3":
+                continue
             my_object = my_class()
             self._objects.append(my_object)
             self._obj_map[my_object.name] = my_object
+        # Then create the dynamo0p3 equivalents, but point them to the
+        # transformations with the new name
+        for my_class in self._classes:
+            if my_class.__name__[:9] != "Dynamo0p3":
+                continue
+            my_object = self.get_trans_name(f"LFRic{my_class.__name__[9:]}")
+            self._obj_map[my_class.__name__] = my_object
         # TODO #620:
         # Transformations that are in psyir and other subdirectories
         # are not found by TransInfo, so we add some that are used in
```

arporter:
Given that the recommended way to use `TransInfo` is to use `list()` first to see what's available, I would argue that we could leave things the way they are. @sergisiso what's your preference?

arporter:
Since TransInfo now raises a DeprecationWarning, I've fired off the integration tests again.

Issue #2781, opened 2024-11-14
Kernels looping over halo cells and stencils


sergisiso:
there is an issue that I hadn't anticipated with removing two of the extended mesh PSyKAl-lite kernels.

Both of those kernels loop over halo cells, and also use stencils. Given the kernel metadata, PSyclone produces the correct psy-layer, specifying that fields to be read by the stencil need to have a halo exchange to a depth of `halo_depth + stencil_depth`, where `halo_depth` is the depth of cells to loop the kernel over.

However these two kernels seem to only use parts of the stencil that are parallel to the halos, and not perpendicular -- thus in the PSyKAl-lite implementations, the halo exchange is only to a depth of `halo_depth`. We therefore trigger errors when trying to perform a halo exchange up to a depth of `halo_depth + stencil_depth` 

_Originally posted by @tommbendall in https://github.com/stfc/PSyclone/issues/2745#issuecomment-2476146076_

arporter:
@tommbendall when you say it triggers errors: is that because we're asking for too deep a halo? Should we do `MAX(halo_depth, stencil_depth)` or would that just be a sticking plaster?

tommbendall:
Thanks for posting this @sergisiso.

Maybe one solution here is to implement a new type of stencil (on the LFRic infrastructure side) that be need a new type of stencil that is explicitly parallel to the halos.

A hack might be to change the halo exchange to happen on:
`MIN(field%get_halo_depth(), halo_depth + stencil_depth)`

(and as I'm typing I've just seen Andy type the same thing!)

arporter:
We could have a rule that for kernels that operate on halo cells and have a stencil, any halo exchange will always be to the maximum depth? (We already have machinery for handling the special case of maximum depth.)

tommbendall:
I realised that I didn't answer your question above. Yes the failures were because `halo_depth + stencil_depth > field%get_halo_depth()`, so we were trying force a halo exchange to a depth that didn't exist.

Maybe our halo exchanges should be performed to a depth `MAX(field%get_halo_depth(), halo_depth)`.

I was just discussing this with @thomasmelvin and the challenge is when we actually want to fail and when we don't. Could we potentially print out a warning if `field%get_halo_depth() < halo_depth + stencil_depth`? There currently isn't a situation where we need to perform a halo exchange to depth `halo_depth + stencil_depth` but what should happen if someone tries to do that?

[...]

PR #3089, opened 2025-08-05
(Towards #2781) Refine halo-depth logic for LFRic "halo" kernels

arporter:


arporter:
In discussion with @tommbendall I've now added TODOs referring to the need to define a new type of stencil to match what is being done in several of the new 'halo' kernels (where apparently the stencil access is parallel to the halo region).

@tommbendall, you could experiment with this branch now if you like.

Ready for review now from either @hiker or @sergisiso.

src/psyclone/domain/lfric/lfric_constants.py
```diff
@@ -249,13 +249,17 @@ def __init__(self):
         # in the LFRic API ('real' and 'integer')
         LFRicConstants.VALID_BUILTIN_DATA_TYPES = ["gh_real", "gh_integer"]
 
+        # Iteration spaces for user-supplied kernels that must include halo
+        # regions for correctness.
+        LFRicConstants.HALO_KERNEL_ITERATION_SPACES = [
```
hiker, line 254: Nice

src/psyclone/lfric.py
```diff
@@ -4783,7 +4786,7 @@ def halo_check_arg(field, access_types):
     call object containing this argument.
 
     :param field: the argument object we are checking
-    :type field: :py:class:`psyclone.lfric.LFRicArgument`
+    :type field: :py:class:`psyclone.lfric.LFRicKernelArgument`
```
hiker: That caused me some confusion, but I take it that class was just renamed, but not all comments were updated :) Thanks for fixing this as you go!

hiker (commented):
Looking good so far. Will trigger CI and finish tomorrow.

src/psyclone/lfric.py
```diff
@@ -5023,12 +5041,20 @@ def _compute_from_field(self, field):
             not (field.access == AccessType.INC
                  and loop.upper_bound_name in ["cell_halo",
                                                "colour_halo"]))
+
+        # Records whether the kernel is a special "halo" kernel that must
+        # iterate into the halo for correctness.
+        is_halo_kernel = False
```
hiker: Can't you initialise the variable here with
```
is_halo_kernel = (call.iterates_over in
                                  const.HALO_KERNEL_ITERATION_SPACES)
```
? I tried, and it passed all lfric tests. I think that would be easier, instead of having this buried two if-statements down somwhere?

src/psyclone/lfric.py
```diff
@@ -4824,7 +4827,7 @@ class HaloWriteAccess(HaloDepth):
     particular loop nest.
 
     :param field: the field that we are concerned with.
-    :type field: :py:class:`psyclone.lfric.LFRicArgument`
+    :type field: :py:class:`psyclone.lfric.LFRicKernelArgument`
```
hiker: Could you just add this using Python typing to the actual parameter below?

src/psyclone/lfric.py
```diff
@@ -4987,16 +4989,32 @@ def stencil_type(self):
...
 
+        For a standard kernel that has been transformed to perform redundant
+        computation and has a field argument with a stencil access, the depth
+        of that field's halo that is read is computed as the *sum* of the
+        depth of the redundant computation and the stencil size. However, for
+        the special case of a Kernel which *must* iterate into the halo region
+        (has an ``OPERATES_ON`` of ``HALO_CELL_COLUMN`` or
+        ``OWNED_AND_HALO_CELL_COLUMN``) to a depth ``ndepth``, say, then the
+        depth of the halo that is read is currently computed as the *maximum*
+        of ``ndepth`` and the stencil size. TODO #2781 - this case should be
```
hiker: Could you add a sentence to explain why? E.g.:
`These kernels will only access the stencil 'parallel' to the halos, i.e. not actually extend the required depth`
It is strange that we don't really have documentation where this could easily be added (as far as I can see). Given that this is a band-aid solution, I am happy to wait for the proper introduction of a new stencil type.

src/psyclone/lfric.py
arporter: Yes, I was confused too but you're right - there were just some outdated comments.

src/psyclone/lfric.py
arporter: Oh yes, that's much nicer - thanks.

src/psyclone/lfric.py
arporter: Oh yes, done. I guess that's another advantage of TypeHints - you can't have typos in them?

src/psyclone/lfric.py
arporter: Good point. I thought I'd covered it but I hadn't. Yes, I too looked through the docs for a place to put it but couldn't find anything.

src/psyclone/lfric.py
hiker: Good point :)

src/psyclone/lfric.py
hiker: Thanks!

hiker:
No worries, I noticed that the CI was busy, so thought I'd better leave the machine for you, since I can run it easily at your night time :)

PR #3095, opened 2025-08-15
(Closes #2916) import pure routine fix

### LINKED ISSUE #2916: [PSyIR] 'pure' routine attribute doesn't seem to be given to associated RoutineSymbol.
Test `psyir/transformations/omp_task_transformations_test.py::test_omptask_apply_kern` has a code fragment containing the declaration of a 'pure' subroutine but then has to manually set the `is_pure` attribute of the associated RoutineSymbol. This should not be necessary.

arporter:
Adds a `RoutineSymbol` specialisation of `copy_properties()`. Also extends the method so that the caller can request that the symbol interface be left alone (as sometimes we're just updating the properties of a local Symbol with those from an imported one).

Updates `Call.get_callees()` so that the properties of `Call.routine.symbol` are updated with those from the target Routine(s).


src/psyclone/psyir/symbols/data_type_symbol.py
```diff
@@ -110,20 +112,24 @@ def datatype(self, value):
...
         self._datatype = value
 
-    def copy_properties(self, symbol_in):
+    def copy_properties(self,
+                        symbol_in: DataTypeSymbol,
+                        exclude_interface: bool = False):
         '''Replace all properties in this object with the properties from
-        symbol_in, apart from the name (which is immutable) and visibility.
+        symbol_in, apart from the name (which is immutable), visibility
+        and (optionally) the interface.
```
sergisiso: I am confused about why we did change the Interface but keep the Visibility. I have the feeling these probably need to change together. However, I can only see this being used with local symbols where we can ignore the visibility. I don't know if is it worth updating it to "exclude_scoping[_properties]" and do both. Or wait until/if we find the issue.

There is another use at src/psyclone/psyir/symbols/symbol_table.py:1773
```
local_sym.copy_properties(imported_sym)
# Restore the interface and visibility as these are
# local (not imported) properties
local_sym.interface = interface
local_sym.visibility = visibility
```
where we could update to
`local_sym.copy_properties(imported_sym, exclude_interface=True)`
the visibility assignment was redundant

src/psyclone/psyir/nodes/call.py
```diff
@@ -498,7 +498,11 @@ def get_callees(self):
                         if psyir:
                             routines.append(psyir)
                     if routines:
-                        rsym.interface = DefaultModuleInterface()
+                        # Update the properties of the target RoutineSymbol
+                        # but don't change its interface as we haven't moved
+                        # anything.
+                        rsym.copy_properties([rt.symbol for rt in routines],
+                                             exclude_interface=True)
```
sergisiso: Do you know why we did `rsym.interface = DefaultModuleInterface()` and why we don't now?

I was going to ask why rsym is not already rt.symbol if this is are local symbols, but then realise that rt can be part of a list and we may be populating the properties for the interface symbol.

However, I don't think this is right place to do this. It is quite unexpected that to fill interface properties you have to find a call to it and do get_callees. I would do this on src/psyclone/psyir/symbols/generic_interface_symbol.py::routines.setter

src/psyclone/psyir/nodes/call.py
```diff
@@ -612,7 +616,11 @@ def get_callees(self):
                 if psyir:
                     routines.append(psyir)
             if all_names and len(routines) == len(all_names):
-                # We've resolved everything.
+                # We've resolved everything. Update the properties of the
+                # target RoutineSymbol but don't change its interface as we
+                # haven't moved anything.
+                self.routine.symbol.copy_properties(
+                    [rt.symbol for rt in routines], exclude_interface=True)
```
sergisiso: I has going to ask why is this not done at this point when the symbol was resolved, but it may be the same case as before that is just populating the interface symbol?

src/psyclone/psyir/nodes/call.py
sergisiso: The routine.setter or the current solution won't be resilient to changes. e.g. if we change the purity of a routine the interface will keep a outdated-wrong value. I would prefer that GenericInterfaceSymbol is_pure and is_elemental are calculated dynamically by looking at its self._routine.is_pure/is_elemental values.

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -188,14 +189,27 @@ def copy(self):
                           visibility=self.visibility,
                           interface=self.interface.copy())
 
-    def copy_properties(self, symbol_in: RoutineSymbol):
+    def copy_properties(self,
+                        symbol_in: Union[RoutineSymbol, list[RoutineSymbol]],
```
sergisiso: We the suggestion above regarding GenericSymbolInterface calculating properties dynamically we may not need to accept list of symbols anymore (since it was only done to copy from routines to interfaces). In that case I would revert this and keep the method simple.

src/psyclone/psyir/symbols/routinesymbol.py
```diff
@@ -170,6 +173,45 @@ def copy(self):
                           is_pure=self.is_pure,
                           is_elemental=self.is_elemental)
 
+    def copy_properties(self,
+                        symbol_in: Union[RoutineSymbol, list[RoutineSymbol]],
```
sergisiso: We the suggestion above regarding GenericSymbolInterface calculating properties dynamically we may not need to accept list of symbols anymore (since it was only done to copy from routines to interfaces). In that case I would revert this and keep the method simple.

src/psyclone/psyir/nodes/call.py
arporter: Good idea - I've implemented this now.

src/psyclone/psyir/nodes/call.py
arporter: This has gone now.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: Yes, reverted.

src/psyclone/psyir/symbols/routinesymbol.py
arporter: Yes, simplified.

src/psyclone/psyir/symbols/data_type_symbol.py
arporter: Updated. Despite what was done here, `copy_properties` never updates visibility so I think we're OK.

src/psyclone/psyir/nodes/call.py
arporter: I don't know why we were setting rsy.interface to DefaultModuleInterface() - probably it was because we were often about to inline it into a module. Obviously, since this is just getting the Routines, we shouldn't be doing this - it should only happen when we do the inlining.

src/psyclone/psyir/nodes/call.py
```diff
@@ -464,14 +463,15 @@ def copy(self):
...
 
-    def get_callees(self):
+    def get_callees(self) -> List[Routine]:
         '''
         Searches for the implementation(s) of all potential target routines
         for this Call without resolving static polymorphism by checking the
-        argument types.
+        argument types. Updates the properties of the RoutineSymbol that is
+        the target of this Call with those associated with the RoutineSymbols
+        of the target routines.
```
sergisiso: Should this be reverted now? The PSyIR tree is not updated anymore (I think)

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -142,6 +159,44 @@ def routines(self, values):
...
 
+    @property
+    def is_pure(self) -> Union[bool, None]:
+        '''
+        :returns: whether the routine represented by this Symbol has no side
+            effects (guarantees that the routine always returns the same
+            result for a given set of inputs).
+        '''
+        # If one Routine in an Interface is pure then they all must be.
+        return self._routines[0].symbol.is_pure
```
sergisiso: _routines can be an empty list, should we return None instead of produce the current KeyError?

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -142,6 +159,44 @@ def routines(self, values):
...
+
+    @is_pure.setter
+    def is_pure(self, _):
+        '''
+        This property is computed dynamically from the RoutineSymbols that
+        this GenericInterfaceSymbol maps to and therefore cannot be set.
+
+        This method exists for compatibility with the RoutineSymbol superclass.
+        '''
```
sergisiso, line 181: Should we raise some type of error here?

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -142,6 +159,44 @@ def routines(self, values):
...
+
+    @property
+    def is_elemental(self) -> Union[bool, None]:
+        '''
+        :returns: whether the routine represented by this Symbol is elemental
+            (acts element-by-element on supplied array arguments) or None if
+            this is not known.
+        '''
+        # If one Routine in an Interface is elemental then they all must be.
+        return self._routines[0].symbol.is_elemental
```
sergisiso: Same. A empty _routines should probably return None

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -142,6 +159,44 @@ def routines(self, values):
...
+
+    @is_elemental.setter
+    def is_elemental(self, _):
+        '''
+        This property is computed dynamically from the RoutineSymbols that
+        this GenericInterfaceSymbol maps to and therefore cannot be set.
+
+        This method exists for compatibility with the RoutineSymbol superclass.
```
sergisiso, line 201: Same. Should we raise and error?

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -184,18 +239,42 @@ def copy(self):
...
 
-    def copy_properties(self, symbol_in: RoutineSymbol):
+    @property
+    def datatype(self) -> Union[DataType, DataTypeSymbol]:
+        '''
+        :returns: the datatype of this symbol if it can be determined.
+        '''
+        dtypes = set([str(rinfo.symbol.datatype) for rinfo in self._routines])
+        if len(dtypes) == 1:
+            return self._routines[0].symbol.datatype
```
sergisiso: Why converting it to str and then searching it again? Would
```
dtypes = set(rinfo.symbol.datatype in ...)
if len(dtypes) == 1
   return next(iter(dtypes))
```
work?

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -184,18 +239,42 @@ def copy(self):
...
+        return UnresolvedType()
+
+    @datatype.setter
+    def datatype(self, _):
+        '''
+        This property is computed dynamically from the RoutineSymbols that
+        this GenericInterfaceSymbol maps to and therefore cannot be set.
+
+        This method exists for compatibility with the RoutineSymbol superclass.
+        '''
```
sergisiso, line 268: Raise issue?

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -328,20 +329,24 @@ def copy(self):
         copy.inline_comment = self.inline_comment
         return copy
 
-    def copy_properties(self, symbol_in):
+    def copy_properties(self,
+                        symbol_in: DataSymbol,
+                        exclude_interface: bool = False):
         '''Replace all properties in this object with the properties from
         symbol_in, apart from the name (which is immutable) and visibility.
+        If `exclude_interface` is True, the interface is also not updated.
```
sergisiso, line 337: Note that a couple of "def copy_properties" have different docstring. It would be good to make them all consistent. I think this one is quite clear.

src/psyclone/psyir/nodes/call.py
arporter: Oh yes, done.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: Oh yes, good spot.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: We can't raise an error because this method *does* get called (because of the `super().__init__()` and `super()._process_arguments()` calls. I think the only way to avoid this would be to alter the class hierarchy and I'm not sure how we'd do that while keeping GenericInterfaceSymbol a sub-class of RoutineSymbol.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: Done.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: Same reasoning.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: I converted to `str` to get objects that were hashable - since we've re-defined the `__eq__` method in DataType it (and its subclasses) are not hashable. I'm not sure of the value of that `__eq__` method:
```python
    def __eq__(self, other):
        '''
        :param Any other: the object to check equality to.

        :returns: whether this type is equal to the 'other' type.
        :rtype: bool
        '''
        return type(other) is type(self)
```
but I'd rather not mess with it here. I've added a comment for now.

src/psyclone/psyir/symbols/generic_interface_symbol.py
arporter: Again we can't, which does make me slightly uncomfortable. I'm open to suggestions.

src/psyclone/psyir/symbols/datasymbol.py
arporter: Done.

arporter:

sergisiso (approved):
@arporter I have some doubts about the empty setters and the str comparison of the datatype. But since this already fixes the bug I will approve this PR and we can care about these later. I will run the integration test one more time just in case.

PR #3100, opened 2025-08-18
(Closes #3097) Disable inheriting options keyword docstring when not appropriate.

### LINKED ISSUE #3097: Generated Transformation documentation includes inherited 'options' argument
For a brand-new transformation that does not support `options` (since we're moving to kwargs), the generated documentation still ends up having 'options' listed as a valid argument:
Image
This is because, it is inherited from the `Transformation` base class.

It would be better if we blocked the inheritance of the 'options' argument so that it is only shown if a Transformation does still accept it. However, this may cause problems for e.g. the OMP transformations which still inherit options from base classes I think? If that's the case, this issue may have to be blocked until we've updated those transformations to use kwargs.

LonelyCat124:
I also checked the `validate_options` command, which has:
```
            if option == "options":
                continue
```

Additionally, the options parameter isn't inherited when generating `get_valid_options`:
```
if k == "options":
    continue 
```

I'm not sure if we want to have the `validate_options` skip over the options therefore? Since `options` being supplied as a `kwargs` we're disallowing? If so I will remove the continue from validate_options if the reviewer thinks it is worthwhile.

Ready for a first look @arporter @sergisiso

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,6 +1,11 @@
+   79) PR #3099 for #3089. Remove is_array from Nodes and rename it to
+   has_indices() in Signature.
+
+   78) PR #3106 for #3105. Add print handlers for Sympy boolean literals.
+
    77) PR #3089 for #2781. Refine halo-depth logic for LFRic "halo" kernels.
 
-   76 PR #2910 for #2896. Adds IncreaseRankLoopArraysTrans.
+   76) PR #2910 for #2896. Adds IncreaseRankLoopArraysTrans.
 
    75) PR #3093 for #3092. Fixes Fortran backend declarations of (1:) arrays.
 
--- a/src/psyclone/core/access_sequence.py
+++ b/src/psyclone/core/access_sequence.py
@@ -125,15 +125,11 @@ class AccessInfo():
                                 f"ComponentIndices, got '{component_indices}'")
         self._component_indices = component_indices
 
-    def is_array(self):
-        '''Test if any of the components has an index. E.g. an access like
-        a(i)%b would still be considered an array.
-
-        :returns: if any of the variable components uses an index, i.e.\
-            the variable is an array.
-        :rtype: bool
+    def has_indices(self) -> bool:
         '''
-        return self._component_indices.is_array()
+        :returns: whether any of the access components uses an index.
+        '''
+        return self._component_indices.has_indices()
 
     @property
     def access_type(self):
@@ -348,34 +344,31 @@ class AccessSequence(list):
                 f" it does not have a 'READ' access.")
         read_access.change_read_to_write()
 
-    def is_array(self, index_variable=None):
-        '''Checks if the variable is used as an array, i.e. if it has
-        an index expression. If the optional `index_variable` is specified,
-        this variable must be used in (at least one) index access in order
-        for this variable to be considered as an array.
+    def has_indices(self, index_variable: str = None) -> bool:
+        ''' Checks whether this variable accesses has any index. If the
+        optional `index_variable` is provided, only indices involving the given
+        variable are considered.
 
-        :param str index_variable: only considers this variable to be used \
-            as array if there is at least one access using this \
-            index_variable.
+        :param index_variable: only consider index expressions that involve
+            this variable.
 
-        :returns: true if there is at least one access to this variable \
-            that uses an index.
-        :rtype: bool
+        :returns: true if any of the accesses has an index.
 
         '''
-        is_array = any(access_info.is_array() for access_info in self)
+        has_indices = any(access.has_indices() for access in self)
 
         # If there is no access information using an index, or there is no
         # index variable specified, return the current result:
-        if not is_array or index_variable is None:
-            return is_array
+        if not has_indices or index_variable is None:
+            return has_indices
 
         # Avoid circular import
         # pylint: disable=import-outside-toplevel
         from psyclone.psyir.nodes import Reference
 
+        lowered_name = index_variable.lower()
         for access_info in self:
-            if any(ref.symbol.name == index_variable
+            if any(ref.symbol.name.lower() == lowered_name
                    for ref in access_info.node.walk(Reference)):
                 return True
 
--- a/src/psyclone/core/component_indices.py
+++ b/src/psyclone/core/component_indices.py
@@ -147,14 +147,9 @@ class ComponentIndices():
         return self._component_indices
 
     # ------------------------------------------------------------------------
-    def is_array(self):
-        '''Test whether there is an index used in any component. E.g. an access
-        like `a(i)%b` with indices `[ [i], [] ]` would still be considered an
-        array.
-
-        :returns: whether any of the variable components uses an index, i.e.\
-            the variable is an array.
-        :rtype: bool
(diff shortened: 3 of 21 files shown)


src/psyclone/tests/utils_test.py
```diff
@@ -257,6 +257,85 @@ def apply(self, node, **kwargs):
     instance.apply(None)
 
 
+def test_transformation_doc_wrapper_uniheritable():
```
arporter: "uninheritable"

src/psyclone/tests/utils_test.py
```diff
@@ -257,6 +257,85 @@ def apply(self, node, **kwargs):
...
+            Super apply docstring
+
+            :param opt1: opt1 docstring.
+            :param opt2: opt2 docstring.
+            :type opt2: opt2 type.
+            :param options: options dictionary.
+            :type options: dict
+            '''
+
+    # Create a base transformation class
```
arporter: Please extend to say how this differs from `BaseTrans` (and perhaps why)

src/psyclone/tests/utils_test.py
```diff
@@ -257,6 +257,85 @@ def apply(self, node, **kwargs):
...
+            '''
+
+        def apply(self, node, opt3: int = 1, **kwargs):
+            '''
+            Sub apply docstring
+
+            :param opt3: opt3 docstring.
+            '''
+
+    transformation_documentation_wrapper(InheritingTrans1, inherit=True)
```
arporter, line 318: Please add a comment explaining the expected behaviour that we are testing for.

src/psyclone/tests/utils_test.py
```diff
@@ -257,6 +257,85 @@ def apply(self, node, **kwargs):
...
+    assert ":param options:" not in InheritingTrans1.apply.__doc__
+
+    class InheritingTrans2(BaseTrans, BaseTrans2):
```
arporter, line 322: Comment explaining what this class is testing for please.

src/psyclone/tests/utils_test.py
```diff
@@ -257,6 +257,85 @@ def apply(self, node, **kwargs):
...
+            '''
+
+        def apply(self, node, opt3: int = 1, **kwargs):
+            '''
+            Sub apply docstring
+
+            :param opt3: opt3 docstring.
+            '''
+
+    transformation_documentation_wrapper(
```
arporter, line 338: Comment again please.

src/psyclone/tests/utils_test.py
```diff
@@ -257,6 +257,85 @@ def apply(self, node, **kwargs):
...
+        InheritingTrans2,
+        inherit=[BaseTrans, BaseTrans2]
+    )
+    assert ":param options:" not in InheritingTrans2.apply.__doc__
```
arporter, line 342: I was going to ask about what happens for an old-style Transformation that still accepts `options` but then I realised that, by definition, it won't use `transformation_documentation_wrapper` so it will be unchanged?

arporter:
> I also checked the `validate_options` command, which has:
> 
> ```
>             if option == "options":
>                 continue
> ```
> 
> Additionally, the options parameter isn't inherited when generating `get_valid_options`:
> 
> ```
> if k == "options":
>     continue 
> ```
> 
> I'm not sure if we want to have the `validate_options` skip over the options therefore? Since `options` being supplied as a `kwargs` we're disallowing? If so I will remove the continue from validate_options if the reviewer thinks it is worthwhile.
> 
> Ready for a first look @arporter @sergisiso

I'm not quite sure I follow. Are we saying that if a Transformation has been updated to use `transformation_documentation_wrapper` that we then forbid it to have an argument named "options"? (I think we are.) In that case I think both `get_valid_options` and `validate_options` need to be updated.

src/psyclone/tests/utils_test.py
LonelyCat124: Done

src/psyclone/tests/utils_test.py
LonelyCat124: Added a comment explaining why this exists - the differences aren't important - only the existence to test multiple inheritance.

src/psyclone/tests/utils_test.py
LonelyCat124: Added

src/psyclone/tests/utils_test.py
LonelyCat124: Added

src/psyclone/tests/utils_test.py
LonelyCat124: Added

src/psyclone/tests/utils_test.py
LonelyCat124: Yes, this only affects inheritance of `options` docstring  - it won't remove any `options` docstrings that are attached to the class manually (which has to be done if `options` is defined on a class).

src/psyclone/tests/psyGen_test.py
```diff
@@ -247,6 +247,10 @@ def apply(self, node, valid: bool = True, untyped=False):
     assert options['untyped'].default is False
     assert options['untyped'].type is None
     assert options['untyped'].typename is None
+    # Check that option is allowed if its specified on the transformation.
+    assert options['options'].default == {}
+
+    # FIXME Add options and check its valid from self.
```
arporter: Is this one done now?

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -577,7 +577,8 @@ def test_omploop_trans_new_options(sample_psyir):
     assert ("'OMPLoopTrans' received invalid options ['fakeoption1', "
```
arporter, line 576: Since we're here, pls could you remove the `print` on the previous line.

src/psyclone/psyGen.py
```diff
@@ -2857,7 +2857,7 @@ def get_valid_options(cls) -> Dict[str, ValidOption]:
             signature = inspect.signature(base_cls_apply)
             # Loop over the arguments to the apply call.
             for k, v in signature.parameters.items():
-                if k == "options":
+                if k == "options" and base_cls is not cls:
```
arporter, line 2863: Comment please, e.g. "Since 'options' argument is deprecated, it is not inherited."

arporter:
Integration tests were green :-)

src/psyclone/tests/psyGen_test.py
LonelyCat124: Yeah, its added after InheritTrans here.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: Done

src/psyclone/psyGen.py
LonelyCat124: Added

LonelyCat124:
@arporter Should be ready for a (hopefully) final look now. I guess the one thing I've not double checked yet is that the docs don't show options where they were being incorrectly inherited before...

Issue #2202, opened 2023-06-30
PSyclone reordering symbol table during output resulting in compiler error


LonelyCat124:
When accessing (at least) one file in Socrates PSyclone changes the order of variable declaration, which is resulting in compiler error due to undefined variable.

The file in question is `gas_list_pcf.f90`, at some point there is a variable defined as:
`INTEGER, PRIVATE :: i`

later in the file during the various `reshape` calls, this `i` variable is used.

IN the file output by PSyclone, the outputs are reordered such that some accesses to `i` are used before its definition:
```
&public :: threshold_wavelength = RESHAPE([REAL(KIND = RealK) :: 246.0E-09_RealK, 242.0E-09_RealK, 175.0E-09_RealK, &
&134.0E-09_RealK, 129.0E-09_RealK, (0.0_RealK, i = 1, npd_products - 5), 227.5E-09_RealK, 167.1E-09_RealK, 128.6E-09_RealK, &
&108.2E-09_RealK, 89.922E-09_RealK, 65.026E-09_RealK, 63.693E-09_RealK, 54.655E-09_RealK, (0.0_RealK, i = 1, npd_products - 8), &
..... rest of this definition
... string array definition
integer, private :: i
```

This happens in some other files too, e.g. refract_re_ccf.f90

arporter:
Strictly I think this is still valid Fortran (which compiler is this?) but we do already do the 'right thing' for parameter declarations. Hopefully that approach can be extended to the other declarations.

LonelyCat124:
I'm using mpifort, wrapping gcc-12.2 on glados.

rupertford:
@sergisiso's suggestion to replace `decl x = expr` with `decl x\nx = expr` in PSyIR would also fix this particular case. Determining a valid order will be complicated by the fact that we will have UnknownFortranType's in some cases (including this one) and we will not know their dependencies.

sergisiso:
We assumed that any parameter (compile-time) value could only depend on other parameter values (compile-time) .

But here it uses a runtime (i) variable (albeit with a runtime given value) to set-up a compile-time value(threshold_wavelength) ?
Or is the 'i' inside the reshape something specific to shape and not related to the outside 'i'? I am not familiar with RESHAPE

sergisiso:
I guess its related because you mention:
> If you move the declaration of i to before threshold_wavelength as the original input does then the output will compile.

arporter:
[...]

PR #3101, opened 2025-08-19
#1312 ScalarArray code generation

mo-alistairp:
Adding code generation for ScalarArrays (towards #1312 ). The metadata support was added in #2173.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   31) PR #3101 for #1312. Implements support for ScalarArray arguments to
+   LFRic kernels.
+ 
    30) PR #3280 towards #2971. Move OMPParallelTrans to psyir.transformations. 
 
    29) PR #3275 for #3216. Allows the specification of private variables


src/psyclone/domain/lfric/kern_call_acc_arg_list.py
```diff
@@ -272,6 +272,20 @@ def scalar(self, scalar_arg, var_accesses=None):
 
         '''
 
+    def scalar_array(self, scalar_arr_arg, var_accesses=None):
+        '''
+        Override the default implementation as there's no need to specify
+        ScalarArrays for an OpenACC data region.
```
arporter: This isn't true I'm afraid - they will need to be added to a data region. However, you could leave this as a TODO as I'm not sure we're ever going to need data regions.

mo-alistairp: Thanks for that - it's currently a very rough draft so some of the comments are currently just borrowed from other functions. But, I will definitely note that down!

arporter:
> I'm unsure how to add the declarations correctly. The ScalarArray needs to have `dimension(dims_array(1), dims_array(2), ... )`. I've tried creating each of `dims_array(n)` as individual symbols which get added to the ScalarArray symbols as a reference. However, this doesn't seem to be added to the dimension correctly and adds the `dims_array(n)` symbols to the reference list.

You'll need to create a Symbol for `dims_array` as an integer array that is an argument to the (kernel) subroutine. Then you create an ArrayType for the ScalarArray with `dims = [ArrayReference.create(dims_array_sim, [Literal(1)]), ArrayReference.create(dims_array_sym, [Literal(2)])]`.

mo-alistairp:
I think this is now ready for review. The code gen and stub gen seem to be performing as I would expect. I've added test coverage for everything I've added. Everything is passing the style checks.

src/psyclone/domain/lfric/kern_call_acc_arg_list.py
```diff
@@ -271,6 +272,9 @@ def scalar(self, scalar_arg, var_accesses=None):
             :py:class:`psyclone.core.VariablesAccessMap`]
 
         '''
+        # TODO: Add implementation of OpenACC data region for ScalarArrays
```
arporter: I suggest adding a check here and if we have a ScalarArray, raise a NotImplementedError. You'll need an associated test too I'm afraid.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -254,7 +254,6 @@ def scalar(self, scalar_arg,
         else:
             sym = self._symtab.lookup(scalar_arg.name)
             self.psyir_append(Reference(sym))
-
```
arporter, line 257: Please revert this file.

src/psyclone/domain/lfric/kern_stub_arg_list.py
```diff
@@ -133,6 +133,35 @@ def cma_operator(self, arg, var_accesses=None):
...
         self.extend(_local_args, var_accesses)
 
+    def scalar(self, scalar_arg, var_accesses=None):
+        '''Add the name associated with the scalar argument to the argument
+        list and optionally add this scalar to the variable access
+        information.
+
+        :param scalar_arg: the kernel argument.
+        :type scalar_arg: :py:class:`psyclone.lfric.LFRicKernelArgument`
+        :param var_accesses: optional VariablesAccessMap instance that \
```
arporter: No need for line-continuation chars ("\") pls.

src/psyclone/domain/lfric/kern_stub_arg_list.py
```diff
@@ -133,6 +133,35 @@ def cma_operator(self, arg, var_accesses=None):
         _local_args += [bandwidth, alpha, beta, gamma_m, gamma_p]
         self.extend(_local_args, var_accesses)
 
+    def scalar(self, scalar_arg, var_accesses=None):
+        '''Add the name associated with the scalar argument to the argument
+        list and optionally add this scalar to the variable access
+        information.
+
+        :param scalar_arg: the kernel argument.
+        :type scalar_arg: :py:class:`psyclone.lfric.LFRicKernelArgument`
```
arporter: We're moving (incrementally) to using type hints where the type is specified in the argument list rather than in the docstring. e.g.:
```python
def scalar(self,
           scalar_arg: LFRicKernelArgument,
           var_accesses: Optiona[VariablesAccessMap] = None):
```
you'll probably need to sort out appropriate imports at the top of the file.

src/psyclone/domain/lfric/kern_stub_arg_list.py
```diff
@@ -133,6 +133,35 @@ def cma_operator(self, arg, var_accesses=None):
...
+            raise InternalError(
+                f"Expected argument type to be one of "
+                f"{const.VALID_SCALAR_NAMES + const.VALID_ARRAY_NAMES}"
+                f" but got '{scalar_arg.argument_type}'")
+
+        if scalar_arg.is_scalar:
+            self.append(scalar_arg.name, var_accesses)
+        else:
+            # ScalarArray
+            self.append("dims_" + scalar_arg.name, var_accesses)
```
arporter, line 163: I think we also need an "array_ndims" to pass how many dimensions the array has?
However, I've looked at the docs and they don't say that so perhaps I'm missing something. Hopefully this will become clear.

arporter: I've solved it by looking at the test cases (thanks) - the kernel metadata specifies the number of dimensions so it isn't required here.

src/psyclone/domain/lfric/lfric_scalar_args.py
```diff
@@ -111,7 +111,8 @@ def invoke_declarations(self):
             intrinsic_type=const.MAPPING_DATA_TYPES["gh_logical"])
 
         for intent in FORTRAN_INTENT_NAMES:
-            scal = [arg.declaration_name for arg in self._scalar_args[intent]]
+            scal = [arg.declaration_name for
```
arporter: Please revert this file.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+
+from psyclone.psyir.frontend.fparser2 import INTENT_MAPPING
+from psyclone.domain.lfric import LFRicCollection, LFRicConstants, LFRicTypes
+from psyclone.errors import GenerationError, InternalError
+from psyclone.psyGen import FORTRAN_INTENT_NAMES
+from psyclone.psyir.nodes import Literal, ArrayReference
+from psyclone.psyir.symbols import (DataSymbol, ArrayType,
+                                    INTEGER_TYPE, ArgumentInterface)
+
+# pylint: disable=too-many-lines
```
arporter: Please remove these pylint-disables. There are definitely not too many lines so maybe the others aren't needed either?

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+# pylint: disable=too-many-locals
+# pylint: disable=too-many-branches
+
+
+class LFRicScalarArrayArgs(LFRicCollection):
+    '''
+    Handles the declarations of ScalarArray kernel arguments appearing in
+    either an Invoke or a Kernel stub.
+
+    :param node: the Invoke or Kernel stub for which to manage the \
```
arporter, line 63: No line-continuation chars please.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+                 ScalarArray arguments.
+    :type node: :py:class:`psyclone.domain.lfric.LFRicKern` or \
```
arporter: Pls update to use typehinting.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+                :py:class:`psyclone.domain.lfric.LFRicInvoke`
+
+    '''
+    def __init__(self, node):
+        super().__init__(node)
+
+        # Initialise dictionaries of 'real', 'integer' and 'logical'
+        # ScalarArray arguments by data type and intent
```
arporter: Please extend this comment to say what "_scalar_array_args" holds as opposed to the "real", "integer"... ones. Please also say what the keys in the dictionary will be. In fact, it might be nicer just to have one dictionary and store the argument objects in it. In fact, could it be a list and you just filter it when required?

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+            self._scalar_array_args[intent] = []
+            self._real_scalar_arrays[intent] = []
+            self._integer_scalar_arrays[intent] = []
+            self._logical_scalar_arrays[intent] = []
+
+    def invoke_declarations(self):
+        '''
+        Create argument lists and declarations for all ScalarArray arguments
+        in an Invoke.
+
```
arporter, line 79: Just one blank line please.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+
+        :raises InternalError: for unsupported argument intrinsic types.
+        :raises GenerationError: if the same ScalarArray argument has \
```
arporter: No line-continuation chars please.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+        self._create_declarations()
+
+    def _create_declarations(self):
+        '''
+        Add declarations for the ScalarArray arguments.
+
+        '''
+        # Real ScalarArray arguments
+        for intent in FORTRAN_INTENT_NAMES:
+            if self._real_scalar_arrays[intent]:
```
arporter: I think the constructor always sets this to be an empty list so you don't need this `if`?

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+                                arg.name,
+                                symbol_type=DataSymbol,
+                                datatype=ArrayType(
+                                    LFRicTypes("LFRicRealScalarDataType")(),
+                                    sym_list))
+                        array_symbol.interface = ArgumentInterface(
+                                            INTENT_MAPPING[intent])
+                        self.symtab.append_argument(array_symbol)
+
+        # Integer ScalarArray arguments
```
arporter: You almost certainly don't need to duplicate this code for the integer and logical types - you should be able to do it all in one loop and just lookup the type for each argument.

arporter: (The order that they get added to the SymbolTable doesn't matter and doesn't affect how the resulting code is generated.)

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+                        f"ScalarArray argument '{arg.declaration_name}'"
+                        f". Supported types are "
+                        f"{const.VALID_SCALAR_DATA_TYPES}.")
+
+        # Create declarations
+        self._create_declarations()
+
+    def _create_declarations(self):
+        '''
+        Add declarations for the ScalarArray arguments.
```
arporter: s/Add declarations/Create the Symbols/

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,313 @@
...
+                        if not self._kernel:
+                            # For code generation, find ScalarArray tag
+                            # and convert it to an ArrayType
+                            array_symbol = self.symtab.lookup_with_tag(
+                                "AlgArgs_" + arg.name)
+                            array_symbol.datatype = ArrayType(
+                                LFRicTypes("LFRicRealScalarDataType")(),
+                                sym_list)
+                        else:
+                            # For stub generation, create the symbol
```
arporter: If you're having to specialise for stub generation then I suggest you move this code into the `stub_declarations` method.

src/psyclone/domain/lfric/lfric_types.py
```diff
@@ -190,6 +190,7 @@ def _create_generic_scalars():
                           LFRicTypes("R_DEF")),
             GenericScalar("LFRicLogicalScalar", ScalarType.Intrinsic.BOOLEAN,
                           LFRicTypes("L_DEF"))]
+        # TODO: Consider adding ScalarArrays as a generic data type
```
arporter: Presumably we don't need this?

src/psyclone/lfric.py
```diff
@@ -1480,7 +1481,7 @@ def initialise(self, cursor: int) -> int:
         init_cursor = cursor
         for arg in self._invoke.psy_unique_vars:
             # We don't have proxies for scalars
```
arporter: "...for scalars or arrays of scalars."

src/psyclone/lfric.py
```diff
@@ -5681,6 +5682,7 @@ def __init__(self, kernel_args, arg_meta_data, arg_info, call, check=True):
         # any-space function spaces.
         self._kernel_args = kernel_args
         self._vector_size = arg_meta_data.vector_size
```
arporter, line 5677: Comment please: "the number of dimensions (for a ScalarArray)"

src/psyclone/lfric.py
```diff
@@ -6152,6 +6154,16 @@ def is_scalar(self):
         const = LFRicConstants()
         return self._argument_type in const.VALID_SCALAR_NAMES
 
+    @property
+    def is_scalar_array(self):
+        '''
+        :returns: True if this kernel argument represents a \
```
arporter: No line continuation char and please use a type hint for the return type.

src/psyclone/lfric.py
```diff
@@ -6545,7 +6557,10 @@ def _find_or_create_type(mod_name: str,
                     kind_name, INTEGER_TYPE,
                     interface=ImportInterface(constants_container))
                 symtab.add(kind_symbol)
-            return ScalarType(prim_type, Reference(kind_symbol))
+            dts = ScalarType(prim_type, Reference(kind_symbol))
+            if self.is_scalar_array and self._array_ndims > 1:
```
arporter: What happens if it's a ScalarArry with ndims==1? Please add a comment.

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
```diff
@@ -603,3 +603,91 @@ def test_three_scalars(tmpdir):
...
     assert expected in generated_code
     assert LFRicBuild(tmpdir).code_compiles(psy)
+
+
+def test_scalar_array(tmpdir):
+    ''' Tests that we generate correct code when a kernel has all three
+    types of valid scalar array argument: 'real', 'integer' and 'logical'.
+
+    '''
+    _, invoke_info = parse(os.path.join(BASE_PATH,
```
arporter: I recommend using `get_invoke(...)` from `test_utilities` here as it's a bit neater.

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
```diff
@@ -603,3 +603,91 @@ def test_three_scalars(tmpdir):
...
+        "    loop0_start = 1\n"
+        "    loop0_stop = mesh%get_last_halo_cell(1)\n"
+        "\n"
+        "    ! Call kernels and communication routines\n"
+        "    if (afield_proxy%is_dirty(depth=1)) then\n"
+        "      call afield_proxy%halo_exchange(depth=1)\n"
+        "    end if\n"
+        "    do cell = loop0_start, loop0_stop, 1\n"
+        "      call testkern_scalar_array_code(nlayers_afield, "
+        "afield_data, real_array, logical_array, integer_array, a_scalar, "
```
arporter: This doesn't match the arguments that the kernel subroutine expects - i.e. there aren't any arrays giving the extents of the ScalarArrays. I recommend you try testing with compilation locally (`pytest --compile lfric_scalar_codegen_test.py`). By default it will use the gfortran compiler but you can use `--f90` to specify a different one.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: I've condensed this down to one dictionary where I can, and removed the repetition later in the file when making the symbols. 

The Errors in invoke_declarations needs to be able to differentiate between all arguments and all valid arguments. I couldn't see a tidy way of getting rid of the specific arrays for intrinsic types. But I believe what I have done tidies it, and reduces the amount of memory being used whilst still functioning enough to check for errors. 

I haven't changed `_scalar_array_args` from a dictionary to a list at the moment, because `self._invoke.unique_declns_by_intent` returns a dictionary. It feels like an unnecessary complication to convert that to a list. But, if you have strong feelings the other way I'm happy to reconsider.

I've also added a couple of comments to hopefully clarify what dictionaries are and the difference between `scal` and `decl_scal`, for example.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Done

src/psyclone/lfric.py
mo-alistairp: Done

src/psyclone/domain/lfric/kern_call_acc_arg_list.py
mo-alistairp: Done

arporter: Please remove the TODO as otherwise our rule is that you need an associated Issue and I don't think we want another Issue :-)

arporter: Also, please document the NotImplementedError as a ":raises ...." in the docstring.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -60,66 +60,58 @@ class LFRicScalarArrayArgs(LFRicCollection):
     Handles the declarations of ScalarArray kernel arguments appearing in
     either an Invoke or a Kernel stub.
 
-    :param node: the Invoke or Kernel stub for which to manage the \
```
arporter: Please put back the description of the `node` argument.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
arporter: Thanks, please update the comment.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -181,93 +168,16 @@ def stub_declarations(self):
...
-                        array_symbol.interface = ArgumentInterface(
-                                            INTENT_MAPPING[intent])
-                        self.symtab.append_argument(array_symbol)
-
-        # Logical ScalarArray arguments
-        for intent in FORTRAN_INTENT_NAMES:
-            if self._logical_scalar_arrays[intent]:
-                for arg in self._logical_scalar_arrays[intent]:
+            for arg in self._scalar_array_args[intent]:
+                if arg.intent in FORTRAN_INTENT_NAMES:
```
arporter: Do you need this `if`? If so, please add a comment.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -279,28 +189,30 @@ def _create_declarations(self):
...
                             array_symbol = self.symtab.lookup_with_tag(
                                 "AlgArgs_" + arg.name)
                             array_symbol.datatype = ArrayType(
-                                LFRicTypes("LFRicLogicalScalarDataType")(),
+                                type_map[arg.intrinsic_type],
                                 sym_list)
+                            # Replace the symbol with itself to ensure
+                            # the ScalarArray is generated after the
+                            # dimensions array to avoid compilation errors
+                            self.symtab.swap(array_symbol, array_symbol)
```
arporter: Please add a TODO for #2202 here.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
arporter: This is still outstanding?

src/psyclone/tests/test_files/lfric/28.scalar_array_invoke.f90
```diff
@@ -42,11 +42,11 @@ program scalar_array_invoke
```
arporter, line 44: Please update the comment in this file to give a bit more detail, i.e. that it exercises the ScalarArray functionality.

external/lfric_infrastructure/src/kernel_metadata/argument_mod.f90
```diff
@@ -60,6 +60,7 @@ module argument_mod
   integer, public, parameter :: GH_FIELD               = 507
   integer, public, parameter :: GH_OPERATOR            = 735
   integer, public, parameter :: GH_COLUMNWISE_OPERATOR = 841
+  integer, public, parameter :: GH_SCALAR_ARRAY        = 973
```
arporter, line 63: We're going to need to add this to https://github.com/stfc/lfric_core and/or https://github.com/MetOffice/lfric_core.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -252,6 +252,11 @@ def scalar(self, scalar_arg,
         if scalar_arg.is_literal:
             self.psyir_append(scalar_arg.psyir_expression())
         else:
+            if scalar_arg.is_scalar_array:
+                # If it's a ScalarArray we need to add the dimensions
+                # array to the call
+                dims_sym = self._symtab.lookup("dims_" + scalar_arg.name)
```
arporter: Probably safer to use "dims_"+scalar_arg.name as a *tag* and then do `lookup_with_tag` here.

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
```diff
@@ -603,3 +604,89 @@ def test_three_scalars(tmpdir):
...
+def test_scalar_array(tmpdir):
+    ''' Tests that we generate correct code when a kernel has all three
+    types of valid scalar array argument: 'real', 'integer' and 'logical'.
+
+    '''
+    psy, invoke = get_invoke("28.scalar_array_invoke.f90", TEST_API, idx=0)
+
+    generated_code = str(psy.gen)
+    expected = (
+        "module scalar_array_invoke_psy\n"
```
arporter: To save pain in the future, please could you chop-down the amount of generated code you compare against. That way, this test will be robust to other, unrelated changes to PSy-layer generation.
i.e. just test the subroutine arg list, the relevant declarations and the kernel call itself.

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
```diff
@@ -603,3 +604,89 @@ def test_three_scalars(tmpdir):
         " map_w3(:,cell))\n")
     assert expected in generated_code
     assert LFRicBuild(tmpdir).code_compiles(psy)
+
+
+def test_scalar_array(tmpdir):
```
arporter: Please also use the `dist_mem` fixture and add ``dist_mem=dist_mem`` to the arguments to `get_invoke`. That way this test is exercised both with and without distributed-memory support.

src/psyclone/tests/test_files/lfric/28.scalar_array_invoke.f90
```diff
@@ -0,0 +1,55 @@
...
+  implicit none
+
+  type(field_type)                            :: afield
+  real(kind=r_def),    dimension(50, 100)     :: real_array
+  logical(kind=l_def), dimension(10)          :: logical_array
+  integer(kind=i_def), dimension(2, 5, 10, 8) :: integer_array
+  integer(kind=i_def)                         :: a_scalar
+
+  call invoke(                                                                            &
+       testkern_scalar_array_type(afield,real_array,logical_array,integer_array,a_scalar) &
```
arporter: Probably worth making this more complicated by including a second kernel call (to one of the existing test kernels) in the invoke. If you could make one of the variable names such that you would get a clash with the `dims_real_field` name that PSyclone generates then that would be great. (e.g. you could re-name "a_scalar" to "dims_real_field")

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,225 @@
...
+            # Check that the same ScalarArray name is not found in either of
+            # 'real', 'integer' or 'logical' ScalarArray lists (for instance
+            # if passed to one kernel as a 'real' and to another kernel as an
+            # 'logical' ScalarArray)
+            scal_multi_type = [item for item, count in
+                               Counter(decl_scal).items() if count > 1]
+            if scal_multi_type:
+                raise GenerationError(
+                    f"ScalarArray argument(s) {scal_multi_type} in Invoke "
+                    f"'{self._invoke.name}' have different metadata for data "
```
arporter: perhaps replace everything after and including "have" with "is/are passed to more than one kernel and the kernel metadata for the corresponding arguments specifies different intrinsic types."? (I really struggled to understand this message at first.)

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,225 @@
...
+        for arg in self.kernel_calls[0].arguments.args:
+            if arg.is_scalar_array:
+                self._scalar_array_args[arg.intent].append(arg)
+
+        const = LFRicConstants()
+        # Filter ScalarArray arguments by intent and data type
+        for intent in FORTRAN_INTENT_NAMES:
+            for arg in self._scalar_array_args[intent]:
+                # Distinguish whether they are ScalarArrays
+                if (arg.descriptor.data_type not in
```
arporter, line 159: I think it would be simpler to put this check at L150 - i.e. before we add `arg` into the _scalar_array_arg dictionary. That would save us having to loop through again to check what we've just added.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -0,0 +1,225 @@
...
+                        const.VALID_SCALAR_DATA_TYPES):
+                    raise InternalError(
+                        f"Found an unsupported data type "
+                        f"'{arg.descriptor.data_type}' for the "
+                        f"ScalarArray argument '{arg.declaration_name}'"
+                        f". Supported types are "
+                        f"{const.VALID_SCALAR_DATA_TYPES}.")
+
+        # Create declarations
+        self._create_declarations()
```
arporter: Regarding my comment on the special handling for kernel stubs in _create_declarations, I think probably you just get rid of this call and inline the relevant part of _create_declarations here.

src/psyclone/tests/domain/lfric/arg_ordering_test.py
```diff
@@ -214,7 +214,8 @@ def test_kernel_stub_invalid_scalar_argument():
     with pytest.raises(InternalError) as excinfo:
         create_arg_list.scalar(arg)
     const = LFRicConstants()
-    assert (f"Expected argument type to be one of {const.VALID_SCALAR_NAMES} "
+    assert (f"Expected argument type to be one of "
```
arporter, line 217: You can probably extend this test to get coverage of that missed line. You'll neeed to create a `create_arg_list = KernCallArgList(kernel)` and then call `scalar` on it with an argument that isn't a scalar. If this works, please update the name and docstring of the test.

external/lfric_infrastructure/src/kernel_metadata/argument_mod.f90
mo-alistairp: I have a draft PR ready in `MetOffice/lfric_core` - https://github.com/MetOffice/lfric_core/pull/223. It'll need to go through review, but it's currently waiting for the PSyclone PR to be implemented.

I'm not entirely sure the process you take to get it into `external/` so let me know if you need me to make a PR in `stfc/lfric_core` too

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Hi @arporter, can I just check what you want here. If I inline the `_create_declarations` function and remove it, it will need to be inlined into both `invoke_declarations` and `stub_declarations`. This will create some code duplication - are you happy with this cost to remove `_create_declarations`?

arporter: Hi Alistair, that's not quite what I meant. You can keep `_create_declarations` but move the stub-specific code out of it. There will be a little bit of duplication but I don't think it's much.

mo-alistairp: Ok sure, thanks for the clarification 😄

src/psyclone/tests/domain/lfric/arg_ordering_test.py
mo-alistairp: I could be misunderstading, but is this not what the test is currently doing? I think that the line is covered but CodeCov is complaining because the test has been diluted by including `scalar_array` in `scalar`

src/psyclone/tests/test_files/lfric/28.scalar_array_invoke.f90
mo-alistairp: Hi @arporter, I've added another invoke to by local branch and that works fine. However, when I change an argument to cause a clash in name I get a generation error, as expected - 

`psyclone.errors.GenerationError: Generation Error: expecting the symbol 'dims_integer_array' to be an array, but found 'Scalar<INTEGER, Reference[name:'i_def']>'.`.

This blocks the code from being generated, so I'm unable to test that the code is being generated as expected. Were you after me writing another test to fail in this case? Because I presume that we definitely do want to test what is generated

arporter: > Hi @arporter, I've added another invoke to by local branch and that works fine. However, when I change an argument to cause a clash in name I get a generation error, as expected -
> 
> `psyclone.errors.GenerationError: Generation Error: expecting the symbol 'dims_integer_array' to be an array, but found 'Scalar<INTEGER, Reference[name:'i_def']>'.`.
> 
> This blocks the code from being generated, so I'm unable to test that the code is being generated as expected. Were you after me writing another test to fail in this case? Because I presume that we definitely do want to test what is generated

Hi @mo-alistairp you'll need to fix what you have so that no generation error is raised. You probably need to switch to using tags in a few more places.

mo-alistairp: Ah ok, I see what you mean - thanks! I'll try and get that working.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Good point, fixed

src/psyclone/tests/test_files/lfric/28.scalar_array_invoke.f90
mo-alistairp: Fixed and added a comment below to explain the potentially confusing use of `dims_integer_array` as a scalar value

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Whoops, fixed!

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: You're right, it wasn't needed - so I removed it

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Done

src/psyclone/domain/lfric/kern_call_arg_list.py
mo-alistairp: Changed it to use tags

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
mo-alistairp: Reduced the size of test

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
mo-alistairp: Added

src/psyclone/tests/test_files/lfric/28.scalar_array_invoke.f90
mo-alistairp: It's been changed and works now

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Done

arporter:
> With regards to the change in `external/` - what is your process for getting something added there? I currently have a draft PR in `MetOffice/lfric_core` but I'm unsure if you need that to be committed first to get your integration tests to work

You can just update it as part of this PR. In order to get the integration tests to work, it will *also* need to be added to stfc/lfric_core. Ideally we'll remove this duplication and perhaps have a git submodule pointing to stfc/lfric_core but we're in a transition stage at the moment.

mo-alistairp:
> > With regards to the change in `external/` - what is your process for getting something added there? I currently have a draft PR in `MetOffice/lfric_core` but I'm unsure if you need that to be committed first to get your integration tests to work
> 
> You can just update it as part of this PR. In order to get the integration tests to work, it will _also_ need to be added to stfc/lfric_core. Ideally we'll remove this duplication and perhaps have a git submodule pointing to stfc/lfric_core but we're in a transition stage at the moment.

Ok, sure - thanks. Is it sufficient then if I wait until this is merged, then go through the merge process of the `MetOffice/lfric_core` PR so you can just update your stfc fork to main? Or does the stfc change need to be done sooner, e.g., by me opening another PR in stfc/lfric_core?

src/psyclone/domain/lfric/kern_call_acc_arg_list.py
```diff
@@ -271,10 +271,10 @@ def scalar(self, scalar_arg, var_accesses=None):
         :type var_accesses: Optional[
             :py:class:`psyclone.core.VariablesAccessMap`]
 
+        :raises NotImplementedError: OpenACC data regions are not currently
+            supported for arrays os scalars.
```
arporter: "of"

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -162,9 +164,37 @@ def stub_declarations(self):
                         f"ScalarArray argument '{arg.declaration_name}'"
                         f". Supported types are "
                         f"{const.VALID_SCALAR_DATA_TYPES}.")
+                self._scalar_array_args[arg.intent].append(arg)
 
         # Create declarations
-        self._create_declarations()
+        for intent in FORTRAN_INTENT_NAMES:
+            for arg in self._scalar_array_args[intent]:
+                if arg._array_ndims >= 1:
```
arporter: I should have spotted this last time but isn't `arg._array_ndims` guaranteed to be >=1 (otherwise it's not a valid ScalarArray)? If so, this `if` too can be removed.

src/psyclone/domain/lfric/lfric_scalar_array_args.py
```diff
@@ -174,49 +204,40 @@ def _create_declarations(self):
...
-                            array_symbol = self.symtab.find_or_create(
-                                arg.name,
-                                symbol_type=DataSymbol,
-                                datatype=ArrayType(
-                                    type_map[arg.intrinsic_type],
-                                    sym_list))
-                        array_symbol.interface = ArgumentInterface(
-                                            INTENT_MAPPING[intent])
-                        self.symtab.append_argument(array_symbol)
+                if arg._array_ndims >= 1:
```
arporter: Same comment about `_array_ndims` here too.

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
```diff
@@ -639,54 +638,26 @@ def test_scalar_array(tmpdir):
...
+        "    do cell = loop1_start, loop1_stop, 1\n"
+        "      call testkern_two_int_scalars_code(nlayers_f1, a, f1_data, "
+        "f2_data, f3_data, f4_data, b, ndf_w1, undf_w1, map_w1(:,cell), "
+        "ndf_w2, undf_w2, map_w2(:,cell), ndf_w3, undf_w3, map_w3(:,cell))\n"
+        "    enddo\n"
     )
-    assert expected in generated_code
+    assert expected_subroutine in generated_code
+    assert expected_declarations in generated_code
+    assert expected_calls[0], expected_calls[1] in generated_code
```
arporter: I don't think this line is correct: a comma after an assert gives a thing to be printed if the assert fails:
```python
>>> assert "andy", "porter" in "andy porter"
>>> assert "andy", "sporter" in "andy porter"
>>> assert 0, "sporter" in "andy porter"
  File "<python-input-2>", line 1, in <module>
    assert 0, "sporter" in "andy porter"
           ^
AssertionError: False
```
You'll need to split this into two separate asserts, one for expected_calls[0] and the other for ...[1].

src/psyclone/tests/test_files/lfric/28.scalar_array_invoke.f90
arporter: Great :-)

src/psyclone/tests/domain/lfric/arg_ordering_test.py
arporter: I think CodeCov's complaint is genuine. This test is currently testing `KernStubArgList` but you *also* need to test `ArgOrdering` itself (because the `scalar` method has been overridden in `KernStubArgList`). Quite how precisely this indirect coverage change has occurred I'm not sure but I don't think that matters. If you follow what I said in the original comment then that should fix it.

src/psyclone/domain/lfric/kern_call_acc_arg_list.py
mo-alistairp: Good spot - fixed

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: I think you're right - I've removed this

src/psyclone/domain/lfric/lfric_scalar_array_args.py
mo-alistairp: Same as above

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
mo-alistairp: I've fixed this

src/psyclone/tests/domain/lfric/arg_ordering_test.py
mo-alistairp: Ahhh, appologies for my confusion - the overridding went right past me! I've fixed this now and it seems to be passing CodeCov

src/psyclone/tests/domain/lfric/arg_ordering_test.py
```diff
@@ -217,6 +217,14 @@ def test_kernel_stub_invalid_scalar_argument():
     assert (f"Expected argument type to be one of "
             f"{const.VALID_SCALAR_NAMES + const.VALID_ARRAY_NAMES} "
             f"but got 'invalid'" in str(excinfo.value))
+    # Now call KernCallArgList to raise an exception
+    create_arg_list = KernCallArgList(kernel)
+    with pytest.raises(InternalError) as excinfo:
+        create_arg_list.scalar(arg)
+    const = LFRicConstants()
```
arporter: You've already got a `const` object but if this is the only thing then I'll do it.

arporter: Actually, you also need to update the name and docstring of the test (now that it does both stub and normal).

Issue #3091, opened 2025-08-11
OMPMinimiseSyncTrans should not fail with a barrier that prevents looping back


arporter:
@LonelyCat124 I have surrounded the minimise transformation with a try/catch so then such transformation issue is found the routine is left alone. However, I think it points to a legitimate issue that was already on master for async, the diff show things like
<img width="1251" height="330" alt="image" src="https://github.com/user-attachments/assets/cd60b185-1103-4bf3-9bde-284a2ad6337e" />
The left-hand-side is master, there is a top-level loop above the conditions. If I follow the control flow I can reach the end of the loop and loop-back to the same target nowait without finding any barrier. My understanding is that this means that a target can overlap with itself from a different iteration. This did not happen in practice probably because the scheduler is more conservative, but my understanding from the standard is that this is possible and the barrier should not be removed/minimised here. I think we need a new issue that the OMPMinimiseSyncTrans should not consider this as errors.

_Originally posted by @sergisiso in https://github.com/stfc/PSyclone/issues/3062#issuecomment-3164485554_

sergisiso:
Now that the branch has been merged, the output is correct (rhs of screenshot above), the only remaining issue is the the transformation throws a TransformationError for such a case, when it should let it through as the output is correct.

LonelyCat124:
I'm a bit confused and think this could be either a bug in the DUC w.r.t nested ifs, or something I didn't consider in async_trans_mixin w.r.t nested if statements / ifelse statements.

PR #3103, opened 2025-08-19
(Closes/towards #3091) Fix to the AsyncTransMixin to handle the new dependencies found betwe…

LonelyCat124:
…en nodes in the if and else part of IfBlocks


sergisiso (approved):
@LonelyCat124 The fix looks good, on the PR that moves the integration test to Rupert I enabled the async for BENCH. Trying to run this PR manually on BENCH I got a signal 11 failure on one file. But the file is quite simple and does not include any cases related to this issue/PR, so I added `OFFLOADING_ISSUES.append("trcldf.f90")` if you are ok with it?

PR #3106, opened 2025-08-20
(Closes #3105) add print handlers for Sympy boolean literals

### LINKED ISSUE #3105: Sympy reader does not work with boolean Literals
A Fortran boolean literal gets converted into e.g. "False" or "True" by the sympy writer. Unfortunately, the sympy reader creates symbols named "False" or "True" from these values. This can be seen by adding a further case to `tests/psyir/frontend/sympy_reader_test::test_sympy_psyir_from_expression`:
```python
                                         (".TRUE. .AND. .FALSE.",
                                          ".FALSE."),
                                         ])
def test_sympy_psyir_from_expression(fortran_reader, fortran_writer,
                                     expressions):
    ...
```
gives:
```bash
FAILED psyir/frontend/sympy_reader_test.py::test_sympy_psyir_from_expression[expressions26] - AssertionError: assert 'False' == '.FALSE.'
```

(I've hit this while working on #3067.)

arporter:


src/psyclone/psyir/frontend/sympy_reader.py
```diff
@@ -63,9 +63,17 @@ def _print_Equivalent(self, expr):
 
     def _print_Xor(self, expr):
         '''Called when converting an XOR expression, which in Fortran
-        is NEQV.'''
+        is .NEQV.'''
```
hiker, line 66: Thanks!

hiker (approved):
Thanks for fixing this.

Issue #2446, opened 2023-12-19
Update OMP/ACCLoopTrans with the ability to mark atomics/reductions variables


sergisiso:
We can insert Atomic directives but currently we also need to force the OMP/ACC loop to parallelise regardless of the dependency analysis, which is not ideal.

We could instead provide an option to the OMP/ACCLoop trans with the specific list of variables to check for the right kind of dependency and add itself the atomic when guaranteed to be True.

Or to make it more general purpose, we could explore letting the OMP/ACCLoopTrans (maybe again with an option) add the atomic when needed to solve any found dependency.

The same applies with reduction clauses when this is added to the PSyIR.

arporter:
One of the reasons the DA will refuse to parallelise a loop is if it writes to a scalar. We could use that information to generate the appropriate clauses (possibly)?

PR #3108, opened 2025-08-20
Infer reduction clauses in parallel loop transformations

mn416:
This is work in progress. The intention is to create a PR early so that CI can identify issues.

To infer reduction clauses, pass `enable_reductions=True` to the `apply()` method of `OMPLoopTrans`.

There are a number of TODOs in the code identifying things that would need to be polished before any potential merge.

The main limitations that I'm aware of are:

  * It only targets loops, not general parallel regions.
  * It only targets scalars, not arrays (array reduction variables have been supported since OpenMP 4.5).
  * No thought has been given to what happens when the same variable occurs both in a reduction clause and a data-sharing clause.
  * We'd like to add `OMPReductionClause`s to loop directives, instead of using the `_reduction_string()` method of `OMPDoDirective`.
  * We'd like an option along the lines of `force_reductions={'+': var_name}` to introduce reduction clauses that are not inferred.

All tests are passing locally, but so far there is only one test that exercises the new functionality.

There is also the question of whether the general approach I have taken is sensible or not. My hope is that this PR will be useful, even if the PSyclone team opts to implement the feature from scratch.

### DIFF
--- a/examples/nemo/scripts/omp_cpu_trans.py
+++ b/examples/nemo/scripts/omp_cpu_trans.py
@@ -82,7 +82,8 @@ def trans(psyir):
         return
 
     omp_parallel_trans = None
-    omp_loop_trans = OMPLoopTrans(omp_schedule="static")
+    omp_loop_trans = OMPLoopTrans(omp_schedule="static",
+                                  enable_reductions=True)
     omp_loop_trans.omp_directive = "paralleldo"
 
     for subroutine in psyir.walk(Routine):
--- a/src/psyclone/psyir/nodes/omp_directives.py
+++ b/src/psyclone/psyir/nodes/omp_directives.py
@@ -1785,6 +1785,11 @@ class OMPDoDirective(OMPRegionDirective, DataSharingAttributeMixin):
                 parts.append(f"reduction("
                              f"{OMP_OPERATOR_MAPPING[reduction_type]}:"
                              f"{reduction})")
+        # TODO[mn416]: use the OMPReductionClause class to insert reduction
+        # clauses into directives, rather than this low-level approach.
+        if isinstance(self.dir_body[0], Loop):
+            for (op, var) in self.dir_body[0].inferred_reduction_vars:
+                parts.append(f"reduction({op}:{var})")
         return ", ".join(parts)
 
     @property
--- a/src/psyclone/psyir/transformations/parallel_loop_trans.py
+++ b/src/psyclone/psyir/transformations/parallel_loop_trans.py
@@ -48,7 +48,8 @@ from psyclone.core import Signature
 from psyclone.domain.common.psylayer import PSyLoop
 from psyclone.psyir import nodes
 from psyclone.psyir.nodes import (
-        Call, Loop, Reference, Routine
+        Call, Loop, Reference, Routine,
+        BinaryOperation, IntrinsicCall, Assignment
 )
 from psyclone.psyir.tools import DependencyTools, DTCode
 from psyclone.psyir.transformations.loop_trans import LoopTrans
@@ -134,6 +135,143 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
 
         return True
 
+    # TODO[mn146]: use a data type for reduction operators not strings.
+    @staticmethod
+    def _get_reduction_operator(node):
+        ''' Return the reduction operator at the root of the given
+        DataNode or None if there isn't one.
+        '''
+        if isinstance(node, BinaryOperation):
+            if node.operator is BinaryOperation.Operator.ADD:
+                return "+"
+            elif node.operator is BinaryOperation.Operator.SUB:
+                return "-"
+            elif node.operator is BinaryOperation.Operator.MUL:
+                return "*"
+            elif node.operator is BinaryOperation.Operator.AND:
+                return ".and."
+            elif node.operator is BinaryOperation.Operator.OR:
+                return ".or."
+            elif node.operator is BinaryOperation.Operator.EQV:
+                return ".eqv."
+            elif node.operator is BinaryOperation.Operator.NEQV:
+                return ".neqv."
+        if isinstance(node, IntrinsicCall):
+            if node.intrinsic is IntrinsicCall.Intrinsic.MAX:
+                return "max"
+            if node.intrinsic is IntrinsicCall.Intrinsic.MIN:
+                return "min"
+            if node.intrinsic is IntrinsicCall.Intrinsic.IAND:
+                return "iand"
+            if node.intrinsic is IntrinsicCall.Intrinsic.IOR:
+                return "ior"
+            if node.intrinsic is IntrinsicCall.Intrinsic.IEOR:
+                return "ieor"
+        return None
+
+    @staticmethod
+    def _get_write_reduction(node, var_name):
+        '''Return the reduction operator for given node if it is an Assignment
+           of the form _either_
+
+             <var_name> = <var_name> <op> <DataNode>
+
+           or
+
+             <var_name> = <DataNode> <op> <var_name>
+
+           where <op> is a valid reduction operator. Otherwise return None.
+
+        :param node: the node to match against
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+        :param str var_name: the candidate reduction variable.
+
+        :returns: the reduction operator, or None
+        :rtype: str
+        '''
+        if isinstance(node, Reference) and node.name == var_name:
+            if isinstance(node.parent, Assignment):
+                op = ParallelLoopTrans._get_reduction_operator(node.parent.rhs)
+                if op:
+                    ok0 = (isinstance(node.parent.rhs.children[0], Reference)
+                           and node.parent.rhs.children[0].name == var_name)
+                    ok1 = (isinstance(node.parent.rhs.children[1], Reference)
+                           and node.parent.rhs.children[1].name == var_name)
+                    if ok0 != ok1:  # exclusive or
+                        return op
+        return None
+
+    @staticmethod
+    def _get_read_reduction(node, var_name):
+        '''Return reduction operator for given node if it is the child
+           of a DataNode which is the RHS of an Assignment of the form
+
+             <var_name> = <var_name> <op> <DataNode>
+
+           or
+
+             <var_name> = <DataNode> <op> <var_name>
+
+           where <op> is a valid reduction operator. Otherwise return None.
+
+        :param node: the node to match against
+        :type node: :py:class:`psyclone.psyir.nodes.Node`
+        :param str var_name: the candidate reduction variable.
+
+        :returns: the reduction operator, or None
+        :rtype: str
+        '''
+        if isinstance(node, Reference) and node.name == var_name:
+            op = ParallelLoopTrans._get_reduction_operator(node.parent)
+            if op:
+                if isinstance(node.parent.parent, Assignment):
+                    if isinstance(node.parent.parent.lhs, Reference):
+                        if node.parent.parent.lhs.name == var_name:
+                            return op
+            return None
+
+    @staticmethod
+    def _attempt_reduction(node, var_name, access_info, dry_run=False):
+        ''' Check and (if dry_run is False) add the variable to the
+        set of variables supporting reduction for the given loop node.
+
+        :param node: the loop that will be parallelised.
+        :type node: :py:class:`psyclone.psyir.nodes.Loop`
+        :param access_info: the variable (and its accesses) that we want to
+           consider as a reduction variable.
+        :type access_info: :py:class:`psyclone.core.SingleVariableAccessInfo`
+        :param bool dry_run: whether to actually add the variable as
+           one that supports reduction.
+
+        :returns: whether the variable supports reduction.
+        :rtype: bool
+        '''
+        # Find all the reduction operators used for the given variable name.
+        # Return early if we ever encounter a use of the variable which is
+        # not in the form of a reduction.
+        ops = []
+        for access in access_info.all_read_accesses:
+            op = ParallelLoopTrans._get_read_reduction(access.node, var_name)
+            if op is None:
+                return False
+            ops.append(op)
+        for access in access_info.all_write_accesses:
+            op = ParallelLoopTrans._get_write_reduction(access.node, var_name)
+            if op is None:
+                return False
+            ops.append(op)
+        if ops == []:
+            return False
+
+        # All potential reductions must use the same operator
+        if any(op != ops[0] for op in ops):
+            return False
+
+        # Add the operator/variable pair to the Loop node
+        if not dry_run:
+            node.inferred_reduction_vars.add((op, var_name))
+        return True
+
     def validate(self, node, options=None, **kwargs):
         '''
         Perform validation checks before applying the transformation
@@ -160,6 +298,8 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
             execution.
         :param bool privatise_arrays: whether to declare as private any
             write after write dependency symbols.
+        :param bool enable_reductions: whether to attempt parallelisation
+            of loops containing reductions.
 
         :raises TypeError: if 'collapse' is not an int or a bool.
         :raises TypeError: if 'ignore_dependencies_for' is not a list of str.
@@ -191,6 +331,7 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
                 ignore_dependencies_for = []
             sequential = self.get_option("sequential", **kwargs)
             privatise_arrays = self.get_option("privatise_arrays", **kwargs)
+            enable_reductions = self.get_option("enable_reductions", **kwargs)
         else:
             verbose = options.get("verbose", False)
             collapse = options.get("collapse", False)
@@ -200,6 +341,7 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
             )
             sequential = options.get("sequential", False)
             privatise_arrays = options.get("privatise_arrays", False)
+            enable_reductions = options.get("enable_reductions", False)
 
         # Check we are not a sequential loop
         if (not sequential and isinstance(node, PSyLoop) and
@@ -283,6 +425,18 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
                                 f"because it is not a plain local array or "
                                 f"it is used after the loop.")
                     continue
+                # See if the scalar in question allow parallelisation of
+                # the loop using reduction clauses.
+                if (enable_reductions and
+                        message.code == DTCode.WARN_SCALAR_REDUCTION):
+                    if (len(message.var_names) == 1 and
+                            len(message.var_infos) == 1):
+                        var_name = message.var_names[0]
+                        access_info = message.var_infos[0]
+                        if (self._attempt_reduction(node, var_name,
+                                                    access_info,
+                                                    dry_run=True)):
+                            continue
                 errors.append(str(message))
 
             if errors:
@@ -310,7 +464,8 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
               collapse: Union[int, bool] = False, force: bool = False,
               ignore_dependencies_for: Union[None, List[str]] = None,
               privatise_arrays: bool = False, sequential: bool = False,
-              nowait: bool = False, **kwargs):
+              nowait: bool = False, enable_reductions: bool = False,
+              **kwargs):
         '''
         Apply the Loop transformation to the specified node in a
         Schedule. This node must be a Loop since this transformation
@@ -351,6 +506,8 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
             execution.
         :param bool privatise_arrays: whether to make the write after write
             dependency symbols declared as private.
+        :param bool enable_reductions: whether to attempt parallelisation
+            of loops containing reductions.
 
         '''
         if not options:
@@ -358,7 +515,8 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
                     verbose=verbose, collapse=collapse,
                     ignore_dependencies_for=ignore_dependencies_for,
                     privatise_arrays=privatise_arrays,
-                    sequential=sequential, nowait=nowait, **kwargs
+                    sequential=sequential, nowait=nowait,
+                    enable_reductions=enable_reductions, **kwargs
             )
             # Rename the input options that are renamed in this apply method.
             # TODO 2668, rename options to be consistent.
@@ -376,12 +534,14 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
             list_of_names = options.get("ignore_dependencies_for", [])
             privatise_arrays = options.get("privatise_arrays", False)
             nowait = options.get("nowait", False)
+            enable_reductions = options.get("enable_reductions", False)
 
         self.validate(node, options=options, verbose=verbose,
                       collapse=collapse,
                       ignore_dependencies_for=ignore_dependencies_for,
                       privatise_arrays=privatise_arrays,
-                      sequential=sequential, nowait=nowait, **kwargs)
+                      sequential=sequential, nowait=nowait,
+                      enable_reductions=enable_reductions, **kwargs)
 
         list_of_signatures = [Signature(name) for name in list_of_names]
         dtools = DependencyTools()
@@ -401,6 +561,19 @@ class ParallelLoopTrans(LoopTrans, AsyncTransMixin, metaclass=abc.ABCMeta):
                 if message.code == DTCode.ERROR_WRITE_WRITE_RACE:
                     for var_name in message.var_names:
                         self._attempt_privatisation(node, var_name)
+
+        # If 'enable_reductions' is specified, see if any of the scalars
+        # preventing parallelisation can be supported by introducing
+        # reduction clauses.
+        if enable_reductions and not node.independent_iterations(
+                 dep_tools=dtools,
+                 test_all_variables=True,
+                 signatures_to_ignore=list_of_signatures):
+            for message in dtools.get_all_messages():
+                if message.code == DTCode.WARN_SCALAR_REDUCTION:
+                    for (var_name, var_info) in zip(message.var_names,
+                                                    message.var_infos):
+                        self._attempt_reduction(node, var_name, var_info)
 
         # If 'collapse' is specified, check that it is an int and that the
         # loop nest has at least that number of loops in it
--- a/src/psyclone/tests/psyir/nodes/omp_directives_test.py
+++ b/src/psyclone/tests/psyir/nodes/omp_directives_test.py
@@ -4651,3 +4651,25 @@ def test_omp_serial_check_dependency_valid_pairing():
     assert test_dir._check_dependency_pairing_valid(
                array_reference1, array_reference2, None, None
            )
+
+
+# TODO[mn416]: more tests needed
+def test_enable_reductions(fortran_reader, fortran_writer):
+    ''' Tests the functionality of the enable_reductions option to
+    the apply() method of OMPLoopTrans.
+    '''
+    psyir = fortran_reader.psyir_from_source('''
+        function sum_arr(arr) result (acc)
+            integer, intent(in) :: arr(:)
+            integer :: i
+            integer :: acc = 0
+
+            do i = 1, ubound(arr)
+                acc = acc + arr(i)
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    omplooptrans.apply(loop, enable_reductions=True)
+    output = fortran_writer(psyir)
+    assert "reduction(+:acc)" in output
--- a/src/psyclone/psyir/nodes/loop.py
+++ b/src/psyclone/psyir/nodes/loop.py
@@ -106,6 +106,10 @@ class Loop(Statement):
         # if this loop is run concurrently. Alternatively this could be
         # implemented by moving the symbols to the loop_body symbol table.
         self._explicitly_private_symbols = set()
+        # Hold the set of operator/variable pairs that can be used to create
+        # reduction clauses if the loop is run concurrently.
+        # TODO[mn416]: it's probably not necessary to store these here
+        self._inferred_reduction_vars = set()
 
     def __eq__(self, other):
         '''
@@ -134,6 +138,16 @@ class Loop(Statement):
         :rtype: Set[:py:class:`psyclone.psyir.symbols.DataSymbol`]
         '''
         return self._explicitly_private_symbols
+
+    @property
+    def inferred_reduction_vars(self):
+        '''
+        :returns: the set of operator/variable pairs that can be used to
+            create reduction clauses if the loop is run concurrently.
+        :rtype: Set[(str, str)]
+        '''
+        # TODO[mn416]: to use a richer datatype than strings.
+        return self._inferred_reduction_vars
 
     @property
     def loop_type(self):
@@ -465,6 +479,9 @@ class Loop(Statement):
                 except KeyError:
                     pass
         super().replace_symbols_using(table_or_symbol)
+
+        # TODO[mn416]: anything to do for self._inferred_reduction_vars
+        # here?
 
     def __str__(self):
         # Give Loop sub-classes a specialised name
--- a/src/psyclone/psyir/tools/dependency_tools.py
+++ b/src/psyclone/psyir/tools/dependency_tools.py
@@ -89,15 +89,25 @@ class Message:
     :param int code: error or warning code.
     :param var_names: list of variable names (defaults to []).
     :type var_names: List[str]
+    :param var_infos: list of variable access infos (defaults to []).
+    :type var_infos: List[:py:class:`psyclone.core.SingleVariableAccessInfo`]
+
 
     '''
-    def __init__(self, message, code, var_names=None):
+    # TODO[mn416]: we want the access info for each variable
+    # in the message, but the above representation does not maintain
+    # a clear mapping between the two. There must be a cleaner way.
+    def __init__(self, message, code, var_names=None, var_infos=None):
         self._message = message
         self._code = code
         if var_names:
             self._var_names = var_names
         else:
             self._var_names = []
+        if var_infos:
+            self._var_infos = var_infos
+        else:
+            self._var_infos = []
 
     # ------------------------------------------------------------------------
     def __str__(self):
@@ -126,6 +136,17 @@ class Message:
         # We convert each expression into a string to support LazyStrings
         # inside of 'var_names'
         return [str(i) for i in self._var_names]
+
+    # ------------------------------------------------------------------------
+    @property
+    def var_infos(self):
+        ''':returns: the access info of each variable to which the message
+        applies.
+
+        :rtype: List[:py:class:`psyclone.core.SingleVariableAccessInfo`]
+
+        '''
+        return self._var_infos
 
 
 # ============================================================================
@@ -170,7 +191,7 @@ class DependencyTools():
         self._messages = []
 
     # -------------------------------------------------------------------------
-    def _add_message(self, message, code, var_names=None):
+    def _add_message(self, message, code, var_names=None, var_infos=None):
         '''Adds an informational message to the internal message
         handling system.
 
@@ -190,7 +211,7 @@ class DependencyTools():
             raise InternalError(f"Unknown message code {code}.")
 
         self._messages.append(Message(f"{message_type}: {message}", code,
-                                      var_names))
+                                      var_names, var_infos))
 
     # -------------------------------------------------------------------------
     def get_all_messages(self):
@@ -775,7 +796,7 @@ class DependencyTools():
         self._add_message(f"Variable '{access_info.var_name}' is read "
                           f"first, which indicates a reduction.",
                           DTCode.WARN_SCALAR_REDUCTION,
-                          [access_info.var_name])
+                          [access_info.var_name], [access_info])
         return False
 
     # -------------------------------------------------------------------------
--- a/src/psyclone/tests/psyir/transformations/transformations_test.py
+++ b/src/psyclone/tests/psyir/transformations/transformations_test.py
@@ -577,7 +577,8 @@ def test_omploop_trans_new_options(sample_psyir):
     assert ("'OMPLoopTrans' received invalid options ['fakeoption1', "
             "'fakeoption2']. Valid options are '['node_type_check', "
             "'verbose', 'collapse', 'force', 'ignore_dependencies_for', "
-            "'privatise_arrays', 'sequential', 'nowait', 'reprod']."
+            "'privatise_arrays', 'sequential', 'nowait', 'enable_reductions', "
+            "'reprod']."
             in str(excinfo.value))
 
     # Check we get the relevant error message when submitting multiple


LonelyCat124:
I set the integration tests running to see if we maintain correctness with this change for NEMO4/5

examples/nemo/scripts/omp_cpu_trans.py
```diff
@@ -111,5 +111,6 @@ def trans(psyir):
                     loop_directive_trans=omp_loop_trans,
                     collapse=False,
                     privatise_arrays=(not NEMOV4 and
-                                      psyir.name not in PRIVATISATION_ISSUES)
+                                      psyir.name not in PRIVATISATION_ISSUES),
+                    enable_reductions=True
```
sergisiso: @mn416 You could move this to the omp_gpu_trans.py in the same folder, in that script we have a flag to enable/disable reproducibility:
https://github.com/stfc/PSyclone/blob/5614b304dec4cdfe332cc53652473f24ab12b8ac/examples/nemo/scripts/omp_gpu_trans.py#L58-L59

so this would be:
`enable_reductions=not REPRODUCIBLE`

During the integration test this is build with both REPRODUCIBLE set to true and false. (the false is not diff against the expected output, but still compiled and run)

@LonelyCat124 For correctness I would suggest manually doing
./makenemo ....
and
REPRODUCIBLE=1 ./makenemo ....

and diff the generated code to see if all the changes make sense

(another option is to make a longer run and see if the run.stat values don't diverge above a tolerance, but we can do this later)

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -402,6 +562,19 @@ def apply(self, node, options=None, verbose: bool = False,
                     for var_name in message.var_names:
                         self._attempt_privatisation(node, var_name)
 
+        # If 'enable_reductions' is specified, see if any of the scalars
+        # preventing parallelisation can be supported by introducing
+        # reduction clauses.
+        if enable_reductions and not node.independent_iterations(
```
sergisiso: In NEMO I see that reductions currently stop loop collapsing, e.g. in:
```
!$omp teams loop collapse(1) reduction(+:peke) default(shared) private(ji,jj,jk)
do jk = 1, jpkm1, 1
    ! Warning: Variable 'peke' is read first, which indicates a reduction. Variable: 'peke'. Consider using the
    !& "ignore_dependencies_for" transformation option if this is a false dependency.
    do jj = loop_start_10, loop_stop_10, 1
        do ji = loop_start_11, loop_stop_11, 1
           peke = peke + zkepe(ji,jj,jk) * (gdept_1d(jk) * (1._wp + r3t(ji,jj,kmm))) * e1e2t(ji,jj) * (e3t_3d(ji,jj,jk) * (1._wp + r3t(ji,jj,kmm) * tmask(ji,jj,jk)))
         enddo
    enddo
enddo
```
Collapse uses a similar analysis with "node.independent_iterations(" below in the same apply, could we extend the same reductions check there?

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -160,6 +298,8 @@ def validate(self, node, options=None, **kwargs):
             execution.
         :param bool privatise_arrays: whether to declare as private any
             write after write dependency symbols.
+        :param bool enable_reductions: whether to attempt parallelisation
+            of loops containing reductions.
```
LonelyCat124: All the :param: options on validate apart from `node` and `options` should be removed as part of this, since this uses the `transformation_docstring_wrapper`.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -283,6 +425,18 @@ def validate(self, node, options=None, **kwargs):
                                 f"because it is not a plain local array or "
                                 f"it is used after the loop.")
                     continue
+                # See if the scalar in question allow parallelisation of
+                # the loop using reduction clauses.
+                if (enable_reductions and
+                        message.code == DTCode.WARN_SCALAR_REDUCTION):
```
sergisiso: This is the case that I was worried about following the DependencyTools errors, as it does not distinguish partial_reductions (OpenMP scans) from full reductions, but I see you handle it well. Well done @mn416 
```
      ! PSyclone: Loop cannot be parallelised because:
      ! Warning: Variable 'npti' is read first, which indicates a reduction. Variable: 'npti'.
      ! The write-write dependency in 'nptidx' cannot be solved by array privatisation because it is not a plain local array or it 
!& is used after the loop.
      ! Consider using the "ignore_dependencies_for" transformation option if this is a false dependency
      ! Consider using the "array_privatisation" transformation option if this is a write-write dependency
      do jj = loop_start_7, loop_stop_7, 1
        do ji = loop_start_8, loop_stop_8, 1
          if (a_i(ji,jj,jl) > epsi10) then
            npti = npti + 1
            nptidx(npti) = (jj - 1) * jpi + ji
          end if
        enddo
      enddo
```
Make sure to add some test for this as this is a common pattern in NEMO.

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Thanks @sergisiso, I will add this (or something similar) as a test case and try to fix the issue. At first glance, we may be able to add reduction variables to the `ignore_signatures` list as they are inferred, so that the subsequent call to `independent_iterations` (for the collapse check) does not fail.

examples/nemo/scripts/omp_cpu_trans.py
sergisiso: I tried the checks mentioned above and both the visual inspection and the results after a longer run look good:
```
it :     500    |ssh|_max:  0.2060750522516920D+01 |U|_max:  0.1156737469059852D+01 |V|_max:  0.9906046447276886D+00 S_min:  0.4637403064969927D+01 S_max:  0.4057228496928409D+02
it :     500    |ssh|_max:  0.2060750522532980D+01 |U|_max:  0.1156737468997449D+01 |V|_max:  0.9906046447204880D+00 S_min:  0.4637403064968112D+01 S_max:  0.4057228496928410D+02
```

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: I've now done this in commit a3d5575.

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
```diff
@@ -106,6 +107,13 @@ def infer_sharing_attributes(self) -> \
         fprivate = set()
         need_sync = set()
 
+        # Collate reduction variables
+        red_vars = []
+        for clause in self.children:
+            if isinstance(clause, OMPReductionClause):
```
LonelyCat124, line 113: Can you add something here like `TODO #2446 Ensure this behaves correctly for OpenACC when reductions are supported.`

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -1695,6 +1716,7 @@ def __eq__(self, other):
         is_eq = is_eq and self.omp_schedule == other.omp_schedule
         is_eq = is_eq and self.reprod == other.reprod
         is_eq = is_eq and self.collapse == other.collapse
+        is_eq = is_eq and self.children[1:] == other.children[1:]
```
LonelyCat124: I think the super.__eq__ should handle children equality (from `Node` which has:
```
        for index, child in enumerate(self.children):
            is_eq = is_eq and child == other.children[index]
```
) - can you add a test to check the reduction clause needs to be equal and remove this addition and check the result is still correct.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -1754,6 +1776,19 @@ def collapse(self, value):
...
+        '''Adds the given OMPReductionClause to the list of reduction clauses.
+
+        :param clause: an OMPReductionClause to be added.
+        :type clause: :py:class:`psyclone.psyir.nodes.OMPReductionClause`
+
+        :raises GenerationError: if the OMPDoDirective is malformed.
+        '''
+        if len(self.children) < 1:
+            raise GenerationError("Calling add_reduction_clause() on a "
+                                  "malformed OMPDoDirective.")
```
LonelyCat124: I'm not sure if this is necessary (since its handled by `_validate_child`) - is this an error that is easy to hit? Otherwise I think we can just leave it to be handled by the `addchild` call

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2002,6 +2040,19 @@ def _validate_child(position, child):
...
+        '''Adds the given OMPReductionClause to the list of reduction clauses.
+
+        :param clause: an OMPReductionClause to be added.
+        :type clause: :py:class:`psyclone.psyir.nodes.OMPReductionClause`
+
+        :raises GenerationError: if the OMPParalleDoDirective is malformed.
+        '''
+        if len(self.children) < 5:
+            raise GenerationError("Calling add_reduction_clause() on a "
+                                  "malformed OMPParallelDoDirective.")
```
LonelyCat124: See previous comment.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2166,6 +2240,7 @@ def __eq__(self, other):
         '''
         is_eq = super().__eq__(other)
         is_eq = is_eq and self.collapse == other.collapse
+        is_eq = is_eq and self.children[1:] == other.children[1:]
```
LonelyCat124: See previous comment on `OMPDoDirective.__eq__`

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2225,6 +2300,19 @@ def collapse(self, value):
...
 
+    def add_reduction_clause(self, clause):
+        '''Adds the given OMPReductionClause to the list of reduction clauses.
+
+        :param clause: an OMPReductionClause to be added.
+        :type clause: :py:class:`psyclone.psyir.nodes.OMPReductionClause`
+
+        :raises GenerationError: if the OMPLoopDirective is malformed.
+        '''
+        if len(self.children) < 1:
```
LonelyCat124: See previous comments on `add_reduction_clause` functions.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+from psyclone.psyir.nodes import (
+        Node, Reference, BinaryOperation, IntrinsicCall, Assignment
+)
+
+
+class ReductionInferenceTool():
+
+    def __init__(self, red_ops: List[Union[BinaryOperation.Operator,
+                                           IntrinsicCall.Intrinsic]] = []):
+        '''Instances of this class are initialsied with a set of allowed
```
LonelyCat124: I think for longer docstrings we prefer to start on a new line and have less indentation, i.e.:
```
def myfunc(...):
    '''
    docstring is long.
    
    :param myarg1: ...
    '''
```
Can you update the docstrings in this file.
Also I think we have the docstring for `__init__` on the class declaration and not the init function itself.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+           :param red_ops: a list of allowed reduction operators.
+        '''
+        self.red_ops = red_ops
+
+    def _get_reduction_operator(self, node: Node) -> \
+            Union[BinaryOperation.Operator, IntrinsicCall.Intrinsic]:
+        '''Return the reduction operator at the root of the given
+           DataNode or None if there isn't one.
+
+           :param node: the node to match against.
```
LonelyCat124: Missing `:returns:` docstring

LonelyCat124: I think with this being a protected member,  I'm happy if you simply move the main docstring of this function into `:returns: [current docstring starting with lower case]`.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+                    return node.operator
+        if isinstance(node, IntrinsicCall):
+            for op in self.red_ops:
+                if node.intrinsic == op:
+                    return node.intrinsic
+        return None
+
+    @staticmethod
+    def _match_sig(ref: Reference, sig: Signature) -> bool:
+        '''Returns True if the Signature of the given Reference
```
LonelyCat124: Swap to `:returns:`. Also add parameter docstrings (even though they're very self explanatory I think its just compulsory).

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+           matches the given Signature, and the Reference involves no
+           array indicies.  Returns False otherwise.
+        '''
+        (ref_sig, ref_indices) = ref.get_signature_and_indices()
+
+        # Check that there are no indices invovled
+        no_indices = sum(ref_indices, []) == []
```
LonelyCat124, line 90: Since OpenMP does support array indices, could you just say that PSyclone doesn't support them (if I understand correctly as this is the reason why we're forcing no indices).

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+
+           where <op> is an allowed reduction operator and the Signature
+           of <Reference> is a scalar reference matching the given Signature.
+           Otherwise, return None.
+
+        :param node: the node to match against.
+        :param sig: the candidate reduction variable.
+        :returns: the reduction operator, or None.
+        '''
+        if isinstance(node, Reference):
```
LonelyCat124, line 121: Since I think this doesn't allow array reductions right now, and structure accesses are forbidden, we can probably just do `type(node) is Reference` here to automatically skip over `ArrayReference` and `StructureReference` nodes.

It should also have a `# pylint: disable=unidiomatic-typecheck` comment before - just a quirk of how `Reference` classes work in `PSyclone`.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+            if self._match_sig(node, sig):
+                if isinstance(node.parent, Assignment):
+                    op = self._get_reduction_operator(node.parent.rhs)
+                    if op:
+                        child_ok = []
+                        for child in node.parent.rhs.children[:2]:
```
LonelyCat124, line 136: Can you add some comments to explain the specific checks here.

Also I need to tag @sergisiso here - does PSyclone now always remember bracket structure from the original source (so this should correctly handle code like `a = (b+c+d+e+f+...)+a` will always find the `a` as children[1]. I think we can live with requiring brackets in reduction statements - I don't remember what the OpenMP standard would require.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+           of <Reference> is a scalar reference matching the given Signature.
+           Otherwise, return None.
+
+        :param node: the node to match against.
+        :param sig: the candidate reduction variable.
+        :returns: the reduction operator, or None.
+        '''
+        if isinstance(node, Reference):
+            if self._match_sig(node, sig):
+                if isinstance(node.parent, Assignment):
```
LonelyCat124: Is the only way to ever reach this function such that `node.parent.lhs is node`? If there is any doubt with how things could end up here, I'd prefer to also add the `node.parent.lhs is node` check here just to be safe (I'd also be tempted to do the read reduction one with `node.parent.lhs is not node`).

I think we're probably safe to skip these, am just trying to convince myself of that (e.g. what happens if `b = b + func(b)` if `func` isn't pure, PSyclone will probably assume this `b` is a `write_access`). Probably worth checking what happens with the code if it contains something like this as I don't think its safe to allow something like that as a reduction either way.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,200 @@
...
+            ops.append(op)
+            ref = access.node
+        for access in access_info.all_write_accesses:
+            op = self._get_write_reduction(access.node, sig)
+            if op is None:
+                return None
+            ops.append(op)
+            ref = access.node
+
+        # No suitable reductions found?
```
LonelyCat124: Remove the ? here I think

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -160,6 +164,12 @@ def validate(self, node, options=None, **kwargs):
             execution.
         :param bool privatise_arrays: whether to declare as private any
             write after write dependency symbols.
+        :param reduction_ops: if non-empty, attempt parallelisation
+            of loops by inferring reduction clauses involving any of
+            the reduction operators in the list.
+        :type reduction_ops: List[
+            :py:class:`psyclone.psyir.nodes.BinaryOperation.Operator` |
+            :py:class:`psyclone.psyir.nodes.IntrinsicCall.Intrinsic`]
```
LonelyCat124: We should check by building docs - but all the params on `validate` docstring should be removed and the docs should still build correctly the validate docstrings with the appropriate options.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -351,14 +382,21 @@ def apply(self, node, options=None, verbose: bool = False,
             execution.
         :param bool privatise_arrays: whether to make the write after write
             dependency symbols declared as private.
+        :param reduction_ops: if non-empty, attempt parallelisation
+            of loops by inferring reduction clauses involving any of
+            the reduction operators in the list.
+        :type reduction_ops: List[
```
LonelyCat124: Since we have the type defined in the arglist, you don't need the `type` docstring, its generated automatically now.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -351,14 +382,21 @@ def apply(self, node, options=None, verbose: bool = False,
...
 
         '''
         if not options:
             self.validate_options(
                     verbose=verbose, collapse=collapse,
                     ignore_dependencies_for=ignore_dependencies_for,
                     privatise_arrays=privatise_arrays,
-                    sequential=sequential, nowait=nowait, **kwargs
+                    sequential=sequential, nowait=nowait,
+                    reduction_ops=reduction_ops, **kwargs
```
LonelyCat124, line 381: Since reduction ops is a list (alongside ignore_dependencies_for) can you explicitly check the type of the input is as expected? `validate_options` isn't advanced enough to handle that case.

Additionally, we usually have the default value for input lists be `None` and set them to `[]` if we find None (as otherwise the same list is reused every time from what I remember) - can you adjust that as well.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4651,3 +4653,418 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+    loop_directive.add_reduction_clause(clause)
+    output = fortran_writer(psyir)
+    assert "reduction(+: acc)" in output
+
+
+def test_reduction_arith_ops(fortran_reader, fortran_writer):
+    ''' Test that reduction loops involing arithmetic reduction operators are
+    parallelised.
+    '''
+    ops = ["+", "-", "*"]
```
LonelyCat124: It would be better if the tests that test multiple ops were done with `@pytest.mark.parametrize` just so if one were to fail it wouldn't prevent the others from running - could you rewrite to use that fixtuer?

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4651,3 +4653,418 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+    with pytest.raises(TransformationError) as err:
+        omplooptrans.apply(loop, enable_reductions=True)
+    assert ("Variable 'count' is read first, which indicates a reduction"
+            in str(err.value))
+
+
+def test_reduction_teams(fortran_reader, fortran_writer):
+    ''' Test that reduction loops with a teams directive are parallelised.
+    '''
+    dirs = ["teamsdistributeparalleldo", "teamsloop"]
```
LonelyCat124: Again - use parametrize please.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4651,3 +4653,418 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+        omplooptrans.apply(loop, enable_reductions=True)
+        output = fortran_writer(psyir)
+        assert "reduction(+: acc)" in output
+
+
+def test_reduction_do_loop(fortran_reader, fortran_writer):
+    ''' Test that reduction loops with a do/loop directive inside a parallel
+        region are parallelised.
+    '''
+    dirs = ["do", "loop"]
```
LonelyCat124: parametrize

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4651,3 +4653,418 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+        end function''')
+    omplooptrans = OMPParallelLoopTrans()
+    loop = psyir.walk(Loop)[0]
+    omplooptrans.apply(loop, options={"enable_reductions": True})
+    output = fortran_writer(psyir)
+    assert "reduction(+: acc)" in output
+
+
+def test_reduction_struct_member(fortran_reader, fortran_writer):
+    ''' Test that reduction loops involing struct members are
```
LonelyCat124: involving (typo)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4651,3 +4653,418 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+            type(struct_type) :: struct
+
+            struct%acc = 0
+            do i = 1, ubound(arr)
+                struct%acc = struct%acc + arr(i)
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    # omplooptrans.apply(loop, enable_reductions=True)
```
LonelyCat124: Since this isn't supported by OpenMP yet I think just remove these commented out lines as they won't be supported any time soon by PSyclone.

src/psyclone/psyir/transformations/omp_loop_trans.py
```diff
@@ -308,18 +339,41 @@ def apply(self, node, options=None,
...
+                    reduction_ops=red_ops,
+                    **kwargs
             )
             self._reprod = reprod
         else:
             self._reprod = options.get("reprod",
                                        Config.get().reproducible_reductions)
-
-        super().apply(node, options, **kwargs)
+            if options.get("enable_reductions", False):
```
LonelyCat124, line 365: Since `enable_reductions` is a new option, I'm ok with us not supporting it from the (now deprecated) `options` dict so you don't need to handle this. You can also propogate that upwards and not support `reduction_ops` as part of the `options` dict.

If you do want to keep the support for reductions from the options dict, then you'll need to instead create a copy of the `options` dict here to modify and pass upwards, as the `options` input is meant to be immutable (which python doesn't have great support for).

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
mn416: Done (commit 4c084944a765a31a2287d8f9ecc3674cc171a65d)

src/psyclone/psyir/nodes/omp_directives.py
mn416: Done (commit e06d93e0ec4c4d517f18bd7e0746836670657f38)

src/psyclone/psyir/nodes/omp_directives.py
mn416: Agreed, I think it's not easy to hit. Now removed (commit 3e074d49e05fa0798a041f4441e3c41a2897d969).

src/psyclone/psyir/nodes/omp_directives.py
mn416: Removed (commit 3e074d49e05fa0798a041f4441e3c41a2897d969).

src/psyclone/psyir/nodes/omp_directives.py
mn416: Removed (commit b563464cbbb7dfa80350c8c97a9fea888ea55a38)

src/psyclone/psyir/nodes/omp_directives.py
mn416: Resolved (commit 3e074d49e05fa0798a041f4441e3c41a2897d969)

src/psyclone/psyir/tools/reduction_inference.py
mn416: Fixed (commit e1f7f605a8f6559bf0b8c2de5144529c2c84f9f5)

src/psyclone/psyir/tools/reduction_inference.py
mn416: Done (commits e1f7f605a8f6559bf0b8c2de5144529c2c84f9f5 and 2fc49872505d2eb719aa5f183bbc2dadac31762b)

src/psyclone/psyir/tools/reduction_inference.py
mn416: Done (commits e1f7f605a8f6559bf0b8c2de5144529c2c84f9f5 and 2fc49872505d2eb719aa5f183bbc2dadac31762b)

src/psyclone/psyir/tools/reduction_inference.py
mn416: Done (commit 2fc49872505d2eb719aa5f183bbc2dadac31762b)

src/psyclone/psyir/tools/reduction_inference.py
mn416: `ReductionInferenceTool` is able to infer struct members as reduction variables if the "singleton check" is removed in the signature-matching function. Although OpenMP doesn't support it, PSyclone could introduce a temporary variable to workaround that restriction (this possibility is documented in the signature matching function). If you're ok with it, I think it's nice to allow all kinds of reference at this point in the code.

src/psyclone/psyir/tools/reduction_inference.py
mn416: > Is the only way to ever reach this function such that `node.parent.lhs is node`?

I don't think so. This function is called on all write accesses to a variable. As you suggest, it is probably possible for a write access to exist that is not in an `Assignment` statement, e.g. a `Call`.

> what happens if `b = b + func(b)`

The access to `b` in `func(b)` will not match either the read reduction form or the write reduction form, so a reduction will not be inferred.

> I'd prefer to also add the `node.parent.lhs is node` check here just to be safe

I like that suggestion. Done in commit 409fb7d59f3b8e9386e7bcab68c5fd101dd4d0d6.

> I'd also be tempted to do the read reduction one with node.parent.lhs is not node

Do you mean `node.parent.parent.lhs is not node`? That seems like a less useful check, and a bit confusing (to me), so I've omitted it for now.

src/psyclone/psyir/tools/reduction_inference.py
mn416: Comments added in commit 9895c6d7137ad36c49cd900d3559df46af70061b.

Regarding the bracketing, I added a new test for this in the same commit.

src/psyclone/psyir/tools/reduction_inference.py
mn416: Done (part of commit 9895c6d7137ad36c49cd900d3559df46af70061b)

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Removed in commit 2f305e97aff0cc4338803b06f55eb386f2227bd9. I built the docs and the `validate` parameters ares still documented.

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Done in commit 2f305e97aff0cc4338803b06f55eb386f2227bd9.

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Both done in commit d11053983169bbac24eb77e33741d38f129fa5eb.

I also added a test to check that the type checking was working (which in turn uncovered a bug where I was passing `dict_keys` instead of a `list`).

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done in commits de1527762cf9634f9993826b2db5a5fa1129f32e and 8c5bef1a73d7af39fe32d1c45799d43b5a2755b7

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done in commits de1527762cf9634f9993826b2db5a5fa1129f32e and 8c5bef1a73d7af39fe32d1c45799d43b5a2755b7

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done in commits de1527762cf9634f9993826b2db5a5fa1129f32e and 8c5bef1a73d7af39fe32d1c45799d43b5a2755b7

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done in commits de1527762cf9634f9993826b2db5a5fa1129f32e and 8c5bef1a73d7af39fe32d1c45799d43b5a2755b7

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done in commit ed424187504c14121d26215a2f0ebed6c347707c.

src/psyclone/psyir/transformations/omp_loop_trans.py
mn416: I decided to keep `options` support, and made changes to keep it immutable in commit ed424187504c14121d26215a2f0ebed6c347707c.

examples/nemo/scripts/omp_cpu_trans.py
mn416: > You could move this to the omp_gpu_trans.py in the same folder, in that script we have a flag to enable/disable reproducibility

Thanks @sergisiso. I'm not really sure how to test modifications to this file, so it may be better for @LonelyCat124 to try it.

sergisiso: These are part of inegration tests. Aidan will run these for you as part of the review process.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -200,6 +184,24 @@ def validate(self, node, options=None, **kwargs):
             )
             sequential = options.get("sequential", False)
             privatise_arrays = options.get("privatise_arrays", False)
+            reduction_ops = options.get("reduction_ops", [])
+
+        # Check type of reduction_ops (not handled by validate_options)
+        reduction_ops_ok = False
+        if isinstance(reduction_ops, List):
```
LonelyCat124: I think `list` here, we only need `List` in type definitions due to 3.9 support, but for type checking we should use `list`.

W.r.t the TypeError below, as far as I can work out this would also happen if the `isinstance(reduction_ops, list)`, but the error message doesn't refer to that case. I think I'd prefer to have two error cases, e.g.:
```
if not isinstance(reduction_ops, list):
   raise TypeError(f"Reduction_ops for {self.class.__name__} should be a list but found {reduction_ops}")
   
for op in reduction_ops:
    if not isinstance(op, (BinaryOperation.Operator, IntrinsicCall.Intrinsic)):
      raise TypeError(...)
```
For this error can you also add what the found element causing the failure was?

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -283,6 +285,17 @@ def validate(self, node, options=None, **kwargs):
                                 f"because it is not a plain local array or "
                                 f"it is used after the loop.")
                     continue
+                # See if the scalar in question allow parallelisation of
```
LonelyCat124: s/allow/allows/

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -1754,6 +1775,14 @@ def collapse(self, value):
 
         self._collapse = value
 
+    def add_reduction_clause(self, clause):
```
LonelyCat124: Since these are just proxy methods for `addchild` can you remove the 3 implementations and just replace their uses with addchild.

src/psyclone/psyir/tools/reduction_inference.py
```diff
@@ -0,0 +1,227 @@
...
+
+from typing import Union, List, Tuple
+
+from psyclone.core import (AccessSequence, Signature)
+from psyclone.psyir.nodes import (
+        Node, Reference, BinaryOperation, IntrinsicCall, Assignment
+)
+
+
+class ReductionInferenceTool():
```
LonelyCat124, line 48: There is no unit testing for this class - can you add it? We're trying to have self contained unit testing code for new modules (i.e. ` pytest --cov psyclone.psyir.tools.reduction_inference tests/psyir/tools/` should show 100%, preferably from a `reduction_inference_test.py` file).

LonelyCat124: (This isn't checked by the code coverage tool we use in the github actions unfortunately).

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Done (commit 7348b5a18b6e44424be484dfefd247fa023837ab)

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Done (commit 7348b5a18b6e44424be484dfefd247fa023837ab)

src/psyclone/psyir/nodes/omp_directives.py
mn416: Done (commit 2da66c25ec0a13b08c22d69c1b61998956905a7f)

src/psyclone/psyir/tools/reduction_inference.py
mn416: Done (commit 12ff32da85e6469774d331b1065ec15bf87e57ad). Thanks for suggesting the coverage-checking command. I see 100% coverage of `reduction_inference_tool.py` from the newly added `reduction_inference_test.py`.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+
+    clause1 = OMPReductionClause(OMPReductionClause.ReductionClauseTypes.ADD)
+    clause1.addchild(Reference(Symbol("foo")))
+    do_directive1.addchild(clause1)
+
+    clause2 = OMPReductionClause(OMPReductionClause.ReductionClauseTypes.ADD)
+    clause2.addchild(Reference(Symbol("bar")))
+    do_directive2.addchild(clause2)
+
+    assert do_directive1 != do_directive2
```
LonelyCat124, line 4796: In case of future changes to equality (in inherited sections), can you also add a test where `==` is true.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    omplooptrans.apply(loop, enable_reductions=True)
+    output = fortran_writer(psyir)
+    assert f"reduction({op}: acc)" in output
+
+
+def test_bracketed_reductions(fortran_reader, fortran_writer):
+    ''' Test brakcted reductions are correctly detected.
```
LonelyCat124: s/brakcted/bracketed/

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+
+            do i = 1, ubound(arr)
+                acc = acc + arr(i)
+                if (arr(i) > 0) then
+                    acc = acc * 2
+                end if
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
```
LonelyCat124, line 4992: Can you add a TODO to this test (to #2446 and #514 ) saying to improve this error message in the future, as I assume this comes from the current dependency analysis, but isn't really the issue we want here. If I'm wrong and this is from your new reduction_inference can you add the TODO to that file too.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+            do i = 1, size(arr)
+                if (arr1(i) > 0) then
+                    count = count + 1
+                end if
+                arr2(i) = count
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    with pytest.raises(TransformationError) as err:
```
LonelyCat124, line 5043: Same TODO as above.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+            integer :: i
+            integer :: count = 1
+
+            do i = 1, size(arr)
+                count = count + count
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    with pytest.raises(TransformationError) as err:
```
LonelyCat124, line 5065: Same TODO

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+            integer :: i
+            integer :: count = 1
+
+            do i = 1, size(arr)
+                count = count / 2
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    with pytest.raises(TransformationError) as err:
```
LonelyCat124, line 5087: Same TODO

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -4658,6 +4678,503 @@ def test_omp_serial_check_dependency_valid_pairing():
...
+            type(struct_type) :: struct
+
+            struct%acc = 0
+            do i = 1, ubound(arr)
+                struct%acc = struct%acc + arr(i)
+            end do
+        end function''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    with pytest.raises(TransformationError) as err:
```
LonelyCat124, line 5176: Another TODO here please.

LonelyCat124 (changes requested):
@mn416 thanks for this, I've asked for some TODOs and one additional test section but otherwise I think we're pretty much ready.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 0c1b0c35d2cb8b30b4891b7169e709d7f6b61245)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 1d2f1c2851ad1b3f4a05e99b6318085967eddef1)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 1d2f1c2851ad1b3f4a05e99b6318085967eddef1). It does indeed come from dependency analysis. It's possible that we could augment the error message along the lines of "... and cannot be inferred as a reduction variable" inside `ParallelLoopTrans`. Let me know if you'd like me to try that.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 1d2f1c2851ad1b3f4a05e99b6318085967eddef1)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 1d2f1c2851ad1b3f4a05e99b6318085967eddef1)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 1d2f1c2851ad1b3f4a05e99b6318085967eddef1)

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit 1d2f1c2851ad1b3f4a05e99b6318085967eddef1)

examples/nemo/scripts/omp_cpu_trans.py
LonelyCat124: These have run successfully - I will remove being enabled before the merge as we haven't yet done the work to evidence enabling it by default.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
LonelyCat124: No, I think I'm happy with the TODOs as when we improve reduction detection I hope we should improve error messages too.

src/psyclone/psyir/tools/reduction_inference.py
LonelyCat124: Somehow i'd missed this conversation until now. I'm happy with this now.

Issue #2659, opened 2024-07-12
LFRicTypes creates orphan Symbols


arporter:
Currently, LFRicTypes just creates DataSymbols and ContainerSymbols (primarily for precision and their imports) as and when it needs them. They are never added to a SymbolTable. This has then meant that we have to take special action in `datatypes::ScalarType.__eq__` as often, two otherwise identical LFRicIntegerScalarTypes have different precision symbols (with the same name).

Clearly, the symbols that LFRicTypes creates should be added to a table. Quite how to manage this isn't clear.

Similarly, `LFRicConstants.precision_for_type` currently uses `LFRicTypes` and returns a Symbol that again is not in a table. Given that this method currently has to locally import LFRicTypes, I suspect that really it should just return a name (str) and let the caller create a suitable symbol in the correct table.

sergisiso:
If I go to a commit before #2633 was merged it works.

arporter:
Now that you say this, I remember hitting it too (but had forgotten). I think the LFRic compiler settings might be such that importing something but never using it will be flagged as an error. It would be better if we could fix it properly.

arporter:
If I just run PSyclone on lfric_xios_setup_mod.x90 then the generated PSy layer does have:
```
  MODULE lfric_xios_setup_mod_psy
    USE constants_mod, ONLY: r_def, i_def
    USE field_mod, ONLY: field_type, field_proxy_type
    IMPLICIT NONE
    CONTAINS
```
Did you get the failure after applying an optimisation script @sergisiso?

arporter:
If I apply the 'standard' `global.py` used by the MO then the generated PSy layer does still have the required USE statement.

sergisiso:
Let me try to replicate it, but for reference I was using lfric_apps 2689 and I just did `./local_build.py -a gungho_model` to get the error

arporter:
Ah, OK. I was just testing with lfric_core (which is where the file lives) @r50758. I've just tried with 'everything-everywhere-all-at-once' and that looked OK too.

sergisiso:
Well, after removing the directories and trying again (the `./local_build.py -a gungho_model -t clean` hanged for me), now it works

arporter:
Yes, the `clean` target always hangs for me too - the `rm` that it does is missing a `-f` and some of the files are protected.

hiker:
[...]

PR #3110, opened 2025-08-21
(Closes #3087) Handle expressions in the kind attribute of a declaration

### LINKED ISSUE #3087: Kind expressions are lost
Given:
`integer(kind=2*i_def), dimension(2) :: c`
psyclone currenty outputs:
`integer(kind=i_def), dimension(2) :: c`

The reason is that fparser has this walk:
`kind_names = walk(kind_selector.items, Fortran2003.Name)`
but even updating it to:
`self.process_nodes(parent=dummy_schedule, nodes=kind_selector.items)`
there the precision only accepts a Symbol or Precison enumerator.

This needs to be updated as this is a passthrough-bug

LonelyCat124:
First implementation to handle kinds like `kind=2*wp`. Not sure if I've missed anything as I'm not super knowledgable about what kinds can be - maybe there's some missing error checking or similar.

Still have a few failing tests to address.


src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -419,12 +419,14 @@ def __init__(self, intrinsic, precision):
                 f"'{type(intrinsic).__name__}'.")
 
         self._intrinsic = intrinsic
-
-        if not isinstance(precision, (int, ScalarType.Precision, DataSymbol)):
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.operation import BinaryOperation
+        if not isinstance(precision, (BinaryOperation, int,
+                                      ScalarType.Precision, DataSymbol)):
```
sergisiso: I have not review the PR yet but I have not seen any limitation in the standard that this could not be a UnaryOp `kind=-(-4)` or a Intrinsic `kind=abs(4)` or accessors `kind=map(3)`, I would be tempted to accept any DataNode, which also includes Literal - instead of int, and Reference instead of DataSymbol.

sergisiso: This is why in the issue I was suggesting to use:
`self.process_nodes(parent=dummy_schedule, nodes=kind_selector.items)`
to parse whatever the expression is without limitations, and end up with an error if it is something not supported, which the current PR implementation doesn't do:
```
        kind_names = walk(kind_selector.items,
                          (Fortran2003.BinaryOpBase, Fortran2003.Name))
```
would still silently skip the remainder of the expression.

LonelyCat124: Ok, I'll try to move it to general DataNode.

LonelyCat124: Main theme I'm unsude on is how to modify:
```
        if not kind_names:
            raise NotImplementedError(
                f"Failed to find valid Name in Fortran Kind Selector: "
                f"{kind_selector}'")
```
this error now.

LonelyCat124: I will continue to allow integer specifications of datatype precisions, as otherwise I get stuck in circular dependency hell when trying to import Literal to this file for the definitions at the end of the file.

LonelyCat124: image
Think I need a bit more energy to fix this, I'll try to finish this off tomorrow. I guess there are other places that are creating DataSymbol precisions.

LonelyCat124: Down to 67 failing tests that probable need addressing one by one now - there were issues in Symbols.reference_accesses, lfric, psyGen and psyad (some still exist in psyad) that covered a lot of the failing tests. A lot of the remaining ones are tests written that create something with a datatype and things, so fewer easy wins.

I expect to be missing coverage and I think docs still need improving in a few places so I'll keep going at it.

LonelyCat124:
Running integration tests to see if #2659 is still an issue there...

LonelyCat124:
> Nemov4 and v5 integration tests show failures like: `psyclone.psyir.symbols.symbol.SymbolError: PSyclone SymbolTable error: Unable to find the scope of node 'Reference[name:'wp']' as none of its ancestors are Container or Schedule nodes.` both in:
> 
> ```
>   File "/archive/psyclone-tests/action-runner-software/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.13/site-packages/psyclone/psyir/nodes/omp_directives.py", line 1470, in infer_sharing_attributes
>     return super().infer_sharing_attributes()
>            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
>   File "/archive/psyclone-tests/action-runner-software/actions-runner/_work/PSyclone-mirror/PSyclone-mirror/.runner_venv/lib/python3.13/site-packages/psyclone/psyir/nodes/data_sharing_attribute_mixin.py", line 120, in infer_sharing_attributes
>     symbol = accesses[0].node.scope.symbol_table.lookup(
>              ^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> and there is a coverage line missing. Could you look at these?

This was a test of a previous push to see if we needed some changes I did last night - I'll have to rerun now to see if they work. I'll check again though, I assume this should be covered but I'm not sure what by. I guess the last changes I did last night broke coverage again so I'll check.

LonelyCat124:
Think I've a potential fix, I'll rerun integration.

LonelyCat124:
Still have some coverage issues from the latest patches - waiting to see if it fixes integration before I resolve those.

LonelyCat124:
@sergisiso I think this is fully functional now and coverage is all done except the 1 currently uncovered line from master we said to ignore. Back to you now.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -232,6 +233,7 @@ def _validate_schedule(self, node, kname, kern_or_call, kernel_schedule):
                 # Compare the routine to be inlined with the one that
                 # is already present.
                 new_rts = self._prepare_code_to_inline([kernel_schedule])
+                print(new_rts[0] == routine, len(new_rts))
```
sergisiso: Remove

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -294,8 +296,10 @@ def _prepare_code_to_inline(
                     if isinstance(symbol.datatype, DataTypeSymbol):
                         symbols_to_bring_in.add(symbol.datatype)
                     elif hasattr(symbol.datatype, 'precision'):
-                        if isinstance(symbol.datatype.precision, Symbol):
-                            symbols_to_bring_in.add(symbol.datatype.precision)
+                        if isinstance(symbol.datatype.precision, DataNode):
+                            for ref in symbol.datatype.precision.walk(
+                                    Reference):
+                                symbols_to_bring_in.add(ref.symbol)
```
sergisiso: Not your fault but what about symbols in a shape expression or in a init expression?

tbh I believe the `code_to_inline.reference_accesses()` should already include them all. So maybe this is leftover code from before we had that functionality? Does it work by deleting it?

Anyway, `src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py::670` seems to be testing this, could you extend it to be `real(kind=r_def*kind_multiplier), dimension(N) :: a = initval`

LonelyCat124:
@sergisiso I'm not sure if `DataType.copy()` may need adjusting? If so let me know.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -419,25 +418,28 @@ def __init__(self, intrinsic, precision):
...
-                not (isinstance(precision.datatype, ScalarType) and
-                     precision.datatype.intrinsic ==
-                     ScalarType.Intrinsic.INTEGER) and
-                not isinstance(precision.datatype, UnresolvedType)):
-            raise ValueError(
-                f"A DataSymbol representing the precision of another "
-                f"DataSymbol must be of either 'unresolved' or scalar, "
-                f"integer type but got: {precision}")
+        if (isinstance(precision, DataNode)):
+            for ref in precision.walk(DataNode):
```
sergisiso: Why don't we do precision.datatype? instead of iterating subexpression

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -485,14 +487,23 @@ def __eq__(self, other):
...
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.datanode import DataNode
+        from psyclone.psyir.nodes.reference import Reference
+        if (isinstance(other.precision, DataNode)
+                and isinstance(self.precision, DataNode)):
+            self_refs = self.precision.walk(Reference)
+            other_refs = other.precision.walk(Reference)
+            precision_match = True
+            # If the precision is defined by DataNode then we need to check
+            # the symbol names and interfaces of all contained References.
```
sergisiso: But this would fail for different expression e.g. `r_kind*4` vs `r_kind*8` vs `r_kind+4`

Since this is a special case to avoid a specific bug, would the follwing be enough?
```
if isinstance(other.precision, Reference):
  # TODO #2659: We compare single symbols by name because LFRicIntegerScalarDataType
  # creates multiple equivalent symbols with the same name
  precision_match = (
                  other.precision.name == self.precision.name and
                  other.precision.interface == self.precision.interface)
else:
   # Everyting else we compare the full expression
   precision_match = self.precision == other.precision
```

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -510,18 +521,22 @@ def replace_symbols_using(self, table_or_symbol):
...
+                # Update any 'precision' information.
+                new_sym = None
+                if isinstance(table_or_symbol, Symbol):
+                    if table_or_symbol.name.lower() == ref.symbol.name.lower():
+                        new_sym = table_or_symbol
+                else:
+                    new_sym = table_or_symbol.lookup(ref.symbol.name,
+                                                     otherwise=None)
+                if new_sym:
+                    ref.symbol = new_sym
```
sergisiso: Would the following work instead?
```
if isinstance(self.precision, DataNode):
    self._precision.replace_symbol_using(table_or_symbol)
```

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -534,15 +549,20 @@ def reference_accesses(self):
...
-            from psyclone.core.access_type import AccessType
+        # Avoid circular import
+        # pylint: disable=import-outside-toplevel
+        from psyclone.core.access_type import AccessType
+        from psyclone.psyir.nodes.datanode import DataNode
+        if isinstance(self.precision, DataNode):
+            precision_ras = self.precision.reference_accesses()
+            # Change all the precision_ras to be of access type
+            # TYPE_INFO instead, as this is a kind declaration not
+            # a read.
```
sergisiso: maybe: "Change all the accesses to be of TYPE_INFO type, since this is a kind expression we know the symbols are not runtime-read" ?

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -1036,17 +1056,6 @@ def replace_symbols_using(self, table_or_symbol):
 
         # TODO #1857: we will probably remove '_precision' and have
         # 'intrinsic' be 'datatype'.
-        if self._precision and isinstance(self._precision, Symbol):
```
sergisiso, line 1039: Shouldn't we still do `self._precision.replace_symbol_using(table_or_symbol)`?

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -1092,10 +1102,15 @@ def reference_accesses(self):
...
 
-        if isinstance(self.precision, Symbol):
-            access_info.add_access(
-                Signature(self.precision.name),
-                AccessType.TYPE_INFO, self)
+        if isinstance(self.precision, DataNode):
+            precision_ras = self.precision.reference_accesses()
+            # Change all the precision_ras to be of access type
+            # TYPE_INFO instead, as this is a kind declaration not
+            # a read.
```
sergisiso: Same as above

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -2736,10 +2736,13 @@ def lower_to_language_level(self):
         datatype = arg_refs[0].symbol.datatype
         if isinstance(datatype, UnsupportedFortranType):
             datatype = datatype.partial_datatype
+        precision = datatype.precision
+        if isinstance(precision, Reference):
+            precision = precision.symbol
```
sergisiso: I don't understand this snippet repeated in several places, if the precision is an expresion  the `Reference(precision)` below will fail. Shouldn't we just do:
`[arg_refs[1], ("kind", precision.datatype.copy())])`

If it is a Precision or int it will also fail, but I think the LFRic DSL says that kinds are always explict.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -93,102 +93,6 @@ def gen_intent(symbol):
...
-        if fortrantype.lower() == "real" and \
-           precision == ScalarType.Precision.DOUBLE:
-            return "double precision"
-        # This logging warning can be added when issue #11 is
-        # addressed.
-        # import logging
-        # logging.warning(
-        #      "Fortran does not support relative precision for the '%s' "
-        #      "datatype but '%s' was specified for variable '%s'.",
-        #      datatype, str(symbol.precision), symbol.name)
```
sergisiso, line 175: Reading this again I belive this was trying to prevent things like:
`scalar_type = ScalarType(ScalarType.Intrinsic.INTEGER, ScalarType.Precision.DOUBLE)`
as there is no double for integers in Fortran, while there is for C, e.g. long.

But I don't know why it is proposed as a warning, if it is not valid fortran it must be an Error.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -420,7 +421,8 @@ def _create_type_map(self, list_of_expressions: Iterable[Node],
                 # have a scope, hence the try...except.
                 orig_sym = sva[0].node.scope.symbol_table.lookup(sig.var_name)
             except SymbolError:
-                if isinstance(sva[0].node, Reference):
+                if (isinstance(sva[0].node, Reference) and
+                        sva[0].access_type != AccessType.TYPE_INFO):
```
sergisiso: Could you add a comment explaining why this is needed?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
                 f"Only real and integer literals are supported as arguments "
                 f"to the KIND intrinsic but found "
                 f"'{type(kind_arg).__name__}' in: {kind_selector}")
 
         # We have kind=kind-param
-        # TODO #3087: This misses expresssions such as "2*wp"
-        kind_names = walk(kind_selector.items, Fortran2003.Name)
-        if not kind_names:
+
+        # Create a dummy Schedule and Assignment to capture the kind=...
```
sergisiso: s/Schedule/Routine/

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+        # so we can capture expressions such as 2*wp.
+        # The input from fparser2 is ['(', kind, ')']
+        kind_items = kind_selector.items[1]
+        fake_routine = Routine(RoutineSymbol("dummy"))
+        dummy_schedule = Assignment()
```
sergisiso: Rename to dummy_assignment

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+        fake_routine.addchild(dummy_schedule)
+        dummy_schedule.addchild(Reference(Symbol("a")))
```
sergisiso: Is this a leftover?

sergisiso: Oh I see, is to create a lhs, but note that there is currently no difference, you could not do this and then get the "dummy_schedule.rhs.copy()". Also you can probably "detach" instead of "copy"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+        self.process_nodes(parent=dummy_schedule, nodes=[kind_items])
+        # Create a copy of the created node.
+        kindvar = dummy_schedule.rhs.copy()
```
sergisiso: s/kindvar/kind_expression/

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+        # For each symbol used in the BinaryOperation, we need to update
```
sergisiso: s/BinaryOperation/kind expression/

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+        # kindvar with the ones from the real symbol_table.
+        for ref in kindvar.walk(Reference):
+            sym_name = ref.symbol.name
+            sym = _kind_find_or_create(sym_name, symbol_table)
+            ref.symbol = sym
+        if len(kindvar.walk(CodeBlock)) != 0:
             raise NotImplementedError(
-                f"Failed to find valid Name in Fortran Kind Selector: "
-                f"{kind_selector}'")
+                "Unsupported kind declaration, error message NYI"
```
sergisiso: Remove ", error message NYI"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+            )
+        # Return the update BinaryOperation
+        return kindvar
 
-        return _kind_find_or_create(str(kind_names[0]), symbol_table)
+#        # Otherwise continue with the previous implementation.
+#        return _kind_find_or_create(str(kind_names[0]), symbol_table)
```
sergisiso: Remove

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,42 @@ def _process_precision(type_spec, psyir_parent):
...
+            sym_name = ref.symbol.name
+            sym = _kind_find_or_create(sym_name, symbol_table)
+            ref.symbol = sym
+        if len(kindvar.walk(CodeBlock)) != 0:
             raise NotImplementedError(
-                f"Failed to find valid Name in Fortran Kind Selector: "
-                f"{kind_selector}'")
+                "Unsupported kind declaration, error message NYI"
+            )
+        # Return the update BinaryOperation
```
sergisiso: Remove

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -695,11 +695,15 @@ def test_module_inline_apply_bring_in_non_local_symbols(
     end module my_mod
     ''')
 
+    print("----------------------------------------------")
```
sergisiso: Remove, and in line 702

src/psyclone/tests/lfric_test.py
```diff
@@ -4335,6 +4335,7 @@ def test_mixed_precision_args(tmpdir):
     psy = PSyFactory(TEST_API, distributed_memory=True).create(invoke_info)
     generated_code = str(psy.gen)
 
+    print(generated_code)
```
sergisiso: Remove

sergisiso (changes requested):
@LonelyCat124 Thanks for doing this change. Most coments are minor.

> I'm not sure if DataType.copy() may need adjusting? If so let me know.

Good question, which I don't know the answer. We need some tests for scalar_type and array_type (both have ._precision). I would expect the symbols to be the same but the references/node to be different.
So if we do a copy and modify the subtree of the kind of the copy, the original doesn't change. But if the symbol table renames a symbol name referenced in the kind expression, both are updated. Does this make sense?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
LonelyCat124: Done

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
LonelyCat124: I'm fairly sure I changed the call to e`_reference_accesses` to give these (as there was a failing test from what I remember) but I'll add it in as well and test removing this.

LonelyCat124: Commenting this out breaks `test_validate_local_routine` as it changes the error message it fails with.
It no longer fails at line `240` of `KernelModuleInlineTrans`, instead failing with the `The target of 'call do_something(a)' is already module inlined` error - I'm not sure how this section works at all (w.r.t versioning) so its a bit confusing.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: I didn't think about `.datatype` of `DataNode` - I suspect this is good enough (as if the `DataNode` doesn't have a `datatype` then it should have failed with the other implementation anyway also).

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Yeah, much neater.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Yes - I also have implemented a conversion function from READ to TYPE_INFO in #3119, so I'll add a TODO to the issue thats being solved so I try that function here later.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Probably yes, I added it back.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Done

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Done

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
sergisiso: The versioning has to do with the equality check of the whole routine. Does `routine == new_rts` in line 237 maybe doesn't return True for the same routine with kinds?

src/psyclone/domain/lfric/lfric_builtins.py
LonelyCat124: Ok - I probably didn't think about expressions, but also didn't know that the kinds were explicit so I will try that

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Needed to do this iff `_precision` and `_precision` is a datanode.

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Do you want to add in handling for this? I thought this was prevented by the ScalarType class.

src/psyclone/psyir/backend/sympy_writer.py
LonelyCat124: Added

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Fixed.

src/psyclone/psyir/frontend/fparser2.py

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I think I had some issues if I only didn't have the lhs. I've updated to use detach after though.

src/psyclone/psyir/frontend/fparser2.py

src/psyclone/psyir/frontend/fparser2.py

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Removed

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Added what the unsupported kind expression is to the error message.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Removed.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
LonelyCat124: Removed.

src/psyclone/tests/lfric_test.py
LonelyCat124: Removed.

src/psyclone/psyir/backend/fortran.py
sergisiso: If it is prevented from ScalarType then it is ok. Maybe then we need a comment where this is preventing saying that if the condition is lifted (e.g. to support C longs) we may want to the check to the backend.

src/psyclone/psyir/symbols/datatypes.py
sergisiso: Yes, that looks better, you just need to remove the superfluous parenthesis to make pylint happy.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -296,6 +295,7 @@ def copy(self):
         '''
         new = copy.copy(self)
         if self._partial_datatype:
+            print(type(self._partial_datatype.copy()))
```
sergisiso: Remove

src/psyclone/psyir/symbols/datatypes.py
sergisiso: Is `if isinstance(self.precision, DataNode):` enough?

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Thanks also add a comment above this statement with:
`# Create a dummy lhs with 'a ='

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -510,18 +513,11 @@ def replace_symbols_using(self, table_or_symbol):
             :py:class:`psyclone.psyir.symbols.Symbol`
 
         '''
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes.datanode import DataNode
         # Only the 'precision' of a ScalarType can refer to a Symbol.
```
sergisiso: I am not sure if this comment is outdated, but in any case I would remove it, the code is clear enough

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -3137,7 +3137,7 @@ def lower_to_language_level(self):
         # Create the PSyIR for the kernel:
         # proxy0%data(df) = REAL(proxy1%data, kind=r_<prec>)
         lhs = arg_refs[0]
-        r_precision = arg_refs[0].datatype.precision
+        r_precision = arg_refs[0].datatype.precision.symbol
```
sergisiso: Delete this line and make the kind below `arg_refs[0].datatype.precision.copy()` ?

src/psyclone/psyad/tl2ad.py
```diff
@@ -408,8 +408,8 @@ def generate_adjoint_test(tl_psyir, ad_psyir,
 
     # If the precision of the active variables is specified by another symbol
     # then we must ensure that it is declared in the harness too.
-    if isinstance(datatype.precision, Symbol):
-        _add_precision_symbol(datatype.precision, symbol_table)
+    if isinstance(datatype.precision, Reference):
```
sergisiso: It seems that here what we want is if datatype.precision is a DataNode, walk to each Reference, and add call the add_precision_symbol with the reference.symbol.

Same in the two other similar instance below

src/psyclone/psyir/backend/fortran.py
sergisiso: I checked and it is not prevented by ScalarType, add the check there or make it a VisitorError here.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -888,14 +882,17 @@ def _gen_parameter_decls(self, symbol_table, is_module_scope=False):
             # The dependence analysis tools do not include symbols used to
             # define precision so check for those here.
             for lit in symbol.initial_value.walk(Literal):
-                if isinstance(lit.datatype.precision, DataSymbol):
-                    read_write_info.add_read(
-                        Signature(lit.datatype.precision.name))
+                if isinstance(lit.datatype.precision, DataNode):
+                    refs = lit.datatype.precision.walk(Reference)
+                    for ref in refs:
```
sergisiso: Remove "refs" and inline expression

src/psyclone/psyir/backend/sympy_writer.py
sergisiso: I am still a bit lost here. Sorry is not your fault but I have a hard time understanding this method in general. But if I remove the added `sva[0].access_type != AccessType.TYPE_INFO` no tests fails. Maybe an example or a test that fails when removing this will help me understand the context.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,39 @@ def _process_precision(type_spec, psyir_parent):
             # arguments to KIND
             if isinstance(kind_arg, (Fortran2003.Int_Literal_Constant,
                                      Fortran2003.Real_Literal_Constant)):
-                return get_literal_precision(kind_arg, psyir_parent)
+                precision = get_literal_precision(kind_arg, psyir_parent)
+                return precision
```
sergisiso: Revert

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2833,22 +2832,39 @@ def _process_precision(type_spec, psyir_parent):
...
+        for ref in kind_expression.walk(Reference):
+            sym_name = ref.symbol.name
+            sym = _kind_find_or_create(sym_name, symbol_table)
+            ref.symbol = sym
+        if len(kind_expression.walk(CodeBlock)) != 0:
+            raise NotImplementedError(
+                f"Unsupported kind declaration: "
+                f"{kind_expression.debug_string()}"
+            )
+        return kind_expression
```
sergisiso, line 2878: Update the rtype from symbol.DataSymbol to nodes.DataNode

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
```diff
@@ -660,6 +660,7 @@ def test_apply_with_hoist_with_dependent_symbols(fortran_reader,
     hoist_trans = HoistLocalArraysTrans()
     hoist_trans.apply(routine)
     code = fortran_writer(psyir)
+    print(code)
```
sergisiso: Remove print (but since we are here, can you make `real(kind=1 * wp) :: a(10)` in the code above and the assert below to ensure that now the hoist validation checks for subexpressions)

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
```diff
@@ -111,6 +111,13 @@ def infer_sharing_attributes(self) -> \
         for signature in var_accesses.all_signatures:
             if not var_accesses[signature].has_data_access():
                 continue
+            # Skip those that are TYPE_INFO accesses.
+            # TODO #3060: This should be all instead of any if we correctly
+            # have TYPE_INFO (or another access type) to handle kind
+            # expressions in IntrinsicCalls labelled correctly.
```
sergisiso: I am not sure I agree with the TODO here. `any` seems the right choice because if any access is TYPE_INFO we know for sure that the variable is a parameter, regardless of how it is used elsewhere. Even if we fix the intrinsics reference_accesses, we may have an unknown function:
`a = 1_r_kind + unknownfunc(r_kind)`, but the literal is enough to know that this is a parameter. I would remove the TODO.

src/psyclone/tests/psyir/frontend/fparser2_select_type_test.py
```diff
@@ -407,6 +407,7 @@ def test_kind(fortran_reader, fortran_writer, tmpdir):
         "    end if\n").lower()
     psyir = fortran_reader.psyir_from_source(code)
     result = fortran_writer(psyir).lower()
+    print(result)
```
sergisiso: Remove

src/psyclone/tests/psyir/frontend/fparser2_kind_params_test.py
```diff
@@ -212,8 +213,8 @@ def test_process_declarations_kind_literals(vartype, kind, precision):
     fake_parent, _ = process_declarations(f"{vartype}(kind=KIND({kind})) :: "
                                           f"var")
     if not precision:
-        assert fake_parent.symbol_table.lookup("var").datatype.precision is \
-            fake_parent.symbol_table.lookup("t_def")
+        assert (fake_parent.symbol_table.lookup("var").datatype.precision ==
+                Reference(fake_parent.symbol_table.lookup("t_def")))
```
sergisiso: I think `precision.symbol is` would be clear than creating the reference for this assert

src/psyclone/tests/psyir/frontend/fparser2_kind_params_test.py
```diff
@@ -131,9 +132,9 @@ def test_process_declarations_kind_use():
     fake_parent, _ = process_declarations("use kind_mod, only: r_def\n"
                                           "real(kind=r_def) :: var2")
     var2_var = fake_parent.symbol_table.lookup("var2")
-    assert isinstance(var2_var.datatype.precision, DataSymbol)
-    assert fake_parent.symbol_table.lookup("r_def") is \
-        var2_var.datatype.precision
+    assert isinstance(var2_var.datatype.precision, Reference)
+    assert (Reference(fake_parent.symbol_table.lookup("r_def")) ==
+            var2_var.datatype.precision)
```
sergisiso: Same here: `is .. precision.symbol` instead of creating the Reference

src/psyclone/tests/psyir/frontend/fparser2_kind_params_test.py
```diff
@@ -102,6 +102,7 @@ def test_process_declarations_kind_new_param():
     processor.process_declarations(fake_parent, [fp2spec[0]], [])
     sym = fake_parent.symbol_table.lookup("var3")
     assert isinstance(sym, DataSymbol)
+    print(sym.datatype)
```
sergisiso: Remove

src/psyclone/tests/domain/lfric/lfric_types_test.py
```diff
@@ -117,9 +117,12 @@ def test_generic_scalars(type_name, symbol_name, intrinsic,
     precision = LFRicTypes(precision_name)
     # datatype
     lfric_datatype = data_type()
+    if isinstance(lfric_datatype.precision, Reference):
+        precision = Reference(precision)
     assert lfric_datatype.intrinsic == intrinsic
-    assert lfric_datatype.precision is precision
+    assert lfric_datatype.precision == precision
     # precision can be set explicitly
+    print(data_type)
```
sergisiso: Remove

src/psyclone/psyir/transformations/mark_routine_for_gpu_mixin.py
```diff
@@ -110,7 +111,8 @@ def validate_it_can_run_on_gpu(self, node, options):
             for sig in vam.all_signatures:
                 name = sig.var_name
                 first = vam[sig][0].node
-                if isinstance(first, (Symbol, DataType)):
+                if (isinstance(first, Reference)
+                        and vam[sig][0].access_type == AccessType.TYPE_INFO):
```
sergisiso: First, not you fault but can you remove the extra 'o' in the comment above this loop and add the indentation to docstring raises field please.

Then, are we saying that "parameters" are valid for the GPU even if they come from an import? Or I am misinterpreting what this code tries? In this case, would ` if vam[sig][0].access_type == AccessType.TYPE_INFO: continue` be enough? And also, like in the OpenMP case, should be do an `any(` instead of just checking the first reference? So maybe we can test that: `use other, only: mykind ... a = myfunc(mykind) + 1_mykind` can be offloaded to the gpu.

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -353,15 +354,24 @@ def _get_local_arrays(node):
...
-                        f"PSyclone warning: '{sym.name}' cannot be hoisted "
-                        f"to the global scope as '"
-                        f"{sym.datatype.precision.name}'"
-                        f" is not guaranteed to be a global symbol")
+            # Precision could include multiple symbols - handle in the same
+            # way as for DataSymbol but check all of them.
+            if isinstance(sym.datatype.precision, DataNode):
+                refs = sym.datatype.precision.walk(Reference)
+                failed = False
+                for ref in refs:
```
sergisiso: Remove refs and inline expression

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1544,14 +1544,19 @@ def precision_datasymbols(self):
...
         precision_symbols = set()
         for sym in self.datasymbols:
             # Not all types have the 'precision' attribute (e.g.
             # UnresolvedType)
             if (hasattr(sym.datatype, "precision") and
-                    isinstance(sym.datatype.precision, DataSymbol)):
-                precision_symbols.add(sym.datatype.precision)
+                    isinstance(sym.datatype.precision, DataNode)):
+                refs = sym.datatype.precision.walk(Reference)
+                for ref in refs:
```
sergisiso: inline refs expression

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Removed.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: It is (at least tests all pass) so I've swapped to that.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Done

src/psyclone/domain/lfric/lfric_builtins.py
LonelyCat124: Done

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Removed.

src/psyclone/domain/lfric/lfric_builtins.py
LonelyCat124: Done

src/psyclone/psyad/tl2ad.py
LonelyCat124: Yeah, done

src/psyclone/psyir/backend/fortran.py
LonelyCat124: @sergisiso I don't understand the comment here - can you explain what/why this would be ok? Is it just that a non-Reference data node isn't allowed as a Literal's precision? i.e. you can't have `1.0_2*wp`?

src/psyclone/psyir/frontend/fparser2.py

src/psyclone/psyir/backend/fortran.py
sergisiso: Sorry for the unclear comment. It is just a formatting suggestion to use: `for ref in lit.datatype.precision.walk(Reference):`, not a change in behaviour

src/psyclone/psyir/backend/sympy_writer.py
LonelyCat124: Ok, so I don't know what test/example failed for me to add this, or if I just made an assumption, but I'm going to need to tag @hiker as I'm not sure I know what the result of this function should be for TYPE_INFO elements.

On the create_type_map test I added `1_wp` to the list of expressions, and my assumption is we don't need the `wp` symbol in the type_map? Which is what the above code is attempting (and failing to do). Is that assumption correct, or should we still have `wp` in the type map?

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Done

src/psyclone/tests/psyir/transformations/hoist_local_arrays_trans_test.py
LonelyCat124: Done

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
LonelyCat124: Yeah I think that makes sense, I feel like I had forgotten about the parameter requirement.

src/psyclone/tests/psyir/frontend/fparser2_select_type_test.py
LonelyCat124: Removed.

src/psyclone/tests/psyir/frontend/fparser2_kind_params_test.py
LonelyCat124: Yeah, done.

src/psyclone/tests/psyir/frontend/fparser2_kind_params_test.py
LonelyCat124: Done

src/psyclone/tests/psyir/frontend/fparser2_kind_params_test.py
LonelyCat124: Removed.

src/psyclone/tests/domain/lfric/lfric_types_test.py
LonelyCat124: Removed.

src/psyclone/psyir/transformations/mark_routine_for_gpu_mixin.py
LonelyCat124: Fixed the former I think.

There's no failing test with this, so I'm not sure why I added this unless it fixed some example, but I'm building those now to check - I'm a bit confused otherwise, maybe it was something in the integration tests that failed but I'm not sure. I guess its to allow something like `1_wp` when `wp` is imported, but I assumed we had a test for this but maybe not.

I can try to add one maybe if examples don't fail.

LonelyCat124: I added a test for this, but this is still not needed as PSyclone knows that the symbol is constant later either way.

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
LonelyCat124: Done

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Done

src/psyclone/psyir/symbols/symbol_table.py
LonelyCat124: Done

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Added here and fixed up tests, as we previously allowed this behaviour in tests.

LonelyCat124:
@sergisiso I think this is ready for another look - the only thing I've not resolved it the sympy_writer issue which I think we need @hiker s help to understand whats happening (and if TYPE_INFO check is correct or not to have there).

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -294,8 +295,10 @@ def _prepare_code_to_inline(
                     if isinstance(symbol.datatype, DataTypeSymbol):
                         symbols_to_bring_in.add(symbol.datatype)
                     elif hasattr(symbol.datatype, 'precision'):
-                        if isinstance(symbol.datatype.precision, Symbol):
-                            symbols_to_bring_in.add(symbol.datatype.precision)
+                        if isinstance(symbol.datatype.precision, DataNode):
```
sergisiso: I still think that we can remove the whole:
```
        if isinstance(symbol, DataSymbol):
            # DataTypes can reference other symbols
            if isinstance(symbol.datatype, DataTypeSymbol):
                symbols_to_bring_in.add(symbol.datatype)
            elif hasattr(symbol.datatype, 'precision'):
                if isinstance(symbol.datatype.precision, DataNode):
                    for ref in symbol.datatype.precision.walk(
                            Reference):
                        symbols_to_bring_in.add(ref.symbol)
```
It's true that removing it makes the `test_validate_local_routine` fail, but on closer inspection it's the test that is wrong. The new error message: "The target of 'call do_something(a)' is already module inlined." is more accurate than the previous error message.

I think the better error message is because this PR actually fixed the equality when there are kinds.

I would also update the test name to test_validate_already_existing_local_routine.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -357,6 +261,100 @@ def _reverse_map(reverse_dict, op_map):
...
+        if isinstance(precision, ScalarType.Precision):
+            # The precision information is not absolute so is either
+            # machine specific or is specified via the compiler. Fortran
+            # only distinguishes relative precision for single and double
+            # precision reals.
+            if precision == ScalarType.Precision.DOUBLE:
+                if fortrantype.lower() == "real":
+                    return "double precision"
+                raise VisitorError(
+                    f"ScalarType.Precision,DOUBLE is not supported for "
```
sergisiso: Thanks for adding this as an actual error.

Dot in ScalarType.Precision.DOUBLE and single-quotes around {fortrantype} please.

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -449,6 +454,7 @@ def _create_type_map(self, list_of_expressions: Iterable[Node],
                     self._specialise_array_symbol(orig_sym, sva)
             else:
                 # A scalar access.
+                print(unique_sym.name)
```
sergisiso: Remove

src/psyclone/psyir/backend/sympy_writer.py
```diff
@@ -420,7 +421,11 @@ def _create_type_map(self, list_of_expressions: Iterable[Node],
                 # have a scope, hence the try...except.
                 orig_sym = sva[0].node.scope.symbol_table.lookup(sig.var_name)
             except SymbolError:
-                if isinstance(sva[0].node, Reference):
+                # Since ScalarType can contain references now, we need to
+                # skip over those used to define data kinds, as we don't
+                # need those here.
+                if (isinstance(sva[0].node, Reference) and
+                        sva[0].access_type != AccessType.TYPE_INFO):
```
sergisiso: Since we are not certain*, revert this change (unless it fails integration tests)

*My understanding is that this method creates a type_map for cases where we have symbol(parenthesis) expressions, because in sympy these symbols need to be declared as a symbol (array access) or function.

For this to matter here we need a symbol hitting both: having AccessType.TYPE_INFO and having parenthesis. So maybe something like these:
```
integer, parameter, dimension(2) :: precision = (/ 4, 8 /)
real(kind = precision(1)) :: value
```
or
` a = 1_precisions(2)` is this even valid Fortran?

So I do not think we should skip those. If you want to add such a test and it fails I would leave it with a TODO.

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
```diff
@@ -111,6 +111,10 @@ def infer_sharing_attributes(self) -> \
         for signature in var_accesses.all_signatures:
             if not var_accesses[signature].has_data_access():
                 continue
+            # Skip those that are TYPE_INFO accesses.
+            if any([x.access_type == AccessType.TYPE_INFO
+                    for x in var_accesses[signature]]):
```
sergisiso: pylint suggests removing the [ ]

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -419,25 +425,31 @@ def __init__(self, intrinsic, precision):
...
-                f"integer type but got: {precision}")
+                    not isinstance(dtype, UnresolvedType)):
+                raise ValueError(
+                    f"A DataNode representing the precision of another "
+                    f"DataSymbol must be of either 'unresolved' or "
+                    f"scalar, integer type but got: ScalarType with "
+                    f"datatype {dtype}")
+        # TODO #3135 If the precision is an int, then we would like to make
+        # a Literal containing it instead, however this is not currently
+        # possible due to circular imports.
```
sergisiso, line 452: Note that if you move the `src/psyclone/psyir/nodes/literal.py:46` inside the Literal init, then it is possible to add `from psyclone.psyir.nodes.literal import Literal` in this init.

(Some of the implementations below will be neater with this, but it is ok to leave this for another PR)

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -419,25 +425,31 @@ def __init__(self, intrinsic, precision):
...
+                f"DataNode, int or ScalarType.Precision, "
+                f"but found '{type(precision).__name__}'.")
         if isinstance(precision, int) and precision <= 0:
             raise ValueError(
                 f"The precision of a DataSymbol when specified as an integer "
                 f"number of bytes must be > 0 but found '{precision}'.")
-        if (isinstance(precision, DataSymbol) and
-                not (isinstance(precision.datatype, ScalarType) and
-                     precision.datatype.intrinsic ==
+        if (isinstance(precision, DataNode)):
```
sergisiso: pylint complains of superfluous outer parenthesis

src/psyclone/psyir/transformations/mark_routine_for_gpu_mixin.py
```diff
@@ -102,15 +102,15 @@ def validate_it_can_run_on_gpu(self, node, options):
...
+        # Check that the routine(s) do(es) not access any data that is
         # imported via a 'use' statement.
         for sched in kernel_schedules:
             vam = sched.reference_accesses()
             ktable = sched.symbol_table
             for sig in vam.all_signatures:
                 name = sig.var_name
                 first = vam[sig][0].node
-                if isinstance(first, (Symbol, DataType)):
+                if (isinstance(first, Reference)):
```
sergisiso: pylint complains of Unnecessary parens

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -661,13 +661,13 @@ def test_module_inline_apply_bring_in_non_local_symbols(
     # Also, if they are in datatype precision expressions
     psyir = fortran_reader.psyir_from_source('''
     module my_mod
-        use external_mod1, only: r_def
+        use external_mod1, only: r_def, kind_multiplier
```
sergisiso: Can you update to this essentially the same but more explicit version:
```
    # Also, if they are inside declaration expressions
    psyir = fortran_reader.psyir_from_source('''
    module my_mod
        use external_mod1, only: r_def
        use external_mod2, only: kind_multiplier
        use external_mod3, only: init_val
        use external_mod4, only: my_user_type
        use not_needed
        implicit none
        contains
        subroutine code()
            real(kind=r_def*kind_multiplier) :: a,b = init_val
            type(my_user_type) :: x
            a = b + x%data
        end subroutine code
    end module my_mod
    ''')

    new_routines = inline_trans._prepare_code_to_inline([routine])
    result = fortran_writer(new_routines[0])
    # The code_to_inline will contain the needed module imports, but
    # will ignore the non-used imports
    assert "use external_mod1, only : r_def" in result
    assert "use external_mod2, only : kind_multiplier" in result
    assert "use external_mod3, only : init_val" in result
    assert "use external_mod4, only : my_user_type" in result
    assert "use not_needed" not in result

```

sergisiso (changes requested):
@LonelyCat124 There is nothing new other than a couple of pylint suggestions, the rest are just the continuation of previous conversations that I brought down here as it was hard to keep track of them having to expand the history.

After this, can you bring it to master and launch the integration tests, as I expect these would  be the last comments.

src/psyclone/psyir/backend/sympy_writer.py
LonelyCat124: This is allowed:
```
program square

integer, parameter, dimension(2) :: precision = (/ 4, 8 /)
real(kind = precision(1)) :: value

value = real(1.0, kind=precision(1))
end program square
```

But gfortran would not compile `1.0_precision(1)`.

sergisiso: This can also happen in inquiry intrinsics, e.g. `value = TINY(precision(1))` which I believe it valid, so I still think the `sva[0].access_type != AccessType.TYPE_INFO` should not be there

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Fixed here and tests.

src/psyclone/psyir/backend/sympy_writer.py
LonelyCat124: Removed

src/psyclone/psyir/backend/sympy_writer.py
LonelyCat124: Ok, I'll remove the TYPE_INFO check.

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
LonelyCat124: Passes tests locally without so I've removed it - I assumed I needed `[]` for the list comprehension so just some syntax I don't know I think

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: I could do it, but once I created a `Literal` from inside `ScalarType` I hit circular imports at that stage, but maybe there is some way but I will leave it for another PR I think.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Removed.

src/psyclone/psyir/transformations/mark_routine_for_gpu_mixin.py
LonelyCat124: Removed.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
LonelyCat124: Swapped.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
LonelyCat124: Done.

LonelyCat124:
@sergisiso I think this is ready for another look - the only missing coverage is the LFRic code that is untested on master I think.

sergisiso:
@LonelyCat124 Unfortunately the very last merged PR also touched declarations and has conflicts, can resolve these and launch the integration once again please.

LonelyCat124:
@sergisiso Ready for review now - I'll set integration running.

Issue #3049, opened 2025-07-11
Old Fortran code requires `free_form=False`


schreiberx:
We need to be able to disable the free_form for reading in some old Fortran files. This is required for old legacy code from Croco with an example attached to this issue.

Please remove `.txt` extension:
[ana_grid.pre.f.txt](https://github.com/user-attachments/files/21182376/ana_grid.pre.f.txt)

schreiberx:
And another thing we'd need in the optimization script handed over to psyclone via `-s` is to get the list of input files to decide whether it's a free-form or not.

PR #3118, opened 2025-09-01
(Towards #3049) First draft of supporting free-form option.

LonelyCat124:
Added a free-form option (to enable fixed form support when set to False) to the PSyclone script. I semi-added it to the APIs, but I'm not sure thats actually worthwhile?

I'm not sure what @schreiberx was wanting for the comment on #3049, so there's nothing towards that here.

Ready for a first look @arporter @sergisiso @hiker

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,9 @@
+   101) PR #3118 towards #3049. Adds the --fixed-form and --free-form flags
+   to the psyclone command. Also automatically determine Fortran format
+   from file extension if they are not supplied.
+
+Release 3.2.0 1st of October 2025
+
    100) PR #3166 for #3120. Add GitHub release workflow and prepare for
    release 3.2.
 
@@ -17,10 +23,10 @@
    94) PR #3095 for #2916. Dynamically propagate pure and elemental
    attributes to GenericInterefaceSymbols.
 
-   93) PR #3147 for #3138. Fixes reference_accesses() for Calls.   
+   93) PR #3147 for #3138. Fixes reference_accesses() for Calls.
 
    92) PR #3133 for #3132. Give specific paths to pytest compilation (
-   to prevent using the unpredictable '.' path) 
+   to prevent using the unpredictable '.' path).
 
    91) PR #3137 for #3096. Add command-flag and config option to disable
    all backend indentations.
@@ -32,7 +38,7 @@
    IAttr constructor (which represents intrinsic calls).
 
    88) PR #3094 for #2851 and #3082. Improves the DA so that PSyclone
-   does not generate unnecessary firstprivate clauses on parallel regions.   
+   does not generate unnecessary firstprivate clauses on parallel regions.
 
    87) PR #3127 for #3126. Disable python module index in latex version
    of docs.


doc/user_guide/psyclone_command.rst
```diff
@@ -107,6 +107,8 @@ by the command:
      --keep-comments       keeps comments from the original code (defaults to False).
                            Directives are not kept with this option (use --keep-directives).
      --keep-directives     keeps directives from the original code (defaults to False).
+     --free-form FREE_FORM
```
arporter: Please could you also extend the section on "Basic Use" to say that, by default, PSyclone assumes the provided file is free-format and that if you have fixed-format, you need to supply the `--free-form=false` option.

arporter: We should also say that, no matter what the input, the output is always free-form Fortran.

src/psyclone/tests/generator_test.py
```diff
@@ -1161,6 +1161,30 @@ def trans(psyir):
...
+c     Comment here.
+      integer n
+
+      n = 3 +
+     &4
+      end subroutine'''
+    inputfile = str(tmpdir.join("fixed_form.f90"))
+    with open(inputfile, "w", encoding='utf-8') as my_file:
+        my_file.write(code)
+    main([inputfile, "--free-form", False])
```
arporter: Does this fail without the "--free-form" flag? Probably worth extending this test to check that.

doc/user_guide/psyclone_command.rst
sergisiso: I am confused, if this is a boolean flag, why does it not generate a help output like the previous `[--keep-comments] [--keep-directives]` and instead seems expects a FREE_FORM value?

LonelyCat124: Those flags are `off` unless set, and there didn't seem to be a `--no-keep-comments` way to do it, its either present or not. Since `free-form` is default on and you set it to false I decided to do it like this.

I could instead do `--enable-fixed-form` as the option to turn on fixed form parsing (or other advice on using argparse better) if you'd prefer @sergisiso ?

sergisiso: Why not just `--fixed--form`?

schreiberx: I wonder whether we could take a different approach here, which might solve the mix-match of files in free and fixed forms.
How about providing a file name ending (or even better, a comma-separated list of file name endings) which should be processed as free-form, and another list to be processed as fixed form?

LonelyCat124: I personally just prefer the option being fully explanatory, but am happy with either.

sergisiso: > How about providing a file name ending (or even better, a comma-separated list of file name endings) which should be processed as free-form

I think it is the responsibility of the build system if the decision is that complex (*I know things like to modulemanager are called directly by psyclone and not the build system).

What about file extensions, gfortran says:
`Source files with .f, .for, .fpp, .ftn, .F, .FOR, .FPP, and .FTN extensions are treated as fixed form. Source files with .f90, .f95, .f03, .f08, .F90, .F95, .F03 and .F08 extensions are treated as free form.`
would that work for your use case @schreiberx

schreiberx: I'd see this as very suitable default values, and if they could be overwritten/set on the command line, this would probably solve it for all cases I'm aware of.

LonelyCat124: I will add `--fixed-form-extensions` and `--free-form-extensions` then as options (with the defaults Sergi suggests)? I'm not sure if/how best to make these accessible to thinks like the modulemanager if needed but I will have a look.

schreiberx: Sounds great to me.
(Alternatively, also just one option, e.g., `--fixed-form-extensions` would work and whatever doesn't match this one is treated as free-form. Whatever you decide upon, it will likely work.)
Cheers!

sergisiso: > @schreiberx is the code you have not following those defaults? Otherwise I don't see why these need to be > configurable. The idea would be that with no flag we use these defaults and then the 
> "--free- form=false/true" for the eventual file that is different (now we need the value because they are distinc of default).

I agree - I don't think we want to go to the trouble of adding extra complexity unless we have to.

schreiberx: It seems that it follows this naming convention, but I can't say this definitely.
We could give it a try once you have a first version, and then we'll see.

src/psyclone/tests/generator_test.py
LonelyCat124: Added

src/psyclone/generator.py
```diff
@@ -104,6 +104,11 @@
               logging.getLevelName(logging.ERROR): logging.ERROR,
               logging.getLevelName(logging.CRITICAL): logging.CRITICAL}
 
+# Default file extensions as defined by gfortran.
```
arporter: Picky but probably best to say that we've just chosen to follow what gfortran does. Otherwise this could be read as meaning we are somehow coupled to gfortran.

src/psyclone/generator.py
```diff
@@ -540,6 +548,12 @@ def main(arguments):
         help="keeps directives from the original code (defaults to False)."
     )
 
+    parser.add_argument(
+        "--free-form", default=argparse.SUPPRESS, type=bool,
+        help="defines whether the original code is free form Fortran "
+             "(defaults to looking at the input file extension)."
```
arporter: "default behaviour is to examine the input file extension" which reminds me that we'll need to document this.

doc/user_guide/psyclone_command.rst
arporter: We need to put a bit more here - see my very first two comments. Also, we will now need to specify the mappings we use to determine whether or not a file is fixed/free format if the user doesn't specify. You may be able to use Sphinx auto-something to pull it in from the source but don't worry if you can't - it's only a couple of lines after all.

arporter: I'm also thinking that Sergi's suggestion of `--fixed-form` and `--free-form` arguments might be slightly nicer and more in-keeping with some of our other arguments.

src/psyclone/generator.py
arporter: If you change this to use `#: blah blah` then Sphinx may pick it up.

src/psyclone/generator.py
```diff
@@ -620,13 +634,30 @@ def main(arguments):
                        "PSyclone enabled keep_comments.")
         args.keep_comments = True
 
+    # If free_form is set in the arguments then it overrides default
+    # behaviour.
+    if "free_form" in args:
+        free_form = args.free_form
+    else:
+        fname = args.filename
+        if fname.endswith(FIXED_FORM):
```
arporter, line 657: Ooh, I didn't know you could provide a list of endings.

src/psyclone/generator.py
```diff
@@ -620,13 +634,30 @@ def main(arguments):
...
+            free_form = False
+        else:
+            free_form = True
+            if not fname.endswith(FREE_FORM):
+                logger.info(
+                    "Filename doesn't end with a known file extension. "
```
arporter: Good idea logging this. Please include the actual filename in the log message.

src/psyclone/tests/generator_test.py
```diff
@@ -1161,6 +1161,72 @@ def trans(psyir):
...
+    correct = """subroutine test()
+  integer :: n
+
+  n = 3 + 4
+
+end subroutine test"""
+    assert correct in captured
+
+    # Check an unknown file extension gives a log message and fails for a
+    # fixed form input.
```
arporter, line 1250: I think you've missed the check on the log message?

src/psyclone/generator.py
```diff
@@ -812,7 +846,8 @@ def code_transformation_mode(input_file, recipe_file, output_file,
         # Parse file
         reader = FortranReader(resolve_modules=resolve_mods,
                                ignore_comments=not keep_comments,
-                               ignore_directives=not keep_directives)
+                               ignore_directives=not keep_directives,
+                               free_form=free_form)
         try:
             psyir = reader.psyir_from_file(input_file)
         except (InternalError, ValueError, IOError) as err:
```
arporter, line 867: At this point, I think it would be good to include in the error message whether we were using/assuming fixed or free form formatting

src/psyclone/tests/generator_test.py
```diff
@@ -1161,6 +1161,72 @@ def trans(psyir):
...
+end subroutine test"""
+    assert correct in captured
+
+    with pytest.raises(SystemExit) as error:
+        main([inputfile])
+    with open(inputfile, "w", encoding='utf-8') as my_file:
+        my_file.write(code)
+    assert error.value.code == 1
+    out, err = capsys.readouterr()
+    assert ("Failed to create PSyIR from file " in err)
```
arporter, line 1222: As mentioned above, I think it would be good to include whether we were expecting free or fixed format.

src/psyclone/generator.py
LonelyCat124: Ok - I think I also need to add xu90 for UKCA

doc/user_guide/psyclone_command.rst
LonelyCat124: Added both arguments back.

src/psyclone/generator.py
LonelyCat124: Added - I'm not sure what docs output would be/where it goes though, but I'll try to find it.

src/psyclone/generator.py
LonelyCat124: Added

src/psyclone/tests/generator_test.py
LonelyCat124: Added

src/psyclone/generator.py
LonelyCat124: Added

src/psyclone/tests/generator_test.py
LonelyCat124: Added.

src/psyclone/generator.py
LonelyCat124: Added

LonelyCat124:
@arporter I think I addressed all the comments and merged in master so back to you

doc/user_guide/psyclone_command.rst
```diff
@@ -231,6 +234,36 @@ this. This first reason is that most compilers are able to cope with
...
+    - Free format extensions: .f90, .f95, .f03, .f08, .F90, .F95, .F03, .F08
+    - Fixed format extensions: .f, .for, .fpp, .ftn, .F, .FOR, .FPP, .FTN
+
+PSyclone also recognises the following extensions as free format to support
+current uses cases: .x90, .xu90
+
+PSyclone also provides the ``--free-form`` and ``-fixed-form`` to override the
+default behaviour, and will use the specified option over the file extension.
+
+If the file extension is not defined, PSyclone defaults to assuming the input
```
arporter: "...is not one of the ones listed above and neither of the ``--free-form``/``--fixed-form`` flags is used then PSyclone defaults..."

doc/user_guide/psyclone_command.rst
```diff
@@ -107,8 +108,10 @@ by the command:
      --keep-comments       keeps comments from the original code (defaults to False).
                            Directives are not kept with this option (use --keep-directives).
      --keep-directives     keeps directives from the original code (defaults to False).
-     --free-form FREE_FORM
-                           defines whether the original code is free form Fortran (defaults to True).
+     --free-form           forces PSyclone to parse this file as free format
+                           (defaults to looking at the input file extension).
```
arporter: suggest: "default is to look at the input file ..."

doc/user_guide/psyclone_command.rst
```diff
@@ -107,8 +108,10 @@ by the command:
...
+     --fixed-form          forces PSyclone to parse this file as fixed format
+                           (defaults to looking at the input file extension).
```
arporter: "default is to look at the input..."

src/psyclone/tests/generator_test.py
```diff
@@ -1161,6 +1167,27 @@ def trans(psyir):
...
 
 
+def test_code_transformation_free_form(tmpdir, capsys):
+    '''Test that the free-form option works for code transformation.'''
+    code = '''
+    subroutine test
+    integer :: n
+    n = 3 + 4
+    end subroutine'''
+    inputfile = str(tmpdir.join("free_form.f90"))
```
arporter: We would default to treating this as free form anyway so perhaps change the file extension to ".f" with a comment.

src/psyclone/generator.py
```diff
@@ -634,10 +645,12 @@ def main(arguments):
                        "PSyclone enabled keep_comments.")
         args.keep_comments = True
 
-    # If free_form is set in the arguments then it overrides default
-    # behaviour.
+    # If free_form or fixed_form is set in the arguments then it overrides
+    # default behaviour.
```
arporter, line 650: Sorry to be a pain but, just in case it comes back to bite us, please could you add a check (and test) that the user hasn't accidentally supplied both --free-form and --fixed-form flags.

src/psyclone/generator.py
```diff
@@ -646,8 +659,8 @@ def main(arguments):
             free_form = True
             if not fname.endswith(FREE_FORM):
                 logger.info(
-                    "Filename doesn't end with a known file extension. "
-                    "Assuming free form."
+                    f"Filename '{fname}' doesn't end with a known file "
```
arporter: s/known/recognised/

arporter (changes requested):
Nice, thanks for adding the new documentation. I've just been a bit picky (no surprise) about the wording - I confess this is partly because I've changed my mind and partly because we now have two separate command-line arguments. I think we also need to protect against a user supplying both --fixed and --free arguments.
Once that's done, this is ready to go.

doc/user_guide/psyclone_command.rst
LonelyCat124: Done

doc/user_guide/psyclone_command.rst
LonelyCat124: Done here and in generator.py

(As an aside is there no way could generate this docstring automatically?)

doc/user_guide/psyclone_command.rst
LonelyCat124: Done here and in generator.py

src/psyclone/tests/generator_test.py
LonelyCat124: Yeah thats my mistake, I'd got confused - I meant to use a fixed format extension - have added a comment explaining why as well.

src/psyclone/generator.py
LonelyCat124: Done* (waiting on teams message)

src/psyclone/generator.py
LonelyCat124: Fixed, test updated.

doc/user_guide/psyclone_command.rst
```diff
@@ -260,8 +260,9 @@ current uses cases: .x90, .xu90
 PSyclone also provides the ``--free-form`` and ``-fixed-form`` to override the
 default behaviour, and will use the specified option over the file extension.
 
-If the file extension is not defined, PSyclone defaults to assuming the input
-source is free form Fortran.
+If the file extension is not one of the ones listed above, and neither of the
+``free-form`` or ``--fixed-form`` flags is used then PSyclone defaults to
```
arporter: Missing a `--` but I'll do that if there's nothing else.

arporter (approved):
All good now, thanks very much Aidan.
I've just taken the liberty of putting back a missing "--" on one of the flags in the docs.

doc/user_guide/psyclone_command.rst
arporter: Done.

PR #3119, opened 2025-09-02
(Closes #3060) intrinsic return types and reference accesses

LonelyCat124:
I'm putting this up early so if anyone has time to have a quick look at the implementation before I go too far down the rabbit hole with how I'm implementing the return type and tests.

I have added two more members to the `IAttr` namedtuple, `return_type` and `reference_accesses`, which can either be a `Callable` or specific value (depending on whats required).

To get the return_type of an `IntrinsicCall`, my plan is to do something like:
```
def return_type(self):
    if isinstance(self.intrinsic.return_type, Callable):
        try:
            return self.intrinsic.return_type(self)
         except:
             # The idea here is to handle all of the "bad input cases", e.g. if we have an Unresolved or UnsupportedType in the input
             # I'd rather have a except here than handle it in every single return_type callable.
             return UnresolvedType()
    # If its not a callable we just return the value.
    return self.intrinsic.return_type
```

The return type implementations are started - there are 3 helper functions at the moment (for cases I expect to be used a lot), e.g. `_get_first_argument_type`, wheras other's have their own lambda (for example see `AINT`).

I'm unsure how much to avoid code duplication here, for example AINT and ANINT have the same lambda for their return_type, so I'm not sure whether its worth moving this out (and whether the result should be a lambda or function) every time I have any 2 intrinsic calls with the same return type? Feedback on this specific question would be appreciated as early as possible (probably one for @arporter to answer perhaps).

To test the return types, my plan was to have standalone test for every "helper" function (or even helper lambda later).
I was then planning to create a parametrize test for all other intrinsics who have their own specific lambda. My one concern is this parametrize would become very large - again feedback/thoughts on this approach would be helpful. You can see an initial versoin of how this parametrize might look at `intrinsic_call_test::650`.

PR #3125, opened 2025-09-08
(3060 reference accesses

LonelyCat124:


LonelyCat124:
Moved this up to date with latest master and #3119, so now its ready to be updated using argument names instead of any indices.

LonelyCat124:
@arporter @sergisiso This is ready for review now.

src/psyclone/tests/psyir/nodes/acc_directives_test.py
```diff
@@ -688,6 +688,8 @@ def test_accdatadirective_update_data_movement_clauses(fortran_reader,
         "  sto_tmp(ji) = sfactor * sto_tmp(ji)\n"
         "end do\n"
         "end program dtype_read\n")
+    print(psyir.view())
+    print(psyir.debug_string())
```
sergisiso: Delete

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -935,13 +935,14 @@ def test_infer_sharing_attributes(fortran_reader):
     assert list(sync)[0].name == 'k'
 
     # Check that kinds on literals are ignored for data sharing clauses
+    # and that if any access to a varible is CONSTANT it is ignored.
```
sergisiso: kinds and CONSTANTS are the same now, maybe the comment should be:
"Check that constants are ignored, even if in the same loop they appear in a location that expect any kind of READ"

src/psyclone/tests/psyir/transformations/omp_target_trans_test.py
```diff
@@ -445,24 +445,24 @@ def test_omptarget_nowait_multiple_dependencies(fortran_reader,
 
 
 def test_kind_parameters_ignored(fortran_reader):
-    '''Test that kind parameters used inside REAL calls don't
-    result in dependencies.'''
-    # TODO #3060: Fixing 3060 should avoid this being necessary.
+    '''Test that CONSTANT variables  used in locations that would also
```
sergisiso: Remove double-whitespace in the middle

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -146,7 +147,7 @@ def test_accparalleltrans_validate(fortran_reader):
         end do
         do i = 1, 10
             do j = 1, 10
-                A(i,j) = GET_COMMAND(2)
+                A(i,j) = GET_COMMAND(command)
```
sergisiso: Why did this make a difference here?

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -173,6 +173,20 @@ def test_intrinsiccall_datatype(fortran_reader):
     assert isinstance(call.datatype, UnresolvedType)
 
 
+def test_intrinsiccall_reference_accesses_error():
+    """Tests the reference_accesses method of intrinsiccall's error
+    case."""
```
sergisiso: This reads a bit strange to me, maybe: "Thest the error case of intrinsicall's reference_accesses method"

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -173,6 +173,20 @@ def test_intrinsiccall_datatype(fortran_reader):
...
+    # Test computing argument names doesn't work when we have 2 arguments for
+    # SUM with no naming, as it can't determine between the SUM variants.
```
sergisiso: "If the intrinsic call cannot compute the argument names (e.g. for SUM with no naming), it cannot guarantee the result of reference_accesses"

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
```diff
@@ -173,6 +173,20 @@ def test_intrinsiccall_datatype(fortran_reader):
...
+    intrinsic = IntrinsicCall(IntrinsicCall.Intrinsic.SUM)
+    intrinsic.addchild(Reference(DataSymbol("a", INTEGER_TYPE)))
+    intrinsic.addchild(Reference(DataSymbol("a", INTEGER_TYPE)))
+    with pytest.raises(InternalError) as err:
+        _ = intrinsic.reference_accesses()
+    assert ("Can't compute reference accesses for 'SUM(a, a)' due to not "
+            "being able to resolve all the argument names." in str(err.value))
```
sergisiso: This doesn't seem a far fetched case, I am not sure about generating an InternalError. Couldn't we just give all arguments a READWRITE? Maybe we can log a warning if you want to keep track of these.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,230 @@
...
+    """Adds a read access for argument into var_acc_map
+
+    :param argument: The argument to add a read access for.
+    :param var_acc_map: The VariablesAccessMap to add the access into.
+    """
+    # Find all the reference children of the argument (e.g. we could
+    # have a binary operation argument with multiple, or a Literal
+    # with none.
+    for ref in argument.walk(Reference):
+        sig, all_indices = ref.get_signature_and_indices()
```
sergisiso: I believe this should be some kind of recursive method calling reference_accesses instead of a walk. For instance if we do:
``` result = sum(a + 1_wp + func_that_rw_to_arg(b)) ```
I expect a is READ but wp is CONSTANT and b is READWRITE (otherwise we miss a dependency) and func_that_rw_to_arg is CALL.

Can you add that test to see what happens?

Maybe the logic is:
```
for child in argument.children:
   if isinstance(ref, Reference):
        sig, _ = ref.get_signature_and_indices()
        var_acc_map.add_access(sig, AccessType.READ, ref)
    var_acc_map.update(child.reference_accesses())
```
(I am not sure because the reference may want to add itself looking at tis parent?)

File: src/psyclone/psyir/nodes/intrinsic_call.py [hunk 9]
```diff
@@ -87,6 +87,230 @@
...
+) -> VariablesAccessMap:
+    """General helper function for creating the reference_accesses for a
+    general IntrinsicCall.
+    This function only uses indices for non-named arguments, e.g. if
+    [0] is passed in as an index, but the argument at position 0 has a name
+    the function will skip it and will assume that the relevant _named_args
+    contains the name.
+
+    :param node: the IntrinsicCall whose reference_accesses to compute.
+    :type node:  :py:class:`psyclone.psyir.nodes.IntrinsicCall`
```
sergisiso: This can be an annotation with a string to forward-delcare

src/psyclone/tests/psyir/nodes/intrinsic_call_reference_accesses_test.py
```diff
@@ -0,0 +1,354 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2022-2025, Science and Technology Facilities Council.
```
sergisiso: 2026

src/psyclone/tests/psyir/nodes/intrinsic_call_reference_accesses_test.py
```diff
@@ -0,0 +1,354 @@
...
+    _add_inquiry_argument(aref, vam)
+    sig, _ = aref.get_signature_and_indices()
+    assert len(vam[sig]) == 1
+    assert vam[sig][0].access_type == AccessType.INQUIRY
+    sig, _ = ref.get_signature_and_indices()
+    assert len(vam[sig]) == 1
+    assert vam[sig][0].access_type == AccessType.READ
+
+
+# FIXME Test _compute_reference_accesses
```
sergisiso: Is this done?

src/psyclone/tests/psyir/nodes/intrinsic_call_reference_accesses_test.py
```diff
@@ -0,0 +1,354 @@
...
+            readwrite_indices=[2],
+            constant_indices=[3],
+            inquiry_indices=[4],
+            read_named_args=["read", "not_present_1"],
+            write_named_args=["write", "not_present_2"],
+            readwrite_named_args=["readwrite", "not_present_3"],
+            constant_named_args=["constant", "not_present_4"],
+            inquiry_named_args=["inquiry", "not_present_5"],
+    )
+    # We should onyl get the 10 accesses present in the Call.
```
sergisiso: "only"

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
LonelyCat124: We could do - I guess this is just giving the worst case scenario instead of the correct result? My only slight concern is that could result in `kind` values becoming READWRITE but I think we have exceptions for those anywhere we need so maybe its ok.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah I think you're right here, these are probably (maybe?) all wrong if/when an argument to an intrinsic is another intrinsic, I'll come back to this and have a look at it again.

LonelyCat124: @sergisiso I've been thinking about this, and for the read accesses I think I probably just want to do:
`argument.reference_accesses()`
as this will already default to read where it should for the argument (be it `BinaryOperation` or `Reference` or something else? (and then merge this with the input var_acc_map).And the other functions I think are correct.

Does that seem reasonable?

sergisiso: Only using `argument.reference_accesses()` for reads I think is correct. I will need to think more about the others. Adding a test with some complex examples of this will help.

LonelyCat124: The writes I'm confident in because they take a  `Reference` as an argument only.
The constant and inquiry ones take a `DataNode` supposedly, but I think its actually only Literal or Reference for both.

src/psyclone/tests/psyir/nodes/acc_directives_test.py
LonelyCat124: Done

src/psyclone/tests/psyir/nodes/omp_directives_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/transformations/omp_target_trans_test.py
LonelyCat124: Done

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: It doesn't, this is kind of a bad test (as its not actual fortran but an abuse of syntax really) but I've reverted it.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
LonelyCat124: Fixed.

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: I am not sure. Even for writes a reference can also be any kind of reference, like:
`intrinsic(argument_array(map_fn(a + 3_wp)))` this should see the map_fn as CALL and the wp as CONSTANT, and a as READ, even if argument_array is write.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Done.

src/psyclone/tests/psyir/nodes/intrinsic_call_reference_accesses_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/nodes/intrinsic_call_reference_accesses_test.py
LonelyCat124: Yeah

LonelyCat124: Removed.

src/psyclone/tests/psyir/nodes/intrinsic_call_reference_accesses_test.py
LonelyCat124: Fixed.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah but the write ones already do:
```
    # For Array accesses, these have reads to their indices
    for indices in all_indices:
        for index in indices:
            var_acc_map.update(index.reference_accesses())
```
so I use what the reference accesses tell me for the indices, I only force the reference itself to be a write/readwrite.

The remaining thing I'm unsure about is whether I should care about indices or not - probably the answer is yes? But I wasn't sure about an example that was valid yet.

LonelyCat124: I updated the comment to say accesses instead of reads in the above snippet too.

LonelyCat124: I'm adding a test for your case as well.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,234 @@
...
+    # For Array accesses, these have accesses to their indices
+    for indices in all_indices:
+        for index in indices:
+            var_acc_map.update(index.reference_accesses())
+    var_acc_map.add_access(sig, AccessType.INQUIRY, argument)
+
+
+def _compute_reference_accesses(
+    node: "IntrinsicCall",
+    read_indices: list[int] = None,
```
sergisiso: All these produce `Default value of type None is not assignable to annotated parameter type list[int]` with the static checker. To solve this I typically use:
`Iterable[int] = (),` which I find nicer than the Optional and accepts tuples and lists. Also you can remove the `if read_indices:` expressions in the body.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -617,7 +864,13 @@ class Intrinsic(IAttr, Enum):
                 arg_names=((None,),)),
             optional_args={},
             return_type=None,
-            reference_accesses=None,
+            reference_accesses=lambda node: (
+                _compute_reference_accesses(
+                    node,
+                    # All arguments are unamed and all are written to.
```
sergisiso: "unnamed"

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: "currently"

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -4285,25 +5159,22 @@ def reference_accesses(self) -> VariablesAccessMap:
             structure acccessors) and the values are AccessSequence
```
sergisiso: "accessors"

sergisiso: and "available" at line 5248 (has the fix_typos PR been merged already?, maybe it does those for you)

src/psyclone/tests/psyir/transformations/transformations_test.py
sergisiso: Oh, that was a good answer, I didn't know what it did but I see it expects a character and is it an output (so cannot be a literal), so I am happy with the change. Can we also remove the `a(i,j) =`  and replace it with `call`

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
sergisiso: I think we should. I don't think we can guarantee that PARAMETERS are always returned as CONSTANT anyway, and in call.py we already do the same worst case decision.

Also, if we allow it to raise, there are many things in the code that would technically require a try:except, so lets go with READWRITE for now.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,234 @@
...
+    # If the argument isn't a Reference then we don't do anything.
+    if isinstance(argument, Reference):
+        sig, _ = argument.get_signature_and_indices()
+        var_info = access_info[sig]
+        try:
+            var_info.change_read_to_constant()
+        except InternalError:
+            # The argument here is also used in some other way
+            # so we do nothing as the other usage has precedence.
+            pass
```
sergisiso: I am not a fan of this InternalError, is there any reason that the functions need to guarantee the starting value. Could they not be simply `\.change_to_constant()` and never raise?

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: I wanted to finish this review today, but I haven't considered this properly yet, I will come back to this or do it in the next review.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: Replacing it with call changes the error so that doesn't work - I will see if there's another function that uses a character string.

LonelyCat124: Swapped it to ADJUSTR

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah thats cleaner, fixed.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Fixed.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: I just copied the structure of the other functions here - I assume its safety to avoid it being called on the wrong things, e.g. if something is written elsewhere we can't change this to be constant as I think some of our other analysis stuff checks for `if(any(type_access == CONSTANT))` and this would break a lot of that code.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: I think this is fixed already? I can't find it/what it refers to.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: These were already fixed by that PR I think

src/psyclone/tests/psyir/nodes/intrinsic_call_test.py
LonelyCat124: It does READWRITEs for references and READS for any other nodes (since we can't write to a non-reference node argument, it doesn't make sense i think).

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: I have been thinking about these middle methods and they still don't make sense to me.

They not only make things more complicated: adding multiple layers, unexpected behaviours (e.g. fail if there is > 1 READ accesses), internal errors... And fail for simple things that should be handled, e.g. the assignment "g(g(1)) = 1" currently fails.

I believe the implementation would be easier if we just delete them and do a more explicit:
`access[the_specific_access_to_update].access_type = the_new_access_type`

I created #3300 to test this. The base of that PR is this branch, so if you agree you can click merge to bring the changes here.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,224 @@
...
+    # For indices, we only apply them if there is no argument name, othrwise
+    # it should be handled by the argument names.
+    for ind in read_indices:
+        if ind < len(node.arguments) and node.argument_names[ind] is None:
+            arg = node.arguments[ind]
+            _add_read_argument(arg, reference_accesses)
+    for ind in write_indices:
+        if ind < len(node.arguments) and node.argument_names[ind] is None:
+            arg = node.arguments[ind]
+            _add_write_argument(arg, reference_accesses)
```
sergisiso: I resolved the previous converstion because it was getting to long, but I still think it would be better to use the recursion (arg.reference_accesses()) that will properly handle all internal accesses, instead of thinking in each case what can or cannot happen in order to come up with the method implementation. Then all the add mehtods will be the same and can be implemented with a single method doing:
```
accesses = arg.reference_accesses()
if isinstance(arg, Reference):
   sig, _ = arg.get_signature_and_indices()
   accesses[sig][-1].access_type = access_type
```
with arg and access_type as a parameter. (note that I am using the setter from the proposed PR in the previous comment)

Or even inlining this expression.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,224 @@
...
+        accesses.
+    :param inquiry_named_args: a list of named arguments that are inquiry
+        accesses.
+
+    :returns: the reference accesses of node.
+    """
+    reference_accesses = VariablesAccessMap()
+    # For indices, we only apply them if there is no argument name, othrwise
+    # it should be handled by the argument names.
+    for ind in read_indices:
```
sergisiso: This would make the AccessSequence have a strange order instead of left-to-right arguments. Tbh I don't think there is a "right" order, but it may be neater to start with a:
`for ind, arg in enumerate(node.arguments)`
and then do each individual:
`if ind in read_indices` ....
`if node.argument_names[ind] in read_named_args` ...
to end up with a more sensible order

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Merged it in.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: I created a single function to do this and tests seemed ok with it.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Done.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -3680,7 +4428,7 @@ class Intrinsic(IAttr, Enum):
                 arg_names=()),
             optional_args={"team": DataNode},
             return_type=INTEGER_TYPE,
-            reference_accesses=None,
+            reference_accesses=None,  # FIXME NYI
```
sergisiso: Can we do this now? If not it needs a TODO, we don't allow bare FIXMEs to pass review

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -4285,25 +5062,33 @@ def reference_accesses(self) -> VariablesAccessMap:
...
+                for i, arg in enumerate(self.arguments):
+                    if isinstance(arg, Reference):
+                        readwrites.append(i)
+                    else:
+                        reads.append(i)
+                return _compute_reference_accesses(
+                        self,
+                        read_indices=reads,
+                        readwrite_indices=readwrites
+                )
```
sergisiso, line 5291: Is this still needed now that _compute_reference_accesses only does the top references and then recurses down? We could assign all assesses as readwirte and the internal references will already be given a read-only if it is part of an expression.

(see also next comment)

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,137 @@
...
+        accesses.
+    :param inquiry_named_args: a list of named arguments that are inquiry
+        accesses.
+
+    :returns: the reference accesses of node.
+    """
+    reference_accesses = VariablesAccessMap()
+    for ind, arg in enumerate(node.arguments):
+        # For indices, we only apply them if there is no argument name,
+        # otherwise it should be handled by the argument names.
```
sergisiso: I was going to say that this comment is in the wrong place, as this happens at the caller with expressions such as:
```
  write_indices=[x for x, i in enumerate(
      node.argument_names) if i is None
  ],
```
but then looking more generally how this is being used  the `*_indices` parameters are a bit strange because we require `self.compute_argument_names()` to succeed to give them accees_types. We only have to use it when we have some variadic argument and all such arguments have the same access. So maybe it would be easier to add a "default_access" instead of `*_indices`. Then, e.g. allocate, will look like:
```
_compute_reference_accesses(
   node,
   read_named_args=["mold", "source"],
   write_named_args=["stat", "errmsg"],
   default_access=AccessType.WRITE
)
```

It would also clarify some current questions about what happens if an index and a names are both specified with different values, or if a value is not specified at all.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -87,6 +87,137 @@
...
+            continue
+        if node.argument_names[ind] in readwrite_named_args:
+            _add_argument_of_access_type(arg, reference_accesses,
+                                         AccessType.READWRITE)
+            continue
+
+    return reference_accesses
+
+
+def _reference_accesses_all_reads_with_optional_kind(
```
sergisiso: If you agree with the previous comment, this could be replaced to calls to:
```
_compute_reference_accesses(
   node,
   constant_named_args=["kind"],
   default_access=AccessType.READ
)
```
(maybe even READ could be the default default_access to make this common case simpler)

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -3968,7 +4749,6 @@ def _find_matching_interface(self) -> Tuple[str]:
             ]
             # Remove any of the interfaces that don't contain
             # a named non-optional argument from the list of potential
-            # candidate interfaces.
```
sergisiso, line 3971: Was this removed on accident?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -4285,25 +5062,33 @@ def reference_accesses(self) -> VariablesAccessMap:
             structure accessors) and the values are AccessSequence
             (a sequence of AccessTypes).
 
+        :raises InternalError: if the IntrinsicCall can't have argument
+                               names computed due to ambiguity in the
+                               intrinsic's interface being used.
```
sergisiso: Is this a leftover from the previous implementation?

sergisiso (changes requested):
Almost there now @LonelyCat124, just one more possible simplification and small clean ups.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah, I'm not sure what happened with this one.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah, readded.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah, removed.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Done.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Done

PR #3127, opened 2025-09-08
(Closes #3126) disable python module index in latex version of docs

### LINKED ISSUE #3126: pdf of docs contains "Python Module Index" section
If you build the `pdflatex` target of the docs locally or look at RTD, you'll see we have:

Image

Presumably this is because we have disabled building the Reference Guide when building the PDF version of the docs. Therefore, we also need to try to disable the generation of this section.

arporter:
Also adds @LonelyCat124 as an author.


arporter:
Very small change to improve the pdf version of the documentation. Also adds Aidan as an author as he must have added a fair bit by now??

Issue #3123, opened 2025-09-04
Compilation problem for driver extraction


hiker:
I actually can't remember what compiler it was (nvidia??) and where it was brought up, but there is a "," missing in compare_variables_mod.F90, I still have the 'fixed' file around, and just realised that it was never merged in???

```
diff --git a/lib/extract/binary/lfric/compare_variables_mod.F90 b/lib/extract/binary/lfric/compare_variables_mod.F90
index 11f44eadc..73ddcd81e 100644
--- a/lib/extract/binary/lfric/compare_variables_mod.F90
+++ b/lib/extract/binary/lfric/compare_variables_mod.F90
@@ -134,7 +134,7 @@ contains
             "#rel<1E-9", "#rel<1E-6", "#rel<1E-3", "#rel>=1E-3", &
             "max_abs", "max_rel", "l2_diff", "l2_cos"
 
-        out_format = trim(out_format)//"' ',6(I12, ' '),4(E12.7,' '))"
+        out_format = trim(out_format)//",' ',6(I12, ' '),4(E12.7,' '))"
 
         ! Then write out the results for each variable:
         do i=1, current_index
```

PR #3128, opened 2025-09-10
#3123 Updated lfric file forgotten in 2878.

hiker:
A tiny fix for gfortran-15 compilation. The main issue was fixed in #2878, but I forgot to update the pre-processed file in lfric/binary (which is required for compilation tests).


arporter:
> I don't think that this is this PR's fault tbh.

Sergi and I agree :-)

PR #3129, opened 2025-09-10
#2709 Fixes #2709.

### LINKED ISSUE #2709: --backend disable-validation does not work for transformation mode
This flag is only used in PSyGen, which is not called in transformation mode . The following patch should fix this:
```
git diff generator.py
diff --git a/src/psyclone/generator.py b/src/psyclone/generator.py
index b405352d1..22fc8a628 100644
--- a/src/psyclone/generator.py
+++ b/src/psyclone/generator.py
@@ -729,7 +729,9 @@ def code_transformation_mode(input_file, recipe_file, output_file,
             Profiler.add_profile_nodes(routine, Loop)
 
         # Generate Fortran
-        output = FortranWriter()(psyir)
+        writer = FortranWriter(
+            check_global_constraints=Config.get().backend_checks_enabled)
+        output = writer(psyir)
         # Fix line_length if requested
         if line_length in ("output", "all"):
             output = fll.process(output)
```

hiker:
Very small change. I did some minor tidying up to the coding style.

### DIFF
--- a/src/psyclone/tests/generator_test.py
+++ b/src/psyclone/tests/generator_test.py
@@ -1169,7 +1169,7 @@ def test_code_transformation_backend_validation(validate: bool,
     the Fortran writer when using generic code transformations.
     '''
 
-    # Create a dummy Fortran writer, which we use the check
+    # Create a dummy Fortran writer, which we use to check
     # the values passed in
     def dummy_fortran_writer(check_global_constraints: bool,
                              disable_copy: bool):
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   86) PR #3129 for #2709. Ensures that the 'disable backends check'
+   configuration option is passed to the FortranWriter.
+
    85) PR #3128 for #3123. Fixes Fortran file in lfric binary extraction
    library so that it works with gfortran 15.
 


src/psyclone/tests/generator_test.py
```diff
@@ -1161,6 +1161,40 @@ def trans(psyir):
...
 
+@pytest.mark.parametrize("validate", [True, False])
+def test_code_transformation_backend_validation(validate: bool,
+                                                monkeypatch) -> None:
+    '''
+    Test that the backend validation flag is passed to
+    the Fortran writer when using generic code transformations.
+    '''
+
+    # Create a dummy Fortran writer, which we use the check
```
arporter: s/the/to/

src/psyclone/tests/generator_test.py
```diff
@@ -1161,6 +1161,40 @@ def trans(psyir):
...
+    # The input file doesn't really matter, so just use a
+    # kernel file from gocean:
+    input_file = Path(get_base_path("gocean")) / "test27_loop_swap.f90"
+
+    if validate:
+        option = "enable-validation"
+    else:
+        option = "disable-validation"
+    main([str(input_file), "--backend", option])
+    # The actual assert is in the dummy_fortran_writer function above
```
arporter, line 1195: Neat.

src/psyclone/tests/generator_test.py
arporter: I took the liberty of fixing this as it's just a comment.

PR #3131, opened 2025-09-12
2049 improved driver base class

hiker:
Move more common functionality into the driver creation base class.


hiker:
@sergisiso, when you have time, can you have a very quick look? My main question is: I admit I was lazy with the tests: I moved tests from `lfric_extract_driver_creator_test.py` to `base_extract_driver_creator_test.py`, till I had all lines covered, then made sure each of the remaining tests (lfric_extract_driver_creator and extract_driver_creator) still covered their corresponding source file to 100%.

'Proper' testing (i.e. setting up data structures so we can call each individual function) would take some serious effort :(

Are you happy to leave the tests the way they are atm, or should I see how I go with having individual tests for individual functions?

hiker:
CI is green now as well.

src/psyclone/domain/lfric/lfric_constants.py
```diff
@@ -56,7 +56,9 @@ class LFRicConstants():
     '''
     HAS_BEEN_INITIALISED = False
 
-    def __init__(self):
+    UTILITIES_MOD_MAP: dict[str, dict[str, str]]
```
sergisiso, line 61: Why is this a class variable while all other constants are instance variables?

examples/gocean/eg5/extract/Makefile
```diff
@@ -148,15 +148,18 @@ psy.o:	$(KERNELS)
 # The driver will be created when creating the psy.f90 file.
 # This dependency is required, otherwise a 'make compile' in a clean
 # directory will fail.
-$(DRIVER_INIT).f90: psy.f90
-$(DRIVER_UPDATE).f90: psy.f90
+$(DRIVER_INIT).F90: psy.f90
```
sergisiso, line 161: Why this change? I assume the driver will not have any preprocessing on it since it is generated by psyclone?

src/psyclone/domain/common/base_driver_creator.py
```diff
@@ -337,3 +347,351 @@ def import_modules(program: Routine):
...
+        module_dependencies = self.collect_all_required_modules(file_container)
+        # Sort the modules by dependencies, i.e. start with modules
+        # that have no dependency. This is required for compilation, the
+        # compiler must have found any dependent modules before it can
+        # compile a module.
+        mod_manager = ModuleManager.get()
+        sorted_modules = mod_manager.sort_modules(module_dependencies)
+
+        # Inline all required modules into the driver source file so that
+        # it is stand-alone.
```
sergisiso: I think this comment has to be moved before `module_dependencies = ...` to explain what this whole section is about

src/psyclone/domain/common/base_driver_creator.py
```diff
@@ -337,3 +347,351 @@ def import_modules(program: Routine):
             new_routine_sym = RoutineSymbol(routine.name, UnresolvedType(),
                                             interface=ImportInterface(module))
             symtab.add(new_routine_sym)
+
+    # -------------------------------------------------------------------------
+    def cleanup_psyir(self, extract_region: Node) -> None:
```
sergisiso: Maybe call it "verify_and_cleanup_psyir"?

src/psyclone/domain/common/base_driver_creator.py
```diff
@@ -337,3 +347,351 @@ def import_modules(program: Routine):
...
+
+        :raises ValueError: if a StructureReference outside of a call
+            is found.
+        """
+        for sref in extract_region.walk(StructureReference):
+            # DSLs use StructureReferences in calls (e.g in LFRic
+            # set_dirty etc), which should be accepted.
+            # So only flag StructureReferences that are not part of
+            # a call.
+            if not isinstance(sref.parent, Call):
```
sergisiso: Maybe this won't be necessary if the specialised method has first the removal of set_dirty/clean and then call the `super().cleanup_psyir ...`

src/psyclone/domain/common/base_driver_creator.py
```diff
@@ -337,3 +347,351 @@ def import_modules(program: Routine):
...
+                raise ValueError(f"The provided PSyIR should not have "
+                                 f"StructureReferences, but found: "
+                                 f"{sref.debug_string()}")
```
sergisiso: This error may be confusing if it reaches the end-user. The provided PSyIR (the extraction region) can have Structures, but these should already be flattened at this point. Maybe a more clear error message would be:

"The DriverCreator does not support StructureReferences, any such references in the extraction region should have been flattened by the ExtractNode, but found: '{sref.debug_string()}'"

src/psyclone/domain/common/driver_creator.py
```diff
@@ -337,3 +347,351 @@ def import_modules(program: Routine):
...
+        original_symbol_table = nodes[0].ancestor(Routine).symbol_table
+
+        # Copy the nodes that are part of the extraction
+        extract_region = nodes[0].copy()
+        self.cleanup_psyir(extract_region)
+        output_symbols = self._create_read_in_code(program, psy_data,
+                                                   original_symbol_table,
+                                                   read_write_info, postfix)
+
+        # Copy the nodes that are part of the extraction
```
sergisiso, line 599: The comment should be:  Add the modified extracted region into the driver program

src/psyclone/domain/common/base_driver_creator.py
```diff
@@ -337,3 +347,351 @@ def import_modules(program: Routine):
...
+                                                   read_write_info, postfix)
+
+        # Copy the nodes that are part of the extraction
+        program.children.extend(extract_region.pop_all_children())
+
+        # Find all imported modules and add them to the symbol table
+        self.import_modules(program)
+        self.handle_precision_symbols(program.scope.symbol_table)
+
+        BaseDriverCreator.add_result_tests(program, output_symbols)
```
sergisiso: `self.`?

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
```diff
@@ -162,25 +152,22 @@ class LFRicExtractDriverCreator(BaseDriverCreator):
 
     :param region_name: the suggested region_name.
     '''
-    def __init__(self, region_name: str = None):
```
sergisiso, line 165: Several part of this docstring are outdated, and most are not specific to LFRic anymore. I suggest removing it, or making an up-to-date version in the Base class.

src/psyclone/parse/module_manager.py
```diff
@@ -120,34 +121,41 @@ def __init__(
...
 
-        self._visited_files = {}
+        self._visited_files: dict[str, FileInfo] = {}
 
         # The list of all search paths which have not yet all their files
         # checked. It is stored as an ordered dict to make it easier to avoid
-        # duplicating entries.
-        self._remaining_search_paths = OrderedDict()
-        self._original_search_paths = []
+        # duplicating entries, but it only stores a 1 as 'value'.
```
sergisiso: I don't understand this addition

src/psyclone/parse/module_manager.py
```diff
@@ -203,6 +210,9 @@ def _add_all_files_from_dir(self, directory):
                 full_path = os.path.join(directory, entry.name)
                 if full_path in self._visited_files:
                     continue
+                # Check if the full path matches an ignore pattern:
```
sergisiso: s/matches/contains/

examples/gocean/eg5/extract/Makefile
hiker: That was already in the driver creator, it just didn't affect gocean in the past. Now that most of the gocean functionality is in the base class, this changed the file name.

TBH, I can't exactly remember why it was required (and git history only indicates that I did this when starting to support inlining modules). 

Note that the including of modules is done from the source code only (not from a parsed tree), so this will handle files with preprocessor directives correctly. If a driver needs a file with preprocessor directives, the file can't be parsed (resulting in a warning atm since the static call tree analysis cannot work on it), but the file will be correctly included. I vaguely remember that I had files with MPI conditional events in it, but not entirely sure where this came from. It might be that this was with the old/current build system, and I picked up un-preprocessed files??

src/psyclone/domain/common/driver_creator.py
hiker: Done

src/psyclone/domain/lfric/lfric_constants.py
hiker: The other variables are (all? most?) class variables. I had to add the declaration here for typing purpose only, to make the lfric_driver_extraction pass `mypy`. And fixing typing snowballs, so I stopped at the one change I needed.

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
hiker: I have removed this, and added a new doc string to the base class. It's not as detailed as previously, but at least we have some documentation still.

src/psyclone/domain/common/base_driver_creator.py
hiker: Done.

src/psyclone/domain/common/base_driver_creator.py
hiker: Good suggestion.

src/psyclone/domain/common/base_driver_creator.py
hiker: Oh, good idea. I remember I was fiddling a bit with testing this, this is much cleaner now.

src/psyclone/domain/common/base_driver_creator.py
hiker: Done.

src/psyclone/domain/common/base_driver_creator.py
hiker: Done (tbh, I don't like it. `add_result_tests` is a static function. If it needs to be overwritten, it should be properly declared with `self`, and only then called using `self`). But I am aware from other reviews that `self` is preferred here, even for static functions.

src/psyclone/parse/module_manager.py
hiker: I've reworded it. I wanted to document why it is a dict (and not a set), hope it's clearer now (without explanation it is odd that there is a dict, but the value is not used, or explained)

src/psyclone/parse/module_manager.py
hiker: Done.

hiker:
It's working again, so ready for next review (but as discussed, not important for the September release).

src/psyclone/domain/lfric/lfric_constants.py
sergisiso: Yes sorry, I just saw that they were declared inside the constructor, but you are right they are class variables. I guess the follow up question is why are they declared inside the constructor?

At least for the one you move out, why don't be have the line 493 on top of line 59, with the comment being a '#:' so it is properly documented in sphinx.

(but I see this was just a typechecking change and not the main think of this PR, so I will be happy either way)

examples/gocean/eg5/extract/Makefile
sergisiso: Ok, I am fine with this explanation.

Not related to this example, but the lfric one creates a lot of files in `examples/lfric/scripts/driver-*.F90` that the gitignore does not ignore. Could you add them to the relevant gitingore please

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
sergisiso: Thanks for cleaning this up. Only small follow-up is that the new point 4 has a double-dot at the end.

sergisiso (changes requested):
@hiker Thanks for addressing the comments, there are just minor follow ups. Also, should we do the renaming of classes/filenames now? I think something like:
BaseDriverCreator -> DriverCreator
ExtractDriverCreator -> GoceanDriverCreator
LFRicExtractDriverCreator -> LFricDriverCreator
would make sense.

examples/gocean/eg5/extract/Makefile
hiker: I've added a .gitignore.

src/psyclone/domain/lfric/lfric_constants.py
hiker: Done (I didn't know about `#:`, but it seems to have worked :)
Note that I left the same comment in both places, since it is just very far away.

The 'advantage' (?) of setting the values in the constructor is that importing the module does not trigger the evaluation (in case that there would be a dependencies on other modules). I don't think there is actually, as far as I can see the constructor just sets constant values.

src/psyclone/domain/lfric/lfric_extract_driver_creator.py
hiker: Fixed, thanks!

hiker:
I had run the CI earlier (and it was all green), but since then I've fixed one documentation issue (old link), and just merged in current master. I won't trigger the CI now, so you all have access to it if you need it for anything else.

The renaming is also all done, so this PR is ready for the next review.

sergisiso (approved):
Thanks for doing the renaming @hiker . This is now ready to merge

PR #3133, opened 2025-09-12
#3132 Properly test build classes for LFRic and GOcean

### LINKED ISSUE #3132: Incorrect test, still working on github CI
The file lfric_build_test.py contains:
```
def test_lf_build_get_infrastructure_flags(monkeypatch, tmpdir):
...
    builder = LFRicBuild(tmpdir)
...
    assert 'configuration' in dir_list
    assert 'function_space' in dir_list
    assert 'field' in dir_list
```
The builder is searching in the wrong directory (`tests`), so it should fail (it worked in the past because `tests/test_files/lfric/infrastructure` was found under `tests`).

While the fix is easy, the question is: why is this test passing on github CI (since it shouldn't have left-over files like  .mod or .o in the old infrastructure directory, which can happens to devs). This test must be improved (so it fails), to ensure it is testing the right thing :) And then the builder needs to be fixed.

hiker:
Trying to fix #3132 - first of all why the test is not failing on github :)

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   92) PR #3133 for #3132. Give specific paths to pytest compilation (
+   to prevent using the unpredictable '.' path) 
+
    91) PR #3137 for #3096. Add command-flag and config option to disable
    all backend indentations.
 


hiker:
A few related issues here. First of all, the existing (at least) LFric build test tested the wrong flags (it would use ANY directory from PSyclone as include flags for LFRic, and the test only passed because all directories included the new external/lfric-infrastructure directories. I've added test to make sure no unexpected directories are returned.

Similarly, the gocean build test did not test the directory at all (and de-fact it returned just `-I` without path as flags). I've fixed that as well.

Then I also updated the tests to have 100% coverage of `lfric_build.py` and `gocean_build.py`, independent if the tests are running with or without compilation. I also added typing.

Then ... I discovered we had actually two `lfric_build.py` files, one in `tests`, one in `tests/domain/lfric`. They seem to be related, but still quite different. The file in `tests/domain/lfric` does not have 100% coverage when compilation is enabled (missing lines 106 and 120 of lfric_build.py). My suggestion: once the review here is done we replace `tests/domain/lfric/lfric_build_test.py` with the modified/improved file `tests/lfric_build_test.py`. In hindsight, it would likely have been easier to update `domain/lfric/lfric_build` .. but too late ;)

Triggering CI now.

hiker:
In spite of the missing coverage, I actually believe this is ready for a review. There are two sections not covered, and those lines are in the tests only, i.e. some tests are not executed. In detail:
1. One line that is only executed if test-compilation is enabled, to cover the non-compilation code when compilation is enabled. Since the github CI does not compile. that line is never executed. But all the lines this test is supposed to cover are actually executes.
2. There is a test to check that a `chdir` is *not* executed (it would be executed if compilation is triggered, and I added this test to reach 100% coverage of the tested class). So it is correct that the exception is not triggered (if the exception is triggered, it does a pytest.fail.

I am happy to use different implementation if that helps.

Otherwise: this PR is low priority, it just fixes properly the compilation tests (which were quite broken before), ensures that the LFRic and GOcean compilation classes are 100% covered, either with compilation enabled or disables, and some coding style improvements.

src/psyclone/tests/domain/gocean/gocean_build_test.py
```diff
@@ -48,30 +50,78 @@
...
+        monkeypatch.setattr(GOceanBuild, "_compilation_path",
+                            "/tmp/path/dl_esm_inf")
+        monkeypatch.setattr(Compile, "TEST_COMPILE", False)
+    flags = GOceanBuild(tmpdir).get_infrastructure_flags()
+    assert len(flags) == 2
+    assert flags[0] == "-I"
+    # This should be a temporary path (*or the above /tmp/path/dl_esm/inf one),
+    # so we can't easily check this (different relative paths when testing in
+    # parallel). So just check for the name, and that it does not include
+    # `external` (so it is not the directory in the PSyclone directory)
```
sergisiso: This comment is the opposite of what the docstring says and the asserts below do.

src/psyclone/tests/gocean_build.py
```diff
@@ -57,7 +61,7 @@ class GOceanBuild(Compile):
     '''
     # A class variable to make sure we compile the infrastructure
     # file only once per process.
-    _infrastructure_built = False
+    _infrastructure_built: Optional[bool] = None
```
sergisiso: Why does it need to be None? Could we make it bool and start with False?

src/psyclone/tests/gocean_build.py
```diff
@@ -67,31 +71,46 @@ class GOceanBuild(Compile):
...
+        if Compile.TEST_COMPILE:
+            # If we are compiling, point to the compilation path, which
+            # contain the compiled mod files.
+            root = GOceanBuild._compilation_path
+        else:
+            # If we are not compiling, point to the external infrastructure
+            # directory, which allows tests (that uses the flags) to pass
+            # even when compilation is disabled (and it will pick up if
+            # the infrastructure should change as well).
+            root = str(self._infrastructure_path)
```
sergisiso, line 107: Why is it not always `str(self._infrastructure_path)`?, "-I" should find the source, not the compiled files

src/psyclone/tests/lfric_build.py
```diff
@@ -72,36 +72,49 @@ class LFRicBuild(Compile):
...
+        if Compile.TEST_COMPILE:
+            # If we are compiling, point to the compilation path, which
+            # contain the compiled mod files.
+            include_root = LFRicBuild._compilation_path
+        else:
+            # If we are not compiling, point to the external infrastructure
+            # directory, which allows tests (that uses the flags) to pass
+            # even when compilation is disabled (and it will pick up if
+            # the infrastructure should change as well).
+            include_root = self._infrastructure_path
```
sergisiso, line 106: Same?

src/psyclone/tests/lfric_build_test.py
```diff
@@ -79,9 +91,59 @@ def test_lf_build_get_infrastructure_flags(monkeypatch, tmpdir):
...
 
+    # In the past there was a bug where the root directory of
+    # PSyclone was used to find the infrastructure directories. To
+    # detect this, check for some unexpected directory names
+    # in there:
+    assert '.git' not in dir_list
+    assert '__pycache__' not in dir_list
+    # One additional test to make sure we are not at the wrong location.
+    # ATM, the infrastructure has 16 directories
+    assert len(dir_list) > 10 and len(dir_list) < 50
```
sergisiso, line 107: I would make it  `== 16` if we want to be that precise, or remove the whole assert. Same in the test below

src/psyclone/tests/domain/gocean/gocean_build_test.py
hiker: Oops - fixed.

src/psyclone/tests/gocean_build.py
hiker: No idea why I did that :( I've simplified the code (which also simplified the test).

src/psyclone/tests/gocean_build.py
hiker: The `-I` path when compilation is enabled must point to the compiled file in the tmpdir. In order to tests, I've made it point to a 'useful' path (originally, it would default to "." in this case somehow, which then means the module manager would pick up ALL directories in psyclone, over 1000 of them).

src/psyclone/tests/lfric_build.py
hiker: Yes, same reason.

src/psyclone/tests/lfric_build_test.py
hiker: The infrastructure might change, which includes adding more directories. I picked this number that any reasonable change to the infrastructure files will not trigger a build error here. But I wanted to pick up the case when we get the default (which starting from the PSyclone root directory would add over 1000 directories). 
I've added more documentation, but if you prefer to have an exact test (which means changing this test if we update the LFRic infrastructure), I am also happy to do that.

Issue #2398, opened 2023-11-14
Atomic needs extra clauses


arporter:
The form of expression for which `!$acc atomic` is valid depends upon the clause given:
![image](https://github.com/stfc/PSyclone/assets/17409212/530ad9b9-c963-4579-b95a-564bdd820797)
This means that, for instance `x(i) = a` is only valid for an `atomic write`. 

At the moment, we do not support any clauses on the atomic directive and thus we generate code that is rejected by the compiler.

arporter:
We currently also only have the `ACCAtomicDirectiveNode` - there is no associated transformation. That could be added as part of this issue unless @sergisiso already has something in hand?

sergisiso:
I don't have any transformations, I started experimenting on the lowering adding them when needed, but I think the Transformation may be a better approach.

>This means that, for instance x(i) = a is only valid for an atomic write.

Ah interesting, it's the same in openmp but the cases I was doing had the read and write in the same statement, e.g.:
`x(i) = x(i) + a`
and did not require any additional clauses.

But tbh, I don't fully understand what the "write" alone does here, presumably we have to fence the operation from the read to the write, otherwise we don't solve the race condition, so we need the "atomic update" (which is the no clause default) or "atomic capture"?

arporter:
> But tbh, I don't fully understand what the "write" alone does here, presumably we have to fence the operation from the read to the write, otherwise we don't solve the race condition, so we need the "atomic update" (which is the no clause default) or "atomic capture"?

I think I've got myself confused. I went a bit gung ho in adding 'atomic' to all kernels but actually, I should only be adding it to those that do GH_INC. I'll refine my optimisation script...

rupertford:
And you only need to add atomic to those gh_inc that are on function spaces that are continuous in the vertical (assuming you are keeping colouring in the horizontal, otherwise ignore me).

arporter:
Thanks Rupert. I've ditched colouring entirely. We're now only 2x slower than the CPU :-)

LonelyCat124:
> But tbh, I don't fully understand what the "write" alone does here, presumably we have to fence the operation from the read to the write, otherwise we don't solve the race condition, so we need the "atomic update" (which is the no clause default) or "atomic capture"?

Atomic write would be for a statement such as `x = max(x, val)` or similar I expect, whether we come across that kind of thing in LFRic or the other codes I don't know.

sergisiso:
[...]

PR #3134, opened 2025-09-15
(Towards #2398) Add atomic types to OMPAtomicDirective

LonelyCat124:


LonelyCat124:
@sergisiso @arporter I put together a simple implementation of this - the change is overall small and it should be ready for review.
I added a couple of TODOs for things we can do later when we look at a full implementation of generated atomics.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2324,14 +2326,57 @@ class OMPAtomicDirective(OMPRegionDirective):
     Note that the standard supports blocks with 2 assignments but this is
     currently unsupported in the PSyIR.
 
+    :param ast: the entry in the fparser2 parse tree representing the code
```
arporter, line 2334: Please could you use type hints where possible.

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2324,14 +2326,57 @@ class OMPAtomicDirective(OMPRegionDirective):
...
+    :param children: the nodes that will be children of this
+                     Directive node or None.
+    :type children: Optional[List[:py:class:`psyclone.psyir.nodes.Node`]]
+    :param parent: PSyIR node that is the parent of this Directive or None.
+    :type parent: Optional[:py:class:`psyclone.psyir.nodes.Node`]
+    :param directive_type: the directive type of the atomic operation.
+    :type directive_type: :py:class:`psyclone.psyir.nodes.OMPAtomicDirective.\
+                          AtomicDirectiveType`
     '''
+    class AtomicDirectiveType(Enum):
```
arporter: Probably should be `OMPAtomicDirectiveType` as it's OpenMP specific.
EDIT: actually, I see we have ACCAtomicDirective too which sub-classes ACCRegionDirective:
image
I wonder whether we should have an `AtomicDirective` base class and then have OMP and ACC sub-classes:
```
                 OMPDirective------\
RegionDirective -> AtomicDirective ->OMPAtomicDirective
```
and similarly for ACC?

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2324,14 +2326,57 @@ class OMPAtomicDirective(OMPRegionDirective):
...
+        self._directive_type = directive_type
+
     def begin_string(self):
         '''
         :returns: the opening string statement of this directive.
         :rtype: str
 
         '''
-        return "omp atomic"
+        directive_type_to_str = {
```
arporter, line 2367: This map might then be the only thing that needs to be specialised in the ACC/OMP subclasses?

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -2324,14 +2326,57 @@ class OMPAtomicDirective(OMPRegionDirective):
...
+        CAPTURE = 4
+
+    def __init__(self, ast=None, children=None, parent=None,
+                 directive_type=None):
+        # The default atomic directive in OpenMP is UPDATE
+        if not directive_type:
+            directive_type = OMPAtomicDirective.AtomicDirectiveType.UPDATE
+        if not isinstance(directive_type,
+                          OMPAtomicDirective.AtomicDirectiveType):
+            raise GenerationError(
```
arporter: Probably this should be a TypeError.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: I've added type hints for 2 - I don't really want to import fparser here (maybe later once we have decided if we're happy with TYPE_CHECKING imports), the last can't be done here as its not yet instantiated.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: I felt it was implied due to being a subclass of `OMPAtomicDirective`, but I don't really like these classes within classes so if we do decide to move this I will have it be a standalone Enum class.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: Fixed

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: begin/end string will need specialising, along with various other methods, but otherwise most of the functionality is similar.

I also moved `_is_valid_atomic_statement` to the `Mixin` and turned it into a classmethod so the valid operators and intrinsics can be specialised individually in the future if required.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: I restructured this to a Mixin, with the Enum being a standalone class and not part of another class.

src/psyclone/psyir/nodes/atomic_mixin.py
```diff
@@ -0,0 +1,115 @@
...
+        if isinstance(stmt.rhs, BinaryOperation):
+            if stmt.rhs.operator not in cls._VALID_OPERATORS:
+                return False
+        # - or intrinsics: MAX, MIN, IAND, IOR, or IEOR
+        if isinstance(stmt.rhs, IntrinsicCall):
+            if stmt.rhs.intrinsic not in cls._VALID_INTRINSICS:
+                return False
+
+        # - one of the operands should be the same as the lhs
+        if stmt.lhs not in stmt.rhs.children:
```
arporter: This may become clear to me but won't this mean we reject the case of a simple assignment that was the start of this, e.g. `a = 1`?

src/psyclone/psyir/nodes/omp_directives.py
arporter: OK, thanks. (Yes, importing fparser here wouldn't be very nice and, in fact, it seems odd that this constructor takes this argument. However, that's not something for this PR.)

src/psyclone/psyir/nodes/atomic_mixin.py
```diff
@@ -0,0 +1,115 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
```
arporter, line 2: Sorry to be a pain but this file needs its own tests. Possibly you can move some existing ones out from somewhere. We also need to check that we can now generate the code that Rob needs - see my comment on the `is_valid_atomic_statement` method. (I see it's called in `validate_global_constraints` so we'd need a test that does the full code generation. We probably have one somewhere that can be extended.)

src/psyclone/psyir/nodes/atomic_mixin.py
LonelyCat124: Yes, this looks to be out of date since the goal is to support all atomic types, I'll remove this check now.

src/psyclone/psyir/nodes/omp_directives.py
LonelyCat124: This is just inherited down from `Node`, its possible it might not be needed on directives but I think I've just always carried it down because i'm not sure.

LonelyCat124: (Indeed directive and omp_directive are some of the only nodes to inherit this property and I don't think need to, but I'm not going to remove them here as I don't understand what ast is for on Nodes).

src/psyclone/psyir/nodes/atomic_mixin.py
LonelyCat124: Yep, added this now. I also test that overriding various properties in a subclass can change the functionality correctly.

src/psyclone/psyir/nodes/omp_directives.py
arporter: We used to keep a reference to it so as to be able to get back to the original parse tree if necessary. I'm not sure we do that anymore.

arporter:

PR #3136, opened 2025-09-16
(#Towards 3060) Adds named arguments to intrinsic_call IAttr definitions

LonelyCat124:
Also placeholders for return_type and reference_accesses for followup PRs.

@arporter I expect this is ready for review basically immediately - it should be fairly straightforward (at least compared to the other intrinsic call changes).

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   89) PR #3136 towards #3060. Adds named arguments to all intrinsic call
+   IAttr definitions. 
+
    88) PR #3094 for #2851 and #3082. Improves the DA so that PSyclone
    does not generate unnecessary firstprivate clauses on parallel regions.   
 


src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -104,641 +108,2087 @@ class Intrinsic(IAttr, Enum):
...
+            optional_args={},
+            return_type=None,
+            reference_accesses=None,
+        )
         BESSEL_JN = IAttr(
-            'BESSEL_JN', True, None, False,
-            ArgDesc(2, 3, DataNode), {})
+            name="BESSEL_JN",
+            is_pure=True,
+            is_elemental=None,
```
arporter, line 465: Could we associate a method with the property and have it check whether there are two or three arguments (as it's pure if there are only two)?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -104,641 +108,2087 @@ class Intrinsic(IAttr, Enum):
...
+            optional_args={},
+            return_type=None,
+            reference_accesses=None,
+        )
         BESSEL_YN = IAttr(
-            'BESSEL_YN', True, None, False,
-            ArgDesc(2, 3, DataNode), {})
+            name="BESSEL_YN",
+            is_pure=True,
+            is_elemental=None,
```
arporter, line 497: Same here?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -104,641 +108,2087 @@ class Intrinsic(IAttr, Enum):
...
+            reference_accesses=None,
+        )
         IBITS = IAttr(
-            'IBITS', True, True, False,
-            ArgDesc(3, 3, (DataNode)), {})
+            name="IBITS",
+            is_pure=True,
+            is_elemental=True,
+            is_inquiry=False,
+            required_args=ArgDesc(3, 3, (DataNode)),
```
arporter: Do you know why (if) we need the parentheses around `DataNode` here and in following examples?

arporter (changes requested):
Thanks Aidan. All looks good apart from where I think we can do a bit better on 'purity' for a couple of methods. If that's not too hard to do then it would be good to have.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: @arporter Is this not pure sometimes? I see its elemental if two arguments otherwise it isn't. I might not be looking at the right docs though.

We can do something similar to what I am proposing for return_type and reference_accesses, with lambdas or functions for some of these and if `isisntance(...intrinsic.is_pure, Callable)` executing the function from the `IntrinsicCall.is_pure` property.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: No, we shouldn't need them unless we have a tuple of types so I can try to remove these.

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: Oops, I meant "elemental" when I said "pure"!

LonelyCat124: I'll do the above then.

LonelyCat124: I spoke to Sergi (as this can have impacts upto the IntrinsicSymbol, which really does not want a Callable as its intrinsic), and he isn't keen on this implementation for this.
I'm adding TODOs into the code, and I will add the implementation I had onto the issue he created.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Fixed

src/psyclone/psyir/nodes/intrinsic_call.py
arporter: Thanks.

changelog
```diff
@@ -1,3 +1,6 @@
+   89) PR #3136 towards #3060. Adds named arguments to all intrinsic call
+   IAttr definitions.
```
sergisiso, line 2: @arporter @LonelyCat124 I really like the change in this PR but reading the changelog had me confused as the intrinsics also have named arguments themselves. I would suggest "Add return_type and reference_accesses fields to the intrinsic call IAttr definitions." If you agree maybe we could update it when we touch the changelog.

PR #3137, opened 2025-09-16
(Closes #3069) Allow all indentation in backend to be disabled.

### LINKED ISSUE #3069: Linebreak issues for heavily nested (and indented) code
The psyclone linebreak output utility produces wrong code for heavily nested (and therefore indented) output code. I think it happens when the number of whitespaces is already bigger than the number we are breaking with:

As a quickfix I suggested removing the indentation in the problematic codebase, and this worked:
```
--- a/src/psyclone/psyir/backend/fortran.py
+++ b/src/psyclone/psyir/backend/fortran.py
@@ -314,7 +314,7 @@ class FortranWriter(LanguageWriter):
     def __init__(self, **kwargs):
         # Construct the base class using () as array parenthesis, and
         # % as structure access symbol
-        super().__init__(("(", ")"), "%", **kwargs)
+        super().__init__(("(", ")"), "%", indent_string="", **kwargs)
```

psyclone has a `--backend {enable-validation,disable-validation}` flag where we could add a `disable-indentation` to do the same. Or we could check if this can be fixed directly in the linebreak utility.

arporter:


arporter:
Ready for a first look from @sergisiso

src/psyclone/generator.py
```diff
@@ -832,8 +838,11 @@ def code_transformation_mode(input_file, recipe_file, output_file,
         # Generate Fortran (We can disable the backend copy because at this
         # point we also drop the PSyIR and we don't need to guarantee that
         # is left unmodified)
+
+        indent = "" if Config.get().backend_indentation_disabled else None
```
sergisiso: I see this here and two more times in psyGen. But would it be better to put in only once at the src/psyclone/psyir/backend/visitor.py constructor? so that it affects all writers no matter what language or where they are instantiated? would this causes issues with sympywriter?

doc/user_guide/configuration.rst
```diff
@@ -150,33 +150,38 @@ supported by PSyclone.
...
+                             for details.
+BACKEND_CHECKS_ENABLED       Optional (defaults to True). Whether or not the PSyIR   bool
+                             backend should validate the tree that it is passed.
+                             Can be overridden by the ``--backend`` command-line
+                             flag (see :ref:`backend-options`).
+BACKEND_INDENTATION_DISABLED Optional (defaults to False). Whether or not the        bool
+                             PSyIR backend should output nested code structures
+			     using indentation. Can be overridden by the
+			     ``--backend`` command-line flag (see
+			     :ref:`backend-options`).
```
sergisiso: The indentation is not the same as the lines above

src/psyclone/generator.py
```diff
@@ -599,8 +601,12 @@ def main(arguments):
...
         # A command-line flag overrides the setting in the Config file (if
         # any).
-        Config.get().backend_checks_enabled = (
-            str(args.backend) == "enable-validation")
+        if "enable-validation" in args.backend:
+            Config.get().backend_checks_enabled = True
+        if "disable-validation" in args.backend:
+            Config.get().backend_checks_enabled = False
+        if "disable-indentation" in args.backend:
+            Config.get().backend_indentation_disabled = True
```
sergisiso: This made me wonder if we really need the "enable-validation"? Technically one could disable in the configuration and then wanting to enable it in one file, but you also could claim the same for indentation. Also this is valid `--backend disable-validation,enable-validation` and a user wouldn't know what to expect. I think I would just remove "enable-validation"

sergisiso (changes requested):
@arporter See inlined some questions I have, but in general is looking great.

src/psyclone/generator.py
arporter: Yes, done. I've tested sympywriter (by altering the Config object returned by the `config_fixture`) and all of its tests passed. I'm pretty sure we never use it to produce multi-line output anyway but I'll tag @hiker here just to check.

doc/user_guide/configuration.rst
arporter: I can't see what you mean? It all lines up in my editor:
image
and the rendered table looks fine too:
image

src/psyclone/generator.py
arporter: Yes, agreed.

doc/user_guide/configuration.rst
sergisiso: @arporter Line 178 starts using tabs, it is spaces before that.

arporter: Ah! I should have thought of that. I've fixed it now.

PR #3147, opened 2025-09-22
(closes #3138) Fix reference_access for Calls

### LINKED ISSUE #3138: Fix reference_accesses of pure subroutines
As reported by Wolfgang: after some praise for the detection of parallel variables (hooray), he had one case where a variable was declared as `intent(out)` (but no LHS assignment), and PSyclone missed it.

sergisiso:


sergisiso:
This is ready for a first review, maybe from @hiker or @arporter.

I commented out a couple of asserts and the coverage has gone down in call_tree_utils. This seems unrelated to the change but maybe you are more familiar with what call_tree_utils does. If not I will explore after merging @hiker infrastructure fix because the test also says `and the infrastructure directory has not been added`, but I am not sure it is currently true and ModuleManager looks at `.`.

src/psyclone/core/access_sequence.py
```diff
@@ -132,11 +132,24 @@ def has_indices(self) -> bool:
...
-    def access_type(self):
-        ''':returns: the access type.
-        :rtype: :py:class:`psyclone.core.access_type.AccessType`'''
+    def access_type(self) -> AccessType:
+        '''
+        :returns: the access type.
+        '''
         return self._access_type
 
+    def is_any_write(self) -> bool:
```
arporter, line 141: Please add unit tests for these methods in `access_sequence_test.py`

src/psyclone/psyir/nodes/call.py
```diff
@@ -309,27 +309,39 @@ def reference_accesses(self) -> VariablesAccessMap:
...
-        else:
-            # We conservatively default to READWRITE otherwise (TODO #446).
-            default_access = AccessType.READWRITE
-
         # The RoutineSymbol has a CALL access.
         sig, indices_list = self.routine.get_signature_and_indices()
         var_accesses.add_access(sig, AccessType.CALL, self.routine)
+
+        try:
+            routine = self.get_callee()[0]
```
arporter, line 319: Comment please: Attempt to find the target of the Call so that the argument intents can be examined.

src/psyclone/psyir/nodes/call.py
```diff
@@ -309,27 +309,39 @@ def reference_accesses(self) -> VariablesAccessMap:
...
-                var_accesses.add_access(sig, default_access, arg)
+                if routine_intents:
+                    if routine_intents[idx] == ArgumentInterface.Access.WRITE:
+                        access_type = AccessType.WRITE
+                    elif routine_intents[idx] == ArgumentInterface.Access.READ:
+                        access_type = AccessType.READ
+                    else:
+                        access_type = AccessType.READWRITE
+                else:
+                    access_type = AccessType.READWRITE
```
arporter, line 346: "We haven't resolved the target of the Call so arguments default to having READWRITE access."

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
```diff
@@ -307,8 +307,8 @@ def test_get_non_local_read_write_info(caplog):
     with caplog.at_level(logging.WARNING):
         ctu.get_non_local_read_write_info(schedule, rw_info)
     assert "Unknown routine 'unknown_subroutine - ignored." in caplog.text
-    assert ("Cannot find module 'constants_mod' - ignoring unknown symbol "
-            "'eps'." in caplog.text)
+    # assert ("Cannot find module 'constants_mod' - ignoring unknown symbol "
```
arporter: The indirect, missing  coverage is a puzzler. Could it be that having extended Call.reference_accesses() to do a `get_callee()`, we don't use the ModuleManager in the same way? It might be worth putting a breakpoint in the uncovered code on master and seeing which test(s) hit it and how they get there.

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
```diff
@@ -627,8 +627,8 @@ def testcall_tree_utils_non_local_inout_parameters(caplog):
         rw_info = ctu.get_in_out_parameters(schedule,
                                             collect_non_local_symbols=True)
     assert "Unknown routine 'unknown_subroutine - ignored." in caplog.text
-    assert ("Cannot find module 'constants_mod' - ignoring unknown symbol "
-            "'eps'." in caplog.text)
+    # assert ("Cannot find module 'constants_mod' - ignoring unknown symbol "
```
arporter: Same here.

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
hiker: Got the reason, not sure about the fix:
1. get_callees in call.py:
container = csym.find_container_psyir(local_node=self)
2. find_container_psyir in  containersymbol.py
 self._reference = self._interface.get_container(self._name)
3. get_container in containersymbol.py:
      mod_manager.add_search_path(Config.get().include_paths)

This adds the lfric external/infrastructure directory to the search path, and as a result constants_mod is found, so the exception is not triggered. And indeed, if I remove this line, the missing lines in call_tree_utils are then covered.

src/psyclone/core/access_sequence.py
sergisiso: Done

src/psyclone/psyir/nodes/call.py
sergisiso: Done

src/psyclone/psyir/nodes/call.py
sergisiso: Done

src/psyclone/tests/psyir/tools/call_tree_utils_test.py
sergisiso: The quick fix is to modify that last line to:
```
mod_manager.add_search_path(Config.get().include_paths, recursive=False)
```
As only the "-d (kernel_path)" should have the recursive behaviour and those are given somewhere else.

However, I am not fully convinced that this method should happily do the import, at least not by default. I started adding a argument "load_external_files", but is touches quite a bit of unrelated code, e.g. inlining. Maybe I should do this in a separate PR?

arporter: Yes, I think a separate PR. Conversely, I'm sure I've written at least one TODO saying it would be useful if we could request a given transformation (I think) to attempt to resolve symbols as required. It will need some care.

PR #3153, opened 2025-09-23
(Closes #3151) small case-related fixes to HoistLocalArraysTrans and Call.replace_named_arg()

### LINKED ISSUE #3151: Bug in HoistLocalArraysTrans
HoistLocalArraysTrans can fail if the variable to be hoisted needs renaming (because of a clash at the Container level) but cannot be because it is e.g. accessed inside a CodeBlock. The `validate` method should catch this.

arporter:


arporter:
...I tried again and I get just one, inconsequential difference:
```
diu_layers.f90
11c11
<   public :: diurnal_sst_coolskin_init, diurnal_sst_bulk_init, diurnal_layers
---
>   public :: diurnal_sst_bulk_init, diurnal_sst_coolskin_init, diurnal_layers
```
I've also tried re-running the action but the results were the same :-(

arporter:
I tried again on Rupert and still got no significant source differences. I then re-ran the integration test a third time and this time it passed. That probably means we have a bug in the asynch. version of NEMO but, happily, it's no the fault of this PR so I declare it ready for review :-)

src/psyclone/psyir/transformations/hoist_local_arrays_trans.py
```diff
@@ -374,14 +376,19 @@ def _get_local_arrays(node):
...
-            # TODO #11 - log the fact that we can't hoist the arrays
-            # listed in 'names_in_cblock'.
+            all_names_in_cblock.update(names_in_cblock)
             for name in names_in_cblock:
                 del local_arrays[name]
+        if all_names_in_cblock:
+            logger.warning(
+                "Cannot hoist local array(s) %s out of Routine '%s' as they "
+                "are accessed in a CodeBlock",
+                sorted(list(all_names_in_cblock)), node.name)
```
sergisiso: I would prefer to report this as preceding_comments as we do with the other two invalid conditions above. 

(I don't mind also reporting it to the logger but then I would do it as suggested in #2670 so the validation issues are consistently reported to the same place, and not some in one place and others in another place)

arporter: Good point. I was focused on removing the TODO for #11 and didn't stop to think.

Issue #3140, opened 2025-09-18
Parameter symbol accessed within derived type is not updated with initial value


arporter:
As reported by @schreiberx, just running PSyclone on the following Fortran:
```fortran
MODULE test_mod
   IMPLICIT NONE

   ! parameters
   INTEGER, PARAMETER :: rsh = 8
   REAL(KIND=rsh), PARAMETER :: pi = 3.14159265358979323846

   TYPE o1dv_param_type
      INTEGER       :: z0bstress_option
      REAL(KIND=rsh) :: c_rho
      REAL(KIND=rsh), ALLOCATABLE :: height_norm(:)

   END TYPE

   TYPE o1dv_out_type
      REAL(KIND=rsh), ALLOCATABLE :: height(:)
      REAL(KIND=rsh) :: dens_mean
   END TYPE

CONTAINS

END MODULE test_mod
```
results in `_gen_parameter_decls` crashing because the Symbol `rsh` is a parameter but is missing its initial value.

I put a breakpoint in `SymbolTable.add()` and that showed that `rsh` is first added to the SymbolTable when parsing a derived type - at that point we know it is a kind parameter and therefore is a constant integer but we don't yet  know more. Presumably, when we come to handle the actual declaration of `rsh` we fail to capture its initial value.

PR #3160, opened 2025-09-25
(Closes #3159, #3140) bug fixes for parameter declaration handling

### LINKED ISSUE #3159: Parameter declaration that refers to itself causes PSyclone to crash
Running PSyclone on:
```fortran
module my_mod
  implicit none
  integer, parameter :: nFieldNames3d = 4
  Integer, Parameter :: InterpolationLevels(nFieldNames3d) = & 
                                       [ 2, 0, &
                                         Huge(InterpolationLevels)/3, 0 ]
end module my_mod
```
causes a crash:
```bash
  File "/home/PSyclone/src/psyclone/psyir/backend/fortran.py", line 1026, in gen_decls
    declarations += self._gen_parameter_decls(symbol_table,
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/PSyclone/src/psyclone/psyir/backend/fortran.py", line 921, in _gen_parameter_decls
    raise VisitorError(
psyclone.psyir.backend.visitor.VisitorError: Visitor Error: Unable to satisfy dependencies for the declarations of ['interpolationlevels']
```

arporter:
There were two bugs in the logic that determined the correct order in which to declare parameters:
 1. it was case sensitive;
 2. a parameter that depended on itself resulted in us raising an exception.
 
(For clarity, 2. is a possibility because of declarations like:
```fortran
real, parameter, dimension(4) :: var = [1, 1, 1, HUGE(var)]
```
)
In addition, while testing, I hit #3140 so have fixed that too - while processing a declaration, if we encountered a kind parameter symbol that we hadn't seen before we would create a DataSymbol. However, we did not update the properties of it when we subsequently found the actual declaration of that symbol because it was already a `DataSymbol` rather than just a `Symbol`.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,5 @@
+   97) PR #3160 for #3159 and #3140. Fix declarations with self-references.a
+
    96) PR #3153 for #3151. Fix some case-related naming comparison issues.
 
    95) PR #3108 for #2446. Adds the ability to automatically add reduction
--- a/src/psyclone/psyir/backend/fortran.py
+++ b/src/psyclone/psyir/backend/fortran.py
@@ -856,13 +856,13 @@ class FortranWriter(LanguageWriter):
 
         :param symbol_table: the SymbolTable instance.
         :type symbol: :py:class:`psyclone.psyir.symbols.SymbolTable`
-        :param bool is_module_scope: whether or not the declarations are in \
+        :param bool is_module_scope: whether or not the declarations are in
                                      a module scoping unit. Default is False.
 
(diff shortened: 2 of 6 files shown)


LonelyCat124:
Hi @arporter - does this address #2202 or are we leaving it until later?

arporter:
We almost had a test that already did this so I've just tweaked it slightly. Should be ready for another look now. The Linkspector failures are because the Tau website is on a go-slow (I can still get to it though).

Issue #1320, opened 2021-06-30
Stop supporting non PSyIR nodes in variable access code (esp ComponentIndices)


hiker:
At this stage we still allow non PSyIR nodes in the variable access code (esp. ComponentIndices, likely in other classes as well). There is still some LFRic code and gocean that relies on strings, once this is all refactored, this can be done.
- [ ] domain/lfric/arg_ordering.py
- [ ] domain/lfric/kern_call_arg_list.py
- [ ] many many tests :(
- [ ] dynamo0p3 ('1')
- [ ] Avoid the usage of the `to_fortran` function - it should be language independent. @arporter suggests to provide a visitor instance, which will be used (language independent) to provide a string for 'nice' error messages (main problem is where this instance is created from? LIkely the dependency tool can take this as an argument in the constructor, but where is this then taken from? Typically the dependency tools will be used from a transformation, so the transformation needs to know the language. I should also point out that e.g. psyGen already contains a `bool_to_fortran` function :) )

Additionally, more consistency checks can then be added:
- [ ] todo: 
```
In src/psyclone/core/access_info.py:

>          if not isinstance(signature, Signature):
             raise InternalError("Got '{0}' of type '{1}' but expected it to "
                                 "be of type psyclone.core.Signature."
                                 .format(signature, type(signature).__name__))
 
-        if component_indices is None:
-            component_indices = [[]]
+        # To make it easier for the user, we allow to implicitly create the
+        # component indices instance here:
+        if not isinstance(component_indices, ComponentIndices):
+            # Handle some convenient cases:
+            # 1. Add the right number of [] if component_indices is None:
+            if component_indices is None:
+                component_indices = [[]] * len(signature)
+            elif isinstance(component_indices, list):
+                # 2. If the argument is a simple list (not a list of lists),

This behaviour needs documenting in the docstring. Also, it should be possible to do some sanity checking by ensuring that the type of the last component permits indexing (must be a subclass of ArrayMixin).

In src/psyclone/core/access_info.py:

>          if not isinstance(signature, Signature):
             raise InternalError("Got '{0}' of type '{1}' but expected it to "
                                 "be of type psyclone.core.Signature."
                                 .format(signature, type(signature).__name__))
 
-        if component_indices is None:
-            component_indices = [[]]
+        # To make it easier for the user, we allow to implicitly create the
+        # component indices instance here:
+        if not isinstance(component_indices, ComponentIndices):
+            # Handle some convenient cases:
[...]

Issue #2424, opened 2023-11-28
`reference_accesses` does not support nested scopes


arporter:
The PSyIR supports nested scoping regions and some transformations will add new Symbols to the symbol table that is closest to the point at which they are needed. Since Fortran does not support nested scoping regions, all nested tables are 'flattened' into a single one at the Routine level by the Fortran backend.

However, the dependence analysis tooling accessed via the `Node.reference_accesses` method (and specialisations thereof) is intended to be used with generic PSyIR. In walking the PSyIR tree, it does not look at the identity of Symbols that are encountered - it only considers their names. This means that two Symbols in local scopes that happen to have the same name will be considered to be the same access, even though they actually refer to different memory locations.

arporter:
The DA tooling is all based on `Signature` which holds a purely text-based description of an access. Therefore, two accesses to different Symbols of the same name will have indistinguishable Signatures. Really, the `Signature` needs to hold information on
the identity of the `Symbol` that is being accessed. This is complicated by the fact that `Signature` is used as a key in `VariablesAccessInfo` - in order to make sure that two different signature objects containing the same variable can be used as a key, this class implements `__hash__` and other special functions.

Unfortunately, my understanding of the implications of this is a bit hazy. Could we include the ID of the Symbol object in the hash @hiker?

LonelyCat124:
> Unfortunately, my understanding of the implications of this is a bit hazy. Could we include the ID of the Symbol object in the hash @hiker?

The "general" rule of Python is that if 2 objects are `==` that their hash should also be equal (so for example when looking up  `mysig` by hash you don't miss a different object with the same key . I think its the same issue as the rest of it where the DA tooling is currently unable to differentiate the same named symbol in different scopes - so its possible its a larger issue.

LonelyCat124:
This primarily occurs dur to `InlineTrans` (which we may change anyway), but if PSyclone supports symboltables at any `ScopingNode` then we technically allow this generally though.

hiker:
I might argue that supporting symbol tables in every block might not have been the best decision :)

I have used very similar code in the driver creation I think: taking an existing PSyIR and putting a copy into a new program unit (the driver). As part of this I transfer all the symbols (and potentially rename them, e.g. I remove 'proxy' from the name to make the variable names easier to understand for the user ... though I just realise we might not have 'proxy' in argument names anymore :)  ).

[...]

Issue #2531, opened 2024-03-26
GOcean accepts kernels that cannot be parallelised


hiker:
In some tests it was discovered that gocean accepts kernels that cannot be parallelised (because of a stencil access, and a write statement to the same field):
https://github.com/stfc/PSyclone/pull/2497/files#r1505787613

I have to admit I wasn't aware that this is a requirement, but I think it also needs to be stressed more in the documentation. ATM we only have (https://psyclone.readthedocs.io/en/latest/kernel_layer.html#kernel-layer):

> The reason for doing this is that it gives the PSy layer the responsibility of calling the Kernel over the spatial domain which is where parallelism is typically exploited in finite element and finite difference codes. The PSy layer is therefore able to call the kernel layer in a flexible way (blocked and/or in parallel for example). Kernel code in the kernel layer is not allowed to include any parallelisation calls or directives and works on raw Fortran arrays (to allow the compiler to optimise the code).

Not sure if I've missed a stronger statement elsewhere.

TODO:
1. Reject kernels that have read-write accesses to a field in a kernel (including checks for LFRic)
2. Update the documentation if required.

arporter:
I think the rules are:

  1. A kernel must only write to the current/'owned' cell.
  2. Any fields with reads from neighbouring locations must be marked in the metadata as having a stencil access.

hiker:
These statements do not mean that a kernel can get parallelised (which always was my understanding). The dependency tool are used to check this (in psyir/transformations/parallel_loop_trans).

So from that point of view, it sounds like this ticket can just be closed, since kernels that cannot be parallelised should be accepted.

arporter:
Probably we are agreeing but just to be clear - it should be possible to parallelise (the loop over cells containing) a kernel but the metadata should contain enough information to allow PSyclone to deduce that e.g. colouring is required. (e.g. in LFRic, we can deduce when a kernel writes to a shared dof and ensure that such an access is protected appropriately.)

hiker:
Hmm - I am not sure we agree :)   Example, with a stencil of 010,010,010, which seems to fulfil the two rules: only write to `a(i,j)`, and declare the stencil access (and we can even remove `a(i,j)` from the RHS):
```
   a(i,j) = a(i,j-1) + a(i,j) + a(i,j+1)
```
[...]

PR #3164, opened 2025-09-29
(towards #2424) Convert component indices to method

sergisiso:
This is a step to reduce internal state of Signatures so they operatate on top of the PSyIR. In thi PR I removed the ComponentIndices class (and associated infrastructure) and the argument in add_accesses. Instead I moved the logic to a method of Reference. So now, we just need to do:

`access_sequence.add_access(AccessType.READ, ArrayReference(Symbol("a")))`

and once the component_indices are needed the AccessSequence can do self.node.component_indicies()


sergisiso:
@arporter @hiker This is ready for review. It was initially a towards #2424 PR, but I found that it may close #1320 as now compoment indices contain the actual psyir nodes from the tree and not strings. So I changed the title to close that one instead.

src/psyclone/core/access_sequence.py
```diff
@@ -62,15 +61,9 @@ class AccessInfo():
     '''
     def __init__(
         self, access_type: AccessType, node: 'Node',
```
arporter, line 62: Please rm component_indices from the docstring too.

src/psyclone/core/access_sequence.py
```diff
@@ -91,45 +84,33 @@ def change_read_to_write(self):
...
-
-        :param component_indices: indices used in the access.
-
-        :raises InternalError: if component_indices is not an instance \
-            of :py:class:`psyclone.core.component_indices.ComponentIndices`.
-
+        :returns: a tuple of each indexed compoment in the accessor with
+            a tuple of each index expression in that compoment. For example,
+            for a scalar it returns `(())`, for `a%b` it returns ((),()) - two
+            components with 0 indicies in each, and for `a(i)%b(j,k+1)` it
```
arporter: "indices"

src/psyclone/core/access_sequence.py
```diff
@@ -91,45 +84,33 @@ def change_read_to_write(self):
...
-            raise InternalError(f"The component_indices object in the setter "
-                                f"of AccessInfo must be an instance of "
-                                f"ComponentIndices, got '{component_indices}'")
-        self._component_indices = component_indices
+        # Only Reference has component_indices, for everything else we assume
+        # it is a scalar
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.nodes import Reference
+        if not isinstance(self._node, Reference):
+            return lambda: tuple(tuple())
```
arporter: Shouldn't this just be `return tuple(tuple())` (as that is what the docstring says it returns) rather than being a Callable?

src/psyclone/core/access_sequence.py
```diff
@@ -321,7 +300,7 @@ def add_access(
         :param component_indices: indices used for each component of the \
```
arporter: Please rm from docstring.

src/psyclone/domain/gocean/transformations/gocean_loop_fuse_trans.py
```diff
@@ -89,15 +89,15 @@ def validate(self, node1, node2, options=None):
...
-        super(GOceanLoopFuseTrans, self).validate(node1, node2,
-                                                  options=options)
-
         if node1.field_space != node2.field_space:
             raise TransformationError(
                 f"Error in {self.name} transformation. Cannot "
                 f"fuse loops that are over different grid-point types: "
                 f"{node1.field_space} and {node2.field_space}")
 
+        super(GOceanLoopFuseTrans, self).validate(node1, node2,
```
arporter: Can this just be `super().validate(....)`?

src/psyclone/gocean1p0.py
```diff
@@ -411,6 +410,11 @@ def independent_iterations(self,
         :rtype: bool
 
         '''
+        # By definition a GOLoop with one GOKern is loop-independent
+        if len(self.children) == 4 and len(self.loop_body.children) == 1:
```
arporter: Why the check on the number of children of the Loop? Is this in case the Loop is not yet fully formed?

src/psyclone/gocean1p0.py
```diff
@@ -994,25 +968,13 @@ def _record_stencil_accesses(self, signature, arg, var_accesses):
...
             for i in [-1, 0, 1]:
                 depth = arg.stencil.depth(i, j)
-                for current_depth in range(1, depth+1):
-                    # Create PSyIR expressions for the required
-                    # i+/- and j+/- expressions
-                    i_expr = GOKern._create_psyir_for_access(symbol_i, i,
-                                                             current_depth)
-                    j_expr = GOKern._create_psyir_for_access(symbol_j, j,
-                                                             current_depth)
+                for _ in range(1, depth+1):
```
arporter: Do we still need the loop over depths? The way the code is now, if there's a stencil access then we will always have one access that is `arg.access` and one or more that are `AccessType.READ`?

src/psyclone/gocean1p0.py
```diff
@@ -1021,8 +983,7 @@ def _record_stencil_accesses(self, signature, arg, var_accesses):
                     else:
                         acc = AccessType.READ
 
-                    var_accesses.add_access(signature, acc, self,
-                                            [i_expr, j_expr])
+                    var_accesses.add_access(signature, acc, self)
```
arporter: Here, `signature` and `self` are loop invariant.

src/psyclone/lfric.py
```diff
@@ -517,12 +516,8 @@ def kern_args(self, stub=False, var_accesses=None,
                 arg_list.append(adj_face)
 
                 if var_accesses and not kern_call_arg_list:
-                    # TODO #1320 Replace [1]
```
arporter, line 520: Whoop!

src/psyclone/core/access_sequence.py
```diff
@@ -91,45 +84,33 @@ def change_read_to_write(self):
         self._access_type = AccessType.WRITE
 
     @property
```
arporter: I think it would be best to make this consistent with the new implementations in the various Nodes. i.e. remove the `@property` decorator.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -549,6 +549,16 @@ def indices(self) -> Tuple[Node]:
                     f"expression but found '{type(child).__name__}'")
         return tuple(self.children)
 
+    def component_indices(self) -> tuple[tuple[Node]]:
+        '''
+        :returns: a tuple of each indexed compoment in the accessor with
```
arporter: "a tuple of tuples of index expressions; one for every component in the accessor. For example...."

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -549,6 +549,16 @@ def indices(self) -> Tuple[Node]:
...
+            a tuple of each index expression in that compoment. For example,
+            for a scalar it returns `(())`, for `a%b` it returns ((),()) - two
+            components with 0 indicies in each, and for `a(i)%b(j,k+1)` it
+            returns `((i,),(j,k+1))`.
+        '''
+        return (self.indices,)
```
arporter, line 567: Since `self.indices` returns a list, shouldn't we return `(tuple(self.indices),)`?

src/psyclone/psyir/nodes/array_mixin.py
arporter: Note you've got a consistent typo of "compoment" :-)

src/psyclone/psyir/nodes/array_of_structures_mixin.py
```diff
@@ -127,6 +127,16 @@ def indices(self) -> Tuple[Node]:
...
 
+    def component_indices(self) -> tuple[tuple[Node]]:
+        '''
+        :returns: a tuple of each indexed compoment in the accessor with
+            a tuple of each index expression in that compoment. For example,
+            for a scalar it returns `(())`, for `a%b` it returns ((),()) - two
+            components with 0 indicies in each, and for `a(i)%b(j,k+1)` it
+            returns `((i,),(j,k+1))`.
+        '''
+        return (self.indices, *self.member.component_indices())
```
arporter, line 138: `tuple(self.indices)` ?

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -267,11 +277,35 @@ def _partition(comp_ind1, comp_ind2, loop_variables):
         :rtype: List[Tuple[Set[str], List[int]]]
 
         '''
+        def get_subscripts_of(
+                component_indices, set_of_vars: set[str]
+        ) -> list[set[str]]:
+            '''This function returns a flat list of which variable from the
```
arporter: "..flat list of sets, one for each index expression within the access. Each set holds the intersection of the set of variables used in that index expression with the input set of variables."

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -267,11 +277,35 @@ def _partition(comp_ind1, comp_ind2, loop_variables):
         :rtype: List[Tuple[Set[str], List[int]]]
```
arporter: Pls update the type of `comp_ind1/2` above as they aren't ComponentIndices anymore.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -715,39 +756,8 @@ def _array_access_parallelisable(self, loop_variables, var_info):
...
-                        # only contains '< kern call: NAME >', so no
-                        # information about the variable and its indices is
-                        # available. For GOKerns, 'reference_accesses' adds
-                        # artificial accesses to the component indices
-                        # depending on the declared stencil. Use these indices
-                        # and the signature to add additional info to the
-                        # error message that indicates which access exactly is
-                        # causing the problem:
-
-                        if isinstance(write_access.node, GOKern):
```
arporter, line 732: Really nice to lose this specialisation :-)

src/psyclone/tests/core/variables_access_map_test.py
```diff
@@ -271,11 +208,13 @@ def test_derived_type_array(array, indices, fortran_writer, fortran_reader):
     assert "j: READ" in str(vai1)
     assert "k: READ" in str(vai1)
 
-    # Verify that the index expression is correct. Convert the index
-    # expression to a list of list of strings to make this easier:
+    # Verify that the index expression is correct. First replace its
```
arporter: I don't understand this Comment - I can't see any References being created?

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
```diff
@@ -141,37 +140,6 @@ def test_gok_reference_accesses(fortran_writer):
...
-    # in a set, so that's the easiest way to compare them)
-    result = set()
-    for access in p_fld:
-        for indices in access.component_indices:
-            result.add(str([fortran_writer(index) for index in indices]))
-
-    # The stencil is 100, 110, 123 - test that appropriate
-    # accesses were added for each direction
-    expected = {
-        # First stencil direction of 123: 1
```
arporter, line 160: So we no longer have information on the stencil accesses of this kernel as provided by the metadata?Obviously this hasn't broken anything in the test suite but does it break anything for @hiker (e.g. in training/tutorials)?

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
```diff
@@ -196,12 +164,5 @@ def test_gok_access_info_scalar_and_property():
     # Check that the derived type using tmask has the corresponding component
     # indices specified. No indices for p_fld and grid:
     tmask = vam[Signature("p_fld%grid%tmask")]
-    comp_ind = tmask[0].component_indices
-    assert comp_ind[0] == []
-    assert comp_ind[1] == []
-
-    # And it should have PSyIR expressions for (i,j) as the last component:
-    assert isinstance(comp_ind[2][0], Reference)
```
arporter, line 204: Again, we no longer have the (artificial) index information?

src/psyclone/tests/psyir/tools/dependency_tools_test.py
```diff
@@ -277,11 +277,6 @@ def test_partition(lhs, rhs, partition, fortran_reader):
         # not matter:
         assert correct[0] == part_info[0]
 
-        # Then check that the partition indices are the same as well.
```
arporter, line 280: How come you've removed this part of the test?

src/psyclone/tests/psyir/tools/dependency_tools_test.py
```diff
@@ -751,10 +745,8 @@ def test_gocean_parallel():
     parallel = dep_tools.can_loop_be_parallelised(loop)
     assert not parallel
 
-    assert ("The write access to 'u_fld(i,j)' in '< kern call: "
-            "stencil_not_parallel_code >' and the read access to "
-            "'u_fld(i,j - 1)' in '< kern call: stencil_not_parallel_code >' "
-            "are dependent and cannot be parallelised. Variable: 'u_fld'."
+    assert ("Variable 'u_fld' is read first, which indicates a reduction."
```
arporter, line 752: I do think this is better now as the message does not refer to artificially-introduced index variables (I don't think we actually looked at the index expressions passed to the kernel did we?).

arporter:
I've set the integration tests going as that will help to set my mind at rest :-)

src/psyclone/core/access_sequence.py
sergisiso: Done

src/psyclone/core/access_sequence.py
sergisiso: Fixed

src/psyclone/core/access_sequence.py
sergisiso: Yes, simplified it now.

src/psyclone/core/access_sequence.py
sergisiso: Removed

src/psyclone/domain/gocean/transformations/gocean_loop_fuse_trans.py
sergisiso: Yes, in fact pylint prefers it

src/psyclone/gocean1p0.py
sergisiso: Ah, it seems that I added this for the test in src/psyclone/tests/domain/gocean/goloop_test.py:325, but in checking this again, I prefer the method to complain that this an incomplete node.

src/psyclone/gocean1p0.py
sergisiso: We don't need the loop, but I had to add a depth > 0 check.

Which bares the question, is a stencil with depth 0 a stencil at all? So I looked at some examples an in fact the arguments that reach this only 0 depths are not stencils but GO_POINTWISE fields. However this still have a GOStencil returned from arg.stencil. So I refactored a bit the code to make this more clear.

src/psyclone/gocean1p0.py
sergisiso: See comment above

src/psyclone/core/access_sequence.py
sergisiso: I agree

src/psyclone/psyir/nodes/array_mixin.py
sergisiso: Ups, fixed all them now I think

src/psyclone/psyir/nodes/array_mixin.py
sergisiso: self.indices returns a tuple. I asked for this change in a review some time ago to make it immutable as I found a bug associate to this.

src/psyclone/psyir/nodes/array_of_structures_mixin.py
sergisiso: See above

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: Done

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: Done

src/psyclone/tests/core/variables_access_map_test.py
sergisiso: The comment was confusing, let me know if the new version is better.

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
sergisiso: Let me know if it does and I can have a look

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
sergisiso: Updated comment

src/psyclone/tests/psyir/tools/dependency_tools_test.py
sergisiso: There are no index expressions passed to the kernel. The generated code is:
```
    do j = u_fld%internal%ystart, u_fld%internal%ystop, 1
      do i = u_fld%internal%xstart, u_fld%internal%xstop, 1
        call stencil_not_parallel_code(i, j, u_fld%data, v_fld%data)
      enddo
    enddo
```
The issue is that the kernel writes outside its given i,j index. There is a larger discussion in 2531 but essentially we need to decide if this is a valid kernel at all, and if it is what semantics does it have in terms of dependencies. This is low priority because nemolite uses new/after fields which are READ or WRITE, but almost never READWRITE (with the exception of bc_flatter). But I think we need to tackle that issue first before we can properly decide which is the appropriate message to give here.

src/psyclone/tests/psyir/tools/dependency_tools_test.py
sergisiso: Aha, I deleted it because the indices were different and I wanted to see where it failed, but since nothing failed I forgot to come back to this. It turns out that the indices were different because since `get_subscripts_of` returned flat indices, I decided to go ahead with those and reshuffle all comonent_indices I found. However, I should instead have gone backwards and make `get_subscripts_of` return nested indices, which is much nicer as it doen't require the constant flattening. I have made this change here https://github.com/stfc/PSyclone/pull/3164/commits/93f8db9b139f5b9e81725dcffce4e32773a78e7f and now this assert can be restored.

src/psyclone/gocean1p0.py
arporter: I have  a vague feeling that we may have abused the stencil functionality for the purposes of the halo-exchange placement logic. I'm happy with what you have though.

src/psyclone/psyir/nodes/member.py
```diff
@@ -144,10 +144,10 @@ def get_base_and_depth(self):
 
     def component_indices(self) -> tuple[tuple[Node]]:
         '''
-        :returns: a tuple of each indexed compoment in the accessor with
-            a tuple of each index expression in that compoment. For example,
-            for a scalar it returns `(())`, for `a%b` it returns ((),()) - two
-            components with 0 indicies in each, and for `a(i)%b(j,k+1)` it
+        :returns: a tuple of tuples of index expressions; one for every
```
arporter, line 147: To save confusion, I think it would be better to specialise this text for the class in question. i.e. "...expressions; since Member has no indices we return a tuple containing an empty tuple."

src/psyclone/psyir/nodes/reference.py
```diff
@@ -393,10 +393,10 @@ def replace_symbols_using(self, table_or_symbol):
 
     def component_indices(self) -> tuple[tuple[Node]]:
         '''
-        :returns: a tuple of each indexed compoment in the accessor with
-            a tuple of each index expression in that compoment. For example,
-            for a scalar it returns `(())`, for `a%b` it returns ((),()) - two
-            components with 0 indicies in each, and for `a(i)%b(j,k+1)` it
+        :returns: a tuple of tuples of index expressions; one for every
```
arporter, line 411: Again, please specialise this text (can be the same as Member).

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -293,14 +282,16 @@ def get_subscripts_of(
                 corresponding array subscripts as strings.
```
arporter: Strictly, this returns a list of lists of tuples now. Also the type-hint on the return type doesn't match.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -315,14 +306,18 @@ def get_subscripts_of(
...
-            # with and without indices, e.g.: a(i) = a*a
-            # In this case we don't add this to the partition, which will
-            # result in an empty partition (which in turns will disable
-            # parallelisation).
-            if i < len(indices_2):
-                partition_infos.append((indices_1[i].union(indices_2[i]), [i]))
+        for i1, idx_exprs in enumerate(comp_ind1):
+            for i2, _ in enumerate(idx_exprs):
+                try:
+                    partition_infos.append(
```
arporter, line 313: The description at L300 needs updating now that the content of `partition_infos` has changed.

src/psyclone/tests/core/variables_access_map_test.py
```diff
@@ -208,12 +207,14 @@ def test_derived_type_array(array, indices, fortran_writer, fortran_reader):
     assert "j: READ" in str(vai1)
     assert "k: READ" in str(vai1)
 
-    # Verify that the index expression is correct. First replace its
-    # strings with references to that symbol
+    # The elements in the paratrized indices are string names, but
```
arporter: "parameterised"

src/psyclone/tests/core/variables_access_map_test.py
arporter: Ah, got it.

src/psyclone/tests/psyir/tools/dependency_tools_test.py
arporter: OK. I agree this is for #2531 to sort out.

arporter (changes requested):
Thanks Sergi, very nearly there now.
I've pinged @hiker in Teams in case there's anything relating to GOcean stencils not covered by our test suite.
Just a bit of tidying to do.

hiker:
I have debugged why loop fusion in my training is now failing. Core reason is that (with this PR merged in) fields as kernel parameters are now detected to be scalars. E.g.:
```
            call invoke(count_neighbours(neighbours, current),)
```
will now consider `current` to be a scalar, because ` is_array = access_info.has_indices(index_variable=index_variable)` returns `False`. I didn't debug this into more detail, but I believe what is happening is that the test:
```
        if access_info:
            # Access Info might not have information if a variable is used
            # as array (e.g. in case of an array expression). In this case
            # we still need to check the type information in the symbol table.
            is_array = access_info.has_indices(index_variable=index_variable)
```
and that seems a likely candidate, since previously this function would ask its ComponentIndices :)

In the past, I've added implicitly accesses to `field(i+-1, j+-1)` (depending on stencil), these indices are now not added anymore.

And then a gocen field is not considered an array (when checking the symbol info), and it fails. Should be easy to reproduce, fusing two gocean loops should trigger this problem (I am actually surprised that this was not picked up by the testing). If you need a test case, merge in branch `1623...`, and check `.../tutorial/training/gocean/2.6-GameOfLife-fuse/solution` (and run `make test`).

hiker:
I checked, surprisingly we seems to have NO tests for fusing gocean loops :(:
```
psyclone/src/psyclone/tests$ find . -iname \*fus\*.py
./psyir/transformations/loop_fusion_test.py

```

sergisiso:
Thanks for checking @hiker , I was expecting this at some point, so it is good to have an example. I'm happy for the training to be merged first if it's close to being ready.

This is expected behaviour, let me explain.

> Core reason is that (with this PR merged in) fields as kernel parameters are now detected to be scalars

Not quite. To know the datatype you would do arg.datatype, and this should return ArrayaType or UnresolvedType, but not ScalarType. Or arg.datatype.is_array (with None and False having different meaning)

The 'has_indices' should return False to be consistent with the psyir because it has no indices, e.g. "call kernel1(arg1%data, ...)"

To fix fusing there are 3 options (coming in the next comment as I need to get off the bus)

hiker:
Oh, we have some gocean fuse tests in ./domain/gocean/transformations/gocean1p0_transformations_test.py, but they are mostly (all?) only testing errors :(

hiker:
Ah, I actually didn't think it all the way through. I thought you were adding the accesses 'after' (i.e. next to the) kernel, but as child of the kernel that makes a lot more sense, I totally missed that.

That sounds actually good, thanks!

sergisiso:
@hiker I have a potential solution implemented (through inlining), and the ideas shared above for a longer term solution. I will share a branch later today or tomorrow to see what you think.

src/psyclone/psyir/nodes/member.py
sergisiso: I don't agree with this because quite a few of them are iterative and quite a few are subclassed. The two that you added comments Member and Reference, both have subclasses where this docstring will not be true.

I think we should avoid that because we do thinks like;
```
refs: List[Reference] = psyir.walk(Reference)
ref[0].component_indices <- And I request the docstring here
```
It will say "has no indices we return a tuple containing an empty tuple" but it could be wrong. Same for Member.

src/psyclone/psyir/nodes/reference.py
sergisiso: Same

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: Fixed, I think

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: Done

src/psyclone/tests/core/variables_access_map_test.py
sergisiso: Done

hiker:
As said in the telco, I am happy to get this merged in!

doc/developer_guide/dependency.rst
```diff
@@ -390,102 +390,6 @@ instances to store all accesses to a single variable. A new instance of
     :no-index:
     :members:
```
hiker, line 392: I wonder if it's worth to have/maintain a (much shorter) section that explains the `component_indices` function that you have added, since the data structure used (tuple of tuples and its mapping to structure accesses) might not be obvious?

src/psyclone/gocean1p0.py
```diff
@@ -411,6 +410,11 @@ def independent_iterations(self,
         :rtype: bool
 
         '''
+        # By definition a GOLoop with one GOKern is loop-independent
```
hiker: Hmm - in theory you are right, but afaik this is not enforced anywhere. We have #2531 that discusses this issue. Unless this change is required, I would prefer to remove this shortcut (and maybe add this as a comment to #2531, so it can be added later??)

src/psyclone/gocean1p0.py
```diff
@@ -988,41 +962,24 @@ def _record_stencil_accesses(self, signature, arg, var_accesses):
...
-                    var_accesses.add_access(signature, acc, self,
-                                            [i_expr, j_expr])
+                if i == 0 and j == 0:
+                    # This is not necessarely a stencil, and the access
+                    # pattern is given by the argument metadata
+                    var_accesses.add_access(signature, arg.access, self)
+                elif arg.stencil.depth(i, j) > 0:
+                    # This is a stencil access, so mark the READ accesses
+                    # to account for reading other indices of the field
+                    var_accesses.add_access(signature, AccessType.READ, self)
```
hiker, line 980: We discussed adding 'artificial' accesses as children of the call node. I now presume that the previous change to 'independent iteration'  is required because of the removal of the accesses here?
I am ok with leaving this temporary, but then we need a ticket and #todo here (and above) to fix this up. I much prefer to run gocean kernels through the normal dependency analysis.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -255,41 +258,67 @@ def _partition(comp_ind1, comp_ind2, loop_variables):
...
-            # This can happen if there is a mixture of accesses to an array
-            # with and without indices, e.g.: a(i) = a*a
-            # In this case we don't add this to the partition, which will
-            # result in an empty partition (which in turns will disable
-            # parallelisation).
-            if i < len(indices_2):
-                partition_infos.append((indices_1[i].union(indices_2[i]),
-                                        [indx]))
+        for i1, idx_exprs in enumerate(comp_ind1):
+            for i2, _ in enumerate(idx_exprs):
```
hiker, line 311: I find that confusing. Would it be clearer (and more efficient ... not that that matters here :) ) to write `for i2 in range(len(idx_exprs)`?

src/psyclone/tests/domain/gocean/goloop_test.py
```diff
@@ -316,12 +316,25 @@ def test_independent_iterations(monkeypatch):
...
+        goloop.independent_iterations()
+    assert (
+        "Loop is incomplete. It should have exactly 4 children, but found"
+        " loop with 0 children" in str(err.value))
+
+    # Add dummy children
+    goloop.addchild(Literal("1", INTEGER_TYPE))
+    goloop.addchild(Literal("1", INTEGER_TYPE))
+    goloop.addchild(Literal("1", INTEGER_TYPE))
+    goloop.addchild(Schedule())
```
hiker, line 324: Should that better be a GOKernSchedule, since we are in gocean? Agreed, it probably doesn't matter, but it feels wrong.

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
hiker: As I've mentioned above, imho this should be improved, e.g. as Sergi suggested by adding reference as child of the call node? I am happy to accept this if we out a #todo here. I think it's too risky to just assume any single gocean kernel is parallelisable (see #2531)

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
```diff
@@ -193,15 +161,8 @@ def test_gok_access_info_scalar_and_property():
...
-    assert isinstance(comp_ind[2][1], Reference)
-    assert comp_ind[2][0].symbol.name == "i"
-    assert comp_ind[2][1].symbol.name == "j"
+    comp_ind = tmask[0].component_indices()
+    assert not comp_ind
```
hiker: I have to admit I had no idea that `tuple(tuple())` is empty :)
It might be a bit clearer to use `assert comp_ind is tuple(tuple())`?

src/psyclone/psyir/nodes/member.py
arporter: Oh yes. In my defence, I was thinking that Member and Reference weren't sub-classed but of course they are.

arporter:
I'm a bit confused as to the status of this PR. Is it OK to proceed with reviewing it @sergisiso and @hiker or should I look at the training PRs first?

sergisiso:
Moving it to 'in progress' I will address @hiker comments first.

doc/developer_guide/dependency.rst
sergisiso: This is part of the PSyIR now so I wouldn't duplicate it here:
image
I will be happy to link to this anywhere that is useful for context, but nowhere else in this page talks about indices.

src/psyclone/tests/domain/gocean/goloop_test.py
sergisiso: GOKernSchedule don't exist :) bodies of GOLoops are always Schedule

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
sergisiso: Good point, done

src/psyclone/gocean1p0.py
sergisiso: Ok, I think you are right, reverting this

src/psyclone/gocean1p0.py
sergisiso: > I much prefer to run gocean kernels through the normal dependency analysis.

I agree, I created https://github.com/stfc/PSyclone/issues/3219 and added a TODO here

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: I don't think I agree on this one, enumerate is more clear to me, but I can change it if you both disagree.

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
sergisiso: As suggested, I added a TODO and brought back (as a comment for now) the expected list of accesses once stencils are captured

src/psyclone/psyir/tools/dependency_tools.py
arporter: I'm happy with the `enumerate` - it's a common pattern.

doc/developer_guide/dependency.rst
arporter: I think we're OK for now. It feels lower-level than the rest of the methods documented here.

hiker:
> > > @hiker I've tried merging this with your training PR and, unsurprisingly, it broke it. I guess this is #3219 ? How would you like to proceed?
> > 
> > 
> > It should work. What breaks?
> > Here is an example from last week #3212
> 
> I thought I had tried tutorial 2.6 and 2.10, but it does indeed break (now??). It doesn't appear to be #3219, it considers the fields to be scalars, not arrays, so it doesn't even reach the state where the (missing) indices would be used. The fix in #3212 is to inline the kernels, but I think we all agree loop fusion should work without the need to inline (which requires lowering).
> 
> I'll try debugging to find more details.

It actually kind of is #3219. When trying to decide if the variable is an array or not, the code is expecting a `Reference` (to then check for `ArrayMixin`), but now/atm we get a `Kernel` object as the source of the access, so that code doesn't trigger.

Then the variable itself is not an arraymixin (since it's a structure type), but at this stage the fact that `%data` is used is not available (due to the missing `Reference`).  It should be possible to do another check and see if the kernel `argument_type` in this case is a field, depending on API? Though that might be a significant change of behaviour. But basically, querying the domain-specific information what a field is feels right :)

Even with this, we still need the References (otherwise the DA rightly says that the access to an array does not depend on the loop variable, but the variable is read and written, so we safely abort).

arporter:
OK, thanks for debugging @hiker. I shall merge this in.

PR #3166, opened 2025-09-29
(Closes #3120) sep25 release

### LINKED ISSUE #3120: September 2025 release
We need to make a release bringing in the functionality developed during NG-Arch.
This issue captures anything else that needs to get onto master or be checked before we do this.

arporter:
Copies what we do in fparser to automate uploading releases to PyPI. Workflow has been updated to use OpenID Connect (OIDC) so there's no need for stored secrets.
Updated fparser version to use to latest release.
Updated version number in version.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,25 +51,25 @@ PACKAGES = find_packages(where=SRC_PATH,
                                   "psyclone.tests.*"])
 
 NAME = 'PSyclone'
-AUTHOR = ('Rupert Ford <rupert.ford@stfc.ac.uk>, '
+AUTHOR = ('Rupert Ford, '
           'Andrew Porter <andrew.porter@stfc.ac.uk>, '
           'Sergi Siso <sergi.siso@stfc.ac.uk>')
 AUTHOR_EMAIL = 'andrew.porter@stfc.ac.uk'
 URL = 'https://github.com/stfc/psyclone'
 DOWNLOAD_URL = 'https://github.com/stfc/psyclone'
-DESCRIPTION = ('PSyclone - a compiler for Finite Element/Volume/Difference'
-               ' DSLs in Fortran')
+DESCRIPTION = ('PSyclone - a source-to-source Fortran compiler for HPC '
+               'applications')
 LONG_DESCRIPTION = '''\
-PSyclone is a compiler for Fortran-embedded Domain Specific Languages
-targetting Finite Element/Volume/Difference methods in earth-system
-modelling.
+PSyclone is a source-to-source Fortran compiler designed to programmatically
+optimise, parallelise and instrument HPC applications via user-provided
+transformation scripts.
 
 See https://github.com/stfc/psyclone for more information.
 '''
 LICENSE = 'OSI Approved :: BSD 3-Clause License'
 
 CLASSIFIERS = [
-    'Development Status :: 3 - Alpha',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -51,25 +51,25 @@
...
+AUTHOR = ('Rupert Ford, '
           'Andrew Porter <andrew.porter@stfc.ac.uk>, '
           'Sergi Siso <sergi.siso@stfc.ac.uk>')
 AUTHOR_EMAIL = 'andrew.porter@stfc.ac.uk'
 URL = 'https://github.com/stfc/psyclone'
 DOWNLOAD_URL = 'https://github.com/stfc/psyclone'
-DESCRIPTION = ('PSyclone - a compiler for Finite Element/Volume/Difference'
-               ' DSLs in Fortran')
+DESCRIPTION = ('PSyclone - a source-to-source Fortran compiler for HPC '
+               'applications')
```
hiker: TBH, I would prefer to keep the DSL aspect visible here. What about:
> PSyclone - a source-to-source Fortran compiler for HPC applications with domain-specific language extensions'

setup.py
```diff
@@ -51,25 +51,25 @@
...
 LONG_DESCRIPTION = '''\
-PSyclone is a compiler for Fortran-embedded Domain Specific Languages
-targetting Finite Element/Volume/Difference methods in earth-system
-modelling.
+PSyclone is a source-to-source Fortran compiler designed to programmatically
+optimise, parallelise and instrument HPC applications via user-provided
+transformation scripts.
```
hiker: And in the same spirit:
> It also offers domain-specific language extension for the Fortran language, which simplifies the implementation of Finite Elements / Volume / Difference codes.

setup.py
```diff
@@ -51,25 +51,25 @@
...
 
 See https://github.com/stfc/psyclone for more information.
 '''
 LICENSE = 'OSI Approved :: BSD 3-Clause License'
 
 CLASSIFIERS = [
-    'Development Status :: 3 - Alpha',
+    'Development Status :: 5 - Production/Stable',
```
hiker, line 80: Ah, I have been wanting to raise this for a long time :)

setup.py
hiker: Also it might be useful to add:
> PSyclone is used by the UK Meteorology Office for the development of their upcoming new weather model LFRic (link?), and for parallelising the Ocean code NEMO.

Let's show off that we do more than just "hello world"

hiker (changes requested):
I can't really comment on the `python_publish.yml` part, that needs to be checked by someone else.

setup.py
arporter: Good idea. I've changed to "...source-to-source and DSL Fortran compiler..." to try to make it clear we don't just do DSLs.

setup.py
arporter: Yes, done.

arporter: Also a good idea :-)

While we're here, would you like to be listed as a maintainer/author Joerg?

hiker: Sure.

hiker:
I am ok with the changes now, but won't approve since I didn't test the yml file, and I want to avoid a misunderstanding :)

setup.py
arporter: Added.

PR #3174, opened 2025-10-02
(#Towards 2668) transformations options on domain.common.transformations

LonelyCat124:


LonelyCat124:
Small PR moving a few more transformations to the new options style. Almost none of these have any keyword arguments, nor call any superclass so this PR is very small.

`AlgTrans` doesn't call the deprecation warning, as I expect its use of `self._invoke_trans.apply` to hit the deprecation warning (and the potential issue is it could hit that many times, but I don't see a reasonable solution to that).

@sergisiso @arporter Ready for a look from either of you - no rush as I expect this is fairly standalone and has no deadline, and I can work on other sections independently (unless i find an inheriting one which I'll leave for later).

PR #3176, opened 2025-10-02
(Towards #2668) Option changes for intrinsic transformations

LonelyCat124:
Again - none of these transformations have options supplied, so this is mostly a trivial change for future-proofing.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,4 +1,4 @@
-   6) PR #3174 towards #2668. Add kwargs to more transformations.
+   6) PR #3174 and #3176 towards #2668. Add kwargs to more transformations.
 
    5) PR #3186 for #3185. Small fix to SymbolicMaths.expand for Boolean
    expressions


src/psyclone/psyir/transformations/intrinsics/abs2code_trans.py
```diff
@@ -119,8 +124,12 @@ def apply(self, node, options=None):
         :type options: Optional[Dict[str, Any]]
```
sergisiso, line 124: I have some questions about what we decided to do with the old docstring. Should we remove the "options" from the docstring now? Should we remove all params from the validate if they are automatically inferred? And how come that the methods are not rendered in the documentation:

Image

Are we missing them in some `__all__` list?

sergisiso (changes requested):
Again, the changes look good @LonelyCat124 but I am wondering if we should update the docstring of those transformation in any way now.

src/psyclone/psyir/transformations/intrinsics/abs2code_trans.py
LonelyCat124: I had been leaving options as being present until they're removed but I can remove them if you feel like we should.

I think its all params other than node shouldn't be on validate. I'll take a look and try to work out why they're not rendered in the docs.

LonelyCat124: They don't have their own `__all__` lists at the bottom of their file (unlike InlineTrans). I will add them.

LonelyCat124: Fixed

PR #3182, opened 2025-10-09
(Closes #2772) fix for case where base of structure access is an array.

### LINKED ISSUE #2772: Problem parsing array of derived types
I've been going through UKCA and editing it so that PSyclone can parse it. For example, removing labelled loops. The one remaining thing that I'm unable to address is a case where there's an array of derived types. Parsing it gives:
```
TypeError: A StructureReference must refer to a symbol that is (or could be) a structure, however symbol 'chch_defs_master' has type 'Array<chch_t1: DataTypeSymbol, shape=[Reference[name:'n_chch_master']]>'
```

arporter:


sergisiso (changes requested):
@arporter I am not sure I agree with this fix. Or at least I want to discuss it first.

The proposed fix is inconsistent with what happens with Reference. In reference we allow References to array symbol. And then we apply Reference2ArrayRange. For a similar solution here we just need to allow the check at `src/psyclone/psyir/nodes/structure_reference.py` to pass with `Array<DataTypeSymbol or Unknown/unresolved types`. 

Another question is if we should go the opposite direction an always add ranges when we know it is an array (for both Reference and StructureReferences). I don't particularly disagree with this, in fact we already do this in "where constructs" in the _array_syntax_to_indexed method, and having this is a prerequisite of other transformations. So should we do it everywhere?

The advantage of the second is that we don't need to apply at the beginning of scripts as we do with NEMO, the disadvantage is that bug in this logic will become passthrough bugs - technically already is if it happens inside the where, and it will be harder to manually create this PSyIR constructs. (To be clear, none of the options allow us to assume the References/StructureReference do not point to an array, as they can still have unresolved symbols. We have a bugs regarding this because some of our transformations assume that if there are no indices is an scalar, which is not true.)

So, even if we don't fix the whole thing here, where do we want to go with this?

arporter:
Hi @sergisiso, yep, I'm happy to scale this back.

PR #3187, opened 2025-10-10
(Closes #3184) Improve logging when chasing imports.

### LINKED ISSUE #3184: Module inlining and symbol resolution
The behaviour of module inlining and symbol resolution is now quite complex.
If we use RESOLVE_IMPORTS = [...] in the transformation script then, when processing a given file, the frontend will resolve all symbols from any imports from the named modules. However, if we then proceed to attempt to module-inline a routine that is called from within the current file, when getting the PSyIR of that routine, we do not appear to do the same symbol resolution.

arporter:
I've also added/updated some logging information to make it easier to track whether or not we're succeeding in processing modules when resolving imports.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   16) PR #3187 for #3184. Improve logging when chasing imports and test
+   inining with the same imported symbol in the caller and the callee.
+
    15) PR #3211 towards #1540. Resolve indirect imports when requested
    by name.
 


src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -575,6 +575,10 @@ def apply(self, node, options=None, **kwargs):
                     newsym = container.symbol_table.lookup(code_to_inline.name)
                     newsym.visibility = Symbol.Visibility.PRIVATE
 
+            # Update any unresolved Symbols in the routine in case they are
+            # resolved in the Container scope.
+            code_to_inline.symbol_table.resolve_imports()
```
sergisiso: I still don't understand why we resolve symbols that are already resolved (in the _rm_imported_routine_symbol just above).

If they are unresolved they shouldn't pass the validate (as we cannot prove that they match the one in the calleer scope). Why is some form of code_to_inline.replace_symbol_with not enough? (I am asking because I tried unfortunately it does not, but I don't fully understand why not).

src/psyclone/parse/file_info.py
```diff
@@ -485,7 +450,8 @@ def get_fparser_tree(
         try:
             config = Config.get()
             reader = FortranStringReader(
-                source_code, include_dirs=config.include_paths
+                source_code, include_dirs=config.include_paths,
+                ignore_comments=(not config.frontend_keep_comments)
```
sergisiso: As I mentioned in the issue, I don't feel this is the behaviour I would expect.

Lets say b.f90 depends on a.f90 and we had to manually annotate some subroutines in a.f90 with "acc routine"

I would expect to compile that with

psyclone --keep-directives a.f90 ...
psyclone b.f90 ...

But with this change the previous commands will not consider directives in the a module, and therefore b will not offload loops with calls to those subroutines.

It is similar with comment, let say we know we have a bug in parsing comments in "file_with_issues.f90", now we can do:

psyclone --keep-comments a.f90
psyclone file_with_issues.f90
psyclone --keep-comments b.f90

With this change we lose the per-file control of the flags, we need to remove all "--keep-comments", as they may be imported from each other.

Note that the current behaviour of being false is also not what I expect. I think that actually the most expected behaviour may be to have them both as true when resolving imports. The inlining can have options to delete them if it really needs to.

src/psyclone/parse/file_info.py
```diff
@@ -499,58 +465,48 @@ def get_fparser_tree(
         # We directly call the cache saving routine here in case that the
         # fparser tree will be modified later on.
         if save_to_cache_if_cache_active:
-            self._cache_save(verbose=verbose)
+            self._cache_save()
```
sergisiso, line 473: Another problem with the current implementation is that the cache will have whatever is requested first, so if a.f90 and b.f90 depend on c.f90 an we do:

psyclone --enable-cache --keep-comments a.f90
psyclone --enable-cache b.f90

b.f90 will have comments from c.f90 imported code, even though we requested not to.

PS: my proposal above of always being true is more consistent because having comments or not will always depend on the "psyclone c.f90" flag, regardless of caching. Not from the a flag from an unrelated file depending if caching is enabled or not.

sergisiso (changes requested):
@arporter I have not done the proper review yet, I have only expanded on the comments I  had in the issue regarding my doubts wiht the proposed behaviour. I am open to change my mind.

src/psyclone/parse/file_info.py
arporter: Reverted this bit.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -1387,6 +1387,14 @@ def test_mod_inline_shared_wildcard_import(monkeypatch, fortran_reader):
...
       subroutine do_it()
         real, dimension(10) :: a
+        ! When finding the routine to inline, the frontend does *not* chase the
+        ! imports and therefore eps20 is Unresolved. However, since it must be
+        ! coming from one of the wildcard imports, those wildcard imports are
+        ! moved down into the routine scope. This is watertight for module
+        ! inlining.
+        ! However, if we subsequently want to inline the routine then this
+        ! becomes a problem that could be resolved by adding a flag to try to
+        ! resolve all symbols.
```
sergisiso: @arporter As we discussed, I would test if https://github.com/stfc/PSyclone/pull/3211 is sufficient

arporter: It is :-)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Reverted.

src/psyclone/parse/file_info.py
arporter: Also reverted :-)

PR #3190, opened 2025-10-14
Fix tests for compatibility with Python 3.14

nmnobre:
pytest seems to be twice as fast?! O.o
I've left python 3.9 and updated the docs to reflect that, though it'll be EOL in two weeks (as agreed in #3043).

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   8) PR #3190. Add python 3.14 in the CI and integration tests (except the
+   LFRic Extraction tests).
+ 
    7) PR #3181 for #3173. Adds protections for CodeBlocks in the
    DataSharingAttributeMixin.   
 


nmnobre:
I tried reproducing the error by executing what's in [here](https://github.com/MetOffice/fab/blob/c0f543ac2fbc04dd0f4e999ca3da9e63c97bb793/source/fab/parse/fortran_common.py#L133-L152) for one of the given files and... it worked :/

```
from fparser.two.utils import FortranSyntaxError
from fparser.two.parser import ParserFactory
from fparser.common.readfortran import FortranFileReader

fpath="/archive/psyclone-tests/latest-run/extraction_test/lfric_apps/applications/gungho_model/fab_workspace/gungho_model_extract-fast-debug-mpif90-gfortran/source/algorithm/diagnostics/checks_and_balances_alg_mod.parsable_x90"

f2008_parser = ParserFactory().create(std=None or "f2008")

reader = FortranFileReader(str(fpath), ignore_comments=False, include_omp_conditional_lines=True)

tree = f2008_parser(reader)
```

In any case, the good news is it's fparser throwing the exception... Unfortunately, I noticed that running the last command on that snippet more than once crashes - I'm not sure if that's related - but I'm also inclined to say that's not intended behaviour?

PR #3196, opened 2025-10-20
(closes #3178) Fixes to handle Directives from fparser

### LINKED ISSUE #3178: Comments include spacing
Given the following program as input
```
program test_prog
    integer :: i, j, k, l
    integer :: N
    integer, dimension(10,10) :: a, b, c, d, e, f, g, h

    !$acc comment1
    !comment1
    do i = 1, n, 1
        do j = 1, n, 1
            c = g(i + 1) * 9
            a(i) = b(i) + c
        enddo
    enddo

    !$acc comment2
    !comment2
    do i = 1, n, 1
        do j = 1, n, 1
            f(i) = a(i) / 2.0
        enddo
    enddo

    !$acc comment3
    !comment3
    do i = 1, n, 1
        do j = 1, n, 1
            k = 8
            h(i) = g(i - 1) + g(i + 1) + k
        enddo
    enddo
end program
```
the output includes spaces in the comments
```
 !$acc comment1



  ! comment1
  do i = 1, n, 1
   do j = 1, n, 1
     c = g(i + 1) * 9
     a(i) = b(i) + c
   enddo
  enddo



  ! $acc comment2
  ! comment2
  do i = 1, n, 1
   do j = 1, n, 1
     f(i) = a(i) / 2.0
   enddo
  enddo



  ! $acc comment3
  ! comment3
  do i = 1, n, 1
   do j = 1, n, 1
     k = 8
     h(i) = g(i - 1) + g(i + 1) + k
   enddo
  enddo
```

Reproducer:
```
from psyclone.psyir.miniposeidon.loop_split import LoopSplit
from psyclone.psyir.nodes import Loop, Assignment, Reference, Literal
from psyclone.psyir.symbols import DataSymbol

import pytest
from psyclone.psyir.transformations.transformation_error import TransformationError


def test_loop_split_to_norm_integration_nested():

    fortran_code = """
program test_prog
    integer :: i, j, k, l
    integer :: N
    integer, dimension(10,10) :: a, b, c, d, e, f, g, h

    !$acc comment1
    !$pos comment1
    !comment1
    do i = 1, n, 1
        do j = 1, n, 1
            c = g(i + 1) * 9
            a(i) = b(i) + c
        enddo
    enddo

    !$acc comment2
    !$pos comment2
    !comment2
    do i = 1, n, 1
        do j = 1, n, 1
            f(i) = a(i) / 2.0
        enddo
    enddo

    !$acc comment3
    !$pos comment3
    !comment3
    do i = 1, n, 1
        do j = 1, n, 1
            k = 8
            h(i) = g(i - 1) + g(i + 1) + k
        enddo
    enddo
end program
"""
    reader = FortranReader(ignore_comments=False, ignore_directives=False, last_comments_as_codeblocks=True)
    psyir = reader.psyir_from_source(fortran_code)

    print(psyir.debug_string())


test_loop_split_to_norm_integration_nested()
```

In addition, the first comment block is treated as a code block, whereas the other ones are stored in `preceding_comment`. I'd have expected that all are treated as `preceding_comment`.

LonelyCat124:
Blocked by https://github.com/stfc/fparser/pull/486

Will need to update PSyclone to use a specific commit of fparser once that is merged. Tests will fail in the mean time.

Main issue right now is this test fails, and I don't know if/how it should be solved:
```
    code = """subroutine x()
    !$omp firstprivate
    integer, dimension(100) :: i !dir$ aligned
    i = 1
    end subroutine x"""
    reader = FortranReader(ignore_comments=False, ignore_directives=False)
    routine = psyir.children[0]
    out = routine.debug_string()
    assert """!$omp firstprivate
integer, dimension(100) :: i  !dir$ aligned""" in out
```

These directives inside the declaration block have to become comments attached to the declarations, so they get bonus ` ` added into them, which may not be correct behaviour, but is not straightforward to fix (other than modifying the backend options).

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -413,7 +413,9 @@ linkcheck_ignore = [
     # the links to anchors to the main README.
     r'^https://github.com/stfc/PSyclone#',
     # Requires authentication.
-    r'^https://code.metoffice.gov.uk/trac'
+    r'^https://code.metoffice.gov.uk/trac',
+    # This often fails, but the link exists
+    r'^https://www.khronos.org'
 ]
 
 # -- Autodoc configuration ---------------------------------------------------


sergisiso:
> Will need to update PSyclone to use a specific commit of fparser once that is merged. Tests will fail in the mean time.

Sorry for the slow review on that one, but you could already update the submodule to point to the HEAD of your fparser branch here, so the tests would pass. In fact, I wanted to suggest this while looking at the fparser PR, as I was unsure what the best options would be there without looking at what they meant for psyclone.

sergisiso:
Maybe cloning does what the submodules file asks, but for git pulls I think you need to commit the new submodule hash. If I do it manually and do a git diff I get:
```
git diff
diff --git a/external/fparser b/external/fparser
index 00d276a6d5..7cf48ed76b 160000
--- a/external/fparser
+++ b/external/fparser
@@ -1 +1 @@
-Subproject commit 00d276a6d575d7c79411ba8ace0dc95cd3b505ab
+Subproject commit 7cf48ed76b613346960bc9e428907417c4736492
```
I think you need to commit this change.

LonelyCat124:
Yeah, I think this was essentially the failing behaviour also from #3196 - I'll need to adapt the regex code from fparser as well to remove the directive comments since at the moment PSyclone is just checking for `!$`.

LonelyCat124:
I believe I've fixed that issue now. I will make the directives in/around declarations as an xfailing test.

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "external/fparser"]
 	path = external/fparser
-	url = https://github.com/stfc/fparser.git
+    url = https://github.com/stfc/fparser.git
+    branch = 483_generic_directives
```
sergisiso: Reverse to fparser master now

doc/user_guide/psyclone_command.rst
```diff
@@ -520,6 +520,10 @@ some limitations:
      nodes. Also PSyclone will not know any details about these nodes
      (including that they contain directives) but this functionality will
      be improved over time.
+  3. Directives that appear before or inside the declaration section of a
+     Subroutine, Program or Module will be converted to comments if directives
+     are kept, and the default output for comments will add a space after the
+     ``!``.
 
 Note that using the ``keep-comments`` option alone means that any comments
```
sergisiso: Was already here but I would do:
add `--` before keep-comments
s/alone/without the ``--keep-directives`` options/
s/lost from the input./excluded./

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3370,14 +3358,20 @@ def _do_construct_handler(self, node, parent):
...
         for child in node.content:
             if isinstance(child, Fortran2003.Comment) and not found_do_stmt:
                 self.process_comment(child, preceding_comments)
                 continue
+            if isinstance(child, Fortran2003.Directive) and not found_do_stmt:
+                directive = self._directive_handler(child, None)
+                # Add the directive before the loop.
+                loop.parent.addchild(directive)
+                continue
+                # TODO
```
sergisiso: ?

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -606,18 +606,31 @@ def test_lost_program_comments():
     assert assignment.preceding_comment == "Comment here"
 
 
-def test_directive_at_end():
-    """Test that the FortranReader stores a directive after all
-    other code in a subroutine."""
+@pytest.mark.parametrize("directive", ["$omp target",
+                                       "$acc kernels",
```
sergisiso, line 610: Add one with "$pos" just to make sure we test Posidon directives.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -606,18 +606,31 @@ def test_lost_program_comments():
...
-    reader = FortranReader(ignore_comments=False, ignore_directives=False)
+  contains
+  subroutine test()
+
+    !$ a = 0 +     &
+    !$&  0
+    !{directive}
+    a = 1
+    a = 2
+    a = 3
```
sergisiso: This two statements are not needed

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3370,14 +3358,20 @@ def _do_construct_handler(self, node, parent):
...
+        # Keep track of the comments and directives before the 'do' statement
         loop_body_nodes = []
         preceding_comments = []
         found_do_stmt = False
         for child in node.content:
             if isinstance(child, Fortran2003.Comment) and not found_do_stmt:
                 self.process_comment(child, preceding_comments)
                 continue
+            if isinstance(child, Fortran2003.Directive) and not found_do_stmt:
+                directive = self._directive_handler(child, None)
```
sergisiso, line 3370: Just to see if I understand this correctly:
- the _directive_handler can be called from the self.handlers map when we find a directive in an execution regions (presumably a standalone directive),
- but it also has to be called from inside dos and ifs because region directives are inside the fparser node, not beside it, e.g.: `If_Construct(Directive('!$omp target'), , If_Then_Stmt(Level_4_Expr(Name('i'), `
?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5838,10 +5841,21 @@ def process_comment(self, comment, preceding_comments):
         :type preceding_comments: List[:py:class:`fparser.two.utils.Comment`]
 
         '''
+        _directive_formats = [
```
sergisiso: This is often called from ` for comment in walk(node, (Fortran2003.Comment, Fortran2003.Directive)):`

Why do we need to pass the Directive comments here? I though if we had "--keep-directives", it would be all deleted or recognised as Fortran2003.Comment. 

Whatever it is, maybe this method name and docstring needs updating now?

src/psyclone/tests/psyir/frontend/fparser2_directive_test.py
```diff
@@ -0,0 +1,209 @@
...
+    assert routine.children[0].debug_string() == "!$omp barrier\n"
+
+
+def test_directive_in_decls():
+    """Test that the FortranReader can handle a directive inside the
+    declarations"""
+    code = """subroutine x()
+    !$omp firstprivate
+    integer, dimension(100) :: i !dir$ aligned
+    i = 1
```
sergisiso, line 70: I didn't think about this in the fparser PR but what happens if we have an inlined comment that looks like a directive.

e.g. `a = a + 1   !$omp atomic`

This is just a comment, but I am worried that fparser converts it to a preceding comment and then to a valid Directive. We need a test for this.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -211,7 +211,7 @@ def get_symbol_names(self) -> List[str]:
         # For directives, we need to analyse all alphanumeric* parts of the
         # comment string and return any names that match a symbol in the
         # symbol table.
-        for node in walk(parse_tree, Fortran2003.Comment):
+        for node in walk(parse_tree, Fortran2003.Directive):
```
sergisiso, line 214: Aha, thats better.

.gitmodules
LonelyCat124: Done

doc/user_guide/psyclone_command.rst
LonelyCat124: Done

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I think this was leftover from earlier code.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Removed.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Yeah - fparser puts them wherever it would have put the comment before, so at the end of declarations, in the main block of an execution region, or attached to a construct.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: If this is called from `walk(... Fortran2003.Directive)` I should remove the Directive part of this I think (subject to any behaviour I may have forgotten) as that should have its own handler now.

Otherwise this should be ok, it should just process comments, and throw away those that are directives (which will feed through as Fortran2003.Comment if we don't request directive specialisation from fparser, in which case we need to throw them away in PSyclone. I can update the docstring to say that though.

src/psyclone/tests/psyir/frontend/fparser2_directive_test.py
LonelyCat124: My feeling on what is happening is this:
fparser passes through a `Fortran2003.Directive`, which gets added to the tree as a codeblock in the wrong place.

@sergisiso do you think we should fix this in fparser instead of working around it here?

LonelyCat124: I think we'll have to do it from fparser.

sergisiso: Yeah I agree

src/psyclone/psyir/frontend/fparser2.py
sergisiso: I added the comment in the code myself as this is likely the last comment

sergisiso (approved):
@LonelyCat124 Thanks for fixing the issue involving both repositories, this now manages properly most cases of directives inside execution statements.

sergisiso:
Just to me sure, I submitted the CI once more, as I am not sure if integration test were using the external fparser before, but they should not impact this PR as they have comments and directives disables anyway.

PR #3202, opened 2025-10-29
Update to Nvidia HPC SDK 26.1

nmnobre:


.github/workflows/nemo_tests.yml
```diff
@@ -202,8 +202,8 @@ jobs:
 
         # Set up envvars
         export PSYCLONE_NEMO_DIR=${GITHUB_WORKSPACE}/examples/nemo/scripts
-        export PSYCLONE_COMPILER=$MPIF90
         export MPIF90=psyclonefc
+        export PSYCLONE_COMPILER=$MPIF90
```
sergisiso: I cancelled the integration test because that is probably not what you want:
- MPIF90 is the envvar that the makenemo arch file uses
- PSYCLONE_COMPILER is the compiler that the psyclonefc wrapper uses

if you set them both to psyclonefc, it will have a recursive call.

nmnobre: Okay, reverted. I wasn't sure `MPIF90` was being set before, I can see now Spack is setting it. These scripts are full of (hidden) side-effects(...), and this resetting of `MPIF90` caught my eye.

nmnobre:
I've managed to build a minimal reproducible example and filed a bug report [here](https://forums.developer.nvidia.com/t/nvfortran-25-9-bug-when-compiling-with-openacc-support-enabled/349648) for the bug causing `nvfortran` to get stuck indefinitely.

nmnobre:
Unfortunately, nvfortran 25.11, whilst fixing the bug above, introduced a fresh new one. I filed a bug report [here](https://forums.developer.nvidia.com/t/nvfortran-25-11-internal-compiler-error-with-openacc-code/353711).

PR #3205, opened 2025-10-31
(Closes #3157, #3306) Initial implementation of maximal parallel region trans.

### LINKED ISSUE #3157: Transformation (or function?) to apply the largest possible parallel regions to a Routine
One clear thing that would be helpful for using nowaits is functionality to generate larger parallel regions, each spanning as large a section as possible.

We should have something (maybe a standalone transformation or similar) to do this instead of everyone having their own.

LonelyCat124:


LonelyCat124:
@MetBenjaminWent Chris said you had some cases worth trying this with as functionality tests?

LonelyCat124:
Yeah, I'd just been looking at the PR earlier today after I looked at this - I guess its still a while until we'd have it ready, but if it could handle cases like this it would definitely help (assuming the rest is otherwise ok).

sergisiso:
@LonelyCat124 Could we finish this PR without the Assignments as there are a few cases that are just contiguous loops that we could already do? And then do the Assignments in a follow up as they seem complicated and worth disucssing separately.

LonelyCat124:
> @LonelyCat124 I thought without assignments it would be encapsulating in a transformation what your already implemented in `nemo/eg7/openmp_cpu_nowait_trans.py`. Does this example also have the problem then?

Yeah it would probably be similar, but I was testing it with some code (https://code.metoffice.gov.uk/trac/lfric_apps/browser/main/trunk/science/physics_schemes/source/boundary_layer/bdy_impl3.F90) and we do a lot worse than the manual implementation. I think `omp_block = ceiling(tdims%j_end/real(omp_get_num_threads()))` is safe to have outside the parallel region, but since we can't put the following loop in an omp do anyway (maybe with force we can) it might not be important.

I'm happy to just go with this simple version initially though and then improve it, probably makes reviewing more straight forward. I'll write some tests for the current functionality and add some stuff to the docstrings.

sergisiso:
These last 2 would also be an issue without this transformation if the do the typical `for loop in psyir.walk(Loop): OMPLoopTrans`. In both cases (new transformation or previous approach) they need to be found and added the Parallel region independently. So I would still go ahead without the Assignments for this first PR.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
```diff
@@ -0,0 +1,191 @@
...
+    Loop,
+    IfBlock,
+    WhileLoop,
+)
+from psyclone.psyir.transformations.region_trans import RegionTrans
+from psyclone.psyir.transformations.transformation_error import \
+    TransformationError
+
+
+class MaximalParallelRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
```
sergisiso: Should this have the @transformation_documentation_wrapper and remove the validate params?

sergisiso: I am wondering if we should remove the "parallel" word from everywhere in this file? There is nothing specific to parallelism, it can be used for any outer_node that we want around a set of valid/invalid inner_nodes.

sergisiso: An example of this is in `examples/nemo/scripts/utils.py::add_profiling` which creates the largest posible ProfileNode to sections excluding Directives (maybe it would be good to add that to should is abstract class is generic enough)

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
```diff
@@ -0,0 +1,191 @@
...
+                current_block.append(child)
+            else:
+                # Otherwise, if the current_block contains any children,
+                # add them to a parallel region if we should and reset
+                # the current_block.
+                if current_block:
+                    for node in current_block:
+                        if node.walk(self._required_nodes,
+                                     stop_type=self._required_nodes):
+                            par_trans.apply(current_block)
```
sergisiso: What happens if this fails the validation? Maybe we want to catch it instead of propagate since we may already partially updated the tree.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
```diff
@@ -0,0 +1,191 @@
...
+                            break
+                    current_block = []
+                # Need to recurse on some node types
+                if isinstance(child, IfBlock):
+                    self.apply(child.if_body)
+                    if child.else_body:
+                        self.apply(child.else_body)
```
sergisiso: It doesn't seem right that the apply is recursive because each apply calls its own validate, and validate is also recursive. So the validations currently grow quadratically.

It maybe better to move that to an independent recursive method and call that after the validation is complete.

src/psyclone/psyir/transformations/omp_maximal_parallel_region_trans.py
```diff
@@ -0,0 +1,80 @@
...
+        OMPLoopDirective,
+        OMPTaskDirective,
+        DynamicOMPTaskDirective,
+)
+from psyclone.psyir.transformations.maximal_parallel_region_trans import (
+        MaximalParallelRegionTrans)
+from psyclone.psyir.transformations.omp_parallel_trans import OMPParallelTrans
+
+
+class OMPMaximalParallelRegionTrans(MaximalParallelRegionTrans):
```
sergisiso: I would prefer the "OMPParallel" words to be together, to make clear it adds OMPParallel directives.

I know I proposed the Maximal word, but do you think it may be confused to mean "the optimally largest" regions? - that we don't do.

Is there any more fitting word we could use?

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
sergisiso: Alternatively, you could use something akind to the sets that you used in `openmp_cpu_nowait_trans.py` and the validate populates the set and the apply applies it which is not recursive

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
sergisiso: For example if the region directive is a OMPTarget and it finds a call to a non-offloaded subroutine.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: Yes, and I should update RegionTrans if it needs it (though is suspect its done).

LonelyCat124: Yeah, that seems reasonable, I'll take a look at add_profiling and see if it can be modified to fit.

src/psyclone/psyir/transformations/omp_maximal_parallel_region_trans.py
LonelyCat124: I can't think of a better work - I can add that the maximal is "for what PSyclone can compute" in the docs but otherwise its hard.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: Yeah, when I was doing the recursive apply Iforgot about the validation steps, thats pretty bad. I'll take a look at the nowait transformation and see if I can reuse (slash remember) the logic as a non recursive method is probably better if it doesn't involve significantly more complex code.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: I need to think if the subclasses need it or not. I think they do (and to therefore to define a very basic apply method).

src/psyclone/psyir/transformations/omp_minimise_sync_trans.py
```diff
@@ -537,6 +549,9 @@ def apply(self, node: Routine, **kwargs) -> None:
             # if its a OMPBarrierDirective as they are unnecessary.
             for parallel in node.walk(OMPParallelDirective):
                 _eliminate_final_parallel_barrier(parallel)
+            # Finally eliminate any barriers leftover outside of parallel
+            # regions, as these are now superfluous
+            self._eliminate_uncontained_barriers(node)
```
LonelyCat124, line 554: @sergisiso This should resolve #3306.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: Done.

LonelyCat124: Renaming done, need to look at add profiling still.

src/psyclone/psyir/transformations/omp_maximal_parallel_region_trans.py
LonelyCat124: I changed this to MaximalOMPParallel... for now, can continue to iterate if we have a better name.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: @sergisiso I think add_profiling needs something like this (though maybe there's just a better explicit correct option?:
```
    from psyclone.psyir.transformations import MaximalRegionTrans
    import itertools
    def all_subclasses(cls):
        return set(cls.__subclasses__()).union(
            [s for c in cls.__subclasses__() for s in all_subclasses(c)])
    disallowed_classes = itertools.chain(all_subclasses(Directive),
                                         all_subclasses(Return),
                                         all_subclasses(ScopingNode),
                                         [Directive, Return, ScopingNode,
                                          Loop, IfBlock])
    class MaximalProfilingTrans(MaximalRegionTrans):
        '''Applies Profiling to the largest possible region.'''
        _allowed_nodes = [cls for cls in Node.__subclasses__() if
                          cls not in disallowed_classes]
```

What do you think for allowed nodes - is this best or shall I just do e.g. `Assignment, Call, ???` I'm trying to think what other statements are needed (since ifblock/loops/while_loops are handled anyway).

LonelyCat124: I settled on
```
    class MaximalProfilingTrans(MaximalRegionTrans):
        '''Applies Profiling to the largest possible region.'''
        _allowed_nodes = [Assignment, Call, CodeBlock]
        _transformation = ProfileTrans
```
for now, we can adjust this later if we find it to be insufficient.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: I've updated this now to try validating all blocks before applying the transformation, however I'm not sure what should be done when the validation fails? I wonder about having a function to call (which can either do nothing and just skip the apply method for this block) or raise a TransformationError or whatever the transformation decides?

LonelyCat124: I've gone with that implemetation for now.

src/psyclone/psyir/transformations/maximal_parallel_region_trans.py
LonelyCat124: The `openmp_cpu_nowait_trans.py` on master still uses recursion for loops and ifblocks, so I think there's no escaping that recursion, so I'll make this its own routine instead of part of apply.

LonelyCat124:
@sergisiso This is ready for another look when you're back off leave.

examples/nemo/scripts/utils.py
```diff
@@ -487,55 +492,4 @@ def add_profiling(children: Union[List[Node], Schedule]):
...
-               "was_single_stmt" in nodes[0].annotations and \
-               isinstance(nodes[0].if_body[0], CodeBlock):
-                # We also don't put single statements consisting of
-                # 'IF(condition) CALL blah()' inside profiling regions
-                return
-        try:
-            ProfileTrans().apply(nodes)
-        except TransformationError:
-            pass
+    MaximalProfilingTrans.apply(children)
```
sergisiso: Using NEMO with this currently fails with:
```
  File "/home/HCHPSE01/nxn02/sss33-nxn02/workspace/psyclone/PSyclone/examples/nemo/scripts/utils.py", line 495, in add_profiling
    MaximalProfilingTrans.apply(children)
TypeError: MaximalRegionTrans.apply() missing 1 required positional argument: 'nodes'
```
if I add parenthesis to instantiate the class I get:
```
  File "/home/HCHPSE01/nxn02/sss33-nxn02/workspace/psyclone/PSyclone/examples/nemo/scripts/utils.py", line 495, in add_profiling
    MaximalProfilingTrans().apply(children)
  File "/home/HCHPSE01/nxn02/sss33-nxn02/workspace/psyclone/PSyclone/src/psyclone/psyir/transformations/maximal_region_trans.py", line 231, in apply
    all_blocks = self._compute_transformable_sections(node_list)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/HCHPSE01/nxn02/sss33-nxn02/workspace/psyclone/PSyclone/src/psyclone/psyir/transformations/maximal_region_trans.py", line 133, in _compute_transformable_sections
    if self._can_be_in_region(child):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/HCHPSE01/nxn02/sss33-nxn02/workspace/psyclone/PSyclone/src/psyclone/psyir/transformations/maximal_region_trans.py", line 90, in _can_be_in_region
    if isinstance(node, self._allowed_nodes):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```
L479 needs to be converted to a tuple, then it works

examples/nemo/scripts/utils.py
```diff
@@ -474,6 +474,11 @@ def add_profiling(children: Union[List[Node], Schedule]):
         attempt to add profiling regions.
 
     '''
+    class MaximalProfilingTrans(MaximalRegionTrans):
+        '''Applies Profiling to the largest possible region.'''
+        _allowed_nodes = [Assignment, Call, CodeBlock]
```
sergisiso: Looking at the outcome it has some differences from the previous implementation due to not having:
```
       if any(ignore in routine_name for ignore in PROFILING_IGNORE):
            return
        if len(nodes) == 1:
            if isinstance(nodes[0], CodeBlock) and \
               len(nodes[0].get_ast_nodes) == 1:
                # Don't create profiling regions for CodeBlocks consisting
                # of a single statement
                return
            if isinstance(nodes[0], IfBlock) and \
               "was_single_stmt" in nodes[0].annotations and \
               isinstance(nodes[0].if_body[0], CodeBlock):
                # We also don't put single statements consisting of
                # 'IF(condition) CALL blah()' inside profiling regions
                return
```
These are important to keep the profile fast and clean.

In a sense this are similar to your _required_nodes (minimum condition to be a region). Do you think we could add (or replace it with) a new: `def satisfies_minimum_region_rules(node) -> bool:` that could include this kind of logic?

I am unsure if "allowed_nodes" also needs to be a subroutine, as mentioned in the issue MaximalOMPParallelTrans could also need that for 'calls that we need to differentiate between pure and not pure' or 'assignments to shared or private variables', ...

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,245 @@
...
+
+    The _transformation should be a transformation class to apply to
+    the computed set of regions.
+
+    The _allowed_nodes is a tuple of Node classes that are allowed as
+    statements in the transformed region. Note that upon finding a Loop or
+    IfBlock, the node's children will be checked to determine whether its safe
+    to contain the Loop or IfBlock in the transformed section.'''
+
+    # The type of transformation to be applied to the input region.
```
sergisiso: Use "#:" to document these 3 class parameters

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,245 @@
...
+    _transformation = None
+    # Tuple of statement nodes allowed inside the _transformation
+    _allowed_nodes = ()
```
sergisiso: Maybe call it allowed_contiguous_nodes, the current name can be confusing as it could mean they are not allowed anywhere inside the region, but we are just talking about the top level here  - with the documentation also mention that loops an ifs are always included a the top-level check will be the nodes in the bodies (I think)

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
```diff
@@ -0,0 +1,93 @@
...
+        OMPTaskDirective,
+        DynamicOMPTaskDirective,
+    )
+
+    def apply(self, nodes: Union[Node, Schedule, list[Node]], **kwargs):
+        '''Applies the transformation to the nodes provided.
+
+        :param nodes: can be a single node, a schedule or a list of nodes.
+        '''
+        super().apply(nodes, **kwargs)
```
sergisiso, line 93: Could you remind me why this empty/super apply methods are needed in order to for the transformation_documentation to work?

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,245 @@
...
+            region.
+
+        :returns: whether it is safe to add the node to a transformed region.
+        '''
+
+        if isinstance(node, self._allowed_nodes):
+            return True
+
+        if isinstance(node, (Loop, WhileLoop)):
+            # Recurse through the loop body.
```
sergisiso: Maybe "Check that all contents of the loop body can be part of the region"

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,245 @@
...
+            for child in node.loop_body:
+                if not self._can_be_in_region(child):
+                    break
+            else:
+                return True
+            return False
+
+        if isinstance(node, IfBlock):
+            # Recurse through the if_body and else_body
```
sergisiso: Maybe: "Check that all contents of each branch body be part of the region"

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,245 @@
...
+        default implementation removes the block from the list of blocks
+        and continues.
+
+        :param err: The TransformationError raised by the transformation's
+            validate function.
+        :param block: The block that failed validation.
+        :param all_blocks: The list of all of the blocks found during
+            application of this transformation.
+        '''
+        all_blocks.remove(block)
```
sergisiso: I see how this function may be needed in the future, but since we don't have a use case for it (all current examples use this default implementation), I would delete this and just add the "all_blocks.remove(block)" inside the except that calls it.

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # Check that the transformation can be applied to all of the found
+        # blocks.
+        for block in all_blocks[:]:
+            try:
+                par_trans.validate(block)
+            except TransformationError as err:
+                # Perform class specific behaviour for a block that fails
+                # transformation validation.
+                self._handle_invalid_block(err, block, all_blocks)
```
sergisiso: I am wondering if this could be done somewhere inside the computation of blocks. If I understand correctly the blocks can now return a large region that may have one transformation blocking node in the middle, and the whole region will be skipped. If this was part of the region computation, we could still create the regions around this invalid node.

src/psyclone/tests/psyir/transformations/maximal_ompparallel_region_trans_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_maximal_ompparallel_region_trans_apply(fortran_reader):
+    ''' Test the apply method of the ompparallel region transformation.'''
+    code = """subroutine x
+    integer :: i
+    i = 1
+    end subroutine x"""
+    psyir = fortran_reader.psyir_from_source(code)
+    MaximalOMPParallelRegionTrans().apply(psyir.children[0].children[:])
+    assert len(psyir.walk(OMPParallelDirective)) == 0
```
sergisiso, line 57: Also add some tests where the transformation adds some directives, and with nested ifs/loops as this are relevant for the transformation.

sergisiso: Oh and the block of only barriers since you mention it in a comment.

sergisiso (changes requested):
@LonelyCat124  Thanks for renaming and adding the Profiling example, I made a few more comments for the implementation, but it is getting closer for the basic functionality.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: The main reason is all of the code is assume to rely on the presence of `cls.apply` when doing things such as inheriting the docstirng and generating the docstring for the validate function.

I'd also assumed it was needed for `get_valid_options` but I was smarter when coding that and it in fact uses the mro to find all of the `apply` methods in the hierarchy and skips over those that don't have one.

This is a flaw with the transformation_documentation_wrapper then - we could do better and only update `apply` if its on `cls`, and update validate according to the first apply method we find looping up the mro (similar to `psyGen.py::2751-2754`). I'm happy to fix this up soon if you'd like as I think it could be a relatively small task - if so can you open an issue and I'll also add a TODO to remove this apply method once that issue is resolved?

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: I can't think of a counter-example for a Transformation which we'd want to do a maximal region operation where:
`par_trans.validate(block)` will succeed but `all(par_trans.validate(element) for element in block)` won't, so I'll give this a try - it would give us much better coverage if so.

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Removed since the structure has changed.

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done.

examples/nemo/scripts/utils.py
LonelyCat124: Fixed.

examples/nemo/scripts/utils.py
LonelyCat124: Added

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done.

src/psyclone/tests/psyir/transformations/maximal_ompparallel_region_trans_test.py
LonelyCat124: Added

examples/nemo/scripts/utils.py
```diff
@@ -477,6 +477,44 @@ def add_profiling(children: Union[List[Node], Schedule]):
         attempt to add profiling regions.
 
     '''
+    class MaximalProfilingTrans(MaximalRegionTrans):
+        '''Applies Profiling to the largest possible region.
```
sergisiso: Maybe call it:
```
class MaximalProfilingOutsideDirectives(MaximalRegionTrans):
''' Applies profiling hooks to the largest possible region outside directives regions '''
```
to make clear what are we going for here.

examples/nemo/scripts/utils.py
```diff
@@ -477,6 +477,44 @@ def add_profiling(children: Union[List[Node], Schedule]):
...
+
+        :param routine_name: The name of the Routine being profiled.
+        '''
+        _allowed_contiguous_nodes = (Assignment, Call, CodeBlock)
+        _transformation = ProfileTrans
+
+        def __init__(self, routine_name: str = ""):
```
sergisiso: I know this may not be possible if we bring this into its own file, but for now the routine_name doesn't feel quite right as an argument, I would prefer the name check to be outside, where it is called:
```
if routine_name not in PROFILING_IGNORE:
    MaximalProfilingTrans().apply(children)
```

examples/nemo/scripts/utils.py
```diff
@@ -477,6 +477,44 @@ def add_profiling(children: Union[List[Node], Schedule]):
         attempt to add profiling regions.
 
     '''
+    class MaximalProfilingTrans(MaximalRegionTrans):
+        '''Applies Profiling to the largest possible region.
+
+        :param routine_name: The name of the Routine being profiled.
+        '''
+        _allowed_contiguous_nodes = (Assignment, Call, CodeBlock)
```
sergisiso: Maybe add a comment saying: "We purposely don't encompase Directive, or Return statements (which would create unclosed hooks)."

Also another picky name change but "_allowed_contiguous_statements" may be better (as we don't really validate other kinds inner nodes I think)

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,262 @@
...
+from psyclone.psyir.transformations.region_trans import RegionTrans
+from psyclone.psyir.transformations.transformation_error import \
+    TransformationError
+from psyclone.utils import transformation_documentation_wrapper
+
+
+@transformation_documentation_wrapper
+class MaximalRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
+    '''Abstract transformation containing the functionality to add
+    the largest allowed transformation to the provided code segment.
```
sergisiso: "Abstract transformation containing the functionality to apply another transformation to the largest code segment possible while satisfying its validation and any additionally provided constraint." ?

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,262 @@
...
+    Subclasses should override the _transformation and
+    _allowed_contiguous_nodes members to control the functionality.
+
+    The _transformation should be a transformation class to apply to
+    the computed set of regions.
+
+    The _allowed_contiguous_nodes is a tuple of Node classes that are allowed
+    as statements in the transformed region. Note that upon finding a Loop or
+    IfBlock, the node's children will be checked to determine whether its safe
+    to contain the Loop or IfBlock in the transformed section.'''
```
sergisiso: I think this is now sufficinenlty documented with the "#:", I would remove it from here.

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -0,0 +1,262 @@
...
+from psyclone.utils import transformation_documentation_wrapper
+
+
+@transformation_documentation_wrapper
+class MaximalRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
+    '''Abstract transformation containing the functionality to add
+    the largest allowed transformation to the provided code segment.
+
+    Subclasses should override the _transformation and
+    _allowed_contiguous_nodes members to control the functionality.
```
sergisiso: and _required_nodes ?

examples/nemo/scripts/utils.py
```diff
@@ -477,6 +477,44 @@ def add_profiling(children: Union[List[Node], Schedule]):
...
+        _transformation = ProfileTrans
+
+        def __init__(self, routine_name: str = ""):
+            super().__init__()
+            self._routine_name = routine_name
+
+        def _satisfies_minimum_region_rules(self, region: list[Node]) -> bool:
```
sergisiso, line 488: There are still some differences from the previous implementation. The most common pattern is that something that before looked like:
```
    if (TRIM(cdrw) == 'WRITE') then
      CALL profile_psy_data_8 % PreStart("zdfosm", "osm_rst-r8", 0, 0)
      if (lwp) then
        ! PSyclone CodeBlock (unsupported code) reason:
        !  - Unsupported statement: Write_Stmt
        WRITE(numout, *) '---- osm-rst ----'
      end if
      call iom_rstput(kt, nitrst, numrow, 'wn', ww)
      call iom_rstput(kt, nitrst, numrow, 'hbl', hbl)
      call iom_rstput(kt, nitrst, numrow, 'dh', dh)
      if (ln_osm_mle) then
        call iom_rstput(kt, nitrst, numrow, 'hmle', hmle)
      end if
      CALL profile_psy_data_8 % PostEnd
      return
    end if
```
Now does not have any profiling hook inserted. Do you know why not?

sergisiso (changes requested):
It is almost there now @LonelyCat124 , just some more minor things to clean up.

examples/nemo/scripts/utils.py
LonelyCat124: @sergisiso Can you link me the file and I'll try to debug, this should just be CodeBlock + Call so i would have thought we should get something inside the if block.

examples/nemo/scripts/utils.py
LonelyCat124: Changed.

examples/nemo/scripts/utils.py

examples/nemo/scripts/utils.py

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done

examples/nemo/scripts/utils.py
LonelyCat124: This is done now, e.g. in zdftke:
```
    else
      CALL profile_psy_data_6 % PreStart("zdftke", "tke_rst-r6", 0, 0)
      if (TRIM(cdrw) == 'WRITE') then
        if (lwp) then
          ! PSyclone CodeBlock (unsupported code) reason:
          !  - Unsupported statement: Write_Stmt
          WRITE(numout, *) '---- tke_rst ----'
        end if
        if (lwxios) then
          call iom_swap(cwxios_context)
        end if
        call iom_rstput(kt, nitrst, numrow, 'en', en, ldxios=lwxios)
        call iom_rstput(kt, nitrst, numrow, 'avt_k', avt_k, ldxios=lwxios)
        call iom_rstput(kt, nitrst, numrow, 'avm_k', avm_k, ldxios=lwxios)
        call iom_rstput(kt, nitrst, numrow, 'dissl', dissl, ldxios=lwxios)
        if (lwxios) then
          call iom_swap(cxios_context)
        end if
      end if
      CALL profile_psy_data_6 % PostEnd
    end if
```

Was an issue with indentation.

Issue #868, opened 2020-08-13
Support custom function spaces in LFRic (multi-data fields, different number of layers and data)


TeranIvy:
~~Relevant LFRic documentation is in [this LFRic wiki](https://code.metoffice.gov.uk/trac/lfric/wiki/LFRicInfrastructure/FieldEnhancements/MultidataPsycloneMetaProposal).~~

**Update**: The support will require a redesign of function space representation. See [this new LFRic wiki](https://code.metoffice.gov.uk/trac/lfric/wiki/LFRicInfrastructure/FieldEnhancements/FunctionSpaceMetadata).

**Update, 21/10/2025**: The design was discussed with STFC and the idea is to not have custom function spaces to capture the changes in `nlayers` and `ndata` as they are properties of the field and not function space continuity (the design of LFRic stores dofmaps in function space object). The current conclusion is to have additional metadata in the field argument types.

TeranIvy:
- [ ] Metadata parsing support and related documentation,
- [ ] Kernel arguments for `k`-first fields (declarations, call argument list, stub argument list),
- [ ] Kernel arguments for `i`-first fields (declarations, call argument list, stub argument list),
- [ ] Kernel arguments for operators (declarations, call argument list, stub argument list),
- [ ] Kernel argument lists for basis functions/quadrature (declarations, call argument list, stub argument list).

arporter:
Having discussed with @TeranIvy, we have agreed that we will need some recognised names for numbers of layers because there are physics 'kernels' with ~100 arguments but only two or three different values of nlayers - we don't want to have to extend the argument list as that would be massive.

arporter:
Therefore, we need an agreed set of names that will be added to argument_mod.F90 and recognised by PSyclone.

arporter:
@sergisiso has pointed out that `LFRicKernelConstTrans` could make use of the nlayers metadata rather than relying on a user providing a value (provided of course that the metadata specifies a value rather than a name). Note that that transformation is still in `psyclone.transformations` so needs to be moved first.

The transformation could also be extended to know about `ndata`.

arporter:
Hi @tommbendall @TeranIvy and @stevemullerworth, is the idea that this new metadata would apply to operator arguments too?

stevemullerworth:
[...]

PR #3209, opened 2025-11-05
(Towards #868) Metadata support for fields with nlayers and ndata

arporter:


doc/user_guide/lfric.rst
```diff
@@ -162,24 +162,30 @@ least rank (number of dimensions) one. Scalar arrays are identified with
...
+
 LFRic fields passed as arguments to any :ref:`LFRic kernel
 <lfric-kernel-valid-data-type>` can be of ``real`` or ``integer``
 primitive type. In the LFRic infrastructure, these fields are
 represented by instances of the ``field_type`` and ``integer_field_type``
 classes, respectively.
 
+Typically, a field will have the same number of vertical layers as the
+model mesh. However, this is not a requirement and the number of layers
+can be as few as one (a 2D field).
```
stevemullerworth: I would suggest avoiding the idea that there is a particular model mesh. For example, we will have apps that regrid data from one number of levels to another.

```suggestion
While an application may model a set of fields that all have the same number of 
vertical levels, this is not a requirement. An application may have fields on a range
of different levels including fields on a single level (a 2D field).
```

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1753,45 @@ meshes cannot be on the same function space while those on the same
 mesh must also be on the same function space.
 
 
+Number of Layers Metadata
+"""""""""""""""""""""""""
+
+If a particular field argument to a kernel has a number of vertical levels
+that is not the same as the extruded mesh then this must be specified using
```
stevemullerworth: I may be misunderstanding this. I am assuming it relates to a kernel which, for example, takes an argument which may be, say, an 70-level field but could be a 50-level or 40-level field. But the kernel should only set 40 levels (or fewer) of the field. Most typically, perhaps, it would be useful when setting the bottom level of a field of any number of levels.

```suggestion
that may be different from the mesh of the input field then this must be specified using
```

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1753,45 @@ meshes cannot be on the same function space while those on the same
...
+values associated with each DoF. This is indicated in the field metadata by
+the optional ``NDATA`` argument to GH_FIELD, e.g.::
+
+  arg_type(GH_FIELD, GH_REAL, GH_READ, W2, NDATA=4)
+
+The value specified for ``NDATA`` may be a literal if it is known at
+compile time. Alternatively, it may be given the special value
+``GH_RUNTIME`` which means that the number of data values at each DoF
+is to be determined at runtime by querying the field object (in the
+generated PSy layer).
```
stevemullerworth: For LFRic atmosphere, It would be useful to have the same flexibility for giving named values as is done for NLAYERS: in fact, it is more useful to have the flexibility here than for NLAYERS as there is less mixing and matching of layer numbers. 

We have a lot of physics kernels with ~100 fields with a handful of multidata choices distinguished by ANY_DISCONTINUOUS_SPACE_1, ANY_DISCONTINUOUS_SPACE_2 etc. If they are converted to GH_RUNTIME we would need separate dofmaps etc. for each of the many fields rather than for each of the ANY_DISCONTINUOUS_SPACE choices. Example kernel here [MOSRS password protected]:

https://code.metoffice.gov.uk/trac/lfric_apps/browser/main/trunk/interfaces/physics_schemes_interface/source/kernel/bl_exp_kernel_mod.F90

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1753,45 @@ meshes cannot be on the same function space while those on the same
...
+``GH_NLAYERSM1`` TODO) or the special value
+``GH_RUNTIME``. A named value means that the number of levels is to be
+determined at runtime by querying the field object (in the generated
+PSy layer). If two different field arguments are on the same function
+space but both have ``NLAYERS=GH_RUNTIME`` then it is assumed that
+they may have *different* values of ``NLAYERS`` and hence a separate
+dofmap is passed to the kernel for each. However, if two or more field
+arguments are on the same function space and have the same, named number
+of layers which is not ``GH_RUNTIME`` then only one dofmap is passed to
+the kernel for those arguments.
```
stevemullerworth:
```suggestion
of layers which is not ``GH_RUNTIME`` then only one dofmap, the dofmap of the 
first field listed in the metadata, is passed to the kernel for those arguments.
```

stevemullerworth (changes requested):
One of my changes (for multi-data fields) suggests a change to the implementation. The others are document changes only, I think.

doc/user_guide/lfric.rst
tommbendall: Just commenting that I agree with this point! In fact I think the most common situation is that `NDATA` will be a runtime variable, but there might be many multidata fields being passed as arguments the same kernel, but all using the same `NDATA` value.

I'm not sure what the best solution to this is... because the `NDATA` values are science-related, I don't think we want any specific values hard-coded in `arguments_mod`.

Could we define a local variable for `NDATA` to take? e.g. 
```
type(ndata_type) :: LAND_TILES
type(ndata_types) :: NUM_AEROSOLS

arg_type(GH_FIELD, GH_REAL, GH_READ, W2, NDATA=4)
type(arg_type) :: meta_args(4) = (/  &
     arg_type(GH_FIELD, GH_REAL, GH_READWRITE, W3),  &
     arg_type(GH_FIELD, GH_REAL, GH_READ,  W3, NDATA=LAND_TILES), &
     arg_type(GH_FIELD, GH_REAL, GH_READ,  WTHETA, NDATA=NUM_AEROSOLS), &
     arg_type(GH_FIELD, GH_REAL, GH_READ,  W3, NDATA=NUM_AEROSOLS)  &
/)
```

stevemullerworth: This is the sort of thing I had envisaged (though originally I was thinking of a direct but more meaningful replacement for ANY_*SPACE entries).

One of the challenges of moving away from ANY_DISCONTINUOUS_SPACE settings was deciding how and where to name scientifically-meaningful versions given that ``arguments_mod`` is in core. One `could` have a physics module to store them which would enable a common set to be used by several kernels.

I think they would need to be initialised so as not to cause any compiler warnings? If so, that's another reason for hiding them in a module.

arporter: Great, that's the same as I was imagining for NLAYERS so it's straightforward to do from a PSyclone point of view. I can see your issue with names in core and physics but I'll let you wrangle that :-)

tommbendall: Something else to consider that Tom M has just mentioned is that we have two types of multidata field. Those with `ndata_first` and those without. We may also need to capture that

doc/user_guide/lfric.rst
arporter: I don't know what the use case is I'm afraid. However, when we discussed with Chris and Iva we were definitely talking about fields that did not have e.g. 70 levels because that then changed the values in the dof map. On the other hand, if all fields actually have the same number of levels but only populate a subset of them then all of their dofmaps will be identical (for a given FS). I think this needs clarification from @TeranIvy or @tommbendall?

arporter: Coming back to this now, we recently agreed with @TeranIvy and @christophermaynard that the nlevels extension was the more straightforward and we would do this first. However, I've now re-discovered this thread and realised I still need an answer to the question: does a field with e.g. 40 levels actually have 70 but only makes use of the first 40? If so, it will have the same dofmap as the 70-level field.
Then there's the special case of 2D fields - perhaps this answers my first question as presumably you don't only make use of the 1st of 70 levels as that's a lot of wasted storage? Tagging @stevemullerworth and @tommbendall too.

stevemullerworth: Each field needs to have correct metadata in the kernel. The metadata can be used to identify the correct dofmap and the correct nlayers for the field. That's a strict rule.

An application can create fields with a 40-level extrusion and a 70-level extrusion (and a 2d field which in practice is a 1-level "extrusion"). Each extrusion requires a different 3D mesh and gives you a different nlayers, dofmap etc.

My suggested change refers to a kernel that, for example, applies `fieldinput` to `fieldoutput`.  `fieldoutput` has 70 levels. `fieldinput` has 40 levels. So the kernel API could have two nlayers arguments and two sets of dofmap arguments. The two fields would be associated with different LFRic function spaces that are each attached to different 3D meshes and so different extrusions.

It's up to the kernel to know that it is looping over the `fieldinput` nlayers value and not the `fieldoutput` nlayers value.

Technically, one could write a kernel that loops over a subset of layers. But in that case, the number of layers would need to be a separate kernel input and the nlayers input would be ignored (except to check that we're not looping over more than nlayers levels).

arporter: Thanks Steve, I understand things a bit more now. Currently, we pass `nlayers` obtained from the first field/operator kernel argument:
image
I think what we're saying therefore is that any field/operator that has a number of levels that differs from the first field/operator argument must have `nlayers=xx` in its metadata. In turn, this will mean an additional dofmap is supplied to the kernel for each distinct combination of function space and nlayers. A kernel developer is of course free to specify `nlayers` for every field/operator kernel argument. That would just mean we would pass one of the dofmaps to the kernel twice but that's not a massive overhead (as opposed to requiring that all kernels be re-written).

arporter: Or, have I misunderstood and the proposal *is* to mandate that all kernel metadata specify `nlayers` for every field/operator argument?

tommbendall: Just catching up on this. A change to enforce `nlayers` metadata for every field/operator would be a very daunting task! So I would support the former proposal: only including an `nlayers` metadata argument for the fields/operators that differ from the first field/operator.

For existing kernels that have fields/operators on different meshes, we are currently managing this by declaring the fields/operators to be on different function spaces.

arporter:
I've had a play and the old metadata parser in PSyclone is quite happy to accept quoted strings, e.g.
```fortran
arg_type(gh_field, gh_real, gh_read, w3, nlevels='double')
```
This means there's no need to have pre-declared and initialised integer values. However, `argument_mod::arg_type` in LFRic core would have to be updated to contain an `nlevels`:
```fortran
 ...
     !> Optional metadata (fields only) for a type of stencil map to use.
     !! One of {XORY1D, X1D, Y1D, CROSS, REGION, CROSS2D}.
     integer :: stencil_map = -1
     !> Optional metadata (fields only) for inter-grid mapping kernels.
     !! One of {GH_FINE, GH_COARSE}.
     integer :: mesh_arg    = -1
     !> Optional metadata specifying the number of vertical levels for a field.
     character(len=SOME_LEN) :: nlevels=""
  end type arg_type
```
where `SOME_LEN` is some parameter with a suitable value. I'm not sure why this approach wasn't taken with other meta-data components so it could be there's some significant drawback to it?

doc/user_guide/lfric.rst
arporter: I now have:
image
If people are OK with that, the next question is whether to require `nlayers` to be an integer (in which case the 'labels' will have to be at least declared and initialised locally in the module containing the kernel) or whether we go for a string of some fixed length. That will avoid the need to declare anything but will require a character member of `arg_type`. As I've said elsewhere, I'm a bit uneasy about this because we've not done it before and there may be a good reason for that (and then again, it may just be because we've always wanted to restrict it to certain recognised values and integer parameters were an efficient way of enforcing that using Fortran alone).

arporter:
Coming back fresh to this issue. I realise that, since the field data is *always* held in a 1D array, if there are multiple data values associated with each dof then that too will affect the dof map. (I was confused about this before.) Therefore, if two fields are on the same space and have the same number of vertical levels but have different `ndata` then we would need to pass a dofmap for each.

doc/user_guide/lfric.rst
arporter: Hi @tommbendall , I'm just starting to pick this back up. Has there been any progress on this point - i.e. is it something that we need to support? I might as welll get it into the design now if so...

arporter:
The documentation now looks like:
 
image

and the rules for providing dofmaps:

image

how does that look @stevemullerworth and @tommbendall ?

arporter:
I've just realised a problem with that suggestion: Fortran isn't going to let us have NDATA/NLEVELS be both an integer and a character string. We could work around this by always having a character string, even if it just contains an integer?

arporter:
With everything as character strings:
image

doc/user_guide/lfric.rst
tommbendall: Yes we would want to support both types (different data layouts will have optimisation advantages for different types of science kernel), but it may actually make no difference to the PSyclone interface!

I think you say this in another comment but fields are always 1D arrays. The only difference it makes is whether the science kernel should be written:
```
do n = 1, ndata
  do k = 1, nlayers
    idx = map(df) + k-1 + nlayers*(n-1)
    ... ! do calculation
```
or
```
do k = 1, nlayers
  do n = 1, ndata
    idx = map(df) + n-1 + ndata*(k-1)
    ... ! do calculation
```
which I assume PSyclone doesn't need to know about

tommbendall:
> The documentation now looks like:
> 
> image
> and the rules for providing dofmaps:
> 
> image
> how does that look @stevemullerworth and @tommbendall ?

Hi Andy, I think this makes sense. The use of strings would probably create the simplest interface (although strings always feel a bit icky to me!), as creating generic enumerators would require us to add extra metadata arguments.

doc/user_guide/lfric.rst
arporter: Done, thanks Steve.

arporter:
I want to extend the PSyclone fork of lfric_core to support the new metadata but am having trouble compiling it (despite using the same commit as the ITs).
EDIT: it was a git configuration thing for the Jules repo. Solved now.

arporter:
Unfortunately, this is the first time we've added keyword arguments to the `arg_type` metadata element. All of the current meta-data handling code asssumes we'll get an `fparser2.Part_Ref` but these keywords mean we get an `fparser2.Structure_Constructor`.

arporter:
Happily, anything that is a Part_Ref can also be a Structure_Constructor so I think I can just replace the former with the latter.

There is a further complication in that a lot of the (new) metadata parsing code works by requiring that the metadata elements be at particular positions in the list of elements. This is unnecessary if we have named elements so I may need to distinguish between positional and optional elements. Validating the latter is also difficult as we need to catch e.g. misspelt keyword options.

arporter:
I was wrong when I said this was the first time we've added keyword arguments - field arguments to inter-grid kernels already accept the `mesh_arg` metadata entry. However, this is handled by `InterGridArgMetadata` which is distinct from `FieldArgMetadata`. The latter will still sometimes be a Part_Ref rather than a Structure_Constructor - I need to allow for both.

arporter:
I'd like to update one or more of the test kernels with an `ndata="1"` in order to exercise the new functionality more thoroughly. To do that I need to update `lfric_core` and I'm unsure what the procedure for that is @LonelyCat124 @sergisiso ? Do I create a branch?

Anyway, in the meantime this is ready for a first look. One for anyone really. It only updates the kernel metadata handling - nothing is done with the new metadata values yet.

arporter:
(The doc failure was due to fortranwiki not responding to the link checker.)

LonelyCat124:
> I'd like to update one or more of the test kernels with an `ndata="1"` in order to exercise the new functionality more thoroughly. To do that I need to update `lfric_core` and I'm unsure what the procedure for that is @LonelyCat124 @sergisiso ? Do I create a branch?
> 
> Anyway, in the meantime this is ready for a first look. One for anyone really. It only updates the kernel metadata handling - nothing is done with the new metadata values yet.

Yeah create a branch with the fix in lfric_core and then just link to this from that PR and to that PR from this one.

sergisiso:
@arporter Are these changes to eventually contribute to lfric_core/apps or only for our testing? Can they be in a kernel in lfric_apps instead of core? That makes it easier to handle

sergisiso:
Oh I see its https://github.com/stfc/lfric_core/pull/5/changes , this is the right place. I will consolidate these into a "psyclone_next" branch but after the current release is out.

doc/user_guide/lfric.rst
```diff
@@ -162,24 +162,34 @@ least rank (number of dimensions) one. Scalar arrays are identified with
 Field
 +++++
 
-LFRic API fields, identified with ``GH_FIELD`` metadata, represent
-FEM discretisations of various dynamical core prognostic and diagnostic
+LFRic API fields, identified with ``GH_FIELD`` metadata, represent FEM
```
sergisiso: Not necessarily for this PR but the lfric documentation has been refreshed and is now public: https://metoffice.github.io/lfric_core/how_to_use_it/lfric_datamodel/field.html
https://metoffice.github.io/lfric_core/how_it_works/core_data_model/science_model_architecture.html#dofs-dof-maps-and-function-spaces

We should at least link to it, and when appropriate don't repeat it.

doc/user_guide/lfric.rst
```diff
@@ -162,24 +162,34 @@ least rank (number of dimensions) one. Scalar arrays are identified with
...
+discretisations of various dynamical core prognostic and diagnostic
 variables. In FEM, variables are discretised by placing them into a
 function space (see :ref:`lfric-function-space`) from which they
 inherit a polynomial expansion via the basis functions of that space.
 Field values at points within a cell are evaluated as the sum of a set
-of basis functions multiplied by coefficients which are the data points.
-Points of evaluation are determined by a quadrature object
+of basis functions multiplied by coefficients which are the data
+points.  Points of evaluation are determined by a quadrature object
```
sergisiso: Not sure if it matters for rst but there is a double space here

doc/user_guide/lfric.rst
```diff
@@ -162,24 +162,34 @@ least rank (number of dimensions) one. Scalar arrays are identified with
...
+
 LFRic fields passed as arguments to any :ref:`LFRic kernel
 <lfric-kernel-valid-data-type>` can be of ``real`` or ``integer``
 primitive type. In the LFRic infrastructure, these fields are
 represented by instances of the ``field_type`` and ``integer_field_type``
 classes, respectively.
 
+Different fields may be defined on different numbers of vertical layers.
+The number of layers can be as few as one (a 2D field). Additionally,
+LFRic has the concet of multi-data fields where multiple data values can be
```
sergisiso: "concept" ?

doc/user_guide/lfric.rst
```diff
@@ -162,24 +162,34 @@ least rank (number of dimensions) one. Scalar arrays are identified with
...
+associated with each DoF. Unfortunately, both the number of layers and the
```
sergisiso: I would remove the "Unfortunately," or replace it with a "Since" and also remove the "Therefore"

doc/user_guide/lfric.rst
```diff
@@ -919,8 +929,8 @@ All three CMA-related kernel types must obey the following rules:
 1) Since a CMA operator only acts within a single column of data,
    stencil operations are not permitted.
 
-2) No vector quantities (e.g. ``GH_FIELD*3`` - see below) are
-   permitted as arguments.
+2) No vector quantities (e.g. ``GH_FIELD*3`` - see below) or
```
sergisiso: Maybe replace "see below" with a reference, I don't know where below is

sergisiso: or just delete it

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
+
+  arg_type(GH_FIELD, GH_REAL, GH_READ, W3, NLAYERS="1")
+
+Alternatively, it may be given a name, e.g.::
+
+  arg_type(GH_FIELD, GH_REAL, GH_READ, W3, NLAYERS="some name")
+
+If two or more field/operator arguments are on the same function space
+and have the same number of layers (whether a literal or a name) then
+only one dofmap (that of the first such field listed in the metadata)
```
sergisiso: Is "(that of the first such field listed in the metadata)" necessary? I think we are saying the the dofmap is the same, but how the psy-layer gets it is not that relevant.

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
+is passed to the kernel for those arguments.
+
+(Since the value of ``NLAYERS`` is looked-up from the corresponding kernel
+argument at run time, the labels given in the kernel metadata are just that
+- they do not have to correspond to anything in the LFRic infrastructure.)
```
sergisiso: I am bit lost with this comment, what do you mean by "are just that ..."?

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
+
+Multi-Data Metadata
+"""""""""""""""""""
+
+A multi-data field is the same as a standard field apart from having multiple
+values associated with each DoF. This is indicated in the field metadata by
```
sergisiso: Since this is a question I have every time I read this? Could we add a sentence about that is the distinction between this and field vectors?

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
+the optional ``NDATA`` argument to GH_FIELD, e.g.::
+
+  arg_type(GH_FIELD, GH_REAL, GH_READ, W2, NDATA="4")
+
+The value contained in the string specified for ``NDATA`` may be a literal if
+it is known at compile time. Alternatively, it may be a name in which
+case the number of data values at each DoF is determined at runtime by
```
sergisiso, line 1820: Is "at each dof" necessary? Can they be different for each dof?

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
+querying the field object (in the generated PSy layer). As with ``NLAYERS``,
+this name is just a label and does not have to correspond to anything in the
+LFRic infrastructure.
```
sergisiso, line 1823: Again, I am not sure I understand "As with ``NLAYERS``, this name is just a label and does not have to correspond to anything in the LFRic infrastructure."

Doesn't the LFRic field need to be declared with the same NDATA? And the dofmap correspond to the storage pattern?

doc/user_guide/lfric.rst
```diff
@@ -2106,21 +2166,26 @@ conventions, are:
...
 
-4) For each function space in the order they appear in the metadata arguments
-   (the ``to`` function space of an operator is considered to be before the
-   ``from`` function space of the same operator as it appears first in
-   lexicographic order)
+4) DoF maps for function spaces are handled in the order they appear in the
+   metadata arguments (the ``to`` function space of an operator is considered
+   to be before the ``from`` function space of the same operator as it appears
+   first in lexicographic order). Note that if two fields on a given function
+   space have differing numbers of vertical layers and/or ``NDATA`` values,
```
sergisiso: maybe s/vertical layers/\`\`NLAYERS\`\`/ to make it consistent with \`\`NDATA\`\` ?

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
 
 
+Number of Layers Metadata
+"""""""""""""""""""""""""
+
+If a particular field/operator kernel argument has a number of vertical
+levels that is *not* the same as the first field/operator argument then
+this must be specified using the ``NLAYERS`` option to ``GH_FIELD``/
+``GH_OPERATOR``. The value specified for ``NLAYERS`` may be an integer
+literal encoded as a string if it is known at compile time, e.g.::
```
sergisiso: Does https://psyclone.readthedocs.io/en/latest/_autogenerated/autoapi/psyclone.transformations.html#psyclone.transformations.LFRicKernelConstTrans take advantage of this?

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1757,52 @@ meshes cannot be on the same function space while those on the same
...
+- they do not have to correspond to anything in the LFRic infrastructure.)
+
+Multi-Data Metadata
+"""""""""""""""""""
+
+A multi-data field is the same as a standard field apart from having multiple
+values associated with each DoF. This is indicated in the field metadata by
+the optional ``NDATA`` argument to GH_FIELD, e.g.::
+
+  arg_type(GH_FIELD, GH_REAL, GH_READ, W2, NDATA="4")
```
sergisiso, line 1816: Not for this PR, but could https://psyclone.readthedocs.io/en/latest/_autogenerated/autoapi/psyclone.transformations.html#psyclone.transformations.LFRicKernelConstTrans
take advantage of this?

src/psyclone/tests/test_files/lfric/testkern_mod.F90
```diff
@@ -42,12 +42,12 @@ module testkern_mod
...
-             arg_type(gh_field,  gh_real, gh_inc,  w1), &
-             arg_type(gh_field,  gh_real, gh_read, w2), &
-             arg_type(gh_field,  gh_real, gh_read, w2), &
-             arg_type(gh_field,  gh_real, gh_read, w3)  &
+     type(arg_type), dimension(5) :: meta_args =                  &
+          (/ arg_type(gh_scalar, gh_real, gh_read),               &
+             arg_type(gh_field,  gh_real, gh_inc,  w1),           &
+             arg_type(gh_field,  gh_real, gh_read, w2),           &
+             arg_type(gh_field,  gh_real, gh_read, w2),           &
+             arg_type(gh_field,  gh_real, gh_read, w3, ndata="1") &
```
sergisiso, line 50: I think I haven't seen any example where nlayers and ndata are not integers. It would be good to have for testing them but also because I am not sure I understand their practical use yet.

.github/workflows/lfric_test.yml
```diff
@@ -42,7 +42,7 @@ on:
   push
 
 env:
-  LFRIC_APPS_HASH: 299d4f786dd2d86a59a184c3e53384005a69eded
+  LFRIC_APPS_HASH: f7714f2726e24574e5b4c42359f6e4728f268314
```
sergisiso: This is the right thing to do to update the integration tests but it has failed because the modified `dependencies.yaml` lfric_core ref does not correspond to the commit in your stfc/lfric_core PR, which should be: `6b8c5ff7c1e1f3a77e042c427174f0489de47eec`

sergisiso (changes requested):
@arporter See comments mostly about the documentation and how to fix the integration test.

doc/user_guide/lfric.rst
arporter: Oh yes, agreed. I've added links and removed a bit of text here that I wasn't totally sure about.

arporter: I've created #3479 to do this properly.

doc/user_guide/lfric.rst
arporter: That's the bit of text I've removed.

doc/user_guide/lfric.rst
arporter: Oops.

doc/user_guide/lfric.rst
arporter: Yes, done.

doc/user_guide/lfric.rst
arporter: Fixed.

doc/user_guide/lfric.rst
arporter: Good point.

doc/user_guide/lfric.rst
arporter: I mean they are only names - they are not defined in the LFRic infrastructure or in PSyclone.

doc/user_guide/lfric.rst
arporter: Yes, done.

doc/user_guide/lfric.rst
arporter: Well, it's not necessary but it's nice to be explicit.

doc/user_guide/lfric.rst
arporter: No - the NDATA is just a name, not necessarily a value. As long as all the field arguments that have NDATA of that name have the same actual value of NDATA then it doesn't matter what it actually is. (Since we can query one of them in the psy layer to get the actual value and the dofmap.)

doc/user_guide/lfric.rst
arporter: Yes, done.

doc/user_guide/lfric.rst
arporter: Ah! I'd forgotten about that. I think the answer is that it *could* take advantage of it but it won't yet. I'll add this to #868 (I looked at eg13 which uses this transformation and it was broken so I've fixed it).

doc/user_guide/lfric.rst
arporter: Potentially. `ndata` is a new concept so we'd need to extend the transformation.

src/psyclone/tests/test_files/lfric/testkern_mod.F90
arporter: I deliberately only made this change because it won't affect any generated code, even once we implement support (since ndata=1 is the default). I'd prefer to hold off adding any new kernels (and associated algorithm) until we come to do the code-generation step if that's OK?

.github/workflows/lfric_test.yml
arporter: I think I've got this now: so this points to a commit of `lfric_apps` that itself has its `dependencies.yaml` set to point to the correct commit of `lfric_core`.

arporter:
LFRic ITs still failed but now because a kernel definition is not consistent with its use. It looks like my lfric_apps and lfric_core branches are inconsistent and out of date. I really don't understand how I should be doing this!

arporter:
Ready for another look/discussion with @sergisiso to help me sort out my lfric_apps/core woes.

doc/user_guide/lfric.rst
sergisiso: Unfortunately this doesn't clarify it for me, I understand that:
- No NLAYER means 'looked-up from the corresponding kernel argument at run time' and provided as an argument
- A integer string, uses that value. No argument is provided? Or the argument is still there but set to the literal value?

But I don't understand what happens if it is "some name" as in the example above, specially because you say that "they do not have to correspond to anything in the LFRic infrastructure". So what value do they take then?

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1755,54 @@ meshes cannot be on the same function space while those on the same
...
+argument at run time, the labels given in the kernel metadata are just that
+- they do not have to correspond to anything in the LFRic infrastructure.)
+
+Multi-Data Metadata
+"""""""""""""""""""
+
+A multi-data field is the same as a standard field apart from having multiple
+values associated with each DoF. (Note that this is not the same as a
+:ref:`field vector <lfric-field-vector>` which corresponds to a bundle of
+fields, all on the same function space and each with a single value at each
```
sergisiso: Are multi-data fields also "all on the same function space"? If they are I would not mention it here to not imply that this is a disctinction factor

doc/user_guide/lfric.rst
```diff
@@ -1748,6 +1755,54 @@ meshes cannot be on the same function space while those on the same
...
+DoF.) This is indicated in the field metadata by
+the optional ``NDATA`` argument to GH_FIELD, e.g.::
+
+  arg_type(GH_FIELD, GH_REAL, GH_READ, W2, NDATA="4")
+
+The value contained in the string specified for ``NDATA`` may be a literal if
```
sergisiso, line 1818: "an integer literal"

doc/user_guide/lfric.rst
sergisiso: > Since we can query one of them

I understand that they don't need to be LFRic infrastructure parameters as other metadata fields (which I think is what this is trying to say).

But I still don't understand how you map from the given string to what you "query". Or do you always query the first field of a specific value, so this is basically a means to reduce the number of ndata arguments to a single one for each distinct ndata string (which string to use is completely arbitrary)?

src/psyclone/tests/test_files/lfric/testkern_mod.F90
sergisiso: > until we come to do the code-generation step if that's OK?

Does it generate an error saying that this is not yet supported? Could we do it until this error and let the "output" part of the test be done in the follow up PR?

sergisiso (changes requested):
@arporter I need some more clarifications. I will wait a bit to resolve the IT issue because it will become much simpler when the LFRic release is done and the intermediate branches are merged in the respective upstream projects.

doc/user_guide/lfric.rst
arporter: I've re-worded this now.

doc/user_guide/lfric.rst
arporter: This has been sorted out now (`nlayers` will be a character string as that then supports both integers and labels).

doc/user_guide/lfric.rst
arporter: OK, I've had a go at re-wording. Does this make it clearer:
image
?

doc/user_guide/lfric.rst
arporter: > do you always query the first field of a specific value, so this is basically a means to reduce the number of ndata arguments to a single one for each distinct ndata string (which string to use is completely arbitrary)?

Yes, exactly this. Hopefully my updated text above now explains this?

src/psyclone/tests/test_files/lfric/testkern_mod.F90
arporter: Hmm, now you ask that question I realise the answer is "no" - it will just silently ignore the new metadata which obviously isn't a good place to be. I will add something to catch this...(and therefore a kernel to exercise it).

doc/user_guide/lfric.rst
arporter: Yep, PSyclone doesn't need to know :-)

doc/user_guide/lfric.rst
arporter: Done (amongst other changes).

doc/user_guide/lfric.rst
arporter: Yes, simplified.

src/psyclone/tests/test_files/lfric/testkern_mod.F90
arporter: I've now fixed this.

arporter:
Thanks for those comments - they made me realise I'd missed some things. This should be ready for another look now @sergisiso.

doc/user_guide/lfric.rst
```diff
@@ -2098,7 +2160,12 @@ conventions, are:
       4) If the field entry stencil access is of type ``XORY1D`` then
          add an additional ``integer`` direction argument of kind
          ``i_def`` and with intent ``in``.
-
+      5) If the field is multi-data then the kernel must be passed the
+	 value of ``NDATA``: add an additional ``integer``, scalar
+	 argument of kind ``i_def`` and intent ``in``.
```
sergisiso: 5) and 6) have different indentations that the points before.

Also note that 5) uses a different uses a pasive "the kernel must be passed" while all others have imperative "add/pass". I would make this the same and also s/:/as/ to match the other formats.

examples/lfric/eg13/kernel_constants.py
```diff
@@ -73,23 +70,21 @@
...
+def trans(psyir: FileContainer):
     '''PSyclone transformation script for the LFRic API to make the
     kernel values of ndofs, nlayers and nquadrature-point sizes constant.
 
     :param psyir: the PSyIR of the PSy-layer.
-    :type psyir: :py:class:`psyclone.psyir.nodes.FileContainer`
 
     '''
+    mod_inline_trans = KernelModuleInlineTrans()
     const_trans = LFRicKernelConstTrans()
```
sergisiso, line 81: Maybe its out of scope to modify this transformation (feel free to defer these) but:
- The top comment on that transformation has a very strange link to a korean website? can we replace it or delete it?
- This examples creates 3 variables per nlayers:
```
    integer, parameter :: nlayers = 20
    integer, parameter :: nlayers_dummy = 20
    integer(kind=i_def), intent(in) :: nlayers_dummy_dummy
```
The first is the parameter used and the third the argument ignored, but what is the second one? Is it a mistake or is it expected?

- The element_order and quadrature do not change any execution statements, is this expected? But they create a bunch of {variable}_1 declarations that are unused, is this expected?

src/psyclone/domain/lfric/kernel/field_arg_metadata.py
```diff
@@ -49,77 +51,96 @@ class FieldArgMetadata(ScalarArgMetadata):
...
     access_arg_index = 2
     function_space_arg_index = 3
     stencil_arg_index = 4
     # The name to use for any exceptions.
     check_name = "field"
-    # The number of arguments in the language-level metadata (min and
-    # max values).
+    # The number of positional arguments in the language-level metadata (min
+    # and max values).
     nargs = (4, 5)
```
sergisiso: There is 4 more instances of this comment in sibling classes, do we need to update those or the change is specific only to this class?

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
```diff
@@ -0,0 +1,90 @@
...
+
+  type, extends(kernel_type) :: testkern_nlevels_ndata_type
+     type(arg_type), dimension(5) :: meta_args =                           &
+          (/ arg_type(gh_scalar, gh_real, gh_read),                        &
+             arg_type(gh_field,  gh_real, gh_inc,  w1),                    &
+             arg_type(gh_field,  gh_real, gh_read, w2),                    &
+             ! Non-default number of levels.
+             arg_type(gh_field,  gh_real, gh_read, w2, nlevels="shallow"), &
+             ! Non-default number of data values per dof.
+             arg_type(gh_field,  gh_real, gh_read, w3, ndata="precip")     &
```
sergisiso: Could this be from  W2?

Otherwise this is the first W3, so if I understand correctly it has to send dofmaps regardless of having ndata or not.

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
```diff
@@ -0,0 +1,90 @@
...
+
+  implicit none
+
+  type, extends(kernel_type) :: testkern_nlevels_ndata_type
+     type(arg_type), dimension(5) :: meta_args =                           &
+          (/ arg_type(gh_scalar, gh_real, gh_read),                        &
+             arg_type(gh_field,  gh_real, gh_inc,  w1),                    &
+             arg_type(gh_field,  gh_real, gh_read, w2),                    &
+             ! Non-default number of levels.
+             arg_type(gh_field,  gh_real, gh_read, w2, nlevels="shallow"), &
```
sergisiso, line 51: I understand that if we had a second:
`arg_type(gh_field,  gh_real, gh_read, w2, nlevels="shallow")`
it would not change the argument list? Can we add it?

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
```diff
@@ -0,0 +1,90 @@
...
+
+contains
+
+  subroutine testkern_nlevels_ndata_code(nlayers, ascalar,          &
+                                         fld1, fld2, fld3, fld4,    &
+                                         ndf_w1, undf_w1, map_w1,   &
+                                         ndf_w2, undf_w2, map_w2,   &
+                                         nlayers_fld3, ndf_fld3,    &
+                                         undf_fld3, map_w2_fld3,    &
+                                         ndf_w3, undf_w3, map_w3)
```
sergisiso: Do we not provide a 'ndata' value? Does it mean labels like "precip" will always be a static known value and never depend on a namelist or anything like that?

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
```diff
@@ -0,0 +1,90 @@
...
+  type, extends(kernel_type) :: testkern_nlevels_ndata_type
+     type(arg_type), dimension(5) :: meta_args =                           &
+          (/ arg_type(gh_scalar, gh_real, gh_read),                        &
+             arg_type(gh_field,  gh_real, gh_inc,  w1),                    &
+             arg_type(gh_field,  gh_real, gh_read, w2),                    &
+             ! Non-default number of levels.
+             arg_type(gh_field,  gh_real, gh_read, w2, nlevels="shallow"), &
+             ! Non-default number of data values per dof.
+             arg_type(gh_field,  gh_real, gh_read, w3, ndata="precip")     &
+           /)
```
sergisiso, line 61: What it we then did a:
`arg_type(gh_field,  gh_real, gh_read, w2, nlevels="shallow", ndata="precip")`
 This should be yet another dofmap different from the already provided?

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
```diff
@@ -0,0 +1,90 @@
...
+     procedure, nopass :: code => testkern_nlevels_ndata_code
+  end type testkern_nlevels_ndata_type
+
+contains
+
+  subroutine testkern_nlevels_ndata_code(nlayers, ascalar,          &
+                                         fld1, fld2, fld3, fld4,    &
+                                         ndf_w1, undf_w1, map_w1,   &
+                                         ndf_w2, undf_w2, map_w2,   &
+                                         nlayers_fld3, ndf_fld3,    &
```
sergisiso: Maybe the names should be [u]ndf_fld3 should be [u]ndf_w2_fld3? To match the map name like in the other cases?

Or maybe make it more readable by including the label, e.g. "[u]ndf_w2_shallow", "map_w2_shallow". But then we have to care that the label characters are valid symbol characters.

sergisiso (changes requested):
@arporter I am now happy with the documentation other than a small formatting aspect and I am more confident that it works thanks to the examples. There are just a few comments left. I will update the integration tests next time so I can wait for the updates I made in the other PR.

doc/user_guide/lfric.rst
arporter: Done.

examples/lfric/eg13/kernel_constants.py
arporter: That is a very strange link! I'm sure it used to resolve to somewhere meaningful. I've removed the sentence.
I'll investigate the code generated.

arporter: Probably the quadrature and element_order are expected. The nlayers change looks like the transformation has been applied twice to the same kernel.  In fact, that is what happens:
```
psyclone -l output --config /home/kbc59144/Projects/PSyclone/config/psyclone.cfg -api lfric -s ./kernel_constants.py \
  ../code/gw_mixed_schur_preconditioner_alg_mod.x90 \
  -oalg alg.f90 -opsy psy.f90
  kernel 'matrix_vector_code'
    Modified nlayers, arg position 2, value 20.
    Skipped dofs, arg position 7, function space any_space_1
    Skipped dofs, arg position 10, function space any_space_2
  kernel 'matrix_vector_code'
    Modified nlayers_dummy, arg position 2, value 20.
    Skipped dofs, arg position 7, function space any_space_1
    Skipped dofs, arg position 10, function space any_space_2
  kernel 'dg_matrix_vector_code'
    Modified nlayers, arg position 2, value 20.
    Skipped dofs, arg position 7, function space any_discontinuous_space_1
    Skipped dofs, arg position 10, function space any_space_1
  kernel 'dg_matrix_vector_code'
    Modified nlayers_dummy, arg position 2, value 20.
    Skipped dofs, arg position 7, function space any_discontinuous_space_1
    Skipped dofs, arg position 10, function space any_space_1
```

arporter: I've updated the transformation script to avoid this.

arporter: I've also opened (:-() #3498 to update the transformation.

src/psyclone/domain/lfric/kernel/field_arg_metadata.py
arporter: Good question. I guess it should apply to field vectors and possibly operators too - I've asked in the ticket.

arporter: Happily, it looks like the field vector will just inherit from field so that should be OK. I'll need to test this though.

arporter: Not sure what I was thinking here - it is the *total* number of arguments, not the positional ones. I've updated the max limit and updated a test to exercise this.

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
arporter: Good idea - you have understood correctly. I'd also got the argument ordering wrong in the kernel subroutine so have fixed that.

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
arporter: Yes, and yes.

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
arporter: No, that was my mistake. Fixed now.

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
arporter: Yes, and I've added this example too. In fact it would mean we don't have to pass ndata and nlevels for this argument so I'll update the rules too.

src/psyclone/tests/test_files/lfric/testkern_nlevels_ndata_mod.F90
arporter: Normally we won't generate these names - it's up to the kernel writer. The exception will be in the kernel-stub generator if we go down the road of implementing support there.

doc/user_guide/lfric.rst
sergisiso: The indentation is still not correct (it uses tabs while other lines use spaces)

sergisiso:
@arporter As discussed, I don't understand my a metadata change (that is not yet used in lfric_app kernels) has changed the performance of a couple of integration tests. So I will take a bit longer to review this as I want to run them manually and try to understand what is changing.

sergisiso (approved):
All good now @arporter, the test issue was that the lfric_core change was deleting some of the recently introduced omp_uniform flags. All runs as expected now.

Issue #1623, opened 2022-02-13
Add training material to PSyclone


hiker:
A more detailed intro to PSyclone, starting with gocean, nemo, LFRic.

PR #3212, opened 2025-11-06
#1623 GOcean training fusing through inlining

sergisiso:
@hiker This is a PR against your training branch, not master. It shows how to inline and then fuse (I only modified the solution files). And it also works with my component_indices refactor.

To inline I had to give a bit more info to the r2d_field inferred_datatype. I think this is a better alternative that what you have in your full_inlining example. With this inlining can be done with:
```
  modinline.apply(kern)
  call = kern.lower_to_language_level()
  inline.apply(call)
```

- I also removed some duplicated KernelInlining (I think)
- I also removed the EnterData, I am not sure we want to show this, it is not faster, and more fragile than the memory_managed as currently implemented.

Let me know what you think.

### DIFF
--- a/src/psyclone/tests/psyir/transformations/omp_task_transformations_test.py
+++ b/src/psyclone/tests/psyir/transformations/omp_task_transformations_test.py
@@ -194,7 +194,7 @@ def test_omptask_inline_kernels(monkeypatch):
     taskt = OMPTaskTrans()
     schedule = invoke.schedule
 
-    # If we monkeypatch the validate() and aget_callee() it should succeed.
+    # If we monkeypatch the validate() and get_callee() it should succeed.
     monkeypatch.setattr(InlineTrans, "validate",
                         lambda _1, _2, routine,
                         use_first_callee_and_no_arg_check,


sergisiso:
> I wonder if this could be moved into the inline transformation? I.e. if it's a kernel, lower first?

Good point, I would agree with this, but that's a separate PR.

> "I also removed the EnterData"
> Not sure if we want to prefer a certain method.

I agree, I meant to say: I "had to" remove the EnterData directive because it is incompatible with inlining. The EnterData looks for KernCalls to decide the clause but inlining removes them. Chaning the order of the transformation doesn't help because EnterData does the search during lowering. This is a regression of this proposal that I don't know how to fix until we can fuse kerncall agian. This is why I was justifing that managed memory is currently the better option. But I agree we should ideally offer choice.

> is inlining required for loop fusion [...] they seem to work now without inlining first

I thought it was required, but maybe adding more type info in the infer_datatype improved something else? Also I only tested one and the copy the solution everywhere else, so maybe different examples have simpler requirements?

> IMHO, we should not require lowering in order to do fusion

Totally agree with this, as you mention in the last sentence, this PR is just to unblock the component_indices change without breaking your traning. So we can both continue merging the respectives PRs and we can work on the long term solution while having them on master.

sergisiso:
So if you are happy with this I will continue with the component_indices and leave this PR open for you to merge or copy the solution at your own convinience. Does that sound ok?

sergisiso:
Closing as this was merged with #3218

PR #3213, opened 2025-11-07
Array index analysis using SMT

mn416:
This PR introduces an `ArrayIndexAnalysis` class to determine whether or not distinct iterations of a given loop can generate conflicting array accesses (if not, the loop can potentially be parallelised). It formulates the problem as a set of SMT constraints over array indices which are then are passed to the third-party [Z3 solver](https://github.com/Z3Prover/z3).

The PR also introduces a new option `use_smt_array_index_analysis` to `DependencyTools`, which enables use of the new analysis in place of `DependencyTools`'s current array conflict analysis. Transformations such as `ParallelLoopTrans` and `OMPLoopTrans` are parameterised by `DependencyTools`, allowing use of the new analysis in existing transformations.

To install the Z3 solver:

```
$ pip install z3-solver
```

This dependency has been added to `setup.py`.

The analysis is broadly explained in [source code comments](https://github.com/stfc/PSyclone/blob/mn416-smt-array-index-analysis/src/psyclone/psyir/tools/array_index_analysis.py).

There is also a set of examples showing what the analysis is capable of, available [here](https://github.com/mn416/array-analysis-examples). None of these examples are solved by PSyclone's existing loop analysis, and all can be solved by the SMT-based analysis.

The comments below include results from applying the analysis to the full UKCA codebase (execution time of the analysis, and number of loops parallelised).


arporter:
This feels like one for @hiker to look at as he did DependencyTools - without looking at the code my feeling is that it would make sense for extra sophistication to go in there rather than into a specific PSyIR node.

EDIT: my mistake - I see you say you've added a new class to encapsulate this functionality. I'd still like to see whether it makes sense to merge it/include it with DependencyTools though.

src/psyclone/psyir/nodes/if_block.py
```diff
@@ -195,3 +196,22 @@ def reference_accesses(self) -> VariablesAccessMap:
...
+
+    def flat(self):
+        '''This method allows a chain of 'if'/'else if'/.../'else'
+        statements to be viewed in its flattened form, without nesting.
+
+        :returns: a list of condition/body pairs. Nested 'else if' chains
+           (if there are any) are recursively gathered. The condition for
+           the final 'else' in the chain (if there is one) is 'None'.
+        '''
+        if self.else_body is None:
```
hiker: I find this harder to understand than necessary (probably because adding 'this' block's data at the end with an `insert`). As far as I can tell, the following code produces the same result:
```
        branches = [(self.condition, self.if_body)]
        if self.else_body:
            if (isinstance(self.else_body, Schedule) and
                    len(self.else_body.children) == 1 and
                    isinstance(self.else_body.children[0], IfBlock)):
                branches.extend(self.else_body.children[0].flat())
            else:
                branches.append((None, self.else_body))
        return branches
```
(if not, we need a test for that, the code passes the existing tests).

src/psyclone/tests/psyir/nodes/if_block_test.py
```diff
@@ -282,3 +283,72 @@ def test_ifblock_children_validation():
         ifblock.addchild(else_body)
     assert ("Item 'Schedule' can't be child 3 of 'If'. The valid format is: "
             "'DataNode, Schedule [, Schedule]'." in str(excinfo.value))
+
```
hiker, line 286: Can you also add a test with no `else` branch? ATM that line in `if_block.py` is not covered. My suggested re-writing avoids this issue (since there is no code to execute if there is no else block :) ), but still we should have a separate test that covers the no-else case.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -162,11 +164,20 @@ class DependencyTools():
         specified in the PSyclone config file. This can be used to
         exclude for example 1-dimensional loops.
     :type loop_types_to_parallelise: Optional[List[str]]
+    :param use_smt_array_index_analysis: if True, the SMT-based
+        array index analysis will be used for detecting array access
+        conflicts. An ArrayIndexAnalysisOptions value can also be given,
+        instead of a bool, in which case the analysis will be invoked
+        with the given options.
+    :type use_smt_array_index_analysis: Union[
```
hiker: We have switched to use 'proper' Python typing (instead of `:type:`), and are fixing files as we touch them. Could you update this to use:
```
                 loop_types_to_parallelise: Bool = None,
                 use_smt_array_index_analysis: Union[...] = False):
```
and remove the existing `:type:` comments? Thanks

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -884,9 +896,15 @@ def can_loop_be_parallelised(self, loop,
             # TODO #1270 - the is_array_access function might be moved
             is_array = symbol.is_array_access(access_info=var_info)
             if is_array:
-                # Handle arrays
-                par_able = self._array_access_parallelisable(loop_vars,
-                                                             var_info)
+                # If using the SMT-based array index analysis then do
+                # nothing for now. This analysis is run after the loop.
+                if self._use_smt_array_index_analysis:
```
hiker, line 918: Would it make sense to first run the existing array analysis? And only run the the Z3 solver if the result is false (and long term, maybe the existing analysis could return three values: yes, no, unknown.

I might try to do some additional performance tests (I noticed the results you have posted ;) ), this obviously only makes sense if the existing one is significantly faster.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -162,11 +164,20 @@ class DependencyTools():
         specified in the PSyclone config file. This can be used to
         exclude for example 1-dimensional loops.
     :type loop_types_to_parallelise: Optional[List[str]]
+    :param use_smt_array_index_analysis: if True, the SMT-based
+        array index analysis will be used for detecting array access
+        conflicts. An ArrayIndexAnalysisOptions value can also be given,
+        instead of a bool, in which case the analysis will be invoked
+        with the given options.
+    :type use_smt_array_index_analysis: Union[
+        bool, ArrayIndexAnalysisOptions]
```
hiker: I have to admit I don't like this kind of `Union` here. Yes, it actually works fine, but I would guess that a linter will have problems with that (and yes, atm we don't do linting for typing, but that might come at some stage :) ).

Could we instead just use `use_smt_array_index_analysis: Optional[...] = None`? Then add a `enabled` option to the `ArrayIndexAnalysisOptions` class. So, if no ArrayIndexAnalysisOptions is specified, you can set `self._smt_array_analysis_option = ArrayIndexAnalysisOptions()` and set the enabled option to false. This way, the dependency analysis member will have a single type only, removing the need for `isinstance` tests later.

I think my main issue is that it does not 'read' nice: the variable is `use_...`, but it's sometimes not a boolean.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -898,6 +916,23 @@ def can_loop_be_parallelised(self, loop,
...
+            if isinstance(self._use_smt_array_index_analysis,
+                          ArrayIndexAnalysisOptions):
+                options = self._use_smt_array_index_analysis
+            else:
+                options = ArrayIndexAnalysisOptions()
+            analysis = ArrayIndexAnalysis(options)
+            conflict_free = analysis.is_loop_conflict_free(loop)
+            if not conflict_free:
+                self._add_message(
+                    "The ArrayIndexAnalysis has determined that the"
```
hiker: Can we get additional information (esp. which variable is the problem)? Ideally, these messages should be added in the new class. One way of doing this would be to pass in the DependencyAnalysis instance to `is_loop_conflict_free` (so it could use the `add_message` method). Or we could even split the message handling into a stand-alone class.
I am happy for now to just open a ticket and add a TODO here.

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -185,6 +188,8 @@ def validate(self, node, options=None, **kwargs):
             sequential = options.get("sequential", False)
             privatise_arrays = options.get("privatise_arrays", False)
             reduction_ops = options.get("reduction_ops", [])
+            use_smt_array_index_analysis = options.get(
```
hiker: Hmm - I see now how the handling of the dependency tools interacts with the transformation.
Looking at the transformation, there is already a place in `apply` where we create dependency tools, and pass it to `independent_iterations`.
What about we use the same pattern? I.e. change this transformation to accept an optional DependencyTool instance (if not, it creates a default with default setting for your analysys).
This would future-proof the design for any future extensions.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -5349,3 +5349,41 @@ def test_firstprivate_with_uninitialised(fortran_reader, fortran_writer):
...
+            enddo
+          enddo
+        enddo
+      end subroutine my_matmul''')
+    omplooptrans = OMPLoopTrans(omp_directive="paralleldo")
+    loop = psyir.walk(Loop)[0]
+    omplooptrans.apply(
+        loop, collapse=True, use_smt_array_index_analysis=True)
+    output = fortran_writer(psyir)
+    assert "collapse(2)" in output
```
hiker, line 5387: Could you make it 'more' obvious that this implies that OpenMP has been applied? Either by adding a comment saying if omp was not applied, this  would raise an exception, or by checking for the longer string `!$omp parallel do collapse(2) `

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2025, University of Cambridge, UK.
```
hiker: ~I need to check with Andy, but I believe this should have the STFC header :(~

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+from psyclone.psyir.symbols import DataType, ScalarType, ArrayType, \
+    INTEGER_TYPE
+
+# Outline
+# =======
+#
+# The analysis class provides a method 'is_loop_conflict_free()' to
+# decide whether or not the array accesses in a given loop are
+# conflicting between iterations. Two array accesses are conflicting
+# if they access the same element of the same array, and at least one
```
hiker: ... the same array in different loop iterations, ...

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# of the accesses is a write.
+#
+# The analysis assumes that any scalar integer or scalar logical
+# variables written by the loop can safely be considered as private
+# to each iteration. This should be validated by the callee.
```
hiker: callee and is typically done in the DependencyTools.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+#
+# Given a loop, we find its enclosing routine, and start analysing the
+# routine statement-by-statement in a recursive-descent fashion.
+#
+# As we proceed, we maintain a set of SMT constraints and a
+# substitution that maps Fortran variable names to SMT variable names.
+# For each Fortran variable, the substitution points to an SMT
+# variable that is constrained (in the set of constraints) such that
+# it captures the value of the Fortran variable at the current point
+# in the code. When a Fortran variable is mutated, the substitution is
```
hiker: 'mutated'? You mean updated?

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# modified to point to a fresh SMT variable, with new constraints,
+# without destroying the old constraints.
+#
+# More concretely, when we encounter an assignment of a scalar
+# integer/logical variable, of the form 'x = rhs', we translate 'rhs'
+# to the SMT formula 'smt_rhs' with the current substitution applied.
+# We then add a constraint 'var = smt_rhs' where 'var' is a fresh SMT
+# variable, and update the substition so that 'x' maps to 'var'.
```
hiker: s/substition/substitution/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# updating the substitution to point to a fresh unconstrained SMT
+# variable; we refer to this process as "killing" the variable.
+#
+# In addition to the current substitution, we maintain a stack of
+# previous substitutions. This allows substitutions to be saved and
+# restored before and after analysing a block of code that may or may
+# not be executed at run time.
+#
+# We also maintain a "current condition".  This can be viewed as a
+# constraint that has not been comitted to the constraint set because
```
hiker: s/comitted/committed/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# we want to be able to grow, contract, and retract it as we enter and
+# exit conditional blocks of code. This current condition is passed in
+# recursive calls, so there is an implicit stack of them.
+#
+# More concretely, when we encouter an 'if' statement, we copy the
```
hiker: s/encouter/encounter/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# the array, whether it is a read or a write, the current condition at
+# the point the access is made, and its list of indices (translated to
+# SMT). When we are analysing code that is inside the loop of
+# interest, we add all array accesses encountered to the list.
+#
+# When we encounter the loop of interest, we perform a couple of steps
+# before recursing into the loop body. First, we kill all variables
+# written by the loop body, because we don't know whether we are
+# entering the loop (at run time) for the first time or not. Second,
+# we create two SMT variables to represent two arbitary but distinct
```
hiker: ... two SMT variables to represents the loop indices of two arbitrary but distinct iterations of the loop.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# iterations of the loop. Each variable is constrained to the start,
```
hiker: Each of these two variables is constrained to ...

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# stop, and step of the loop, and the two variables are constrained to
+# be not equal. After that, we analyse the loop body twice, each time
+# mapping the loop variable in the substitution to each of the SMT
+# loop variables.  After analysing the loop body for the first time,
```
hiker: Single space between `.` and `After` (that happened elsewhere here as well, maybe just search for `.  ` (double space)

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# we save the array access list and start afresh with a new one.
+# Therefore, once the analysis is complete, we have two array access
+# lists, one for each iteration.
+#
+# When we encounter a loop that is not the loop of interest, we follow
+# a similar approach but only consider a single arbitrary iteration of
+# the loop.
+#
+# When the recursive descent is complete, we are left with two array
+# access lists. We are interested in whether any pair of accesses to
```
hiker: What about replacing 'We are interested ...  conflict' with something like:

A conflict happens if there is an access in the first list of accesses that is the same as an access in the second list, and one of which is a write.

hiker: ... access lists, representing two different iterations of the same loop.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+       that a conflict exists even though it has not yet found one.
+       This can be set to 'None' to disable the timeout.
+
+    :param prohibit_overflow: if True, the analysis will tell the solver
+       to ignore the possibility of integer overflow. Integer overflow is
+       undefined behaviour in Fortran so this is safe.
+
+    :param handle_array_intrins: handle array intrinsics 'size()',
+       'lbound()', and 'ubound()' specially. For example, multiple
+       occurences of 'size(arr)' will be assumed to return the same value,
```
hiker: s/occurences/occurrences/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+
+        :param psyir_node: PSyIR node for the access (useful for reporting
+           conflict messages / errors).
+        '''
+        def __init__(self,
+                     cond:       z3.BoolRef,
+                     is_write:   bool,
+                     indices:    list[list[z3.ExprRef]],
+                     psyir_node: Node):
+            self.cond = cond
```
hiker: Could you use 'private' (I know, they are not really private) variable names, i.e. starting with `_` (`self._is_write = is_write`). And add a getter (setter) if required.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+        determine whether or not distinct iterations of a given loop
+        can generate conflicting array accesses.
+
+        :param options: these options allow user control over features
+           provided by, and choices made by, the analysis.
+        '''
+        self.opts = options
+
+    def _init_analysis(self):
+        '''Intialise the analysis by setting all the internal state
```
hiker: s/Intialise/Initialise/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+        varibles accordingly.'''
```
hiker: s/varibles/variables/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+        # the substitution to cover only the free variables present
+        # in the expression, and then apply it.
+        subst_pairs = []
+        for fv in _free_vars(expr):
+            if fv in self.subst:
+                subst_pairs.append((fv, self.subst[fv]))
+        return z3.substitute(expr, *subst_pairs)
+
+    def _translate_integer_expr_with_subst(self, expr: Node):
+        '''Translate the given integer expresison to SMT, and apply the
```
hiker: s/expresison/expression/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+# Translating Fortran expressions to SMT formulae
+# ===============================================
+
+
+def translate_integer_expr(expr_root: Node,
+                           opts: ArrayIndexAnalysisOptions
+                           ) -> (z3.ExprRef, z3.BoolRef):
+    '''Translate a scalar integer Fortran expression to SMT. In addition,
+    return a constraint that prohibits/ignores bit-vector overflow in the
+    expression.'''
```
hiker: Could you add a `:returns:... ` line

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+
+    constraints = []
+
+    def trans(expr: Node) -> z3.ExprRef:
+        # Check that type is a scalar integer of unspecified precision
```
hiker: Could you add `:param ...` and `:returns:` documentation.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,1099 @@
...
+        type_ok = _is_scalar_integer(expr.datatype)
```
hiker: I am not sure I understand the `type_ok` handling. It is only tested for the first two tests, after that is ignored, so a (bad-style) assignment (`int_var = 1.23`) would not set `int_var` to 1?

hiker (commented):
I don't have enough time today to finish the review, so I have to abort somehow in the middle.
Two 'general' questions:
1. Is there any advantage of supporting bit vectors? It feels like the transformation of expressions to Z3 can be cut in half if we would only support integers (and since we kind of agree to ignore overflows).  Or of course the reverse ;)
2. I feel it might be useful to have a command line option to generically enable your analysis and set the main configurations (timeout and #threads?). E.g. the Fab build system should be able to support to directory-specific flags, so the build system could enable your analysis for all files under `../ukca` or so, and not affecting anything else?

That might be useful if the same transmutation script is used across different source directories, but only a subset will benefit from Z3.

I will continue hopefully tomorrow.

src/psyclone/psyir/tools/array_index_analysis.py
hiker: Please ignore. Andy confirmed that this is fine (as long as the license is BSD 3).

hiker:
...
> Here is a histogram of the times taken by each method to complete on each loop:

Do you have the file names for the slow ones? I would be interested to look at them in more detail (for both approaches).

...
> As expected, `DependencyTools` is generally quicker. Its run time is proportional to the size of the loop whereas the run time of `ArrayIndexAnalysis` is proportional to the size of the routine containing the loop. Plus there is the cost of running a general-purpose solver.

I assume you are looking at the code 'around' a loop to know additional constraints on variables that might be used in the loop? Is this actually required (i. e. leads to more parallelisable loops)? I still need to fully understand you code ;)

Might an option to restrict the Z3 analysis to just the loops be useful?? 


> All loops taking longer than 15s to analyse by `ArrayIndexAnalysis` contain auto-generated code (unnatural loops containing hundreds/thousands of array access sites). 

Would that imply that you get hundreds of duplicated constraints (which could slow things down)? Maybe this could be filtered??


> I have not investigated the `DependencyTools` outliers.
> 
> The `ArrayIndexAnalysis` was run with an SMT-solver timeout of 5s, and none of the loops in UKCA timed out. This is an encouraging sign that the SMT solver is generally good at solving array-conflict problems.

Sounds promising.
> 
> Using `ArrayIndexAnalysis` led to 12 exceptions. These were all due to Fortran code not supported by PSyclone (referring to the same array multiple times on the LHS of an assignment).
> 
> Using `DependencyTools` led to 54 exceptions. I have not yet explored the cause of these.

Again, I would be interested in a list :)

If you don't have the data ready, don't worry, I should be able to reproduce them:
```
science/ukca/src$ find  -iname "*.f90"|wc
    308     308   15702
```
which seems to match the number of files you mentioned :)

hiker:
I just realised that I am running on a slightly different ukca version, but here outliers  (runtime > 5 seconds) I see: (from 256 files, after which psyclone appears to hang, which I will look at separately):
```
File name                   loop#                           DepTools    parallel?   z3        z3 result

ukca_um_strat_photol_mod.F90	2	Execution time loop i:	11.899324	True	1.511083	True
ukca_um_strat_photol_mod.F90	3	Execution time loop j:	6.690157	True	1.471593	True
ukca_chemistry_ctl_col_mod.F90	2	Execution time loop j:	0.032098	False	5.312659	False
ukca_deriv_aero.F90				1	Execution time loop n:	0.173906	False	16.631266	False
ukca_deriv_aero.F90				4	Execution time loop i:	0.331547	False	14.088031	False
ukca_deriv_raqaero_mod.F90		1	Execution time loop n:	0.379767	False	25.602987	False
ukca_deriv_raqaero_mod.F90		3	Execution time loop i:	0.126291	False	24.398631	False
ukca_surfddr.F90			   17	Execution time loop j:	0.196217	False	8.116066	False
ukca_aerod.F90					1	Execution time loop j:	38.493939	True	5.474127	True
ukca_deriv_raq.F90				1	Execution time loop n:	0.291267	False	20.539170	False
ukca_deriv_raq.F90				3	Execution time loop i:	0.110052	False	21.570237	False
ukca_deriv.F90					1	Execution time loop n:	0.268167	False	11.196480	False
ukca_deriv.F90					3	Execution time loop i:	0.109296	False	10.146189	False

```

hiker:
I will stop for now, and hand it back to you. I started to get into the actual analysis, but I thinks splitting this into two files (one to convert expressions (or Fortran statement), one to do the actual analysis) will be a good start.

Note that I also did #3311, which should fix some (all?) of the timing outliers with the dependency tools ;) Thanks for your performance results.

src/psyclone/psyir/nodes/if_block.py
mn416: Done (commit b9154cd).

src/psyclone/tests/psyir/nodes/if_block_test.py
mn416: Done (commit b9154cd).

src/psyclone/psyir/tools/dependency_tools.py
mn416: I had the same thought myself. I might return to this at some point, as I think a bit more work would be needed to justify it right now. One advantage of keeping them separate was that it allowed me to find some issues with the existing DA (which I think you have now fixed).

src/psyclone/psyir/tools/dependency_tools.py
mn416: Done (commit df713f2) but only for the the method in question, not throughout the file.

src/psyclone/psyir/tools/dependency_tools.py
mn416: Done (commit df713f2).

src/psyclone/psyir/tools/dependency_tools.py
mn416: The analysis now generates conflict messages (commit 80c3428). As an example, the loop

```
  n = size(arr)
  do i = 1, n-1
    do j = i+1, n
      arr(j) = arr(j) + arr(i)
    end do
  end do
```

yields the following message for the `i` loop:

```
Iterations 1 and 2 have conflicting accesses to arr(2)
```

These messages get converted to DA `Message` values inside DA. I could add a TODO to use the DA's `Message` class in future, but I think it's preferable from a modularity perspective to keep `ArrayIndexAnalysis` separate from DA, as it can be understood in isolation.

src/psyclone/psyir/transformations/parallel_loop_trans.py
mn416: Yes, that seems cleaner. Done in commit 08a1d93.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
mn416: Done (commit e7e3413).

src/psyclone/psyir/tools/array_index_analysis.py
mn416: This and all the suggested documentation tidy ups (below) have been done in commit b228cb0.

src/psyclone/psyir/tools/array_index_analysis.py
mn416: Oops, missed this one. Done in commit 80e831e.

src/psyclone/psyir/tools/array_index_analysis.py
mn416: Done (commit f69fa4f).

src/psyclone/psyir/tools/array_index_analysis.py
mn416: Yes, you're right. It's not consistent to do the checks only for the first two constructs. In any case, I don't think they are needed because the translation routine is only called when the LHS of an assignment is known to be a scalar (or, in the case of a `logical`, on an if/loop condition which must be scalar by definition). So, if the program is type correct, which seems reasonable to assume, then these `type_ok` checks are redundant.

Removed in commit 8fdde19.

I've added some comments clarifying the assumptions in the translation functions in commit 3d99d60.

mn416:
@hiker, thank you very much for all your comments and suggestions. I think I've addressed all of them, except for a few outstanding ones which I address below.

First, and most importantly, you had a concern about the use of a solver timeout. Paraphrasing:

> On a slow (or highly loaded) machine, the time out could be reached and a loop would not parallelised, but on a faster (not highly loaded) machine, the loop would be parallelised. That's ... really not something we should be doing.

This is a valid concern and, in fact, I experienced it when the PSyclone CI started running my tests. As a result, I modified my tests so that timeouts are not treated as failures. However, it does bring into question whether the analysis should be used in transmutation scripts where the user depends on a consistent result. I added the following warning to the developer guide:

> The main drawback of the SMT-based analysis is that the time-to-solve is not guaranteed to be consistent. It might be fast using one version of Z3 but slow using another. It might vary significantly with seemingly innocuous code changes. Having said that, our experience is that it generally performs well, especially when setting the option `num_sweep_threads=4` to explore different constraint orderings in parallel (this is the default setting). Nevertheless, it is probably best suited to static checking / linting, e.g. spotting bugs in parallel loops or missed opportunities for parallel loops. It is probably not advisable to use it in cases where a consistent result is relied upon, such as a transmutation script.

I will leave it to you to decide whether or not it is a good idea to go ahead with this PR. Obviously, please do what is right for the PSyclone project (I will not be dissappointed or offended). I think the analysis has some great potential uses in a static checker, but that may not align with the goals of the PSyclone project.

Elsewhere, you suggested splitting out the translation-to-Z3 code from the analysis. This has been done in commit dc0bba0. The code has also been brought up-to-date with master.

Things I can look at further if you're happy to go ahead with the PR:

  * I've added documentation to the developers guide (commit 1bb4705), but not done the `argpase` suggestion and associated changes to user guide.

  * Your suggestion to restrict the analysis to just the loops (rather than looking at the whole routine) probably would help with analysis time and might work well as a new option, but I've not tried it.

BTW, all CI checks were passing before the I made some documentation changes, but now code coverage is failing even though all new lines of have been covered.

doc/developer_guide/dependency.rst
```diff
@@ -585,6 +585,34 @@ can be parallelised:
...
+guaranteed to be consistent. It might be fast using one version of Z3 but slow
+using another. It might vary significantly with seemingly innocuous code
+changes. (Having said that, our experience is that it generally performs well,
+especially when setting the option `num_sweep_threads=4` to explore different
+constraint orderings in parallel). It is therefore best suited to static
+checking / linting, e.g. spotting bugs in parallel loops or missed
+opportunities for parallel loops. If the user depends on a timely result, such
+as in a build-time transmutation script, it should be used with the
+`succeed_on_timeout=True` option. This means the analysis may report no
+conflicts when there is one, but it will have tried to find one.
```
hiker, line 614: That needs more careful wording (we want to avoid that developers take this statement as: 'just disable it, all will be fine' ... I know, that's not what you are writing, still). What about:
> ... when there is one, and therefore it's up to the user to ensure validity of the transformation in this case. For example, the user could investigate the reasons for the rejection by the standard `DependencyTools` and verify that the reason for the rejection has not changed. Or a checksum of the instrumented region could be used to detect a change to the original source, which then could be flagged to the user to verify again.

src/psyclone/psyir/tools/__init__.py
```diff
@@ -43,11 +43,16 @@
...
 __all__ = ['CallTreeUtils',
            'DTCode',
            'DependencyTools',
            'DefinitionUseChain', 
            'ReadWriteInfo',
-           'ReductionInferenceTool']
+           'ReductionInferenceTool',
+           'ArrayIndexAnalysis',
+           'ArrayIndexAnalysisOptions',
+           'FortranToZ3']
```
hiker, line 58: I get several linting errors:
```
./__init__.py:47:5: E128 continuation line under-indented for visual indent
./__init__.py:53:33: W291 trailing whitespace
************* Module psyclone.psyir.tools
__init__.py:44:0: W0404: Reimport 'DefinitionUseChain' (imported line 41) (reimported)
__init__.py:58:11: E0603: Undefined variable name 'FortranToZ3' in __all__ (undefined-all-variable)

```

src/psyclone/psyir/tools/__init__.py
```diff
@@ -43,11 +43,16 @@
 from psyclone.psyir.tools.read_write_info import ReadWriteInfo
 from psyclone.psyir.tools.definition_use_chains import DefinitionUseChain
 from psyclone.psyir.tools.reduction_inference import ReductionInferenceTool
+from psyclone.psyir.tools.array_index_analysis import (ArrayIndexAnalysis,
```
hiker, line 46: Could you sort them alphabetically, this helps with avoiding duplicates :)

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -40,6 +40,7 @@
     dependency analysis.'''
 
 from enum import IntEnum
+from typing import Optional, List
```
hiker, line 43: Please sort alphabetically

src/psyclone/psyir/tools/dependency_tools.py
hiker: Can you a ticket for further Z3 work and add a #TODO here. 
In the ticket we can include all other issues that we find and address later.

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -903,9 +913,15 @@ def can_loop_be_parallelised(self, loop,
...
             is_array = symbol.is_array_access(access_info=var_info)
             if is_array:
-                # Handle arrays
-                par_able = self._array_access_parallelisable(loop_vars,
-                                                             var_info)
+                # If using the SMT-based array index analysis then do
+                # nothing for now. This analysis is run after the loop.
+                if self._use_smt_array_index_analysis:
+                    # This analysis runs after the loop
+                    par_able = True
```
hiker, line 920: I think it would be a bit easier to understand if `par_able = True` is replaced with `continue` (and of course the comment). Maybe extend the comment to:
```
# Z3 will analyse all arrays at once after this loop
```

src/psyclone/tests/psyir/nodes/omp_directives_test.py
hiker: Thanks

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
+# -----------------------------------------------------------------------------
```
hiker, line 1: I get many pylint errors. Not sure if you are aware of this:
```
              exprs_to_eval: list[z3.ExprRef] = []) 
```
is considered unsafe (since the function could modify the list, changing the default value). Standard way to avoid this is to set this to `None` (and declare as optional), and then in the body test if `expres_to_eval` is None, then set it to `[]`.

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+           :param expr_root: the expression to translate. This is assumed
+             to have scalar integer type.
+           :return: a pair containing the translated expression and a
+             list of Z3 constraints introduced by the translation.
+             One use of the constraint list is to prohibit bit-vector
+             overflow.
+        '''
+        constraints = []
+
+        def trans(e: Node) -> z3.ExprRef:
```
hiker, line 109: Could you rename `e` to (e.g) `node` or `expr`? I am actually surprised that `e` isn't flagged by linting _ I thought somehow only standard loop variables (i, j) or so would be allowed.

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+                                   IntrinsicCall.Intrinsic.MOD]:
+                    left_smt = trans(e.children[1])
+                    right_smt = trans(e.children[2])
+
+                    if (e.intrinsic == IntrinsicCall.Intrinsic.MODULO):
+                        return left_smt % right_smt
+
+                    if (e.intrinsic == IntrinsicCall.Intrinsic.MOD):
+                        m = left_smt % right_smt
+                        return (z3.If(z3.And(
```
hiker, line 203: I admit I had to check `mod` and `modulo`. Doesn't your transformation indicate that a mod with different signs in its operands is just ignored? Is that correct? If so, could you comment on this? Maybe even briefly mention the difference between mod and Z3's `%`?

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+                                          m != 0,
+                                          (left_smt < 0) != (right_smt < 0)),
+                                      m-right_smt, m))
+
+                    if self.use_bv:
+                        # TODO: when fparser supports shift operations (#428),
+                        # we can uncomment test and remove the "no cover"
+                        # block
+                        if True:  # pragma: no cover
```
hiker, line 212: Shouldn't you be able to explicitly create a PSyclone `shiftl` (etc) node, and use this to test the lines of this function?

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+
+                # N-ary operators
+                if e.intrinsic in [IntrinsicCall.Intrinsic.MIN,
+                                   IntrinsicCall.Intrinsic.MAX]:
+                    smt_args = [trans(arg) for arg in e.children[1:]]
+                    reduced = smt_args[0]
+                    for arg in smt_args[1:]:
+                        if e.intrinsic == IntrinsicCall.Intrinsic.MIN:
+                            reduced = z3.If(reduced < arg, reduced, arg)
+                        elif e.intrinsic == IntrinsicCall.Intrinsic.MAX:
```
hiker, line 269: This can just be an `else`

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+                        else:
+                            return z3.Int(array_intrins_pair[1])
+
+            # Fall through: return a fresh, unconstrained symbol
+            if self.use_bv:
+                return z3.FreshConst(z3.BitVecSort(self.int_width))
+            else:
+                return z3.FreshInt()
+
+        # Incoke the recursive function
```
hiker, line 289: s/Incoke/Invoke/

src/psyclone/psyir/tools/fortran_to_z3.py
hiker: Also, can you rename this function to be e.g. `_trans_integer` or so? Helps with finding all calls.
It might also be more test-friendly if this function would be moved out of `translate_integer_expression`, and pass in the outer `constraints` variable instead (as in: I'll leave it for you to decide what you prefer ;) ).

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+           :param expr_root: the expression to translate. This is assumed
+             to have scalar logical type.
+           :return: a pair containing the translated expression and a
+             list of Z3 constraints introduced by the translation.
+             One use of the constraint list is to prohibit bit-vector
+             overflow.
+        '''
+        constraints = []
+
+        def trans(expr: Node):
```
hiker, line 306: As above, please rename (and consider moving it out and replacing the outer `constraints` variable with an argument).

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+              If the intrinisic call is not recognised then None is returned.
+        '''
+        if call.intrinsic == IntrinsicCall.Intrinsic.SIZE:
+            var = "#size"
+        elif call.intrinsic == IntrinsicCall.Intrinsic.LBOUND:
+            var = "#lbound"
+        elif call.intrinsic == IntrinsicCall.Intrinsic.UBOUND:
+            var = "#ubound"
+        else:
+            return None
```
hiker, line 399: What are the implications if an unsupported 'inquiry` function is called? It is de-facto ignored? I am not 100% certain about the potential impact, but maybe we need a `unsupported` section in the manual. Also, if this could result in incorrect results, should we return that this is not parallelisable?

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+
+        array = call.children[1]
+        if isinstance(array, Reference):
+            (sig, indices) = array.get_signature_and_indices()
+            indices_flat = [i for inds in indices for i in inds]
+            if indices_flat == [] and len(sig) == 1:
+                var = var + "_" + sig.var_name
+                if len(call.children) == 3:
+                    rank = call.children[2]
+                    if isinstance(rank, Literal):
```
hiker, line 412: Please comment on what is happening here.

src/psyclone/psyir/tools/fortran_to_z3.py
```diff
@@ -0,0 +1,536 @@
...
+           These constraints are implicitly ANDed with the 'constraints'.
+           If the sweeper is enabled, elements of the inner lists are
+           randomly shuffled by each solver thread.
+        :param exprs_to_eval: a list of expressions to evaluate, assuming
+           the constraints are satisfiable.
+        :return: the result of the solver and a list of evaluated expressions
+           (the list is empty if the constraints were not satisifiable)
+        '''
+        if self.num_sweep_threads <= 1:
+            s = z3.Solver()
```
hiker, line 449: Rename `s` to e.g. `solver` (I really need to check what is happening with 1 letter variable names, I am sure in the past these were flagged).

src/psyclone/tests/psyir/tools/array_index_analysis_test.py
```diff
@@ -0,0 +1,381 @@
...
+from psyclone.psyir.tools import (
+    ArrayIndexAnalysis, ArrayIndexAnalysisOptions)
+
+
+# -----------------------------------------------------------------------------
+def conflict_free(loop, opts):
+    '''Helper function to determine if a loop is conflict free.
+    Note that this function returns True if the solver times out,
+    so that the test suite does not fail due to slow solving.'''
+    opts.succeed_on_timeout = True
```
hiker, line 50: This is of course not ideal - if something should break the Z3 solver handling (and it always hangs), the test would not pick that up :(

Can you add this to the ticket I've mentioned elsewhere, add a #TODO here and mark the test in this case as `xfail`?

src/psyclone/tests/psyir/tools/array_index_analysis_test.py
```diff
@@ -0,0 +1,381 @@
+# -----------------------------------------------------------------------------
```
hiker, line 1: Could you please check with `pylint`, I get quite a few warnings.

src/psyclone/tests/psyir/tools/array_index_analysis_test.py
```diff
@@ -0,0 +1,381 @@
...
+    '''
+    psyir = fortran_reader.psyir_from_source('''
+        subroutine reverse(arr)
+          real, intent(inout) :: arr(:)
+          real :: tmp
+          integer :: i, n
+          n = size(arr)
+          do i = 1, n/2
+            tmp = arr(i)
+            arr(i) = arr(n+1-i)
```
hiker, line 69: Could you parameterize this function to also include an overlapping access (`arr(i) = arr(n-i` I think should do)

src/psyclone/tests/psyir/tools/array_index_analysis_test.py
```diff
@@ -0,0 +1,381 @@
...
+    '''
+    psyir = fortran_reader.psyir_from_source('''
+        subroutine odd_even_transposition(arr, start)
+          real, intent(inout) :: arr(:)
+          integer, intent(in) :: start
+          real :: tmp
+          integer :: i
+          do i = start, size(arr), 2
+            if (arr(i) > arr(i+1)) then
+              tmp = arr(i+1)
```
hiker, line 95: Could you parameterise this (or add a new, similar test) that uses a step value, but is not conflict free (`arr(i) > arr(i+2)`).

src/psyclone/tests/psyir/tools/array_index_analysis_test.py
```diff
@@ -0,0 +1,381 @@
...
+                enddo
+              enddo
+            enddo
+          enddo
+        end subroutine my_matmul''')
+    results = []
+    opts = ArrayIndexAnalysisOptions(smt_timeout_ms=500)
+    for loop in psyir.walk(Loop):
+        results.append(conflict_free(loop, opts))
+    assert results == [True, True, False, True, True, False]
```
hiker, line 142: That is actually impressive!

src/psyclone/tests/psyir/tools/array_index_analysis_test.py
```diff
@@ -0,0 +1,381 @@
...
+        contains
+
+        subroutine chunking(arr, chunk_size)
+          integer, dimension(:), intent(inout) :: arr
+          integer, intent(in) :: chunk_size
+          integer :: n, chunk_begin, chunk_end
+
+          n = size(arr)
+          do chunk_begin = 1, n, chunk_size
+            chunk_end = min(chunk_begin+chunk_size-1, n)
```
hiker, line 160: Again, add another example that has an overlap and is detected.

src/psyclone/tests/psyir/tools/fortran_to_z3_test.py
```diff
@@ -0,0 +1,115 @@
...
+                    integer :: arr(10)
+                    integer :: i
+                    x = {expr}
+                  end subroutine''')
+        for assign in psyir.walk(Assignment):
+            (rhs_smt, cs) = trans.translate_logical_expr(assign.rhs)
+            (result, expr_vals) = trans.solve(
+                                      constraints=[rhs_smt] + cs,
+                                      exprs_to_eval=[rhs_smt])
+            assert result == z3.sat or result == z3.unknown
```
hiker, line 73: Can you add some comments here?
Also, z3.unkown indicates it didn't finish, so should we do an x-fail in this case?

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,827 @@
...
+       one using a different constraint ordering (and potentially different
+       solver parameters in future). This reduces the solver's sensitivity
+       to the order of constraints.
+
+    :param sweep_seed: the seed for the random number generator used
+       by the sweeper.
+
+    :param succeed_on_timeout: interpret a timeout as a non-conflict.
+       This means that analysis may report no conflicts when there is one,
+       but it will have tried to find one.
```
hiker, line 101: Add a warning to make it obvious that it is then in the responsibility of the script to ensure correct behaviour.

doc/developer_guide/dependency.rst
```diff
@@ -585,6 +585,34 @@ can be parallelised:
...
+
```
hiker, line 615: Can you also add documentation (e.g. autodoc) about the `ArrayIndexAnalysisOptions`, since this is being used by developers.

hiker (commented):
A bit more coming, just making sure what I have so far is not lost

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,827 @@
...
+    the point the access is made, and its list of indices (translated to
+    SMT). When we are analysing code that is inside the loop of
+    interest, we add all array accesses encountered to the list.
+
+    When we encounter the loop of interest, we perform a couple of steps
+    before recursing into the loop body. First, we kill all variables
+    written by the loop body, because we don't know whether we are
+    entering the loop (at run time) for the first time or not. Second,
+    we create two SMT variables to represent the loop variables of two
+    arbitary but distinct iterations of the loop. Each of these two
```
hiker, line 201: s/arbitary/arbitrary/

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,827 @@
...
+    list that can have the same loop indices as an access to the same
+    array in the second list, and one of which is a write.  This is
+    determined by asserting an equality constraint between each access's
+    indices which, when combined with the current condition of each
+    access and the global constraint set, will be satisfiable if and
+    only if there is a conflict.  In this way, we check every access
+    pair and determine whether or not the loop contains conflicts.
+    '''
+
+    class ArrayAccess:
```
hiker, line 226: This might be a perfect example of a dataclass (see e.g. lfric.py), which saves you the trouble of declaring properties.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,827 @@
...
+                            stop:  DataNode,
+                            step:  DataNode):
+        '''Constrain a loop variable to given start/stop/step.'''
+        zero = self._integer_val(0)
+        var_begin = self._translate_integer_expr_with_subst(start)
+        var_end = self._translate_integer_expr_with_subst(stop)
+        if step is None:
+            step = Literal("1", INTEGER_TYPE)  # pragma: no cover
+        var_step = self._translate_integer_expr_with_subst(step)
+        i = self._fresh_integer_var()
```
hiker, line 468: Can you move this to before line #476, where it is used for the first time?

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,827 @@
...
+
+    def get_loop_conflicts(self, loop: Loop, all_conflicts: bool = False) -> \
+            list[Tuple[Signature, Optional[str]]]:
+        '''Determine whether or not distinct iterations of the given loop
+           can generate conflicting array accesses.
+
+           :param loop: loop to be analysed.
+           :param all_conflicts: if True, enumerate all conflicts, otherwise
+              stop after the first conflict. Defaults to False.
+           :return: a list of pairs array-name/message pairs. If the list
```
hiker, line 538: Can you add an empty line between the parameter and `:return:`? I seem to remember that it was required in the past (otherwise sphinx would fail). I tried, and in this case it seems to work. Not sure if I have a specific newer/older version installed. Since for now it's coding style, let's play it safe and add a new line here.

src/psyclone/psyir/tools/array_index_analysis.py
```diff
@@ -0,0 +1,827 @@
...
+            self._step(stmt, z3.BoolVal(True))
+
+        # Check that we have found and analysed the loop to parallelise
+        if not (self.finished and len(self.saved_access_dicts) == 2):
+            return None  # pragma: no cover
+
+        # A list of conflicts to return
+        conflicts = []
+
+        # Forumlate constraints for solving, considering the two iterations
```
hiker, line 596: s/forumlate/formulate/

hiker (changes requested):
Thanks a lot for the work here, good stuff. I have a few issues to be addressed, but we are making progress - thanks!

PR #3145, opened 2025-09-21
1623 add training

hiker:
Adds the PSyclone training material (based on #3131)

PR #3218, opened 2025-11-13
2705 gocean inlining

### LINKED ISSUE #2705: Inlining in GOcean
The full inlining does not yet work with GOcean based code. Here the issues I have identified so far:
1) GOcean imports `field mod` without listing `r2d_field`, therefore the field type is not known, and validation of inlining aborts. It might be that just importing `r2d_field` explicitly (and whatever else is needed) explicitly would fix this.
2) field_r2d is declared as an unresolved type in gocean1p0.py:
https://github.com/stfc/PSyclone/blob/92fbf609e448f750dd2b420db7ac29d12a66c0c4/src/psyclone/gocean1p0.py#L1498-L1500
2) There are two import statements added inGOPsy: the constructor adds it to the PSy-layer symbol table (https://github.com/stfc/PSyclone/blob/92fbf609e448f750dd2b420db7ac29d12a66c0c4/src/psyclone/gocean1p0.py#L100-L102), and the `gen` code (https://github.com/stfc/PSyclone/blob/92fbf609e448f750dd2b420db7ac29d12a66c0c4/src/psyclone/gocean1p0.py#L123) adds this as well. I believe atm the `gen` code is responsible for the `use` statement that turns up in the output psy-layer file, but the symbol table is used to avoid name clashes.

When I tried adding a proper declaration to the module symbol table, the code mentioned in item 2 would create a new symbol avoiding the name clash with `r2d_field`. I realise now that this was likely because I created the symbol without a tag :(

I also tried to disable the validation, and while them inlining worked, the backend visitor crashed because it hit the undefined symbol r2d_field.

I could get the inlining to work after some hacky changes to gocean.py - adding the r2d_field symbol and fixing up the renaming (due to the tag I missed I believe), so there doesn't appear to be anything mayor missing.

hiker:
A small PR that fixes #2705. It is basically Sergi's patch for the training, moved into a stand-alone PR and updated tests.
This way the training does not affect any files outside of the training directory.

The PR adds a manual basic definition of `r2d_fields`, so that inlining finds the type information and proceeds as expected.

This blocks #3145 (adding training).

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   1) PR #3218 for #2705. Adds explicit type information for the r2d_field
+   type in GOcean so that kernel inlining succeeds.
+
 Release 3.2.1 12th of November 2025
 
    17) PR #3214 for #3188. Improves the validation when doing automatic
--- a/src/psyclone/version.py
+++ b/src/psyclone/version.py
@@ -43,5 +43,5 @@ __MAJOR__ = 3
 __MINOR__ = 2
 __MICRO__ = 1
 
-__SHORT_VERSION__ = f"{__MAJOR__:d}.{__MINOR__:d}"
-__VERSION__ = f"{__MAJOR__:d}.{__MINOR__:d}.{__MICRO__:d}"
+__SHORT_VERSION__ = f"{__MAJOR__:d}.{__MINOR__:d}-dev"
+__VERSION__ = f"{__MAJOR__:d}.{__MINOR__:d}.{__MICRO__:d}-dev"


hiker:
Integration tests passed as well. Ready for review.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -201,65 +202,33 @@ def test_apply_array_access(fortran_reader, fortran_writer, tmpdir):
...
+    # replace invokes with calls)
+    for kern in invoke.schedule.walk(Kern):
+        kmit.apply(kern)
+        kern.lower_to_language_level()
+    # Then inline the calls
+    for call in invoke.schedule.walk(Call):
+        inline_trans.apply(call)
+
+    output = fortran_writer(invoke.schedule)
+    expected = ("  do j = cu_fld%internal%ystart, cu_fld%internal%ystop, 1\n"
```
arporter, line 225: Whoop!

arporter (approved):
Excellent. I love it when things start to work :-)
I see the integration tests were all green.
I'll proceed to merge.

Issue #2884, opened 2025-02-04
Add support for WHERE with imported symbols


sergisiso:
WHEREs with imported symbols create a CodeBlock with:
`PSyclone doesn't yet support reference to imported symbols inside WHERE clauses.`

Now that we can resolve modules in the frontend before iterating down to the execution section we can potentially fix this. Or improve the message to provide the instructions to fix it.

RobWatersMet:
Just to second this as we have a number of these in UKCA. 

For example:

https://code.metoffice.gov.uk/trac/ukca/browser/main/trunk/src/science/core/chemistry/asad/asad_hetero.F90#L283
https://code.metoffice.gov.uk/trac/ukca/browser/main/trunk/src/science/core/chemistry/asad/asad_hetero.F90#L322

PR #3221, opened 2025-11-13
(towards #2884) Cleanup Reference2ArrayRange

sergisiso:
Looking how we can solve the WHERE with imports I found that we need to replicate logic to check if a plain reference is an array or a scalar or unresolved, ... this is also needed in InlineTrans and in ArrayAssignment2LoopsTrans, so I am not fond in duplicating this a third time.

I believe the problem is that Reference2ArrayRangeTrans validates that is is being applied to an array, which we don't know in advance (unless we did the proper validation), and often use it as:
```
try:
    Reference2ArrayRangeTrans().apply(reference)
except TransformationError:
    pass
```
I think it would be more useful to change the interface of this transformation to pass validation for any given ref (regardless of it being transformed to an ArrayRef or not), but fail validation in the case that we cannot guarantee what it is. Which is really what we need to continue with the WHERE, Inline, ArrayAssignment2Loop. They know what to do with scalars, but not with symbols that we don't know what they are.

However, before chaning the interface, I wanted to do a smaller PR cleaning up the current implementation: deleting some dead code and the 'allow_call_arguments', which I think is responsibility of the InlineTrans. Doing that I also thought the InlineTrans would be clearer changing the boundary between two internal methods.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,5 @@
+   10) PR #3221 towards #2884. Cleans-up Reference2ArrayRangeTrans.
+ 
    9) PR #3235 for #3234 and #3232. Add support for OpenMP conditional
    compilation statements.
 


src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -777,9 +732,10 @@ def _replace_formal_struc_arg(
         use_first_callee_and_no_arg_check: bool = False,
     ) -> Reference:
         '''
-        Called by _replace_formal_arg() whenever a formal or actual argument
-        involves an array or structure access that can't be handled with a
-        simple substitution, e.g.
+        Called by _replace_formal_args_in_expr() whenever a reference to
+        the formal argument is found. This will been to be replaced with
```
arporter: s/been/need/

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -506,41 +493,9 @@ def _replace_formal_arg(
...
-        #
-        #   subroutine my_sub(var)
-        #     ...
-        #     var%data(i,j) = 0.0
-        #
-        if type(actual_arg) is Reference:
-            ref.symbol = actual_arg.symbol
-            return ref
-
         # Neither the actual or local references are simple, i.e. they
```
arporter: Please update this comment.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -815,6 +771,54 @@ def _replace_formal_struc_arg(
...
+        # and if so, what are their boundaries. To do so we can use the
+        # Reference2ArrayRangeTrans
+        if type(actual_arg) is Reference:
+            if isinstance(actual_arg.datatype, ArrayType):
+                dummy = Assignment()
+                dummy.addchild(actual_arg)
+                Reference2ArrayRangeTrans().apply(actual_arg)
+                actual_arg = dummy.children[0]
+
+        # Local reference is not simple but the actual argument is, e.g.:
```
arporter: This comment also needs updating now I think.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -815,6 +771,54 @@ def _replace_formal_struc_arg(
...
+        #
+        #   call my_sub(my_struc)
+        #
+        #   subroutine my_sub(var)
+        #     ...
+        #     var%data(i,j) = 0.0
+        #
+        if type(actual_arg) is Reference:
```
arporter, line 810: Perhaps it needs a new comment along the lines of "Actual argument is a scalar quantity"?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -815,6 +771,54 @@ def _replace_formal_struc_arg(
...
+                    call_node,
+                    formal_args,
+                    routine_node,
+                    use_first_callee_and_no_arg_check=(
+                        use_first_callee_and_no_arg_check),
+                )
+
+            return new_ref
+
         # The final stage of this method creates a brand new
```
arporter: If I'm understanding correctly, this comment also needs updating as we may just have actual_arg being an ArrayReference now too?

src/psyclone/psyir/transformations/reference2arrayrange_trans.py
```diff
@@ -84,71 +84,25 @@ class Reference2ArrayRangeTrans(Transformation):
     structures, see issue #1858.
 
     '''
-    @staticmethod
-    def _get_array_bound(symbol, index):
```
arporter, line 88: Ah! So this method was unused?

arporter (changes requested):
Thanks for this tidying Sergi. I think some of the comments need updating now to match the new logic. Good work on deleting dead code! Although, now that I write that, how come we hadn't spotted that in the coverage? Had it just slipped through at some point previously?

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: Ups, fixed

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: Done, also updated the method name to make it more clear

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: I updated the comment, but I think this was also true before the change.

src/psyclone/psyir/transformations/reference2arrayrange_trans.py
sergisiso: yeah, I think at some point be moved this functionality to a method of ArrayMixin and forgot about this

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: Done

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: Done

Issue #3219, opened 2025-11-13
Add children nodes to Kern to represent their computation pattern


sergisiso:
This follows from the discussion in https://github.com/stfc/PSyclone/pull/3164#issuecomment-3421413588

In order to do some advanced validation (e.g. check if two loops can be fused), it is not enough to know that the nodes are kernels, but we also need to know their computation pattern (e.g. which stencil position it accesses in each of its fields) in order to check for dependencies with other kernels.

Currently we could inline the kernel body to do this validations but this has several limitations: we lose the DSL semantics from that point, inline could fail, the body could be hard to analyse (e.g. it has CodeBlocks), ...

A different solution to this would be to add as children of KernCall a prototype of the computation pattern it represents, e.g.:
```
 KernelCall
    arg1%data(i,j) = ptwise%data(i,j) + stencil%data(i, j-1) + stencil%data(i, j+1)
```

This can be generated from the metadata and would still work with the core PSyIR tools (they children with the correct read/write accesses to specific data that represents the computation pattern of the kernel).

These won't make it to code because the lowering replaces the whole KernelCall with the actual Call. The only transformation where we need to be careful is in Hoisting, as we must prevent this expression be hoisted outside of the KernelCall.

arporter:
#3223 was only a 'towards' - LFRicKern still needs to be done.

PR #3223, opened 2025-11-15
(towards #3219) Add a prototype of the kernel computation pattern as its children

sergisiso:
The idea is that the code PSyIR tools (e.g. dependency analysis) will do the right thing without needed a specialised implementation.


sergisiso:
@hiker This is ready for a first review. Let me know if this helps with the training.

@hiker @arporter I have added and then commented out some validations regarding what it is or it isn't a valid kernel. Do you think kernels should have one (and only one) write location? In the tests we have several that don't follow this.

src/psyclone/gocean1p0.py
```diff
@@ -207,6 +208,8 @@ class GOLoop(PSyLoop):
 
         :param parent: optional parent node (default None).
         :type parent: :py:class:`psyclone.psyir.nodes.Node`
+        :param variable: the loop iteration variable.
+        :type variable: :py:class:`psyclone.psyir.symbols.Symbol`
```
hiker: Please use typing (and you are also an unfortunate victim of 'you touch it, you update it' :) )

src/psyclone/gocean1p0.py
```diff
@@ -218,8 +221,8 @@ class GOLoop(PSyLoop):
     '''
     _bounds_lookup = {}
 
-    def __init__(self, parent, loop_type="", field_name="", field_space="",
-                 iteration_space="", index_offset=""):
+    def __init__(self, parent, variable, loop_type="", field_name="",
+                 field_space="", iteration_space="", index_offset=""):
```
hiker: Ah, that's actually the better place to request typing ;)

src/psyclone/gocean1p0.py
```diff
@@ -232,44 +235,18 @@ def __init__(self, parent, loop_type="", field_name="", field_space="",
...
-            suggested_name = "i"
-        elif self.loop_type == "outer":
-            tag = "noncontiguous_kidx"
-            suggested_name = "j"
-        else:
+        if self.loop_type not in ["inner", "outer"]:
             raise InternalError(f"While the loop type '{self._loop_type}' is "
                                 f"valid, it is not yet supported.")
 
-        # In the GOcean API the loop iteration variables are declared in the
```
hiker, line 255: It's nice to see this go :)

src/psyclone/gocean1p0.py
```diff
@@ -232,44 +235,18 @@ def __init__(self, parent, loop_type="", field_name="", field_space="",
...
-                suggested_name, tag, symbol_type=DataSymbol,
-                datatype=INTEGER_TYPE)
-
         # Initialise bounds lookup map if it is not already
         if not GOLoop._bounds_lookup:
             GOLoop.setup_bounds()
 
     @staticmethod
-    def create(parent, loop_type, field_name="", field_space="",
+    def create(parent, variable, loop_type, field_name="", field_space="",
```
hiker: Typing (I won't comment about this any further though)

src/psyclone/gocean1p0.py
```diff
@@ -891,12 +870,32 @@ def create(call, parent=None):
         :rtype: :py:class:`psyclone.gocean1p0.GOLoop`
 
         '''
+        # Check that the GOKern is inside the GOcean PSy-layer
+        if (
```
hiker, line 887: That feels like a better place to test this (and allow more flexibility in testing I would assume, since we can create loops without parents now :) )

src/psyclone/gocean1p0.py
```diff
@@ -938,7 +939,7 @@ class GOKern(CodedKern):
     :type parent: :py:class:`psyclone.psyir.nodes.Node`
 
     '''
-    def __init__(self, call, parent=None):
+    def __init__(self, call, inner_symbol, outer_symbol, parent=None):
```
hiker: Documentation of the new parameter is missing (and typing).

src/psyclone/gocean1p0.py
```diff
@@ -947,74 +948,122 @@ def __init__(self, call, parent=None):
...
+                        # until reaching the depth value
+                        depth = arg.stencil.depth(i, j)
+                        for offset in range(1, depth+1):
+                            if i == 0 and j == 0:
+                                # In this case it just refers to pointwise
+                                offset = 0
+
+                            inner_index = Reference(inner_symbol)
+                            outer_index = Reference(outer_symbol)
+                            if j > 0:
```
hiker: That feels like it should have one small function here, which takes the offset, and variable, and creates the `variables+-offset` psyir.

src/psyclone/gocean1p0.py
```diff
@@ -947,74 +948,122 @@ def __init__(self, call, parent=None):
...
+        #     raise InternalError(
+        #         "This is not a valid kernel, a kernel must write "
+        #         "to one field.")
+        if not read_accesses:
+            # There can be write-only kernels, in this case we just put
+            # a constant literal
+            rhs = Literal("1", INTEGER_TYPE)
+        else:
+            rhs = read_accesses.pop(0)
+            while read_accesses:
```
hiker, line 1079: I have to admit I felt a bit uneasy about creating a proper Fortran expression (`write=read+read+...`), but I can't find anything wrong with that, and indeed it should have the beauty that the variable access sequences (and then the DA) just work out of the box.

src/psyclone/gocean1p0.py
```diff
@@ -947,74 +948,122 @@ def __init__(self, call, parent=None):
...
+                rhs = BinaryOperation.create(
+                        BinaryOperation.Operator.ADD,
+                        rhs, read_accesses.pop(0))
+        return Assignment.create(write_access, rhs)
```
hiker: I do realise a small change here, which might have huge impact. As far as I can see, now all psydata libraries will not get the field provided anymore, but the `field%data` 2D Fortran array.

Now, the good news of this is of course it might mean we don't need domain-specific implementations for gocean anymore.

The downside - it's hard to evaluate if we might break something somewhere :(  That statement is triggered by the fact that only very few changes are required in the tests. Looking at `readonly`, indeed, the only change would be the error message (not talking of GOcean 'field' anymore, instead it would be array). The gocean example passes. I will have to think about this a bit, and look into other psy data libraries.

hiker (changes requested):
Nice change. I have a few small changes requested.

The for me potential bigger issue: the psy-data libraries will now receive a 2d array instead of a gocean field. That should work, but I need to have a closer look to check if we are not breaking something (by removing the DSL knowledge and going to Fortran arrays only). Of course, being able to remove gocean-specific libs would be nice ... just scary ;)

hiker:
Before I forget this:
1. this PR would then also need to update the examples (at least the README in gocean/eg5/readonly contains the old output (field instead of array)
2. We also need to check the documentation, this should be explained somewhere in the dev guide.

sergisiso:
> "> Do you think kernels should have one (and only one) write location? In the tests we have several that don't follow this."
>IMHO, we should not require. In the MOST Tsunami code for example, we have three output variables in one kernel:

Ah, I see no problem with those, each writing to the i,j index of a different variable. These can be prototyped as:
output1(i,j) = ...all reads...
output2(i,j) = ...all reads...

My question was more for several location on the same variable e.g. writes that also have go_stencil kernel type. The dependency analysis of these are really complex anyway. What I am suggesting is to require all writes to be pointwise.

What about the question about kernels with no writes at all (supposedly kernel can not have side-effects), so what would be the point of such kernel?

src/psyclone/gocean1p0.py
sergisiso: Done

src/psyclone/gocean1p0.py
sergisiso: Yeah, added them here

src/psyclone/gocean1p0.py
sergisiso: Done

src/psyclone/gocean1p0.py
sergisiso: Done

src/psyclone/gocean1p0.py
sergisiso: Done

doc/developer_guide/dependency.rst
```diff
@@ -139,8 +139,44 @@ PSyKAl analysis methods:
...
+    cu_fld%data(i,j) = p_fld%data(i,j) + p_fld%data(i,j + 1) + u_fld%data(i,j)
+
+This assignment will not be generated to code because the lowering method
+replaces the entire CodedKern with the appropriate call. But this child is
+sufficient to provide the `VariablesAccessMap` used by generic PSyIR the
+necessary access information to perform the needed depenency checks.
+
+LFric does not currently use this virtual assignments, instead it uses
+information provided by the `Argument` class attached to the CodedKern to
+evaluate the ordering constraints between between kernels.
```
hiker, line 179: Thanks, nice

src/psyclone/gocean1p0.py
hiker: I am happy to accept this. Removing the need for domain-specific libraries is a very nice thing to get. I believe, if we should really have a case of a psydata library that needs the actual fields, the corresponding code could implements its own way of providing the fields when lowering.

src/psyclone/tests/psyir/transformations/read_only_verify_trans_test.py
```diff
@@ -105,7 +105,7 @@ def test_read_only_options():
                     options={"region_name": ("a", "b")})
     code = str(psy.gen)
 
-    assert 'CALL read_only_verify_psy_data % PreStart("a", "b", 4, 4)' in code
+    assert 'CALL read_only_verify_psy_data % PreStart("a", "b", 5, 5)' in code
```
hiker: Why do we only get `j` added here? To some degree that is correct (since `i` is a kind of 'local only' variable, but ... I actually don't know why it is not added.  The expression as kernel child has both `i` and `j`.

hiker:
@arporter , we might want your feedback here as well about what a gocean kernel should be allowed to look like (e.g. does it have to have a write?).

TBH, I am not sure if this consistency check is strictly related to this PR, and I would be happy to just have a ticket for that, and get the rest merged in.

I think I had one question (why is `j` added and not `i`), I have to admit that I am a bit confused by this behaviour, which is kind of right, but I just don't understand why it works :)

src/psyclone/tests/psyir/transformations/read_only_verify_trans_test.py
sergisiso: It is what read_only_verify is provided with as pre_list `src/psyclone/psyir/nodes/value_range_check_node.py::85`
```
read_write_info = ctu.get_in_out_parameters(self)
return {'pre_var_list': read_write_info.read_list,
        'post_var_list': read_write_info.write_list}
```
here read_list means input variables to the selected node.

Note that this is different from exctraction `src/psyclone/psyir/nodes/extract_node.py:203` which provides all used variables in the pre_list:
```
  options = {'pre_var_list': read_write_info.all_used_vars_list,
             'post_var_list': read_write_info.write_list,
             'post_var_postfix': self._post_name}
```

sergisiso: If I remember correctly, in extraction this was changed to all_used_vars because get_in_out_parameters can not distinguish variables that are writing just in part of their range as input. For 'i' this is not a concern because it is scalar (it cannot have partial writes). Maybe it is a concern for arrays and we should treat them as we do in extraction? But this is probably out of scope for this PR.

hiker: After some debugging: not your fault, but can you change this test to use
```
read_only.apply(invoke.schedule[0], ...
```
in line 104. This will apply the readonly test at the appropriate level.
Then the `prestart` will get `..., 6, 6` as parameter (adding two new loop boundaries for the j loop, and removing 'j' since it's now properly identified as a local variable).

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
```diff
@@ -126,61 +209,43 @@ def test_gok_get_callees():
...
+    # (because they can only be accessed as pointwise reads anyway)
+    assert kern1.children[0].debug_string() == (
+        "p_fld%data(i,j) = p_fld%data(i,j) + "
+        "p_fld%grid%subdomain%internal%xstop + p_fld%grid%tmask\n"
+    )
 
     # Check that we get the grid properties listed:
+    vam = kern1.reference_accesses()
     assert (str(vam) ==
+            "i: READ, j: READ, p_fld%data: READ+WRITE, "
```
hiker: I just noticed this change. We basically go from `READWRITE` (which was taken from the kernel specification) to a `READ+WRITE`.  Now looking at this, I think the kernel is declared incorrectly (that field is actually written only??), but ignoring this:
is it a problem that we change from `READWRITE` (based on kernel) to `READ+WRITE` (based on the Kernel child assignment)? I can't see any negative side-effect, but am a bit fuzzy with the various READ+WRITE, READWRITE, READINC, SUM differences ... admittedly they seem to be mostly LFRic related?

Anyway, I am happy with this change, but just want to confirm that this is indeed ok.

src/psyclone/tests/psyir/transformations/value_range_check_trans_test.py
```diff
@@ -104,7 +104,7 @@ def test_value_range_check_options(fortran_writer):
     value_range_check.apply(invoke.schedule[0].loop_body[0],
                             options={"region_name": ("a", "b")})
     code = fortran_writer(invoke.schedule)
-    assert 'CALL value_range_check_psy_data % PreStart("a", "b", 4, 2)' in code
+    assert 'CALL value_range_check_psy_data % PreStart("a", "b", 5, 2)' in code
```
hiker: Same problem here. This changes to `6,3`: two more loop boundaries as input, and one more loop variable (j) as output.

src/psyclone/tests/psyir/transformations/read_only_verify_trans_test.py
sergisiso: Done as suggested

src/psyclone/tests/psyir/transformations/value_range_check_trans_test.py
sergisiso: Done as suggested

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
sergisiso: Well spotted, I updated the kernel metadata.

Regarding READWRITE vs READ+WRITE there should be no difference as code requesting this information should always use `access_info.is_any_read() ` / `access_info.is_any_write() `, otherwise it is probably faulty because as you mention there are other types that also imply reads or writes.

As I see it the only difference is that in READ+WRITE we can guarantee that the READ happens before the WRITE, and in the READWRITE we cannot (e.g. is an inout argument). But this is not a problem because when we don't know the safe choice is that the READ comes first, so for safety it should always be treated equivalent to READ+WRITE.

sergisiso:
@hiker I addressed the two ongoing comments, ready for another review.

src/psyclone/tests/psyir/tools/dependency_tools_test.py
```diff
@@ -740,13 +740,20 @@ def test_reserved_words(fortran_reader):
...
     _, invoke = get_invoke("test31_stencil_not_parallel.f90",
                            api="gocean", idx=0, dist_mem=False)
 
     loop = invoke.schedule.children[0]
     dep_tools = DependencyTools()
     parallel = dep_tools.can_loop_be_parallelised(loop)
+
+    if parallel is True:
+        pytest.xfail(
+            reason=("TODO #2531: GOcean metadata validation should not allow"
```
hiker, line 754: While I don't think this comment is entirely correct (since the kernel actually does not write outside pointwise location. It reads from non-pointwise, but only writes to pointwise), we can leave this for #2531 (atm I don't think we can actually distinguish based on the meta-data alone if a kernel writes outside of pointwise if an argument is readwrite).

src/psyclone/tests/psyir/transformations/read_only_verify_trans_test.py
hiker: Thanks!

src/psyclone/tests/domain/gocean/kernel/gokern_test.py
hiker: Thanks for the comments :)

hiker (approved):
Integration tests were green - approved.

Issue #984, opened 2020-11-16
reproducible reductions broken


rupertford:
When reproducible reductions are switched on we get:

`"Could not find the tag 'omp_num_threads' in the Symbol Table."`

rupertford:
Create branch [984_reproducible_reductions](https://github.com/stfc/PSyclone/tree/984_reproducible_reductions)

PR #3226, opened 2025-11-18
(Closes #514) refactor existing reduction suport to use OMPReductionClause

### LINKED ISSUE #514: [PSyIR] Add support for reductions and 'omp parallel do'

arporter:


src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
```diff
@@ -274,3 +275,30 @@ def _should_it_be_fprivate(
...
         # If not, it can be just 'private'
         return False
+
+    def _add_reduction_clauses(self):
+        '''
+        Analyses the code beneath this node and adds the necessary
+        OMPReductionClause nodes.
+
+        '''
+        _, _, need_sync = self.infer_sharing_attributes()
```
sergisiso: infer_sharing_attrbutes is one of the most computationally espensive steps in psyclone. I see that when lowering "omp do" we need to do it because otherwise we don't perform this analysis for this loop. But for "omp parallel do" (src/psyclone/psyir/nodes/omp_directives.py line 1365 we could provide as an optional agument the already calculated need_sync, and here only compute it if no argument is provided). What do you think?

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
```diff
@@ -274,3 +275,30 @@ def _should_it_be_fprivate(
...
+
+        from psyclone.psyir.tools.reduction_inference import (
+            ReductionInferenceTool)
+        from psyclone.psyir.transformations.omp_loop_trans import (
+            MAP_REDUCTION_OP_TO_OMP)
```
sergisiso: This map now feels in the wrong place, should we move it to src/psyclone/psyir/nodes/omp_directives.py ?

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -2409,43 +2417,46 @@ def test_two_reprod_reductions_real_do(tmpdir, dist_mem):
...
+            "    local_bsum = 0.0_r_def\n"
             "\n"
             "    ! Call kernels\n"
             "    !$omp parallel default(shared) private(df,th_idx)\n"
             "    th_idx = omp_get_thread_num() + 1\n"
             "    !$omp do schedule(static)\n"
             "    do df = loop0_start, loop0_stop, 1\n"
             "      ! Built-in: X_innerproduct_Y (real-valued fields)\n"
-            "      l_asum(1,th_idx) = l_asum(1,th_idx) + "
+            "      local_asum(1,th_idx) = local_asum(1,th_idx) + "
```
sergisiso, line 2445: In `tutorial/practicals/LFRic/single_node/1_openmp/` we do

```
    otrans = LFRicOMPParallelLoopTrans()
    ptrans = OMPParallelTrans()
    ltrans = LFRicOMPLoopTrans()

    # Make sure reductions are reproducible
    if loop.reductions():
        ptrans.apply(loop)
        ltrans.apply(loop, {"reprod": True})
    else:
        otrans.apply(loop)
```
Is there aray reason we don't support `LFRicOMPParallelLoopTrans().apply(loop, {"reprod": True})`? naively it should be the same

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -83,8 +84,7 @@
 from psyclone.psyir.nodes.schedule import Schedule
 from psyclone.psyir.nodes.structure_reference import StructureReference
 from psyclone.psyir.symbols import (
-    INTEGER_TYPE, DataSymbol, ImportInterface, ContainerSymbol,
-    RoutineSymbol)
```
sergisiso, line 87: In `src/psyclone/tests/psyir/nodes/omp_directives_test.py` there are multiple instances of 
`# TODO #2446 and #514: improve this error message in future`

We need to delete at least the 514 part to close this issue.

Optionally (it could be self to another PR), coudl we make the error message
say something more sensible, e.g.:
"but is not supported by current PSyclone reductions implementation."

src/psyclone/psyir/nodes/omp_directives.py
```diff
@@ -1332,37 +1298,32 @@ def lower_to_language_level(self):
                         " region with reductions and containing children of "
                         "different types.")
 
-        # pylint: disable=import-outside-toplevel
```
sergisiso, line 1335: Could you add a comment at the top of this method explaining that "Generic code reductions have already been introduced as directive clauses when applying the Parallelisation transformation, but DSL reductions (given by self.reductions) need to be added during the lowering."

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -1904,8 +1904,8 @@ def test_reduction_real_pdo(tmpdir, dist_mem):
     if dist_mem:
         assert "loop0_stop = f1_proxy%vspace%get_last_dof_owned()" in code
         assert (
-            "    !$omp parallel do reduction(+:asum) default(shared) "
-            "private(df) schedule(static)\n"
+            "    !$omp parallel do default(shared) "
+            "private(df) schedule(static) reduction(+: asum)\n"
```
sergisiso, line 1908: In this or another test with two reductions, can you check if #984 is still a problem?

Another thing we coudl try is fusing (which may work with dof kernels?) and see if it generates different variables names and the appropriate clause. But no worries if it doesn't.

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
```diff
@@ -274,3 +275,30 @@ def _should_it_be_fprivate(
...
+
+        vam = self.children[0].reference_accesses()
+        # Currently we only support summation reductions.
+        red_tool = ReductionInferenceTool(
+            [BinaryOperation.Operator.ADD])
```
sergisiso: Maybe "Currenlty DSL-code only supports summation reductions. Generic code supports more reduction operations but these are inserted during the transformation"

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
arporter: That's a good idea. I've done it.

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
arporter: Yes, done.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
arporter: I thought so too so I tried it and then realised that it isn't the same: we currently query and store the index of the current thread so that we can store its reduction result in a separate element of the result array:
```fortran
    !$omp parallel default(shared) private(df,th_idx)
    th_idx = omp_get_thread_num() + 1
    !$omp do schedule(static)
    do df = loop6_start, loop6_stop, 1
      ! Built-in: X_innerproduct_X (real-valued field)
      local_sum(1,th_idx) = local_sum(1,th_idx) + p_data(df) * p_data(df)
    enddo
    !$omp end do
    !$omp end parallel
```
The lookup-and-store of the thread index needs to happen in the same parallel region as the loop. The only way we could do this for a `parallel do` is lookup the thread index in each loop iteration which is probably not ideal.

src/psyclone/psyir/nodes/omp_directives.py
arporter: I've added a bit to the error message and added logging to ReductionInferenceTool.

src/psyclone/psyir/nodes/omp_directives.py
arporter: I've updated the docstring using your words.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
arporter: I chatted this through with @LonelyCat124 today and he thought that doing an in-place `omp_get_thread_num()` might be painless so it is an option. Perhaps separate to this PR though?

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
arporter: We already have a test with two reductions:
https://github.com/stfc/PSyclone/blob/114473775376b5792830c7cb446c9ea0c8682454/src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py#L3637-L3641
and it is fine so I've closed #984.

arporter: I've extended some tests and discovered that we have:
https://github.com/stfc/PSyclone/blob/114473775376b5792830c7cb446c9ea0c8682454/src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py#L225-L235
so we refuse to fuse two loops that both do reductions. We could improve on this at some point. Also, most dof kernels have arguments on 'any space' so we can't fuse them without the additional argument to say that the spaces should be assumed to be the same.

src/psyclone/psyir/nodes/data_sharing_attribute_mixin.py
arporter: Yes, done.

PR #3239, opened 2025-11-28
(Closes #3199) Initial implementation of OMPCritical

### LINKED ISSUE #3199: omp critical directives
Simialr to #3198 

Ideally we'd like to replace exisiting omp directives with identical equivalents from a transformer - This will help ensure the directives are not mistakenly stripped out in the future and will improve consistency in how omp directives are being applied to UKCA.

E.g.:
```fortran
! OMP CRITICAL will only allow one thread through this code at a time,
! while the other threads are held until completion.
!$OMP CRITICAL (asad_bedriv_init)
IF (first_pass) THEN
  IF (first) THEN
    first = .FALSE.
    ! find N2O5, NO3, and NO2 species in species array
    DO js = 1,jpspec
      SELECT CASE (speci(js))
      CASE ('N2O5      ')
        in2o5 = js
      CASE ('NO3       ')
        ino3  = js
      CASE ('NO2       ')
        ino2  = js
      END SELECT
    END DO

    ! find N2O5 + M and NO2 + NO3 + M in termolecular reactions array
    DO js = 1,jptk
      SELECT CASE (spt(js,1))
      CASE ('N2O5      ')
        tn2o5   = ntrkx(js)
      CASE ('NO2       ')
        IF (spt(js,2) == 'NO3       ') tno2no3 = ntrkx(js)
      CASE ('NO3       ')
        IF (spt(js,2) == 'NO2       ') tno2no3 = ntrkx(js)
      END SELECT
    END DO

```

LonelyCat124:
This PR contains the implementation of OpenMP critical regions.

At this point, correctness is left to the user. I'm adding this for use in UKCA, with the idea that correct usage is currently left to the user, as for their use case even with more improvements PSyclone would not be able to do analysis reasonably (as the `parallel do` and `omp critical` regions are in different files).

In terms of future correctness, I think post 3238 improvements we might be able to do better, as critical regions have atomicity guaranteed.

I also chose not to care about hint or name, as these massively complicated this and I'm not sure its worth adding without a use-case.

@sergisiso @arporter ready for review when we have spare resource - no rush I think for this to be merged, though I think it is a simple change unless you dislike leaving its correct usage to the user.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,5 @@
+   13) PR #3239 for #3199. Add OMPCritical Directive and Transformation.
+
    12) PR #3226 for #514. Refactor LFRic reductions to use OMPReductionClause.
 
    11) PR #3249 for #3246. Add the ArrayBounds.copy method.


src/psyclone/tests/psyir/transformations/omp_critical_trans_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+    psyir = fortran_reader.psyir_from_source(code)
+
+    routine = psyir.walk(Routine)[0]
+
+    ctrans.apply(routine.children[:])
+    assert isinstance(routine.children[0], OMPCriticalDirective)
+
+    output = fortran_writer(psyir)
+    print(output)
```
sergisiso: Remove

doc/user_guide/transformations.rst
```diff
@@ -382,6 +382,15 @@ can be found in the API-specific sections).
 
 ####
 
+.. autoclass:: psyclone.psyir.transformations.OMPCriticalTrans
+    :members: apply
+    :no-index:
+
+.. note:: PSyclone does not perform any checks on code generated with
+          OMPCriticalTrans, and it's correct usage is up to the user.
```
sergisiso: This seems duplicated because it is already mentioned inside the docstring embeded just above. So I would deleted it. However, I agree the ".. note::" syntax make it look more highlighted. Could you instead move this tag inside the OMPCriticalTrans docsring second paragraph.

src/psyclone/psyir/nodes/omp_critical_directive.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class OMPCriticalDirective(OMPRegionDirective):
+    '''
+    OpenMP directive to inform that the contained region must only be executed
+    by a single thread at any time.
+
+    PSyclone does not currently do any analysis about critical regions, and
+    their correct usage is up to the user. In the future, we may try to
+    improve this (after issues like #3238).
```
sergisiso: I would remove this from here. This seems the concern of the transformation that already has this message.

src/psyclone/psyir/nodes/omp_critical_directive.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def end_string(self) -> str:
+        '''
+        :returns: the ending string statement of this directive.
+        '''
+        return "omp end critical"
+
+
+# For automatic API documentation generation
+__all__ = ["OMPCriticalDirective"]
```
sergisiso: I know this was mentioned in Teams and I didn't speak up, but I disagree in splitting omp to new files. I believe these will end up being very shallow classes (with similarly shallow tests). And I think splitting them all will just create more complexity that they really represent.

I don't disagree that the size of omp_directives.py needs to be reduced, but I think that other logic currently there has to be moved to a better place, not the shallow clases. And if we start moving them to multiple files it will make the actual refactoring more difficult. So unless @arporter and you disagree, I think I would still put this class in the main omp file.

src/psyclone/psyir/transformations/omp_critical_trans.py
```diff
@@ -0,0 +1,84 @@
...
+from psyclone.psyir.transformations.region_trans import RegionTrans
+from psyclone.utils import transformation_documentation_wrapper
+
+
+@transformation_documentation_wrapper
+class OMPCriticalTrans(RegionTrans):
+    '''
+    Adds an OpenMP critical directive around a region of code.
+
+    PSyclone does not currently do any analysis about critical regions, and
```
sergisiso: Maybe s/PSyclone/This transformation/ s/analysis/validation/ and the ".. note::"

src/psyclone/tests/psyir/transformations/omp_critical_trans_test.py
LonelyCat124: Done

doc/user_guide/transformations.rst
LonelyCat124: Added the note syntax to the transformation and removed this section from here.

src/psyclone/psyir/nodes/omp_critical_directive.py
LonelyCat124: Removed.

src/psyclone/psyir/nodes/omp_critical_directive.py
LonelyCat124: Ok - perhaps if we create new complex classes it might be better to have them in separate files, but small ones like this might be unnecessary.

src/psyclone/psyir/transformations/omp_critical_trans.py
LonelyCat124: Changed.

sergisiso (approved):
Thanks @LonelyCat124 all looks good now, this is currently used only  in its own test so I won't lauch integration tests. I will queue this for merging.

Issue #3124, opened 2025-09-04
The Loop variable (and maybe the codeblocks and kernels symbols?) should be references


sergisiso:
Currently the loop variable in a Loop is directly in a symbol attribute, not in a Children of type Reference. This makes some intrastructure meant for Refernces stligly odd, for instances the reference_accesses() sometimes returns Loop nodes, similarly the dependency analysis return the Loop node, and we need special casing every time we use them and get something which is not a Reference.

Similarly for CodeBlocks, but I am less sure (about what to do with these) so we may decide clossing the issue with just the Loop variable References

PS: kernels is another place, but as with codeblocks it may be complicated

schreiberx:
I can add my experience to this:

I had to deal with do loops and loop variables that serve as counters for the number of iterations in some iterative solvers. In this case, the loop variable itself is not used anywhere as another Reference.
This case didn't work for automatic variable elimination of unused variables because the loop variable is not accessible as `.walk(Reference)`, and the algorithm assumed it was not used at all. I had to use a workaround for this.
It's probably another very clear case where creating a reference would be much clearer.

sergisiso:
Loops, CodeBlocks and GOcean kernels are now done, the remaining part is LFRic kernels.

The idea is to implement the same solution as in GOcean, having children equivalent to the prototype computation of the kernel as provided by the metadata access pattern.

Once this is done `reference_accesses` should provide the correct information to fix https://github.com/stfc/PSyclone/issues/2498

PR #3247, opened 2025-12-02
(towards #3124, closes #831) Codeblocks with children references

### LINKED ISSUE #831: code block symbols should be added to a symbol table
The fparser2 frontend should look in code blocks and add any symbols to the symbol table in order to avoid name clashes when creating new symbols. There was some code in the where handler that did this. This code has now been removed as the logic should be associated with code blocks and therefore work in all situations, not just for where statements.

sergisiso:


src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -105,6 +109,33 @@ def __init__(self, fp2_nodes, structure, parent=None, annotations=None):
...
+
+    def _insert_representative_references(self):
+        ''' Insert Reference children under this codeblock that
+        represent each of the symbols used inside the CodeBlock.
+        '''
+        for symbol_name in self.get_symbol_names():
+            try:
+                symtab = self.scope.symbol_table
+            except SymbolError:
+                symtab = SymbolTable()
```
arporter, line 139: Why is this branch necessary? Please add a comment.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -249,9 +294,11 @@ def reference_accesses(self) -> VariablesAccessMap:
 
         '''
         var_accesses = VariablesAccessMap()
-        for name in self.get_symbol_names():
-            var_accesses.add_access(Signature(name), AccessType.READWRITE,
-                                    self)
+        for child in self.children:
```
arporter: Comment please: all symbols accessed within the CodeBlock are captured as Reference nodes and stored as children of the CodeBlock node (or something).

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -105,6 +109,33 @@ def __init__(self, fp2_nodes, structure, parent=None, annotations=None):
             self.ast_end = None
         # Store the structure of the code block.
         self._structure = structure
+        self._insert_representative_references()
```
arporter, line 117: Comment pls.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -47,6 +47,10 @@
 from psyclone.core import AccessType, Signature, VariablesAccessMap
 from psyclone.psyir.nodes.statement import Statement
 from psyclone.psyir.nodes.datanode import DataNode
+from psyclone.psyir.nodes.reference import Reference
+from psyclone.psyir.nodes.node import Node
+from psyclone.psyir.symbols import (
+    SymbolTable, SymbolError, UnresolvedInterface)
 
 
 class CodeBlock(Statement, DataNode):
```
arporter, line 57: Please update the class docstring to record that it does now have children.

arporter: (I've checked the description of CodeBlocks in the documentation and it doesn't require any updating.)

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -186,12 +217,26 @@ def get_symbol_names(self) -> List[str]:
...
+                else:
+                    check = node
+                if isinstance(check.parent, Fortran2003.Data_Ref):
+                    # The first child is the base reference, the others are
+                    # accessor names, which are not symbols
+                    if check.parent.children[0] is not check:
+                        continue
                 result.append(node.string)
         # Precision on literals requires special attention since they are just
         # stored in the tree as str (fparser/#456).
```
arporter, line 350: Not yours but L263 is not covered by the associated test file. Perhaps one for @LonelyCat124 to fix while he's looking at directives?

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -1164,7 +1164,7 @@ def validate(
                         continue
                     exprn = prev.ancestor(Statement, include_self=True)
                     stmt = exprn.debug_string().strip()
-                    if isinstance(prev, (CodeBlock, Call, Kern, Loop)):
+                    if isinstance(prev, (Kern, Loop)):
```
arporter, line 1160: Just to check my understanding, both `Call` and `CodeBlock` are covered by the `isinstance(prev, Reference)` below?

src/psyclone/psyir/transformations/scalarisation_trans.py
```diff
@@ -307,13 +301,6 @@ def _value_unused_after_loop(sig: Signature,
             if has_complex_index:
                 return False
 
-            # If next access is a Call or CodeBlock or Kern then
-            # we have to assume the value is used. These nodes don't
-            # have the is_read property that Reference has, so we need
-            # to be explicit.
-            if isinstance(next_access, (CodeBlock, Call, Kern)):
```
arporter, line 316: Presumably this check for Call or Kern was already redundant before this PR?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1069,11 +1071,11 @@ def filecontainer_node(self, node):
         for symbol in node.symbol_table.symbols:
             # TODO #2201 - ContainerSymbols should be accepted but
```
arporter: With this change, ContainerSymbols (and bare Symbols) are accepted. Is this OK? Either way, the comment needs updating.

src/psyclone/tests/psyir/backend/fortran_test.py
```diff
@@ -950,18 +950,18 @@ def test_fw_filecontainer_2(fortran_writer):
 def test_fw_filecontainer_error1(fortran_writer):
     '''Check that an instance of the FortranWriter class raises the
     expected exception if the symbol table associated with a
-    FileContainer node contains any symbols.
+    FileContainer node contains any data symbols.
```
arporter, line 962: I'm a bit concerned we could now let Symbols through here.

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
```diff
@@ -583,8 +583,7 @@ def test_module_contains_subroutine_contains_subroutine(
     end function func_a
     end module my_mod"""
     psyir = fortran_reader.psyir_from_source(code)
-    # s symbol should not be in my_mod
-    with pytest.raises(KeyError):
-        psyir.children[0].symbol_table.lookup("s")
+    # FIXME: explain
```
arporter: Yes please :-)

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -128,24 +131,24 @@ def test_where_unknown_selector_type():
 
     '''
     fake_parent, fparser2spec = process_where(
-        "WHERE (ptsu(myfunc(), :, :) /= 0._wp)\n"
+        "WHERE (ptsu(unres, :, :) /= 0._wp)\n"
```
arporter: What was the reason for this change?

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -120,18 +121,30 @@ def test_codeblock_get_symbol_names(parser):
...
-    assert "b" in sym_names
-    assert "c" in sym_names
-    assert "mytest" in sym_names
-    assert "subroutine" not in sym_names
-    assert "sqrt" not in sym_names
-    assert "myloop" not in sym_names
-    assert "myifblock" not in sym_names
-    assert "this_is_true" in sym_names
-    assert "that_is_true" in sym_names
+    refs = block.walk(Reference)
```
arporter, line 227: Please could you also check that these refs are immediate children of the CodeBlock (as this is new).

Your use of 'virtual refs' to describe them did make me wonder whether we should have a new `VirtualReference` subclass of  `Reference` but I'm not sure that buys us anything other than making it clear that these aren't "real" references.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -676,37 +676,43 @@ def test_infer_sharing_attributes_with_explicitly_private_symbols(
...
+@pytest.mark.parametrize("code, loop_idx", [
+   (lambda x, y: f"write(*,*) {x}, {y}\n", 0),
+   (lambda x, y: f"do {x} = 1,2\nenddo\ndo {y} = 1,2\nenddo\n", 2)
+])
+def test_infer_sharing_attributes_with_hidden_references(
+        code, loop_idx, fortran_reader, fortran_writer):
     ''' Tests the infer_sharing_attributes() method when some of the loops have
-    Codeblocks inside it. We check that the infer_sharing attribute analysis
-    succeed by assuming worst case.
+    potentially hidden references (e.g. inside codeblocks or loop varialbes).
```
arporter: "variables"

src/psyclone/tests/psyir/nodes/omp_directives_test.py
```diff
@@ -676,37 +676,43 @@ def test_infer_sharing_attributes_with_explicitly_private_symbols(
...
-            write(*,*) scalar1, scalar2
+            {code("scalar1", "scalar2")}
         end subroutine''')
     omplooptrans = OMPLoopTrans()
     omplooptrans.omp_directive = "paralleldo"
-    loop = psyir.walk(Loop)[0]
+    loop = psyir.walk(Loop)[loop_idx]
     # Make sure that the write statements inside the loop are CodeBlocks,
     # otherwise we need a new test example
-    assert loop.has_descendant(nodes.CodeBlock)
```
arporter, line 736: We still need something like this check - probably has to be a `walk` now.

arporter (changes requested):
Nice, thanks Sergi.
The only concern I have is that, with this change, we now won't warn if a Symbol incorrectly gets added to the table of a Container. Can we tighten up on that?
I see you ran the integration tests so I won't do that again this time around.

src/psyclone/psyir/nodes/codeblock.py
sergisiso: Done

src/psyclone/psyir/nodes/codeblock.py
sergisiso: Added your words

src/psyclone/psyir/nodes/codeblock.py
sergisiso: Done

src/psyclone/psyir/nodes/codeblock.py
sergisiso: Done

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: With this PR 'previous_accesses' will not return CodeBlocks (it will return the references inside them with appropriate READWRITE).

Call has not been returned with 'previous_accesses' for quite some time now, it may be a leftover from them the past? Anyway all its symbols are children, the routine is a reference with CALL access and the arguments are references with the apporpriate READ/WRITE.

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Fixed

src/psyclone/psyir/transformations/scalarisation_trans.py
sergisiso: Yes

src/psyclone/psyir/backend/fortran.py
sergisiso: Yes, but the logic was confusing. I refactored it again to look precisely for the conditions that are acceptable and added comments explaining why.

src/psyclone/tests/psyir/backend/fortran_test.py
sergisiso: Reverted

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
sergisiso: Reverted

src/psyclone/tests/psyir/nodes/omp_directives_test.py
sergisiso: Re-introduced

src/psyclone/tests/psyir/nodes/codeblock_test.py
sergisiso: Done  the check.

(I am unconvinced about the VirtualReference, but I am exploring in a separate branch the concept of locked nodes, which would provide an early error if anybody tries to change/update any of these.)

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
sergisiso: Done. Its an important consideration, feel free to discuss.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1063,15 +1076,19 @@ def filecontainer_node(self, node):
...
         for symbol in node.symbol_table.symbols:
-            # TODO #2201 - ContainerSymbols should be accepted but
-            # currently are stored in its containing scope.
-            if isinstance(symbol, DataSymbol):
+            # Only RoutineSymbols and ContainerSymbol can be declared here
+            # pylint: disable=unidiomatic-typecheck
+            if type(symbol) is Symbol and symbol.is_unresolved:
+                # However we also accept symbols that we don't know where
+                # they are declared, so we propagated upwards.
+                continue
```
arporter, line 1116: Unfortunately this line isn't covered by the fortran_*test.py tests.

arporter: Since I'm here, I'm also not sure what "we propagated upwards" means as we're already handling a FileContainer at this point. Do you mean they will have been propagated upwards and have ended up in this table?

arporter (changes requested):
Very nearly there. There's just one line in the fortran backend that isn't covered by the associated tests. I'll fire off the ITs in the meantime.

sergisiso:
@arporter It's been a while from the last review, but this is now unblocked and ready to continue.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -54,13 +55,15 @@
 
 
 class CodeBlock(Statement, DataNode):
-    '''Node representing some generic Fortran code that PSyclone does not
-    attempt to manipulate. It has a child Reference for each symbol used in
-    the CodeBlock.
-
-    :param fp2_nodes: the fparser2 parse-tree nodes representing the
+    '''Node representing any generic Fortran code that PSyclone does not
+    attempt to manipulate. As such it is a leaf in the PSyIR. A CodeBlock
```
arporter: Strictly speaking, it is no longer a leaf as it has child References.

src/psyclone/psyir/transformations/scalarisation_trans.py
```diff
@@ -79,11 +79,10 @@ class ScalarisationTrans(LoopTrans):
       integer :: j
       real, dimension(100) :: a
       real, dimension(100) :: b
-      real :: a_scalar
     <BLANKLINE>
       do i = 1, 100, 1
-        a_scalar = i
-        b(i) = a_scalar * a_scalar
+        a(i) = i
```
arporter: This doesn't look right - `a(i)` should have been replaced by `a_scalar`?

src/psyclone/tests/psyir/transformations/scalarisation_trans_test.py
```diff
@@ -78,13 +78,6 @@ def test_scalararizationtrans_is_local_array(fortran_reader):
     assert ScalarisationTrans._is_local_array(Signature("local"),
                                               var_accesses)
 
-    # Test b - the RHS of the assignment is a codeblock so we do not
-    # count it as a local array and invalidate it, as otherwise the
-    # local array test can fail. Also we can't safely transform the
-    # CodeBlock anyway.
-    assert not ScalarisationTrans._is_local_array(Signature("b"),
```
arporter, line 85: Presumably we do now know that `b` *is* a local array (because we now support the `len=2` specifier)?

arporter (changes requested):
Thanks Sergi, good to get this done. I'm a bit puzzled by the change to the scalarisation transformation docstring. Hopefully the new doctests will back me up.
Apart from that, there's just a tiny bit of tidying.
I've fired off the ITs.

src/psyclone/psyir/nodes/codeblock.py
sergisiso: True, I removed the sentence

src/psyclone/psyir/transformations/scalarisation_trans.py
sergisiso: Ups, thats embarrassing, I think I was blanked accepting docsting changes assuming they were coming from my doctest fixes but this is clearly wrong.

sergisiso: Fixed now

src/psyclone/tests/psyir/transformations/scalarisation_trans_test.py
sergisiso: Yes, is a combination of both: the dependency is not to the codeblock anymore, but to the specific reference inside the codeblock. So we can do a ref.symbol to know the declaration, and in the decaration we support everything to know that it is local. So the _is_local_array of the Signature(b) can be resolved, even if we still have the Codeblock.

sergisiso:

src/psyclone/tests/psyir/transformations/scalarisation_trans_test.py
arporter: Nice.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -926,10 +930,13 @@ def _compute_backward_uses(self, basic_block_list: list[Node]):
                     sig = reference.get_signature_and_indices()[0]
                     # Work out if its read only or not.
                     assign = reference.ancestor(Assignment)
-                    # RHS reads occur "before" LHS writes, so if we
-                    # hit the LHS or an assignment then we won't have
-                    # a dependency to the value used from the LHS.
-                    if assign is not None:
+                    if reference.ancestor(Call):
```
arporter: Does this need to be `(Call, CodeBlock)` rather than just `Call` - the comment below mentions CodeBlocks? The original code (at L980) had a `if defs_out[sig] is not None` - do we need the same protection here?

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -472,8 +472,11 @@ def test_definition_use_chain_find_backward_accesses_codeblock(
     sig = ref.get_signature_and_indices()[0]
     chains = DefinitionUseChain(ref)
     reaches = chains.find_backward_accesses()[sig]
-    assert len(reaches) == 1
+    # We don't know if the CodeBlock is definitely a write (in this case it
+    # is not), to the backwards could also be the previous write
```
arporter: s/to/so/ ?

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -547,7 +550,7 @@ def test_definition_use_chain_find_backward_accesses_codeblock_and_call_local(
     sig = ref.get_signature_and_indices()[0]
     chains = DefinitionUseChain(ref)
     reaches = chains.find_backward_accesses()[sig]
-    assert len(reaches) == 1
+    assert len(reaches) == 2
     assert reaches[0] is routine.walk(CodeBlock)[0].children[0]
```
arporter, line 554: Pls also add a check that the second `reaches` is as expected.

arporter (commented):
Thanks Sergi. I wasn't expecting to see changes to the DUC - is that because of other things that have gone onto master? Some of the changes don't seem entirely consistent - please see inline. I'll hold off the ITs for now.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
sergisiso: Fixed

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
sergisiso: Added

src/psyclone/psyir/tools/definition_use_chains.py
sergisiso: I could do with @LonelyCat124 assistance here as I don't fully understand the DUC, but there are 2 aspects in your question:

> Does this need to be (Call, CodeBlock) rather than just Call

Codeblocks now have references with the read-write attribute, so they don't need an special path if they are local symbols (other interfaces are handled before this point). I updated the comment.

I don't know why the Call has this special path as arguments also have read/write/rw depending on the intents or rw if unknown. I think we are currently more strict than intents for Calls but @LonelyCat124 was considering changing this. But modifying the Call behaviour is out of scope for this PR.

> The original code (at L980) had a if defs_out[sig] is not None - do we need the same protection here?

Tbh is because I couldn't tigger the condition, but I may be wrong. For reference this line use to be:
```
if defs_out[sig] is not None:
    self._killed[sig].append(defs_out[sig])
defs_out[sig] = reference
```
and I removed the condition part. These is reached for codes like this (starting on the assingment and going backwards):
```
        print *, a
        call b(a)
        a = a + 2
```
But I don't understand what "killed" is when going backwards. Is it to mark before is alive? aka. before a guaranteed write? In this case the Call RW is by convention, not necessarely guaranteed. So I am unsure if this should be killed.
But as I say, I may be misunderstnding that these lines did.

sergisiso:
@arporter I mark this as ready for review again, but would be good to have @LonelyCat124 view on the remaining point before progressing.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -171,13 +171,21 @@ def __init__(
...
+        # These are maps of signatures to node for the region of code provided
+        # to the DUC, for every singature:
+        # - The uses collects every possible read reached from the start of the
+        # region (until we find any guaranteed write).
         self._uses = {}
+        # - The defout collects the nodes that could represent end points of
+        # the signature use chain (all writes before the guaranteed write).
         self._defsout = {}
+        # - The killed is a working collection used by compute_x_uses to store
+        # writes found that are defenitely are not the final write.
```
arporter: "...definitely not..."

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -171,13 +171,21 @@ def __init__(
...
         self._killed = {}
-
-        # The output map, mapping between nodes and the reach of that node.
+        # - The reaches collects all nodes that can be reach for each signature
```
arporter: "reached"

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -930,8 +938,10 @@ def _compute_backward_uses(self, basic_block_list: list[Node]):
                     sig = reference.get_signature_and_indices()[0]
                     # Work out if its read only or not.
                     assign = reference.ancestor(Assignment)
-                    if reference.ancestor(Call):
-                        # For calls and Codeblocks we assume read/write access
+                    if reference.ancestor((Call, CodeBlock)):
+                        # For calls we always assume read/write access
```
arporter: "...Calls and CodeBlocks we..."

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -473,10 +473,9 @@ def test_definition_use_chain_find_backward_accesses_codeblock(
     chains = DefinitionUseChain(ref)
     reaches = chains.find_backward_accesses()[sig]
     # We don't know if the CodeBlock is definitely a write (in this case it
```
arporter: Please update this comment - we have to assume the worst for a CodeBlock.

arporter (changes requested):
Thanks Sergi, all good now bar some comments that need updating/correcting.
I've fired off the ITs in the meantime.

src/psyclone/psyir/tools/definition_use_chains.py
sergisiso: Done

src/psyclone/psyir/tools/definition_use_chains.py
sergisiso: Done

src/psyclone/psyir/tools/definition_use_chains.py
sergisiso: Done

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
sergisiso: Done

sergisiso:
@arporter Ready for another look

arporter (approved):
Thanks Sergi. All changes look good now and I see the ITs were green when I ran them.

PR #3249, opened 2025-12-03
(Closes #3246) ArrayBounds.copy

### LINKED ISSUE #3246: ArrayBounds .copy()
In #3119 we have discussed adding a .copy() to the ArrayBounds dataclass to avoid reconstructing repeats.

LonelyCat124:


src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -652,6 +654,16 @@ def _dangling_parent(
             object.__setattr__(self, 'lower', _dangling_parent(self.lower))
             object.__setattr__(self, 'upper', _dangling_parent(self.upper))
 
+        def copy(self) -> ArrayType.ArrayBounds:
+            '''Creates a copy of this ArrayBounds object'''
+            # pylint: disable-next=import-outside-toplevel
+            from psyclone.psyir.nodes import DataNode
+            lower_copy = (self.lower.copy() if isinstance(
+                self.lower, DataNode) else self.lower)
```
sergisiso: self.lower must be a DataNode, so maybe the else is not needed?

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -652,6 +654,16 @@ def _dangling_parent(
...
+            upper_copy = (self.upper.copy() if isinstance(
+                self.upper, DataNode) else self.upper)
```
sergisiso: I see that Extent (which is what this is if its not a DataNode) has a copy defined. Maybe we don't need an else here either?

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Yeah I think so, remove the if for both lower and upper as I think copy should always be defined for them, so its back to being a one-liner.

PR #3254, opened 2025-12-09
(closes #3253) Fix openmp parallel lowering with kernels and improve dependency messages

### LINKED ISSUE #3253: GOcean omp partially broken
Openmp for gocean seems to be broken. Taking the `1623_add_training` branch, and running `psyclone/tutorial/training/gocean/2.8-GameOfLife-openmp/solution` using:
```
psyclone --config $PSYCLONE_ROOT/config/psyclone.cfg --psykal-dsl gocean -l output -nodm -d $PSYCLONE_ROOT/tutorial/training/gocean/gol-lib -s ./openmp_combined.py -oalg time_step_alg_mod.f90 -opsy time_step_alg_mod_psy.f90 time_step_alg_mod.x90
```
results in:
```
    !$omp parallel default(shared) private(born,die,i,j,neighbours) firstprivate(current)
    !$omp do schedule(static)
    do j = neighbours%internal%ystart, neighbours%internal%ystop, 1
      do i = neighbours%internal%xstart, neighbours%internal%xstop, 1
        call count_neighbours_code(i, j, neighbours%data, current%data)
      enddo
    enddo
    !$omp end do
    !$omp do schedule(static)
    do j = born%internal%ystart, born%internal%ystop, 1
      do i = born%internal%xstart, born%internal%xstop, 1
        call compute_born_code(i, j, born%data, current%data, neighbours%data)
      enddo
...
So the fields are all private :cry: It appears that gocean actually usually passed the tests?? Will need to check this cleanly, I would have expected that at least the CI in my training would have picked that up (but I am currently slightly updating the training to ensure all the tests run as expected, esp with compilation and running).

Imho, a gocean field should never become private (since it would be an uninitialised structure pointing to an uninitialised 2d array), and not even sure what openmp would expect (should it be `current%data` etc?? Sorry, I can't quote the OpenMP standard off the top of my head :) )

### PR DESCRIPTION
**Pull request by sergisiso**

### CODE REVIEW

File: src/psyclone/psyir/tools/dependency_tools.py [hunk 1]
```diff
@@ -738,37 +738,33 @@ def _array_access_parallelisable(self, loop_variables, var_info):
                 if not self._is_loop_carried_dependency(loop_variables,
                                                         write_access,
                                                         other_access):
+                    # We can store the loop variable in the lambdas because
+                    # we immidiately return after them.
```
hiker: I have to admit that this is kind of beyond me, but I assume you know better than me :)
So, even if in the caller there would be an unrelated variable `write_access`, it would not be used?

If that is indeed the case, that's fine. I would just recommend to change the comment: `immediately return after them` (them=lambdas?) might not be clear. Something like: `immediately return after the definition of the lambdas, so any local variable of the same name will not be used`??

hiker (changes requested):
Thanks a lot for fixing this, it was great that I could mention this on the next day in my training.
One thing I would really like to see added: a test case to make sure that bug is not re-introduced in the future. Since testing for 'expression not there' is fragile, maybe just a small example to check that only `i, j` are declared private? So that it would fail on current master, and succeeds with your patch.

Otherwise maybe just a comment update for the lambdas (and if you think it's obvious enough, I am happy to say that my understanding of the implementation of lambdas and when variables are actually used is not good enough :) )

src/psyclone/psyir/tools/dependency_tools.py
hiker: And I forgot to mention: thanks a lot for improving the error messages, they are indeed a lot more useful now.

sergisiso: Your suggested comment is what I meant and I updated to something similar (but just to clarify your "even if in the caller there would be an unrelated variable", the lambdas have no input parameters, the caller can not provide anything, is that if we capture a local variable this shouldn't be changed after the fact because these are lazily evaluated)

hiker: OK, I decided to dig into this, since as far as I can see this is wrong. Simple python example:
```
class C:
	@property
	def f(self):
		print("Evaluating c.f")
		return "f"

c = C()
x = lambda : f"lambda using {c.f}."
x()

class D:
	@property
	def f(self):
		print("Evaluating d.f")
		return "f"

print("Setting c to be D()")
c = D()
x()
```
Output:
```
Evaluating c.f
Setting c to be D()
Evaluating d.f
```
So, the first evaluation of lambda uses the `c` which is an instance of `C`. When `c` is changed to be an instance of `D`, and the lambda is evaluated again, as expected it calls `D`'s `f` property.

BUT: I could not reproduce this problem in the dependency tool, which really really confused me. Some debugging later: it works because the `Message` class in `_add_message` does:
```
        self._messages.append(Message(f"{message_type}: {message}", code,
                                      var_names, var_infos))
```
Basically, the lazy string is evaluated when adding the message to the Message list in DependencyTools. Or: the LazyString evaluation does not work as expected in the DependencyTools.

As far as I can see, this will require to:
1. Remove the string evaluation when adding a message
2. Converting the lambdas to use default values, e.g. in the example above:
```
y = lambda c=c: f"lambda using {c.f}."
```
If this is used, it will as expected print `Evaluating c.f` in both cases, even if the local `c` is modified.

This is out of scope for this PR, but could you open a ticket for this?

hiker (changes requested):
Thanks a lot for the updated. It let me on a long chase to fix my understanding of lambda, which actually ended up with detecting a but in LazyString handling in `DependencyTools`, see separate comment.
Unfortunately, none of your extended tests seems to test this particular error. When I copied the examples to master, all tests still passed on master. So, we still need a better test for this, sorry.

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: Sorry but I don't understand what you mean, doesn't your example show that the dependency_tools comment (and subsequent implementation) is correct? We should (and do) return after capturing a variable in a lambda.

sergisiso:
@hiker back to you. I am afraid I don't understand where you think the bug is, could you point me to some steps to trigger it.

sergisiso:
I added a small commit to let Message propagate, instead of evaluate the LazyString, to show that this still doesn't change the results.

src/psyclone/psyir/tools/dependency_tools.py
hiker: OK, I went back to it. And it turns out you are right, my example wasn't 100% what happens in the dependency tools. I assume the lambda keeps a copy of `locals()`, and as such can access the local variables even if they are out of scope. In my example, I modified the local variable (i.e. I modified `locals()`), which is why it showed the wrong value). Once I created the lambda in a function (with a local `c`), it all worked.

So, please ignore this part. I have opened #3261 , since there is still the fact that the dependency tools will evaluate the LazyString (even if the user never converts a message to a string).

sergisiso:
It was quite involved to make the error happen, but now one of the tests fails in master, it has to:
- do the lowering before, so the reference in the call doesn't have indices and follow the "scalar logic" (this is the step actually fixed in this PR)
- Inline it and have intents other than READWRITE (otherwise it says the READ could be first so it must be shared)
- Have the field used by 2 kernels (otherwise we have a check that if it only has one access is an error or shared)

src/psyclone/psyir/tools/dependency_tools.py
sergisiso: Wasn't the last bit sorted by https://github.com/stfc/PSyclone/pull/3254/changes/2f10a29c953ded8972a62c48aec91f4f2e29479a ?

PR #3255, opened 2025-12-09
(Closes #3251) Implementation for the PSyDirective.

### LINKED ISSUE #3251: PSyclone specific directives to improve script portability
While the whole idea of adding PSyclone-specific directives runs again the idea of PSyclone (after all, what would be the difference to use say omp directives), there is one long-term problem that we need to address, and a **very limited** support for PSyclone-specific directives would imho be useful.

In discussion with NG-ARCH partner (talking about generic Fortran transformation scripts) it was raised that many scripts are very fragile. E.g. they count loops to identify the loop that needs a special handling (like adding a compiler directive to enable vectorisation).

This is very fragile, since modifications of the source code (later) can mean that incorrect transformations are done (following up on the example above, a vectorisation directive could be added to a loop that is now not safe to vectorise anymore, and/or performance is lost since the loop that needs it won't get the directive added anymore).

It adds burden to the script programmer (e.g. they could add tests to make sure it is the 'right' loop, e.g. checking the variables used??), and additional burden for further development (when a test like the one mentioned before is triggered, the script needs to be updated).

Being able to name a statement (esp. a loop) would significantly help with portability. Instead of (say) getting the 5th loop, user scripts could ask for `psy_ir.get_statement("stochastic_physics")` or so.

Simple Fortran-comments markup would be sufficient for this, e.g.:
```
       !PSY time-stepping-loop
        do time =1, time_steps

```

LonelyCat124:
Example and doc TODO

Basic implementation of PSyDirective including reading and writing.

At the moment I implemented it as a Standalone Directive (as it doesn't make sense to me to be a RegionDirective) but that could be more difficult to use in some ways, but I'll put together an example and see.


sergisiso:
I still don't think we need psyclone (`!$psy`) specific directives, I thought the plan was to add a generic `Directive` node, so that we could do `psyir.walk(Directive)` as an slight improvement to the current `psyir.walk(Codeblock)` (which tbh is almost the same as this PR implements, just with a slighly different `_directive_handler`)

LonelyCat124:
We could do that instead, but it would mean more false positives when people are looking for something specific to their script, e.g. after `psyir.walk(???Directive)` (name is not straightforward as we can't make it `Directive` since thats a necessary abstract class, along with `StandaloneDirective` and `RegionDirective` nodes. The other pain point is taking things out of CodeBlocks is it could be misinterpreted as support for them, and having non-RegionDirective nodes having to be `StandaloneDirective` subclasses.

I would maybe with ok with doing some `GenericDirective` and excluding all directives that start `omp`, `ompx`, `omx`, `acc` if you think that would be better?

sergisiso:
> name is not straightforward as we can't make it Directive since thats a necessary abstract class, ...
> The other pain point is taking things out of CodeBlocks is it could be misinterpreted as support for them
> I would maybe with ok with doing some GenericDirective

Good points, what about [Unrecognised/Unsupported/Unknown]Directive (being sibling of Standalone/RegionDirective)?

So it is clear that we don't support them and if we ever try to parse OpenMP, OpenACC into their own Directive the name will still make sense for those that we have not done yet. (my preferd is UnknownDirective, but I am unsure)

> but it would mean more false positives when people are looking for something specific to their script

I don't understand this. It will be better than the current Codeblock, and it would allow searching for "ivdep" or other know compiler directive that may exist in the code.

> excluding all directives that start omp, ompx, omx, acc

Why excluding those?

LonelyCat124:
@sergisiso this is ready for review, but no rush to do so.

src/psyclone/tests/psyir/frontend/fparser2_directive_test.py
```diff
@@ -191,8 +191,8 @@ def test_directive_before_while():
 
 
 def test_directive_before_allocate():
-    """Test that the FortranReader stored a directive before an allocate as a
-    CodeBlock."""
+    """Test that the FortranReader stored a directive before an allocate as an
+    UnknownDirective."""
```
sergisiso: Revert (this one is still a Codeblock)

src/psyclone/psyir/nodes/unknown_directive.py
```diff
@@ -0,0 +1,78 @@
...
+# -----------------------------------------------------------------------------
+# Authors A. B. G. Chalk, STFC Daresbury Lab.
+# -----------------------------------------------------------------------------
+
+''' This module contains the UnknownDirective node implementation.'''
+
+from psyclone.psyir.nodes.directive import StandaloneDirective
+
+
+class UnknownDirective(StandaloneDirective):
```
sergisiso, line 42: Maybe StandaloneDirective is the right choice, but why not just Directive?

src/psyclone/psyir/nodes/unknown_directive.py
```diff
@@ -0,0 +1,78 @@
...
+    '''
+    Directive representing PSyclone-specific directives in the tree.
+
+    :param directive_string: The content after the !$psy part of this
+                             node in the tree.
```
sergisiso, line 47: param kwargs is missing

src/psyclone/psyir/nodes/unknown_directive.py
```diff
@@ -0,0 +1,78 @@
...
+    '''
+    _children_valid_format = ""
```
sergisiso: we typically write `"<LeafNode>"` as format for leafs

examples/psyir/unknown_directives/README.md
```diff
@@ -0,0 +1,51 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+Authors: A. B. G. Chalk, STFC Daresbury Lab
+-->
+
+# PSyclone PSyIR directives.
```
sergisiso: Maybe:

```
# Use custom directives

This directory contains an example of how write PSyclone scripts to transform code
by taking advantage of information provided by the use of custom Fortran code directives.

```

I would also call the directory "custom_directives" (although the node name is Unknown, the use case is probably what people will look for in the examples)

examples/psyir/unknown_directives/README.md
```diff
@@ -0,0 +1,51 @@
...
+This directory contains examples of how to create and/or modify
+instances of PSyIR and how to use directives in the tree to influence
+PSyclone's behaviour.
+
+The Fortran code contains a custom directive, ``!$psy no_par`` which is
+used by the PSyclone script to skip attempting to parallelise a specific loop.
+
+The example can be compiled using the Makefile, but this is a simple operation:
+``psyclone --keep-directive -s omp.py directive_filtering.F90``. The
+ ``-keep-directives`` option is essential to enabling the directive of interest
```
sergisiso: double dash?

examples/psyir/unknown_directives/omp.py
```diff
@@ -0,0 +1,56 @@
...
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: A. B. G. Chalk, STFC Daresbury Lab
+
+from psyclone.psyir.transformations import OMPLoopTrans
+from psyclone.psyir.nodes import Loop, Routine, UnknownDirective
+
+
+# Find the PSyDirectives, check what the directive is for and then find the
```
sergisiso: Maybe "Find the custom directive and add the associated loop to the list of loops to skip. Parallelise the rest."

And better as the method docstring instead of a comment, to make pylint happy.

examples/psyir/unknown_directives/Makefile
```diff
@@ -0,0 +1,53 @@
...
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+# Modified: A. B. G. Chalk, STFC Daresbury Lab
```
sergisiso: Only you as author

examples/psyir/unknown_directives/directive_filtering.F90
```diff
@@ -0,0 +1,58 @@
...
+    character, dimension(13), parameter :: j = "some_test_par"
+    integer :: i
+    integer, dimension(1000) :: vals
+
+    do i = 1, 1000
+        vals(i) = i
+    end do
+
+    ! Don't parallelise this next loop due to the string comparison
+    !$psy no_par
```
sergisiso: I would prefer a different string than "psy" for the example, maybe "mydirective" or "custom_directive" to not confuse anybody just looking at this file that this is a psyclone directive (change the readme with the same).

examples/psyir/custom_directives/Makefile
```diff
@@ -0,0 +1,53 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# -----------------------------------------------------------------------------
+# Author: M. Naylor, University of Cambridge, UK
+# Modified: A. B. G. Chalk, STFC Daresbury Lab
+
```
sergisiso, line 35: We typically add "transform, compile and run" standardised target names to all examples, and you will need to make ../Makefile recurse to it (at least for the transform target) so that this is tested in the CI.

Also, it will be good to mention the example in the parent folder README

src/psyclone/tests/psyir/frontend/fparser2_directive_test.py
LonelyCat124: Fixed.

src/psyclone/psyir/nodes/unknown_directive.py
LonelyCat124: I guess it doesn't really matter here, just more conceptually i see `StandaloneDirective` or `RegionDirective` as the real base classes for the actual PSyIR and `Directive` is just a class that contains their shared functionality.

src/psyclone/psyir/nodes/unknown_directive.py
LonelyCat124: I think thats just up to examining parent classes docs - I don't think I want to document kwargs here right? I think they just result in the base `Node` class keyword arguments here so are simply provided for internal use wherever the node is constructed if required.

src/psyclone/psyir/nodes/unknown_directive.py
LonelyCat124: Added

examples/psyir/unknown_directives/README.md
LonelyCat124: Changed, and yeah custom directives is a better name for the example.

examples/psyir/unknown_directives/README.md
LonelyCat124: Fixed

examples/psyir/unknown_directives/omp.py
LonelyCat124: Done.

examples/psyir/unknown_directives/Makefile
LonelyCat124: Changed.

examples/psyir/unknown_directives/directive_filtering.F90
LonelyCat124: Yeah, changed it to `!$my_dir`.

examples/psyir/custom_directives/Makefile
LonelyCat124: Done, only this example works from the example testing (as before none in this directory did).

src/psyclone/psyir/nodes/unknown_directive.py
sergisiso: I am unsure about this, but I am happy to leave it like this and revisit later if necessary.

src/psyclone/psyir/nodes/unknown_directive.py
sergisiso: We typically have `:param kwargs: additional keyword arguments provided to the PSyIR node.`, I will add it myself if there is nothing else.

examples/psyir/custom_directives/omp.py
```diff
@@ -0,0 +1,56 @@
...
+            if psy_dir.directive_string == "my_dir no_par":
+                position = psy_dir.position
+                parent = psy_dir.parent
+                # Remove the directive as we don't need it in the output now.
+                psy_dir.detach()
+                loops_to_skip.append(parent.children[position])
+    for loop in routine.walk(Loop):
+        if loop in loops_to_skip:
+            continue
+        OMPLoopTrans(omp_directive="paralleldo").apply(loop)
```
sergisiso: This should be indented in (otherwise it just parallelises the last routine)

examples/psyir/custom_directives/identify_custom_directives.py
```diff
@@ -0,0 +1,56 @@
+# -----------------------------------------------------------------------------
```
sergisiso, line 1: Also change the name of this file to something more descriptive like: `identify_custom_directives.py`

examples/psyir/create_and_modify_psyir/Makefile
```diff
@@ -0,0 +1,50 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+# ------------------------------------------------------------------------------
+# Authors: A. R. Porter and S. Siso, STFC Daresbury Laboratory
+
+include ../common.mk
```
sergisiso: This is broken now (so `make transform` in this folder does not work)

sergisiso: Also, two of the examples in this folder's README should not be here, and instead should be in a parent readme

examples/psyir/Makefile
```diff
@@ -31,20 +31,8 @@
...
-	$(CONFIG_ENV) ${PYTHON} create.py
-	$(CONFIG_ENV) ${PYTHON} create_structure_types.py
-	$(CONFIG_ENV) ${PYTHON} modify.py
-
-compile:
-	@echo "No compilation supported for the PSyIR examples"
-
-run: compile
-	@echo "No run targets for PSyIR creation examples"
+EXAMPLES=custom_directives
```
sergisiso: You should also list the new `create_and_modify_psyir` as this was tested before (it was what this Makefile was doing). Also test the `matmul` and `transpose` directories (they both work fine for me)

examples/psyir/custom_directives/omp.py
LonelyCat124: Fixed.

examples/psyir/custom_directives/identify_custom_directives.py
LonelyCat124: Done

examples/psyir/create_and_modify_psyir/Makefile
LonelyCat124: Fixed, and moved the README sections into their own examples readme's (not in the parent).

examples/psyir/Makefile
LonelyCat124: Fixed those and added them too.

LonelyCat124:
@sergisiso I think I added everything and the gitignore should ignore the .F90 files generated in those directories now.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5801,16 +5801,30 @@ def process_comment(self, comment, preceding_comments):
...
         :param parent: The parent to add the PSyIR node to.
 
-        :returns: a CodeBlock containing the input Directive.
+        :returns: a CodeBlock containing the input Directive or a
+                  UnknownDirective.
         '''
+        # We don't turn OpenMP, OpenACC or directives we can't output
+        # correctly into Directive nodes. PSyclone currently always
+        # outputs directives starting with !$
+        dont_match = ["!$omp", "!$acc", "!$ompx", "!dir$"]
```
sergisiso: @LonelyCat124 I was a bit on the fence on these and wanted to wait for a use case. With what we discussed today: add profiling hooks around openmp in existing directives maybe it sways my opinion that these should be UnknownDirective so they are easier to identify. What do you think?

The !dir$ (and other !xxx$ ?) should still be CodeBlocks.

LonelyCat124: My main resoning for this is preventing them occurring in parallel loops. I could also add `UnknownDirective` into `excluded_node_types` in ParallelLoopTrans if you'd prefer, though this would block compiler directives appearing inside Parallel loops.

sergisiso: >  I could also add UnknownDirective into excluded_node_types in ParallelLoopTrans if you'd prefer, though this would block compiler directives appearing inside Parallel loops.

I am ok with this.

LonelyCat124: I'm still disallowing `ompx` extensions since we don't know at the moment what these might be and thus they're not useful to keep I think.

LonelyCat124: @sergisiso This is done now.

Issue #3250, opened 2025-12-03
Complete Codecov coverage


sergisiso:
We are at 99.91% coverage, but looking at the misses, I doubt some of them were introduced by PRs that had no coverage (because the Codecov CI complains about these). I think the most likely explanation is that they were 'Indirect changes' in PRs that didn't decrease the total percentage (because the Codecov CI does not complain about these) - I admit I often fail to manually check that on my reviews.

It may be worth adding "pragma no-cover" (if a fix is not obvious) to the remaining cases so that Codecov would complain for any indirect change, as it would make it below 100%.

PR #3258, opened 2025-12-11
(closes #279) Remove Arguments.dofs and generally improve code coverage

### LINKED ISSUE #279: Refactor psyGen.Invoke to remove need for Arguments.dofs property.
All of the APIs have a `dofs` property in their Arguments sub-class with a comment to the effect that it's only there because the constructor of the Invoke base class expects it to be. I'm not sure if any API actually uses that constructor.

sergisiso:


src/psyclone/domain/lfric/lfric_kern.py
```diff
@@ -976,7 +976,7 @@ def _validate_kernel_code_arg(self, kern_code_arg, interface_arg,
             # This is used to identify the correct kernel subroutine for a
             # mixed-precision kernel.
             alg_precision = api_config.precision_map[alg_arg.precision]
-            if alg_precision != actual_precision:
+            if alg_precision != actual_precision:  # pragma: no-cover
```
arporter: I think adding the following test to `lfric_kern_test.py` will cover this line. I'll let you write a suitable docstring :-)
```python
def test_validate_kernel_code_arg_against_alg_arg(monkeypatch):
    '''
    '''
    psy, invoke = get_invoke("26.8_mixed_precision_args.f90", TEST_API,
                             name="invoke_0", dist_mem=False)
    sched = invoke.schedule
    # Second kernel has args of precision r_solver.
    kernel = sched.walk(LFRicKern, stop_type=LFRicKern)[1]
    read_access = ArgumentInterface(ArgumentInterface.Access.READ)
    lfric_real_scalar_symbol = LFRicTypes("LFRicRealScalarDataSymbol")(
        "scalar", interface=read_access)
    with pytest.raises(GenerationError) as err:
        kernel._validate_kernel_code_arg(lfric_real_scalar_symbol,
                                         lfric_real_scalar_symbol,
                                         kernel.args[0])
    assert ("Precision (4 bytes) of algorithm-layer argument "
            "'scalar_r_solver' does not match that (8 bytes) of the "
            "corresponding kernel subroutine argument 'scalar' for kernel "
            "'mixed_code'" in str(err.value))
```

src/psyclone/gocean1p0.py
```diff
@@ -1207,13 +1206,6 @@ def find_grid_access(self):
         # if the kernel requires a grid-property argument.
         return None
 
-    @property
```
arporter, line 1210: I see you've also update lfric.py (thanks) so I've added #279 to the issues this PR will close (as I've grepped and those are the only mentions).

src/psyclone/psyGen.py
```diff
@@ -850,7 +833,7 @@ def args(self):
         base method and simply return our argument. '''
         return [self._field]
 
-    def check_vector_halos_differ(self, node):
+    def check_vector_halos_differ(self, node):  # pragma: no-cover
```
arporter: psyGen_test does have some tests of this method so it can't all be uncovered? In fact, I've grepped and it seems this method is completely unused?

src/psyclone/psyir/nodes/node.py
```diff
@@ -57,6 +57,7 @@
 # We use the termcolor module (if available) to enable us to produce
 # coloured, textual representations of Invoke schedules. If it's not
 # available then we don't use colour.
+# no cover: start
 try:
     # pylint disable=import-outside-toplevel
     from termcolor import colored
```
arporter, line 62: Can't we monkeypatch the packages that appear to be available from `termcolor`? I had a google and found https://stackoverflow.com/questions/8658043/how-to-mock-an-import

arporter: (I think it's quite important to get coverage for this because if we somehow broke it it would break a lot of stuff!)

src/psyclone/tests/gocean1p0_test.py
```diff
@@ -115,6 +115,29 @@ def test_field(tmpdir, dist_mem):
     assert GOceanBuild(tmpdir).code_compiles(psy)
 
 
+def test_invalid_field_accesses_for_iteration_space():
+    ''' Tests that we when retrieving the iteration_space_arg we validate
```
arporter: s/we when/when/

src/psyclone/tests/psyir/backend/fortran_unsupported_declns_test.py
```diff
@@ -118,7 +118,7 @@ def test_fw_unsupportedtype_routine_symbols_error(fortran_writer):
     class OtherType(UnsupportedType):
         ''' UnsupportedType is abstract so sub-class it for this test '''
         def __str__(self):
-            return "OtherType"
+            ...
```
arporter, line 121: I didn't know you could do that!

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
```diff
@@ -47,6 +47,20 @@
 from psyclone.psyir.symbols import RoutineSymbol
 
 
+def test_empty_program_handler(parser):
+    '''Test that an empty program handler still returns a FileContainer
```
arporter: "the program handler still returns a FileContainer for an empty program" perhaps?

src/psyclone/transformations.py
```diff
@@ -2115,35 +2114,12 @@ def apply(self, node: Schedule, options: Optional[Dict[str, Any]] = {}):
         # extract async. Default to False.
         async_queue = options.get('async_queue', False)
 
-        # check
```
arporter, line 2118: How come you've removed this? I see we still support the specification of an `async_queue`?

src/psyclone/transformations.py
```diff
@@ -2473,12 +2444,14 @@ def validate(self, node, options=None):
                 f"'{type(invoke_schedule).__name__}'")
 
         # Check that there are no unqualified imports or undeclared symbols
+        # no cover: start
         try:
             kernels = node.get_callees()
-        except SymbolError as err:
+        except SymbolError as err:  # pragma: no-cover
```
arporter: This should be easy to cover as we can `monkeypatch` `node.get_callees()` with a function that just raises `SymbolError`.

arporter (changes requested):
Thanks very much for tackling this Sergi.
There are a couple of cases where I think we can add tests rather than no-cover pragmas. Also, you've removed some validation code from ACCEnterDataTrans and I don't understand why.

src/psyclone/domain/lfric/lfric_kern.py
sergisiso: Done

src/psyclone/gocean1p0.py
sergisiso: Great!

src/psyclone/tests/gocean1p0_test.py
sergisiso: Done

src/psyclone/tests/psyir/frontend/fparser2_program_handler_test.py
sergisiso: Yes, that's better

src/psyclone/transformations.py
sergisiso: I don't understand why we checked that all directives in the same schedule had the same async_queue? Wouldn't that mean that we cannot do ACCEnterData asynchronously at all?

src/psyclone/psyGen.py
sergisiso: I delete it and there was no issues

src/psyclone/transformations.py

src/psyclone/psyir/nodes/node.py
sergisiso: I have reverted this for now, as I see we have the same problem with graphviz and I have not yet figure out how to solve it

src/psyclone/transformations.py
arporter: Yes, it would mean that. I think we have this limitation because we hadn't got as far as doing the necessary dependence analysis to check which parts of a schedule could safely be dispatched to different queues. We know that just putting lots of kernel launches into the same queue is beneficial and, since we have never done much with ACCEnterData, we just kept this check for safety. Therefore, I think we still need this check until such time as it's worth us doing the work to support the use of different queues.

arporter: Also, presumably we can start the asynchronous ACCEnterData region and then overlap it with other setup code before we get to kernel launches? Perhaps??

arporter (changes requested):
Thanks Sergi. Apart from the need to put back (or discuss) the check for different asynch. queues in EnterData, this is pretty much ready to go now. I see the integration tests were all OK previously so won't re-run them now.

src/psyclone/transformations.py
sergisiso: I reverted this change as it was not the purpose of this PR (but I think we can probably delete this validation in the future)

sergisiso:
I reverted the controversial change and removed #3250 from the list of issues that the PR will close since finally it doesn't reach 100% coverage and I still want to look to the remaining lines at some point.

Ready for another look @arporter

arporter (approved):
All looks good now, thanks Sergi.
As master has now moved on, I'll re-run the IT and then proceed to merge.

Issue #2674, opened 2024-07-26
New LFRic BuiltIns - field_min_max and copy_field routines based on intrinsic Fortran precisions


rjlmalone:
Introduction: The release of LFRic version 1.1 has seen psykal_builtin_light module replaced by sci_psykal_builtin_light_mod,
which, at present, provides support for field_min_max and two other routines (pending Psyclone issue #489), but does 
not include the copy_field invokes, which use similar parallelised methods, based on intrinsic Fortran precisions. 
Field_min_max and copy_field are covered by LFRic tickets #4109 and #4024.

**1/ LFRic Core version 1.0**

Ref: [log:LFRic/branches/dev/richardmalone/apps1.0_b6]

1.0/ The relevant modules are located at (apps1.0_b6):

	       ./components/science/source/algorithm/field_minmax_alg_mod.f90
	       ./components/science/source/algorithm/copy_field_alg_mod.f90
               ./components/science/source/psy/psykal_builtin_light_mod.f90

1.1/ Psykal_builtin_light_mod contained two new classes of invoke subroutines for field min_max and field copy operations, 
optimised for parallel processing, utlilising only intrinsic iso Fortran numeric precisions. These are distinct from 
similar functions based on science numeric classes (r_single, r_double, r_def, r_tran, r_solver).

1.2/ These new invoke functions, used in field_min_max_alg_mod and copy_field_alg_mod, are as follows:

     		invoke_r32_field_min_max
      	 	invoke_r64_field_min_max
       		invoke_copy_field_32_64
       		invoke_copy_field_64_32
       		invoke_copy_field_32_32
       		invoke_copy_field_64_64

       where: [32, 64] are intrinsic iso Fortran real precisions 

**2/ LFRic Core version 1.1**

Ref: [log:LFRic/branches/dev/richardmalone/core1.1_b6N]

2.0/ The relevant modules are located at (core1.1_b6N):

	./components/science/source/algorithm/field_minmax_alg_mod.f90
	./components/science/source/algorithm/copy_field_alg_mod.f90
        ./components/science/source/psy/sci_psykal_builtin_light_mod.f90

2.1/ In Core version 1.1, field_min_max invoke routines currently reside in sci_psykal_builtin_light_mod, with note that:

 	! This is a PSyKAl-lite implementation of a built-in that will be implemented 
  	! under PSYclone issue #489. See that issue for further details.

[...]

Issue #2381, opened 2023-10-27
Introduce GlobalReduction object


TeranIvy:
This functionality is required for global reduction built-ins in #2021 and #489, as we currently only have `GlobalSum` and LFRic child class.

There are two attempts to adding this functionality, PR #2045 and PR #2111, both linked with issue #2021. In PR #2045, he codte and documentation changes were done together, so we agreed to have two separate PRs, one for code changes and one for the documentation changes.

I will link #2111 to this issue.

arporter:
Language-level PSyIR has support for various intrinsics including `MAXVAL` and `MINVAL`. At the DSL level, as @TeranIvy says, we have `psyclone.lfric.LFRicGlobalSum` which subclasses `psyclone.psyGen.GlobalSum`. `GlobalSum` is not sub-classed by anything else (i.e. GOcean doesn't have support for global sums).

One possible design is to introduce a new `psyclone.domain.common.psylayer.GlobalReduction` class. `GlobalSum` would then be moved to `psyclone.domain.common.psylayer.GlobalSum` and inherit from the new class. Most of the important work in the current LFRicGlobalSum is done in the `lower_to_language` method which generates language-level PSyIR for the reduction. Presumably this would be relatively simple to parameterise such that it would generate appropriate code for the different types of reduction (min, max as well as sum).
Looking at:
 https://github.com/stfc/PSyclone/blob/dc66e55c00740c39580c76b5b353ab9f0825ae4c/src/psyclone/lfric.py#L3945-L3964
it is probably as simple as having two class properties - one holding the name of the reduction ("min", "max", "sum") and one (e.g. "get_sum") the name of the method that must be called on the LFRic field object.

arporter:
Excluding the tests, `LFRicGlobalSum` is *only* used in the constructor of `lfric.LFRicInvoke`:
https://github.com/stfc/PSyclone/blob/dc66e55c00740c39580c76b5b353ab9f0825ae4c/src/psyclone/domain/lfric/lfric_invoke.py#L182-L189
Here, we'd need to add the appropriate LFRic reduction node, depending on the access of the scalar argument to the kernel.

arporter:
Actually, a simpler hierarchy might be:
```
psy_layer.GlobalReduction -> domain.lfric.GlobalReduction
   |                                               |
   ---- GlobalReduction.Reduction.SUM              ---- method_name
   |
   ---- Argument
```

arporter:
[...]

PR #3263, opened 2025-12-18
(Towards #2381 #2674) generalise GH_SUM Kernel metadata to GH_REDUCTION and begin support for new min/max LFRic builtins

arporter:
This is a step towards both #2381 (generalising global reductions) and #2674 (new field_min_max builtin). It attempts to reduce the size of #3222 by *only* adding the two new access types for reductions as well as the skeletons for the new builtins.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   28) PR #3263 towards #2381 and #2674. Generalise GH_SUM to GH_REDUCTION
+   and begin support for new reduction LFRic Builtins.
+
    27) PR #3293. Fixes various typos throughout the PSyclone codebase.
 
    26) PR #3267 for #3265 and #1418. Improves the parsing of Fortran
--- a/src/psyclone/tests/core/access_sequence_test.py
+++ b/src/psyclone/tests/core/access_sequence_test.py
@@ -102,6 +102,9 @@ def test_access_info_is_any_read_or_write():
     access_info = AccessInfo(AccessType.WRITE, Node())
     assert not access_info.is_any_read()
     assert access_info.is_any_write()
+    access_info = AccessInfo(AccessType.REDUCTION, Node())
+    assert not access_info.is_any_read()
+    assert access_info.is_any_write()
 
     # read and write types
     access_info = AccessInfo(AccessType.INC, Node())


arporter:
Ready for a first look from anyone still working @sergisiso or @hiker ?
It *shouldn't* affect the IT but I'll fire them off to be on the safe side.

config/psyclone.cfg
```diff
@@ -64,7 +64,8 @@ FORTRAN_STANDARD = f2008
 # ==================================
 [lfric]
 access_mapping = gh_read: read, gh_write: write, gh_readwrite: readwrite,
-                 gh_inc: inc, gh_readinc: readinc, gh_sum: sum
+                 gh_inc: inc, gh_readinc: readinc, gh_sum: sum,
+		 gh_min: min, gh_max: max
```
sergisiso: tab to spaces :)

doc/user_guide/lfric.rst
```diff
@@ -2797,7 +2797,8 @@ are listed in the table below.
 +===============+=====================+================+====================+
 | GH_SCALAR     | GH_INTEGER          | n/a            | GH_READ            |
 +---------------+---------------------+----------------+--------------------+
-| GH_SCALAR     | GH_REAL             | n/a            | GH_READ, GH_SUM    |
+| GH_SCALAR     | GH_REAL             | n/a            | GH_READ, GH_SUM,   |
+|               |                     |                | GH_MIN, GH_MAX     |
```
sergisiso: I don't understand why do we need GH_SUM, GH_MIN or GH_MAX. These is the output scalar of the built-in which will only be written to, Why not GH_WRITE?

src/psyclone/core/access_type.py
```diff
@@ -75,6 +76,10 @@ class AccessType(Enum):
     #: is available at compile-time and can be used for type properties such
     #: as kinds or dimensions.
     CONSTANT = 10
+    #: Is the output of a MIN reduction (i.e. global minimum value).
+    MIN = 11
+    #: Is the output of a MAX reduction (i.e. global maximum value).
+    MAX = 12
```
sergisiso: Again MIN/MAX is not an AccessType (same with the pre-existing SUM). It is even more confusing that this refer to the output scalar, which is just written to. So why do we differentiate them?

doc/user_guide/lfric.rst
arporter: This is a good point. Currently, we rely on knowing the type of reduction being performed by a built-in when creating the Invoke (in lfric_invoke.py):
image
At this point, we don't have the PSyIR for the built-in - just its name and metadata. Therefore, if the scalar argument was just `GH_WRITE` we would know that it was a reduction but could not determine the type. I feel we may have conflated DSL metadata "access" information with PSyIR access concepts.

src/psyclone/core/access_type.py
arporter: As commented elsewhere, I agree. I'll look to see if I can restrict the MIN/MAX/SUM concept to the kernel metadata.

arporter:
My feeling is that we currently have quite a lot of domain-specific information in AccessType which should not be there (e.g. INC). All of the generic PSyIR should, as you say, use READ/WRITE/READWRITE. The concept of a reduction or INC access belongs with a domain. I was therefore planning on separating the two out and ensuring that the domain-specific access types are only referred to in appropriate places in the code.

I'm not that keen on the idea of having a mapping of particular built-ins to the type of reduction they perform - it feels more natural to keep that information as part of the metadata associated with the kernel.

The potential future re-use of `ATOMIC_WRITE` is attractive but I'd be concerned about letting that out into the wild, even unadvertised (we could have it as an internal access type but not as kernel metadata perhaps?).

config/psyclone.cfg
arporter: Ah! Done.

arporter:
Having now realised that we also have `AccessType.INC` and `AccessType.READINC`, it's going to be no small matter to attempt to unpick this (see e.g. https://github.com/stfc/PSyclone/pull/3272). I therefore propose that we keep things as they are for the moment and have a more in-depth discussion about whether we want to tackle this.

arporter:
@sergisiso ready for another look and/or continued discussion.

arporter:
> My point is that we don't need to repeat this information again in the access type, because the access type is the same for all of them. Would calling the metadata GH_REDUCTION and the AccessType.REDUCTION instead of GH_SUM, GH_MIN, GH_MAX and SUM,MIN,MAX be alright?

That's a good idea. Fortunately, since only built-ins can do reductions this change wouldn't affect any existing LFRic code (I think). I can try doing that and see how it goes.

arporter:
This is ready for another look now @sergisiso.

src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -162,12 +163,27 @@ class LFRicBuiltIn(BuiltIn, metaclass=abc.ABCMeta):
...
+        '''
+        #: No reduction.
+        NONE = auto()
+        #: Summation reduction.
+        SUM = auto()
+        #: Global minimum value.
+        MIN = auto()
+        #: Global maximum value.
+        MAX = auto()
+
```
sergisiso, line 182: This is much better than before, but I still don't know why we need this. This is only used in:
`if kern.reduction_type == LFRicBuiltIn.ReductionType.SUM:`
why can we use:
`if isinstance(kern, LFRicSumXKern):`
?

arporter: There are several builtins that perform global sums. Therefore, we either do it this way or we maintain a separate list of which builtin performs which type of reduction. I think this way is nicer as the information lives with the kernel class.

sergisiso: OK, it's not how I would have done it but I won't linger on this point

PR #3264, opened 2025-12-19
(Closes #3259) shorten generated names for basis/diff-basis arrays

### LINKED ISSUE #3259: Names of generated LFRic basis functions can be too long for compiler
I've just had the following error trying to compile some new LFRic code:
```
  103 |       REAL(KIND=r_def), allocatable :: basis_aspc2_source_surface_altitude_on_aspc1_surface_altitude_w0_k0(:,:,:)
      |                                                                                                        1
Error: Name at (1) is too long
```
This is because my kernel takes the arguments `source_surface_altitude` (in `ANY_SPACE_2`) and samples it at `surface_altitude_w0_k0` in (`ANY_SPACE_1`). The variable name above was generated for the basis function variable in the psy-layer.

Could instances like this be avoided by not appending the field variable names to the basis function variable name?

arporter:


arporter:
This PR shortens the generated names used for various variables in order to avoid hitting compiler limitations as reported by Tom in #3259. Since we're no longer basing the names on existing variable names (which must be unique) I've had to increase the use of tags to ensure we cope in the face of name collisions.
One for @hiker, @LonelyCat124 or @sergisiso.
I've fired off the integration tests.

src/psyclone/lfric.py
```diff
@@ -3820,7 +3822,7 @@ def __init__(self, node):
                 bc_fs = op_arg.function_space_to
                 self._boundary_dofs.append(self.BoundaryDofs(op_arg, bc_fs))
 
-    def invoke_declarations(self):
+    def invoke_declarations(self) -> None:
```
sergisiso, line 3828: Maybe this is a stylistic choice, but is the "-> None" needed when there is no return? None of my tooling complains with or without

src/psyclone/lfric.py
```diff
@@ -3257,8 +3258,9 @@ def initialise(self, cursor: int) -> int:
         # Allocate basis arrays
         for basis in basis_arrays:
             dims = "("+",".join([":"]*len(basis_arrays[basis]))+")"
-            symbol = self.symtab.find_or_create(
-                basis, symbol_type=DataSymbol, datatype=UnsupportedFortranType(
+            symbol = self.symtab.find_or_create_tag(
+                basis,
+                symbol_type=DataSymbol, datatype=UnsupportedFortranType(
                     f"real(kind=r_def), allocatable :: {basis}{dims}"
```
sergisiso: As you did somewhere else, you will need to do:
`new_name = self.symtab.next_available_name(basis)`
if you embed the name in the datatype (as it may end up being different)

src/psyclone/lfric.py
```diff
@@ -3436,7 +3438,7 @@ def _initialise_xyoz_qr(self, cursor):
                 const.QUADRATURE_TYPE_MAP["gh_quadrature_xyoz"]["intrinsic"]
             kind = const.QUADRATURE_TYPE_MAP["gh_quadrature_xyoz"]["kind"]
             for name in self.qr_weight_vars["xyoz"]:
-                self.symtab.find_or_create(
+                self.symtab.find_or_create_tag(
```
sergisiso, line 3444: `new_name = self.symtab.next_available_name(...)`

src/psyclone/domain/lfric/function_space.py
```diff
@@ -159,17 +158,37 @@ def _mangle_fs_name(self):
...
+        Shortens the provided name by splitting on any underscore characters
+        and reconstructing the name using the beginning and
+        ending chars from each token.
+
+        :param name: the name to shorten.
+
+        :returns: a shortened form of the name.
+
+        '''
+        new_parts = []
```
sergisiso, line 139: Maybe its a pain to do not that you already changed lots of tests, but some of the generated code is more cryptic now, have you considered doing:
```
if len(name) < threshold:
   return name
```
to keep already short names descriptive?

src/psyclone/domain/lfric/function_space.py
```diff
@@ -197,9 +216,9 @@ def _shorten_fs_name(self):
                 f"{const.VALID_ANY_DISCONTINUOUS_SPACE_NAMES} spaces.")
 
         # Split name string to find any_*_space ID and create a short name as
-        # "<start>" + "spc" + "ID"
+        # "<start>" + "s" + "ID"
```
sergisiso, line 171: Is this function still necessary now?, we could do:
`self._shorten_arg_name(self.orig_name + "_" + arg.name)}`

But if you decide not to remove the short_name do we really need to keep the _short_name and _mangled_name as state, it may be safer to compute them in the property method.

src/psyclone/domain/lfric/function_space.py
```diff
@@ -159,17 +158,37 @@ def _mangle_fs_name(self):
...
+                if (fspace and fspace.orig_name.lower() == lorig_name):
+                    return (f"{self._short_name}_"
+                            f"{self._shorten_arg_name(arg.name)}")
         # Raise an error if there are no kernel arguments on this
         # function space
         raise FieldNotFoundError(f"No kernel argument found for function "
                                  f"space '{self._orig_name}'")
 
+    @staticmethod
+    def _shorten_arg_name(name: str) -> str:
```
sergisiso: Maybe just "_shorten_name", it doesn't use the fact is an arg.

src/psyclone/lfric.py
arporter: I believe some tools do complain - @hiker mentioned this.

src/psyclone/lfric.py
arporter: Oh yes, done.

src/psyclone/lfric.py
arporter: Good spot.

src/psyclone/domain/lfric/function_space.py
arporter: Ah! That's a good idea. I've done some googling and realised that the standard actually mandates a maximum length. F2008:
image
while in F95 it was 31 apparently.

src/psyclone/lfric.py
sergisiso: I suppose this is mypy, I am using ty for typechecking and it does not complain. Anyway, if we need something I suggest NoReturn https://docs.python.org/3/library/typing.html#typing.NoReturn , the method does not return None.

arporter: I think it does. `NoReturn` indicates that a function never returns while a function without an explicit `return` does return None (at least, that's what pylint complains about if you ever have a function where it thinks you've missed a `return` statement).

sergisiso: Ah yes, I agree

src/psyclone/domain/lfric/function_space.py
arporter: I've kept this one but have, as you suggest, removed the internal state in favour of dynamic computation.

src/psyclone/domain/lfric/function_space.py
arporter: Oh yes.

src/psyclone/domain/lfric/function_space.py
```diff
@@ -137,50 +97,64 @@ def _mangle_fs_name(self):
...
+        lorig_name = self._orig_name.lower()
         for arg in args:
             for fspace in arg.function_spaces:
-                if (fspace and fspace.orig_name.lower() ==
-                        self._orig_name.lower()):
-                    mngl_name = self._short_name + "_" + arg.name
-                    return mngl_name
+                if (fspace and fspace.orig_name.lower() == lorig_name):
+                    return (f"{self.short_name}_"
+                            f"{self._shorten_name(arg.name)}")
```
sergisiso: This can still surpass 63 if only part of the name is shortened, should we provide self.short_name+'_'+arg.name as the argument to _shorten_name?

src/psyclone/tests/domain/lfric/arg_ordering_test.py
```diff
@@ -267,9 +267,9 @@ def test_arg_ordering_generate_cma_kernel(dist_mem, fortran_writer):
         'cell', 'nlayers_lma_op1', 'ncell_2d', 'lma_op1_proxy%ncell_3d',
         'lma_op1_local_stencil', 'cma_op1_cma_matrix', 'cma_op1_nrow',
         'cma_op1_ncol', 'cma_op1_bandwidth', 'cma_op1_alpha', 'cma_op1_beta',
-        'cma_op1_gamma_m', 'cma_op1_gamma_p', 'ndf_adspc1_lma_op1',
-        'cbanded_map_adspc1_lma_op1', 'ndf_adspc2_lma_op1',
-        'cbanded_map_adspc2_lma_op1']
+        'cma_op1_gamma_m', 'cma_op1_gamma_p', 'ndf_ads1_lma_op1',
+        'cbanded_map_ads1_lma_op1', 'ndf_ads2_lma_op1',
+        'cbanded_map_ads2_lma_op1']
```
sergisiso, line 272: How come that these (and most other test cases) are still being shorten if these are below the threshold?

arporter: I kept the shortening of "aspc" -> "as" and "adspc" -> "ads" as I thought it was worth having.

src/psyclone/domain/lfric/function_space.py
arporter: Oh yes, good plan.

src/psyclone/tests/test_files/lfric/11.3_any_space.f90
```diff
@@ -46,10 +46,10 @@ program any_space_example
 
   implicit none
 
-  type(field_type)           :: a, f
-  type(operator_type)        :: b, c, d, e
+  type(field_type)           :: a, a_field_with_a_very_long_name
+  type(operator_type)        :: b, c, an_operator_with_a_very_long_name, e
```
sergisiso, line 50: I was ready to approve but I checked this example (from src/psyclone/tests/lfric_test.py) and the basis_/diff_basis_ /map_/ndf_/undf_ do get shortened to, .e.g: diff_basis_a4_an_or_wh_a_vy_lg_ne_qr
but the proxy and local stencil are still:
- an_operator_with_a_very_long_name_proxy
- an_operator_with_a_very_long_name_local_stencil

I know this is not exactly in the title of the PR, but do you want to also solve those?
And why did the new backend check not complain here?

arporter: My feeling is that those are OK as the bulk of the length is due to the original name chosen by the scientist and they have control over that (and I'd like to get on to new things :-) ). The new backend check did not complain because it is generous and follows the F2008 standard of 63 chars.

sergisiso: Ah sorry, I didn't realise the limit is 63 and the fs_name max length is set to 21, this is more than enough

PR #3275, opened 2026-01-12
(closes #3216) Allow specifying explicit private variables for parallel regions

### LINKED ISSUE #3216: Explore how to specify explictly_private_symbols for region directives
The method also needs better documentation/examples.

sergisiso:
Currently we have `Loop.explictly_private_symbols` to force a different result than the automatic sharing attributes provide. However it is not possible to do the same with parallel regions. To fix this I moved this attribute to the SharingAttributeMixin. It makes more sense here as it is being set and used by the same class (better encapsulation) and can be used by ALL subclases adding sharing attribute clauses.
 
There was also the problem that users didn't find out this capability easily. This can be improved by adding a transformation option, as this is the documentation that people is looking at while transforming code.

So when LFRic_atm scripts currently do:
```
# set some symbols to be PRIVATE
mark_explicit_privates(loop, symbols_to_add)

OMPParallelLoopTrans(omp_schedule="dynamic").apply(
    loop, options=opts
)
```
can be:
```
OMPParallelLoopTrans(omp_schedule="dynamic").apply(
    loop, force_private=symbols_to_add, **opts
)
```
The transformation option logs (but doesn't fail) for symbols not found, to have the same behaviour as `mark_explicit_privates`.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,5 +1,8 @@
-   28) PR #3275 for #3216. Allows the specification of private variables
+   29) PR #3275 for #3216. Allows the specification of private variables
    for parallel regions in user scripts and removes the option from Loops.
+
+   28) PR #3263 towards #2381 and #2674. Generalise GH_SUM to GH_REDUCTION
+   and begin support for new reduction LFRic Builtins.
 
    27) PR #3293. Fixes various typos throughout the PSyclone codebase.
 
--- a/doc/user_guide/lfric.rst
+++ b/doc/user_guide/lfric.rst
@@ -793,7 +793,7 @@ intents there are always ``intent(in)``.
 The Fortran intent of :ref:`scalars <lfric-scalar>` is still defined
 by their :ref:`access metadata <lfric-kernel-valid-access>` as they are
 actual data. This means ``intent(in)`` for ``GH_READ`` and ``intent(out)``
-for ``GH_SUM`` (more details in :ref:`meta_args <lfric-api-meta-args>`
+for ``GH_REDUCTION`` (more details in :ref:`meta_args <lfric-api-meta-args>`
 section below).
 
 The intent of other data structures is mandated by the relevant
@@ -1143,7 +1143,7 @@ The third component of argument metadata describes how the Kernel
 makes use of the data being passed into it (the way it is accessed
 within a Kernel). This information is mandatory. There are currently 6
 possible values of this metadata ``GH_READ``, ``GH_WRITE``,
-``GH_READWRITE``, ``GH_INC``, ``GH_READINC`` and ``GH_SUM``. However,
+``GH_READWRITE``, ``GH_INC``, ``GH_READINC`` and ``GH_REDUCTION``. However,
 not all combinations of metadata entries are valid and PSyclone will
 raise an exception if an invalid combination is specified. Valid
 combinations are specified later in this section (see
@@ -1174,9 +1174,9 @@ combinations are specified later in this section (see
   subsequently incremented. Therefore this is equivalent to a
   ``GH_READ`` followed by a ``GH_INC``.
 
-* ``GH_SUM`` is an example of a reduction and is the only reduction
-  currently supported in PSyclone. This metadata indicates that values
-  are summed over calls to Kernel code.
+* ``GH_REDUCTION`` indicates a reduction. Only Built-ins may perform
+  reductions. The type of reduction (sum, maximum value, minimum value)
+  is a property of the particular Built-in.
 
 For example::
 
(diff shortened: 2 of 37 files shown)


LonelyCat124:
I'm not going to review this today, just having a quick look at it now - is it worth having an example showing how to use this? We had people in NG-Arch using explicit private on `Loop` nodes which will no longer function and an example/docs to explain how to use this new functionality on the Directives (which is currently likely to be quite common) might be useful? @sergisiso

sergisiso:
When merging the NG-Arch tickets its use has been standarised to the "mark_explicit_privates" from the LFRic psyclone_tools.py that I show in my top comment, so it is a one-line change. But I agree regarding the example/more docs comment.

sergisiso:
I added and example as suggested (currently in nemo/eg8, but we could move it to the psyir folder once there is no conflict with @LonelyCat124 current changes).

The documentation is now part of the Transformation options. This is done on purpose to make it more discoverable, it is what people will probably open. And it is automatically inherited to all subclasses through kwargs:
image

@LonelyCat124 @arporter This is ready to review again

examples/nemo/eg8/Makefile
```diff
@@ -0,0 +1,47 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2020-2026, Science and Technology Facilities Council.
```
LonelyCat124: Copyright date should just be 2026 right? Since its a new file

examples/nemo/eg8/add_parallelism.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2018-2025, Science and Technology Facilities Council
```
LonelyCat124: Copyright year.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -429,11 +416,10 @@ def variable(self, var):
 
     def replace_symbols_using(self, table_or_symbol):
```
LonelyCat124: Can you update this to use type hints instead of the doc `:type:`?

src/psyclone/psyir/transformations/parallel_loop_trans.py
```diff
@@ -101,14 +105,21 @@ def _directive(self, children, collapse=None):
         '''
 
     @staticmethod
-    def _attempt_privatisation(loop, symbol_name, dry_run=False):
```
LonelyCat124, line 105: What happened to the dry_run option here? I guess its not needed because we no longer modify the tree, only affect some local variables to the transformation and resultant directive>

LonelyCat124 (commented):
@sergisiso This mostly looks ok to me, I've not done all the local stuff yet - couple of copyright dates need updating but I'll do a final review once we can run integration with a local lfric_atm/core

examples/nemo/eg8/Makefile
sergisiso: Done

examples/nemo/eg8/add_parallelism.py
sergisiso: Done

src/psyclone/psyir/nodes/loop.py
sergisiso: Done

src/psyclone/psyir/transformations/parallel_loop_trans.py
sergisiso: Exactly, there is no modification of the tree. The output now is the optional output_set, and in the validation this is not provided, as we don't need them anymore once they are checked.

examples/nemo/eg8/output.f90
```diff
@@ -0,0 +1,52 @@
+program add_parallelism_example
```
LonelyCat124: @sergisiso I think this needs excluding from the example?

LonelyCat124 (changes requested):
@sergisiso Almost there - I think a result file from eg8 made it into the commits, can you remove that and I'll set integration running.

examples/nemo/eg8/output.f90
sergisiso: Oh yes, and now that we merged your restructuring of examples/psyir I also wanted to move the example there as there is nothing NEMO-specific and I think is a good introduction to how to use the parallelisation transformation.

sergisiso:
@LonelyCat124 This is ready for another look. I added new Reviewer's steps in the Wiki for changing the lfric hash https://github.com/stfc/PSyclone/wiki/CodeReview/_compare/3483b8f58015f2ae2b4e4f12fde0b1ca45eaa11c

LonelyCat124 (approved):
Running integration one last time for sanity, but ready to merge if all is green.

PR #3280, opened 2026-01-13
(Towards #2971) Moved OMPParallelTrans to psyir.transformations

LonelyCat124:


LonelyCat124:
@sergisiso @arporter Mostly a small PR ready for a look, just moves the transformation and fixes the scripts I found. I'm going to set integration tests running now as well in case I missed anything.

.github/workflows/lfric_test.yml
```diff
@@ -291,9 +291,14 @@ jobs:
...
 
+        # PSyclone scripts must now be under 'optimisation/psykal' and be called 'global.py'
+        export OPT_DIR=${LFRIC_DIR}/applications/lfric_atm/optimisation/psyclone-test/psykal
+        mkdir -p ${OPT_DIR}
+        cp ${PSYCLONE_LFRIC_DIR}/everything_everywhere_all_at_once.py ${OPT_DIR}/global.py
+
         # Build the 'meto-ex1a' platform, that includes psyclone-transmutation
         export BUILD_START="${SECONDS}"
-        ./build/local_build.py -a lfric_atm -p meto-ex1a -j 16 -w workdir-lfric-atm-transmuted
+        ./build/local_build.py -a lfric_atm -p psyclone-test -j 16 -w workdir-lfric-atm-transmuted
```
arporter: Does this mean we're no longer testing the 'transmutation' if we're using our own "everything_everywhere..." script? That doesn't seem right?

arporter: Could we patch the transformation script(s) used by LFRic?

src/psyclone/psyir/transformations/omp_parallel_trans.py
```diff
@@ -0,0 +1,135 @@
...
+# Authors R. W. Ford, A. R. Porter, S. Siso and N. Nobre, STFC Daresbury Lab
+#         A. B. G. Chalk STFC Daresbury Lab
+#         J. Henrichs, Bureau of Meteorology
+# Modified I. Kavcic, J. G. Wallwork, O. Brunt and L. Turner, Met Office
+#          S. Valat, Inria / Laboratoire Jean Kuntzmann
+#          M. Schreiber, Univ. Grenoble Alpes / Inria / Lab. Jean Kuntzmann
+#          J. Dendy, Met Office
+'''This module provides the OMPParallelTrans transformation.'''
+
+from psyclone.psyir.transformations.parallel_region_trans import (
```
arporter: If possible (i.e. you don't get circular imports), pls put these import statements in alpha-order by module name.

LonelyCat124: Fixed.

.github/workflows/lfric_test.yml
LonelyCat124: I'm a bit clueless here I just did what @sergisiso suggested I thought - if I misunderstood I can fix that, let me know if so Sergi.
Its probably patchable but I'd need to find the script, its just removing OMPParallelTrans from the transformations import and swapping it to psyir.transformation.

sergisiso: > Does this mean we're no longer testing the 'transmutation' if we're using our own "everything_everywhere..." script?

Yes, but not because we use this file for psykal transformations, but because we created a new optimisation target "psyclone-test" without a transmute folder. 

A quick solution would be to also copy the transmute files from meto-ex1a `cp applications/lfric_atm/optimisation/meto-ex1a/transmute/ applications/lfric_atm/optimisation/psyclone-test/transmute/` however if any of those also gets OMPParallelTrans from TransInfo it will also fail.

LonelyCat124: Oh yeah I'd forgotten this change means `OMPParallelTrans` is no longer available from TransInfo.

LonelyCat124: One place its defined in lfric that is a problem for lfric_atm is unfortunately in `lfric_core/infrastructure/build/psyclone/psyclone_tools.py` where they use:
```
from psyclone.transformations import (
    Dynamo0p3ColourTrans,
    Dynamo0p3OMPLoopTrans,
    Dynamo0p3RedundantComputationTrans,
    OMPParallelTrans,
)
```
This might be possible to fix but I'm not sure how, we'd need to patch it somehow - @arporter I think up to you if you want me to try.

Its also imported in `applications/lfric_atm/optimisation/meto-ex1a/psykal/algorithm/casim_alg_mod.py`, for lfric atm. It is used in a few other application scripts but I assume we don't hit those in our tests.

arporter: > This might be possible to fix but I'm not sure how, we'd need to patch it somehow - @arporter I think up to you if you want me to try.
>
Yes please, I think it's important we keep this. We do already patch lfric for the nvidia compiler:
https://github.com/stfc/PSyclone/blob/2823f9df2fb018f077d76cc1b9c9cde4827d88b2/.github/workflows/lfric_test.yml#L101
so hopefully it's doable for this too. Or have a little bash script that `find`s all .py files, checks for OMPParallelTrans and updates as necessary?

sergisiso: It could get complicated with patches (I am also breaking lfric_atm scripts in #3275). But since I am updating to git in another PR and lfric_apps is public and on git what do you think if we create a fork of MO:lfric_apps with we do the scripts breaking changes there. Then with the release we can push the changes to the MO repository or at least point it to them in order to do the update.

arporter: That's a good idea Sergi - sounds like the most straightforward way to go.

LonelyCat124: @sergisiso Should we do this for lfric_core too? There are some scripts in there that contain base functionality used by lfric_apps (e.g. `lfric_core/infrastructure/build/psyclone/psyclone_tools.py`)

arporter (dismissed):
All looks good now, thanks Aidan.
I'll re-run the ITs just to be sure.

PR #3284, opened 2026-01-15
#3283 Fix #3283 by using a set instead of a list.

### LINKED ISSUE #3283: Incorrect "More than one match for kernel" message
If a directory with a kernel source is (directly or indirectly) used more than once specified using the `-d` command line option, `get_kernel_filepath` reports that `More than one match for kernel`, because the same kernel file is matched twice (based on different search paths). Example command line option:
```
psyclone -d $lfric_source/kernel -d $lfric_source
```
Imho, if there is only one file (found several times), this should be accepted, and this is a rather trivial change.

FWIW, the double `-d` option happens with Fab when using kernel extraction (adding both options is neater design compared with providing either of the two as required). Alternative, we could sort out #2793 (i.e. adding more command line options that add paths to the module manager, but not kernel search path, then I could use the other option for one of the paths).

hiker:
Besides fixing #3283, I've fixed some pylint errors in the test.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   24) PR #3284 for #3283. Bug fix for case where same file found via
+   multiple search paths.   
+
    23) PR #3255 for #3251. Identify uncategorised directives as
    UnknownDirective (instead of CodeBlock) when possible.
 


arporter (approved):
All looks good, thanks Joerg.
I see ITs were green so I won't re-run them.
Errors building docs are not the fault of this PR.

PR #3289, opened 2026-01-19
(Closes #3288) fix the docs build

### LINKED ISSUE #3288: Docs build is failing
It seems that the GHA runner has recently upgraded its version of Sphinx and the docs now 'fail' because we get 39 warnings from Sphinx:

Image

arporter:
I got some way to improving things by being more precise about imports and adding some type hints. However, the ultimate 'fix' was to pin sphinx to version 8.3.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,5 @@
+   25) PR #3289 for #3288. Pin Sphinx to 8.3 to prevent doc build issues.
+
    24) PR #3284 for #3283. Bug fix for case where same file found via
    multiple search paths.   
 


src/psyclone/psyir/nodes/__init__.py
```diff
@@ -132,7 +132,6 @@
         'Literal',
         'Loop',
         'Member',
-        'Node',
```
sergisiso, line 135: What is the rational of this change?

We won't be able to do `from psyclone.psyir.nodes import Node` anymore?

arporter: I think that will still work. This change just stops AutoAPI from attempting to document Node as part of the `nodes` module. I was actually thinking of preventing the import of `Node` from `nodes` as that kind of thing seems to lead to circular imports and `Node` is an internal thing anyway. However, given that this change is not strictly necessary and needs wider agreement I'm happy to revert this.

Issue #3144, opened 2025-09-21
Module Inlining handling of wildcard imports is unsafe.


hiker:
I am using transformation only (to make it confusing, I am using the dl_esm_inf infrastructure, BUT I am not using any invokes ... I never had time to remove the infrastructure from this example in my training), so I am just using plain PSyIR transformations.

The main work is done here:
```
module time_step_mod
...
time_step(grid, current, time_steps)
        USE grid_mod, only             : grid_type
        USE field_mod, only            : r2d_field, go_t_points
        use compute_die_mod, only      : compute_die
        use compute_born_mod, only     : compute_born
        use combine_mod, only          : combine
        use count_neighbours_mod, only : count_neighbours
        use output_field_mod, only     : output_field

        implicit none
        TYPE(grid_type), intent(in), target :: grid
        TYPE(r2d_field), intent(inout) :: current
        integer, intent(in)            :: time_steps

        ! Declare the fields required
        TYPE(r2d_field)                :: neighbours, die, born
        integer                        :: time

        ! Create fields required:
        neighbours = r2d_field(grid, GO_T_POINTS)
        die        = r2d_field(grid, GO_T_POINTS)
        born       = r2d_field(grid, GO_T_POINTS)

        if(time_steps<=20) call output_field(current)
        ! In each time step:
        ! 1. Count neighbours
        ! 2. Compute newly born cells
        ! 3. Compute dying cells.
        ! 4. Combine the results from 2. and 3. with current state
        do time =1, time_steps
            call count_neighbours(neighbours, current)
            call compute_born(born, current, neighbours)
            call compute_die(die, current, neighbours)
            call combine(current, die, born)
            if(time_steps<=20) call output_field(current)
        enddo
```

When I inline any of the four subroutines called in the time stepping loop using:
```
    kmit = KernelModuleInlineTrans()
    inline = InlineTrans()
    for call in psyir.walk(Call):
        if call.routine.name != "output_field":
            kmit.apply(call)
            inline.apply(call)
```

the modified `time_step.f90` file contains:
```
    use grid_mod, only :

```
The `grid_type` is missing. If I import other symbols from that modules, they are all there, but `grid_type` is still missing :( I can't see an obvious reason why `grid_type` is special.

I will provide a reproducer (on a branch) later.

hiker:
See https://github.com/stfc/PSyclone/tree/1623_add_training/tutorial/training/transformation/3.4/solution
[...]

Issue #2683, opened 2024-08-02
Rename (and move) KernelModuleInlineTrans


arporter:
{Copy/Inline}RoutineToLocalContainerTrans are possible naming options.

sergisiso:
I prefer to drop the "inline" which is confusing as this is not the typical meaning of inline, so my vote is for:
`CopyRoutineToLocalContainerTrans`

PR #3290, opened 2026-01-19
(Closes #2846) extend KernelModuleInlineTrans so that it re-names inlined routines

### LINKED ISSUE #2846: Module inlining a public routine results in uncompilable code if there is a wildcard import.
If we have a module that does `use my_mod` and contains a routine that calls `my_sub` and then we "module inline" `my_sub` into that module then we get un-compilable code:
```
module my_mod
    public
contains
    subroutine my_sub
    end subroutine my_sub
end module my_mod
```
and
```
module caller_mod
  use my_mod
contains
  subroutine other_sub
    call my_sub()
  end subroutine
end module caller_mod
```
This is because the compiler then has two routines named `my_sub` in scope. 

We can spot this situation by checking for a 'wildcard' import from the module containing the routine being inlined.
We could potentially solve it by adding a renaming to the import of the original routine symbol.

arporter:


arporter:
This is now proceeding OK. However, I've realised that, as things currently stand, we'll end up generating multiple OpenCL implementations of the same kernel (since the call to each is module-inlined separately and given a new name). This isn't ideal and the code duplication that could result from having multiple copies of the same kernel will substantially bloat the executable. We could have an option to the transformation to use a pre-existing version of the routine if it exists? What do @sergisiso @LonelyCat124 and @hiker think?

arporter:
I realised over lunch that my question over-simplifies things - the very first time we module-inline a routine, it gets re-named so a simple check on the name of any subsequent target routine won't reveal that it was previously module inlined.

sergisiso:
@arporter I believe the previous version was already doing a subroutine comparison (with the equal operator in the routine node) and decide if it was the same than the already existing or not. Could we do something similar? If the subroutine is equal in all but name* to the inlining candidate (e.g. after updating symbols), use the existing one.

* the 'all by name' is tricky because its not what the equal operator does.

Another option is change the behaviour of ModuleInline so it updates ALL the calls to that subroutine.

arporter:
> Another option is change the behaviour of ModuleInline so it updates ALL the calls to that subroutine.

I like that idea. We can have it as an option so that users can do either.

arporter:
If I run the failing tutorial test locally under 3.14.6(?) it works. If I revert to 3.13 then it fails!

arporter:
The failure occurs because the example *does* try to module-inline the same kernel(s) twice.

After module-inlining, the debug-string of the PSyIR shows:
```
...
    do j = neighbours%internal%ystart, neighbours%internal%ystop, 1
      do i = neighbours%internal%xstart, neighbours%internal%xstop, 1
        < kern call: count_neighbours_code >
...
  pure subroutine count_neighbours_code_inlined_(i, j, neighbours, c)
    double precision, dimension(:,:), intent(out) :: neighbours
...
```
so I'm not convinced the name of the GOcean kernel is updated correctly?

arporter:
That was a red herring - `__str__` of `CodedKern` used `self._name` rather than `self.name`. ~~The better question to ask is, why with python 3.14 are we failing to spot that the kernel is already module inlined?~~ We weren't - it was embarrassing user error.

arporter:
ITs were finally green. This is ready for review. One for @sergisiso or @LonelyCat124.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -330,117 +296,27 @@ def _prepare_code_to_inline(
                                 module_symbol.name)
         return copied_routines
 
-    @staticmethod
-    def _rm_imported_routine_symbol(symbol: Symbol,
```
sergisiso, line 334: Would this PR also close https://github.com/stfc/PSyclone/issues/3144?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -330,117 +296,27 @@ def _prepare_code_to_inline(
...
-                                    schedule: Routine,
-                                    table: SymbolTable):
-        '''
-        If the named symbol is in the supplied table (or an ancestor) *and* is
-        an import then it is removed. If the Container from which it was being
-        imported no longer has any imports associated with it then the
-        ContainerSymbol is also removed.
-
-        TODO #2846 - this method is only required because at present we do not
```
sergisiso, line 343: There remaining TODOs 2846:
```
554:      local copy. #2846 will lift this limitation.

examples/nemo/scripts/utils.py
128:                 "fatal_error"  # TODO #2846 - is brought into scope via
```

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -502,144 +377,68 @@ def apply(self, node, options=None, **kwargs):
...
-                sym = node.scope.symbol_table.lookup(external_callee_name)
-                table = sym.find_symbol_table(node)
-                table.rename_symbol(sym, caller_name)
-                new_sym = sym
-            else:
-                new_sym = node.scope.symbol_table.lookup(caller_name)
+            # Deal with the interface symbol - create a new, local
+            # private version.
+            new_sym = container.symbol_table.new_symbol(
+                interface_sym.name+"_inlined_",
```
sergisiso, line 455: As I mention in https://github.com/stfc/PSyclone/issues/2683 , I prefer to stop using "inline" in this transformation, I would use `_psyclone_internal_copy` (also dropping the final `_` makes it compose better with other postfixes, e.g. in OpenCL generation)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -330,117 +296,27 @@ def _prepare_code_to_inline(
...
-                    :py:class:`psyclone.psyir.nodes.Call`
         :param options: a dictionary with options for transformations.
-        :type options: Optional[Dict[str, Any]]
+        :param update_all: by default, only the supplied Kernel or Call is
+           updated by this transformation. If this argument is set to True then
+           this behaviour is modified and all calls to this Kernel/Routine in
+           the associated program unit are updated.
 
         '''
-        if isinstance(node, CodedKern) and node.module_inline:
```
sergisiso, line 440: Can the module_inline getter/setter/attribute be deleted now? It is only used in:
```
src/psyclone/psyGen.py
1290:                str(self.module_inline) + "]")

src/psyclone/tests/psyGen_test.py
812:    assert ckerns[0].module_inline is False
816:    assert ckerns[0].module_inline is True
817:    assert ckerns[1].module_inline is do_all

src/psyclone/domain/gocean/transformations/gocean_opencl_trans.py
205:            if not kern.module_inline:
```

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -502,144 +377,68 @@ def apply(self, node, options=None, **kwargs):
...
-                    kern.routine = Reference(new_sym)
+            node.routine = Reference(new_sym)
+            # pylint: disable=protected-access
+            node._schedules = updated_routines
+        else:
+            # Update the Call to point to the inlined routine.
+            node.routine.symbol = new_sym
+
+        # In theory we could remove the import of the original routine but
+        # that is dangerous and unnecessary so we don't.
```
sergisiso: " ... if it is a subroutine-scoped or module-private symbol ..." (as we can never remove public symbols)

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -330,117 +296,27 @@ def _prepare_code_to_inline(
...
-            f"old_{name}",
-            symbol_type=RoutineSymbol,
-            interface=ImportInterface(
-                csym, orig_name=name))
-
-    def apply(self, node, options=None, **kwargs):
+    def apply(self,
+              node: Union[CodedKern, Call],
+              options: dict[str, Any] = None,
+              update_all: Optional[bool] = False,
```
sergisiso: I am tempted to say that "update_all" is the expected default behaviour, would that cause may issues?

I assume this is mostly used as:
```
for kern in schedule.kernels():
    modinline.apply(kern)
```
Do we currently create a copy for each call?

Could we make this the transformation pass for "already locally-scope routines" so that the above doesn't need to be changed if we update the behaviour? It would be similar to things like Reference2ArrayRangeTrans that succeeds if it already has ranges.

Feel free to push back if this creates problems.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -83,17 +84,17 @@ def __str__(self):
                 "Container of the call site.")
 
     # pylint: disable=too-many-branches
-    def validate(self, node, options=None, **kwargs):
+    def validate(self,
```
sergisiso, line 129: If you are up to it, we could convert the example in the docstring to a doctested example.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -213,37 +213,36 @@ def test_validate_name_clashes():
...
-    schedule = psy.invokes.invoke_list[0].schedule
+    psy, invoke = get_invoke("4.6_multikernel_invokes.f90", "lfric", idx=0,
+                             dist_mem=False)
+    schedule = invoke.schedule
     coded_kern = schedule.children[0].loop_body[0]
     inline_trans = KernelModuleInlineTrans()
 
-    # Check that name clashes which are not subroutines are detected
+    # Check that name clashes which are not subroutines are handled
     schedule.symbol_table.add(DataSymbol("ru_code", ScalarType.real_type()))
```
sergisiso: I think this is meant to have an overlap with the candidate name, e.g. `ru_code_inlined_` or `ru_code_psyclone_internal_copy`

tutorial/training/gocean/2.10-GameOfLife-mpi/solution/openmp_combined.py
```diff
@@ -61,15 +60,12 @@ def trans(psyir: FileContainer) -> None:
...
     omp_do = OMPLoopTrans(omp_schedule="dynamic")
-    module_inline = KernelModuleInlineTrans()
 
     # We know that there is only one schedule
     schedule = psyir.walk(InvokeSchedule)[0]
 
-    # Module inline all kernels to help gfortran with inlining.
-    for kern in schedule.kernels():
-        module_inline.apply(kern)
-
```
sergisiso, line 72: Was this intentionally deleted? If yes, remove the comment.

sergisiso: I see this is repeated in other tutorial examples, but I see the PR has a comment regarding this file saying:

> The failure occurs because the example does try to module-inline the same kernel(s) twice.

Is this something that the behaviour change I proposed doing update_all by default and letting already local routines succeed, would fix?

sergisiso (changes requested):
@arporter this changes makes the "module-inlining" simpler and safer, well done.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: I was going to say, "oh yes, I'll fix it" but then I discovered that none of our test kernels have `only` clauses on their `use` statements :-( I'll put what I've found in #3144.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes, good spot.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Good point. However, I'm a bit worried we'll end up with very long routine names if we add `_psyclone_internal_copy`. How about `_psy_copy`?

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: It can and I have...

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Added.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: That seemed like a good idea so I have gone with it.  I've added an 'info'-level logging message if we attempt to copy in a routine that is already in local scope.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Done.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
arporter: Oh yes, good spot. Done modulo choosing how we name our copies.

tutorial/training/gocean/2.10-GameOfLife-mpi/solution/openmp_combined.py
arporter: It would fix it, yes, but I don't think there's any need to keep these scripts the way they were.

The comment here is correct in that `fuse_trans` does make local copies of the routines and then fuses. Probably I should rename it.

arporter: I've renamed the (multiple copies) of this script now.

arporter:
Compilation of lfric/eg13 failed because it can't find the kernel source file - presumably because we no longer remove the associated use statement?

arporter:
Ready for another look now @sergisiso - we also need to agree on a naming scheme for the private copies of routines.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -330,117 +314,56 @@ def _prepare_code_to_inline(
...
-            f"old_{name}",
-            symbol_type=RoutineSymbol,
-            interface=ImportInterface(
-                csym, orig_name=name))
-
-    def apply(self, node, options=None, **kwargs):
+    def apply(self,
+              node: Union[CodedKern, Call],
+              options: dict[str, Any] = None,
+              update_all: Optional[bool] = True,
```
sergisiso: The argument is optional, but the type is not Optional, its just bool

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
sergisiso: Thanks, there is a bunch of docs that need updating now:
```
rg "\[module_inline"
doc/user_guide/psyke.rst
115:              0: CodedKern testkern_code_w2_only(f3,f2) [module_inline=False]
119:              0: CodedKern testkern_wtheta_code(f4,f5) [module_inline=False]
123:              0: CodedKern testkern_code(scalar,f1,f2,f3,f4) [module_inline=False]
146:                      0: CodedKern testkern_code_w2_only(f3,f2) [module_inline=False]
150:              0: CodedKern testkern_wtheta_code(f4,f5) [module_inline=False]
154:              0: CodedKern testkern_code(scalar,f1,f2,f3,f4) [module_inline=False]
180:                      0: CodedKern testkern_code_w2_only(f3,f2) [module_inline=False]

doc/user_guide/profiling.rst
291:                                           cv_fld,h_fld,tdt,dy) [module_inline=False]
299:                                           cu_fld,h_fld,tdt,dy) [module_inline=False]
307:                                           cv_fld,tdt,dx,dy) [module_inline=False]
328:                                        cv_fld,h_fld,tdt,dy) [module_inline=False]
339:                                        cu_fld,h_fld,tdt,dy) [module_inline=False]
350:                                        cu_fld,cv_fld,tdt,dx,dy) [module_inline=False]
372:                                                ...) [module_inline=False]
384:                                                ...) [module_inline=False]
396:                                                ...) [module_inline=False]
547:                                     [module_inline=False]

tutorial/practicals/LFRic/distributed_memory/1_distributed_memory/README.md
86:             0: CodedKern scaled_matrix_vector_code(grad_p,p,div_star,hb_inv) [module_inline=False]
166:            0: CodedKern scaled_matrix_vector_code(grad_p,p,div_star,hb_inv) [module_inline=False]

doc/developer_guide/dependency.rst
265:    ... ... CodedKern testkern_stencil_vector_code(f1,f2) [module_inline=False]
302:         CodedKern testkern_stencil_vector_code(f1,f2) [module_inline=False]

tutorial/training/gocean/2.6-GameOfLife-fuse/README.md
81:                        0: CodedKern count_neighbours_code(neighbours,current) [module_inline=True]
86:                        0: CodedKern compute_born_code(born,current,neighbours) [module_inline=True]
111:                    0: CodedKern count_neighbours_code(neighbours,current) [module_inline=True]
114:                    0: CodedKern compute_born_code(born,current,neighbours) [module_inline=True]
145:                        0: CodedKern count_neighbours_code(neighbours,current) [module_inline=True]
148:                        0: CodedKern compute_born_code(born,current,neighbours) [module_inline=True]
163:                    0: CodedKern count_neighbours_code(neighbours,current) [module_inline=True]
164:                    1: CodedKern compute_born_code(born,current,neighbours) [module_inline=True]

examples/lfric/README.md
251:            CodedKern matrix_vector_kernel_code(m_lumped,ones,mb) [module_inline=False]
```

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -474,6 +397,9 @@ def apply(self, node, options=None, **kwargs):
             interface_sym = codes_to_inline[0].symbol_table.lookup(
                 external_callee_name)
 
+        if self._target_is_local(node):
+            return
```
sergisiso, line 408: Maybe this should also be called during validation to skip the remainder of the validation.

E.g. If we attempt to module-inline a subroutine which has one of out limitations, but it is already a local subroutine, it should succeed (the user still has a module-inlined routine afterwards)

Also the validation of subsequent module-inline kernels may be slighly faster with this on top.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
sergisiso: I changed my mind and prefer defering the decision to when we decide the semantics in #2683

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -490,156 +416,63 @@ def apply(self, node, options=None, **kwargs):
...
-                    if call.routine.symbol is sym_in_ctr:
-                        call.routine.symbol = target_sym
+        for code_to_inline in updated_routines:
+            # Create a new name for the routine.
+            new_name = code_to_inline.name+"_inlined_"
+            new_sym = container.symbol_table.new_symbol(
+                new_name, symbol_type=RoutineSymbol)
+            new_sym.copy_properties(code_to_inline.symbol,
+                                    exclude_interface=True)
+            new_sym.visibility = Symbol.Visibility.PRIVATE
```
sergisiso, line 443: Maybe comment:
`# Don't expose the new symbol externally to prevent unexpected collisions`

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -490,156 +416,63 @@ def apply(self, node, options=None, **kwargs):
...
+            # then it must be because the target of the call is
+            # renamed on import.
+            target_sym = name_map.get(external_callee_name)
+
+        for call in all_calls:
+            name = call.routine.symbol.name.lower()
+            if name == caller_name:
+                if isinstance(node, Call):
+                    call.routine.symbol = target_sym
+                else:
```
sergisiso, line 480: Add comment : `# Otherwise is a CodedKern`

sergisiso (changes requested):
Thanks for updating the update_all behaviour @arporter . Again, I think this is a really good step for ModuleInlining. There is just minor details now.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: OK.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes, thanks. Gone now.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oops.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes, done.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Yes, done.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Yep, done.

arporter:
CI permitting, this is ready for another look now.

sergisiso (approved):
Thanks for addressing all comments @arporter , the last IT run build LFRic with the typical build time, so this is approved for merging.

Issue #1013, opened 2020-11-27
Refactor CodedKernel rename_and_write method


sergisiso:
There are several issue with the rename_and_write() implementation:
- As the name implies, it couples together the renaming (possible versioning) and writing operations. This is problematic when module-inline also needs versioning because to get the suffix the current implementation just looks at the files already created in the output directory, so for a module-inline modified in different ways to work properly it has to create dummy files to keep track of the suffixes already used.
- It also doesn't check if previous files are exact match of the new version of the kernel, creating more versions than really are needed (although that is a minor problem).
- It decides itself the file extension (e.g. '.f90', '.cl'), this will have to grow when adding more languages, but maybe this should be given by the back-end instead?
- The logic is implemented inside of the PSyIR CodedKernel AST node, which is an odd place to have all the IO operations. Also the _rename_ast part has to rename of symbols in the whole tree, it seems something more global than the inner Kern node should do this operation.

arporter:
I've just run into this in #1021 as we are now outputting derived-type definitions and that means that the original kernel metadata gets written out. This then needs modifying if the kernel name has been changed. In that PR I've just done a temporary fix (reviewer permitting) but I agree this needs handling properly.

arporter:
#1892 (for #1824) updates get_kernel_schedule() so that GOcean-specific PSyIR is created. In turn (thanks to Rupert's new metadata functionality) this means the Container now has a proper metadata object and it is very easy to update it properly. I've therefore been able to remove the temporary fix that I had to add as per my previous comment.

arporter:
Slightly over 3 years on from this Issue's creation, I think it's time we tackled it. To my mind, the key problem we need to solve is how to choose a new name for a particular Container. We should probably also support renaming a particular `RoutineSymbol` or `GenericInterfaceSymbol` within a `Container` so that we are able to transform a single call of a particular `Routine` in the presence of other calls of that `Routine`.

A `Container` (module) name needs to be unique in the namespace of the application that is being compiled. In LFRic we can guarantee that we are creating a unique name thanks to the naming convention in use. In the absence of a naming convention, we will need to know the names of all modules (Containers) in an application which I think we can obtain from @hiker's `ModuleManager`. We could perhaps have an application-wide SymbolTable that contained all ContainerSymbols used by the application. We could then use the SymbolTable's existing functionality for generating unique names.

arporter:
[...]

Issue #1015, opened 2020-11-28
modify lfric tutorial example


rupertford:
The scaled matrix vector code in the lfric tutorial was taken from the redundant computation example and has the matrix replaced with a field. This did not affect the tutorial. However, once #901 is on master the openacc example will fail as the declarations do not match the metadata specification. Therefore (in #901) the tests in the makefile have been temporarily commented out. The metadata needs to be put back how it should be, the tutorial needs to be updated appropriately and the makefile tests need to be uncommented.

rupertford:
Created branch [1015_fix_lfric_tutorial_example](https://github.com/stfc/PSyclone/tree/1015_fix_lfric_tutorial_example)

PR #3294, opened 2026-01-21
(Closes #3334 #1013) Permit only module-inlined Kernels to be transformed

### LINKED ISSUE #3334: Make module inlining a required precursor to kernel transformations
Currently it is possible to transform a PSyKAl kernel implementation and have PSyclone create a new Fortran source file for the modified (and renamed) version. This has always been a bit clunky and causes problems for build systems which are not generally expecting new Fortran source files to appear during a build.

Now that the module-inlining functionality has matured, we can do away with this behaviour by requiring that any kernel that is to be transformed be module-inlined into the module containing the PSy-layer subroutine. This will then allow us to remove the machinery for renaming and outputting a modified kernel. Consequently, this will also affect the command-line flags and the documentation.

arporter:
Require that all kernel transformations act only on module-inlined Kernels.
Do away with the 'rename-and-write' functionality currently used for modified kernels.
We have to keep the `-okern` command-line flag as that is used when creating OpenCL versions of Kernels.

### DIFF
--- a/src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
+++ b/src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
@@ -538,9 +538,9 @@ def test_module_inline_apply_polymorphic_kernel_in_multiple_invokes(tmpdir):
     # throughout the whole module use the newly-inlined version.
     assert """subroutine invoke_1(scalar_r_bl, field_r_bl, \
 operator_r_def, f1, f2, m1, a, m2, istp, qr)
-    use function_space_mod, only : basis, diff_basis
     use quadrature_xyoz_mod, only : quadrature_xyoz_proxy_type, \
-quadrature_xyoz_type""" in output
+quadrature_xyoz_type
+    use function_space_mod, only : basis, diff_basis""" in output
     assert "mixed_kernel_mod" not in output
     assert LFRicBuild(tmpdir).code_compiles(psy)
 
--- a/src/psyclone/lfric.py
+++ b/src/psyclone/lfric.py
@@ -1799,13 +1799,14 @@ class LFRicCMAOperators(LFRicCollection):
                 f"{op_name}_{suffix}")
             tag = f"{op_name}:{suffix}"
             arg = self._cma_ops[op_name]["arg"]
-            precision = LFRicConstants().precision_for_type(arg.data_type)
+            # Ensure that the appropriate precision symbol exists.
+            kind_sym = self.symtab.add_lfric_precision_symbol(arg.precision)
             array_type = ArrayType(
-                LFRicTypes("LFRicRealScalarDataType")(precision),
+                LFRicTypes("LFRicRealScalarDataType")(kind_sym),
                 [ArrayType.Extent.DEFERRED]*3)
             index_str = ",".join(3*[":"])
             dtype = UnsupportedFortranType(
-                f"real(kind={arg.precision}), pointer, "
+                f"real(kind={kind_sym.name}), pointer, "
                 f"dimension({index_str}) :: {new_name} => null()",
                 partial_datatype=array_type)
             self.symtab.new_symbol(new_name,
@@ -3021,10 +3022,12 @@ class LFRicBasisFunctions(LFRicCollection):
                     dims.append(Literal(value, INTEGER_TYPE))
                 except ValueError:
                     dims.append(Reference(self.symtab.find_or_create(value)))
+            kind_sym = self.symtab.add_lfric_precision_symbol("r_def")
+            arr_type = ArrayType(ScalarType(ScalarType.Intrinsic.REAL,
+                                            Reference(kind_sym)), dims)
             arg = self.symtab.find_or_create_tag(
                 basis, symbol_type=DataSymbol,
-                datatype=ArrayType(LFRicTypes("LFRicRealScalarDataType")(),
-                                   dims))
+                datatype=arr_type)
             arg.interface = ArgumentInterface(ArgumentInterface.Access.READ)
             self.symtab.append_argument(arg)
 
@@ -3037,11 +3040,9 @@ class LFRicBasisFunctions(LFRicCollection):
                 raise InternalError(
                     f"Quadrature shapes other than {supported_shapes} are not "
                     f"yet supported - got: '{shape}'")
-            kind_sym = self.symtab.find_or_create(
-                const.QUADRATURE_TYPE_MAP[shape]["kind"],
-                symbol_type=DataSymbol, datatype=UnresolvedType(),
-                interface=ImportInterface(
-                    self.symtab.lookup("constants_mod")))
+
+            kind_sym = self.symtab.add_lfric_precision_symbol(
+                const.QUADRATURE_TYPE_MAP[shape]["kind"])
 
             # All quatratures are REAL
             intr_type = ScalarType(ScalarType.Intrinsic.REAL,
@@ -3106,6 +3107,8 @@ class LFRicBasisFunctions(LFRicCollection):
         # We need BASIS and/or DIFF_BASIS if any kernel requires quadrature
         # or an evaluator
         if self._qr_vars or self._eval_targets:
+            # Quadrature weights and basis functions are in r_def precision.
+            self.symtab.add_lfric_precision_symbol("r_def")
             module = self.symtab.find_or_create(
                 const.FUNCTION_SPACE_TYPE_MAP["function_space"]["module"],
                 symbol_type=ContainerSymbol)
@@ -3205,6 +3208,7 @@ class LFRicBasisFunctions(LFRicCollection):
             # to evaluate basis/diff-basis functions
             nodes_name = "nodes_" + fspace.mangled_name
             kind = api_config.default_kind["real"]
+            self.symtab.add_lfric_precision_symbol(kind)
             symbol = self.symtab.new_symbol(
                 nodes_name, symbol_type=DataSymbol,
                 datatype=UnsupportedFortranType(


examples/gocean/eg4/acc_transform.py
```diff
@@ -77,7 +77,7 @@ def trans(psyir):
     # Convert any accesses to imported data into kernel arguments, put an
     # 'acc routine' directive inside, and module-inline each kernel
     for kern in schedule.coded_kernels():
+        itrans.apply(kern)
```
sergisiso: Could you rename it to mod_inline_trans (to prevent confusions with i(nline_)trans)

src/psyclone/domain/common/transformations/kernel_transformation_mixin.py
```diff
@@ -0,0 +1,97 @@
...
+
+        """
+        if not isinstance(node, CodedKern):
+            return
+
+        msg_text = (f"Cannot transform this Kernel call to '{node.name}' "
+                    f"because")
+
+        rsymbol = node.scope.symbol_table.lookup(node.name, otherwise=None)
+        if not rsymbol or rsymbol.is_import or rsymbol.is_unresolved:
```
sergisiso: maybe: `if not rsymbol or not rsymbol.is_modulevar:` ?

src/psyclone/tests/domain/gocean/transformations/globalstoargs_test.py
```diff
@@ -84,76 +84,28 @@ def test_kernelimportsstoargumentstrans_no_outer_module_import():
     kernel = invoke.schedule.coded_kernels()[0]
     with pytest.raises(TransformationError) as err:
         trans.validate(kernel)
-    assert ("contains accesses to 'alpha' which is declared in the callee "
-            "module scope." in str(err.value))
+    assert ("Cannot transform this Kernel call to 'kernel_with_global_code' "
+            "because it is not module inlined" in str(err.value))
```
sergisiso: I am wondering if we shouldn't assume that users know what "module inline" means here, as we use it but it has been confusing before. Maybe the error could be:
"Cannot transform this kernel because its implementation resides in a different file, apply the KernelModuleInlineTrans first to bring it to this module".

sergisiso: Oh looking at KernelTransformationMixin message it is different. Do we double check this condition?

src/psyclone/tests/psyGen_test.py
```diff
@@ -662,126 +660,6 @@ def test_codedkern_node_str():
     assert expected_output in out
 
 
-def test_codedkern_module_inline_getter_and_setter():
```
sergisiso, line 665: The getter still exists, do we want to keep a test for it?

sergisiso (changes requested):
@arporter I am really happy to see that we are getting rid of rename_and_write.

This is an important change with implications to the CLI and output behaviour. I would prefer that you create an issue for it instead of being just a towards a meta-issue with many other things.

>  Now that I write that, this probably means the user guide needs to be updated and I'm pretty sure I haven't done that.

This. But also looking at the documentation, can we delete the "--kernel-renaming" option from generator.py?

> Now that we require a kernel to be module-inlined before it is transformed,

Not necessarily, LFRic can now ask psyclone to process kernel files (it is essentially what 'transmute' is). It is more "now we require kernels specifically modified for an invoke to be module-inlined first". I don't know if it belongs to our documentation, but if we do we want to be clear about it.

examples/gocean/eg4/acc_transform.py
arporter: Done.

src/psyclone/domain/common/transformations/kernel_transformation_mixin.py
arporter: I'll try that. I've seen some weird errors with this part anyway.

src/psyclone/tests/domain/gocean/transformations/globalstoargs_test.py
arporter: The full message is actually longer than the string the test checks for. However, I like your suggested words so I've changed it and updated the test too.

src/psyclone/tests/psyGen_test.py
arporter: Yes, I've put one back.

arporter:
I've now updated all the docs and have removed the "kernel-renaming" command-line option.
Ready for another look. I also realised we still had the `CodedKern._modified` property so I've removed that too.

src/psyclone/psyir/transformations/omp_declare_target_trans.py
```diff
@@ -106,9 +110,6 @@ def apply(self, node, options=None):
         self.validate(node, options)
 
         if isinstance(node, Kern):
-            # Flag that the kernel has been modified
-            node.modified = True
```
sergisiso, line 110: There is one more use in:
```
tutorial/practicals/LFRic/single_node/3_sequential/matvec_opt.py
67:            kernel.modified = True
```

Which made me realise this example will also need the ModuleInlining (and the ` kernel_schedule.walk(IntrinsicCall)` is also broken because kernel.get_callees() returns a list now)
(also the Makefile commands have the `--psykal-dsl lfric` missing)

sergisiso: Grepping Matmul2CodeTrans, also examples/lfric/eg15/matvec_opt.py may be missing ModuleInling

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
```diff
@@ -536,8 +540,7 @@ def test_module_inline_apply_polymorphic_kernel_in_multiple_invokes(tmpdir):
 operator_r_def, f1, f2, m1, a, m2, istp, qr)
     use function_space_mod, only : basis, diff_basis
     use quadrature_xyoz_mod, only : quadrature_xyoz_proxy_type, \
-quadrature_xyoz_type
-    use testkern_qr_mod, only : testkern_qr_code""" in output)
+quadrature_xyoz_type""" in output)
```
sergisiso: pylint says "superfluous-parens"

src/psyclone/domain/common/transformations/kernel_transformation_mixin.py
```diff
@@ -0,0 +1,100 @@
...
+
+        :raises TransformationError: if the supplied Kernel has not been
+            module inlined (is not present in the current Container).
+        :raises TransformationError: if the supplied Kernel (call) is not
+            within a Container or the Container does not contain the
+            implementation of the Kernel.
+
+        """
+        if not isinstance(node, CodedKern):
+            return
```
sergisiso: When I first saw that I though this is to let Builtins pass, but then I found somewhere that this is provided with a call_node as returns None. But then I realised there is nothing here specific to kernels. If this a transfomration is provided a call and its callee is being modified, the same check must be done. Maybe it's worth it doing this generic: `class CalleeTransformationMixin` and `_check_callee_implementation_is_local` ?

Then I would do this more strict and not return None for everything else, if somewhere a builtin is valid I would do:
```
if not isinstance(node, Builtin):
    node.check_callee_implementation_is_local(node)
```

What do you think?

(I have not look everywhere this is used, it could be good to check is applicable everywhere before doing the change)

sergisiso: Oh I also see that the KernelModuleInlineTrans that the error message suggest is only applicable to kernels. So maybe this can be left with a TODO together with the plans to make KernelModuleInlineTrans generic which we have discussed in the past (maybe we could repurpose #2683 to include this?)

(I would still rename the method to `check_callee_implementation_is_local` and make it stricter now)

src/psyclone/domain/gocean/transformations/gocean_opencl_trans.py
```diff
@@ -198,15 +198,17 @@ def validate(self, node, options=None):
...
-                if global_variables.difference(prec_symbols):
-                    names = sorted([sym.name for sym in
-                                    global_variables.difference(prec_symbols)])
+                global_variables = set(sym.name for sym in
+                                       ksched.symbol_table.imported_symbols)
+                prec_sym_names = set(sym.name for sym in
+                                     ksched.symbol_table.precision_datasymbols)
+                non_prec_vars = global_variables.difference(prec_sym_names)
+                if non_prec_vars:
+                    names = sorted(non_prec_vars)
```
sergisiso: Why was this change needed?

src/psyclone/tests/domain/gocean/transformations/globalstoargs_test.py
```diff
@@ -84,76 +84,29 @@ def test_kernelimportsstoargumentstrans_no_outer_module_import():
...
-    assert ("'kernel_with_use_code' contains accesses to 'rdt' which is "
-            "unresolved" in str(err.value))
-
-
-@pytest.mark.xfail(reason="Transformation does not set modified property "
-                   "of kernel - #663")
-@pytest.mark.usefixtures("kernel_outputdir")
-def test_kernelimportstoargumentstrans(monkeypatch):
+def test_kernelimportstoargumentstrans(monkeypatch, fortran_writer):
     ''' Check the KernelImportsToArguments transformation with a single kernel
```
sergisiso, line 93: The monkeypatch argument is unused now

src/psyclone/tests/domain/gocean/transformations/globalstoargs_test.py
```diff
@@ -84,76 +84,29 @@ def test_kernelimportsstoargumentstrans_no_outer_module_import():
...
     invoke and an imported variable.'''
     from psyclone.psyGen import Argument
```
sergisiso, line 95: It was already here, but can not this be at the top?

src/psyclone/tests/domain/gocean/transformations/globalstoargs_test.py
```diff
@@ -162,55 +115,58 @@ def set_to_real(variable):
...
+    assert "use model_mod, only : magic, rdt" in generated_code
+    assert ("call kernel_with_use_code(i, j, oldu_fld, cu_fld%data, "
+            "cu_fld%grid%tmask, rdt, magic)" in generated_code)
     assert invoke.schedule.symbol_table.lookup("model_mod")
     assert invoke.schedule.symbol_table.lookup("rdt")
 
 
 @pytest.mark.usefixtures("kernel_outputdir")
-def test_kernelimportstoargumentstrans_constant(monkeypatch):
+def test_kernelimportstoargumentstrans_constant(monkeypatch, fortran_writer):
```
sergisiso, line 167: pylint says "unused variable psy", can you replace it with _

sergisiso: Also can you move the import outside

src/psyclone/tests/domain/gocean/transformations/gocean_move_iteration_boundaries_inside_kernel_trans_test.py
```diff
@@ -145,7 +150,7 @@ def test_go_move_iteration_boundaries_inside_kernel_trans():
     # - It has the boundary symbol as kernel arguments
     assert isinstance(kschedule.symbol_table.lookup("xstart").interface,
                       ArgumentInterface)
-    assert isinstance(kschedule.symbol_table.lookup("xstop").interface,
+    assert isinstance(kschedule.symbol_table.lookup("xstop_1").interface,
```
sergisiso, line 153: Why has this changed? I am a bit worried about this one because if the call had argument names, this would not match anymore.

src/psyclone/tests/domain/gocean/transformations/gocean_opencl_trans_test.py
```diff
@@ -73,6 +74,12 @@ def setup():
     Config._instance = None
 
 
+@pytest.fixture(name="mod_inline_trans")
+def make_mod_inline_trans() -> KernelModuleInlineTrans:
+    '''Creates and returns a KernelModuleInlineTrans transformation.'''
+    return KernelModuleInlineTrans()
```
sergisiso, line 80: This seems unnecessary

doc/user_guide/psyclone_command.rst
```diff
@@ -56,7 +56,6 @@ by the command:
                     [-l {off,all,output}] [-p {invokes,routines,kernels}]
                     [-o OUTPUT_FILE] [-api DSL] [-oalg OUTPUT_ALGORITHM_FILE] [-opsy OUTPUT_PSY_FILE]
                     [-okern OUTPUT_KERNEL_PATH] [-dm] [-nodm]
-                    [--kernel-renaming {multiple,single}]
```
sergisiso, line 59: Also clean these please:
```
582:two different kernel-renaming schemes: "multiple" and "single"
583:(specified via the ``--kernel-renaming`` command-line flag). In the

doc/user_guide/getting_going.rst
179:                    [--kernel-renaming {multiple,single}]
223:      --kernel-renaming {multiple,single}
```

src/psyclone/psyGen.py
```diff
@@ -1315,38 +1317,14 @@ def dag_name(self):
         return f"kernel_{self.name}_{position}"
 
     @property
-    def module_inline(self):
-        '''
-        :returns: whether or not this kernel is being module-inlined.
-        :rtype: bool
+    def module_inline(self) -> bool:
```
sergisiso, line 1325: Maybe another TODO for 2054 saying that it subclasses Call, this probably should live in the super class

src/psyclone/transformations.py
```diff
@@ -2451,8 +2434,7 @@ def apply(self, node, options=None):
                 # If we have a new symbol then we must update the symbol table
                 if updated_sym is not imported_var:
                     kernel.symbol_table.swap(imported_var, updated_sym)
-
-            if updated_sym in kernel.symbol_table.precision_datasymbols:
+            if updated_sym.name.lower() in precision_sym_names:
```
sergisiso, line 2438: Why was this changed?

src/psyclone/psyir/transformations/omp_declare_target_trans.py
arporter: I see from https://github.com/stfc/PSyclone/issues/1015 that this has been broken since 2020. Perhaps we should just remove this material in favour of @hiker's recent training material?

arporter: Oh, eg15 shows up a bit of a problem: there's nothing stopping a user getting the kernel schedule and then applying transformations to it themselves. However, those transformations will now have no effect as the modified PSyIR will never be written to file :-(

arporter: Anyway, I've now fixed-up the example and the tutorial.

src/psyclone/tests/domain/common/transformations/kernel_module_inline_trans_test.py
arporter: Fixed.

src/psyclone/domain/common/transformations/kernel_transformation_mixin.py
arporter: Good points. I have generalised it, renamed it and moved it into `psyir.transformations`. In some places a `Routine` is valid so I've added checks for those.

src/psyclone/domain/gocean/transformations/gocean_opencl_trans.py
arporter: `SymbolTable.precision_datasymbols` collects the Symbols referred to  by the `precision` property (if any) of all Symbols in the table. This problem must have arisen because we have added new symbols that refer to a precision symbol from another table. I will investigate as it would be better if those references were updated.

arporter: Probably we're missing a call to `update_symbols_using` or similar.

arporter: I have 'fixed' this within KernelModuleInlineTrans but I'm still not happy - I suspect the problem is originating earlier so I will do some further investigation.

src/psyclone/psyGen.py
arporter: Yes, done.

PR #3299, opened 2026-01-24
3297 fix intermittent test failures

### LINKED ISSUE #3297: Tests might fail (randomly and rarely).
I got:
```
       raise FileNotFoundError(f"Could not find source file for module "
                                f"'{module_name}' in any of the directories "
>                               f"'{', '.join(self._original_search_paths)}'. "
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                f"You can add search paths using the '-d' "
                                f"command line option.")
E       TypeError: sequence item 4: expected str instance, LocalPath found
```
in the lfric_extration_test (during testing with compilation). 

I managed to reproduce this issue by running:
```
pytest -s ./domain/common/transformations/kernel_module_inline_trans_test.py    ./domain/lfric/transformations/lfric_extract_test.py
```
It appears that the line 1403 of `teests/domain/common/transformations/kernel_module_inline_trans_test.py`
```
    ModuleManager.get().add_search_path([tmpdir])
```
adds a `LocalPath` - which is fine, unless an error message is raised and the LocalPath is used in a `".".join(...)`.

All in all, there would be the following issues to fix - each one of them will likely fix this crash, but all should make PSyclone more stable:
- [ ] In the join statement, use `[str(i) for i in self._original_search_paths]` - this works for strings and paths
- [ ] either add type checking when adding search path (I would suggest we accept `str` and `Path`, and verify that the later works as expected - I assume long term we should try to use `Path` I assume??)
- [ ] In the above test replace `tmpdir` with `tmp_path` (which is a Path and the recommended modern version I believe)

hiker:
This fixed #3297:
- I added support to add paths as strings or Paths (turns out the actual bug did not use a `Path`, but a `LocalPath`, which is a different thing :) ). I left Path support in (assuming that we will try to replace the current string handling with Paths at some stage??) - but internally the Path is immediately converted to a string. And I added a type check for directories.
- Added tests for the Path handling, including the original bug (a non-string in a directory triggers a failure when a 'module not found' error is raised)
- I fixed the test to use `Path` (instead of `LocalPath`) - basically replacing `tmpdir` with `tmp_path`. Additionally, I reset the module manager after this test.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   32) PR #3299 for #3297. Fix intermittent test failures by more consistent
+   use of str and Path in the ModuleManager.
+ 
    31) PR #3101 for #1312. Implements support for ScalarArray arguments to
    LFRic kernels.
  


src/psyclone/parse/module_manager.py
```diff
@@ -210,23 +211,32 @@ def resolve_indirect_imports(self, value: Union[bool, Iterable[str]]):
...
 
         '''
-        if isinstance(directories, str):
+        if not isinstance(directories, list):
             # Make sure we always have a list
             directories = [directories]
 
         for directory in directories:
+            if not isinstance(directory, (str, Path)):
+                raise TypeError(f"ModuleManager.add_search_path expects a "
```
arporter, line 236: Please add a :raises ... to the docstring

src/psyclone/tests/psyir/backend/fortran_unsupported_declns_test.py
```diff
@@ -105,7 +104,7 @@ def test_fw_unsupported_interface_decln(tmpdir, fortran_writer):
     # Routines add their own symbols into the Container's symbol table
     container.addchild(Routine.create("eos1d", eos1d_table, []))
     container.addchild(Routine.create("eos2d", SymbolTable(), []))
-    assert Compile(tmpdir).string_compiles(fortran_writer(container))
+    assert Compile(tmp_path).string_compiles(fortran_writer(container))
```
arporter, line 107: I see that Compile() says that the supplied path is a `LocalPath`. Please could you update its docstring to allow for `Path` too.

src/psyclone/parse/module_manager.py
hiker: Oops - done.

src/psyclone/tests/psyir/backend/fortran_unsupported_declns_test.py
hiker: Now that is annoying. First of all, we had two type declarations, old-style (`LocalPath`), and new style (`[str|path]`). I have remove the old and incorrect one. But ... I could not figure out how to add `LocalPath` properly to the new style (since it is still widely used  .... 177 times I believe :) ). It is supposed to be either imported from `py._path`, or `py.path`, and various variations of this ... none of them worked. Even just using `"LocalPath"` as string was not accepted by mypy.

So, I gave up :(  (adding `py` as dependency, which apparently might be a solution seems not to be ideal :) ). Any ideas? I'd say to postpone till we have a typing linter, and then fix this later (or I can replace all 177 existing `tmpdir` usages :cry:)

arporter: Yes, I say leave it for now. We ought to have a new policy to update `tmpdir` to `tmp_path` on a you-touch-it you-fix-it basis.

PR #3301, opened 2026-01-26
DataNodeExtractTrans implementation

LonelyCat124:
Initial implementation - transformation is quite straightforward so should be good for a review assuming CI is ok.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   44) PR #3301 towards #1646. Adds new DataNodeToTempTrans() that allows
+   an expression to be replaced with a temporary variable.
+ 
    43) PR #3326 for #3323 - bug fix for comments before directives on
    loops or if statements.
  


src/psyclone/psyir/transformations/__init__.py
```diff
@@ -181,4 +184,5 @@
     "OMPDeclareTargetTrans",
     "OMPCriticalTrans",
     "OMPParallelTrans",
+    "DataNodeExtractTrans",
```
arporter: I think the doc generation works better if we don't put the entry here but instead have a `__all__` at the end of the file containing the class implementation. (We can keep the earlier import though since that makes it available from the `transformations` module.)

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -0,0 +1,157 @@
...
+        self.validate_options(**kwargs)
+
+        if isinstance(node.datatype, UnresolvedType):
+            raise TransformationError(
+                f"Input node's datatype cannot be computed, so the "
+                f"DataNodeExtractTrans cannot be applied. Input node was "
+                f"'{node.debug_string().strip()}'."
+            )
+
+        dtype = node.datatype
```
arporter, line 121: Please mv up to L93 so that we can re-use the result rather than re-computing.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -0,0 +1,157 @@
...
+                if element in [ArrayType.Extent.DEFERRED,
+                               ArrayType.Extent.ATTRIBUTE]:
+                    raise TransformationError(
+                        f"Input node's datatype is an array of unknown size, "
+                        f"so the DataNodeExtractTrans cannot be applied. "
+                        f"Input node was '{node.debug_string().strip()}'."
+                    )
+
+    def apply(self, node: DataNode, storage_name: str = "", **kwargs):
+        """Applies the DataNodeExtractTransApplies to the input arguments.
```
arporter: "...Trans to the input ..."

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -0,0 +1,157 @@
...
+
+        :param node: The datanode to extract.
+        :param storage_name: The name of the temporary variable to store
```
arporter: Perhaps "The (base) name of the..." to indicate that it won't necessarily be exactly that name that is used. Also, please could you de-dent the following lines to just use four spaces so we don't need as many lines.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -0,0 +1,157 @@
...
+        if isinstance(dtype, ArrayType):
+            for element in dtype.shape:
+                if element in [ArrayType.Extent.DEFERRED,
+                               ArrayType.Extent.ATTRIBUTE]:
+                    raise TransformationError(
+                        f"Input node's datatype is an array of unknown size, "
+                        f"so the DataNodeExtractTrans cannot be applied. "
+                        f"Input node was '{node.debug_string().strip()}'."
+                    )
+
```
arporter, line 192: Please add a check that `storage_name` is a `str` and raise a `TypeError` if not.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -0,0 +1,157 @@
...
+            symbol = node.scope.symbol_table.new_symbol(
+                root_name="tmp",
+                symbol_type=DataSymbol,
+                datatype=datatype
+            )
+        else:
+            symbol = node.scope.symbol_table.new_symbol(
+                root_name=storage_name,
+                symbol_type=DataSymbol,
+                allow_renaming=False,
```
arporter: How come you've set this?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -0,0 +1,157 @@
...
+                allow_renaming=False,
+                datatype=datatype
+            )
+
+        # Create a Reference to the new symbol
+        new_ref = Reference(symbol)
+
+        # Find the parent and position of the statement containing the
+        # DataNode.
+        parent = node.ancestor(Statement).parent
```
arporter, line 305: Please add a check that `node.ancestor(Statement)` is not None in `validate()`

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -0,0 +1,157 @@
...
+        :raises TypeError: if the input arguments are the wrong types.
+        :raises TransformationError: if the input node's datatype can't be
+                                     resolved.
+        :raises TransformationError: if the input node's datatype is an array
+                                     but any of the array's dimensions are
+                                     unknown.
+        """
+        # Validate the input options and types.
+        self.validate_options(**kwargs)
+
```
arporter, line 114: Please add a check that `node` is a `DataNode` and raise a TransformationError if not.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -0,0 +1,157 @@
...
+        :raises TransformationError: if the input node's datatype can't be
+                                     resolved.
+        :raises TransformationError: if the input node's datatype is an array
+                                     but any of the array's dimensions are
+                                     unknown.
+        """
+        # Validate the input options and types.
+        self.validate_options(**kwargs)
+
+        if isinstance(node.datatype, UnresolvedType):
```
arporter: Also need to check for `UnsupportedType` I think?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -0,0 +1,157 @@
...
+        pos = node.ancestor(Statement).position
+
+        # Replace the datanode with the new reference
+        node.replace_with(new_ref)
+
+        # Create an assignment to set the value of the new symbol
+        assign = Assignment.create(new_ref.copy(), node)
+
+        # Add the assignment into the tree.
+        parent.addchild(assign, pos)
```
arporter, line 315: It's really nice that this is (after a lot of hard work) so simple to do :-)

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -0,0 +1,104 @@
...
+    dtrans = DataNodeExtractTrans()
+    code = """subroutine test(a, b, c)
+        integer, dimension(:,:), intent(inout) :: a, b, c
+        c = b + a
+    end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    with pytest.raises(TransformationError) as err:
+        dtrans.validate(assign.rhs)
+    assert ("Input node's datatype is an array of unknown size, so the "
```
arporter, line 66: In this case, we know that we can query e.g. `a` for its L/UBOUND but I guess that this requires an extension to the `datatype` method which you discussed with @sergisiso elsewhere. Might be worth a comment.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -0,0 +1,104 @@
...
+        c = b + a
+    end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    with pytest.raises(TransformationError) as err:
+        dtrans.validate(assign.rhs)
+    assert ("Input node's datatype cannot be computed, so the "
+            "DataNodeExtractTrans cannot be applied. Input node "
+           "was 'b + a'" in str(err.value))
+
```
arporter, line 136: Please add a test for an UnsupportedType declaration.

arporter: Please also add tests when the transformation is given the wrong type of node and a non-str name.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -0,0 +1,104 @@
...
+        integer, dimension(10,100) :: a
+        integer, dimension(100,10) :: b
+        integer, dimension(10,10) :: c, d
+        d = c + MATMUL(a, b)
+    end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    dtrans.apply(assign.rhs.operands[1])
+    out = fortran_writer(psyir)
+    assert ("integer, dimension(SIZE(a, dim=1),SIZE(b, dim=2)) :: tmp"
```
arporter: Ah, *this* was what you discussed. I guess the same comment applies (that we *could* do better but it's up to `datatype` to do that).

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+        b = INT(a)
+        end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    dtrans.apply(assign.rhs, storage_name="temporary")
+    out = fortran_writer(psyir)
+    assert "integer :: temporary" in out
+    assert "temporary = INT(a)" in out
+    assert "b = temporary" in out
```
arporter, line 307: Please could you add at least one test with a more complicated code structure - e.g. an expression inside a loop. This will exercise the hierarchy of symbol tables.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        # Call validate to check inputs are valid.
+        self.validate(node, storage_name=storage_name, **kwargs)
+
+        # Find the datatype
+        datatype = node.datatype
+
+        # Create a symbol of the relevant type.
+        if not storage_name:
+            symbol = node.scope.symbol_table.new_symbol(
```
arporter: Talking with @sergisiso about possible use-cases for this transformation, he mentioned things like:
```fortran
   a(2:4) = 3*a(1:3)
```
where we need to explicitly store the copy of `a(1:3)` in a temporary in order to parallelise. With this transformation now we'd generate:
```fortran
  tmp = a(1:3)
  a(2:4) = tmp
```
That has made me think that we need to check that we create `tmp` with dimensions starting at 1. (I think we do because that's what `datatype` returns?)
Are there any other considerations @sergisiso?

sergisiso: > That has made me think that we need to check that we create tmp with dimensions starting at 1.

I am not sure about this, if it doesn't start at the same lbound as the lhs it will generate the offset expression.

However, impure calls could be a problem:

`a = globalvar + b(fn(4))`

If we try to move b, it could internally update 'globalvar' which would change the order of the use-update. I would not allow impure or unknown purity functions.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
sergisiso: If `datatype` does not work you can check the expression produced by `get_effective_shape`.

But this made me thing that expressions using imported symbols (even with explicit bounds) may be a problem. Do the symbol in the explicit bound is available locally? Does it have the global or the local interface?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: This is done by the validate_options function so we don't need to.

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: I was thinking the user-defined name should be honoured exactly, but I'm happy to relax that if you think that would be better.

src/psyclone/psyir/transformations/__init__.py
LonelyCat124: Changed.

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Fixed.

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: I guess we need to decide if this should be a base name of if the storage_name argument is provided the name must be exact? Happy to go either way - if the name must be exact I should probably add it to validate.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Added.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Added

LonelyCat124: with test.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Added

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Added tests for both of these cases with appropriate errors in validate. (I think the function one is currently already caught by the datatype but I was explicit in case we make improvements in the future).

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: @sergisiso Is there a good way to test this? I'm a bit unfamiliar with making codes that correctly follow imports for the test suite.

sergisiso: Yes, there are a few tests that use a code and an imported module code. See for example `src/psyclone/tests/psyir/symbols/symbol_table_test.py::test_resolve_imports_from_child_symtabs`

src/psyclone/psyir/transformations/datanode_extract_trans.py
sergisiso: Just to be clear, the problem is there even without moving the the fn()
`a = fn(3) +  b(globalvar)`
moving b still changes the order if fn also uses globalvar, and this one is not caught by datatype.

Also if the result_type of fn is known, we will have a valid datatype but still be impure.

LonelyCat124: Yeah - probably just my inexperience, I couldn't create a test that could correctly resolve the datatype of a function (but i do fail on any impure function call in the provided node). I guess what I don't currently handle is:

`a = fn(somevar) + b(globalvar)` - If b was provided as the argument (and PURE) in theory we'd happily move it which is probably incorrect (if fn is impure)? I'm not sure what the best solution to this is - probably check the statement containing the input node for impure calls?

LonelyCat124: FIXME:
Ah thats what you wrote - I'll make that change when I'm next on the branch

sergisiso: On a second thought, this will be too restrictive for the first use case: taking argument expressions out of io calls. These are always impure.

So now I think we don't need this check. Sorry for the distraction

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: This works until an array index is already an imported symbol in wherever its imported from. I need to think of a good solution for that

LonelyCat124: This works up to a point and I'm not sure how to resolve it further.

The easy thing to do is check all symbols in the node don't have a symbol with `is_unknown_interface`. The issue is for a case like this:
```
    with open(filename, "w", encoding='UTF-8') as module:
        module.write('''
        module c_mod
            use some_mod, only: i
            integer, dimension(25, i) :: some_var
        end module c_mod
        ''')
    code = """subroutine test()
        use c_mod, only: some_var
        integer, dimension(25, 50) :: b
        b = some_var
        end subroutine test"""
    psyir.children[0].symbol_table.resolve_imports()
    dtrans.apply(assign.rhs)
    out = fortran_writer(psyir)
    print(psyir.walk(Assignment)[0].rhs.symbol.shape[1].upper.symbol.interface)
    print(psyir.walk(Assignment)[0].rhs.symbol.shape[1].upper.symbol.is_unknown_interface)
```

The interface is known, but its to this `some_mod` which PSyclone can't resolve (because i made it up and its not found), but PScylone will still says `is_unknown_interface` is False. I had a look at InlineTrans to try to match that but it only seems to inquire about `is_unknown_interface` - I assume because you can safely call symbol_table.merge().

Can i just add the symbols that aren't in this module's symbol table safely if they don't collide? Or will that cause some other side effects? @sergisiso @arporter

arporter: `is_unknown_interface` is False because it has a known interface - we know where it comes from, we just can't find it.

arporter: You could add imports for them if they are already imported. Alternatively, we could fall back to SIZE?

LonelyCat124: This is awkward a bit as some errors shadow others, e.g.
```
        symbols = node.get_all_accessed_symbols()
        scope_symbols = node.scope.symbol_table.get_symbols()
        for sym in symbols:
            if sym.is_unknown_interface:
                raise TransformationError(
                    f"Input node contains a symbol with an unknown "
                    f"interface, so the DataNodeExtractTrans cannot be "
                    f"applied. Input node was "
                    f"'{node.debug_string().strip()}'."
                )
```

This shadows this other check (or whichever is first is the only one that I can have happen):
```
        if isinstance(dtype, (UnresolvedType, UnsupportedFortranType)):
            raise TransformationError(
                f"Input node's datatype cannot be computed, so the "
                f"DataNodeExtractTrans cannot be applied. Input node was "
                f"'{node.debug_string().strip()}'."
            )
```

Are both checks necessary? I think I should just keep the latter and then import symbols where possible (so I just need to check for symbol collision).

arporter: Yes, I agree that it's only the latter check that's necessary. In general, it's best *not* to have an individual transformation import symbols because that performs very badly. It's much better to get the frontend to import them when it first constructs the PSyIR (by having the module names added to RESOLVE_IMPORTS in the transformation script). Therefore, if you could update the error message to say this that would be great - you'll probably need to finesse it a bit depending on whether it is Unresolved or Unsupported.

arporter: Further, it's probably going to really help the user if we could say which Symbol(s) is/are causing the problem. Therefore, once we know we're going to raise an error, it would be worth examining `node` to get its Symbols and seeing which of them are the problem.

LonelyCat124: I need to think more about the "symbol merging" as it seems kinda hard to do - the existing symbol can be not equal but the essentially the same symbol (same module, name, type etc.) so I need to work out how to handle this.

LonelyCat124: Ah by import symbols I was only meaning adding the relevant `use x_mod: symbol` symbols when we find them when importing known symbols.

LonelyCat124: I've added some code now to handle imported symbols including those in shape declarations, I'm hoping I've not overengineered it but i'll start working on the tests for it next.

src/psyclone/psyir/transformations/datanode_extract_trans.py
arporter: I can't imagine anyone ever caring that it was exact so I vote for it just being a suggestion (i.e. base_name).

src/psyclone/psyir/transformations/datanode_extract_trans.py
arporter: As discussed, I think this can be relaxed.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -108,15 +130,58 @@ def validate(self, node: DataNode, **kwargs):
...
+                if isinstance(element.lower, DataNode):
+                    symbols.update(element.lower.get_all_accessed_symbols())
+                if isinstance(element.upper, DataNode):
+                    symbols.update(element.upper.get_all_accessed_symbols())
+                scope_symbols = node.scope.symbol_table.get_symbols()
+                for sym in symbols:
+                    scoped_name_sym = scope_symbols.get(sym.name, None)
+                    if scoped_name_sym and sym is not scoped_name_sym:
+                        raise TransformationError(
+                            f"Input node contains an imported symbol whose "
```
arporter: It took me a while to work out what this check was for. Could you extend the text to say that the "The type of the node supplied to {self.name} depends upon an imported symbol ('{sym.name}') which has a name clash with a symbol in the current scope."

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -108,15 +130,58 @@ def validate(self, node: DataNode, **kwargs):
...
+                for sym in symbols:
+                    scoped_name_sym = scope_symbols.get(sym.name, None)
+                    if scoped_name_sym and sym is not scoped_name_sym:
+                        raise TransformationError(
+                            f"Input node contains an imported symbol whose "
+                            f"name collides with an existing symbol, so the "
+                            f"DataNodeExtractTrans cannot be applied. "
+                            f"Clashing symbol name is '{sym.name}'."
+                        )
+                    # If its an imported symbol we need to check if its
```
arporter, line 176: I think we want this check as part of the previous one as we may have the same symbol imported from the same location but represented by different Symbol instances. i.e. `sym is not scoped_name_sym` but, if they are both imported from the same module, they must be representing the same variable.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -108,15 +130,58 @@ def validate(self, node: DataNode, **kwargs):
                         f"so the DataNodeExtractTrans cannot be applied. "
                         f"Input node was '{node.debug_string().strip()}'."
                     )
+                # Otherwise we have an ArrayBounds
```
arporter: I need a bit more help following this code. Perhaps, "...ArrayBounds - check to see which Symbols are referenced in those bounds."

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -108,15 +130,58 @@ def validate(self, node: DataNode, **kwargs):
                         f"so the DataNodeExtractTrans cannot be applied. "
                         f"Input node was '{node.debug_string().strip()}'."
                     )
+                # Otherwise we have an ArrayBounds
+                symbols = set()
+                if isinstance(element.lower, DataNode):
+                    symbols.update(element.lower.get_all_accessed_symbols())
+                if isinstance(element.upper, DataNode):
+                    symbols.update(element.upper.get_all_accessed_symbols())
+                scope_symbols = node.scope.symbol_table.get_symbols()
```
arporter, line 150: "Compare these symbols with those already in scope at `node`"

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -108,15 +130,58 @@ def validate(self, node: DataNode, **kwargs):
...
+                                None
+                        )
+                        if scoped_name_sym and not isinstance(
+                                scoped_name_sym, ContainerSymbol):
+                            raise TransformationError(
+                                f"Input node contains an imported symbol "
+                                f"whose containing module collides with an "
+                                f"existing symbol. Colliding name is "
+                                f"'{sym.interface.container_symbol.name}'."
+                                )
```
arporter, line 191: Unfortunately we also have to allow for a situation when the type of the imported variable (`a_var` in this case) depends on private variables:
```fortran
module some_other_mod
    integer, parameter :: dim1 = 4, dim2 = 5
    real(kind=wp), dimension(dim1, dim2) :: a_var
    public :: a_var
    private
end module
```
in which case we can't import the symbols that define the shape of the array. You'll need to check their visibility in the symbol_table associated with the Container pointed to by the import interface of `a_var`.

src/psyclone/psyir/transformations/datanode_extract_trans.py
```diff
@@ -108,15 +130,58 @@ def validate(self, node: DataNode, **kwargs):
...
+                                )
+
+        if node.ancestor(Statement) is None:
+            raise TransformationError(
+                "Input node to DataNodeExtractTrans has no ancestor "
+                "Statement node which is not supported."
+            )
+
+        if isinstance(dtype, (UnresolvedType, UnsupportedFortranType)):
+            raise TransformationError(
```
arporter: It's a bit of a faff but I think it would be very useful to the user if you could examine the symbols referenced in `node` and see if you can establish which of them has unsupported/unresolved type. It may be that they could be resolved by setting RESOLVE_IMPORTS appropriately in the transformation script.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -139,6 +204,44 @@ def apply(self, node: DataNode, storage_name: str = "", **kwargs):
...
+                if isinstance(element.lower, DataNode):
+                    symbols.update(element.lower.get_all_accessed_symbols())
+                if isinstance(element.upper, DataNode):
+                    symbols.update(element.upper.get_all_accessed_symbols())
+                scope_symbols = node.scope.symbol_table.get_symbols()
+                for sym in symbols:
+                    scoped_name_sym = scope_symbols.get(sym.name, None)
+                    # If no symbol with the name exists then create one.
+                    if not scoped_name_sym:
+                        sym_copy = sym.copy()
```
arporter, line 256: I hadn't understood what you meant by "import symbols" before but now I do. I guess the premise here is that if a symbol appears in the definition of an *imported* symbol's shape, then it must be a compile-time constant? (Because if it was allocatable, we wouldn't have sizes for its dimensions.) In which case I *think* it's OK. Presumably also, we're only adding these symbols into the table associated with a Routine? (I started worrying that they might have module scope, in which case we would need to make sure they were private.)

arporter: Worth summarising this in a comment I think - possibly instead of the FIXME above?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -68,10 +76,146 @@ def test_datanodeextracttrans_validate(fortran_reader):
...
+            "'storage_name' option expects type 'str' but received '1' of "
+            "type 'int'.\nPlease see the documentation and check the "
+            "provided types." in str(err.value))
+
+    with pytest.raises(TransformationError) as err:
+        dtrans.validate(Reference(DataSymbol("a", INTEGER_TYPE)))
+    assert ("Input node to DataNodeExtractTrans has no ancestor Statement "
+            "node which is not supported." in str(err.value))
+
+    code = """module some_mod
```
arporter, line 115: This test is getting quite large :-) Please could you split it up.

src/psyclone/tests/psyir/transformations/datanode_extract_trans_test.py
```diff
@@ -68,10 +76,146 @@ def test_datanodeextracttrans_validate(fortran_reader):
...
+        integer :: a, b
+
+        a = a + some_func(a,b)
+    end subroutine test
+    end module"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[2]
+    with pytest.raises(TransformationError) as err:
+        dtrans.validate(assign.rhs)
+    assert ("Input node to DataNodeExtractTrans contains a call that is not "
```
arporter: I may have already suggested this but how about: "...a call ('some_func(a,b)`) that is not..."

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -68,10 +76,146 @@ def test_datanodeextracttrans_validate(fortran_reader):
...
+        end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    psyir.children[0].symbol_table.resolve_imports()
+    assign = psyir.walk(Assignment)[0]
+    with pytest.raises(TransformationError) as err:
+        dtrans.validate(assign.rhs)
+    assert ("Input node contains an imported symbol whose name collides "
+            "with an existing symbol, so the DataNodeExtractTrans cannot be "
+            "applied. Clashing symbol name is 'i'." in str(err.value))
+
```
arporter, line 187: As commented earlier, please could you add a similar test but have `i` imported from `some_mod` in subroutine test too.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -68,10 +76,146 @@ def test_datanodeextracttrans_validate(fortran_reader):
...
+        module.write('''
+        module some_mod
+            integer, parameter :: i = 25
+            integer, parameter :: j = 30
+        end module some_mod
+        ''')
+    code = """subroutine test()
+        use a_mod, only: some_var
+        use some_mod, only: j
+        j = some_var(1,3)
```
arporter, line 209: Picky but `j` is a parameter so can't be assigned to.

src/psyclone/tests/psyir/transformations/datanode_extract_trans_test.py
```diff
@@ -68,10 +76,146 @@ def test_datanodeextracttrans_validate(fortran_reader):
...
+    code = """subroutine test()
+        use f_mod, only: some_var
+        integer :: tmpmod
+        tmpmod = some_var
+        end subroutine test"""
+    psyir = FortranReader(resolve_modules=True).psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    with pytest.raises(TransformationError) as err:
+        dtrans.validate(assign.rhs)
+    assert ("Input node contains an imported symbol whose containing module "
```
arporter: "...symbol ('sym.name') whose ..."

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -102,3 +246,144 @@ def test_datanodeextractrans_apply(fortran_reader, fortran_writer):
...
+        integer :: i
+
+        do i = 1, 100
+          b(i) = REAL(i)
+        end do
+     end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    dtrans.apply(assign.rhs, storage_name="temporary")
+    out = fortran_writer(psyir)
```
arporter, line 321: Pls could you add checks that the code compiles. (`assert Compile(tmpdir).string_compiles(out)`)
Also, following @hiker pointing out that `tmp_path` is the replacement for `tmpdir`, please could you switch to using that?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -102,3 +246,144 @@ def test_datanodeextractrans_apply(fortran_reader, fortran_writer):
...
+    end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    assign = psyir.walk(Assignment)[0]
+    dtrans.apply(assign.rhs)
+    out = fortran_writer(psyir)
+    assert "  integer, dimension(3) :: tmp" in out
+    assert """  tmp = 3 * b
+  a(:4) = tmp""" in out
+
+    # Test the imports are handled correctly.
```
arporter, line 357: I think the following could be a separate test.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -102,3 +246,144 @@ def test_datanodeextractrans_apply(fortran_reader, fortran_writer):
...
+        use b_mod
+        integer, dimension(25, 50) :: b
+        b = some_var
+        end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    psyir.children[0].symbol_table.resolve_imports()
+    assign = psyir.walk(Assignment)[0]
+    dtrans.apply(assign.rhs)
+    out = fortran_writer(psyir)
+    assert """  integer, dimension(25,50) :: tmp
```
arporter: Compilation checks for these various checks please (if possible).

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Changed this behaviour now.

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: @arporter I did wonder about this initially, but couldn't find a way to create a test where it happened, so abandoned it, but I'll have another go since you also think it should be possible.

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Added.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: @arporter I'm a bit confused when trying to do this test - when I examine the interface of `dim1` and `dim2` (as found in the size of the a_var.`datatype`)I don't get an `ImportInterface` but instead a `StaticInterface`. Is this expected behaviour? If so, how can I differentiate between two private static variables where one is in the current scope and one is from a "resolve imports" module?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Not quite, if we have a symbol (that needs to be imported from another module and currently isn't) in the shape then we add it to the symbol table with an ImportInterface, so the output should be `use the_module, only: the_shape_symbol`. Updating the comment to explain this I think.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Yeah, I'm not sure how to tell the difference between the dim/dim2 symbols in your example, and x in this code:
```module test
  implicit none
  integer, parameter, private :: x = 1
  private

  contains
  subroutine sub()
    integer :: i

    i = x

  end subroutine sub

end module test
```

I could check that if its a static variable then it has to be in the scope? But why are those not ImportInterface symbols despite being static?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Fixed

src/psyclone/tests/psyir/transformations/datanode_extract_trans_test.py
LonelyCat124: Added

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Yeah I found a way to test this now when looking at another comment.

LonelyCat124: Done I think - at least for validate, need to think about apply.

LonelyCat124: Actually apply should be ok, but i'm a bit confused, i make symbol copies and put them in the symbol table and then just don't reference them anywhere (including not reforming the shape appropriately) so I'm a little confused.

LonelyCat124: I'm so confused by symbols every time i try to work with them, I think I need to pause and double check all the symbols being created are used where i expect afterwards and are in the symbol table.

LonelyCat124: Fixed now I think.

src/psyclone/psyir/transformations/datanode_extract_trans.py
LonelyCat124: Added, not 100% sure about the interfaces here but I think they should be sufficient.

LonelyCat124: It also won't add the extra bit to the message if the unresolved type comes from an IntrinsicCall that has an unknown datatype (for example).

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Split into timported and not imported sections.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/transformations/datanode_extract_trans_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: I added the option, but I'll need to submit integration later to see if they work, I don't seem to be able to test compilation locally.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Added

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: I think I've overthought this, if the symbol whose shape we're examining is imported, and any of the shape symbols are private we just reject it and that should be sufficient I think.

LonelyCat124: I think they should still have an ImportInterface and not a StaticInterface perhaps, but we can discuss that later.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Fixed this via already existing functionality.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
arporter: Perhaps we're (incorrectly) giving them a StaticInterface when we chase down their definition? This is a bit awkward as we will want to know that they are static symbols but also that they are imported.

arporter: Perhaps we're (incorrectly) giving them a StaticInterface when we chase down their definition? This is a bit awkward as we will want to know that they are static symbols but also that they are imported.

doc/user_guide/transformations.rst
```diff
@@ -529,6 +529,12 @@ can be found in the API-specific sections).
 
 ####
 
+.. autoclass:: psyclone.psyir.transformations.DataNodeToTempTrans
```
arporter: Sorry to be a pain but please put this in alpha-order with the other transformations.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -107,7 +114,7 @@ def validate(self, node: DataNode, **kwargs):
 
         if not isinstance(node, DataNode):
             raise TypeError(
-                f"Input node to DataNodeExtractTrans should be a "
+                f"Input node to DataNodeToTempTrans should be a "
```
arporter: I think you can use {self.name} here rather than putting in the actual transformation name.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -117,8 +124,9 @@ def validate(self, node: DataNode, **kwargs):
         for call in calls:
             if not call.is_pure:
                 raise TransformationError(
-                    f"Input node to DataNodeExtractTrans contains a call "
-                    f"that is not guaranteed to be pure. Input node is "
+                    f"Input node to DataNodeToTempTrans contains a call "
```
arporter: Same here.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -127,24 +135,46 @@ def validate(self, node: DataNode, **kwargs):
                                ArrayType.Extent.ATTRIBUTE]:
                     raise TransformationError(
                         f"Input node's datatype is an array of unknown size, "
-                        f"so the DataNodeExtractTrans cannot be applied. "
+                        f"so the DataNodeToTempTrans cannot be applied. "
```
arporter: and here.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -127,24 +135,46 @@ def validate(self, node: DataNode, **kwargs):
...
                 for sym in symbols:
                     scoped_name_sym = scope_symbols.get(sym.name, None)
+                    # If sym is not scoped_name_sym, then there is a
+                    # symbol collision from an imported symbol.
                     if scoped_name_sym and sym is not scoped_name_sym:
+                        # If the symbol in scoped is imported from the same
+                        # container then we can skip this.
+                        if (isinstance(scoped_name_sym.interface,
+                                       ImportInterface) and
+                            (scoped_name_sym.interface.container_symbol.name
```
arporter: I've had a look and I think the `ImportInterface.__eq__` method does what you want so just `scoped_name_sym.interface == sym.interface` should be fine here.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -127,24 +135,46 @@ def validate(self, node: DataNode, **kwargs):
...
                     symbols.update(element.upper.get_all_accessed_symbols())
+                # Compare the symbols in the array bounds with the symbols
+                # already in the scope.
                 scope_symbols = node.scope.symbol_table.get_symbols()
                 for sym in symbols:
                     scoped_name_sym = scope_symbols.get(sym.name, None)
+                    # If sym is not scoped_name_sym, then there is a
+                    # symbol collision from an imported symbol.
                     if scoped_name_sym and sym is not scoped_name_sym:
+                        # If the symbol in scoped is imported from the same
```
arporter: "scope"

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -157,29 +187,46 @@ def validate(self, node: DataNode, **kwargs):
...
+                                f"'{sym.name}' whose containing module "
+                                f"collides with an existing symbol. Colliding "
+                                f"name is "
                                 f"'{sym.interface.container_symbol.name}'."
                                 )
 
         if node.ancestor(Statement) is None:
             raise TransformationError(
-                "Input node to DataNodeExtractTrans has no ancestor "
+                "Input node to DataNodeToTempTrans has no ancestor "
```
arporter: `self.name` instead of "DataNodeToTempTrans" pls.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -157,29 +187,46 @@ def validate(self, node: DataNode, **kwargs):
...
+            for sym in symbols:
+                if isinstance(sym.interface, (UnresolvedInterface,
+                                              UnknownInterface)):
+                    failing_symbols.append(sym.name)
+            # Sort the order of the list to get consistant results for tests.
+            failing_symbols.sort()
+            message = (
                 f"Input node's datatype cannot be computed, so the "
-                f"DataNodeExtractTrans cannot be applied. Input node was "
+                f"DataNodeToTempTrans cannot be applied. Input node was "
```
arporter: self.name pls

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -157,29 +187,46 @@ def validate(self, node: DataNode, **kwargs):
...
                 f"'{node.debug_string().strip()}'."
             )
+            if failing_symbols:
+                message += (
+                    f" The following symbols in the input node are not "
```
arporter: Perhaps s/are not resolved in the scope:/have not been resolved by PSyclone:/

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -189,26 +236,14 @@ def apply(self, node: DataNode, storage_name: str = "", **kwargs):
...
-                datatype=datatype
-            )
-
-        # FIXME Make sure the shape is all in the symbol table. We know that
+        # Make sure the shape is all in the symbol table. We know that
         # all symbols we find can be safely added as otherwise validate will
         # fail.
-        # This is an oversimplification because we could have multiple
-        # references to the same symbol...
+        # Symbols used to reference shapes that are from imported modules but
```
arporter: "Symbols occurring within shape definitions that are" perhaps? I might be misunderstanding though as you say "symbol will reference the original ... shape" below? I don't think I understand how a symbol references a shape?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
arporter: s/the the/to the/ and I'm not sure what "and we're ok with this symbol" means?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -33,7 +33,7 @@
 # -----------------------------------------------------------------------------
 # Authors A. B. G. Chalk STFC Daresbury Lab
 
-'''This module contains the DataNodeExtractTrans class.'''
+'''This module contains the DataNodeToTempTrans class.'''
```
arporter: "tests for the ..."

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -75,8 +77,11 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
     with pytest.raises(TransformationError) as err:
         dtrans.validate(assign.rhs)
     assert ("Input node's datatype cannot be computed, so the "
```
arporter: Now that I look at this in context, how about making it a bit more user friendly:  "The datatype of the supplied expression cannot be computed so ... The input expression was ..."

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -125,12 +130,20 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
     assign = psyir.walk(Assignment)[2]
     with pytest.raises(TransformationError) as err:
         dtrans.validate(assign.rhs)
-    assert ("Input node to DataNodeExtractTrans contains a call that is not "
+    assert ("Input node to DataNodeToTempTrans contains a call "
```
arporter, line 132: Again, I think "expression" rather than "node", although, I'm not totally certain: I guess a user in this context *is* supplying a node to the transformation so perhaps it's OK?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -149,31 +162,51 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
...
+        end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    psyir.children[0].symbol_table.resolve_imports()
+    assign = psyir.walk(Assignment)[0]
+    dtrans.validate(assign.rhs)
 
     # Check validation works when the shape contains a symbol from an
     # existing module
-    filename = os.path.join(str(tmpdir), "a_mod.f90")
+    filename = os.path.join(str(tmp_path), "a_mod.f90")
```
arporter: The nice thing about `tmp_path` is that you can just do: `filename = tmp_path / "a_mod.f90"`

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -149,31 +162,51 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
...
     with open(filename, "w", encoding='UTF-8') as module:
         module.write('''
         module a_mod
             use some_mod, only: i
             integer, dimension(25, i) :: some_var
         end module a_mod
         ''')
-    filename = os.path.join(str(tmpdir), "some_mod.f90")
+    filename = os.path.join(str(tmp_path), "some_mod2.f90")
```
arporter: `tmp_path / "some_mod2.f90"`

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -184,15 +217,15 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
 
     # Check validation raise an error when the shape contains a symbol from
     # a module that overlaps with a symbol in the scope.
-    filename = os.path.join(str(tmpdir), "tmpmod.f90")
+    filename = os.path.join(str(tmp_path), "tmpmod.f90")
```
arporter: `tmp_path / "tmpmod.f90"`

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -184,15 +217,15 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
...
     with open(filename, "w", encoding='UTF-8') as module:
         module.write('''
         module tmpmod
             integer, parameter :: i = 25
             integer, parameter :: j = 30
         end module tmpmod
         ''')
-    filename = os.path.join(str(tmpdir), "f_mod.f90")
+    filename = os.path.join(str(tmp_path), "f_mod.f90")
```
arporter: and again.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -209,15 +242,40 @@ def test_datanodeextracttrans_validate(fortran_reader, tmpdir, monkeypatch):
...
     with pytest.raises(TransformationError) as err:
         dtrans.validate(assign.rhs)
-    assert ("Input node contains an imported symbol whose containing module "
-            "collides with an existing symbol. Colliding name is 'tmpmod'."
+    assert ("Input node contains an imported symbol 'i' whose containing "
+            "module collides with an existing symbol. Colliding name is "
+            "'tmpmod'."
             in str(err.value))
 
+    filename = os.path.join(str(tmp_path), "some_other_mod.f90")
```
arporter: and again.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -276,13 +337,27 @@ def test_datanodeextractrans_apply(fortran_reader, fortran_writer, tmpdir,
...
+
+    TODO #284: Compilation tests for these are blocked as the Compile
+    utility doesn't look for imported modules yet.
+    """
+    dtrans = DataNodeToTempTrans()
     # Test the imports are handled correctly.
-    monkeypatch.setattr(Config.get(), '_include_paths', [str(tmpdir)])
-    filename = os.path.join(str(tmpdir), "a_mod.f90")
+    monkeypatch.setattr(Config.get(), '_include_paths', [str(tmp_path)])
+    filename = os.path.join(str(tmp_path), "a_mod.f90")
```
arporter: `tmp_path / "a_mod.f90"`

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -304,7 +379,7 @@ def test_datanodeextractrans_apply(fortran_reader, fortran_writer, tmpdir,
   tmp = some_var
   b = tmp""" in out
 
-    filename = os.path.join(str(tmpdir), "b_mod.f90")
+    filename = os.path.join(str(tmp_path), "b_mod.f90")
```
arporter: and throughout please :-)

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -344,25 +420,27 @@ def test_datanodeextractrans_apply(fortran_reader, fortran_writer, tmpdir,
...
+  use some_mod, only : i
   integer, dimension(25,50) :: b
-  integer, dimension(25,i) :: tmp
+  integer, allocatable, dimension(:,:) :: tmp
 
+  ALLOCATE(tmp(1:25,1:i))
   tmp = some_var
   b = tmp""" in out
 
     # Check that modules in a shape from an imported module are
```
arporter: s/modules/symbols/

src/psyclone/psyir/transformations/datanode_extract_trans.py
arporter: Nice, thanks.

arporter (changes requested):
Thanks Aidan. Just small things to tidy now. I've submitted the ITs again too.

doc/user_guide/transformations.rst
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Fixed.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Yeah, done.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Yeah that is correct, updated the comment.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: The we're ok with this symbol is unnecessary, it just meant we don't need to change anything else with this symbol.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: I think I'd rather stay with node for most of these (including the previous one) as its explicit as to what a node is in the PSyIR, while expression is more of a language concept that we don't express, but I'm not sure what is better for user experience?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Done throughout.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
arporter: OK. I guess we might find out anecdotally in the future.

arporter:
All looks good now, thanks very much Aidan. I've fired off the ITs one last (hopefully) time...

PR #3309, opened 2026-01-29
(closes #3281) Increase rank in zdftke and fix solfrac issue

### LINKED ISSUE #3281: Add more file's to NEMO's increase array rank
tke_tke can be added if the validate considers as safe arrays used in an intrinsic with a 'dim' argument (as the resulting rank will always be the same - that of the given dim)

sergisiso:


src/psyclone/psyir/transformations/increase_rank_loop_arrays_trans.py
```diff
@@ -196,6 +196,14 @@ def validate(
                     # Assignments to the variable are fine, because the value
                     # we will just be repeated to each index of the new rank
                     continue
+                if (isinstance(check.parent, IntrinsicCall) and
+                        check.parent.is_inquiry and
+                        "dim" in check.parent.argument_names):
+                    # Inquiry intrinsics calls to a given 'dim' are still ok
```
arporter, line 202: Great comment, thanks :-)

Issue #3310, opened 2026-01-30
Slow dependency analysis


hiker:
While doing some performance tests for #3213, I noticed some extreme slow performance of the existing dependency analysis - below the runtime in seconds of `can_loop_be_parallelised` for two ukca files:
```
File name                   loop#                           DepTools    parallel? 

ukca_um_strat_photol_mod.F90	2	Execution time loop i:	11.899324	True	
ukca_um_strat_photol_mod.F90	3	Execution time loop j:	6.690157	True	
ukca_aerod.F90					1	Execution time loop j:	38.493939	True	
```
One reason for this are rather expensive comparisons of identical terms (`j == j`) - the dependency analysis goes all the way to convert this to a sympy expression, solves `j+dj == j` for `dj`, then with `dj=0` concludes that there is no loop carried dependency :)

Adding a shortcut for this special case improved the timings to:
```
ukca_um_strat_photol_mod.F90	2	Execution time loop i:	0.221416	True
ukca_um_strat_photol_mod.F90	3	Execution time loop j:	0.188218	True
ukca_aerod.F90	                1	Execution time loop j:	0.829583	True	
```
A speedup of between 35 and 45 (potentially more, I see some runtime variation, these are not the fastest measurements)

The patch:
```
@@ -632,9 +632,21 @@ class DependencyTools():
                 elif len(set_of_vars) == 1:
                     # One loop variable used in both accesses.
                     # E.g. `a(2*i+3) = a(i*i)`
+                    if (isinstance(index_write, Reference) and
+                        isinstance(index_other, Reference) and
+                        index_write == index_other):
+                        if index_write.name == loop_var:
+                            return True
+                        # The expression does not depend on the loop variable
+                        # at all (dependency distance would return None),
+                        # again no need for an explicit test, we can continue
+                        # the outer loop
+                        continue
```
This still passed all tests.

PR #3311, opened 2026-01-30
3310 optimise dependency tool

hiker:
Delivers significant speedup in the dependency analysis for files that have many array indices that just use a  simple reference as index (`a(i,j)`, not `a(i+1, h)`) by avoiding calling sympy in these simple cases.

In two ukca files some loops are 30 to over 40 times faster (measurements in #3310).

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   38) PR #3311 for #3310. Small optimisation to the dependence-analysis
+   functionality for array indices consisting of simple References.
+ 
    37) PR #3295 for #3292. Add command-line flag to permit config-file
    settings to be overridden and extend possible settings for the
    runtime warnings option.


hiker:
I've verified that the 'speedup' of the DA on master was caused because the references were not resolved (turned up as unknown, so DA aborted early). Resolving the symbol fixes the 'speedup' (which marked the loop as not parallelisable, even though it is), it is then slow again with the correct result, and this patch fixes it:
```
master:
ukca_um_strat_photol_mod.F90	2	Execution time loop i:	14.372405	True	
ukca_um_strat_photol_mod.F90	3	Execution time loop j:	10.499314	True	
ukca_aerod.F90	                1	Execution time loop j:	35.180508	False	

3310:
ukca_um_strat_photol_mod.F90	2	Execution time loop i:	0.155515	True	
ukca_um_strat_photol_mod.F90	3	Execution time loop j:	0.132040	True	
ukca_aerod.F90	                1	Execution time loop j:	0.695516	False	

```

PR #3316, opened 2026-02-03
(Closes #3314) fix line limit bug for in-line comments

### LINKED ISSUE #3314: Bug in wrapping of comment
If I process icethd_dh.f90 from SI3 with `--keep-comments` and `-l output` we get a line like:
```fortran
          ! remaining "potential" evap is sent to ocean
          wfx_err_sub(ji,jj) = wfx_err_sub(ji,jj) - pevap_rema(ji,jj) * a_i(ji,jj,jl_cat) * r1_dt_ice  ! <=0 (net evap for the &
&ocean in kg.m-2.s-1)
```
where the line-length limiter has failed to spot that it's an in-line comment that is taking us over the length limit. The original source line was:
```fortran
            ! remaining "potential" evap is sent to ocean
            wfx_err_sub(ji,jj) = wfx_err_sub(ji,jj) - pevap_rema(ji,jj) * a_i(ji,jj,jl_cat) * r1_Dt_ice  ! <=0 (net evap for the ocean in kg.m-2.s-1)
```

arporter:


src/psyclone/line_length.py
```diff
@@ -143,18 +140,17 @@ def long_lines(self, fortran_in):
         return False
```
LonelyCat124: Should `long_lines` also be updated? I'm not sure how hard it would be to update this to only return True for non comments/inline comments?

LonelyCat124: At the very least the docstring and type hints should be updated.

LonelyCat124 (changes requested):
Mostly looks good to me - I had one question (depending on how complex long_lines would need to become) and one update to the documentation.

You also need to update the author list for these files.

src/psyclone/line_length.py
arporter: I'm not sure we'd want to do that - presumably the line-length limit applies to comments too?
I've updated the docstring and type hints.

arporter: (I've just googled and it looks like the line-length limit *does* apply to comments.)

arporter:
Should be ready for another look now.

src/psyclone/line_length.py
```diff
@@ -195,7 +221,7 @@ def process(self, fortran_in):
         # We add an extra newline so remove it when we return
         return fortran_out[:-1]
 
-    def _get_line_type(self, line):
+    def _get_line_type(self, line) -> str:
```
LonelyCat124: Can you also add the str typehint to line?

src/psyclone/line_length.py
LonelyCat124: While we're here, can you add the type hint to line_length

LonelyCat124: Also the docstring for the `line_length` on the class, as this option is not currently documented.

LonelyCat124: Can we also add type hints and return type hint here while we're here

arporter: Yes, done.

arporter: Oh yes, good spot.

src/psyclone/line_length.py
arporter: It's all lovely now (I claim).

PR #3317, opened 2026-02-03
Add profiling option for wrapping ROCTx

thomasgibson:
This PR introduces Fortran interfaces for wrapping ROCTx, allowing for instrumenting Fortran codes for trace visualization on AMD GPUs. I have tested this using the `tra_adv.f90` example so far, but still experimenting with it.

Want to create this so others can try this out and review


lib/profiling/amd/README.md
```diff
@@ -0,0 +1,86 @@
...
+must be provided:
+
+```shell
+amdflang <my object files> \
+    -L<PATH-TO-PSYCLONE>/lib/profiling/amd -lroctx_prof \
+    -L$ROCM_PATH/lib -lrocprofiler-sdk-roctx
+```
+
+**Note**: The ``<PATH-TO-PSYCLONE>`` differs depending on whether the
+wrapper library is compiled in a clone of PSyclone repository or in a
```
arporter: "...in a clone of the PSyclone..."

arporter (changes requested):
Thanks very much @thomasgibson. It's looking good. I only found one small typo!
Please could you add a brief mention of the new wrapper to lib/profiling/README.md.

We had slightly crossed wires about updating the Makefile :-) I was thinking you could perhaps update the one in examples/gocean/eg2 so that it is NVIDIA/AMD agnostic? Up to you.

lib/profiling/amd/README.md
thomasgibson: Fixed in https://github.com/stfc/PSyclone/pull/3317/commits/641bc4e2991776729ad58df5525021161fdaaa89

examples/nemo/eg8/omp_gpu_profile_trans.py
```diff
@@ -0,0 +1,116 @@
...
+PROFILING_ENABLED = os.environ.get("ENABLE_PROFILING", False)
+
+
+def add_omp_region_profiling_markers(children: Union[List[Node], Schedule]):
+    """Insert profiling markers around all top-level OpenMP directives.
+
+    :param children: a Schedule or sibling nodes in the PSyIR to which to
+        attempt to add profiling regions.
+    """
+    from utils import add_profile_region
```
arporter: I think you've fallen foul of a recent update to `utils.py` here - it  now contains `add_profiling` rather than `add_profile_region`.

examples/nemo/README.md
```diff
@@ -149,3 +149,13 @@ supported for generic transformations.
 A simple stand-alone example that shows verification that read-only data
 is not modified, e.g. by out-of-bounds accesses to other variables.
 This uses the PSyData interface to instrument generic Fortran code.
+
+## Example 7
+
+OpenMP parallelisation (for CPU and GPU) of `tra_adv` over levels, using
```
arporter, line 155: Thanks for adding this :-)

examples/nemo/eg8/Makefile
```diff
@@ -0,0 +1,81 @@
...
+
+GENERATED_FILES = traadv_instrumented.F90 \
+	traadv_instrumented.o \
+	traadv.exe \
+	output.dat
+
+PROFILER ?= rocprofv3
+PROFILER_FLAGS ?= --runtime-trace --output-format pftrace
+
+# Profiling wrapper settings
```
arporter, line 56: For this to work out of the box (and as part of the CI) you'll need to change this to use Gnu and something like simpletimer. I suggest you keep have these AMD-specific settings as a commented block to show a user what to do if they have the AMD compiler.

arporter (changes requested):
Very nice, thanks Thomas. I'd missed that the example I pointed to was OpenACC only so apologies (I only tested it running on the CPU). Thanks for adding the new example.

Unfortunately `utils.py` got updated last week to make use of @LonelyCat124 's new `MaximalRegionTrans`. You'll need to update your transformation script.
Since we run the examples as part of the CI, please could you update your new example so that it 'works' (i.e. at least compiles) with Gnu and e.g. simple timer by default. If you keep the AMD/rocm flags as comments in the Makefile/README then a user will easily be able to switch if they have the appropriate compiler and hardware.

examples/nemo/eg8/omp_gpu_profile_trans.py
thomasgibson: Fixed here: https://github.com/stfc/PSyclone/pull/3317/commits/401c3dd49139de42850110b31aed80e040ed31be

examples/nemo/eg8/Makefile
thomasgibson: Rewrote the example and set CI-safe defaults: https://github.com/stfc/PSyclone/pull/3317/commits/07f3bd9fe555c695fcd4e660210542b05b5a7e52 Let me know if this looks better

arporter (approved):
Thanks very much Thomas, sorry for all the to-and-fro.
I'll proceed to merge.

PR #3321, opened 2026-02-05
(Closes #3320) remove null (integer) operations when creating PSyIR

### LINKED ISSUE #3320: Removing null operations in the PSyIR frontend
### Discussed in https://github.com/stfc/PSyclone/discussions/3313

<div type='discussions-op-text'>

<sup>Originally posted by **arporter** February  2, 2026</sup>
The pre-processed code of NEMO often ends up with expressions like `limit + 0`, presumably because of the tiling or some such. It will make comparing loop bounds simpler (in the future) if we simplified these expressions. This might feed in to #3310 too as the shortcut that @hiker has found for that problem won't work if we have a `+ 0` expression.

I'm therefore proposing that we tweak the fparser2 frontend slightly to spot this situation. We can update the handler for binary_operation like so:
```
--- a/src/psyclone/psyir/frontend/fparser2.py
+++ b/src/psyclone/psyir/frontend/fparser2.py
@@ -4943,6 +4943,14 @@ class Fparser2Reader():
        binary_op = BinaryOperation(operator, parent=parent)
         self.process_nodes(parent=binary_op, nodes=[arg_nodes[0]])
         self.process_nodes(parent=binary_op, nodes=[arg_nodes[1]])
+        if operator in [BinaryOperation.Operator.ADD,
+                        BinaryOperation.Operator.SUB]:
+            arg1 = binary_op.operands[1]
+            if isinstance(arg1, Literal) and arg1.value == "0":
+                arg0 = binary_op.operands[0].detach()
+                del binary_op
+                return arg0
+            
         return binary_op
```
Apart from the fact that the binary-operation handler can now return something that isn't a BinaryOperation, are there any other drawbacks @LonelyCat124 and @sergisiso?</div>

arporter:


arporter:
NEMOv5 integration tests failed again - with a signal 11 from nvfortran during the GPU offload async build. Since all the 'passthrough' tests and other offload tests passed, I think this is not strictly the fault of this PR but obviously we do need the tests to be passing. I'll see if I can reproduce the failure myself.

LonelyCat124 (changes requested):
It all looks good, just one remaining issue that I asked on teams about multiplication, as I think we can probably do something similar with that?

PR #3308, opened 2026-01-28
Update spack compilers

sergisiso:

PR #3324, opened 2026-02-09
Load required env vars for intel compiler

nmnobre:


.github/workflows/nemo_v5_tests.yml
```diff
@@ -162,6 +162,7 @@ jobs:
         # Set up environment
         source /archive/psyclone-spack/psyclone-spack-Dec25/spack-repo/share/spack/setup-env.sh
         spack unload && spack load nemo-build-environment %${INTEL_TOOLCHAIN}
+        spack load intel-oneapi-compilers@2025.3.1
```
arporter: Thanks Nuno, I confirm that the ITs are OK now. However, I see that we don't have to do this step for the other spack environments (Gnu, NVIDIA) so perhaps it's the intel-onapi nemo-build environment that needs to be updated @sergisiso ?

nmnobre: I just noticed @sergisiso actually added the exact same line in `nemo_tests.yml` in #3308, but missed this because we only run this test on re-attempts. I'm waiting for him to remove version 2025.3.0, at which point I can remove the version specifier from this line on both files.

sergisiso: To clarify, this is indeed needed as otherwise the intel passthrough fails, but my idea is to omit the compilers loads following spack strategy to RPATH everything (or set the needed LD_LIBRARY_PATHs in the package.py), but I didn't have time to look into this specific intel compiler issue in the last PR and wanted to get it merged before my AL. I may have another look in the future but this is not very high in my list of priorities.

nmnobre: > but my idea is to omit the compilers loads following spack strategy to RPATH everything

Yeah, that won't work, because we're using the compiler outside a spack compilation/installation flow. EDIT: Unless we edit `LDFLAGS` in `package.py`, but that's uglier imho.

> (or set the needed LD_LIBRARY_PATHs in the package.py)

~I think we'd have to change the path for the environment, if that's possible. I don't really understand how Spack decides/knows what to add to the path when loading an environment.~ EDIT: I now realise what we call environments are in reality bundle packages and, yes, we'd just need to add the required path next to the netcdf pair. EDIT: I've changed `package.py` for the bundle to add the required path - useless but harmless for the other variants.

nmnobre: Thanks for retriggering the workflow, seems everything is working as expected. :-D

PR #3327, opened 2026-02-10
defusechain update to go beyond calls.

schreiberx:


schreiberx:
> @schreiberx Can you add a test for backward accesses as well? After that I will probably try to also fix the start/stop_point stuff as well.

I just uploaded it.

LonelyCat124 (changes requested):
Hi @schreiberx , I was adding a bit more onto this PR and I don't think this works correctly and the tests are insufficient which is why it didn't catch it.

I was looking at your forward accesses test while adding more to this PR, and decided to update the test from:
```
    assert len(reaches) == 5
    assert isinstance(reaches[1], Call)
```
to check all the elements of reaches are as expected, but I don't think they are:
```
    assert reaches[0] is assigns[2].rhs.children[1]
    assert reaches[1] is calls[0]
    assert reaches[2] is assigns[3].rhs.children[0]
    assert reaches[3] is assigns[3].lhs
```

This fails as reaches[3] is a Call, but I think it should be the lhs of the Assignment, and also that should end the chain.

Can you fix this for the two tests, and can you also propogate this argument to `Reference.next_accesses` and `Reference.previous_accesses` and add tests to check that the argument is propogated correctly (its mostly just a case of repeating the tests but doing them on Reference.next_accesses` instead of using the DUC.

PR #3332, opened 2026-02-16
(Closes #3143) Bugfix for DUC and "previous" calls

### LINKED ISSUE #3143: DefUseChain - unexpected result
I am trying to get the location of where an argument to a call was written previoulsy, and the DUC gives me the call itself. Reproducer:
```
#! /usr/bin/env python3

from psyclone.psyir.nodes import Call, Statement
from psyclone.psyir.tools. definition_use_chains import DefinitionUseChain

code = """
subroutine foo(a, b)
real, intent(inout) :: a
real, intent(inout) :: b
real :: c, d, e, f

c = d * a
b = c + d
call bar(c, b)
b = b + c
end subroutine foo
subroutine bar(x, y)
real, intent(in) :: x
real, intent(inout) :: y
!x = x + 1.0
y = exp(x**2)
end subroutine bar
"""

reader = FortranReader()
routine = psyir.find_routine_psyir("foo")
call = psyir.walk(Call)[0]
arg = call.children[2]
print("ARG", arg)
chain = DefinitionUseChain(arg)
# chain = DefinitionUseChain(arg, stop_point = call.abs_position)
all_prev = chain.find_backward_accesses()
writer = FortranWriter()
for i in all_prev:
    print("prev access: ", writer(i.ancestor(Statement, include_self=True)))

```
So I take the second argument of the call ('b'), ask for previous accesses, and get the call again:

```
ARG Reference[name:'b']
prev access:  call bar(c, b)
```
I would expect to get the `b=c+d` assignment. I can make it work by adding a stop location (the commented out code), then I get as expected:
```
ARG Reference[name:'b']
prev access:  b = c + d
```

Additionally (I did look a bit into DUC), I am wondering if line 756 is correct:
```
                if (isinstance(reference.parent, IntrinsicCall) and
```
If the intrinsic call is using an operation (`sin(a+b)`, then asking for e.g. `a`), the parent will be the `BinaryOperation`. I think that should be `ancestor(Statement, include_self=True)`?

A third question: it looks like the `find_backwards_accesses` can return a Reference, but also any(?) other node type, e.g. a call. Is that expected? I am aware that a loop might have an issue (no reference for the loop variable in the loop statement), but otherwise, shouldn't that function always return a Reference?

LonelyCat124:


src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -820,3 +820,31 @@ def test_definition_use_chain_find_backward_accesses_pure_call(
...
+    c = d * a
+    b = c + d
+    call bar(c, b)
+    b = b + c
+    end subroutine foo
+    """
+    psyir = fortran_reader.psyir_from_source(code)
+    routine = psyir.find_routine_psyir("foo")
+    call = psyir.walk(Call)[0]
+    arg = call.children[2]
```
arporter: Slightly picky but it would be better to use `call.arguments[1]` here.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -820,3 +820,31 @@ def test_definition_use_chain_find_backward_accesses_pure_call(
...
+    chain = DefinitionUseChain(arg)
+    all_prev = chain.find_backward_accesses()
+    # Check that the ancestor call of b isn't a backward access.
+    assert not isinstance(all_prev[0], Call)
+    assert all_prev[0] is routine.children[1].lhs
```
arporter, line 852: Comment  pls - "It should be the LHS of the second (assignment) statement"

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
LonelyCat124: Added a comment with a similar meaning.

PR #3333, opened 2026-02-16
(Towards #2381) Re-structure classes for reduction nodes.

arporter:


arporter:
ITs were green. This is ready for a first look from either @hiker, @sergisiso or @LonelyCat124. There's a fair bit of code movement as I've taken code out of the monoliths that are `lfric.py` and `psyGen.py` and put it into separate modules. I've done the same thing for the tests. Essentially though, it just restructures the class hierarchy and introduces LFRicGlobal{Max,Min,Sum}.

doc/user_guide/transformations.rst
```diff
@@ -723,11 +723,11 @@ PSyclone supports parallel scalar reductions.  If a scalar reduction is
 specified in the Kernel metadata (see the API-specific sections for
 details) then PSyclone ensures the appropriate reduction is performed.
 
-In the case of distributed memory, PSyclone will add **GlobalSum's**
-at the appropriate locations. As can be inferred by the name, only
-"summation" reductions are currently supported for distributed memory.
+In the case of distributed memory, PSyclone will add **GlobalReduction's**
```
LonelyCat124: No apostrophe in `GlobalReductions`

src/psyclone/domain/lfric/lfric_invoke.py
```diff
@@ -57,26 +62,25 @@ class LFRicInvoke(Invoke):
...
+    :raises InternalError: if an unrecognised global reduction operation
+                           is encountered.
     '''
     # pylint: disable=too-many-instance-attributes
     # pylint: disable=too-many-locals
-    def __init__(self, alg_invocation, idx, invokes):
+    def __init__(self,
+                 alg_invocation: InvokeCall,
+                 idx: int,
+                 invokes: "LFRicInvokes"):
```
LonelyCat124, line 80: See teams discussion

src/psyclone/psyGen.py
```diff
@@ -2217,9 +2158,9 @@ def call(self, value):
     def backward_dependence(self):
         '''Returns the preceding argument that this argument has a direct
         dependence with, or None if there is not one. The argument may
-        exist in a call, a haloexchange, or a globalsum.
+        exist in a call, a haloexchange, or a GlobalReduction.
```
LonelyCat124: Since Call and HaloExchange are also classes can you update them as well?

src/psyclone/psyGen.py
```diff
@@ -2230,14 +2171,14 @@ def backward_dependence(self):
     def forward_write_dependencies(self, ignore_halos=False):
         '''Returns a list of following write arguments that this argument has
         dependencies with. The arguments may exist in a call, a
-        haloexchange (unless `ignore_halos` is `True`), or a globalsum. If
-        none are found then return an empty list. If self is not a
+        haloexchange (unless `ignore_halos` is `True`), or a GlobalReduction.
```
LonelyCat124: HaloExchange

src/psyclone/psyGen.py
```diff
@@ -2249,15 +2190,15 @@ def forward_write_dependencies(self, ignore_halos=False):
     def backward_write_dependencies(self, ignore_halos=False):
         '''Returns a list of previous write arguments that this argument has
         dependencies with. The arguments may exist in a call, a
-        haloexchange (unless `ignore_halos` is `True`), or a globalsum. If
-        none are found then return an empty list. If self is not a
+        haloexchange (unless `ignore_halos` is `True`), or a GlobalReduction.
```
LonelyCat124: HaloExchange

src/psyclone/psyGen.py
```diff
@@ -2269,9 +2210,9 @@ def backward_write_dependencies(self, ignore_halos=False):
     def forward_dependence(self):
         '''Returns the following argument that this argument has a direct
         dependence on, or `None` if there is not one. The argument may
-        exist in a call, a haloexchange, or a globalsum.
+        exist in a call, a haloexchange, or a GlobalReduction.
```
LonelyCat124: Call, HaloExchange

src/psyclone/psyGen.py
```diff
@@ -2282,11 +2223,11 @@ def forward_dependence(self):
     def forward_read_dependencies(self):
         '''Returns a list of following read arguments that this argument has
         dependencies with. The arguments may exist in a call, a
-        haloexchange, or a globalsum. If none are found then
+        haloexchange, or a GlobalReduction. If none are found then
```
LonelyCat124: HaloExchange

src/psyclone/domain/common/psylayer/global_reduction.py
```diff
@@ -0,0 +1,117 @@
...
+    _text_name = "GlobalReduction"
+    #: The colour to use when creating a view of this node.
+    _colour = "cyan"
+
+    def __init__(self, scalar: KernelArgument, parent: Node = None):
+        super().__init__(children=[], parent=parent)
+        # Check that distributed memory is enabled
+        if not Config.get().distributed_memory:
+            raise GenerationError(
+                f"It makes no sense to create a {self._text_name} object "
```
LonelyCat124: I think "It makes no sense" is a bit weird for an error message, maybe just "Attempting to create a {self.name} object when distributed memory is not enabled (dm=False)."

src/psyclone/domain/common/psylayer/global_reduction.py
```diff
@@ -0,0 +1,117 @@
...
+                f"when distributed memory is not enabled (dm=False).")
+
+        # Check that the global sum argument is indeed a scalar
+        if not scalar.is_scalar:
+            raise InternalError(
+                f"{self._text_name}.init(): A global reduction argument should"
```
LonelyCat124: This also feels old fashioned, I think just something like `scalar argument to {self.name} should be a scalar but found argument of type...`

src/psyclone/domain/common/psylayer/global_reduction.py
```diff
@@ -0,0 +1,117 @@
...
+                f" be a scalar but found argument of type "
+                f"'{scalar.argument_type}'.")
+
+        self._scalar = copy.copy(scalar)
+        if scalar:
+            # Update scalar values appropriately
+            # Here "readwrite" denotes how the class GlobalSum
+            # accesses/updates a scalar
+            self._scalar.access = AccessType.READWRITE
```
LonelyCat124, line 86: Related to this - what does `reference_accesses` return on a GlobalReduction? Should it be supported or is that something for a different PR to handle? Also tagging @sergisiso for his thoughts.

doc/user_guide/transformations.rst
arporter: Oops, done.

src/psyclone/domain/lfric/lfric_invoke.py
arporter: I've double-checked and mypy needs the import so I'm leaving this unchanged.

arporter: I've also done a bit of tidying to make mypy a bit happier with this file in general.

src/psyclone/psyGen.py
arporter: Done plus typing.

src/psyclone/psyGen.py

src/psyclone/psyGen.py

src/psyclone/psyGen.py

src/psyclone/psyGen.py
arporter: - Done.

src/psyclone/domain/common/psylayer/global_reduction.py
arporter: Agreed. I've gone for "Refusing to create a .... because distributed memory is not enabled.." Is that OK?

src/psyclone/domain/common/psylayer/global_reduction.py
arporter: Updated.

src/psyclone/domain/common/psylayer/global_reduction.py
arporter: I just checked and it returns an empty dict. I did wonder about the fact that `GlobalReduction` has the `._scalar` property rather than children. However, I think changing that is for a different PR.

sergisiso: I agree that this is for a different PR, LFRic kernels also miss the accesses, which are the same varaibles that would interact with the GlobalReduction ones. So it is not a problem for now.

In GOcean we started experimenting giving them accesses representative to their compuational pattern and I want to come back to this in the "replacing legacy functionality with the use of PSyIR" WP in the next LFRic project to see if this is possible also in LFRic.

LonelyCat124: Can you open the issue? I don't think we need any TODOs here, we could have an xfailing test though if you feel up to it, but it depends how feasible that is.

src/psyclone/domain/common/psylayer/global_reduction.py
LonelyCat124: Yeah thats fine.

src/psyclone/domain/common/psylayer/global_reduction.py
arporter: Sure. Opened #3346 for this and added an x-failing test.

src/psyclone/tests/domain/common/psylayer/global_reduction_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+def test_globalreduction_reference_accesses():
+    '''Test the reference_accesses method returns the appropriate
+    information for a GlobalReduction.'''
+    _, invoke = get_invoke("15.14.3_sum_setval_field_builtin.f90",
+                           api="lfric", dist_mem=True, idx=0)
+    schedule = invoke.schedule
+    global_sum = schedule.children[2]
+    vam = global_sum.reference_accesses()
+    if len(vam.all_data_accesses) != 1:
```
LonelyCat124, line 158: Hi Andy,

I have a slight concern that this test will pass if we change something, wheras I think it should fail if it doesn't reach the xfail. I'm guessing from the tests design something like `assert len(vam.all_data_accesses != 1` shouldn't pass if we correctly implement the reference accesses function?

arporter: If we correctly implement `reference_accesses` then I would expect `len(vam.all_data_accesses) == 1`. I've added a check on the name of the variable being updated that will definitely fail, even when that check passes. That way, we can't forget this test when doing the work on `reference_accesses`. Is that what you were after?

LonelyCat124: Yeah, I just prefer to have xfailing tests fail when we fix the issue so we can't miss that the xfailing test should be updated.

LonelyCat124:
Set CI running, otherwise I think everything is resolved.

Issue #1857, opened 2022-09-07
[PSyIR] Arrays of scalars should be arrays of ScalarType


arporter:
Currently an array of Scalars is represented by an `ArrayType` and the type and precision of an element are given by the `intrinsic` and `precision` properties of that class, respectively. However, this is inconsistent with other sorts of an array - an array of Structures for instance has the `intrinsic` as the `StructureType`.

We therefore need to make arrays of scalars consistent with other sorts of arrays. For clarity, we will also rename `ArrayType.intrinsic` to be `ArrayType.elemental_type` and (I think) do away with `precision` altogether since that is captured in `ScalarType`.

PR #3336, opened 2026-02-17
(closes #3325) Improving .datatype to return size intrinsics instead of DEFERRED

### LINKED ISSUE #3325: `.datatype` limits some operations for allocatables.
While working on #3301 it has become apparent that the result of `.datatype` can limit what PSyclone can do. For example, for the expression (where all arrays are allocatable) `-avt_k(:,:,:) * rn2(nis0:nie0,njs0:nje0,:) * wmask(nis0:nie0,njs0:nje0,:)`, PSyclone's returned datatype is only `Array<Scalar<REAL, UNDEFINED>, shape=['DEFERRED', 'DEFERRED', 'DEFERRED']>`. If PSyclone assumes that the input code is valid Fortran, I think we can do better for the shape, where it is instead either: `[size(avt_k, 1), size(avt_k, 2), size(avt_k,3)]` or `[(nis0:nie0), (njs0:nje0), size(avt_k,3)]`.

Without information on the shape of the result, PSyclone cannot create a temporary variable to store the result in, so we can't store the result in a temporary, to then parallelise.

LonelyCat124:
Still in progress, this so far changes the arrayreference datatype to never give shapes with `ArrayType.Extent` in the result if we can resolve the shape.

I want to look at binaryoperation and how it chooses what result to use next, as I think we can do better than the current resul.


src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -168,12 +177,13 @@ def datatype(self):
                 # the variable name should be (TODO #2137).
                 base_type = UnresolvedType()
             else:
-                base_type = self.symbol.datatype
+                # Create a copy of the base datatype.
+                base_type = self.symbol.datatype.datatype.copy()
```
sergisiso: Not the fault of this PR but I always find the datatype.datatype very confusing and I have to dig into it to understand what it tries to do (and we have it in several places).  I propose renaming the ArrayType property as "elemental_type", I think it would make it more readable everywhere we use it? @arporter any opinions?

(I would also remove the comment then as is the same as the statement)

src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -150,16 +150,25 @@ def datatype(self):
...
+                    idx in range(len(shape)))):
+                return self.symbol.datatype
+            if (orig_shape and len(orig_shape) == len(shape) and
+                all(self.is_full_range(idx) for idx in range(len(shape)))
+                and all(isinstance(idx, ArrayType.ArrayBounds) for idx in
+                        orig_shape)):
+                # Although this access has a shape, it is for the
+                # whole array, and the original array has no ArrayType.Extent
+                # elements, so we should return the original shape as its
+                # fully defined without using size intrinsics.
```
sergisiso: I don't agree with this special-case, at least not for the reason in the comment.

Only 1 lfric test fails and is probably due to a wrong assert, if we want to test the symbol declaration type we need to do `isinstance(psyir_args[4].symbol.datatype, ....`

As now ref.datatype (resulting reference datatype) and ref.symbol.datatype (declaration datatype ) have slightly different meanings.

I notice there are also differences in WHERE tests that I haven't looked at, are this the same reason?

src/psyclone/psyir/nodes/operation.py
```diff
@@ -429,8 +430,36 @@ def datatype(self):
                     f"{len(argtypes[1].shape)}")
             # In general there is no way we can check that the extents of each
             # dimension match so we have to assume that they do.
-        shape = (argtypes[0].shape if isinstance(argtypes[0], ArrayType) else
-                 argtypes[1].shape)
+        # FIXME We can do better here, pick whichever we have more information
+        # about, where information is literal > reference > intrinsiccall
+
+        # If only one of the argtypes is an ArrayType then we return that
```
sergisiso: Not yours but I am confused by the use of "argtypes" in this method, can we rename it to "operand_types"

src/psyclone/psyir/nodes/operation.py
```diff
@@ -429,8 +430,36 @@ def datatype(self):
...
+                            not isinstance(x.lower, IntrinsicCall) and
+                            isinstance(x.upper, DataNode) and
+                            not isinstance(x.upper, IntrinsicCall)])
+            # If the second shape has more defined elements then
+            # return the shape of that.
+            if dnodes_1 > dnodes_0:
+                return ArrayType(base_type, shape=argtypes[1].shape)
+            # Otherwise use the default.
+            shape = (argtypes[0].shape if isinstance(argtypes[0], ArrayType)
+                     else argtypes[1].shape)
```
sergisiso: Is the else in this statement ever used? the parent if already tells us that all are ArrayType. If not the comment on top should say "... the first"

src/psyclone/psyir/nodes/operation.py
```diff
@@ -429,8 +430,36 @@ def datatype(self):
                     f"{len(argtypes[1].shape)}")
             # In general there is no way we can check that the extents of each
             # dimension match so we have to assume that they do.
-        shape = (argtypes[0].shape if isinstance(argtypes[0], ArrayType) else
-                 argtypes[1].shape)
+        # FIXME We can do better here, pick whichever we have more information
+        # about, where information is literal > reference > intrinsiccall
```
sergisiso: Remove the FIXME and this comment may be better inside the else block below

src/psyclone/psyir/nodes/operation.py
```diff
@@ -429,8 +430,36 @@ def datatype(self):
...
+
+        # If only one of the argtypes is an ArrayType then we return that
+        if not all(isinstance(argtype, ArrayType) for argtype in argtypes):
+            shape = (argtypes[0].shape if isinstance(argtypes[0], ArrayType)
+                     else argtypes[1].shape)
+        else:
+            # Otherwise we try to work out the best. Find out how many
+            # or non-intrinsic DataNodes shape definitions each
```
sergisiso: Remove "or"?

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -1137,6 +1137,7 @@ def test_elemental_function_to_loop(fortran_reader, fortran_writer):
       end if
     enddo'''
     out = fortran_writer(psyir)
+    print(out)
```
sergisiso: Delete

src/psyclone/psyir/nodes/operation.py
```diff
@@ -429,8 +430,36 @@ def datatype(self):
...
+                            isinstance(x.upper, DataNode) and
+                            not isinstance(x.upper, IntrinsicCall)])
+            # If the second shape has more defined elements then
+            # return the shape of that.
+            if dnodes_1 > dnodes_0:
+                return ArrayType(base_type, shape=argtypes[1].shape)
+            # Otherwise use the default.
+            shape = (argtypes[0].shape if isinstance(argtypes[0], ArrayType)
+                     else argtypes[1].shape)
         return ArrayType(base_type, shape=shape)
```
sergisiso, line 461: Add tests in src/psyclone/tests/psyir/nodes/operation_test.py to show/test this functionality

src/psyclone/psyir/nodes/operation.py
LonelyCat124: I don't think it can be any more since I added the if statement so I'll remove it.

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: I think we already have an issue about ArrayType's subtypes so I think the improvement should be suggested there probably @sergisiso

sergisiso: What is the issue number, I am not able to find it.

LonelyCat124: https://github.com/stfc/PSyclone/issues/1857 I was thinking it sounded related to this.

sergisiso: Ok, agreed

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: I fixed the test for the LFRic test to use symbol.datatype.

I've not looked at the WHERE tests, I don't think I changed any of those? Or if I did I missed it.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: I changed it to optype, I can expand it still if you'd prefer but I'd like it to be a bit concise as a name.

src/psyclone/psyir/nodes/array_reference.py
sergisiso: To clarify, when I tried commenting out this special case code something in the frontend where tests failed. But maybe it was unrelated - no worries if you don't see it.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: Removed the whole block, the commont in the else already describes what we're doing.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: Yeah, it remained from a previous implementation where it was a bit more conditional, but I decided this was sufficient.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
LonelyCat124: Done

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: The second block is needed as otherwise you get:
```
do widx1 = 1, SIZE(a, dim=1), 1
```
instead of just
```
do widx1 = 1, 100, 1
```

src/psyclone/psyir/nodes/operation.py
LonelyCat124: This has shown up a slight issue where I will often get `1 + SIZE(blah) - 1` as the .upper, which doesn't show the behaviour I want so I need to add some slight improvement to the check I think.

LonelyCat124: Fixed this to get the correct behaviour. I don't love that the upper shape is "5-1+1", but I think that's a more fundamental issue with _extent and using integer types that aren't exactly `INTEGER_TYPE`. This test (for whatever reason) uses INTEGER_SINGLE_TYPE (which I just copied), but then this fails at comparing to "one" in `ArrayMixin._extent` because the precisions are different.

@sergisiso @arporter I discussed adding a "get_python_representation" (more concise name is appreciated) to the `Literal` class - can I add that into this PR and use it in _extent to check start/step being one? Or would you rather I make a separate PR and come back to this later.

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: Oh yeah these are the tests where you now get SIZE instead of just `100` because the second block isn't the special case the original comment is referring to - its just the basic case for a full range where we have a non-allocatable array. I'll fix the comment.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: The new implementation is a little more verbose but doesn't care about precision any more.

src/psyclone/psyir/nodes/array_mixin.py
```diff
@@ -609,14 +608,15 @@ def _extent(self, idx):
                 [start.arguments[0].copy(),
                  ("dim", Literal(str(idx+1), INTEGER_TYPE))])
 
-        if start == one and step == one:
+        if (isinstance(start, Literal) and start.value_as_python == 1 and
+                isinstance(step, Literal) and step.value_as_python == 1):
```
sergisiso, line 628: Just for me to understand, what is the difference here? Is that now the precision doesn't matter in the comparison?

Also, I find the "value_as_python" slightly odd, for me it should be the other way around: "as_python_value" or even excluding the python and just being "as_value"

src/psyclone/psyir/nodes/operation.py
```diff
@@ -320,15 +321,15 @@ def has_explicit_grouping(self, value: bool):
...
         '''
         Examines the two operand types to determine the base type of the
         operation using the rules in Section 7.2 of the Fortran2008 standard.
         If the type cannot be determined then an instance of `UnresolvedType`
         is returned.
 
-        :param argtypes: the types of the two operands.
-        :type argtypes: list[:py:class:`psyclone.psyir.symbols.DataType`,
+        :param optypes: the types of the two operands.
+        :type optypes: list[:py:class:`psyclone.psyir.symbols.DataType`,
```
sergisiso: typehint, maybe `Iterable[DataType]`?

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -1073,6 +1073,7 @@ def test_elemental_intrinsic_to_loop(fortran_reader, fortran_writer):
       b(widx1) = a(widx1)
     end if
   enddo'''
+    print(out)
```
sergisiso: Delete

src/psyclone/tests/psyir/nodes/operation_test.py
```diff
@@ -292,6 +292,22 @@ def test_binaryop_array_datatype():
...
+    ref6 = ArrayReference.create(sym2, [Range.create(
+        Literal("1", INTEGER_SINGLE_TYPE),
+        Literal("5", INTEGER_SINGLE_TYPE))])
+    ref7 = ArrayReference.create(sym1, [Range.create(
+        Literal("1", INTEGER_SINGLE_TYPE),
+        IntrinsicCall.create(
+            IntrinsicCall.Intrinsic.SIZE,
+            [Reference(sym2)])
+        )])
+    print("------------")
```
sergisiso: Remove

src/psyclone/tests/psyir/nodes/operation_test.py
```diff
@@ -292,6 +292,22 @@ def test_binaryop_array_datatype():
...
+    binop7 = BinaryOperation.create(oper, ref7.copy(), ref6.copy())
```
sergisiso, line 308: Can you also do another BinOp with the operands switched? I think you are trying to check that the "datatype" prefers the type without the SIZE intrinsic (add a comment saying that this is the purpose), but in this case its the first operand which would also be chosen by default.

src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -168,12 +170,13 @@ def datatype(self):
                 # the variable name should be (TODO #2137).
                 base_type = UnresolvedType()
             else:
-                base_type = self.symbol.datatype
+                # Create a copy of the base datatype.
+                base_type = self.symbol.datatype.datatype.copy()
+
             # TODO #1857 - passing base_type as an instance of ArrayType
             # only works because the ArrayType constructor just pulls out
             # the intrinsic and precision properties of the type.
```
sergisiso: Is this the reason the method have the desirable property that "the datatype method should not propagate the DEFERRED attribute when constructing the datatype of an expression". In this case this should not be a TODO, just a comment. And it will be good to add tests with Extent.ATTRIBUTE and Extent.DEFERRED in src/psyclone/tests/psyir/nodes/array_reference_test.py which are the moment there is none.

Also if the above is true can you update the docstring to say:
```
def datatype(self) -> Datatype:
''' Note that if the resulting datatype is an ArrayType,
this will convert Extent attributes to the appropriate inquiry intrinsic,
which is convenient to directly use the resulting expression as executable
code. Use `node.symbol.datatype` to obtain the type declaration of the
array symbol (with its Extent attributes).

:returns: the resulting datatype of the array access expression.
'''
```
(We may want the first sentence to be in each node.datatype docstring, as it is true by extension. E.g. operation.datatype must also have this property)

src/psyclone/psyir/nodes/array_reference.py
```diff
@@ -150,16 +150,18 @@ def datatype(self):
...
-                # of this method.)
+
+            # When we access the full array, and the array's shape has no
+            # ArrayType.Extent elements we can return the original shape as
+            # its fully defined without needing to use size intrinsics.
+            if (orig_shape and len(orig_shape) == len(shape) and
+                all(self.is_full_range(idx) for idx in range(len(shape)))
+                and all(isinstance(idx, ArrayType.ArrayBounds) for idx in
+                        orig_shape)):
                 return self.symbol.datatype
```
sergisiso, line 168: I am still usure about this:
- Are not all shapes ArrayType.ArrayBounds?, what is the new condition checking?
- What happens if there is a DEFERRED array bounds e.g. `real, dimension(3:) :: a`. There is no tests for this.
- Wouldn't we want the same behaviour if it is not the full range, for instance in . `real, dimension(100,100) :: a` and `a(:,1)`, the resulting shape should preferable be [100] and not SIZE(a, 1).

src/psyclone/psyir/nodes/array_reference.py
sergisiso: By the way, if you extend the src/psyclone/tests/psyir/nodes/array_reference_test.py::test_array_datatype I would prefer to have it with a fortran_writer and a bunch of fortran declaration and array_references. I find it much easier to quickly verify it than reading the constructed PSyIR.

src/psyclone/tests/psyir/nodes/operation_test.py
```diff
@@ -292,6 +292,22 @@ def test_binaryop_array_datatype():
...
+    dtype6 = binop7.datatype
+    assert dtype6.shape[0].lower.value == "1"
+    assert dtype6.shape[0].upper.debug_string() == "5"
```
sergisiso, line 311: Another case we could implement/test is UnresolvedType/UnknwonType + ArrayType

In fortran the operand of an ArrayType can only be an array of the same rank or an scalar. But both will always results as an array of the exact same shape than the ArrayType of the known operand.

src/psyclone/psyir/nodes/array_mixin.py
LonelyCat124: Yes, before it specifically requried the input to be exactly a `INTEGER_TYPE` which meant it didn't work for e.g. `INTEGER_SINGLE_TYPE` which is used in several places in the test suite at least, so now we can ignore that.

In theory this might do something pythony (i.e. 1.0 == 1) if the start/step is a REAL, but that sunsupported Fortran so I don't want to check.

src/psyclone/psyir/nodes/operation.py
LonelyCat124: I've gone for list[DataType] as it matches the current typehint and the typehint on the called function.

LonelyCat124: I also used annotations here as I thinkI can't import `DataType` in this file.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/nodes/operation_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/nodes/operation_test.py
LonelyCat124: Added a comment and the inverted test.

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: 1. No, shapes are either `ArrayBounds` or `Extent`. Since `Extent` (e.g. DEFERRED) is not an `ArrayBounds` so we don't return the datatype in that case as we need to return sizes.
2. This goes into the rest of the function, so should return something involving size but messaged on teams.
3. Ideally we probably would but I'm not sure how to do that, I can take al ook.

LonelyCat124: ArrayBounds can also contain extents (happens for the `3:` case) apparently so I need to handle that too...

LonelyCat124: I was overthinking the solution to this a lot, its actually straightforward.

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: I moved the ones that were simple to generate to be generated by code + fortran_reader now.

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: I think we already get this for 3. I've added a test.

src/psyclone/tests/psyir/nodes/operation_test.py
LonelyCat124: I had a dive on this and my answer is no, and my reason is because we don't know the scalar result, i.e. the Unresolved/UnknowType can be a double and the other array be a real, so while we know the shape we can't know the result type (it could be int + real or float + double for example).

src/psyclone/psyir/nodes/operation.py
LonelyCat124: Done.

src/psyclone/tests/psyir/nodes/operation_test.py
sergisiso: Good point, I did not consider the change of the types.

FYI: I may still give it a go propagating an ArrayType with elements of UnresolvedType and the shape of the known operand, in some situations I am interested to understand the shape-rank of the expression, but I don't care about the type. But this is outside the scope of this PR.

src/psyclone/psyir/nodes/array_reference.py
LonelyCat124: Done

src/psyclone/tests/psyir/nodes/array_reference_test.py
```diff
@@ -472,45 +472,76 @@ def test_array_indices():
...
+    thing = test(:, 1)
+
+
+    end subroutine code"""
+    psyir = fortran_reader.psyir_from_source(code)
+    refs = psyir.walk(ArrayReference)
+
     # Reference to a single element of an array.
-    aref = ArrayReference.create(test_sym, [one])
+    aref = refs[0]
```
sergisiso, line 496: Thanks for using code, just to make it a bit more redable, can you add the associated array accessor in each comment every time you use refs[x], e.g. in this case:
` # Reference to a single element of an array (test(1))`

src/psyclone/tests/psyir/nodes/array_reference_test.py
```diff
@@ -472,45 +472,76 @@ def test_array_indices():
...
+    assert dtype.intrinsic == ScalarType.Intrinsic.REAL
+    assert len(dtype.shape) == 1
+    assert dtype.shape[0].lower.value == "1"
+    assert isinstance(dtype.shape[0].upper, IntrinsicCall)
+    assert dtype.shape[0].upper.intrinsic.name == "SIZE"
+    assert dtype.shape[0].upper.arguments[0].symbol.name == "test3"
+
+    # Test we don't get a size for the 1D access to a 2D array.
+    dref = refs[4]
+    dtype = dref.datatype
```
sergisiso: This tests the contrary to what the comment says, a 2D access to a 1D array
```
real, dimension(10) :: test
thing = test(:, 1)
```
Shouldn't this produce an error? It is not valid fortran, so I would say it shouldn't be valid PSyIR.

Also, the opposite, what the comment says, maybe could aslo be an error as I belive Fortran accessors must be complete - exactly match the rank?

sergisiso (changes requested):
@LonelyCat124 I was almost ready to approve it but I don't understand one of the tests. Could you have a look?

Also, is there anything pending to close the associated issue?

src/psyclone/tests/psyir/nodes/array_reference_test.py
LonelyCat124: Added.

src/psyclone/tests/psyir/nodes/array_reference_test.py
LonelyCat124: Yeah I'd meant to use the `test_2d` array for this test, and now I need to see about fixing the result / if its feasible.

LonelyCat124: @sergisiso This is due to the definition of `_extent` that we get SIZE and not just a literal (or expression) here:

```
            # Access is to full range and start and stop are expressed in terms
            # of LBOUND and UBOUND. Therefore, it's simpler to use SIZE.
            return IntrinsicCall.create(
                IntrinsicCall.Intrinsic.SIZE,
                [start.arguments[0].copy(),
                 ("dim", Literal(str(idx+1), INTEGER_TYPE))])
```

I can modify this where if we know the bounds of the array we return an expression instead of the IntrinsicCall, but I don't know if this will have any side effects elsewhere. What do you think?

sergisiso: As you prefer, a literal would be better but SIZE is acceptable (its not a DEFERRED). If it has side effects can leave the literal improvement for another PR.

LonelyCat124: I have an implementation that doesn't appear to break much, but is quite specific so I'll let you look.

LonelyCat124: Lost some coverage so I'll try to sort that.

sergisiso (approved):
@LonelyCat124 All comments have been addressed. Nicely done simplifying in the frontend and letting more cases output literals instead of intrinsics.

I expect some conflicts with the PR that is marked as "Ready to be merged" but since this are partially my fault, I will mark this one as approved and resolve the conflicts myself.

PR #3344, opened 2026-02-23
(Closes #2202) Update the generation of Fortran declarations to observe dependencies

### LINKED ISSUE #2202: PSyclone reordering symbol table during output resulting in compiler error

arporter:
This is a relatively simple PR - it takes the inter-dependence handling out of `_gen_param_decls` and puts it into `gen_decls`. This also then means we can delete `_gen_param_decls`. I've added tests for the various tricky cases described in the original Issue and they all pass now :-)


arporter:
Although things are now better, `complex` types are still a problem - I've added a (failing) test which demonstrates this.

arporter:
All the ITs passed. However, I've realised that the order of output is not guaranteed - if more than one Symbol has all its dependencies declared (satisfied) at a given point then we should sort those Symbols and declare them in alphabetical order.

Issue #2837, opened 2025-01-07
Switch test suite to use ubuntu-latest when Python 3.7 support is removed.


LonelyCat124:
Python 3.7 is no longer supported as part of ubuntu-latest, so #2836 has switched the test suite to use 22.04. Once python 3.7 is removed we can switch this back to ubuntu-latest.

LonelyCat124:
Also depending on timeline, there is a import statement for python 3.7 in #2877 which can be removed.

LonelyCat124:
Also affects some of the code for #2668

arporter:
![Image](https://github.com/user-attachments/assets/7ea14291-eddf-4790-af56-24afca758cc0)
so I propose we move to 3.10 as the oldest supported Python version.

PR #3345, opened 2026-02-23
(Closes #3330) MetaTransformation for OpenMPCPURoutineTrans

### LINKED ISSUE #3330: Create meta transformation to apply OpenMP CPU parallelism with nowaits to a routine
Sergi suggests that once we have the maximal region transformation finished we can create a meta transformation to just apply the whole tool chain to a full routine for OpenMP

This minimises the effort for the basic cases in codes and allows specialised scripts where required whilst a single transformation can be applied to full routines for the majority.

There are a few outstanding questions:
1. Should this go in psyir/transformations or a new metatransformation location or some other name? My opinion is not the former as it's not really a transformation.
2. What should the validate function really do? One difficulty is avoiding modifying the tree if e.g. the maximal region trans fails, but this is difficult because without having the previous transformation steps applied. This one seems difficult short of doing everything to a copy of the tree first...

LonelyCat124:
Initial implementation which is very bare bones and unfinished. I got a bit confused with inheritance, and there are still various issues with this.

Things that I was hoping would work, e.g. using superclasses validate/apply methods is a bit messy due to the MRO, since super().apply(...) does result in a call to `self.validate()` which doesn't work if this has any type checking in its own validate function.

I can work around this by making it's own validate function do very little and manually calling each superclasses inherited functionality.

The alternative (and potentially neater way to do it) is to only define the inherited classes for the purposes of validating available options and having documentation fixes (potentially with other improvements) but then create instances of the superclasses as required. The issue with this is pulling out the appropriate options for each superclass (as the option validation would no longer work for just passing in all options)

I'm also not entirely sure how `__init__ ` works on this class right now, I think it will just use `OMPLoopTrans`, but I'm not sure I even want the user to have the option to change that, so I should potentially override `__init__` to not allow these options to be changed.

I will show its current "working state" in the next post.

@sergisiso @arporter Do either of you have thoughts on a best direction for this?

Edit:
I also havd to make a change in the Transformation base class to never fail on the `"options"` kwarg, as this could be being provided to the class by some superclass call and then breaking valid options checking in a different base classes' superclass. It was a bit weird, but is a problem of options still being defined on some classes but not newer ones.


LonelyCat124:
This is now reimplemented using the new `_SUB_TRANSFORMATIONS` infrastructure and works for the documentation. I still need to create a test from the above cases to ensure it works correctly.

LonelyCat124:
I'm hopeful this is almost ready once #3472 is on.

LonelyCat124:
@sergisiso @arporter This is ready for a first look now.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
+# -----------------------------------------------------------------------------
```
arporter, line 1: I thought I'd build the docs first to see what they looked like and that made me realise that this transformation needs to be added to the UG - possibly here: https://psyclone.readthedocs.io/en/latest/user_guide/transformations.html#openmp. Perhaps we need a new section on meta-transformations too?

arporter: I got myself confused by looking at the Doxygen-generated reference guide, only to realise that our clever Python magic doesn't work there and therefore the keyword options are not documented. Perhaps it's time to drop the Doxgen version (not for this PR though)?

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
...
+from psyclone.psyGen import Transformation
+from psyclone.psyir.nodes import (
+    Directive, Loop, Routine
+)
+from psyclone.psyir.transformations.transformation_error import (
+    TransformationError)
+from psyclone.psyir.transformations.omp_loop_trans import OMPLoopTrans
+from psyclone.psyir.transformations.maximal_omp_parallel_region_trans import (
+    MaximalOMPParallelRegionTrans)
+from psyclone.psyir.transformations.omp_minimise_sync_trans import (
```
arporter, line 45: Alpha order please (if that works).

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
...
+                            OMPMinimiseSyncTrans]
+
+    def validate(self, node: Routine, **kwargs):
+        '''
+        Validates the input options of the OpenMPCPURoutineTrans.
+
+        :param node: The Routine node to validate.
+        '''
+        # Validate the provided options are allowed and typed correctly.
+        self.validate_options(**kwargs)
```
arporter, line 74: Is there some magic to call the `validate` of the sub transformations or is that done on the way through the apply? Please add a comment.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
...
+    def apply(self, node: Routine, **kwargs):
+        '''
+        Applies the OMPLoopTrans, MaximalOMPParallelRegionTrans and
+        OMPMinimiseSyncTrans to the relevant parts of the input
+        node.
+
+        :param node: The Routine node to transform
+        '''
+
+        # Split the options for the subtransformations.
```
arporter: Presumably these come back in the order that the sub-transformations are specified in `_SUB_TRANSFORMATIONS`? Please add a comment.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Find all of the loops.
+        loops = node.walk(Loop)
+        ltrans = OMPLoopTrans()
+        for loop in loops:
+            if loop.ancestor(Directive):
+                continue  # Skip if an outer loop is already parallelised
+            try:
+                # Validate that this loop can be parallelised.
+                ltrans.validate(loop, **loop_kwargs)
```
arporter: I'm not sure there's much to be gained by calling `validate` separately here, so long as `apply` calls it (which it should)?

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
...
+                # If it is, then apply the OMPLoopTrans
+                ltrans.apply(loop, **loop_kwargs)
+            except TransformationError:
+                # If we fail to parallelise a loop we just skip it.
+                continue
+
+        # Apply the maximal openMP parallel region transformation to the
+        # routine.
+        momprtrans = MaximalOMPParallelRegionTrans()
+        momprtrans.validate(node.children[:], **maxpar_kwargs)
```
arporter: Again, no need for separate `validate`?

This does raise an issue though: for normal transformations the rule is that if `validate` succeeds then the transformation must succeed. Presumably that's not true here? Is it possible to make it work like this? Maybe by refactoring this functionality into a separate method that can be used by both `validate` and `apply` but with a `dryrun` option for the former use case? (There's code somewhere that already works like this.)

Linked to this, if this `apply` (or `validate`) fails then the whole transformation will fail but the tree will have been modified by the previous OMPLoopTrans applications? Do we need to take a copy of the sub-tree with `node` at its root before proceeding?

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
```diff
@@ -0,0 +1,124 @@
...
+        momprtrans.apply(node.children[:], **maxpar_kwargs)
+
+        nowait = self.get_option("nowait", **kwargs)
+        # If the asynchronous option was specified, then we need to apply the
+        # OMPMinimiseSyncTrans as well.
+        if nowait:
+            minstrans = OMPMinimiseSyncTrans()
+            minstrans.validate(node, **minsync_kwargs)
```
arporter: Same considerations here.

src/psyclone/psyir/transformations/omp_minimise_sync_trans.py
```diff
@@ -35,9 +35,7 @@
 
 '''Contains the OMPMinimiseSyncTrans.'''
 
-# TODO #2837: Once we leave python 3.8 we can use list instead of List for
-# type hints.
-from typing import List, Union
+from typing import Union
```
arporter, line 38: I got excited thinking we could close another issue but I see #2837 is already closed. However, I've grepped and can't find any other references to it either so that's good.

src/psyclone/tests/psyir/transformations/metatransformations/omp_cpu_routine_trans_test.py
```diff
@@ -0,0 +1,180 @@
...
+    a(i) = i
+  enddo
+  !$omp end do nowait
+  !$omp do schedule(auto)
+  do i = 1, 100, 1
+    b(i) = 4 + i
+  enddo
+  !$omp end do nowait
+  !$omp end parallel"""
+    assert correct in out
```
arporter, line 180: Please could you add some more contrary tests where things may fail. Is it possible for us to have put some OMP do's on loops but then still fail to do the parallel region? Or would we get multiple parallel regions?

src/psyclone/psyGen.py
```diff
@@ -2502,6 +2502,11 @@ def validate_options(self, **kwargs):
         wrong_types = {}
         for option in kwargs:
             if option not in valid_options:
+                # This is needed to enable metatransformations where
+                # only some inherited classes have options set on
+                # superclasses
+                if option == "options":
```
arporter, line 2505: Should this have a TODO for getting rid of `options` dict?

arporter (changes requested):
Thanks Aidan. I'm new to the meta-transformation thing so some of my requests might be a bit odd but I thought it would help me understand.
You need to add some tests to check what happens when things don't go according to plan.
Is it worth modifying the NEMO script to make use of this new functionality? That would stress test it!

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
sergisiso: > Perhaps we need a new section on meta-transformations too?

Is this a detail that we want to expose to the user? I thought of this as just an implementation detail to compose transformations. (Unless you mean a DevGuide section)

In the same line, this PR exposes them in the import:
`from psyclone.psyir.transformations.metatransformations import OMPCPURoutineTrans`
Are we sure we want to distiguish them? Would it be simpler with:
`from psyclone.psyir.transformations import OMPCPURoutineTrans`

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: At the moment its done in the validate, but some of the vlidation steps are only done after some of the application.

There is a possibility that we could copy the tree, and proceed through the operations (until the final validate) in the validate method instead, which would avoid some unexpected errors (though I'm not sure what they could be)? But would be very computationally intensive since copying an entire Routine and doing multiple steps of application to the copy.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: Yeah I think its unneeded.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: My concern with the validation is purely cost-wise (copying an entire tree and performing basically the whole transformation), but leaving the tree in an unfinished state is probably not ideal.
However, I don't believe this transformation can ever fail in a way that results in an illegal tree. If it only performs the transformation that adds `omp do` then the resulting tree isn't invalid (as they're ignored if outside a parallel region) - though PSyclone may fail to output the code due to its validate global constraints maybe.

I'm happy to do the dry_run validate if you think its worthwhile.

src/psyclone/tests/psyir/transformations/metatransformations/omp_cpu_routine_trans_test.py
LonelyCat124: No I don't think its possible (at least I can't think of a way it is, and its still standard compliant code if we did I believe).

src/psyclone/psyGen.py
LonelyCat124: Yeah added.

LonelyCat124:
@arporter I've addressed most of the comments but before I send it back for review properly let me know if you want me to use a dry run to create a validate function - see my comments for more details.

Edit:
I'm also not sure how it would be possible to test various of the failure cases (assuming I need to stop on various TransformationErrors) since I don't believe there is a way to have them occur.

LonelyCat124:
Add a test to code that has no parallelisable loops in.

LonelyCat124:
~~@arporter This is ready for review now I think.~~

Edit: I forgot I still need to do the docs.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: I'm happy to do either - currently I've not changed it, but if its as simple of moving the declaration from one __init__ to another then I'm happy to do it.

LonelyCat124: There's already a section on meta transformations in the developer guide in the transformations section titled: `Transformations options and meta-transformations`.

LonelyCat124:
@arporter This is ready for another look now.

doc/developer_guide/transformations.rst
```diff
@@ -589,18 +589,12 @@ kwargs as in the example below:
             self._trans2().validate(node, **tr2_kwargs)
             self.validate_options(**self_kwargs)
             super().validate(node, **self_kwargs)
-
+    
 		def apply(self, node, my_option):
```
arporter: The indentation of this doesn't appear to be correct (when I build the pdf):

Image

doc/user_guide/transformations.rst
```diff
@@ -477,6 +477,12 @@ can be found in the API-specific sections).
 
 ####
 
+.. autoclass:: psyclone.psyir.transformations.metatransformations.OMPCPURoutineTrans
```
arporter: Sorry to be a pain but please could you move this so that it's in alphabetical order with the other OMP* transformations?

doc/user_guide/transformations.rst
```diff
@@ -750,6 +756,17 @@ caused by Taskloops, and adds OpenMP Taskwait statements to satisfy those
 dependencies. An example of using OpenMP tasking is available in
 `PSyclone/examples/nemo/eg1/openmp_taskloop_trans.py`.
 
+OpenMP CPU Routine Transformation
++++++++++++++++++++++++++++++++++
+The OpenMP CPU Routine transformation applies simple OpenMP parallelism to
```
arporter: I think you're underplaying it - s/simple//

arporter: Also, can you make a link to the earlier section where its docs are included?

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
arporter: Oh yes, that's fine then. I'm with Sergi on the simpler import:
`from psyclone.psyir.transformations import OMPCPURoutineTrans` so if you could move it that would be great.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
arporter: We've agreed not to bother with the dry-run approach for now. If we do hit problems in future we could re-visit this decision.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
arporter: We've agreed not to do the dry_run for now.

arporter (changes requested):
Thanks Aidan, almost there now. The formatting of the UG looks like it needs a bit of a tweak and there's just making OMPCPURoutineTrans available from `psyir.transformations`.

doc/user_guide/transformations.rst
LonelyCat124: Done

doc/user_guide/transformations.rst
LonelyCat124: Yeah, I think I meant "basic" instead of simple, but nothing is cleaner.

doc/developer_guide/transformations.rst
LonelyCat124: I think there was some mismatch of tabs and spaces in this, my editor was unhappy but I couldn't work out why (I'd assumed it was my editor). I think its fixed now.

LonelyCat124: Fixed.

doc/user_guide/transformations.rst
LonelyCat124: I could make a reference work to the autodoc, so I just added a comment to check the documentation - if either you or  @sergisiso knows of a way to link directly to the documentation I'll swap to doing that, but attempting to label the autodoc didn't work for me.

src/psyclone/psyir/transformations/metatransformations/omp_cpu_routine_trans.py
LonelyCat124: Done.

LonelyCat124:
@arporter Addressed everything I could - the reference to the autodoc wasn't working for me and I couldn't find a solution on google, so I've just put text saying to refer to the docs for more detail.

doc/user_guide/transformations.rst
```diff
@@ -742,14 +748,27 @@ transformation.
...
++++++++++++++++++++++++++++++++++
+The OpenMP CPU Routine transformation applies OpenMP parallelism to
+an entire routine. The transformation attempts to apply OpenMP loop directives
+to every loop in the routine, and then builds the largest OpenMP parallel
+region(s) possible to apply parallelism. If the ``nowait`` option is specified,
+then the transformation will attempt to maximise the number of nowait clauses
+on the loop directives, and minimise the number of barriers required to
+keep the output code correct.
+
+See the transformation's (``OpenMPCPURoutineTrans``) documentation for more details.
```
arporter: It's `OMPCPURoutineTrans` but if that's the only thing I'll fix it...

doc/developer_guide/transformations.rst
arporter: Not quite but I used the power of vi and removed another tab.

doc/user_guide/transformations.rst
arporter: Yeah, I think you do need a header to link to. Not to worry.

doc/user_guide/transformations.rst
arporter: Fixed.

arporter (approved):
All good now, thanks very much Aidan.
(I've taken the liberty of fixing a couple of minor things in the docs.)

PR #3353, opened 2026-02-25
(Closes #3328) fixes duplicated nodes in definition use chains

### LINKED ISSUE #3328: DefUseChain can output the same reference multiple times with nested loops
There is a bug in DefUseChains with nested loops where the same loop can occur multiple times.

Helpfully the comment in the code says:
```
                        # We will only ever reach a reference once, so
                        # we don't need to check uniqueness.
```
at line 367, which seems to be incorrect, so we just need a uniqueness check here.

LonelyCat124:


LonelyCat124:
This reminds me I wanted to do type hints for this file so I'll check that before this is ready for review.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -86,10 +84,10 @@ class DefinitionUseChain:
 
     def __init__(
         self,
-        reference,
-        control_flow_region=None,
-        start_point=None,
-        stop_point=None,
+        reference: Reference,
+        control_flow_region: Iterable[Node] = (),
+        start_point: int = None,
```
sergisiso: Optional[int] if these 2 can be None

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -364,9 +357,12 @@ def find_forward_accesses(self):
...
                     for ref in chain._reaches:
-                        # We will only ever reach a reference once, so
-                        # we don't need to check uniqueness.
-                        self._reaches.append(ref)
+                        # We need to check for uniqueness
+                        for ref2 in self._reaches:
+                            if ref2 is ref:
+                                break
+                        else:
+                            self._reaches.append(ref)
```
sergisiso: There are several self._reaches.append in the code. Some have comments saying it will be unique but some don't. Does the problem affect any other append? Is it a problem for _uses, _defouts, _killed?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -364,9 +357,12 @@ def find_forward_accesses(self):
                             return self._reaches
 
                     for ref in chain._reaches:
-                        # We will only ever reach a reference once, so
-                        # we don't need to check uniqueness.
-                        self._reaches.append(ref)
+                        # We need to check for uniqueness
```
sergisiso: Can you expand this a bit, it was not clear to me at first look. Maybe: "# Add node to the _reaches list if it is not already there (note that a "not in" check is not enough because the references are equal, but this needs identity uniqueness)"

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Added

PR #3354, opened 2026-02-26
(closes #2429) Make all structure constructors and functions with unknown declarations Calls

### LINKED ISSUE #2429: Function calls without argument become code blocks
A simple example:
```
module module_call_tree_mod
contains
  integer function module_function()
    module_function = 1
  end function module_function

  subroutine call_local_function()
    integer i
    i = module_function()
  end subroutine call_local_function

end module module_call_tree_mod
```

The RHS of the assignment in `call_local_function` becomes a code block: fparser returns this as `Structure_Constructor`, for which no handler is found. Around lines 2820 in fparser2 it will do:
```
            generic_type = type(child).__bases__[0]
            handler = self.handlers.get(generic_type)
```
But still gets no handler, which raises an ImplenetationError (which turns this into a code block). I would think that this might be an fparser issue, but I didn't dig any deeper - if fparser should return something else instead of `Structure_Constructor`, let me know and I'll open a ticket in fparser instead.

The code works when I add a parameter, and fparser returns a `Part_ref` instead. Not sure how a non-argument function should be returned.

sergisiso:
In turn this simplifies the Invoke/Kernel uplifting, because we don't have to deal with the case with Codeblocks anymore.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,7 @@
+   58) PR #3354 for #2429. Extends the fparser2 frontend such that all
+   structure constructors and functions with unknown delcarations are
+   represented by Call nodes.
+
    57) PR #3355 towards #3252. Improves PSY_VALUE_RANGE format, introduces a
    generic value_range_check library and runs examples during the tests.
  


sergisiso:
@arporter @LonelyCat124 @hiker This is ready for a first review.

src/psyclone/tests/core/symbolic_maths_test.py
```diff
@@ -232,7 +232,7 @@ def test_symbolic_math_never_equal(fortran_reader, exp1, exp2, result):
     source = f'''program test_prog
                 use some_mod
                 integer :: i, j, k, x
-                type(my_mod_type) :: a, b
+                real, dimension(10) :: a, b
```
arporter, line 235: How come you had to change this?

arporter (changes requested):
Nice, impressive simplification :-)
I just have a question about one of the changes to a test.

src/psyclone/tests/core/symbolic_maths_test.py
sergisiso: The last parameterise case fails with: "Expected parent PSyIR node to be either a Reference or a Member but got 'Call' when processing ':'" while handling the a(:) _subscript_triplet_handler. And this is fair, this is not valid Fortran.

I assume these declarations were just copied from another test is the same file?

arporter: Oh yes, I hadn't spotted that it wasn't valid Fortran.

Issue #3252, opened 2025-12-09
Value Range Check is broken for DSL (at least gocean)


hiker:
Since the change to gocen to use the arrays instead of fields (i.e. `field%data` instead of `field`), there need for domain-specific implementations of PSydata has gone, but as a side effect, the value range check functionality is broken:

Since the Fortran range check function received `field%data` as string, it checks for environment variables called `PSYVERIFY__module__kernel__field%data`, instead of looking for `PSYVERIFY__module__kernel__field`.

And of course `%` is not a valid part of an environment variable :(

Not sure how to best fix this. Immediate idea: in the Fortran psydata libraries remove everything after the first `%` from the variable name.  This would work for gocean, but in generic Fortran, a user might have variables `a%b` and `a%c`, and might want to specify different valid ranges for both of them, which then would not be possible. In gocean we also do get variables like `neighbours%internal%xstart`, i.e. multiple `%` (though I would accept that loop boundaries should not need to be supported?)

Or we replace any `%` with `_`. But then the user would have to specify `field_data` to specify the range for `field`, which is not entirely intuitive, the user should not need to know that there is a `data` anywhere :(

Suggestions welcome.

Second issue: this bug should have been detected by the CI. We do have a test, but what obviously happened was that the test was updated to expect `field%data`. We also have examples/gocean/eg5, but that only tests for infinity, not that setting a variable work. Similarly, the training did not actually test this at all (so that will become part of #1623).

Summary:
- [ ] Find and implement (likely in the Fortran wrapper and the transformations?) the best way to support gocean.
- [ ] Add/Update tests or examples to check that setting environment variables works
- [ ] Add/update training to test this properly (this will be done as part of #1623, just adding this here to remind me :) )

PR #3355, opened 2026-02-27
(towards #3252) improve value range env variable

hiker:
Changes the implementation to use a single environment variable `PSY_VALUE_RANGE` to specify all variables, which adds flexible support for structures etc (`field%data=...`).

It does not address the fact that now a user needs to know the 'internal' name (after lowering) of a variable that is passed to a kernel.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,4 +1,7 @@
-   56) PR 3343 towards #1960 and #1799. Improves ArrayAssignment2Loop
+   57) PR #3355 towards #3252. Improves PSY_VALUE_RANGE format, introduces a
+   generic value_range_check library and runs examples during the tests.
+ 
+   56) PR #3343 towards #1960 and #1799. Improves ArrayAssignment2Loop
    by using call/RoutineSymbol datatype information.
  
    55) PR #3333 for #2381 - Restructures global reduction nodes to support


hiker:
IT passed, ready for review.

hiker:
Note that I had a README update after IT, they will obviously not affect the results.

examples/gocean/eg5/value_range_check/Makefile
```diff
@@ -75,12 +77,16 @@ transform: test.x90 value_range_check_transformation.py
 
 run: compile
 	# Note that if the output line is not found, the make will abort with error
-	./value_range_check 2>&1| grep "PSyData: Variable a_fld has the invalid value"
+	./value_range_check 2>&1| grep "PSyData: Variable 'a_fld%data' has the invalid value"
```
sergisiso, line 80: Great.
Maybe we should update ".github/workflows/compilation.yml" to do
 `F90=gfortran make -C examples run` instead of compile?
(I tried and most succeeded but there are some errors, but I reckon we can bypass those since they are broken also for command line execution, and we need to fix them at some point anyway)

doc/user_guide/psy_data.rst
```diff
@@ -260,47 +260,48 @@ test all input and output parameters of a kernel to make sure they are
...
-The range is specified as a ``:`` separated tuple::
+At runtime, the environment variable PSY_VALUE_RANGE is used to specify 
+which variables are within what expected range, and optionally also at
+which location. The range is specified as a ``:`` separated tuple::
 
     1.1:3.3   A value between 1.1 and 3.3 (inclusive).
     :3.3      A value less than or equal to 3.3
     1.1:      A value greater than or equal to 1.1
 
 The syntax for the environment variable is one of:
```
sergisiso: Should this mention that is ;-separated entries?

doc/user_guide/psy_data.rst
```diff
@@ -260,47 +260,48 @@ test all input and output parameters of a kernel to make sure they are
...
 
-``PSYVERIFY__module__kernel__variable``
+``PSY_VALUE_RANGE="module.kernel.variable=...``
```
sergisiso: Now that this is domain-agnostic we probably shouldn't use 'kernel'. I see the lib implementation uses "region" which is a better name, but it is also not documented anywhere what this means. I believe is the region_name that any PSyData can give to its application region? Could you add some explanation about this in the lines below.

examples/gocean/eg5/value_range_check/Makefile
hiker: I am trying. I've fixed one failing example on my laptop, but I get stuck in code for accelerators. I'll check how it behaves in the IT.

doc/user_guide/psy_data.rst
hiker: Oops - of course.

doc/user_guide/psy_data.rst
hiker: That made me think :) Using 'generic' was not immediately supported (mainly because the 'generic' base class has the wrong module name, so the code would not compile). But I've added a generic implementation (which basically inherits all, just in a different module - same thing I do with extraction :) ), and added a generic example as well. Then I reworded the paragraph.

PR #3357, opened 2026-03-02
(Closes #1857) Rename ArrayType.datatype to elemental_type

### LINKED ISSUE #1857: [PSyIR] Arrays of scalars should be arrays of ScalarType

sergisiso:
"datatype" is an overloaded term and not very precise for arrays. In particular there currently are several node.datatype.datatype expressions that are hard to follow, and this change makes them more redable.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   60) PR #3357 for #1857. Re-names ArrayType.datatype to
+   ArrayType.elemental_type and precision and intrinsic properties.
+
    59) PR #3353 for #3328. Fix a case of duplicated nodes returned by DUC.
 
    58) PR #3354 for #2429. Extends the fparser2 frontend such that all


src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -723,16 +726,15 @@ def _node_from_int(var):
...
+        self._elemental_type = elemental_type
 
     @property
-    def datatype(self):
+    def elemental_type(self) -> Union[DataType, DataTypeSymbol]:
         '''
         :returns: the datatype of each element in the array.
-        :rtype: :py:class:`psyclone.psyir.symbols.DataSymbol`
         '''
         # TODO #1857: This property might be affected.
```
arporter: Can we remove this TODO now?

arporter: Actually, I think it stays.

Issue #2949, opened 2025-04-09
Symbol.find_symbol_table(node) should consider shadowing


sergisiso:
This can be done by adding `and current.lookup(self.name) is self` in the condition but was temporary disabled in #2834 because LFRic has some duplicated ContainerSymbols

Issue #2975, opened 2025-05-02
Recover LFRic constants_mod qualified imports


sergisiso:
#1010 uses a wilcard "use constants_mod", but LFRic conventions require qualified imports, so it would be good to recover the "only:" clause here

arporter:
We seem to have qualified imports now so I'm closing this.

Issue #1823, opened 2022-08-03
KernelInline does not consider name clashes from different Invokes


sergisiso:
When kernelInline transformation is applied to multiple kernels in an Algorithm with multiple Invokes (e.g. LFRic gw_pressure_operator_alg_mod.x90) there are situations when 2 Invokes call the same kernel and PSyclone will Inline both without noticing the name clash.

The output Fortran will fail compilation due to the duplicated declaration. (manually deleting on of the inlined subroutine make the code valid again)

Repeated kernels inside the same Invoke are not a problem.

sergisiso:
I started #1844 to fix these issues. I could deal with the duplicated declarations when the same kernel is in multiple invokes in the same file and a more subtle issues I found when psy.gen is called multiple times.

But I need your opinion how to solve issues with missing symbols inside the inlined routines. The problem is when the kernel code uses a globally imported variable. e.g.:

```
module dg_inc_matrix_vector_kernel_mod
     use constants_mod,           only : i_def, r_single, r_double
     ....
contains
    subroutine dg_inc_matrix_vector_code_r_single(cell, ...)
       ...
       real(kind=r_single), dimension(undf2),              intent(in)    :: x
       ...
```

When this routine is inlined the r_single is undeclared in the _psy.f90 file. There is a rule/check that no global variables are used inside inlined kernels but this is just checking the code and not the declarations kinds. I could:

1) Always import r_single and r_double in the psy module. This will solve it for my case but will potentially fail for some other in the future.
2) Check the imported symbols and bring them to the psy module. But I am not sure if having no global symbols in kernel code is a PSyKAl rule that makes ensuring the kernel is thread_safe easier to verify. Although kind symbols are safe. So should I made and exception for them?
3) Make the PSyclone check fail for this kernels because of the global import. Then they will need to be written as:
```
module dg_inc_matrix_vector_kernel_mod
     
     ....
contains
    subroutine dg_inc_matrix_vector_code_r_single(cell, ...)
       use constants_mod,           only : i_def, r_single, r_double
       ...
       real(kind=r_single), dimension(undf2),              intent(in)    :: x
       ...
```
this will require updating a few LFRic kernel code.

@rupertford @arporter @TeranIvy what do you think?

arporter:
I think we should allow/make an exception for kind parameters. In #924 I copy any imports over to the call site, even if they are in module scope.

sergisiso:
[...]

Issue #2874, opened 2025-01-27
Remove arg_ordering legacy code


sergisiso:
Currently we store 2 duplicated args list: _arglist and _psyir_arglist. After #1010 the former could be eliminated and the `append(cell_ref_name)` should be fully deleted in favour of the already present `psyir_append(cell_ref)`

This represent changes in quite a few places because many method return both (str and reference) to populate each list. For example, "cell_ref_name" should be "cell_ref" and only return the Reference. Additionally, all these methods have a "var_accesses" argument. They can now be deleted in favour of doing `cell_ref.reference_accesses(...)` at the caller site, as all information should be in the tree now.

PR #3358, opened 2026-03-02
(Closes #1734 #2975 #2949) Add stricter checking for symbols

### LINKED ISSUE #1734: SymbolTable: imported symbol may be added to a table that does not include the originating container
Currently there's nothing stopping us adding a Symbol with an import interface to a table, even if that import interface refers to a ContainerSymbol that is not in that table but instead in an ancestor. If this is done then a `deepcopy` of the former table fails as the container symbol cannot be found.

arporter:
I've broken this out of the work on #1823 - that work revealed issues with our Symbol handling so I thought I might as well do the job properly.

We now no longer add a wildcard import of `constants_mod` in LFRic generated code. This revealed a few bugs/misssing symbols that I've now fixed.

### DIFF
--- a/src/psyclone/psyir/symbols/symbol_table.py
+++ b/src/psyclone/psyir/symbols/symbol_table.py
@@ -52,14 +52,14 @@ from typing import Any, List, Optional, Union, TYPE_CHECKING
 
 from psyclone.configuration import Config
 from psyclone.errors import InternalError
-if TYPE_CHECKING:
-    from psyclone.psyir.nodes.scoping_node import ScopingNode
 from psyclone.psyir.symbols import (
     DataSymbol, ContainerSymbol, DataTypeSymbol,
     ImportInterface, RoutineSymbol, Symbol, SymbolError, UnresolvedInterface)
 from psyclone.psyir.symbols.intrinsic_symbol import IntrinsicSymbol
 from psyclone.psyir.symbols.typed_symbol import TypedSymbol
 
+if TYPE_CHECKING:
+    from psyclone.psyir.nodes.scoping_node import ScopingNode
 
 # Used to provide a unique default value for methods within the
 # SymbolTable class. This enables us to determine when the user has
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   70) PR #3358 for #1734, #2975 and #2949. Improve symbol table consistency
+   with better copies, validation checks and symbol localisation operations.
+
    69) PR #3384 for #3383. Adds a pytest fixture to set the logging level
    in fparser to CRITICAL. This simplifies the error messages produced
    when a test fails.


arporter:
I'll need to add new tests for the new functionality I've added to SymbolTable.

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -119,12 +119,18 @@ def get_user_type(self, module_name, user_type, name, tag=None):
...
+        # create the ContainerSymbol for the import for the type:
         try:
             # Check if the module is already declared:
             module = self._symtab.lookup(module_name)
-            # Get the symbol table in which the module is declared:
-            mod_sym_tab = module.find_symbol_table(self._kern)
+            # Get the symbol table in which the module is declared. Check
+            # the current table first:
+            if module in self._symtab.symbols:
+                mod_sym_tab = self._symtab
```
sergisiso: Seems odd that the `find_symbol_table(self._symtab.node)` below doesn't already search the given scope as the parameter is `:param node: the PSyIR node from which to search.` but the implementation starts with a node.scope, wich means the ancestor scope.

Could you check if including the given scope in the search would break other things. I think this would match the expected behaviour of this method name. And even if it breaks other things should those other calls then be `...find_symbol_table(parent)` instead?

src/psyclone/domain/lfric/lfric_symbol_table.py
```diff
@@ -56,54 +58,23 @@ class LFRicSymbolTable(SymbolTable):
...
-        :py:class:`psyclone.psyir.symbols.Symbol.Visibility`
-
     '''
-
-    # The container symbol for all precision variables
-    _constants_mod = None
-    # A mapping of 'i_def' etc. to the corresponding DataSymbol
-    _precision_map = {}
-
-    def __init__(self, node=None, default_visibility=Symbol.Visibility.PUBLIC):
```
sergisiso, line 75: Always happy to see methods deleted from this class, as we have a good strategy to do DSL nodes (by uplifting-lowering), but subclassing other parts of PSyclone for API-specific logic is problematic (discussed here: #1954).

This PR is big enough, but do you think we could fully remove this class? This is now just helpers to create symbols with the LFRic precisions, but the generic find_or_create already accepts a datatype argument to specify these.

src/psyclone/domain/lfric/lfric_symbol_table.py
```diff
@@ -56,54 +58,23 @@ class LFRicSymbolTable(SymbolTable):
...
-        super().__init__(node, default_visibility)
```
sergisiso, line 76: Also, asking about related issues, does this PR bring any update to #2659 #1258 or #2975?

src/psyclone/psyir/nodes/extract_node.py
```diff
@@ -470,7 +470,9 @@ def bring_external_symbols(read_write_info: "ReadWriteInfo",
                 continue
             container = symbol_table.find_or_create(
                 module_name, symbol_type=ContainerSymbol)
-
+            # Take care in case we've found an existing ContainerSymbol and
+            # it's in an outer scope.
+            actual_table = container.find_symbol_table(symbol_table.node)
```
sergisiso: I am a bit confused about why do we need here, is it to "declare the new imported symbols in the same scope as container they come from"? If thats the case this may be a better comment.

src/psyclone/psyir/nodes/scoping_node.py
```diff
@@ -128,7 +130,7 @@ def _refine_copy(self, other):
         '''
         # Reorganise the symbol table construction to occur before we add
         # the children.
-        self._symbol_table = other.symbol_table.deep_copy(self)
+        self._symbol_table = other.symbol_table.deep_copy(new_node=self)
```
sergisiso: Not the fault of this PR but I always found that argument name strange. Could we replace "new_node" to "attached_to"? Since the symtab method that does the same is call "attach"

src/psyclone/psyir/nodes/scoping_node.py
```diff
@@ -201,10 +205,12 @@ def replace_symbols_using(self, table_or_symbol):
         this node (if any), the one passed to the child nodes is updated to be
         the one associated with this node.
 
+        The symbols within the associated symbol table are also updated to
+        ensure that any symbols upon which they depend are replaced with the
+        new copies.
```
sergisiso: Is this still true? I am asking because the body hasn't change, so I want to know if this is a clarificaiton of this method behaviour or is a leftover?

src/psyclone/psyir/symbols/symbol.py
```diff
@@ -438,8 +438,7 @@ def find_symbol_table(self, node):
         try:
             current = node.scope.symbol_table
             while current:
-                if self.name in current:
-                    # TODO #2949: Add: and current.lookup(self.name) is self:
+                if self.name in current and current.lookup(self.name) is self:
```
sergisiso, line 441: Nice! Does this PR close #2949?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -42,15 +42,18 @@
...
 from collections.abc import Iterable
 import inspect
 import copy
 import logging
-from typing import Any, List, Optional, Union
+from typing import Any, List, Optional, Union, TYPE_CHECKING
 
 from psyclone.configuration import Config
 from psyclone.errors import InternalError
+if TYPE_CHECKING:  # pragma: no cover
```
sergisiso: I belive Codecov automatically ignore this without the "#pragma: no cover"

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
```diff
@@ -42,7 +42,6 @@
 and move it to psyir/transformations/.
 
 '''
-from typing import List
```
sergisiso, line 45: There is a remaining TODO to #1734 in this file, line 377

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -291,7 +292,8 @@ def deep_copy(self, new_node=None):
                 new_sym = symbol.copy()
                 if new_sym.is_import:
                     name = new_sym.interface.container_symbol.name
-                    new_sym.interface.container_symbol = new_st.lookup(name)
+                    new_sym.interface.container_symbol = new_st.find_or_create(
+                        name, symbol_type=ContainerSymbol)
```
sergisiso: It is a bit unexpected that deep_copy creates new symbols or changes the name of symbols (find_or_create behaviour). But I understand this is necessary if the symbol has a dependency to an outer scope AND we change the associated node, which could make a lookup go wrong. Is this the reason?

I will be happier if we extend the second paragraph of the docstring to explain this behaviour as a consequence of providing a different associated scope node.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -791,28 +799,116 @@ def _add_container_symbols_from_table(self, other_table):
...
+                    # name imported from different containers. We don't
+                    # support renaming an imported symbol but the
+                    # symbol in this table can be renamed so we do that.
+                    self.rename_symbol(
+                        other_sym,
+                        self.next_available_name(
+                            other_sym.name, other_table=other_table))
+            self.update_import_interface(isym)
+
+    def update_symbol_dependencies(self):
```
sergisiso: "thinking out loud" comment, no need to agree, but I am not sure if "update_" is too generic to find out what these two methods do without reading the docstrings but I am not sure of good alternatives. What I could think of is:
- localise_symbol_dependencies
- bind_symbol_dependencies_to_scope

same for the next one where I also wonder if adding "_of" would help make clear is the argument that we update, .e.g. "localise_import_interface_of(symbol)"

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -791,28 +799,116 @@ def _add_container_symbols_from_table(self, other_table):
...
+                    # Symbols to update.
+                    break
+
+    def update_import_interface(self, isym: Symbol):
+        '''
+        Update the import interface of the supplied Symbol so that it
+        refers to a Container in this or an ancestor scope.
+
+        If no ContainerSymbol is found with the name of the one in the
+        ImportInterface of the Symbol, then the ContainerSymboll in the
```
sergisiso: "ContainerSymbol"

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -1078,6 +1103,111 @@ def test_add_container_symbols_from_table():
...
+    # Repeat but have the array sized by a variable with the same name as
+    # an existing (Container)Symbol.
+    clash = symbols.DataSymbol("b_module", symbols.INTEGER_TYPE)
+    artype2 = symbols.ArrayType(stype, [Reference(clash)])
+    table.new_symbol("my_array2", symbol_type=symbols.DataSymbol,
+                     datatype=artype2)
+    with pytest.raises(symbols.SymbolError) as err:
+        table.update_symbol_dependencies()
+    assert ("which is not of the same type as the original dependency"
+            in str(err.value))
```
sergisiso, line 1170: If I am not wrong, this test also needs a case with an outer symbol that has been shadowed in the inner scope, but the "update" should not change the interface of an import in the inner scope (something the implementation already accounts for if I understand the comments correctly)

src/psyclone/lfric.py
```diff
@@ -6465,14 +6473,8 @@ def _find_or_create_type(mod_name: str,
             except KeyError:
                 mod_map = LFRicConstants().UTILITIES_MOD_MAP
                 const_mod = mod_map["constants"]["module"]
-                try:
-                    constants_container = symtab.lookup(const_mod)
-                except KeyError:
-                    # TODO Once #696 is done, we should *always* have a
```
sergisiso, line 6471: Where are we with #696?

src/psyclone/domain/lfric/kern_call_arg_list.py
arporter: That is odd. I've just checked `node.scope` though and it does:
```python
 node = self.ancestor(ScopingNode, include_self=True)
```
so it *will* return its own table first. I think the problem is that I hacked this code in towards the end because I hit situations where `self._symtab.node` was `None`. I'll refactor it.

src/psyclone/domain/lfric/lfric_symbol_table.py
arporter: Yes we could. I started doing it and it was going well but it got quite big so I decided to do the Right Thing and break it into #3366.

src/psyclone/domain/lfric/lfric_symbol_table.py
arporter: Arguably, #1258 could already be closed but I'll add it to the list for this PR. Unfortunately this PR doesn't do anything to address the root of #2659 - that is a harder problem. (Although it *will* improve the consistency of a tree as soon as it is copied.) #2975 will be closed by this (thanks for spotting).

src/psyclone/psyir/nodes/extract_node.py
arporter: Yes, updated

src/psyclone/psyir/nodes/scoping_node.py
arporter: Oh yes, that's a good idea. I also realised that we didn't check the type of the supplied node so I've added that plus test.

src/psyclone/psyir/nodes/scoping_node.py
arporter: I think it must have been a leftover - this method doesn't touch symbol tables.

src/psyclone/psyir/symbols/symbol.py
arporter: Aha! Yes, I'll add it to the list :-)

src/psyclone/psyir/symbols/symbol_table.py
arporter: Removed.

src/psyclone/domain/common/transformations/kernel_module_inline_trans.py
arporter: Oh yes. Fixed now

src/psyclone/psyir/symbols/symbol_table.py
arporter: Agreed. Note that it won't change the name of a Symbol but it will add a new ContainerSymbol if none is found. I tried removing this and ~70 tests broke, primarily because of the way we create a temporary copy of a SymbolTable within `LFRicKern.reference_accesses()`. That's already the subject of #2874 so I've put in a few more TODOs. I've extended the paragraph as you suggested.

However, I've also realised that perhaps I should be stricter about imported Symbols. As things are now, I check that a suitable ContainerSymbol is in scope when a Symbol with ImportInterface is added to a table. We could tighten this further to require that the ContainerSymbol is in the *same* table as the imported symbol. What do you think?

src/psyclone/psyir/symbols/symbol_table.py
arporter: Oops

src/psyclone/psyir/symbols/symbol_table.py
arporter: Agreed. I've gone for `localise_all_symbol_dependencies` and `localise_import_interface_of`.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
arporter: Good point. Test added - let me know if it isn't quite what you had in mind.

src/psyclone/lfric.py
arporter: Another good spot - #696 is done so I've closed it. I've just grepped and found no remaining TODOs for it.

src/psyclone/domain/lfric/lfric_symbol_table.py
sergisiso: I can see two TODOs to #1258, should we drop it from closes?

src/psyclone/domain/lfric/kern_call_arg_list.py
sergisiso: Has this been adressed? If I remove the lines:
```
  if module in self._symtab.symbols:
      mod_sym_tab = self._symtab
  else:
```
all unit tests still pass.

src/psyclone/psyir/symbols/symbol_table.py
sergisiso: I still get a wrong-import-position for this line. It has to be moved below the other imports.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -791,28 +819,114 @@ def _add_container_symbols_from_table(self, other_table):
...
+                        other_sym,
+                        self.next_available_name(
+                            other_sym.name, other_table=other_table))
+            self.localise_import_interface_of(isym)
+
+    def localise_all_symbol_dependencies(self):
+        '''
+        For each Symbol in this table, examines the Symbols upon which it
+        depends and updates them to be the Symbols in scope in this table.
+        If there is no corresponding Symbol in scope then the dependency
```
sergisiso, line 847: I am not sure of the use of "in scope" "in this table", to me it is one or the other (depending whether it looks to parent tables or not). Maybe the second "in" should be an "of"?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -791,28 +819,114 @@ def _add_container_symbols_from_table(self, other_table):
...
+        is added to this table.
+
+        :raises SymbolError: if the Symbol found in the current scope is
+                             of a different type to the original dependency.
+        '''
+        all_symbols: set[Symbol] = set(self.symbols)
```
sergisiso: I have no problem with the `set[Symbol]` annotation if it is for visual clarification, but if it is for the tooling it should be `self.symbols` that needs to be updated to return the right type, not a local annotation.

src/psyclone/psyir/nodes/node.py
```diff
@@ -1674,6 +1674,15 @@ def _refine_copy(self, other):
...
         self.children.extend([child.copy() for child in other.children])
+
+        # Now that the copied children are connected into the tree, ensure any
+        # dangling symbols are re-connected. We can't do this in ScopingNode
+        # because the re-attachment only happens after the copy has been
+        # made (above).
+        for child in self.children:
+            if hasattr(child, "symbol_table"):
+                child.symbol_table.localise_all_symbol_dependencies()
+
```
sergisiso, line 1686: The first question is why can not this be made in the ScopingNode after `super()._refine_copy`?

The seconds, looking there, there already is a:
`self.replace_symbols_using(self._symbol_table)` which already iterates over the dependent symbols and updates them to be from the new scope. What does the `localise_all_symbol_dependency` does differently than the `replace_symbols_using`?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -791,28 +819,114 @@ def _add_container_symbols_from_table(self, other_table):
...
+                    # We found a new Symbol so re-start the loop over
+                    # Symbols to update.
+                    break
+
+    def localise_import_interface_of(self, isym: Symbol):
+        '''
+        Update the import interface of the supplied Symbol so that it
+        refers to a Container in this or an ancestor scope.
+
+        If no ContainerSymbol is found with the name of the one in the
```
sergisiso: Maybe just:
"If no ContainerSymbol with the same name as the one in the ImportInterface is found, the ImportInterface symbol is added to this table."

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -791,28 +819,114 @@ def _add_container_symbols_from_table(self, other_table):
...
+        ImportInterface of the Symbol, then the ContainerSymbol in the
+        ImportInterface is added to this table.
+
+        :param isym: a Symbol with an import interface.
```
sergisiso: I think just "symbol" would be a better name for this argument

src/psyclone/tests/domain/lfric/lfric_types_test.py
```diff
@@ -95,7 +95,9 @@ def test_constants_mod():
         # pylint: disable=no-member
         assert isinstance(symbol, DataSymbol)
         assert isinstance(symbol.interface, ImportInterface)
-        assert symbol.interface.container_symbol is module
+        # Just check the name as we don't actually use the generated module
+        # symbol instance.
+        assert symbol.interface.container_symbol.name == module.name
```
sergisiso: Would `==` work? instead of .name

src/psyclone/tests/psyir/transformations/profile_test.py
```diff
@@ -646,23 +661,27 @@ def test_region(fortran_writer):
...
     assert ("CALL profile_psy_data_3 % PreStart(\"multi_functions_multi_"
             "invokes_psy\", \"invoke_0-r0\", 0, 0)" in result)
     assert ("CALL profile_psy_data % PreStart(\"multi_functions_multi_"
             "invokes_psy\", \"invoke_0-r1\", 0, 0)" in result)
     assert ("CALL profile_psy_data_1 % PreStart(\"multi_functions_multi_"
-            "invokes_psy\", \"invoke_0-r2\", 0, 0)" in result)
+            "invokes_psy\", \"invoke_0-r3\", 0, 0)" in result)
     assert ("CALL profile_psy_data_2 % PreStart(\"multi_functions_multi_"
-            "invokes_psy\", \"invoke_0-testkern_code-r3\", 0, 0)" in result)
+            "invokes_psy\", \"invoke_0-r2\", 0, 0)" in result)
```
sergisiso: Do you know why these changed?

src/psyclone/lfric.py
```diff
@@ -1355,7 +1355,8 @@ def _add_symbol(self, name, tag, intrinsic_type, arg, rank):
             self.symtab.new_symbol(name,
                                    symbol_type=DataSymbol,
                                    datatype=dtype,
-                                   tag=tag)
+                                   tag=tag,
+                                   interface=AutomaticInterface())
```
sergisiso: Several explicit AutomaticInterface have been added to this file, but this is also the default if not interface is provided. Was this just to make the code clearer, or was it needed?

(No problem if you prefer to have them to make the code more explicit)

src/psyclone/domain/lfric/kern_call_arg_list.py
arporter: That's very strange. I could have sworn I changed that code. I too see no test failures if I modify it. I'll try the ITs too...

src/psyclone/domain/lfric/lfric_symbol_table.py
arporter: Oh yes, I should have checked. We still hit those situations so I've removed #1258 from the closes list.

src/psyclone/domain/lfric/kern_call_arg_list.py
arporter: We do get a failure in tutorial/practicals/LFRic/single_node/1_openmp due to #2874. I've updated the code structure to make it clearer that that is what's going on. I still need to create a test case that exercises this.

arporter: I'm beginning to think this may be a bug because it's very hard to get a test case to exercise it.

arporter: I finally tracked it down - it's due to the temporary copy we make of the symbol table. This misses symbols declared in the container scope. I started to fix that but then fell foul of all the new consistency checks so decided to leave it as it is for now with additional comments.

src/psyclone/psyir/symbols/symbol_table.py
arporter: Oh yes, fixed.

src/psyclone/psyir/symbols/symbol_table.py
arporter: Yes. I've changed it to just "in this table".

src/psyclone/psyir/symbols/symbol_table.py
arporter: I'm not sure why I did that. I started running mypy on symbol_table.py but it all got a bit out of hand so I've backed off from that.

src/psyclone/psyir/nodes/node.py
arporter: I tried putting this code inside ScopingNode originally but found that, at that point, there was no connection to the ancestor - this is only updated within `self.children.extend` and that happens after the copy has been completed. I've extended the comment to try to explain.

`replace_symbols_using` updates the tree at and below a given node. It doesn't update the symbol table and symbol definitions. In contrast, `localise_all_symbol_dependencies` updates the Symbol definitions but does not touch the tree.

src/psyclone/psyir/symbols/symbol_table.py
arporter: Done.

src/psyclone/psyir/symbols/symbol_table.py
arporter: Agreed.

src/psyclone/tests/domain/lfric/lfric_types_test.py
arporter: Apparently it does.

src/psyclone/tests/psyir/transformations/profile_test.py
arporter: Because I messed up the changes to the nodes being transformed. Fixed now.

src/psyclone/tests/domain/lfric/lfric_types_test.py
arporter: I take that back. It does in isolation but if other things have run then we get different Symbol instances. As this is #2659 I've added a TODO to the test.

src/psyclone/lfric.py
arporter: Ah, no. They were a hangover from when I was having problems. I've reverted them.

PR #3371, opened 2026-03-12
(towards #3012) exclude feature for psyclone

schreiberx:
Salut Sergi,
This is a simple additional feature where one can use the environment variable `PSYCLONE_EXCLUDE_FILES` to provide a comma separated list of files that shouldn't be processed with psyclone.
We could also tweak the build system, but in cases such as NEMO that's more challenging, hence this option.
Cheers,
Martin


sergisiso:
> I like to skip files that, e.g., fparser fails to parse or which I simply want to skip.

This is the purpose of FILES_TO_SKIP. If it doesn't work like this, it is a bug.

> does that mean asdf.f would just be copied to asdf.f90, right?

Hopefully it should still be copied as "asdf.f", but psyclonefc adds a "psycloned" to identify them, so it should be "asdf.psycloned.f" (but still be a copy, no psyclone involved). The output filename is:
```
  stem = Path(argument).stem
  suffix = Path(argument).suffix
  output = f"{stem}.psycloned{suffix}"
```

schreiberx:
Yet another thought I'm currently facing:
It would also be nice to have Psyclone parse only a single file. E.g., to have a whitelist
`PSYCLONEFC_WHITELIST_FILES=...`
and then only these files are processed if it's specified.
E.g., I just want to parse a single NEMO file with Psyclone, but nothing else to speed up the compilation steps.

sergisiso (approved):
@schreiberx I acknowledge the problem and added to my TODO list in https://github.com/stfc/PSyclone/issues/3012. I am also happy to merge the `PSYCLONEFC_EXCLUDE_FILES` workaround if it is useful for you, but note that psyclonefc is not a stable interface yet and things may still change.

Issue #1445, opened 2021-10-11
Extensions to hoist transformations


hiker:
Left over issues from #1387:

1. The current implementation of the hoist transformation is limited to assignments. It should be extended to any type of node, as long as it valid to move.
2. We might also want to move more than just one statement, which can help with certain code constructs:
```
   a=2
   b=sqrt(a)
   a=3
   c(i) = a*b
```
The assignment `a=2` alone cannot be moved out, even `a=2;b=sqrt(a)` is difficult (with the current dependency check it would not be allowed due to the second assignment to a in the loop). But all three statements would be allowed to be moved.

PR #3373, opened 2026-03-13
Changes to OpenMP scripts to extract arguments from iom_put

LonelyCat124:


LonelyCat124:
ITs all pass now, the remaining question is whether there's any performance degredation - I don't think there should be but its possible this is being done "too widely". One for either @sergisiso or @arporter to review.

examples/nemo/scripts/utils.py
```diff
@@ -526,3 +527,17 @@ def _satisfies_minimum_region_rules(self, region: list[Node]) -> bool:
...
+     in a temporary if they are an Operation with an array datatype.'''
+    for call in calls:
+        if call.symbol.name == "iom_put":
+            arg = call.arguments[1]
+            dtype = arg.datatype
+            if isinstance(dtype, ArrayType) and isinstance(arg, Operation):
+                try:
+                    DataNodeToTempTrans().apply(arg)
+                except TransformationError:
+                    pass
```
sergisiso: The integration tests don't show any performance advantage which is not what we expected (we are changing from multiple gpu->cpu reads to one, and to maybe preventing the data touched from the gpu to be brought back):
- I can check with a grep how many more loops are offloaded but for the places that it was not applied, could you add as preceding comment the reason why not (if not all transformation errors provide useful information, a verbose option like other transformation have can help)
- There is nothing specific of iom_put, other than we know it is a common pattern. We want to avoid touching things from the CPU as much as possible, could this be applied to all subroutine calls (not functions)?

sergisiso (changes requested):
@LonelyCat124 The changes look good, but I would like that the transformation provide more feedback in order to understand why it hasn't given a performance improvement. Also, see if it can be more generic.

examples/nemo/scripts/utils.py
LonelyCat124: I've added a preceding comment now. I'll try generalising it as well.

LonelyCat124: This currently causes stuff to fail, I'll try to see if I can get my VPN to start working again and see if I can try manually building NEMO5 to find the cause.

LonelyCat124: This has shown a few issues with the DataNodeToTempTrans (partly because some things are `Statement`s that I didn't think, e.g. an IfBlock's condition).

LonelyCat124: Fixed the bugs now

LonelyCat124:
Oh ok - I'm not sure all the cases are with intrinsics (but I can't rememeber), but I'm happy to remove the code that tries to move the allocation outside of parent loops for now (as I can't constryct code to test it anyway) and see where we are up to with the ITs.

LonelyCat124:
This still results in NEMO5 failures on accuracy after 5 timesteps for GPU Bench:

```
<  it :       5    |ssh|_max:  0.4949503007019763D+01 |U|_max:  0.1145357177490683D-01 |V|_max:  0.5364770396338010D-01 S_min:  0.2996918706295251D+02 S_max:  0.3101392651315916D+02
<  it :       6    |ssh|_max:  0.5140472974504294D+01 |U|_max:  0.1064859943349158D-01 |V|_max:  0.6818865538921472D-01 S_min:  0.2996921262561763D+02 S_max:  0.3101392594056643D+02
<  it :       7    |ssh|_max:  0.5229361171698658D+01 |U|_max:  0.7814316351505365D-02 |V|_max:  0.8358086738711774D-01 S_min:  0.2996923864577587D+02 S_max:  0.3101392538498657D+02
<  it :       8    |ssh|_max:  0.5220719217849735D+01 |U|_max:  0.1141515836389971D-01 |V|_max:  0.9761604183737975D-01 S_min:  0.2996926417117689D+02 S_max:  0.3101392490495337D+02
<  it :       9    |ssh|_max:  0.5145297949564593D+01 |U|_max:  0.1416399592481446D-01 |V|_max:  0.1152759253497915D+00 S_min:  0.2996929035879930D+02 S_max:  0.3101392444612748D+02
<  it :      10    |ssh|_max:  0.4979557010366685D+01 |U|_max:  0.1986785874282287D-01 |V|_max:  0.1303543987480510D+00 S_min:  0.2996931641421842D+02 S_max:  0.3101392405137852D+02
---
>  it :       5    |ssh|_max:  0.4949503007019763D+01 |U|_max:  0.1145357177490683D-01 |V|_max:  0.5364770396338010D-01 S_min:  0.2996918706295250D+02 S_max:  0.3101392651315916D+02
>  it :       6    |ssh|_max:  0.5140472974504294D+01 |U|_max:  0.1064859943349160D-01 |V|_max:  0.6818865538921472D-01 S_min:  0.2996921262561763D+02 S_max:  0.3101392594056643D+02
>  it :       7    |ssh|_max:  0.5229361171698658D+01 |U|_max:  0.7814316351505386D-02 |V|_max:  0.8358086738711780D-01 S_min:  0.2996923864577586D+02 S_max:  0.3101392538498657D+02
>  it :       8    |ssh|_max:  0.5220719217849735D+01 |U|_max:  0.1141515836389971D-01 |V|_max:  0.9761604183737996D-01 S_min:  0.2996926417117689D+02 S_max:  0.3101392490495337D+02
>  it :       9    |ssh|_max:  0.5145297949564593D+01 |U|_max:  0.1416399592481445D-01 |V|_max:  0.1152759253497916D+00 S_min:  0.2996929035879929D+02 S_max:  0.3101392444612748D+02
>  it :      10    |ssh|_max:  0.4979557010366685D+01 |U|_max:  0.1986785874282303D-01 |V|_max:  0.1303543987480510D+00 S_min:  0.2996931641421842D+02 S_max:  0.3101392405137852D+02
```

@sergisiso At this point I think its best if we just move back to iom_put for now and look at this later in more detail?

sergisiso:
Sure, this already spots and fix quite a few things, we can revert to only iom_put and do a separate issue to continue the more generic application.

LonelyCat124:
Checking the status on integration, NEMO4 and NEMO5 worked, but LFRic failed, but I'm not sure why since I don't think I touched that, I'm waiting to see if it fails again, if so I'll finish updating this branch and then try again.

LonelyCat124:
Integration passed this time - I think the previous failure may have been running on Thursday when the machine restarted so not worried about it for now. Will try to cover the remaining lines next.

LonelyCat124:
@sergisiso This is ready for another look now at last.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -565,8 +565,11 @@ def _maxval_return_type(node: IntrinsicCall) -> DataType:
 
     :returns: the computed datatype for the IntrinsicCall.
     """
+    dtype = ScalarType(
```
sergisiso, line 572: Could twe call `_type_of_named_arg_accounting_for_dim_arg(node, "array")` instead of having a maxval specialisation?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -588,8 +591,8 @@ def _dot_product_return_type(node: IntrinsicCall) -> DataType:
     from psyclone.psyir.tools.type_info_computation import (
         compute_scalar_type
     )
-    veca_datatype = node.argument_by_name("vector_a").datatype
-    vecb_datatype = node.argument_by_name("vector_b").datatype
+    veca_datatype = node.argument_by_name("vector_a").datatype.elemental_type
```
sergisiso, line 598: Woudn't this fail if `node.argument_by_name("vector_a").datatype` or the same in vector_b has UnresolvedType?

Should be do a if not ArrayType for both return UnresolvedType at the top?

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -3215,7 +3218,9 @@ class Intrinsic(IAttr, Enum):
             optional_args={"kind": DataNode},
             return_type=lambda node: (
                 _type_of_scalar_with_optional_kind(
-                    node, node.argument_by_name("l").datatype.intrinsic,
+                    node,
+                    node.argument_by_name("l").
+                    datatype.intrinsic,
```
sergisiso: Keep this in the same line.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -287,7 +287,9 @@ def test_datanodetotemptrans_apply(fortran_reader, fortran_writer, tmp_path,
     out = fortran_writer(psyir)
     assert """  integer, allocatable, dimension(:,:) :: tmp
 
-  ALLOCATE(tmp(1:SIZE(a, dim=1),1:SIZE(b, dim=2)))
+  if (.NOT.ALLOCATED(tmp)) then
```
sergisiso, line 305: This test has an unused monkeypatch argument.

Line 54 also has a unused tmp_path argument

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -287,7 +287,9 @@ def test_datanodetotemptrans_apply(fortran_reader, fortran_writer, tmp_path,
...
+    ALLOCATE(tmp(1:SIZE(a, dim=1),1:SIZE(b, dim=2)))
```
sergisiso, line 306: That's fine but all the information is static, how come we end up with the SIZE?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -338,7 +340,9 @@ def test_datanodetotemptrans_apply(fortran_reader, fortran_writer, tmp_path,
     out = fortran_writer(psyir)
     assert """  integer, allocatable, dimension(:) :: tmp
 
-  ALLOCATE(tmp(1:3))
+  if (.NOT.ALLOCATED(tmp)) then
+    ALLOCATE(tmp(1:3))
+  end if
```
sergisiso: And when we have static bound information, wouldn't it be better to use it in an automatic variable as: `integer, dimension(1:3) :: tmp` instead of the allocatable. This would allow the compiler to avoid the heap allocations for this small cases.

examples/nemo/scripts/utils.py
```diff
@@ -530,3 +530,22 @@ def _satisfies_minimum_region_rules(self, region: list[Node]) -> bool:
...
+                if (isinstance(dtype, ArrayType) and
+                        isinstance(arg, Operation)):
+                    try:
+                        DataNodeToTempTrans().apply(arg)
+                    except TransformationError as err:
+                        call.append_preceding_comment(
+                            f"Couldn't pull the argument {arg} to a "
+                            f"temporary due to the following error: "
+                            f"{str(err.value)}"
+                        )
```
sergisiso: Thanks that helps, but could you prefix it with something like "PSyclone Warning" and use `'{arg.debug_string()}'`, otherwise we currently get:

```
        ! Couldn't pull the argument BinaryOperation[operator:'MUL']
        ! Reference[name:'rnf']
        ! ArrayReference[name:'pts']
        ! Range[]
        ! Range[]
        ! Literal[value:'1', Scalar<INTEGER, UNDEFINED>]
        ! Reference[name:'jp_tem']
        ! Reference[name:'kmm']
        !  to a temporary due to the following error: Transformation Error: Input node's datatype is an array of unknown size, so 
!& the DataNodeToTempTrans cannot be applied. Input node was 'rnf * pts(:,:,1,jp_tem,kmm)'.
```

examples/nemo/scripts/utils.py
```diff
@@ -530,3 +530,22 @@ def _satisfies_minimum_region_rules(self, region: list[Node]) -> bool:
...
+    for call in calls:
+        if call.symbol.name == "iom_put":
+            for arg in call.arguments:
+                dtype = arg.datatype
+                if (isinstance(dtype, ArrayType) and
+                        isinstance(arg, Operation)):
+                    try:
+                        DataNodeToTempTrans().apply(arg)
+                    except TransformationError as err:
+                        call.append_preceding_comment(
```
sergisiso: Also, consider doing the same but inside the transformation with a verbose option.

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -199,6 +200,10 @@ def trans(psyir):
             if "pp_len" not in symtab:
                 symtab.add(symtab.lookup("pp_len"))
 
+        # Extract any array operations from iom_put calls to temporary
+        # expressions that can be parallelised.
+        iom_put_argument_to_temporary(subroutine.walk(Call))
+
```
sergisiso: I would prefer this to be at the top, but inside the normalise loops. I think it is still taking out implicit loops to their explicit form so they can be parallelised, which is the purpose of the normalise_loops method.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -4897,7 +4910,21 @@ def datatype(self) -> DataType:
         if isinstance(self.intrinsic.return_type, Callable):
             try:
                 return self.intrinsic.return_type(self)
+            except TypeError as err:
```
sergisiso, line 4908: I know we discussed this before but can you remind me why we don't simply use:
```
try:
      return self.intrinsic.return_type(self)
except (TypeError, AttributeError):
      return UnresolvedType()
```
I know for debugging the more explicit error could help, but from a user point of view it doesn't matter that much it would be better to get the UnresolvedType to communicate that we can't do it but let it continue.

examples/nemo/scripts/omp_gpu_trans.py
sergisiso: I also wouldn't mind preemptively calling it the method and the normalise_loop flag "hoist_argument_expressions", and just putting a comment inside saying that this is currently limited to "iom_put" (and a TODO to generalise it)

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: Same for _iparity_return_type?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -329,9 +338,40 @@ def apply(self, node: DataNode, storage_name: str = "", **kwargs):
...
+            for shape in ref.indices:
+                for ref2 in shape.walk(Reference):
+                    if isinstance(ref2, ArrayReference):
+                        hoistable = False
+            # If we can hoist the allocate, find the highest level Loop
+            # ancestor and set the schedule and position to place the
+            # allocate before this loop.
+            if hoistable:
+                loop_anc = schedule.ancestor(Loop)
+                finger = loop_anc
```
sergisiso: Maybe "cursor" instead of "finger"?

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -329,9 +338,40 @@ def apply(self, node: DataNode, storage_name: str = "", **kwargs):
...
+
+            ifblock = IfBlock.create(
+                UnaryOperation.create(
+                    UnaryOperation.Operator.NOT,
+                    allocated),
+                [intrinsic]
+            )
+            # If the shape doesn't contain array references then we can hoist
+            # the allocate statement outside of any ancestor loops.
+            hoistable = True
```
sergisiso, line 404: I see the HoistTrans is not applicable here becuase it is only for Assingments, but maybe we need a issue/TODO to generalise HoistTrans to any statement?

sergisiso (changes requested):
@LonelyCat124 The implementation is getting closer, and the generated NEMO diff seems to make sense, but it is a bit puzzling that the performance is not getting better. I will have a look with the profiler but I won't stop the PR while I do it.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah we can now, so I've removed those as they're replaced by the newer functionality.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yes, but its caught in IntrinsicCall.datatype and handled there, explicitly to catch cases where unexpected types are present when resolving datatypes of arguments we correctly found (at least the original design). I'm not sure if we gain much by moving the error handling here, as we'll always need the checks in `datatype` due to things being resolved inside lambdas.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Removed both, not sure why flake8 doesn't find those?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Its just from how the matmul return type is implemented, it always gives SIZE instead of just values, probably this can be improved at some point if we feel its worthwhile though it does make the code more messy.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: I'm not keen do it generally, as we have to decide what we think "small" should be too, for example I think you could have
```
subroutine x(input_array_that_was_allocatable_somewhere)
use size_def_mod
integer, dimension(a1, a2, a3) :: input_array_that_was_allocatable_somewhere

 ...
end subroutine
```
Where we know the static bound information in that subroutine, but the array could be very large (the bounds could even be integer defined in this subroutine I think?) and we can't stack allocate a temporary without a crash, so we went with allocate always for now. If I'm wrong let me know and I can look into it.

examples/nemo/scripts/utils.py
LonelyCat124: Done - I used .rstrip() on the debug string to avoid random end of line characters too.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
sergisiso: I don't think we need to decide what is small, that's the compiler's job. For instance Intel allocates on the heap any automatic array that is more than 10kb, other compilers have a similar heuristic.

examples/nemo/scripts/utils.py
LonelyCat124: Done, should be covered by tests too.

examples/nemo/scripts/omp_gpu_trans.py
LonelyCat124: If I understood correctly this is done.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: I think I'd prefer to be more precise as from a user point of view I think it should never reach the InternalError, so its mostly being defensiveness from a development standpoint to avoid reaching things that are unexpected, but I'm happy to change it if you would prefer.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Sure

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: We have https://github.com/stfc/PSyclone/issues/1445 already, I'll add a TODO.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: This actually makes some other stuff in the code fail, because of Structure % array accesses that fail to get through:
```
    if (
        not isinstance(arg_dt, ArrayType) or
        not isinstance(arg_dt.elemental_type, ScalarType) or
        not isinstance(arg_dt.elemental_type.intrinsic, ScalarType.Intrinsic)
    ):
```

but I'm not sure why? The arg_dt.elemental_type still returns an ArrayType. Somehow we have this as the datatype:
`Array<Array<Scalar<REAL, UNDEFINED>, shape=[10, 10]>, shape=[10, 10]>` for 
```
            "subroutine test(n,m)\n"
            "  integer :: n, m\n"
            "  type :: array_type\n"
            "      real :: array(10,10)\n"
            "  end type\n"
            "  type(array_type) :: ref\n"
            "  real :: result\n"
            "  integer :: dimension\n"
            "  result = maxval(ref%array)\n"
            "end subroutine\n"
```

Is this expected @sergisiso ? ref%array should should have a type of `Array<Scalar<Real>, shape=[10,10]>` right?

sergisiso: Yeah, `Array<Array<` should never happen

LonelyCat124: I'll try to work out whats causing the bug then, I might just have to make an issue, readd the removed stuff and then add a TODO on the relevant issue if its not straightforward.

LonelyCat124: Ok - it looks like a bug in StructureReference.datatype, the result of this section:
```
        # Walk down the structure, collecting information on any array slices
        # as we go.
        while isinstance(cursor, (StructureMember, StructureReference)):
            cursor = cursor.member
            if isinstance(cursor_type, ArrayType):
            if not isinstance(cursor_type, (UnresolvedType, UnsupportedType)):
                # Once we've hit an Unresolved/UnsupportedType the cursor_type
                # will remain set to that as we can't do any better.
                cursor_type = cursor_type.components[
                    cursor.name.lower()].datatype
            try:
                cursor_shape = _get_cursor_shape(cursor, cursor_type)
            except NotImplementedError:
                return UnresolvedType()
            if cursor_shape:
                if shape:
                    raise NotImplementedError(
                        f"Array of arrays not supported: the ultimate member "
                        f"'{cursor.name}' of the StructureAccess represents "
                        f"an array but other array notation is present in the "
                        f"full access expression: '{self.debug_string()}'")
                shape = cursor_shape
```
is `Array<Scalar<REAL, UNDEFINED>, shape=[10, 10]>` (for the datatype accessed in `tests/psyir/transformations/intrinsics/array_reduction_base_trans_test.py::test_structure_error `, and then the code does
```
        if shape:
            return ArrayType(cursor_type, shape)
```
resulting in the `ArrayType<ArrayType>` datatype. I'm not really sure how to fix this right now, so I'm going to open an issue and revert some of the changes I made yesterday with a TODO on the new issue.

LonelyCat124: Reverted and TODOs added.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Ok, if compiler's are smart enough to make that decision for us I can have a look at doing this.

LonelyCat124: I've implemented it to work for Literal sizes for now, I'm a bit scared to do it for references as we'd need to check they're safe to use and that seems a bit more complex.

LonelyCat124: Maybe its safe, but since references can change during the symbol table in the weird inline loop structures its a bit less straightforward, but the literals are an easy win at least.

LonelyCat124:
@sergisiso This is ready for another look, I am going to set ITs going again since output code changes will likely happen.

examples/nemo/scripts/omp_gpu_trans.py
```diff
@@ -190,14 +190,17 @@ def trans(psyir):
             if "pp_len" not in symtab:
                 symtab.add(symtab.lookup("pp_len"))
 
+        # Extract any array operations from iom_put calls to temporary
+        # expressions that can be parallelised.
```
sergisiso: Delete this now?

examples/nemo/scripts/utils.py
```diff
@@ -530,3 +538,23 @@ def _satisfies_minimum_region_rules(self, region: list[Node]) -> bool:
     routine_name = parent_routine.name if parent_routine else ""
     if routine_name not in PROFILING_IGNORE:
         MaximalProfilingOutsideDirectivesTrans().apply(children)
+
+
+def iom_put_argument_to_temporary(calls: list[Call]):
```
sergisiso, line 543: "param:" missing

examples/nemo/scripts/utils.py
sergisiso: Thanks, now you can delete this in favour of the option.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
```diff
@@ -276,13 +320,27 @@ def apply(self, node: DataNode, storage_name: str = "", **kwargs):
...
-            # We want to create an allocatable symbol for Array entities, so
-            # create a new datatype for the symbol and keep the
-            # datatype around for the ALLOCATE statement later.
-            allocatable_datatype = datatype
-            datatype = ArrayType(allocatable_datatype.elemental_type,
-                                 [ArrayType.Extent.DEFERRED for x in
-                                  allocatable_datatype.shape])
+            # If any of the bound information aren't static then we need
+            # to create an allocatable array.
+            is_static = True
```
sergisiso: maybe r/is_static/has_static_bounds/ to not confuse it with symbols with static interface

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: Wasn't done? There is also the same unecessary line breaks in lines 3775, 4458, 4730 and 4755

src/psyclone/psyir/nodes/intrinsic_call.py
sergisiso: Ok (it still feels not ideal that we compare against the error message, hopefully we can improve on this in the furture)

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -4897,7 +4914,21 @@ def datatype(self) -> DataType:
...
+                # If we get an invalid argument to a ScalarType constructor it
+                # means we attempted to pass either an UnresolvedType into the
+                # datatype
+                if ("ScalarType expected 'intrinsic' argument to be of type "
+                    in str(err)
+                    or "ScalarType expected 'precision' argument to be of "
+                        "type " in str(err)):
+                    return UnresolvedType()
+                # Is this reachable? Tested via monkeypatch as there may be
+                # some edge case I can't think of.
```
sergisiso: I would prefer a comment like "This should never happen, so propagate as an InternalError" instead.

sergisiso (changes requested):
@LonelyCat124 I am happy with the current implementation and bringing the remaining rough edges to separate issues/TODOs. There is just final clean up to do before merging it.

examples/nemo/scripts/omp_gpu_trans.py
LonelyCat124: Yeah.

examples/nemo/scripts/utils.py
LonelyCat124: Added. I also  have slightly changed the behaviour so I will rerun integration.

examples/nemo/scripts/utils.py

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Ah my bad, I reverted some commits to the file I think and so this was also reverted.

LonelyCat124: Fixed those now.

src/psyclone/psyir/transformations/datanode_to_temp_trans.py
LonelyCat124: Yeah thats probably better.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: Yeah, thats neater, the previous comment probably should have been labelled as FIXME, updated now.

LonelyCat124:
@sergisiso I've set integration running again, assuming no issues arise this is ready for another look. If the tests fail on treesitter related issues then I think we should wait until #3408 is merged.

sergisiso (approved):
@LonelyCat124 All comments have been addressed and this is ready for merging, but I will wait until the PR that fixes the treesitter issue is merged to proceed.

PR #3374, opened 2026-03-13
(Closes #3366) rm LFRicSymbolTable

arporter:
This branch was taken from the one for #1734 and therefore needs #3358 to be merged first.


src/psyclone/domain/lfric/lfric_builtins.py
```diff
@@ -487,8 +488,9 @@ def get_dof_loop_index_symbol(self):
         table = self.scope.symbol_table
         # The symbol representing the loop index is created in the LFRicLoop
         # constructor.
-        return table.find_or_create_integer_symbol(
-            "df", tag="dof_loop_idx")
+        return table.find_or_create(
+            "df", tag="dof_loop_idx", symbol_type=DataSymbol,
+            datatype=LFRicTypes("LFRicIntegerScalarDataType")())
```
LonelyCat124, line 492: Do we need to check with find_or_create that the datatype matches?

LonelyCat124: Same question whenever this type of call happens.

src/psyclone/transformations.py
```diff
@@ -52,8 +52,11 @@
 from psyclone import psyGen
 from psyclone.configuration import Config
 from psyclone.core import Signature, VariablesAccessMap
-from psyclone.domain.lfric import (KernCallArgList, LFRicConstants,
-                                   LFRicInvokeSchedule, LFRicKern, LFRicLoop)
+from psyclone.domain.lfric.kern_call_arg_list import KernCallArgList
```
LonelyCat124, line 55: Was this to avoid some circular import issues?

LonelyCat124 (changes requested):
Hi @arporter  - mostly looks good, I have a couple of questions about functionality because I'm not sure what happens if find_or_create finds a symbol already in the table with the wrong datatype (or if thats possible)? From what I can see I don't see any type checking in find_or_create.

I'm somewhat loathe to suggest it, but could we get more type hints in the functions you've modified if that feels reasonable? I'm not sure what our rules w.r.t adding those in is so feel free to ignore if you think its too much.

src/psyclone/domain/lfric/lfric_builtins.py
arporter: That's a good point and I was going to add some checks but then I realised that we already have a TODO for #1057 where Sergi suggests a good solution. One of us could tackle that next?

src/psyclone/transformations.py
arporter: Yes.

src/psyclone/domain/lfric/lfric_builtins.py
LonelyCat124: I don't think its a high priority to fix it right now, but if we have a quick solution to add it could be worthwhile - but as you say not something that needs fixing here.

LonelyCat124 (approved):
Looks good now, I've set integration going and assuming no issues arise I'll proceed to merge.

arporter:
The LFRic IT with nvfortran failed with a signal 11 so I've launched it a second time.

PR #3388, opened 2026-03-26
(Closes #3334) Make module-inlining mandatory for Kernel transformations

### LINKED ISSUE #3334: Make module inlining a required precursor to kernel transformations

arporter:
A second attempt at this after my git fu let me down badly.
Original changes are here: https://github.com/stfc/PSyclone/pull/3294/changes#diff-436bf7f3b5f30b4c1d6ff1f9a0a8e144f924bae78cbcc9595b30fc5954e715e0


doc/user_guide/psyclone_command.rst
```diff
@@ -98,12 +97,10 @@ by the command:
       -opsy OUTPUT_PSY_FILE
                             (psykal mode) filename of generated PSy-layer code
       -okern OUTPUT_KERNEL_PATH
-                            (psykal mode) directory in which to put transformed kernels, default
+                            (psykal mode) directory in which to put any generated kernels, default
                             is the current working directory
       -dm, --dist_mem       (psykal mode) generate distributed memory code
       -nodm, --no_dist_mem  (psykal mode) do not generate distributed memory code
-      --kernel-renaming {multiple,single}
-                            (psykal mode) naming scheme to use when re-naming transformed kernels
```
sergisiso, line 106: `psyclone -h` still has this flag because the changes to generator.py has not been ported to this second take branch.

There are also other kernel-renaming mentions here:

```
582:two different kernel-renaming schemes: "multiple" and "single"
583:(specified via the ``--kernel-renaming`` command-line flag). In the

src/psyclone/configuration.py
742:                            kernel-renaming scheme.

src/psyclone/generator.py
239:    :raises GenerationError: if an invalid kernel-renaming scheme is specified.
276:            f"Invalid kernel-renaming scheme supplied: {str(verr)}") from verr
534:        '--kernel-renaming', default="multiple",

src/psyclone/tests/generator_test.py
1681:    kernel-renaming scheme is supplied.
1693:    assert "Invalid kernel-renaming scheme supplied" in str(err.value)
```

src/psyclone/transformations.py
```diff
@@ -2308,31 +2305,17 @@ def validate(self, node, options=None):
...
-            try:
-                kernel.check_outer_scope_accesses(
-                    node, "Kernel",
-                    permit_unresolved=False,
-                    ignore_non_data_accesses=True)
-            except SymbolError as err:
-                raise TransformationError(
-                    f"Cannot apply {self.name} to Kernel '{node.name}' "
-                    f"because it accesses data from its outer scope: "
-                    f"{err.value}") from err
```
sergisiso, line 2329: I am not sure we can delete this.

You added the comment "This also implies that there are no unqualified imports or undeclared symbols.", do you mean that module-inlining would have failed? Can we not have a pre-existing local routine with unqualified imports?

Or a kernel that was module-inlined, then another transformation or script adds an unqualifired import, then we apply this.

src/psyclone/tests/domain/gocean/transformations/gocean_move_iteration_boundaries_inside_kernel_trans_test.py
```diff
@@ -145,36 +151,42 @@ def test_go_move_iteration_boundaries_inside_kernel_trans():
     # - It has the boundary symbol as kernel arguments
     assert isinstance(kschedule.symbol_table.lookup("xstart").interface,
                       ArgumentInterface)
-    assert isinstance(kschedule.symbol_table.lookup("xstop").interface,
+    assert isinstance(kschedule.symbol_table.lookup("xstop_1").interface,
```
sergisiso, line 154: Why did this name change?

src/psyclone/tests/psyir/transformations/callee_transformation_mixin_test.py
```diff
@@ -0,0 +1,146 @@
...
+    '''
+    A test 'transformation' class that inherits from CalleeTransformationMixin.
+    '''
+    def validate(self, node, **kwargs):
+        '''
+        '''
+        self._check_callee_implementation_is_local(node)
+
+    def apply(self, node, **kwargs):
+        '''
```
sergisiso, line 58: Maybe add "Do nothing" here and "Call CalleeTransformationMixin validation" in the validate docstring.

sergisiso (changes requested):
@arporter Most comments raised in the other PR are now resolved. There is just some minor clean up/clarfications to do.

doc/user_guide/psyclone_command.rst
arporter: Oops, thanks for spotting those. It turns out we had two sections on "Kernel Transformations", one in the UG and one in the DG. The latter was quite good so I've simply moved that to the UG as it's not really developer information (although that statement applies to quite a lot of the DG).

src/psyclone/transformations.py
arporter: Oh yes, good point. Re-instated.

arporter: and in fact my comment was wrong: we do permit Kernels with wildcard imports to be module-inlined as we just bring all the wildcard imports inside the Kernel.

src/psyclone/tests/domain/gocean/transformations/gocean_move_iteration_boundaries_inside_kernel_trans_test.py
arporter: We're doing this transformation after module-inlining now and so the 'create new symbol' spots the `xstop` in the outer scope and makes a new name.

src/psyclone/tests/psyir/transformations/callee_transformation_mixin_test.py
arporter: Oops, yes.

arporter:
Ready for another look now, I've just fired-off the ITs again.

sergisiso (approved):
@arporter All comments have been addressed and the integration test pass. I could not find more mentions of the old functionallity. Thanks for the multiple takes on this, it's a good simplification of the codebase and the documentation.

PR #3389, opened 2026-03-30
Updates for inlining with common variables

schreiberx:


src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1007,6 +1042,33 @@ def _handle_symbol_clash(self, old_sym, other_table):
...
             return
 
+        if old_sym.is_commonblock and self_sym.is_commonblock:
+            return
+
+        from psyclone.psyir.symbols.datatypes import UnsupportedFortranType
+        if (isinstance(old_sym.datatype, UnsupportedFortranType) and
+                isinstance(self_sym.datatype, UnsupportedFortranType)):
+            if old_sym.datatype.declaration == self_sym.datatype.declaration:
+                # Identical COMMON-block markers – already present in self.
```
arporter: This comment isn't right - they are just the same unsupported type, not necessarily a common block.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -947,13 +953,42 @@ def _add_symbols_from_table(self, other_table, symbols_to_skip=()):
             already been updated to refer to a Container in this table.
 
         '''
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.symbols.datatypes import UnsupportedFortranType
+
+        # Names (normalised) of commonblock variables already in this table.
```
arporter: "...of variables with a 'common block' interface already in this table."

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -947,13 +953,42 @@ def _add_symbols_from_table(self, other_table, symbols_to_skip=()):
...
+        # Used below to decide whether a _PSYCLONE_INTERNAL_COMMONBLOCK marker
+        # from other_table is a duplicate that must be skipped.
+        self_cb_names = frozenset(
+            self._normalize(s.name)
+            for s in self.symbols if s.is_commonblock)
+
+        # Normalised names of commonblock variables in other_table.  Any
```
arporter: Same change here.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -947,13 +953,42 @@ def _add_symbols_from_table(self, other_table, symbols_to_skip=()):
...
+        # overlap with self_cb_names means the COMMON block is already
```
arporter: Worth extending this comment to say that the same variable cannot appear in more than one common block so if they have the same name and a 'common block' interface then they must originate from the same common. I'm slightly uneasy as the actual names of symbols within a common block don't have to match: just their order and type:

Image

So, two symbols from different scopes but with the same name and a common block interface are not guaranteed to be referring to the same thing.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -947,13 +953,42 @@ def _add_symbols_from_table(self, other_table, symbols_to_skip=()):
...
                 # We've dealt with Container symbols in _add_container_symbols.
                 continue
 
+            # Skip _PSYCLONE_INTERNAL_COMMONBLOCK marker symbols when any of
+            # the COMMON-block variables they represent are already present in
+            # this table.  Adding such a marker would produce a duplicate
+            # COMMON declaration and a Fortran compile error like
+            # "Symbol 'x' is already in a COMMON block".
+            if (cb_overlap and
+                    self._normalize(old_sym.name).startswith(
```
arporter: I don't think this code should be here. Any special action should be deferred to the `_handle_symbol_clash` method (which in fact I see has been updated).

src/psyclone/psyir/symbols/symbol_table.py
arporter: Consider:
```fortran
integer :: var1, var2
common /block1/ var1, var2
```
and (in a different scope):
```fortran
real :: var1, var2
common /block2/ var1, var2
```
as written, the code would assume that it is safe to ignore this clash because both var1 and var2 have a 'common block' interface. Unfortunately we currently make no attempt to capture the name of a common block - the whole `common /....` statement is just reproduced blindly.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1007,6 +1042,33 @@ def _handle_symbol_clash(self, old_sym, other_table):
             # refer to the same thing and we don't have to do anything.
             return
 
+        if old_sym.is_commonblock and self_sym.is_commonblock:
```
arporter: As discussed above, unfortunately this check is insufficient but it is all that PSyclone is capable of supporting at the moment. To do better, we would need to extend CommonBlockInterface so that it holds the name of the common block. (Note that this name is not in the same space as the datasymbols themselves so can't just be a normal Symbol.)

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1007,6 +1042,33 @@ def _handle_symbol_clash(self, old_sym, other_table):
...
+            # Markers have different declarations.  Skip the incoming one only
+            # if its COMMON-block name(s) overlap with those already in self:
+            # that means the block is already declared and adding a second
+            # marker for it would produce a "Symbol X is already in a COMMON
+            # block" compile error.  If the block names are different this is
+            # a genuinely new COMMON block and we fall through to the
+            # rename-and-add path below.
+            if self._normalize(old_sym.name).startswith(
+                    "_psyclone_internal_commonblock"):
+                import re  # pylint: disable=import-outside-toplevel
```
arporter: OK, this is definitely a common block at this point. It would be nice not to resort to a regex but the alternative (of exposing the parse tree) is even less good. The `import re` needs to be moved up to the top of the module and the regex needs commenting to explain what it's extracting - presumably the name(s) of the common blocks?

src/psyclone/psyir/symbols/symbol_table.py
arporter: In fact, this is quite confusing. I think what we have here is that the two symbols that have a name collision both have a 'common block' interface and *also* happen to be of UnsupportedFortranType. I think (but could be wrong) that it would also be possible for them to be of a supported type. Unfortunately, we have no way of knowing whether they are both the (same) variable from the same common block.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1007,6 +1042,33 @@ def _handle_symbol_clash(self, old_sym, other_table):
...
+                _blk_re = re.compile(r"/\s*(\w*)\s*/", re.IGNORECASE)
+                old_blocks = set(_blk_re.findall(
+                    old_sym.datatype.declaration))
+                self_blocks = set(_blk_re.findall(
+                    self_sym.datatype.declaration))
+                if old_blocks & self_blocks:
```
arporter: OK, so if the two unsupported datatypes are defining the same common blocks then we accept the clash with no further action? This part seems safe.

arporter (commented):
Thanks @schreiberx, I learnt (or reminded myself) quite a lot about common blocks :-)
Unfortunately, it is currently impossible to do what you want safely as PSyclone does not capture which common a variable with a 'common block' interface is associated with. Since the names of common blocks are global and don't live in the same space as variables, I think extending PSyclone to at least capture those names and the symbols associated with them should be relatively straightforward. The code in question is:
https://github.com/stfc/PSyclone/blob/8dd90f56f18b359082cb70b79600d037138b8a36/src/psyclone/psyir/frontend/fparser2.py#L2771-L2798
Given that we've learnt to our cost that safety needs to come first, I suggest we need to extend PSyclone's handling of common a little before we can safely handle common blocks when merging tables.

src/psyclone/psyir/symbols/symbol_table.py
schreiberx: done

arporter:
Thanks Martin, if you could fix the linting so that the test suite runs through then I'll take another look.

schreiberx:
linting fixed.

arporter:
Hi @schreiberx, please could you fix the test failure and then make sure the coverage is OK too?

schreiberx:
Hi Andy, I fixed this and added some new code to address another issue with Croco.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1000,11 +1029,45 @@ def _handle_symbol_clash(self, old_sym, other_table):
 
         self_sym = self.lookup(old_sym.name)
         if old_sym.is_unresolved and self_sym.is_unresolved:
+            # Update after fixing issue #3392
```
arporter: Since this routine assumes `check_for_clashes` has been called successfully, this comment doesn't belong here. However, we could do with a comment at 1039 to say that `check_for_clashes` must have determined that the two variables are from the same common block.

src/psyclone/psyir/symbols/symbol_table.py
arporter: I'm afraid I'm still confused here. The check at L1038 means that only one or none of the clashing symbols has a common-block interface. Therefore, old_sym and self_sym can't both be common-block markers and in fact both might not be? Am I misunderstanding something?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -945,13 +953,34 @@ def _add_symbols_from_table(self, other_table, symbols_to_skip=()):
...
         '''
+
         for old_sym in other_table.symbols:
 
             if old_sym in symbols_to_skip or isinstance(old_sym,
                                                         ContainerSymbol):
                 # We've dealt with Container symbols in _add_container_symbols.
                 continue
 
+            # Avoid duplicate COMMON-block marker symbols when multiple
```
arporter: This new logic belongs in `_handle_symbol_clash` - that's where we determine whether or not to take any action to resolve a clash.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -219,6 +227,15 @@ def apply(self,
         #     just delete the if statement.
         self._optional_arg_eliminate_ifblock_if_const_condition(routine)
 
+        # If parameter_cloning is disabled, identify duplicate constant
+        # (PARAMETER) symbols and redirect their references *before* the
+        # routine body is extracted, so that the extracted statements already
+        # carry references to the call-site symbols.
+        extra_skip: List[DataSymbol] = []
```
arporter: You can use "list" rather than "List".

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -219,6 +227,15 @@ def apply(self,
...
+        if not parameter_cloning:
+            extra_skip = self._redirect_duplicate_parameters(
+                table, routine, routine_table)
```
arporter: Please don't pass `routine_table` - it must always be the one that belongs to `routine`.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+
+        A constant is only considered a duplicate when every routine-local
+        symbol referenced inside its initial-value expression is itself a
+        confirmed duplicate.  This prevents false positives for expressions
+        like ``negflag = .NOT. flag`` when ``flag`` has different values in
+        the caller and the callee (the names would match but the semantics
+        would differ).
+
+        :param table: the call-site symbol table.
+        :type table: :py:class:`psyclone.psyir.symbols.SymbolTable`
```
arporter: Please replace the ":type ..." entries here with type-hints on the arguments.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+        :param routine: the (copy of the) routine being inlined.
+        :type routine: :py:class:`psyclone.psyir.nodes.Routine`
+        :param routine_table: the symbol table of the routine copy.
```
arporter: As mentioned earlier, pls remove this argument.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+        :type routine_table: :py:class:`psyclone.psyir.symbols.SymbolTable`
+
+        :returns: the list of routine symbols that are duplicates of
```
arporter: s/routine//

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+            call-site constants and should be excluded from the subsequent
+            table merge.
+        :rtype: List[:py:class:`psyclone.psyir.symbols.DataSymbol`]
```
arporter: Type-hint instead please and you can just use `list`.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+
+        '''
+        # The names of all local data symbols in the routine table (used to
+        # identify references that point to routine-local constants).
+        routine_local_names = {
+            s.name.lower() for s in routine_table.datasymbols
+            if not s.is_argument
```
arporter: This will include symbols imported from modules which might not be what you want? You probably want to check for `is_automatic` instead.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+            tsym = table.lookup(rsym.name)
+            # Update all References in the routine body.
+            routine.replace_symbols_using(tsym)
+            # Update any references to rsym embedded in the datatypes or
+            # initial-value expressions of other symbols in routine_table.
+            for sym in routine_table.symbols:
+                if sym is rsym:
+                    continue
+                sym.replace_symbols_using(tsym)
+                if hasattr(sym, 'datatype') and sym.datatype is not None:
```
arporter: You shouldn't need this part as `TypedSymbol.replace_symbols_using()` does this for you. (L265 of typed_symbol.py)

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+                )
+            ]
+            for name in to_remove:
+                del candidates[name]
+            if to_remove:
+                changed = True
+
+        duplicates: List[DataSymbol] = list(candidates.values())
+
+        # Redirect all references from duplicate routine symbols to their
```
arporter: "...duplicate symbols in the routine to..." (just to avoid confusion with RoutineSymbols)

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -329,6 +347,116 @@ def apply(self,
...
+                    and dep.name.lower() not in candidates
+                    for dep in rsym.initial_value.get_all_accessed_symbols()
+                )
+            ]
+            for name in to_remove:
+                del candidates[name]
+            if to_remove:
+                changed = True
+
+        duplicates: List[DataSymbol] = list(candidates.values())
```
arporter: s/List/list/

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -1325,6 +1325,90 @@ def test_handle_symbol_clash_imported_symbols():
...
+    # Include a non-marker symbol so the marker filter condition also takes
+    # the false branch while scanning existing symbols.
+    table1.add(symbols.DataSymbol("plain", symbols.INTEGER_TYPE))
+    # Existing marker with different internal number but same block as incoming
+    # marker. This exercises the scan of all existing markers in table1.
+    table1.add(symbols.DataSymbol(
+        "_PSYCLONE_INTERNAL_COMMONBLOCK_2",
+        symbols.UnsupportedFortranType("common /overlap/ c")))
+    table2.add(symbols.DataSymbol(
+        marker_name, symbols.UnsupportedFortranType("common /overlap/ a")))
```
arporter: The trouble here is that `a` and `c` in fact refer to the same thing and, pending #3392, we can't tell that. Therefore we must refuse to handle such a clash.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    trans.apply(calls[0])
+
+    result = fortran_writer(caller)
+    # /ocean_zeta/ must appear exactly once – not duplicated.
+    assert result.count("COMMON /ocean_zeta/") == 1
+    # The extra block from the caller must be preserved.
+    assert result.count("COMMON /comm_setup_mpi1/") == 1
+    assert "zeta" in result
+    assert "lmmpi" in result
+
```
arporter, line 3072: Good tests, thanks. Please could you add one to show that we refuse to do inlining if there are overlapping common-block declarations that use different variable names, e.g.:
```fortran
COMMON /ocean/ depth
```
and
```fortran
COMMON /ocean/ height
```

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    result = fortran_writer(caller)
+    # Each COMMON block must appear exactly once.
+    assert result.count("COMMON /ocean_zeta/") == 1
+    assert result.count("COMMON /ocean_ubar/") == 1
+    assert result.count("COMMON /ocean_vbar/") == 1
+    # All three variables must still be present.
+    assert "zeta" in result
+    assert "ubar" in result
+    assert "vbar" in result
```
arporter, line 2981: Please add a compilation check, if possible (see other tests for examples).

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    trans.apply(calls[0])
+    calls = caller.walk(Call)
+    trans.apply(calls[0])
+
+    result = fortran_writer(caller)
+    # Exactly one COMMON declaration must appear.
+    assert result.count("COMMON /blk/") == 1
+    # Both variables must still be present.
+    assert "volume" in result
+    assert "lmmpi" in result
```
arporter, line 2919: Please add a compilation check if possible.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    trans = InlineTrans()
+    calls = caller.walk(Call)
+    trans.apply(calls[0])
+
+    result = fortran_writer(caller)
+    # /ocean_zeta/ must appear exactly once – not duplicated.
+    assert result.count("COMMON /ocean_zeta/") == 1
+    # The extra block from the caller must be preserved.
+    assert result.count("COMMON /comm_setup_mpi1/") == 1
+    assert "zeta" in result
```
arporter, line 3025: Please add a compilation check.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    bar = psyir.walk(Routine)[0]
+    foo = psyir.walk(Routine)[1]
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo)
+
+    result = fortran_writer(bar)
+    # With cloning enabled the inlined constant must appear at least once;
+    # it may be renamed to avoid the clash.
+    assert "constval" in result
```
arporter, line 3153: Please add a check that the code compiles.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    foo = psyir.walk(Routine)[1]
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # constval should be declared exactly once (no duplicate parameter).
+    assert result.count("parameter :: constval") == 1
+    # The inlined assignment should still use constval correctly.
+    assert "constval" in result
```
arporter, line 3188: Please add a compilation check.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    bar = psyir.walk(Routine)[0]
+    foo = psyir.walk(Routine)[1]
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # Both constant declarations must survive since they have different values.
+    assert result.count("constval") >= 2
+
```
arporter, line 3222: Compilation check please.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    psyir = fortran_reader.psyir_from_source(code)
+    bar = psyir.walk(Routine)[0]
+    foo = psyir.walk(Routine)[1]
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # constval from foo must be added to bar because bar didn't have it.
+    assert "constval" in result
```
arporter, line 3251: Compilation check pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # n should appear only once as a parameter declaration.
+    assert result.count(", parameter ::", result.lower().find("n =")) <= 1 \
+        or result.count("n = 5") == 1
+    # The inlined array tmp should still be present and use n.
+    assert "tmp" in result
+    assert "n" in result
```
arporter, line 3289: Compilation pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    foo = psyir.walk(Routine)[1]
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # shared must be declared exactly once (no duplicate parameter).
+    assert result.count("parameter :: shared") == 1
+    # local_only is unique to foo, so it must be added to bar.
+    assert "local_only" in result
```
arporter, line 3324: Compilation pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # Neither base_val nor constval should be duplicated.
+    assert result.count("parameter :: constval") == 1
+    assert result.count("parameter :: base_val") == 1
+    # The inlined body should still reference constval.
+    assert "constval" in result
```
arporter, line 3362: Compilation pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    call = bar.walk(Call)[0]
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # constval has different values in bar and foo, so both must appear.
+    assert result.count("parameter :: constval") >= 2 or (
+        "constval" in result and "constval_1" in result)
+    # base_val is identical and should be deduplicated.
+    assert result.count("parameter :: base_val") == 1
```
arporter, line 3399: Compilation pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+
+
+def test_apply_parameter_cloning_false_unary_op_identical(fortran_reader,
```
arporter: I don't think you need this test - it's really just testing the `__eq__` of `DataNode`.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+
+    InlineTrans().apply(call, routine=foo, parameter_cloning=False)
+
+    result = fortran_writer(bar)
+    # flag has different values so both must appear (foo's renamed).
+    assert result.count("parameter :: flag") >= 2 or "flag_1" in result
+    # negflag depends on flag which differs, so foo's negflag must also
+    # appear (renamed), and the inlined if must use foo's (renamed) negflag.
+    assert "negflag_1" in result
+    assert "if (negflag_1)" in result
```
arporter: Compilation check pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    assert """\
+subroutine bar(b)
+  integer, parameter :: constval_1 = 10
+  integer :: b
+  integer :: constval
+
+  constval = 7
+  b = constval_1
+
+end subroutine bar""" in result
```
arporter, line 3530: Pls check compilation too.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+    # deduplicated.
+    assert """\
+subroutine bar(b)
+  integer, parameter :: constval = 10
+  real, parameter :: constval_1 = 10.0
+  integer :: b
+
+  b = INT(constval_1)
+
+end subroutine bar""" in result
```
arporter, line 3574: Compilation pls.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -2843,3 +2874,588 @@ def test_apply_array_access_check_unresolved_override_option(
...
+contains
+  subroutine bar(b)
+    logical, parameter :: flag = .true.
+    logical, parameter :: negflag = .not. flag
+    integer :: b
+    call foo(b)
+  end subroutine bar
+  subroutine foo(a)
+    logical, parameter :: flag = .false.
+    logical, parameter :: negflag = .not. flag
```
arporter: Worth doing a similar test where a kind parameter is duplicated but different between the call site and the callee, e.g.:
```fortran
integer, parameter :: wp = kind(1.0d0)
real(kind=wp), parameter :: pi = 3.14592
```
and the same in the callee but with:
```fortran
integer, parameter :: wp = kind(1.0)
```

arporter (changes requested):
Nice, thanks Martin. It's impressive to see what can be done now.
There's a bit more code moving to do and I'm still a bit worried that we'll let through overlapping common-block declarations that use different local variable names.
Apart from that, it would be great to add compilation checks to all of the new tests you've added. You need to:
 * Add the `tmp_path` test fixture;
 * `output = fortran_writer(psyir)`
 * Add a `Compile(tmp_path).string_compiles(output)`

schreiberx:
Thanks for your feedback. I'm trying to summarize all changes briefly:
- The compilation feature was added to most of the inlining test cases. This requires running with `pytest --compile`.
- I moved almost all `common`-related parts in the symbol table into a dedicated function. We need to treat "common" variables partly differently and moving this handling into a separate function was IMHO the best solution to keep the code clearer.
- I added also further support / test cases to handle merges if the variable names aren't matching. They don't need to match; they just need to match the type.

schreiberx:
@arporter Next round :-)

schreiberx:
There is a new version available, but I'm not sure whether it's better than this one.
You can find it here: https://github.com/stfc/PSyclone/pull/3444
The new version directly supports common statements in the symboltable rather than using workarounds.

PR #3399, opened 2026-04-09
(Towards #3367) def use chain update for assignments.

LonelyCat124:
I now have this at a stage that it can handle multiple inputs and behave somewhat sensibly. I haven't yet finished, as it currently requires inputs to be members of the same parent, but I think we want them to simply share an ancestor statement.

There are a few remaining FIXMEs for things that I don't know what they were preventing previously and didn't cause any previous test failures, but presumably were things I thought mattered previously (though I did also just find some code that didn't matter so...). I need to do more testing for multi-inputs, but I think I might do that through Statement.next_accesses or some similar option.


LonelyCat124:
I resolved the last of the FIXMEs. I need to do some tests with loops and things where I check I get the same results for Assignment.next_accesses/previous_accesses that I would calling it for all of the References individually.

LonelyCat124:
Having an issue with this code:
```
subroutine test
    integer :: a, b
    logical :: x
    a = b
    if (x) then
        a = b + a
    else
        b = a * b
    end if
    a = 1
    b = 1
    end subroutine test
```

For the single reference case, it doesn't find b=1 for a forward check for the original `b`  Reference, and I can't quite work out the issue. There seems to be something with the `b = a * b` accesses seemingly not belonging to a control flow node, so the DUCs are assuming it must always happens and ends the chain, which is incorrect. I'm trying to sort this out but its a bit confused (as far as I can work out it also doesn't happen for the `a` access which is super weird.

LonelyCat124:
Fixed the previous case, but immediately created another non-equivalent case where I do
```
a = b
b = 2
...
```
as the def use chains output for b when the full statement is input goes past the `b = 2` statement incorrectly.

LonelyCat124:
I fixed the above case going forwards, so things are looking like they're getting there. I suspect there will be issues going backwards that I'll need to match as well.

LonelyCat124:
NEMO ITs failing in the async runs. I need to manually look at whats happening - I think barriers are now being missed, which if there's not a simple solution to it might need a full DUC rewrite 😢

LonelyCat124:
The failure is pretty confusing. Somehow a directive has a next_dependency computed of the if statement that contains it without the if statement being in an ancestor loop, which isn't making sense to me. I'll try to work it out.

Looks like we have
```
if cond:
   a = 2
else: 
   a = 4
endif
```

And for some reason its finding the a=4 as a dependency for a = 2. I need to see if I can create a standalone replicator.

Edit: This basic case has no issue, I need to dive into it more.

LonelyCat124:
Couldn't work out the issue from working with NEMO, am creating a local test that just follows the NEMO5 script and will try to replicate my issues.

LonelyCat124:
I have a replicator locally (though its mostly just an entire nemo subroutine) with the relevant reference that is causing an issue, but I'm now at the "trying to remember what is meant to be happening" phase of the DUCs, but there is very clearly a subchain created on its ancestor if/else statement, where it must be searching in the else part incorrectly (there's no ancestor loop), so I need to track down which part of the logic creates that chain and why (when it clearly never did before (?)).

LonelyCat124:
I'm still not totally certain whats going wrong at this stage. I'm definitely getting something like this:

```
if (...) 
    ...
        if (condition including ref(1,2,3,4)) then
                     ref(1,2,3,4) = ...
         end if
    ...
else
   ...
    loops
        ref(a,b,c,d) = ...
   ....
end if
```

where forward_accesses from either the reference in the original if condition or the if condition body points to the `ref(a,b,c,d)` access as a forward access, which is both definitely wrong and wasn't possible previously.

I think I really need to make a simple replicator as working with a full NEMO function is a bit of a nightmare.

Edit: Have a standalone replicator now thankfully, looks like some weird behaviour with nested if statements, I need to try it with current master as well and check that its a new bug.

Edit 2: This is a new bug so I need to try to work out why these sections of code are appearing in the DUCs wheras before they didn't. I did modify some code to do with how the basic blocks and the control flow nodes fit together, so I probably need to see how those modifications break things.

LonelyCat124:
Found the source of this bug now, I was a bit aggressive in using `ref.copy()` instead of `ref`, which led to some errors when checking if the nodes were in the if part of an if statement. I'm going to tidy up the test I made and then run ITs again.

LonelyCat124:
I think this is ready for a first look now @arporter @sergisiso

src/psyclone/psyir/nodes/assignment.py
```diff
@@ -243,3 +244,39 @@ def is_literal_assignment(self):
 
         '''
         return isinstance(self.rhs, Literal)
+
+    def previous_accesses(self) -> dict[Signature, list[Node]]:
+        '''
+        :returns: the nodes accessing the same symbols directly before this
```
arporter, line 250: "the nodes containing the previous accesses of the symbols accessed within this node. It can be ..."

src/psyclone/psyir/nodes/assignment.py
```diff
@@ -243,3 +244,39 @@ def is_literal_assignment(self):
...
+                refs.append(ref)
+        # Avoid circular import
+        # pylint: disable=import-outside-toplevel
+        from psyclone.psyir.tools import DefinitionUseChain
+        chain = DefinitionUseChain(refs)
+        return chain.find_backward_accesses()
+
+    def next_accesses(self) -> dict[Signature, list[Node]]:
+        '''
+        :returns: the nodes accessing the same symbols directly after this
```
arporter, line 268: Similar change to this docstring please.

src/psyclone/psyir/nodes/reference.py
arporter, line 254: You can remove thsi line now :-)

src/psyclone/psyir/nodes/reference.py
```diff
@@ -245,10 +256,11 @@ def previous_accesses(self):
         # Avoid circular import
         # pylint: disable=import-outside-toplevel
         from psyclone.psyir.tools import DefinitionUseChain
-        chain = DefinitionUseChain(self)
-        return chain.find_backward_accesses()
+        chain = DefinitionUseChain([self])
```
arporter, line 259: It's a small thing but would it be nicer if you could provide the DUC with either a single Reference or a list?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -79,25 +81,57 @@ class DefinitionUseChain:
...
     def __init__(
         self,
-        reference: Reference,
+        references: list[Reference],
         control_flow_region: Iterable[Node] = (),
         start_point: Optional[int] = None,
         stop_point: Optional[int] = None,
     ):
-        if not isinstance(reference, Reference):
+        if not isinstance(references, list):
```
arporter, line 94: I think a small tweak here to make a list of Reference if passed a single Reference?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -79,25 +81,57 @@ class DefinitionUseChain:
...
+                )
+        # We need all the references to have the same ancestor Schedule and
+        # belong to the same child of the Schedule.
+        # Skip this check if we only have 1 input.
+        if len(references) > 1:
+            parent = references[0].ancestor(Schedule)
+            # Skip this check for detached nodes, since we get copies
+            # provided to the recursive calls.
+            if parent:
+                parent_path = references[0].path_from(parent)[0]
```
arporter, line 115: For clarity, perhaps s/parent_path/parent_idx/ since it holds an index (integer)?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -79,25 +81,57 @@ class DefinitionUseChain:
...
+                for ref in references:
+                    if (ref.ancestor(Schedule) is not parent or
+                            ref.path_from(parent)[0] != parent_path):
+                        raise InternalError(
+                            "All references provided into a "
+                            "DefinitionUseChain must have the same parent in "
+                            "the ancestor Schedule."
+                            )
+        # Make a copy of the list so we can modify it.
+        self._references = [ref for ref in references]
```
arporter, line 125: I think you can just do `... = references[:]`. Also, worth specifying that it's a "shallow" copy?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -135,34 +169,40 @@ def __init__(
...
+            self._killed[sig] = []
+            self._reaches[sig] = []
 
     @property
-    def uses(self) -> list[Node]:
+    def uses(self) -> dict[list[Node]]:
         """
-        :returns: the list of nodes using the value that the referenced symbol
-                  has before it is reassigned.
+        :returns: the lists of nodes using the value that the referenced
```
arporter, line 188: "a map holding, for each referenced Symbol, the list of nodes that use the value that the Symbol had before it is reassigned" ?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -135,34 +169,40 @@ def __init__(
...
+                  symbols has before it is reassigned.
         """
         return self._uses
 
     @property
-    def defsout(self) -> list[Node]:
+    def defsout(self) -> dict[list[Node]]:
         """
-        :returns: the list of nodes that reach the end of the block without
+        :returns: the lists of nodes that reach the end of the block without
```
arporter, line 196: Similar change here?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -135,34 +169,40 @@ def __init__(
...
                   of this block.
         """
         return self._defsout
 
     @property
-    def killed(self) -> list[Node]:
+    def killed(self) -> dict[list[Node]]:
         """
-        :returns: the list of nodes that represent the last use of an assigned
+        :returns: the lists of nodes that represent the last use of an assigned
```
arporter, line 205: and here.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -274,25 +316,32 @@ def find_forward_accesses(self) -> list[Node]:
...
-                    else:
+                    elif (all([ref is not ancestor.lhs for
+                               ref in self._references])):
                         # Add the lhs as a potential basic block with
-                        # different start and stop positions.
+                        # different start and stop positions, but don't
+                        # include the lhs if the lhs is present.
                         chain = DefinitionUseChain(
-                            self._reference,
+                            [ref.copy() for ref in self._references if
```
arporter, line 337: If I'm reading this correctly, you don't need the `if ref is not ancestor.lhs` here because the `elif` above has already checked that none of `self._references` is `ancestor.lhs`?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -274,25 +316,32 @@ def find_forward_accesses(self) -> list[Node]:
...
+                             ref is not ancestor.lhs],
                             [ancestor.lhs],
                             start_point=ancestor.lhs.abs_position - 1,
                             stop_point=ancestor.lhs.abs_position + 1,
                         )
-                        control_flow_nodes.append(None)
+                        index = len(chains)
+                        control_flow_nodes.insert(index, None)
```
arporter, line 344: How come you don't just append `None` and then `chain` or even, `control_flow_nodes.extend([None, chain])`?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -306,12 +355,13 @@ def find_forward_accesses(self) -> list[Node]:
                 if len(block) == 0:
                     continue
                 chain = DefinitionUseChain(
-                    self._reference,
+                    [ref for ref in self._references],
```
arporter, line 358: Just `self._references[:]` I think?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -320,15 +370,37 @@ def find_forward_accesses(self) -> list[Node]:
...
-                        # Add unique references to reaches. Since we're not
-                        # in a control flow region, we can't have added
-                        # these references into the reaches array yet so
-                        # they're guaranteed to be unique.
-                        self._reaches.append(ref)
-                    # If we have a defsout in the chain then we can stop as we
-                    # will never get past the write as its not conditional.
-                    if len(chain.defsout) > 0:
+                    # Some signatures may already have been removed by being
+                    # killed, so we only add those if they've not already been
```
arporter, line 374: "...those that haven't already been killed." if that's right?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -344,38 +416,49 @@ def find_forward_accesses(self) -> list[Node]:
                     # if the variable is the same symbol as the _reference.
```
arporter, line 416: This needs updating now that _references is a list.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -389,37 +472,46 @@ def find_forward_accesses(self) -> list[Node]:
...
+                for ref in self._uses[sig]:
+                    self._reaches[sig].append(ref)
+                # If this block doesn't kill any accesses, then we add
+                # the defsout into the reaches array.
+                if len(self.killed[sig]) == 0:
+                    for ref in self._defsout[sig]:
+                        self._reaches[sig].append(ref)
+                else:
+                    # If this block killed any accesses, then the first element
+                    # of the killed writes is the access access that we're
```
arporter, line 512: s/access access/access/

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -478,85 +572,95 @@ def _compute_forward_uses(self, basic_block_list: list[Node]):
...
+                            # Note that this assumes two symbols are not
+                            # aliases of each other.
+                            continue
+                        if reference.is_pure:
+                            # Pure subroutines only touch their arguments, so
+                            # we'll catch the arguments that are passed into
+                            # the call later as References.
+                            continue
+                        # For now just assume calls are bad if we have a
+                        # non-local variable and we treat them as though they
```
arporter, line 608: "...variable: we treat them..."

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -907,8 +1032,10 @@ def find_backward_accesses(self) -> list[Node]:
                 # statement.
                 if len(block) == 0:
                     continue
+                # Create a copy of the list as it can modify elements
+                # in the list.
                 chain = DefinitionUseChain(
-                    self._reference,
+                    [ref for ref in self._references],
```
arporter, line 1038: `self._references[:]`

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -936,18 +1063,20 @@ def find_backward_accesses(self) -> list[Node]:
...
                     # the parents of the reference.
-                    if self._reference.ancestor(Assignment) is not None:
-                        sub_start_point = self._reference.ancestor(
+                    if self._references[0].ancestor(Assignment) is not None:
+                        sub_start_point = self._references[0].ancestor(
                             Assignment
                         ).abs_position
                     else:
-                        sub_start_point = self._reference.abs_position
+                        sub_start_point = max(list(
```
arporter, line 1071: Should the "start" point be the min absolute position?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -194,21 +234,23 @@ def find_forward_accesses(self) -> list[Node]:
...
         # Setup the start and stop positions
         save_start_position = self._start_point
         save_stop_position = self._stop_point
         # If there is no set start point, then we look for all
         # accesses after the Reference.
         if self._start_point is None:
-            self._start_point = self._reference_abs_pos
+            # Find the highest abs position, as all of these are
+            # contained in the same parent.
+            self._start_point = max(list(self._references_abs_pos.values()))
```
arporter, line 253: As in, we only start the search *after* the last of the provided references?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -879,15 +1002,17 @@ def find_backward_accesses(self) -> list[Node]:
...
         # Setup the start and stop positions
         save_start_position = self._start_point
         save_stop_position = self._stop_point
         # If there is no set start point, then we look for all
         # accesses after the Reference.
         if self._stop_point is None:
-            self._stop_point = self._reference_abs_pos
+            # Find the max abs position, as all of these are
+            # contained in the same parent.
+            self._stop_point = min(list(self._references_abs_pos.values()))
```
arporter, line 1015: You use `min` but say "max" in the comment :-)

arporter: In fact, as with my comments on the start_point, is `min` correct?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -1030,67 +1172,79 @@ def find_backward_accesses(self) -> list[Node]:
...
+                for ref in self._uses[sig]:
+                    self._reaches[sig].append(ref)
+                # If this block doesn't kill any accesses, then we add
+                # the defsout into the reaches array.
+                if len(self.killed[sig]) == 0:
+                    for ref in self._defsout[sig]:
+                        self._reaches[sig].append(ref)
+                else:
+                    # If this block killed any accesses, then the first element
+                    # of the killed writes is the access access that we're
```
arporter, line 1245: s/access access/access/

arporter (changes requested):
Thanks Aidan, that is extremely complex code! I've had a look through most of it except the tests so far. You'll see I'm a bit confused about the correct way of determining the "start point" and "end point". Apart from that, it's just minor things.

src/psyclone/psyir/nodes/assignment.py
LonelyCat124: Done.

src/psyclone/psyir/nodes/assignment.py
LonelyCat124: Done.

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Done

src/psyclone/psyir/nodes/reference.py
LonelyCat124: Fixed.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah, thats better.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done - I think these were originally taken from the book excerpt Sergi provided me so I tried to update them with my current understanding.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah I think so - the elif had come later when I had failing tests so I suspect I just missed the conflict.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: (This was a surprise to me when it broke) I think the control_flow_nodes already contains a "full set" of control flow populated by another function, which includes `None`s for the chains that haven't yet been added, so if I didn't do insert and appended onto both then for some multi-reference cases I got non-matching lists.

I think this is one of those "it would be really nice to rewrite" sections of the DUCs to use a difference data structure (e.g. list of tuples) but I'm not sure how simple it would be - I can have a look if you think its worthwhile.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Fixed.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah I think so.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yes, I felt like for a next_accesses on a statement such as `b = a + a` we wouldn't want to return the second reference to `a`. I guess in theory with how the chains are currently setup for an example like `b = a + a + a` you could provide a list with only `b` and the first and last `a` references in which case we'd miss the middle `a`, but I think that isn't a real case so I'm not worried.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Updated the comment - I think `min` is correct as discussed in the previous comment, but I could be persuaded either way.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Done

LonelyCat124:
@arporter Review comments addressed I think. I've tried to explain my reasoning for using max or min, but happy to change it if you disagree with what the result should be from the DUC.

LonelyCat124:
I guess to also note this only improves the behaviour for Assignments, and the issue says statements, so I'm going to change this to towards 3367 and update the title and fix that later.

src/psyclone/psyir/tools/definition_use_chains.py
arporter: I seem to be generally confused here. You get the length of `chains` and use that as the point at which to insert into `control_flow_nodes` - I don't think I'd fully understood that last time. Please add a comment to explain - this could also go some way to investigating whether (future) refactoring would help.

arporter: I'm not advocating for a re-write of anything at this stage - or at least, not in this PR. You understand the code a lot better than I do so if you think it would simplify things then that would be great, otherwise, don't worry.

src/psyclone/psyir/tools/definition_use_chains.py
arporter: That makes sense, thanks. Worth saying this (the first sentence) in a comment.

src/psyclone/psyir/tools/definition_use_chains.py
arporter: Please add a comment as discussed on the previous comment.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -65,18 +65,19 @@ def test_definition_use_chain_compute_backward_uses(fortran_reader):
     a_3 = psyir.walk(Reference)[3]
     # Check this is the lhs of the assignment
     assert a_3 is psyir.walk(Assignment)[1].rhs
-
+    sig = a_3.get_signature_and_indices()[0]
     duc = DefinitionUseChain(
-        a_3, control_flow_region=[routine]
+        [a_3], control_flow_region=[routine]
```
arporter: You don't need to put `a_3` in a list now.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -91,20 +92,20 @@ def test_definition_use_chain_compute_backward_uses(fortran_reader):
     psyir = fortran_reader.psyir_from_source(code)
     routine = psyir.walk(Routine)[0]
     a_3 = psyir.walk(Reference)[4]
-
+    sig = a_3.get_signature_and_indices()[0]
     duc = DefinitionUseChain(
-        a_3, control_flow_region=[routine]
+        [a_3], control_flow_region=[routine]
```
arporter: No list again.

src/psyclone/tests/psyir/tools/definition_use_chains_backward_dependence_test.py
```diff
@@ -284,10 +295,12 @@ def test_definition_use_chain_find_backward_accesses_loop_example(
     psyir = fortran_reader.psyir_from_source(code)
     routine = psyir.walk(Routine)[0]
     # Start the chain from A = a + i.
+    ref = routine.walk(Assignment)[1].lhs
+    sig = ref.get_signature_and_indices()[0]
     chains = DefinitionUseChain(
```
arporter: Just `(ref)`?

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -79,25 +81,59 @@ class DefinitionUseChain:
...
+                f"found '{type(references).__name__}'."
             )
-        self._reference = reference
-        # Store the absolute position for later.
-        self._reference_abs_pos = reference.abs_position
+        for ref in references:
+            if not isinstance(ref, Reference):
+                raise TypeError(
+                    f"The 'references' argument passed into a "
+                    f"DefinitionUseChain must be a list of References "
```
arporter: "...must be a Reference or a list of References but found ...." and possibly drop "in the list" since the caller might not have provided a list.

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
```diff
@@ -605,10 +647,12 @@ def test_definition_use_chain_find_forward_accesses_loop_example(
     psyir = fortran_reader.psyir_from_source(code)
     routine = psyir.walk(Routine)[0]
     # Start the chain from b = A +2.
+    ref = routine.children[1].loop_body.children[1].rhs.children[0]
+    sig = ref.get_signature_and_indices()[0]
     chains = DefinitionUseChain(
```
arporter: Just "(ref)"?

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
```diff
@@ -1178,6 +1250,83 @@ def test_forward_accesses_nested_loop(fortran_reader):
...
+    k = i
+    end subroutine x"""
+    psyir = fortran_reader.psyir_from_source(code)
+    routine = psyir.walk(Routine)[0]
+    assigns = routine.walk(Assignment)
+    rhs = assigns[0].rhs
+    chains = DefinitionUseChain([rhs.children[0],
+                                 rhs.children[1]])
+    reaches = chains.find_forward_accesses()
+    sig0, _ = rhs.children[0].get_signature_and_indices()
```
arporter: Easier to follow if you name these e.g. `jsig` and `ksig`.

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
```diff
@@ -1178,6 +1250,83 @@ def test_forward_accesses_nested_loop(fortran_reader):
...
+            some_array(i,j,k) = i*k*j
+          end do
+        end do
+      end do
+    end if
+    end subroutine"""
+
+    psyir = fortran_reader.psyir_from_source(code)
+    references = psyir.walk(Reference)
+    next_accesses = references[4].next_accesses()
```
arporter, line 1327: Please add some comments to explain which bits of the code are being examined and found.

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
```diff
@@ -0,0 +1,199 @@
...
+       a = 1
+    end if
+    a = 2 * b
+    end subroutine test""",
+    ])
+def test_duc_forward_equivalence(code, fortran_reader):
+    '''Test the DUCs give the same results for multiple inputs as each of
+    the inputs individually.
+    '''
+
```
arporter: It's a small thing but you could reduce duplication in the fixture by doing:
```python
code = f"subroutine test\n{code}\nend subroutine test"
```

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
```diff
@@ -0,0 +1,199 @@
...
+    end if
+    b = 2
+    a = b + a
+    end subroutine test""",
+    ])
+def test_duc_backward_equivalence(code, fortran_reader):
+    '''Test the DUCs give the same results for multiple inputs as each
+    of the inputs individually.
+    '''
+    psyir = fortran_reader.psyir_from_source(code)
```
arporter, line 160: Same comment here.

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+    assign = psyir.walk(Assignment)[-1]
+
+    all_refs = assign.walk(Reference)
+    lhs_ref = all_refs[0]  # First Reference is lhs a
```
arporter, line 165: Probably worth sanity-checking this in case of future changes, e.g. `assert lhs_ref.symbol.name == "a"` and same for `rhs_ref`.

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
```diff
@@ -0,0 +1,199 @@
...
+    '''Test the DUCs give the same results for multiple inputs as each of
+    the inputs individually.
+    '''
+
+    psyir = fortran_reader.psyir_from_source(code)
+
+    assign = psyir.walk(Assignment)[0]
+
+    all_refs = assign.walk(Reference)
+    lhs_ref = all_refs[0]  # First Reference is lhs a
```
arporter, line 95: Pls add sanity checks that the refs are the ones you expect, e.g. `assert lhs_ref.symbol.name == "a"`

arporter (changes requested):
Thanks very much Aidan. I've made it all the way through now and it just comments and tidying left to do :-)
(ITs were all green bar a signal 11 on the LFRic tests which I've just restarted.)

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Added a comment here.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Added.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Added

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah, fixed.

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
LonelyCat124: Renamed

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/tools/definition_use_chains_multiref_test.py
LonelyCat124: Added

LonelyCat124:
@arporter Ready for another look.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -1016,6 +1025,10 @@ def find_backward_accesses(self) -> dict[Signature, list[Node]]:
         if self._stop_point is None:
             # Find the min abs position, as all of these are
             # contained in the same parent.
+            # We start before any of the provided references, as
```
arporter: Do you mean "stop" rather than "start"?

src/psyclone/tests/psyir/tools/definition_use_chains_forward_dependence_test.py
arporter: Great, thanks.

arporter (changes requested):
Thanks very much Aidan. I just need clarification on one comment and then this is good to go. I'll re-check the ITs as there's been quite a lot of code under the bridge since I last did it.

arporter:
Unfortunately the ITs were not happy - please could you take a look?

LonelyCat124:
I had a quick look and I think the fixes on your branch will resolve most of them. The LFRic extraction ones I'm hoping will be fixed but there were a lot of build system errors there.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yes, I've rewritten the comments describing what this start/stop point setters mean.

LonelyCat124:
@arporter Fixed up the issues, ready for another look.

arporter (approved):
All looks good now, thanks Aidan. I also see the ITs are all green so this is ready to go.

PR #3404, opened 2026-04-15
(closes #3403) Modernise runme examples

### LINKED ISSUE #3403: Modernise examples that use the runme script
Some examples still have a runme python script instead of calling psyclone. This is problematic because they still use the `from psyclone.parse.algorithm import parse` that now could be deleted. We need to mordernise this examples.

sergisiso:


sergisiso:
@LonelyCat124 @arporter This is ready for review, it brings the last example that was using the old runme.py style scripts to modern psyclone scripts. Github lost track of the renaming/changes, so it is best to start from the Makefile modification to understand the changes.

LonelyCat124 (approved):
Looks good, will proceed to merge.

PR #3407, opened 2026-04-22
(Closes #3406) Fix subroutine handler to find the subroutine stmt if it exists instead of assuming its the first child.

### LINKED ISSUE #3406: Comment creates problem for fparser2
NEMO dev main branch (version from 2026-04-21) has some code that psyclone can't parse with `--keep-comments` enabled.

```
MODULE tsltde
CONTAINS
! Remove this comment and this empty line to make it compile
   SUBROUTINE tsl_tde_osc()
   CONTAINS
      RECURSIVE SUBROUTINE nodal_factort()
            CALL nodal_factort()
      END SUBROUTINE nodal_factort
   END SUBROUTINE tsl_tde_osc
END MODULE tsltde
```
Command to reproduce: `psyclone --keep-comments  ~/tmp/tsltde.f90`
`PSyclone version: 3.3.1-dev`
`FParser Version: 0.2.2` also reproducible with version `0.2.3.dev1+ga197a857a`

Error message:
```
  File "/home/martin/workspace/psyclone_REPOS/.venv/lib/python3.14/site-packages/psyclone/psyir/frontend/fparser2.py", line 5720, in _subroutine_handler
    name = str(x.parent.children[0].children[1])
```

If the comment is removed, it works.

Bringing @LonelyCat124 and @arporter  into the loop.

LonelyCat124:
@schreiberx This worked for the reproducer you gave us - if you have time to try it on NEMO itself it'd be appreciated.

Otherwise ready for a look from @arporter or @sergisiso. I will set ITs running in the mean time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -171,7 +171,7 @@ if __name__ == '__main__':
         classifiers=CLASSIFIERS,
         packages=PACKAGES,
         package_dir={"": "src"},
-        install_requires=['pyparsing', 'fparser==0.2.2', 'configparser',
+        install_requires=['pyparsing', 'fparser>=0.2.2', 'configparser',
                           'sympy', "Jinja2", 'termcolor', 'graphviz'],
         # Have to pin Sphinx to a pre-9.0 version because of
         # https://github.com/sphinx-doc/sphinx/issues/14223
--- a/setup.py
+++ b/setup.py
@@ -171,7 +171,7 @@ if __name__ == '__main__':
         classifiers=CLASSIFIERS,
         packages=PACKAGES,
         package_dir={"": "src"},
-        install_requires=['pyparsing', 'fparser>=0.2.2', 'configparser',
+        install_requires=['pyparsing', 'fparser==0.2.2', 'configparser',
                           'sympy', "Jinja2", 'termcolor', 'graphviz'],
         # Have to pin Sphinx to a pre-9.0 version because of
         # https://github.com/sphinx-doc/sphinx/issues/14223
--- a/src/psyclone/psyir/frontend/fparser2.py
+++ b/src/psyclone/psyir/frontend/fparser2.py
@@ -5717,7 +5717,16 @@ class Fparser2Reader():
         try:
             x = _first_type_match(node.children,
                                   Fortran2003.Internal_Subprogram_Part)
-            name = str(x.parent.children[0].children[1])
+            subroutine_stmt = None
+            for child in x.parent.children:
+                if isinstance(child, Fortran2003.Subroutine_Stmt):
+                    subroutine_stmt = child
+                    break
+            # Backup branch to maintain old functionality for edge cases
+            # and test suite.
+            if not subroutine_stmt:
+                subroutine_stmt = x.parent.children[0]
+            name = str(subroutine_stmt.children[1])
             # If we will make a CodeBlock to represent this subroutine then
             # we still need to ensure the symbol is in the parent's symbol
             # table. For this case the best we can do is place the symbol
--- a/src/psyclone/psyir/frontend/fparser2.py
+++ b/src/psyclone/psyir/frontend/fparser2.py
@@ -5717,15 +5717,12 @@ class Fparser2Reader():
         try:
             x = _first_type_match(node.children,
                                   Fortran2003.Internal_Subprogram_Part)
-            subroutine_stmt = None
-            for child in x.parent.children:
-                if isinstance(child, Fortran2003.Subroutine_Stmt):
-                    subroutine_stmt = child
-                    break
-            # Backup branch to maintain old functionality for edge cases
-            # and test suite.
-            if not subroutine_stmt:
-                subroutine_stmt = x.parent.children[0]
+            # Find the Subroutine_Stmt or Function_Stmt to find the Name
+            # from.
+            subroutine_stmt = _first_type_match(
+                x.parent.children,
+                (Fortran2003.Subroutine_Stmt, Fortran2003.Function_Stmt)
+            )
             name = str(subroutine_stmt.children[1])
             # If we will make a CodeBlock to represent this subroutine then
             # we still need to ensure the symbol is in the parent's symbol
--- a/src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
+++ b/src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
@@ -631,24 +631,54 @@ def test_module_contains_comment_before_subroutine(
     '''Test to check that subroutines contained in a module are
     correctly handled when there are comments before the subroutine
     statement after the contains.'''
-    code = """MODULE tsltde
+    code = """MODULE test_mod
 CONTAINS
 ! Here is a comment
-   SUBROUTINE tsl_tde_osc()
-   END SUBROUTINE tsl_tde_osc
-END MODULE tsltde"""
+   SUBROUTINE test()
+     contains
+     Subroutine subsub()
+     end subroutine
+   END SUBROUTINE test
+END MODULE test_mod"""
     fortran_reader = FortranReader(ignore_comments=False)
     psyir = fortran_reader.psyir_from_source(code)
     out = fortran_writer(psyir)
-    assert """module tsltde
-  implicit none
-  public
-
-  contains
+    assert """  contains
+  ! PSyclone CodeBlock (unsupported code) reason:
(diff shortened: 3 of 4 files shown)


LonelyCat124:
This PR also ticks fparser from 0.2.2 to >= 0.2.2 (since fparser has a 0.2.3 version now at least it helps us to use newer versions during development).

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5717,7 +5717,16 @@ def _subroutine_handler(self, node, parent):
...
+            subroutine_stmt = None
+            for child in x.parent.children:
+                if isinstance(child, Fortran2003.Subroutine_Stmt):
+                    subroutine_stmt = child
+                    break
+            # Backup branch to maintain old functionality for edge cases
+            # and test suite.
+            if not subroutine_stmt:
+                subroutine_stmt = x.parent.children[0]
+            name = str(subroutine_stmt.children[1])
```
sergisiso: I am confused about this (and the previous code was also lacking), but my understanding is that this whole block is to find the CONTAINS section inside subroutines and functions (first_type_match).
If we find it we create a codeblock but still put its name in the parent symbol table (but not the ones from the CONTAINS subroutines, hence only taking the first one).

Now we look at isinstance(child, Fortran2003.Subroutine_Stmt) instead of assuming that it is children[1], that is good, but wouldn't that miss the case when this is a Function. Even worse, a function with a contains with a subroutine may get the name of the internal subroutine in the parent symbol table?

Is the backup branch related to missing function names?

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
```diff
@@ -46,6 +46,7 @@
 from psyclone.errors import InternalError
 from psyclone.psyir.frontend.fparser2 import (
     Fparser2Reader, TYPE_MAP_FROM_FORTRAN)
+from psyclone.psyir.frontend.fortran import FortranReader
```
sergisiso, line 49: Use the fortran_reader fixture instead

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
```diff
@@ -622,3 +623,32 @@ def test_module_contains_subroutine_contains_subroutine(
...
+def test_module_contains_comment_before_subroutine(
+    fortran_writer
+):
+    '''Test to check that subroutines contained in a module are
+    correctly handled when there are comments before the subroutine
+    statement after the contains.'''
+    code = """MODULE tsltde
+CONTAINS
+! Here is a comment
+   SUBROUTINE tsl_tde_osc()
```
sergisiso: I am not sure this is the problem in the issue, was this also not working before?

The problem is the issue in then the subroutine has a contains and it ends up being a Codeblock.

Could you also test a function that contains a subroutine and in both cases check that the resulting module symbol tables have the correct symbol in it?

setup.py
```diff
@@ -171,7 +171,7 @@ def get_files(directory, install_path, valid_suffixes):
         classifiers=CLASSIFIERS,
         packages=PACKAGES,
         package_dir={"": "src"},
-        install_requires=['pyparsing', 'fparser==0.2.2', 'configparser',
+        install_requires=['pyparsing', 'fparser>=0.2.2', 'configparser',
```
sergisiso: I would reverse this to `==` as this is the version that we created for the release and is the version still pointed by the submodule

sergisiso (changes requested):
@LonelyCat124 I have some doubts about the implementation and its test, see inline.

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
LonelyCat124: I can't i need `ignore_comments=True`

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: The failing case was that it would look for the name of the function at ` x.parent.children[0].children[1]`, which in this case `x.parent.children[0]` was a Comment and not the Subroutine statement.

the printing of `x.parent.children[0]` in the failing case was:
```
[Comment('! Remove this comment and this empty line to make it compile'), Subroutine_Stmt(None, Name('tsl_tde_osc'), None, None), Internal_Subprogram_Part(Contains_Stmt('CONTAINS'), Subroutine_Subprogram(Subroutine_Stmt(Prefix(' ', (Prefix_Spec('RECURSIVE'),)), Name('nodal_factort'), None, None), Execution_Part(Call_Stmt(Name('nodal_factort'), None)), End_Subroutine_Stmt('SUBROUTINE', Name('nodal_factort')))), End_Subroutine_Stmt('SUBROUTINE', Name('tsl_tde_osc'))]
```

I could just walk to the first Name but walking in fparser trees is a bit gross so I just did this. I can see the issue if we have a Function so I can see if I need to change that, otherwise all I changed here was the discovery of the subroutine name I think?

The not found statement was to avoid ever passing this with None, which happened for some of the tests but I can investigate that more detailed as maybe they are Functions and I'm just missing those.

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
LonelyCat124: No this wasn't working before, the comment would fail because fparser2.py would ask the comment for `children[1]` to find the `Name` which didn't exist.

setup.py
LonelyCat124: Ok, I think there is something weird going on then. My fparser2 _version.py in external (which is up to date with fparser master) shows:
```
__version__ = version = '0.2.3.dev1+ga197a857a'
__version_tuple__ = version_tuple = (0, 2, 3, 'dev1', 'ga197a857a')
```
and the only change after the tagged release is `Update README.md`, but with == I couldn't do `pip install -e .` on the latest master of fparser, which is why I did >= - it only changes development for now unless we do more changes to fparser, and is probably necessary for ongoing development branches I think? That or `_version.py` in fparser has done something weird.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: The test does require a contains within the subroutine to test this, so I did have that wrong, but I think the name we're aiming to find is the original subroutine or function. I've addressed the not finding functions and removed the fail to find subroutine_stmt, and tidied up the search for it using the inbuilg utilties.

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
LonelyCat124: It did need the contains and now it tests it properly.

LonelyCat124:
@sergisiso Ready for another look, I fixed the main changes not being tested properly but the fparser chjange I didn't revert yet as I'm confused as to why my local fparser reports a later version.

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Ah yeah, this new implementation is much clearer. Thanks

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
sergisiso: Can you add the symbol_table checks in both tests please:
```
assert "test" in psyir.children[0].symbol_table
assert "subsub" not in psyir.children[0].symbol_table
```

setup.py
sergisiso: Can you move this to a separate as I think we need a better solution here but the rest of the PR is ready.

sergisiso (changes requested):
@LonelyCat124 See replies above, I like the new solution, there are just some minor comments now.

setup.py
LonelyCat124: Done

src/psyclone/tests/psyir/frontend/fparser2_subroutine_handler_test.py
LonelyCat124: Added

LonelyCat124:
@sergisiso Reverted the setup.py change and added the additional checks into the tests.

sergisiso (approved):
@LonelyCat124 Everything looks good now and the integration test succeeded, this is approved for merging

PR #3408, opened 2026-04-22
(Closes #2812) max and min code update

### LINKED ISSUE #2812: max2code and min2code assume args are real scalars.
Embarrassingly, I fixed this problem for Abs2Code and Sign2Code but somehow failed to spot that Max2Code and Min2Code do exactly the same thing (with a TODO for #658). We need to take the fix used for the former two transformations and add it to MinOrMax2CodeTrans.

LonelyCat124:
Small PR to ensure the max2code and min2code transformations use the datatype of their arguments correctly, and don't assume scalar real.

Ready for review from @arporter when he's back - its possible I missed something I should have added but I'm not sure.

### DIFF
--- a/src/psyclone/psyir/transformations/intrinsics/minormax2code_trans.py
+++ b/src/psyclone/psyir/transformations/intrinsics/minormax2code_trans.py
@@ -47,11 +47,13 @@ transformations.
 from abc import ABC
 import warnings
 
-from psyclone.psyir.nodes import BinaryOperation, Assignment, \
-        Reference, IfBlock
-from psyclone.psyir.symbols import DataSymbol, REAL_TYPE
-from psyclone.psyir.transformations.intrinsics.intrinsic2code_trans import \
-        Intrinsic2CodeTrans
+from psyclone.psyir.nodes import (
+    BinaryOperation, Assignment, Reference, IfBlock
+)
+from psyclone.psyir.symbols import DataSymbol
+from psyclone.psyir.transformations.intrinsics.intrinsic2code_trans import (
+    Intrinsic2CodeTrans
+)
 from psyclone.utils import transformation_documentation_wrapper
 
 
--- a/src/psyclone/psyir/transformations/intrinsics/minormax2code_trans.py
+++ b/src/psyclone/psyir/transformations/intrinsics/minormax2code_trans.py
@@ -48,7 +48,7 @@ from abc import ABC
 import warnings
 
 from psyclone.psyir.nodes import (
-    BinaryOperation, Assignment, Reference, IfBlock
+    BinaryOperation, Assignment, Reference, IfBlock, IntrinsicCall
 )
 from psyclone.psyir.symbols import DataSymbol
 from psyclone.psyir.transformations.intrinsics.intrinsic2code_trans import (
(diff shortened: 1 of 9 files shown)


src/psyclone/psyir/transformations/intrinsics/minormax2code_trans.py
```diff
@@ -85,6 +87,19 @@ def __init__(self):
         super().__init__()
         self._compare_operator = None
 
+    def validate(self, node, options=None):
+        '''
+        Check that it is safe to apply the transformation to the supplied node.
+
+        :param node: the SIGN call to transform.
+        :type node: :py:class:`psyclone.psyir.nodes.IntrinsicCall`
```
arporter: typehints please.

arporter (changes requested):
Nice, thanks Aidan. Just a bit of docstring tidying to do and there's a (simple) test failure that needs fixing. Once you've done that, please could you launch the ITs too...

src/psyclone/psyir/transformations/intrinsics/minormax2code_trans.py
LonelyCat124: Added, also added the kwargs that I'd missed and fixed the typehints. I removed the options typehints since they're deprecated now.

LonelyCat124:
Am hoping the test will just work this time as I'm a bit confused as I can't replicate it locally (any more). If not I'll dive deeper.

LonelyCat124:
@arporter I have "fixed" the tree sitter issues now I think. I would ask if you could test locally as well and check you have no issues with your local version. I'm not sure why these tests have suddenly started failing, I tried running on another branch as well and they also have failures that weren't present last week. As far as I can find there has been no treesitter or treesitter-fortarn new release so its very strange.

arporter:
Hi @LonelyCat124, I've just tried locally (with 3.13) and I get one treesitter-related failure with this branch. If I switch to latest master it goes away. It could be something to do with tests leaking state?

LonelyCat124:
Set the ITs going - assume they are successful this is ready for another look.

src/psyclone/tests/psyir/frontend/fortran_treesitter_reader/ftr_test.py
```diff
@@ -74,6 +75,10 @@ def test_generate_parse_tree(tmpdir_factory, caplog):
     Test that generate_parse_tree returns treesitter trees or appropriate
     error messages.
     '''
+    # TODO #3416: Skip treesitter tests below 3.10 as they're unsupported by
```
arporter: Since 3.9 is going to be with us a while and (hopefully) the number of treesitter tests is only going to increase, I thought it was worth looking to see whether there's a better way of doing this and there is:

Image

arporter: Please could you do the equivalent of the `@minversion` described above?

arporter (changes requested):
ITs were all green and tests were fine (apart from a link-check on Tau which has an out-of-date certificate).
I was going to green-light it this time but then thought we should perhaps do the test skipping for tree sitter in a nicer fashion - see inline.

src/psyclone/tests/psyir/frontend/fortran_treesitter_reader/ftr_test.py
LonelyCat124: Added

LonelyCat124:
@arporter Created something in tests.utilities to do this and use that now.

arporter (approved):
Nice, thanks for that Aidan. All good now.

PR #3413, opened 2026-04-23
(closes #3411) Remove OpenMP SUB reduction operator (use ADD instead)

### LINKED ISSUE #3411: The minus reduction operator is deprecated since OpenMP 5.2 and is not supported
This issue was observed when building the full NEMO (v5) application using amdflang (22.2 and above). The default transformation script for GPU offloading using OpenMP (`omp_gpu_trans`) generates `reduction(-: var)` for loops, which fails for an  OpenMP 5.2 standard-compliant compiler

This manifests at build time like:
```text
error: Semantic errors in /home/thgibson/nemo_workspace/nemo/cfgs/BENCH_AMD_GPU_AFAR23_OMP_UPSTREAM/BLD_SCT_PSYCLONE/obj/obs_inter_h2d.f90
/home/thgibson/nemo_workspace/nemo/cfgs/BENCH_AMD_GPU_AFAR23_OMP_UPSTREAM/BLD_SCT_PSYCLONE/obj/obs_inter_h2d.f90:840:66: error: The minus reduction operator is deprecated since OpenMP 5.2 and is not supported in the REDUCTION clause.
          !$omp teams loop collapse(1) default(shared) private(jk) reduction(-: zsum)
                                                                   ^^^^^^^^^^^^^^^^^^
```

The workaround for this is to apply a post-pass to rewrite `reduction(-: var)` to `reduction(+: var)`; a recurrence of the form `acc = acc - val` is equivalent to a `+` reduction on `-val`. I think this is worth addressing in Psyclone so this gets addressed automatically, but some care should be taken I imagine.

sergisiso:


sergisiso:
@LonelyCat124 @arporter This should be ready for review. I launched the integration test to see if the other compilers also agree.

LonelyCat124 (approved):
Looks good to me, simple fix, assuming ITs are ok will merge.

Issue #3279, opened 2026-01-13
NEMOv5 GPU ready transformations


addy419:
Hi @arporter @sergisiso,

I am back working on the GPU benchmarking for NEMO and have encountered issues with the `master` branch of PSyclone.
The ORCA2_ICE_PISCES config produces a signal 11 with nvhpc@24.11 on `bdylib.f90`
BENCH config produces a signal 11 on `trabbc.f90`. Both on the `v5.0-RC` branch of NEMO.

Can you please check if the integration tests are successful as expected? Are there any changes to the branch or the compiler in the integration tests?

sergisiso:
Oh and `examples/nemo/scripts/KGOs/namelist_cfg_bench_small` as `namelist_cfg`, I don't know if there is any difference to the one in the repo

addy419:
@sergisiso I tried nvhpc 25.1 and 24.11 with both managed and unified memory (and same flags as the CI), but BENCH is unable to compile `trabbc.f90` without encountering `fort2: terminated by signal 11`

sergisiso:
Isn't this the same solfrac.F90 issue that we had with nvfortran and had to patch that file?

I recently added:

https://github.com/stfc/PSyclone/blob/c9a794cfff17020f8d27742af045e1f4a2f18ca5/examples/nemo/scripts/omp_gpu_trans.py#L194C1-L199C48

into master's omp_gpu_trans.py that copies the global parameters inside the subroutine's body. Are you using the 'single_nemo_script' branch? I can bring the changes there if helpful or just copy the lines above.

addy419:
@sergisiso Just moved to the latest master, but it doesn't seem the code block is executed.

```
        if subroutine.name == "solfrac_mod.f90":
            # Bring these solfrac parameters to the subroutine as nvidia
            # does not permit offloaded kernels to access module parameters
            print("#######################################################INSOLFRAC######################################")
            symtab = subroutine.symbol_table
            symtab.add(symtab.lookup("pp_wgt"))
            symtab.add(symtab.lookup("pp_len"))
```

PR #3414, opened 2026-04-23
(towards #3279) Improves NEMOv5 performance

sergisiso:
... by removing exclusions that don't need to be there anymore (no signal 11 observed), and avoid doing the lbclnk which has lots of small control loops on the GPU.


sergisiso:
@LonelyCat124 @arporter This is ready for review. It provides a small performance improvement over all NEMOv5 GPU targets and increases the script coverage, which is useful for testing purposes.

LonelyCat124 (approved):
Yeah if they work for the integration tests without performance issues then this all looks fine, will proceed to merge.

Issue #3398, opened 2026-04-09
LFRic: perform field-copy to full depth of clean halo


arporter:
Currently any builtin will, by default, iterate to the last owned (or annexed) dof. Although this can be extended into the halo using `LFRicRedundantComputationTrans`, the depth of the redundant computation is specified at compile time. This does not support the use case of an iterative loop where, to amalgamate halo exchanges, a single, deep halo exchange is performed prior to the loop. At each trip round the loop, the depth of the clean halo is reduced by one. Thus, the depth to which to do the redundant computation is determined at run-time.

Currently, `LFRicRedundantComputationTrans` accepts a `depth` option that specifies the literal depth to which to perform redundant computation. If `depth` is not set then redundant computation is performed to the full halo depth. What we could do with is a third option that says "do redundant computation but only out to the depth of the currently-clean halo."

arporter:
An alternative approach is to have all built-ins default to iterating out to the currently-clean halo.

arporter:
Either of these approaches require that we be able to query the field proxy for the depth of the clean halo. This information can be computed but isn't readily available at the moment:
https://github.com/MetOffice/lfric_core/blob/3ef2ace5104efe01580c672e1837a584d428f5af/infrastructure/source/field/field_parent_mod.f90#L123-L131
i.e. we can ask whether a certain depth of halo is clean but not what the maximum clean depth is.

arporter:
Met with Mike and Chris and agreed this as the long(er) term plan. Mike will add the necessary functionality to the LFRic infrastructure. It will also be necessary to check that the clean halo depth is indeed decreasing for every trip around the loop (it *should* be but it pays to be sure).

arporter:
Thinking some more about this, there are other complexities. Currently we have:

<img width="692" height="312" alt="Image" src="https://github.com/user-attachments/assets/f282e5c7-46d9-4d5b-b27d-b68aefaa1695" />

so if depth is supplied, it must currently be an integer literal. This information is also used to determine what depth to do a halo exchange, e.g.:
```fortran
    if (mt_lumped_inv_proxy%is_dirty(depth=1)) then
      call mt_lumped_inv_proxy%halo_exchange(depth=1)
    end if
```
I don't yet know whether the halo-exchange logic can currently cope with a variable for the depth.

arporter:
`LFRicRedundationComputationTrans` has been moved and modernised so we're now ready to add support for this new functionality (of querying and then using the depth of the clean halo).

[...]

PR #3417, opened 2026-04-27
(Towards #3398) move LFRicRedundantComputationTrans and update to use kwargs.

arporter:
An initial step before updating the options accepted by this transformation.


arporter:
Test failure is the treesitter one that #3408 fixes.

arporter:
A fairly straightforward PR that just moves LFRicRedundantComputationTrans to the correct location and updates it to use kwargs. One for @LonelyCat124 @sergisiso or @hiker.

src/psyclone/domain/lfric/transformations/__init__.py
```diff
@@ -48,5 +48,7 @@
     import RaisePSyIR2LFRicAlgTrans
 from psyclone.domain.lfric.transformations.lfric_loop_fuse_trans \
     import LFRicLoopFuseTrans
+from psyclone.domain.lfric.transformations.lfric_redundant_computation_trans \
```
LonelyCat124: Use parentheses here instead of `\`. You don't need to change the whole file unless you want.

src/psyclone/domain/lfric/transformations/lfric_redundant_computation_trans.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def validate(self, node: Loop, options=None, **kwargs):
+        '''Perform various checks to ensure that it is valid to apply the
+        RedundantComputation transformation to the supplied node
+
+        :param node: the supplied node on which we are performing
+                     validity checks.
+        :param options: a dictionary with options for transformations.
+        :type options: Optional[Dict[str, Any]]
+        :param int options["depth"]: the depth of the stencil if the value
```
LonelyCat124: You don't need to keep the options declarations in the docstring now.

LonelyCat124: Once thats done I'll check the generated docs are correct.

src/psyclone/domain/lfric/transformations/lfric_redundant_computation_trans.py
```diff
@@ -0,0 +1,312 @@
...
+
+        '''
+        # TODO #2668: remove options dictionary.
+        self.validate(node, options=options, **kwargs)
+
+        if options:
+            # TODO #2668: Deprecate options dictionary.
+            depth = options.get("depth")
+        else:
+            depth = self.get_option("depth", **kwargs)
```
LonelyCat124: You don't need the else block for the `apply` method - we just have the depth kwarg value. Do we have a test for this behaviour as I think the current implementation should fail that test as `self.get_option("depth", **kwargs)` should always return `None` here I think.

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
```diff
@@ -0,0 +1,1231 @@
...
+def test_rc_invalid_depth():
+    ''' Test that LFRicRedundantComputationTrans raises an exception if the
+    supplied depth is less than 1. '''
+    _, invoke = get_invoke("1_single_invoke.f90", TEST_API,
+                           idx=0, dist_mem=True)
+    schedule = invoke.schedule
+    rc_trans = LFRicRedundantComputationTrans()
+    loop = schedule.children[4]
+    with pytest.raises(TransformationError) as excinfo:
+        rc_trans.apply(loop, {"depth": 0})
```
LonelyCat124, line 173: Update the test to use keywords.

LonelyCat124: Keep one test (or duplicate one test) with kwargs and add a TODO to remove it.

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
```diff
@@ -0,0 +1,1231 @@
...
+def test_rc_invalid_depth_continuous():
+    ''' Test that LFRicRedundantComputationTrans raises an exception if the
+    supplied depth equals 1 when modifying a continuous field. '''
+    _, invoke = get_invoke("1_single_invoke.f90", TEST_API,
+                           idx=0, dist_mem=True)
+    schedule = invoke.schedule
+    rc_trans = LFRicRedundantComputationTrans()
+    loop = schedule.children[4]
+    with pytest.raises(TransformationError) as excinfo:
+        rc_trans.apply(loop, {"depth": 1})
```
LonelyCat124, line 223: Here also

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
```diff
@@ -0,0 +1,1231 @@
...
+    the redundant computation transformation with a fixed value for
+    halo depth.
+
+    '''
+    psy, invoke = get_invoke("1_single_invoke.f90", TEST_API,
+                             idx=0, dist_mem=True)
+    schedule = invoke.schedule
+    rc_trans = LFRicRedundantComputationTrans()
+    loop = schedule.children[4]
+    rc_trans.apply(loop, {"depth": 3})
```
LonelyCat124: And again

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
```diff
@@ -0,0 +1,1231 @@
...
+    schedule = invoke.schedule
+    rc_trans = LFRicRedundantComputationTrans()
+    if annexed:
+        # there are no halo exchange calls
+        index = 0
+    else:
+        # there are 3 halo exchange calls
+        index = 3
+    loop = schedule.children[index]
+    rc_trans.apply(loop, {"depth": 3})
```
LonelyCat124: also

LonelyCat124: Also lines 324, 370, 424, 473, 561, 614, 728, 783, 792, 809, 830, 853, 859, 866, 902, 910, 1014, 1045-1046, 1052, 1062, 1087, 1094, 1118, 1126, 1153, 1160, 1168, 1229

LonelyCat124 (changes requested):
Hi @arporter - few things to sort out, mostly in tests/formatting but I think there are a couple of errors in the kwargs update also.

Edit: Also can you double check if this is used in lfric_apps/core scripts and make sure they're updated with the new import location (conditionally with importError probably).

LonelyCat124:
Looking at the ITs failures it lokos like this was handled previously by the Dynamo compatability layer in transformations.py - so either we can update that or we can update lfric_apps/core scripts (the latter is probably preferable at this point).

sergisiso:
@arporter @LonelyCat124 I have the new branches that force push to our lfric_apps almost ready, let me do the change and give you a hash to try, otherwise it will conflict with my changes. I will look at this tomorrow.

arporter:
OK, thanks Sergi.

sergisiso:
~~@arporter Updating the LFRic hash to `300225d72ffca5c843cdeceeedb353f6eb01dfa9` should work, but you need to bring the branch to master first.~~

PS: Another branch is already bringing this update, so not hash update should be necessary once this is brought to master

src/psyclone/domain/lfric/transformations/__init__.py
arporter: This took some digging. The only way I could find to solve it was to use relative imports (https://peps.python.org/pep-0328/#guido-s-decision) as that reduces the length of the import path. I think in this case it is also probably more readable too.

src/psyclone/domain/lfric/transformations/lfric_redundant_computation_trans.py
arporter: Done.

src/psyclone/domain/lfric/transformations/lfric_redundant_computation_trans.py
arporter: I'm confused here - I thought we were keeping 'options' as an argument to `apply`? Why will `get_option("depth", **kwargs)` always return None?

arporter: Ah, having updated one of the tests i now see what you mean. Fixed.

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
arporter: Done.

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
arporter: Done.

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
arporter: Done.

src/psyclone/tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py
arporter: Done all.

arporter:
The NEMO integration test fails. This requires further investigation.

arporter:
The problem was a circular dep. on Transformation in psyclone.utils which was only triggered if Sphinx wasn't installed (which meant we used our own implementation of the Sphinx-supplied method in utils.py).

arporter:
Assuming the ITs are all fine, this is ready for another look now. (I cancelled the compilation IT because it would have failed due to a test failure that I've subsequently fixed.)

src/psyclone/domain/lfric/transformations/__init__.py
LonelyCat124: Ah yeah, I'd missed the formatting here, thanks for resolving that.

src/psyclone/psyGen.py
```diff
@@ -2539,7 +2539,7 @@ def validate_options(self, **kwargs):
                 except TypeError:
                     # Type checking for Generics, e.g. Union[...], doesn't
                     # work so we skip this check - it is done in the
-                    # relevant function instead.
+                    # validate method of the transformation instead.
```
LonelyCat124, line 2542: Thanks for updating this text.

LonelyCat124 (approved):
All looks good now, will being merging process.

PR #3419, opened 2026-04-28
(closes #3418) Fortran backend elseif

### LINKED ISSUE #3418: Add backend support for "elseif"
Currently Fortran constructs from else if or select case or select type end up as nested if. This is done to avoid an additional construct for an elseif and simplify the IR, which is good thing, but when generating back the fortran it can lead to heavily indented code.

The nvfortran has problems with this heavily indented code and we need to use the "--backend disable-indentation" flag. This makes the output code difficult to read (but the heavily indented code is even harder to read), and looking at the output and understanding what it does is difficult.

Without adding a elseif construct, the backend could try to recreate the elseif in order to produce nicer looking outputs.

sergisiso:


sergisiso:
@arporter @LonelyCat124 This is ready for review. As discussed in the issue, the main purpose of this is to prevent the deeply nested if that psyclone sometimes generate from 'select cases' that break the nvfortran compiler. With this we can compile lfric_atm without using --disable-backend-indentation and the output files are much nicer to read.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -354,16 +354,14 @@ def test_comments_and_codeblocks(last_comments_as_codeblocks):
       ! Comment on assignment 'a = 2'
       a = 2
       ! Comment on elseif block 'elseif (a == 2) then' => CodeBlock
```
LonelyCat124, line 356: Indentation of the comment on the else if seems wrong to me - it should be outdented to the same point as the else if right?

LonelyCat124: I guess this is a question of where this was originally placed in the frontend - it looks like its added at the end of the previous if body, but if we had it on the next if statement we'd probably get it in the right indentation level - can you see if this is possible for elseif?

LonelyCat124 (changes requested):
@sergisiso It mostly looks good - I had one small issue I see to do with comment indentation where I think we can probably do better for elseif now.

If you get it to work, can you also put a test with the word `if` in the preceding comment as I suspect this will break your implementation as well

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
sergisiso: These are Codeblocks placed in the if_body (because they cannot be associated to a preceding comment since they are at the end of a schedule?) so I would leave them at if_body indentation (as they were before this PR).

Trying to change the indentation is tricky because we can not distinguish between:

```
   ! End of if block comment
else
```
and
```
! Else comment
else
```

sergisiso:
@LonelyCat124 Ready for another review, see if you are happy with the response, I don't think we can do better as (specially in the backend) we don't know what the comment referred to.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Ok thats no problem - there's still an issue with your replace code though. I threw together a quick example taht fails:
```
code = """subroutine test
integer :: i, j
logical :: a

if(a) then
  i = 1
else
  !Here is an if comment
  if(i > 2) then
    j = 1
  end if
end if
end subroutine
"""

freader = FortranReader(ignore_comments=False)
fwriter = FortranWriter()
psyir = freader.psyir_from_source(code)
print(fwriter(psyir))
```

This outputs:
```
  if (a) then
    i = 1
  ! Here is an elseif comment
  if (i > 2) then
    j = 1
  end if
```

LonelyCat124 (changes requested):
@sergisiso I'm happy to not worry about those comments, but there's still a bug in the implemmentation at the moment.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
sergisiso: Oh that's a good catch, I did not consider this

sergisiso: Ok, attempted new implementation and added your test

sergisiso:
@LonelyCat124 Ready for another review (CI-permitting)

LonelyCat124:
All looks ok, am going to rerun integration as we've changed code output again just to be safe, if thats all good I'll accept and merge.

LonelyCat124:
ITs passed, will proceed to merge.

LonelyCat124 (approved):

PR #3422, opened 2026-04-30
(closes #440) Clean remaining todos for this issue

### LINKED ISSUE #440: Extend Loops to be language independent

sergisiso:


sergisiso:
@LonelyCat124 This is ready for review, it is a small PR to clean up the remaining TODOs to close this issue. I removed some of the C checks because there is no support for the Call node, and C and Fortran reuse the same common directive backend, so the code is goes through is the same (no coverage lost).

I also fixed some issues with links as I wanted to see green CI checks again.

LonelyCat124 (dismissed):
@sergisiso Looks good to me, removing the C checks makes sense. I' don't think I need to run CI for this so will merge after group meeting.

sergisiso:
@LonelyCat124 Unfortunately Codecov did report a couple of missing lines, let me have a quick look before merging

sergisiso:
@LonelyCat124 Ok, all green now. One of the tests I removed was the only example of standalonedirective in C. I added this back but in the c_test.py as an omp barrier now.

LonelyCat124 (approved):
Yeah, after jumping the gun last time this looks fine now. I'll merge this shortly after andy's merged the other and i fix the changelog conflict that will arise.

sergisiso:
Aah I can't belive its red again! But now is due to the nemo link, we were warned this morning that the nemo webpage certificate has expired and they will solve this shortly, so I would still go ahead without fixing the nemo link.

PR #3428, opened 2026-05-06
Update compilation tests to gfortran 16

nmnobre:


sergisiso:
https://github.com/stfc/PSyclone-mirror/actions/runs/25484600410/job/74776858025

nmnobre:
```
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7fa286068fff in ???
#1  0x4a527b in __local_mesh_mod_MOD_initialise_full
	at /redacted/PSyclone-mirror/PSyclone-mirror/external/lfric_infrastructure//src/mesh/local_mesh_mod.f90:440
#2  0x406d11 in MAIN__
	at /redacted/PSyclone-mirror/PSyclone-mirror/examples/lfric/eg14/main_alg.f90:50
#3  0x409d0a in main
	at /redacted/PSyclone-mirror/PSyclone-mirror/examples/lfric/eg14/main_alg.f90:2
make[2]: *** [Makefile:85: run] Segmentation fault (core dumped)
```
This seems to be a gfortran bug and, unfortunately, the bug still exists in master/trunk (17.0.0 20260507). I'll revisit this in three months time when gcc 16.2.0 releases. If it's not fixed by then, well... let's not think about that yet I guess...
Compiling with ` -Wall -Wextra -fno-strict-aliasing -fwrapv -fno-aggressive-loop-optimizations -fsanitize=address,undefined` moves the segfault to the first occurrence of this same gid<->lid conversion pattern.
I did run this through gdb, and the data structures seem fine. I also tried manually applying the subroutine, as if inlining on the spot inside gdb, and it carries on a bit further until:
```
At line 520 of file /redacted/PSyclone/external/lfric_infrastructure//src/mesh/local_mesh_mod.f90
Fortran runtime error: Index '1' of dimension 2 of array 'self%vert_coords' outside of expected range (25606685:192)
```
Even though:
```
(gdb) p self%_data%n_unique_vertices
$26 = 16
(gdb) p self%_data%vert_coords
$27 = ((0, 0, 0) <repeats 16 times>)
(gdb) p self%_data%vert_coords(:,1)
$28 = (0, 0, 0)
```
So it just seems things have been corrupted beyond repair.

... Fortran is tough business.

nmnobre:
Update: good news, this has been reported in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=125192, and I've confirmed the patch suggested there works for us.

PR #3343, opened 2026-02-20
(towards #1960 and #1799) Improve ArrayAssingment2Loop using call/routinesymbol datatype information

sergisiso:
I started this branch to tackle the WHERE with reduction intrinsics, but as in my previous PRs I think this should be solved first in ArrayAssignment2Loop and then integrate it to the WHERE.

The reductions that end up with a scalar can are already be done because we know that that term does not need additional indexing after applying ArrayAssignment2Loop. However the call.datatype was quite limited. It worked for Intrinsics but not generic calls (even when we had a RoutineSymbol with information). And RoutineSymbols defaulted to NoType, even when that was not necessarely the case (I think UnresolvedType is a better default and use the frontend type, e.g. is a CallStmt, to give it the NoType).

I have made the changes needed to fix the above and I feel its already enough for a PR, the WHEREs will come in a follow up.

PR #3429, opened 2026-05-07
Generalise hoist_arguments_to_temporaries in NEMO utils to all arguments with array operations 

LonelyCat124:


LonelyCat124:
Some failures after this change can be resolved with a change to reference2arrayrange_trans:

```
            if not ((isinstance(node.parent, IntrinsicCall) and
                 node.parent.intrinsic in REDUCTION_INTRINSICS)
                or node.parent.is_elemental):
                return
```
instead of
`if not node.parent.is_elemental`
However, this still fails for stpctl.f90 for somer aeason, as we end up with an UnresolvedType in the gen_decls part of the backend, which is a bit confusing, I'm trying to track that down.

LonelyCat124:
I'm trying building today and it just seems to work, I will test full integration and see if any others fail.

LonelyCat124:
From discussion with sergi the previous direction has potential to produce incorrect code.

Instead we should change operation order:
1) Do the array transformations that don't involve new temporaries
2) Create new temporaries
3) Reapply the array transformations.

LonelyCat124:
> Another option is to move the
- convert_array_notation
- loopify_array_intrinsics
- convert_range_loops
- hoist_argument_expressions
parts of normalise_loops into a loopifyTrans metatransformation which I wanted to do for while, no individual enabling flags needed.
Then the DataNodeToTempTrans apply can call this loopifyTrans on the extracted expression as it still have them inside the apply.

We don't have metatransformation support yet for this, but I'll make an issue for this as well.

LonelyCat124:
@sergisiso I think this is ready for a first look now.

examples/nemo/scripts/utils.py
```diff
@@ -225,6 +221,9 @@ def normalise_loops(
                 Reference2ArrayRangeTrans().apply(reference)
             except TransformationError:
                 pass
+            except Exception as err:
+                print(reference, reference.parent.parent.debug_string())
+                raise err
```
sergisiso: What happens here? Is this a leftover?

examples/nemo/scripts/utils.py
```diff
@@ -280,6 +279,22 @@ def normalise_loops(
                 except TransformationError:
                     pass
 
+    # TODO #3412: This is currently limited to iom_put, we want to expand it
+    # throughout the code
+    if hoist_argument_expressions:
+        iom_put_argument_to_temporary(schedule.walk(Call))
```
sergisiso: This is now no longer limited to iom_put right? So the TODO can be removed and the method name updated?

examples/nemo/scripts/utils.py
```diff
@@ -550,13 +565,17 @@ def iom_put_argument_to_temporary(calls: list[Call]):
...
-            for arg in call.arguments:
-                dtype = arg.datatype
-                if (isinstance(dtype, ArrayType) and
-                    (isinstance(arg, Operation) or
-                        isinstance(arg, IntrinsicCall))):
-                    try:
-                        DataNodeToTempTrans().apply(arg, verbose=True)
-                    except TransformationError:
-                        pass
+        # if call.symbol.name == "iom_put":
```
sergisiso: Remove

examples/nemo/scripts/utils.py
```diff
@@ -550,13 +565,17 @@ def iom_put_argument_to_temporary(calls: list[Call]):
...
+        for arg in call.arguments:
+            dtype = arg.datatype
+            if (isinstance(dtype, ArrayType) and
+                (isinstance(arg, Operation) or
+                    isinstance(arg, IntrinsicCall))):
+                try:
+                    if (isinstance(dtype.elemental_type, ScalarType)
+                        and dtype.elemental_type.intrinsic ==
+                            ScalarType.Intrinsic.CHARACTER):
+                        continue
```
sergisiso, line 577: Is this an issue with the transformation? I am wondering if it should be inside the validation instead?

examples/nemo/scripts/utils.py
```diff
@@ -550,13 +565,17 @@ def iom_put_argument_to_temporary(calls: list[Call]):
...
-                    try:
-                        DataNodeToTempTrans().apply(arg, verbose=True)
-                    except TransformationError:
-                        pass
+        # if call.symbol.name == "iom_put":
+        for arg in call.arguments:
+            dtype = arg.datatype
+            if (isinstance(dtype, ArrayType) and
+                (isinstance(arg, Operation) or
+                    isinstance(arg, IntrinsicCall))):
```
sergisiso, line 570: Maybe a comment explaining the condition: "Only extract expressions that can potentially be loopyfied (e.g. operations over arrays)"

src/psyclone/tests/psyir/transformations/intrinsics/sum2loop_trans_test.py
```diff
@@ -141,3 +141,21 @@ def test_apply(fortran_reader, fortran_writer, tmpdir):
     result = fortran_writer(psyir)
     assert expected in result
     assert Compile(tmpdir).string_compiles(result)
+
+
+def test_nemo5_transerror_case(fortran_reader):
+    '''Test that the transformation fails correctly if we have nested sums
+    as the reference can't be converted to an arrayreference.'''
```
sergisiso: Can you change the name and description of the test to make it generic about the problem tested here, e.g. "the array reference is inside a non-elemental function and therefore won't be converted"?

src/psyclone/tests/psyir/transformations/intrinsics/sum2loop_trans_test.py
```diff
@@ -141,3 +141,21 @@ def test_apply(fortran_reader, fortran_writer, tmpdir):
...
+    code = """subroutine sum_test()
+    integer :: n, m
+    real , dimension(:, :) :: array
+
+    result = sum(sum(array, dim=2)) * array(n,m)
```
sergisiso, line 155: Since the validation counts ArrayReference, maybe test that expressions with other arrays like:
`sum(sum(array  + array(:,:), dim=2))`
or
`sum(sum(array, dim=dimensions(2)))`

work as expected.

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
```diff
@@ -299,6 +299,9 @@ def test_datanodetotemptrans_apply(fortran_reader, fortran_writer, tmp_path):
     psyir = fortran_reader.psyir_from_source(code)
     assign = psyir.walk(Assignment)[0]
     dtrans.apply(assign.rhs.operands[1])
+    # Check the tmp symbol is at the routine scope.
+    routine = psyir.walk(Routine)[0]
+    assert routine.symbol_table.lookup("tmp") is not None
```
sergisiso: `assert "tmp" in routine.symbol_table` ?

sergisiso (changes requested):
Thanks @LonelyCat124, its good to see it has passed all integration tests without making it exclusive for iom_put. Just some extra clarifications and testing needed.

examples/nemo/scripts/utils.py
LonelyCat124: Yeah this was debugging.

examples/nemo/scripts/utils.py

examples/nemo/scripts/utils.py
LonelyCat124: Removed.

examples/nemo/scripts/utils.py
LonelyCat124: I'm not sure - the transformation does it successfully but there was some error, I don't remember whether compiler or elsewhere in PSyclone. I can have a further look if you want, I just decided to skip it here for strings (not sure if there would be some other valid use case?)

examples/nemo/scripts/utils.py
LonelyCat124: Added

src/psyclone/tests/psyir/transformations/intrinsics/sum2loop_trans_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/transformations/intrinsics/sum2loop_trans_test.py
LonelyCat124: So the first one already doesn't work as expected.

```
    code = """subroutine sum_test()
    integer :: n, m
    real, dimension(:, :) :: array
    result = sum(sum(array  + array(:,:), dim=2))
    end subroutine"""
    intrinsic_node = psyir.children[0].children[0].rhs
    trans = Sum2LoopTrans()
        trans.apply(intrinsic_node)
    print(fortran_writer(psyir))
    print(err.value)
    assert False
```
This outputs an error message on the fortran_writer stage, and the err.value is also different (if it were reached):
```
E           psyclone.psyir.backend.visitor.VisitorError: Visitor Error: The following symbols are not explicitly declared or imported from a module and there are no wildcard imports which could be bringing them into scope: 'result'
```
and
```
Transformation Error: ArrayAssignment2LoopsTrans could not convert the expression:
array(:,:) = SUM(array(:,:) + array(:,:), 2)

 into a loop because:
Transformation Error: ArrayAssignment2LoopsTrans does not support statements containing dependencies that would generate loop-carried dependencies when naively converting them to a loop, but found:
array(:,:) = SUM(array(:,:) + array(:,:), 2)
```

I'm not sure if this is the right place to fix this - I can make some xfailing tests and an issue?

src/psyclone/tests/psyir/transformations/datanode_to_temp_trans_test.py
LonelyCat124: Yeah

src/psyclone/tests/psyir/transformations/intrinsics/sum2loop_trans_test.py
LonelyCat124: @sergisiso I've addressed the other comments but this one is still open if you have any time to let me know how to proceed before another full review.

sergisiso: Both errors seem correct:
- result is not declared in the example
- And TransformationErrors are fine, what we don't want is to validate and then generate incorrect code.

So I would say no xfail needed?

LonelyCat124: Ah thats my mistake in the example, I was concerned the resulting code may be changed but I still get the input so thats fine. I'll check the other example you suggested as well.

LonelyCat124: I tested the other examples - the error messages are different, but the transformation does fail as expected.

LonelyCat124:
@sergisiso I think this is ready for review again? I'm not sure why i didn't mark it as so last week so if I've missed something let me know.

examples/nemo/scripts/utils.py
sergisiso: I found out in #3343 that node.datatype of a array_reference with a character string is special because those are not array with shapes but ScalarType with length. And therefore its datatype result is wrong. Could this be the same problem?

(If this PR is ready before #3443 I can explore it, e.g. remove the condition, there)

sergisiso (approved):
@LonelyCat124 This is approved for merging, we still don't know what's the issue with characters and why we don't see the expected speedup, but it is a big step to generalise the functionality and it passes all tests

sergisiso:
@LonelyCat124 The linkspector action is failing, but I will proceed for now and monitor if it gets fixed eventually.

PR #3433, opened 2026-05-14
(closes #3432) Transformation functionality to split and propagate kwargs to subtransformations

### LINKED ISSUE #3432: Split and propagate kwargs into multiple sub-transformations
>I think if you make a metatransformation it would work anyway (i.e. if you have class MaximalOMPParallelRegionTrans(MaximalRegionTrans, OMPParalllelTrans)) via get_valid_options

Ah, we had a naming confusion here, I was using metatransformations to mean transformations that call other transformations, not that inherit from multiple transformations. I haven't though about apply/validate MRO (but I don't think this PR should go that far)

> I don't like the if not found branch though , as it already duplicates checks done by validate_options

What about:
```
self_kwargs = dict(kwargs)
trans_kwargs = {}
for key in kwargs:
    if key in type(self._trans).get_valid_options():
        trans_kwargs[key] = kwargs[key]
            del self_kwargs[key]
```
This would only remove self_kwards IFF they are in _trans and not self, so self.validate_options still works as expected for invalid options.

> I have a slight reservation with just automatically making all of the options of self._trans for all MaximalRegionTrans going forward without requiring full testing of all options (which could be a pain), but maybe its just handled by the failure states we have already.

I don't think that's true, it does full testing of all options some through
self.validate(self_kwargs)
others with
self._compute_transformable_sections -> trans.validate(current_block + [child], trans_kwargs)

(the trans.validate is missing but I think we need it there otherwise we validate a potentially different thing)

_Originally posted by @sergisiso in https://github.com/stfc/PSyclone/issues/3431#issuecomment-4449771813_

sergisiso:


sergisiso:
@LonelyCat124 This is ready for a first look. 

For now I added a manual docstring mention that creates the correct link (see image). Maybe it can also be done automatically in `docstring_parser.py`, but I couldn't figure out where.
image

src/psyclone/tests/psyGen_test.py
```diff
@@ -205,6 +205,105 @@ def apply(self, node, valid: bool = True):
...
+            # If we want a keyword argument to not be exclusively consumed by
+            # this transformation and propagate it, put it back into kwargs
+            kwargs['common_option'] = common_option
+            # If we want to consume it use it by name
+            self.validate(
+                node,
+                meta_option=meta_option,
+                **kwargs)
+
+            self._trans1().apply(node, **kwargs)
```
LonelyCat124: I'm a little concerned about this section of the test.

In general, self._trans1().apply would have a validate that calls `validate_options(**kwargs)`, and won't know anything about the other parts of the metatransformations.

I think this should fail as is written here, and the apply also needs to do:
`self_kwargs, tr1_kwargs, tr2_kwargs = self.split_kwargs(meta_option=meta_option, common_option=common_option, kwargs)` 

I also don't think we need to put `common_option` into kwargs, but it should be fine to just pass it to validate with as `common_option=common_option` (it then just appears in `kwargs` in the `validate` anyway).

LonelyCat124: I think we should test this fails if we don't use split_kwargs as well as the working version (which should just involve calling validate_options in the Called1Trans and Called2Trans).

src/psyclone/psyGen.py
```diff
@@ -2353,6 +2358,36 @@ def name(self):
...
+        kwargs: dict[str, Any],
+    ) -> tuple[dict[str, Any]]:
+        '''
+        :param kwargs: the list of kwargs to split.
+
+        :returns: a tuple of the kwargs dictionaries that are valid for this
+            transformation and every other transformation listed in the
+            _SUB_TRANSFORMATIONS list. The first kwargs (the ones for itself)
+            will also include any key that is not valid in any of the other
+            transformation (e.g. unsupported options).
```
LonelyCat124: Probably this should say why (i.e. to ensure validate_options correctly shows errors in user input).

src/psyclone/psyGen.py
```diff
@@ -2353,6 +2358,36 @@ def name(self):
...
+        # and delete them from the first one if they are valid somewhere
+        # else but not in the self options
+        for key in kwargs:
+            for idx, trans in enumerate(self._SUB_TRANSFORMATIONS):
+                if key in trans.get_valid_options():
+                    other_dicts[idx][key] = kwargs[key]
+                    if key not in type(self).get_valid_options():
+                        del first_dict[key]
+
+        return tuple([first_dict] + other_dicts)
```
LonelyCat124: Is this best or can we just do `return first_dict, other_dicts`? I'm happy with either approach, just not sure which is more pythonic (vs which is easier to type)

LonelyCat124 (changes requested):
@sergisiso I think this looks mostly good to me - I have some slight reservations with the tests (and not reaching some fail cases that I think it should as written now, with a better way to write some parts) so I've asked for a few changes and an additional test.

src/psyclone/psyGen.py
sergisiso: I am not particularly happy having the list addition involved and would prefer a more pythonic solution here but
`return first_dict, other_dicts`  returns a `tuple(item, tuple(other_items))` which is not what I want as I quite like the current destructuring in the caller

src/psyclone/psyGen.py

src/psyclone/tests/psyGen_test.py
sergisiso: > In general, self._trans1().apply would have a validate that calls validate_options(**kwargs), and won't know anything about the other parts of the metatransformations.

Ah good point, I incorrectly thought that was probably what the base class was already doing.

> I think this should fail as is written here

Indeed, after adding `def validate(self, node, **kwargs): self.validate_options(**kwargs)` it fails.

> and the apply also needs to do ...

And indeed, splitting them in the apply fixes the previous faliure

> I think we should test this fails if we don't use split_kwargs as well as the working version

I am not sure about this, it was my mistake not adding a validate_options. But we already test somewhere else that if a validate_options does recieve keywords not supported by that transformation it fails. But let me know if you disagree.

> I also don't think we need to put common_option into kwargs

If we don't put it back it gets consumed by this transformation but not propagated to the sub-transformations (e.g. the `assert common_option` on the CalledTransformation will fail). For things like 'verbose' we typically want to propagate it, but it will be up to each metatransformation to decide if it consumes it or it propagates it.

sergisiso:
@LonelyCat124 This is ready for another review

src/psyclone/tests/psyGen_test.py
LonelyCat124: This mostly looks good now.

> If we don't put it back it gets consumed by this transformation but not propagated to the sub-transformations (e.g. the assert common_option on the CalledTransformation will fail). For things like 'verbose' we typically want to propagate it, but it will be up to each metatransformation to decide if it consumes it or it propagates it.

I think the best fix for this is not to add to kwargs, but instead do `_, tr1_kwargs, tr2_kwargs = self.split_kwargs(meta_option=meta_option, common_option=common_option, kwargs)` and let the python kwargs packing on the subclass handle this, rather than us needing to do it ourselves, the additional keyword options will appear in the kwargs for split_kwargs and appear as needed.

Modifying kwargs can potentially have a side effect - see https://stackoverflow.com/questions/45883655/is-it-always-safe-to-modify-the-kwargs-dictionary top answer.

src/psyclone/psyGen.py
LonelyCat124: @sergisiso Perhaps using the * to unpack `other_items`?:
```
>>> def a():
...    b = [1, 2]
...    return 3, *b
...
>>> z = a()
>>> z
(3, 1, 2)
>>> x, y, z, = a()
>>> x
3
>>> y
1
>>> z
2
```
The length isn't constant, but since its always `1 + len(...)` we can know it in the code anyway without much difficulty.

LonelyCat124:
@sergisiso Have a couple more suggestions but I think we're closer now.

We probably need a section in the docs discussing metatransformations (or transformations that primarily call other transformations, or some other title you prefer) and what is supported (This stuff) and what isn't (doc option inheritance) with perhaps a TODO to either #3330 or a new issue. I think this subtransformation stuff should probably make doc option inheritance a lot easier for metatransformations, but I'll investigate it in #3345 .

I think w.r.t #3431 this will enable option support (with small changes) and then we add a TODO for the undocumented inherited arguments again.

src/psyclone/tests/psyGen_test.py
sergisiso: Ah, I didn't understand what you meant before, I like your solution

src/psyclone/psyGen.py

src/psyclone/tests/psyGen_test.py
sergisiso: One problem of this is that now split_kwargs needs to accept: `**kwargs` which I still don't quite know how to typehint. Apparently there is a solution coming https://peps.python.org/pep-0692/ , but not available to our minimum 3.9, so I will just leave it untypehinted for now.

sergisiso:
@LonelyCat124 I adressed your comments, added documentation and fixed outdated links to bitbucket that were making the linkcheck fail. Ready for another view

LonelyCat124 (approved):
All looks good and ITs pass, will proceed to merge.

PR #3436, opened 2026-05-19
(closes #3135) Fix cyclic dependencies by replacing Scalartype import-time instantiations

### LINKED ISSUE #3135: Circular imports resulting in impossible imports in PSyclone internals
There are some issue with how we currently resolve/designed our import structure meaning certain files cannot import certain other PSyclone components, for example there is no way to import `Literal` from inside `psyir.symbols.datatypes`.

It appears this is due to our use of `__init__.py` files containing the imports for every file in a module, and being used as a reference to that module in places in PSyclone. This makes user scripting significantly easier, but is the most likely cause of our headaches.

Our currently discussed solution to try is:
1. Introduce a `psyclone.api` file (or similar if we want to have a bit more granularity we could have a few, e.g. `psyclone.psyir.api`, `psyclone.domain.lfric.api` etc.) that contains all of the expected externally visible classes, e.g. nodes/transformations/symbols/datatypes.
2. Use fully explicit paths for internal imports, e.g. rather than `psyclone.psyir.nodes import BLAH` we would need to move to an import structure where everything is `psyclone.psyir.nodes.blah import BLAH`.

The difficulty with this change is that it will be disruptive to users for us to have any internal benefit - we will need to add deprecation messages when users use `from psyclone.psyir.transformations import my_transformation`, but I expect this may just happen from random files in psyclone while the `__init.py__`. Perhaps one way around this is to add new files like `node.py` inside the `psyir/` directory to use for imports for now, but I'm not sure if/how that would work.

I think realistically this is quite urgent, so deprecation and the `psyclone.api` being available in 3.3 would be good, and depending on whether improvements can be used internally from that, removal from 3.4 (i.e. early-mid 2026) if users are ok with that. It is a big task though, and the suggested proposal here may not function as expected/hoped, so we will have to see.

sergisiso:
I hit #3135 again where some cyclic dependencies cannot be prevented by adding the import inside methods because they happen when symbols are initialised at import-time. To remove this import time instantiations I opted for creating ScalarType staticmethods instead of the factory proposed in the issue. The change itself is small, but it is repeated many many times.


sergisiso:
@LonelyCat124 @arporter This is ready for review

doc/developer_guide/psyir.rst
```diff
@@ -167,10 +167,10 @@ node can be integrated and used in the PSyIR tree:
 ::
 
     >>> from psyclone.psyir.nodes import Literal, Schedule
-    >>> from psyclone.psyir.symbols import INTEGER_TYPE
+    >>> from psyclone.psyir.symbols import ScalarType.integer_type()
```
LonelyCat124: Will this import work? I suspect not?

doc/user_guide/psyir.rst
```diff
@@ -618,12 +625,12 @@ A more complicated access involving arrays of structures such as
 
 .. code-block:: python
 
-  from psyclone.psyir.symbols import INTEGER_TYPE
+  from psyclone.psyir.symbols import ScalarType.integer_type()
```
LonelyCat124: Bad import. Check this everywhere with grep or similar I think.

LonelyCat124 (changes requested):
@sergisiso I struggled to look through every single change (my browser is dying looking at the diffs...) but I found some minor issues to check with imports in examples and docs - can you check these? Then I can probably merge.

doc/developer_guide/psyir.rst
sergisiso: These are certainly invalid, it probably gives away that I did the bulk of the conversion with regex and this docstrings escaped my replacement expressions.

PS: I have made progress in enabling docs and docstrings snippet checking that would have spotted this. I will attempt to commit it soon.

doc/user_guide/psyir.rst
sergisiso: Done, I found another one with grep

sergisiso:
@LonelyCat124 This is ready for another review.

LonelyCat124:
Set ITs running as a sanity check but otherwise all is good.

sergisiso:
@LonelyCat124 the integration test spotted a INTEGER_TYPE in lfric_apps. I will push a fix now

sergisiso:
~~@LonelyCat124 This is passing all integration test now, my lfric change from yesterday failed with signal 11, but after rerunning the action it has succeeded.~~

sergisiso:
@LonelyCat124 This has finally passed all tests. I had to do it with multiple lfric_apps hash update attempts as I had no access to the system to test it directly.

LonelyCat124:
Ok, now ITs pass I'll proceed to merge.

LonelyCat124 (approved):

PR #3438, opened 2026-05-20
(Closes #3435) Metatrans docs

### LINKED ISSUE #3435: SUB_TRANSFORMATION docstirng inheritance
The new _SUB_TRANSFORMATION member of transformations needs docstring inheritance for their options. My idea here is to have sections like this in the docstring:
```
:param blah: ...
:param blah2: ...

Options for sub transformations SUBTRANS1

:param blah3: ...
:param blah 4: ...

Options for sub transformation SUBTRANS2

:param asd: ...

...
```

Rather than just one long list of keyword options that don't distinguish what they affect (note this won't affect the inherited options from parents).

This needs a couple of bits of changes:
1. DocstringData needs some sub_arguments member (or similar), which is a (Ordered)Dict of transformationname to arguments (OrderedDict) for sub members. We need to be able to add sub_arguments into the DocstringData from another DocstringData. `merge` will also need this update.
2. DocstringData gen_docstring needs to use these correctly.
3. The transformation_documentation_wrapper inherit option will be removed, and instead inheriting from ancestors will always happen, and SUB_TRANSFORMATION options will always be included. The main reason for this is that doing `@trasnforamtion_documentation_wrapper(inherit=False)` doesn't work anyway, so this is a pretty bad option. I will do some research though and see if there is a way to add options onto class decorators, as being able to disable sub transformation options is probably something we might want in the future.

LonelyCat124:
This is the initial stage to have support for docs from _SUB_TRANSFORMATIONS supported.

There are remaining issues to fix:
1. No unit tests cover this code.
2. We can get repeated options which result in repeated docs:
```
>>> print(MaximalOMPParallelRegionTrans.apply.__doc__)
    Applies the transformation to the nodes provided.


    :param nodes: can be a single node, a schedule or a list of nodes.
    :param bool node_type_check: whether or not the type of the
        nodes enclosed in the region should be tested to avoid using
        unsupported nodes inside a region.
    :param bool node_type_check: (Option used for OMPParallelTrans) whether or not the type of the
        nodes enclosed in the region should be tested to avoid using
        unsupported nodes inside a region.
```

`MaximalOMPParallelRegionTrans` inherits the `node_type_check` from `RegionTrans`, but then also finds it imported from its `SUB_TRANSFORMATION` class. @sergisiso Do you think I should avoid repeated docstrings? Should it be limited in some way to only be if the strings are `==`? I think I can do that check (as the "Option used for ...") is added later.


LonelyCat124:
@sergisiso This should be ready for a first look now. Also let me know if you're happy with all the behaviour shown in `test_transformation_doc_wrapper_subtrans` from utils_test.py

src/psyclone/utils.py
```diff
@@ -88,12 +89,31 @@ def stringify_annotation(annotation) -> str:
     return str(annotation)
 
 
-def transformation_documentation_wrapper(cls, *args, inherit=True, **kwargs):
+def transformation_documentation_wrapper(*args, inherit=True,
+                                         add_subtransformations: bool = True,
+                                         **kwargs):
     '''
     Updates the apply and validate methods' docstrings for the supplied cls,
```
sergisiso: cls is not supplied anymore?

src/psyclone/utils.py
```diff
@@ -88,12 +89,31 @@ def stringify_annotation(annotation) -> str:
...
+
+    whilst the decorator without any argument is simply:
+
+    >>> transformation_documentation_wrapper(myclass)
+
+    We use this to vary the behaviour of the wrapper slightly depending on
+    whether any arguments are present.
 
-    :param Class cls: The class whose docstrings are to be updated.
     :param inherit: whether to inherit argument docstrings from cls' parent's
```
sergisiso, line 136: cls not available anymore?

src/psyclone/utils.py
```diff
@@ -102,11 +122,13 @@ def transformation_documentation_wrapper(cls, *args, inherit=True, **kwargs):
                     Transformation's validate docstring from its own
                     apply docstring.
     :type inherit: Union[list[Class], bool]
```
sergisiso: Make this typehints

src/psyclone/utils.py
```diff
@@ -102,11 +122,13 @@ def transformation_documentation_wrapper(cls, *args, inherit=True, **kwargs):
...
+    :param add_subtransformations: Whether to add parameter docstrings from
+        sub transformations used by this Transformation.
     '''
     # List of argument doctrings to never inherit.
```
sergisiso, line 146: inherited?

src/psyclone/utils.py
```diff
@@ -153,16 +176,42 @@ def wrapper():
                  )
         else:
             added_parameters = None
+        if add_subtransformations and len(cls._SUB_TRANSFORMATIONS) > 0:
+            if added_parameters is None:
+                added_parameters = DocstringData(
+                    desc=None, arguments=OrderedDict(),
+                    raises=[], returns=None,
```
sergisiso: The types of desc and return don't match, should they be declared Optional[...]  ?

src/psyclone/utils.py
```diff
@@ -141,7 +163,8 @@ def wrapper():
         if isinstance(inherit, list):
```
sergisiso, line 181: Isn't inherit a boolean in the argument list?

src/psyclone/utils.py
```diff
@@ -88,12 +89,31 @@ def stringify_annotation(annotation) -> str:
...
 
 
-def transformation_documentation_wrapper(cls, *args, inherit=True, **kwargs):
+def transformation_documentation_wrapper(*args, inherit=True,
+                                         add_subtransformations: bool = True,
+                                         **kwargs):
     '''
     Updates the apply and validate methods' docstrings for the supplied cls,
-    according to the value of inherit.
+    according to the value of inherit. args is either a length 1 set argument
```
sergisiso: I find the sentence explaining 'args' confusing, could you make the examples below part of the main sentence, something like (if I understood correctly):

Args is either a 1-argument set containing the class to be wrapped:

`>>> transformation_documentation_wrapper(inherit=True)(myclass)`

or  an empty set if the options are specified in the decorated classs:
```
>>> transformation_documentation_wrapper
class myclass():
    pass
```

Also it may be good to explain the relation between args and class, even if it overexplains how decorators works. It is not clear to me just looking at this implementation.

src/psyclone/utils.py
```diff
@@ -88,12 +89,31 @@ def stringify_annotation(annotation) -> str:
...
+    containing the class to be wrapper, or a length 0 argument set if
+    options are specified on the transformation_docstring_wrapper that is
+    handled by python. This works due to:
+
+    >>> @transformation_documentation_wrapper(inherit=True)
+    >>> class myclass():
+    >>>     pass
```
sergisiso: Also if you use the `>>>` make sure the example works, right now I get:
```
pytest --doctest-modules src/psyclone/utils.py
>>> @transformation_documentation_wrapper(inherit=True)
SyntaxError: invalid syntax
```

These are becauses `>>>` have to be complete blocks and use the `...` for continuations
```
>>> @transformation_documentation_wrapper(inherit=True)
... class myclass():
...    pass
```

src/psyclone/utils.py
sergisiso: Can we create the empty DocstringData at the top instead of inside each condition?

src/psyclone/utils.py
```diff
@@ -153,16 +176,42 @@ def wrapper():
...
+            if add_subtransformations:
+                for trans in cls._SUB_TRANSFORMATIONS:
+                    for arg in list(
+                            added_parameters.sub_arguments[
+                                trans.__name__
+                            ].keys()
+                    ):
+                        if arg in _uninheritable_args:
+                            del added_parameters.sub_arguments[
+                                    trans.__name__][arg]
```
sergisiso, line 223: Would it be simpler to remove them from "inherited_params" sets before they are added/merged to added params, we would not need to replicate the whole if conditions here I think

src/psyclone/docstring_parser.py
```diff
@@ -85,16 +87,22 @@ class ArgumentData():
...
+    def gen_docstring(self, function: Union[None, Callable[..., Any]] = None,
+                      cls_name: str = "") -> str:
         '''
         :param function: The function who the generated docstring will be for.
                     Default option is None. If no function is supplied, there
                     can be no type annotation and so the type information is
                     included inline in the :param or in a separate :type
                     entry.
+        :param cls_name: The cls_name to use for saying which subclass (or
+            otherwise) this argument is from. Used for documenting
```
sergisiso: What does the "or otherwise" refer here?

src/psyclone/docstring_parser.py
```diff
@@ -85,16 +87,22 @@ class ArgumentData():
...
+            sub transformation options.
+
         :returns: The docstring represented by this ArgumentData.
         '''
+        if cls_name:
+            cls_name = f"(Option used for {cls_name}) "
```
sergisiso: Maybe: `Option provided to {cls_name}` /

src/psyclone/tests/utils_test.py
```diff
@@ -99,7 +99,8 @@ def test_transformation_doc_wrapper_non_transformation():
 def test_transformation_doc_wrapper_single_inheritance():
     '''Test the transformation_doc_wrapper.'''
 
-    # Create a base transformation class
+    # Createa base transformation class
```
sergisiso: remove 'a'

src/psyclone/tests/utils_test.py
```diff
@@ -359,3 +354,129 @@ def func(temp: bool, temp2: Union[bool, int]):
...
+    correct = """Super apply docstring
+    
+    
+    :param opt1: opt1 docstring.
+    :param opt2: opt2 docstring.
+    :type opt2: opt2 type.
+    :param opt3: (Option used for SubTrans1) opt3 docstring.
+    :type opt3: int
+    :param int opt3: (Option used for SubTrans2) opt3 docstring.\
+"""  # noqa: W293
```
sergisiso: Why is `# noqa: W293` needed, since this is a "in" assert, can we finish the string at the end of the last line?

sergisiso (changes requested):
I am happy with the behaviour @LonelyCat124, but the implementation is a bit hard to follow, specially the utils.py could benefit with more clear typehints and docstrings.

src/psyclone/utils.py
LonelyCat124: So sometimes it is, but its hidden in `*args` in the cases its provided here instead of to wrapper. I've rewritten these docs to not mention cls.

src/psyclone/utils.py
LonelyCat124: The rewrite at 214-217 needs `cls` defined on wrapper, this allowed both
`@transformation_documentation_wrapper` and `@transformation_documentation_wrapper(arguments, go, here)` to be usable. The argument list being provided changes whether `cls` is provided as an argument to the outer function, so `cls` is now defined on `def wrapper` and then passed according to the definition.

src/psyclone/utils.py
LonelyCat124: Done

src/psyclone/utils.py
LonelyCat124: I don't think so, sub transformations aren't inherited from in the current implementations.

src/psyclone/utils.py
LonelyCat124: From a quick google they're not allowed to be Optional. I'll update these calls to do something valid.

src/psyclone/utils.py
LonelyCat124: No, its either a list of transformations or a boolean (`list[Type[Transformation]] | bool`).

src/psyclone/utils.py
LonelyCat124: I can't see how to build it into the original sentence, however I've separated it into its own  paragraph and rewritten a lot of it. I'm a bit concerned the docstring here is overly technical (vs useful), though since its really a PSyclone developer facing utility maybe thats ok.

src/psyclone/utils.py
LonelyCat124: Fixed.

src/psyclone/utils.py
LonelyCat124: I could but I struggle to see the benefit - we'd sometimes (most of the time in the actual PSyclone code) be creating a DocstringData object we never use - the cost is small but I suspect this may get called a lot so everything may count.

src/psyclone/utils.py
LonelyCat124: I could move the sub transformation inherited parameters removal into the add_subtransformations bit, but we'd still end up with this section for the normal parameters. Its mostly up to you regarding whether you'd rather have all the argument removal together or the if statements together.

I could also refactor this into its own function and then you just have a call to `remove_uninheritable_args`.

Let me know your preference.

src/psyclone/docstring_parser.py
LonelyCat124: At the moment it refers to SUB_TRANSFORMATION (as they're not actually subclasses necessarily), but I can just remove it and `subclass` and just say `class`

src/psyclone/docstring_parser.py
LonelyCat124: Sure

src/psyclone/tests/utils_test.py
LonelyCat124: Fixed.

src/psyclone/tests/utils_test.py
LonelyCat124: Its not due to that, its due to the two blank lines that contain whitespace.  The \ and """ are just to make sure I can add the noqa: W293. The comment before the string explains why.

LonelyCat124:
@sergisiso Ready for another look, most comments addressed, a couple I've just responded to with comments/questions for this review cycle.

src/psyclone/docstring_parser.py
sergisiso: I suggest: "The cls_name to use for saying which argument is this from. Used for documenting sub transformation options."

src/psyclone/tests/utils_test.py
sergisiso: I see, I think I prefer `correct = """Super apply docstring\n    \n    \n` to avoid this issue

sergisiso (approved):
@LonelyCat124 This is approved for merging, I end up doing the final suggestions myself since only involved comments and formatting, but let me know if you disagree with anything.

PR #3439, opened 2026-05-20
(Closes #3415) Avoid nested ArrayTypes on StructureType types

### LINKED ISSUE #3415: StructureType.datatype can return an ArrayType with an ArrayType intrinsic.
the result of this section:
```
        # Walk down the structure, collecting information on any array slices
        # as we go.
        while isinstance(cursor, (StructureMember, StructureReference)):
            cursor = cursor.member
            if isinstance(cursor_type, ArrayType):
            if not isinstance(cursor_type, (UnresolvedType, UnsupportedType)):
                # Once we've hit an Unresolved/UnsupportedType the cursor_type
                # will remain set to that as we can't do any better.
                cursor_type = cursor_type.components[
                    cursor.name.lower()].datatype
            try:
                cursor_shape = _get_cursor_shape(cursor, cursor_type)
            except NotImplementedError:
                return UnresolvedType()
            if cursor_shape:
                if shape:
                    raise NotImplementedError(
                        f"Array of arrays not supported: the ultimate member "
                        f"'{cursor.name}' of the StructureAccess represents "
                        f"an array but other array notation is present in the "
                        f"full access expression: '{self.debug_string()}'")
                shape = cursor_shape
```

is `Array<Scalar<REAL, UNDEFINED>, shape=[10, 10]>` (for the datatype accessed in `tests/psyir/transformations/intrinsics/array_reduction_base_trans_test.py::test_structure_error` , and then the code does
```
        if shape:
            return ArrayType(cursor_type, shape)
```
resulting in the ArrayType<ArrayType> datatype.
From discussion in #3373 we should never get an ArrayType with an ArrayType intrinsic so this is unintended and needs to be updated/fixed.

LonelyCat124:


LonelyCat124:
@sergisiso This is ready for a first look, I'm hoping this is simple enough.

sergisiso (approved):
@LonelyCat124 I agree with the fix and it has passed all test, this is approved for merging.

sergisiso:
@LonelyCat124 Oh before merging, there are 4 todos in other tests that now can be clean up. Could you do that before we close the issue?

LonelyCat124:
@sergisiso Done.

PR #3440, opened 2026-05-26
Update to fparser 0.2.4

sergisiso:


sergisiso:
One pytest failed, I am a bit confused because adding the same fortran snippet into a file and doing: `fparser2 test.f90` and `psyclone test.f90` both succeed.

arporter:
> One pytest failed, I am a bit confused because adding the same fortran snippet into a file and doing: `fparser2 test.f90` and `psyclone test.f90` both succeed.

Could it be a Python version thing?

sergisiso:
No, I am using the same environment for everything. The problem is that the test requests a fixed_form parser and the snippet is free_form. Maybe before we just ignored the syntaxerror because it starts with a blankline and we fixed this issue now.

This is not a problem in most cases because our FortranReader defaults to free_form True. But this particular tests uses the FparserReader that defaults to free_form False.

sergisiso:
@arporter @LonelyCat124 This is ready for review

arporter:
I suggest we pause this while we get 0.2.4 of fparser out.

sergisiso:
@arporter I unblocked the PR and marked ready for review again.

.github/workflows/python-package.yml
```diff
@@ -102,10 +117,10 @@ jobs:
           python -m pip install --upgrade pip
           # Some of the examples use Jupyter.
           pip install jupyter
-          # We need to install sphinx to get correct doc testing
           # Uncomment the below to use the submodule version of fparser rather
           # than the latest release from pypi.
-          # pip install external/fparser
+          pip install external/fparser
```
arporter: This needs commenting out again now that 0.2.4 is on pypi.

.github/workflows/python-package.yml
```diff
@@ -63,12 +63,27 @@ jobs:
       GITHUB_PR_NUMBER: ${{ github.event.number }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: recursive
```
arporter, line 67: I'm confused as to why this is showing as added? Any ideas?

arporter (changes requested):
Thanks Sergi, you just need to put the GHA workflow back to using pypi and I have a question about why other changes in that file are showing.

.github/workflows/python-package.yml
sergisiso: Done

.github/workflows/python-package.yml
sergisiso: Before this PR the docs where build with the pip fparser and not the submodule. This was the only action not using the submodule which is wrong (and it will be more important then enabling the doctests)

To build with the submodule I have to this recursive checkout. And I thought it also made sense to respect the skip ci we use elsewhere.

arporter: Oh I see! I failed to drive github correctly and thought I was looking at the place where we already have this check - I hadn't realised it was a new addition to the doc-test job. That's all good then :-)

sergisiso:
@arporter ready for another look

arporter (commented):
All good now. I'll double-check the integration tests since we've updated the workflow files.

arporter (approved):
ITs were all green so I'll proceed to merge.

Issue #2088, opened 2023-03-29
doctest issue(s) - e.g. LFRicTypes created before config is loaded


hiker:
While trying to set up some new doctests, I had problems with ``make doctest`` in the developer guide now and again failing. Root cause seems to be the order in which doctest is running, which seems to be pseudo-random. If ``psyir.rst`` is processed first, we get the exception that LFRicType is used before a config is loaded:
```
$ make doctest
sphinx-build -b doctest -d _build/doctrees   . _build/doctest
Running Sphinx v6.1.3
loading pickled environment... checking bibtex cache... up to date
done
building [mo]: targets for 0 po files that are out of date
writing output... 
building [doctest]: targets for 16 source files that are out of date
updating environment: 0 added, 1 changed, 0 removed
reading sources... [100%] psy_data                                                                                                                                               
looking for now-outdated files... none found
pickling environment... done
checking consistency... /home/joerg/work/psyclone/doc/developer_guide/developers_gocean1p0_stencils.rst: WARNING: document isn't included in any toctree
done
running tests...

Document: psyir
---------------
**********************************************************************
File "../../src/psyclone/domain/lfric/lfric_types.py", line ?, in default
Failed example:
    num_dofs_class = LFRicTypes("NumberOfUniqueDofsDataSymbol")
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib/python3.10/doctest.py", line 1350, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest default[1]>", line 1, in <module>
        num_dofs_class = LFRicTypes("NumberOfUniqueDofsDataSymbol")
      File "/home/joerg/work/psyclone/src/psyclone/domain/lfric/lfric_types.py", line 95, in __new__
        LFRicTypes.init()
      File "/home/joerg/work/psyclone/src/psyclone/domain/lfric/lfric_types.py", line 122, in init
        LFRicTypes._create_precision_from_const_module()
      File "/home/joerg/work/psyclone/src/psyclone/domain/lfric/lfric_types.py", line 146, in _create_precision_from_const_module
        lfric_const = LFRicConstants()
      File "/home/joerg/work/psyclone/src/psyclone/domain/lfric/lfric_constants.py", line 65, in __init__
        raise InternalError("LFRicConstants is being created before the "
    psyclone.errors.InternalError: PSyclone internal error: LFRicConstants is being created before the config file is loaded

```
I tried to fix this by loading the config object in testsetup:
```
[...]

PR #3441, opened 2026-05-26
(closes #2936, #1220, #2088) Fix doctests and add CI tests

### LINKED ISSUE #2936: Documentation "make doctest" fails
```
Doctest summary
===============
 1149 tests
  384 failures in tests
    0 failures in setup code
    0 failures in cleanup code
build finished with problems.
```

sergisiso:


sergisiso:
@arporter @LonelyCat124 This PR fixes the doctests using both:
`cd docs; make doctest`
and
`pytest -n auto --doctest-modules src/psyclone/`

And enables both as checks in CI (We need both because there isn't a complete overlap between both sets and their working dir is different, so we need to be sure that paths have been normalised).

There are some doctests that didn't work, for example because they used non-existing example files. In this cases I converted the `>>>` to `... code-block ::` which is not tested, and we can bringing them back to doctests over time.

It is ready for review.

doc/developer_guide/module_manager.rst
```diff
@@ -134,15 +134,14 @@ which prints the filenames of all modules used in ``tl_testkern_mod``:
         mod_info = mod_manager.get_module_info(module_name)
         print("Module:", module_name, os.path.basename(mod_info.filename))
 
-.. testoutput::
 
+.. The snippet above fails a `.. testcode ::` block, but if it didn't we
```
arporter: Why does it fail? Is it because the file doesn't exist or is it because of the relative path given to the module manager?

src/psyclone/tests/utilities.py
```diff
@@ -600,7 +595,56 @@ def get_invoke(algfile: str,
...
+    filepath = os.path.join(get_base_path(api), algfile)
+    config = Config.get()
+    if api == "gocean-examples":
+        # gocean-examples in this utility is a shorthand for using the gocean
+        # API but from its examples directory as a basepath
+        api = "gocean"
+    config.api = api
+
+    _, info = parse(filepath, api=api)
+    psy = PSyFactory(api, distributed_memory=dist_mem).create(info)
```
arporter: This could be simplified a lot if you used `get_invoke`. In fact, I'm not sure that this new method brings much benefit over the latter?

src/psyclone/domain/gocean/transformations/gocean_extract_trans.py
```diff
@@ -46,24 +46,19 @@
 
 
 class GOceanExtractTrans(ExtractTrans):
-    ''' GOcean1.0 API application of ExtractTrans transformation \
+    ''' GOcean1.0 API application of ExtractTrans transformation
```
arporter: Just "GOcean" now please.

src/psyclone/domain/lfric/transformations/lfric_extract_trans.py
```diff
@@ -49,22 +49,6 @@ class LFRicExtractTrans(ExtractTrans):
     ''' LFRic API application of ExtractTrans transformation
     to extract code into a stand-alone program. For example:
 
-    >>> from psyclone.parse.algorithm import parse
```
arporter, line 52: How come you've removed this? Also, the paragraph above still ends with "For example:..."

src/psyclone/psyir/transformations/acc_kernels_trans.py
```diff
@@ -67,16 +67,16 @@ class ACCKernelsTrans(RegionTrans):
...
+    >>> psyir = FortranReader().psyir_from_file(filename)
     >>>
     >>> from psyclone.psyir.transformations import ACCKernelsTrans
     >>> ktrans = ACCKernelsTrans()
     >>>
     >>> schedule = psyir.children[0]
-    >>> # Uncomment the following line to see a text view of the schedule
-    >>> # print(schedule.view())
-    >>> kernels = schedule.children[9]
+    >>> kernels = schedule.children[26]
```
arporter: This could be a bit fragile. Could we make it a little better by walking over Loops?

src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
```diff
@@ -49,23 +49,8 @@
 class OMPParallelLoopTrans(OMPLoopTrans):
     ''' Adds an OpenMP PARALLEL DO directive to a loop.
 
-        For example:
```
arporter, line 56: It seems a shame to lose this. Is it possible to keep a version?

src/psyclone/tests/psyir/nodes/profile_node_test.py
```diff
@@ -116,7 +116,7 @@ def test_lower_to_lang_level_single_node():
     a single ProfileNode.
 
     '''
-    Profiler.set_options([Profiler.INVOKES], api="nemo")
+    Profiler.set_options([Profiler.INVOKES], api="gocean")
```
arporter: This implies `Profiler` was missing a check on the validity of the API string? Have you fixed that?

src/psyclone/tests/exceptions_test.py
```diff
@@ -112,8 +114,11 @@ def test_exception_repr():
 
     for psy_except in psy_excepts:
 
-        # Check if the exception inherits PSycloneError
-        assert issubclass(psy_except, PSycloneError)
+        # Check if the exception inherits PSycloneError (we don't use
+        # issubclass because the import_modules used in this test re-imports
+        # already loaded modules and create two super classes PSycloneError)
```
arporter: Does that mean we should be monkeypatching or something or are those changes local to this test?

arporter: (Well done on fixing this BTW - presumably this was the test error we see occasionally?)

src/psyclone/transformations.py
```diff
@@ -1038,21 +1011,6 @@ class LFRicAsyncHaloExchangeTrans(Transformation):
     '''Splits a synchronous halo exchange into a halo exchange start and
     halo exchange end. For example:
 
-    >>> from psyclone.parse.algorithm import parse
-    >>> from psyclone.psyGen import PSyFactory
-    >>> api = "lfric"
-    >>> ast, invokeInfo = parse("file.f90", api=api)
```
arporter, line 1058: Is it possible to recover this one? If not, please update the docstring above.

arporter (changes requested):
Good job Sergi, it's great to have this working and the PR wasn't quite as big as I feared :-)
Only significant quibble is with the new utility method - I'm not sure it buys us much over `get_invoke` and wonder whether the two can be merged?

doc/developer_guide/module_manager.rst
sergisiso: It failed because some of the module files where moved to `../external/lfric_infrastructure/` and this was not updated. Fixed now.

(Hopefully these are the things that enabling doc-tests back will avoid)

src/psyclone/domain/gocean/transformations/gocean_extract_trans.py
sergisiso: Done

src/psyclone/domain/lfric/transformations/lfric_extract_trans.py
sergisiso: The filename does not exist: `find . -name solver_alg.x90` returns nothing.

Removed "For example:"

src/psyclone/tests/psyir/nodes/profile_node_test.py
sergisiso: This didn't use the api value, and it feels it is not the right place to validate api strings. So I changed it to a boolean "is_psykal"

sergisiso: (btw, I initially changed that because I did a grep for remaining api="nemo", not because it failed)

src/psyclone/tests/utilities.py
sergisiso: I have now one calling the other, so the code is not duplicated.

One of the reasons that I prefered the new method is because all examples had the "invoke.schedule" before and this is not user-facing infrastructure. It's not what they have to do in the trans script and think this could be unecessarely confusing.

src/psyclone/psyir/transformations/acc_kernels_trans.py
sergisiso: Done.

Note that a lot of doctests like this one are currently not very interesting, they only show how to apply a transformation (which a python user should already know). What is more interesting is to show the before and after of the associated code. I suppose this was the idea behing the `print(schedule.view())` but these are too lang and hard to read and they where commented in the documentation anyway (so not checked nor shown in docs).

My idea is to make a second chosing snippets of fortran that demonstrate each transformation, but it became too much for this PR. So for now I just wanted to enable the automatic testing and then we should make them apply and showcase more relevant inout/output nodes/strings.

src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
sergisiso: As mention, I didn't find these very valuable as they were because:
1) the psyfactory/invoke should not be exposed to the user
2) the trans.apply is trivial and doesn't need a example imo
3) the input file does not exist, so it was not possible to see the behaviour (which is the interesting part).

But since you asked I redid this example as I think we should be doing them.

src/psyclone/transformations.py
sergisiso: Updated docstring.

src/psyclone/tests/exceptions_test.py
sergisiso: Yes, this was not isolated. But under further inspection I think we were overcomplicating a lot this test. I updated to something much simpler. Let me know what you think.

sergisiso:
@arporter This is ready for another look, I provided in the comments some more explanation about why I did certain additions/deletetions in the PR that were missing for context.

src/psyclone/tests/utilities.py
arporter: Fair point. Thanks for simplifying.

src/psyclone/psyir/transformations/acc_kernels_trans.py
arporter: That sounds good. For this PR though this file isn't showing as modified since my last review (but the OMP one is)?

src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
arporter: Thanks :-)

src/psyclone/tests/exceptions_test.py
arporter: I asked Claude "I want to write a test in Python that checks that all classes within my package (PSyclone on github) that inherit from Exception do so via a package-specific sub-class of Exception (namely PSycloneError). What's the best way to do this?" and it came up with:
```
import ast
import pathlib
import pytest

# Adjust this to point at the root of the PSyclone package source
PACKAGE_ROOT = pathlib.Path(__file__).parent.parent / "src" / "psyclone"

ALLOWED_BASE = "PSycloneError"

# Names that are acceptable bases — add aliases here if PSycloneError is
# imported under a different name anywhere
EXCEPTION_BASES = {"Exception", "BaseException"}


def is_exception_base(name: str) -> bool:
    """Return True if `name` looks like a raw exception base class."""
    return name in EXCEPTION_BASES or (
        name.endswith(("Error", "Exception")) and name != ALLOWED_BASE
    )


def get_base_name(node: ast.expr) -> str | None:
    """Extract a simple name or attribute name from a base class node."""
    if isinstance(node, ast.Name):
        return node.id
    if isinstance(node, ast.Attribute):
        return node.attr  # e.g. `builtins.Exception` → `Exception`
    return None


def find_bad_exception_classes(path: pathlib.Path) -> list[tuple[str, str, int]]:
    """
    Return a list of (file, class_name, line_no) for every class that
    inherits directly from a raw exception base instead of PSycloneError.
    """
    violations = []
    source = path.read_text(encoding="utf-8")
    try:
        tree = ast.parse(source, filename=str(path))
    except SyntaxError:
        return []  # skip unparseable files

    for node in ast.walk(tree):
        if not isinstance(node, ast.ClassDef):
            continue
        # Skip PSycloneError itself
        if node.name == ALLOWED_BASE:
            continue
        for base in node.bases:
            base_name = get_base_name(base)
            if base_name and is_exception_base(base_name):
                violations.append((str(path), node.name, node.lineno))

    return violations


def all_python_files():
    return list(PACKAGE_ROOT.rglob("*.py"))


@pytest.mark.parametrize("py_file", all_python_files())
def test_exceptions_use_psyclone_error(py_file):
    violations = find_bad_exception_classes(py_file)
    assert violations == [], (
        f"Classes inheriting directly from Exception/BaseException "
        f"instead of {ALLOWED_BASE}:\n"
        + "\n".join(f"  {f}:{line} — {cls}" for f, cls, line in violations)
    )```

File: src/psyclone/tests/exceptions_test.py [hunk 10]
```diff
@@ -32,108 +32,41 @@
...
-
-    for psy_except in psy_excepts:
-
-        # Check if the exception inherits PSycloneError (we don't use
-        # issubclass because the import_modules used in this test re-imports
-        # already loaded modules and create two super classes PSycloneError)
-        assert 'PSycloneError' in [str(cls.__name__) for cls in
-                                   psy_except.__bases__]
-
+    psy_exception_classes = [
```
arporter: The idea of the old test was to make sure no developer added their own Exception class that failed to extend `PSycloneError`. However, it wasn't nice and it did fail occasionally. I'm wondering whether we could use the `inspect` module instead? EDIT - I asked Claude and it recommended parsing the AST - see other comment.

arporter (changes requested):
Thanks @sergisiso, almost there now. There's just one comment that needs some more attention and we need to think a bit more about the troublesome test for PSycloneError.

src/psyclone/psyir/transformations/acc_kernels_trans.py
sergisiso: Ah done, and it was not too difficult to use the format I was mentioning above

src/psyclone/tests/exceptions_test.py
sergisiso: This would not work, ast only only analyses the code, it would be easy to break with inheritance and of course with eval tricks like we use in fparser

src/psyclone/tests/exceptions_test.py
sergisiso: I reintroduced something similar to the previous version but hopefully simpler, and so far I haven't seen side effects.

sergisiso:

src/psyclone/tests/exceptions_test.py
arporter: LLM 0 Sergi 1 (although we don't/shouldn't use eval tricks for error classes...)

arporter:
Unfortunately recent additions/alterations have broken some of the doctests. Please could you take a look @sergisiso?

sergisiso:
@arporter Fixed now, the doctest environment needs to be editable because it uses functions and files that are not part of the package.

arporter (approved):
Thanks Sergi - apologies for messing up the conflict resolution and missing the `-e`.

PR #3447, opened 2026-05-29
(closes #3250) 100% coverage

### LINKED ISSUE #3250: Complete Codecov coverage

schreiberx:
For 100% coverage.

Warning: This is AI generated.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,5 @@
+   25) PR #3447 for #3250. Add missing tests to achieve 100% code coverage.
+
    24) PR #3429 for #3412. Generalise hoist_arguments_to_temporaries in NEMO utils
    to all arguments with array operations.
 
--- a/src/psyclone/domain/lfric/lfric_invoke.py
+++ b/src/psyclone/domain/lfric/lfric_invoke.py
@@ -44,9 +44,6 @@ from typing import TYPE_CHECKING
 
 from psyclone.configuration import Config
 from psyclone.domain.lfric.lfric_builtins import LFRicBuiltIn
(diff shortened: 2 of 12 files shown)


sergisiso:
@schreiberx There is 1 line missing, can we cover that one?
image

It is an odd one because it is a pytest fixture and has a no_cover (which apparently doesn't work)
image

I will attempt fixing it chaning the nocover, because this is tested in the compilation integration test, but not on the CI one that Codecov uses.

sergisiso (approved):
@schreiberx Aha, the last one took a bit of trial and error but is finally 100%! This is approved for merging.

sergisiso:
@schreiberx One more thing, in the future, instead of commit messages like "intermediate commit" could you write a more informative message, like "Requested Claude Code to cover remaining code coverage lines"

schreiberx:
> @schreiberx One more thing, in the future, instead of commit messages like "intermediate commit" could you write a more informative message, like "Requested Claude Code to cover remaining code coverage lines"

I didn't expect Claude code to work through this on its first attempt and assumed that's an intermediate commit.

PR #3450, opened 2026-06-02
(closes #3262) Update lfric gpu script and collect aggregated transformation stats

### LINKED ISSUE #3262: Error in LFRic GPU Offloading Script
The example script in
https://github.com/stfc/PSyclone/blob/master/examples/lfric/scripts/gpu_offloading.py
contains:
```
        # Make setval_* compute redundantly to the level 1 halo if it
        # is in its own loop and only if it has an iteration space that is
        # *not* restricted to owned dofs.
        for loop in subroutine.loops():
            if loop.iteration_space == "dof":
                if len(loop.kernels()) == 1:
                    if loop.kernels()[0].name in ["setval_c", "setval_x"]:
                        rtrans.apply(loop, options={"depth": 1})
```
But modifying `setval_x` introduces changes to the KGO. Since this example might be used in 'real' LFRic, we should remove `setval_x` and only modify `setval_c`.

sergisiso:


sergisiso:
@arporter @LonelyCat124 This is ready for review, its a small modification to the gpu offloading script but also adds the aggregated stats that I have been reporting. The ITs now show:
image

Should we also add integration checks for the numbers, eg "if number of module-inlined successes is < 277: fail"?

LonelyCat124:
I think having checks that we don't make this do worse is probably a good idea, if we have a good reason to make things worse then we can manually make that change as/when it happens.

LonelyCat124:
> I think having checks that we don't make this do worse is probably a good idea, if we have a good reason to make things worse then we can manually make that change as/when it happens.

I thought about this more - having an option to see which expected files are missed would be helpful for future PRs that cause issues.

sergisiso:
@arporter @LonelyCat124 For now I added the suggested threshold checks. Listing which kernels diverged is harder because it requires keeping a full log of a working run, and this are not very stable: parallel runs, reflect any change, ... so for now I haven't done this second step but I maybe it can be a follow on PR.

LonelyCat124:
> @arporter @LonelyCat124 For now I added the suggested threshold checks. Listing which kernels diverged is harder because it requires keeping a full log of a working run, and this are not very stable: parallel runs, reflect any change, ... so for now I haven't done this second step but I maybe it can be a follow on PR.

@sergisiso Yeah thats reasonable - I think probably a script later (that isn't stored/run automatically) to let this be computed manually is good enough, but that doesn't need to be in this PR I think.

LonelyCat124 (changes requested):
@sergisiso This is good I think - the only thing I wonder is if we can have something explicit to state when we increase the specific ones we care about, e.g. in `check_above`
```
if [[$value -gt $2]]; then
   echo
   echo "Improvement: Number of $1 has increased to $value"
 fi
```
so when we want to review changes its spelled out without having to search the values so much. It is a bit unnecessary since the information is output anyway, but I'll leave it up to you - if you don't see the value I'll proceed to merge.

sergisiso:
@LonelyCat124 Good point, I took your suggestion. Ready for another look

LonelyCat124 (approved):
Looks good now, we can merge after the release.

Issue #3257, opened 2025-12-10
LFRic OMPParallelLoop with reproducible reductions


sergisiso:
Currently only separate OMPParallel and OMPLoop support reproducible reductions because we create a summation variable inside the parallel region, however, we can potentially get rid of the variable holding the thread number and add support for:
 `LFRicOMPParallelLoopTrans().apply(loop, {"reprod": True})`

See discussion in https://github.com/stfc/PSyclone/pull/3226#discussion_r2585504335

sergisiso:
The precedence of the config variable or the transformation option is also inconsistent,  see comment below _Orginally posted by @LonelyCat124 in https://github.com/stfc/PSyclone/pull/3452#discussion_r3586068654_ :

This depends on the route of the code, my understanding is at the moment if `reprod` is provided as an option or kwarg, then it overrides the Config, while the config provides the default behaviour. From what I can see this is how it is setup in `OmpLoopTrans`.

However, the LFRicOMPParallelLoopTrans has its own behaviour, where the config overrides anything else (which I kept the behaviour in this PR, just had to refactor it to work with the new changes):
```
        if (Config.get().reproducible_reductions
                and self.omp_directive == "do"):
```

This is echoed in LFricOMPLoopTrans where the options are set by the transformation itself:

```
        options = options.copy() if options else {}
        # Make sure the default is set:
        options["reprod"] = options.get("reprod",
                                        Config.get().reproducible_reductions)
```

and always uses options to call its superclass. I think since this PR hasn't modified this transformation I won't try to refactor it more, but I will add a TODO to #2668 on that Transformation since its a larger required change than removing the dict.

sergisiso:
I would suggest to remove the config variable and let the script decide what needs to be generated. In addition, now that we have script kwargs, this can be exposed to the build system that is who should be responsible to decide what level of optimisations are enabled.

Issue #3207, opened 2025-11-04
Update and explore differences in LFRic KGOs


sergisiso:
We need to keep LFRic integration tests up-to-date
- [x] Use the new git repos (or stfc forks)
- [x] Move patches to the forks
- [ ] Add a lfric_atm passthrough that goes through all files (cp do_nothing.py transmute/global.py)
- [ ] Update LFric extraction tests

Without exact match is difficult to guarantee that we haven't introduced small errors

Some ideas why they could be differences:
- [ ] It is compiler-dependent and the baseline is gcc
- [ ] It is number-of-ranks dependent but we just have one for all tests
- [ ] We may need the config REPRODUCIBLE_REDUCTIONS=true
- [ ] REPRODUCIBLE_REDUCTIONS=true fails with the colour-tiling script
- [ ] We may need the 'math_uniform' in gpu run

Issue #2648, opened 2024-07-05
Duplicated OMPParallelLoopTrans transformation


sergisiso:
We have at least one duplicated transformation with:
```
OMP_TRANS_PAR_DO = OMPParallelLoopTrans(omp_schedule="static")
```
```
OMP_TRANS_PAR_DO = OMPLoopTrans(omp_schedule="static", omp_directive="paralleldo")
```
But the first one does not accept the collapse option.

arporter:
In #3480 @LonelyCat124 has also pointed out that the way the `validate` of OMPParallelLoopTrans interacts with that of ParallelLoopTrans isn't great. The inheritance is:

<img width="768" height="75" alt="Image" src="https://github.com/user-attachments/assets/7b8e68b5-f74f-4ed4-b55e-17651293a217" />

`ParallelLoopTrans.validate` also sets some state in the transformation and it really shouldn't:
https://github.com/stfc/PSyclone/blob/5287120c4e96066f2ac597925ea18445154a9488/src/psyclone/psyir/transformations/parallel_loop_trans.py#L341-L352

I see it's 18 months since @sergisiso had a go at refactoring this. Can you remember what happened?

sergisiso:
I don't remember when/why I was doing this. Feel free to give it another go

PR #3452, opened 2026-06-09
(Closes #3409) reproducible reduction performance improvements

### LINKED ISSUE #3409: Improve LFRic reproducible reductions

LonelyCat124:
Assuming tests are all successful here, then ready for a first look. This does the changes we discussed in the issue, there are some rogue newline statements that seem to commen from adding preceding_comments, though I could also be tempted to remove the preceding comments if they seems a bit over the top.


LonelyCat124:
@sergisiso I think this is ready for a first look now.

sergisiso:
@LonelyCat124 Today LFRic meeting discussion has reminded me that we still don't test reproducible reductions in our integration tests and we should because this is what the MetOffice always use. Can you update:
`examples/lfric/scripts/KGOs/*.cfg`
to have
`REPRODUCIBLE_REDUCTIONS = true`
and then launch the integration tests on this branch?

sergisiso:
A second step would be to update the:
`python ${PSYCLONE_LFRIC_DIR}/compare_output.py ${PSYCLONE_LFRIC_DIR}/KGOs/gungho_model-checksums.txt gungho_model-checksums.txt`
to just:
`diff ${PSYCLONE_LFRIC_DIR}/KGOs/gungho_model-checksums.txt gungho_model-checksums.txt`
but I do not expect this to work because `${PSYCLONE_LFRIC_DIR}/KGOs/gungho_model-checksums.txt` is very likely generated with different number of ranks.

We could regenerate the KGO by doing manual run of
`./build/local_build.py -j ${NUM_PARALLEL} -w working-gh-passthrough gungho_model` (without the `-p`)
and then
`OMP_NUM_THREADS=1 mpirun -n 12 ../bin/gungho_model configuration.nml`

There is another issue for this https://github.com/stfc/PSyclone/issues/3207, so I am not sure if it should be here or a separate PR. But it will be great to be confident that this produced the exact expected results (and continues doing so going forward).

LonelyCat124:
The LFRic colour tiling IT fails - I'll investigate.

LonelyCat124:
@sergisiso So I setup running this and its an issue due to `OMPParallelDoDirective` doing a reproducible reduction. This is not supportable with the new way of creating them, since we have
```
!$omp parallel do
do i = 1, n
   ! some statement with a reproducible reduction
end do
!$omp end parallel do
```

This means we can't have `asum_temp_local = 0.0` after the `omp parallel`, nor `asum_local(1, th_idx) = asum_temp_local` after the `end do`

This gives 2 options as I see, first is to make the transformation do something different if `reprod` is True (i.e. use parallel then do directives separately) or disallow reprod on OmpParallelDoTransformation. Do you have any other ideas or preference in which I should do?

sergisiso:
The standard has this reproducible order-modifier https://www.openmp.org/spec-html/5.2/openmpse59.html, Is this something we can use / the compilers implement?

sergisiso:
Any other suggestions @markb-epcc ?

markb-epcc:
@sergisiso AFAIK, the reproducible order-modifier applies to the mapping of loop iterations to threads, not to the order of reduction operations. The spec says "The location in the OpenMP program at which values are combined and the order in which values are combined are unspecified. Thus, when comparing sequential and parallel executions, or when comparing one parallel execution to another (even if the number of threads used is the same), bitwise-identical results are not guaranteed." 

I would have thought that padding the temporary array would solve the problem, but that is not done in https://github.com/MetOffice/lfric_core/blob/bb10f897f7ecc4a931906b8d54bf1692f4a66b7f/components/science/source/psy/sci_psykal_light_mod.f90 where I noticed the false sharing problem coming from. 

If you want to implement this with !$omp parallel do then would you not need to split the parallel and do in any case, so that omp_get_thread_num() is only called once and not in every loop iteration?

LonelyCat124:
Yeah, I'm not sure what the best solution is here - essentially how should PSyclone do the splitting (its a mostly a bit of an inheritance nightmare).

I could simply skip the reproducible reductions being created if the node is a OMPParallelDoDirective, but weirdly that requires checking the type in OMPParallelDirective which is a mess.

The entire LFRicOMPParallelLoopTrans infrastructure currently never checks to see if reproducible reductions are enabled, it just (now? Maybe it did before and we never tested but the old version just worked) magically happens at lowering time, when it hits OMPParallelDirective.lower_to_language_level which performs the new reproducible reductions setup, and results in :
```
E           psyclone.errors.GenerationError: Generation Error: An OMPParallelDoDirective can only be applied to a single loop but this Node has 4 children: [<psyclone.psyir.nodes.assignment.Assignment object at 0x7bb197a97da0>, <psyclone.psyir.nodes.assignment.Assignment object at 0x7bb197a97350>, <psyclone.psyir.nodes.loop.Loop object at 0x7bb197a97590>, <psyclone.psyir.nodes.assignment.Assignment object at 0x7bb197b38050>]
```

This is a good sanity error checking test at least, but the only thing I can thing to do is when lowering the OMPParallelDoDirective, to do the following:
```
    reprod_red_call_list = self.reductions(reprod=True) # I don't love this argument name being reprod...
    if reprod_red_call_list:
       # Replace this node with a parallel and a do
       dir_body = self.dir_body.detach()
       dodir = OMPDoDirective(
   omp_schedule=self.omp_schedule,
   collapse = self.collapse,
    reprod=self.reprod,
    nowait=self.nowait,
    children=dir_body
      )
     pdir = OMPParallelDirective.create(dodir)
     self.replace_with(pdir)
     return pdir.lower_to_language_level()


        # Calling the super() explicitly to avoid confusion
        # with the multiple-inheritance
        OMPParallelDirective.lower_to_language_level(self)
        self.children[4].replace_with(OMPScheduleClause(self._omp_schedule))

        return self
```

I'm not sure how you feel about changing the tree like this during lowering @sergisiso ? This is a rough problem to solve in another way though - its possible I could do it at transformation application (or just in general, if someone requests LFRICParallelLoopTrans with reprod set to true in the config it applies the 2 transformations separately instead regardless of whether reductions exist), but I don't know if we know whether there are reductions required until after transformation time.

Let me know if you have a preference between the options.

LonelyCat124:
I decided to implement the transformation approach, and have a test in PSyclone now that works (we never tested that functionality anywhere before...). I'm hoping the ITs pass now but waiting on them for now.

LonelyCat124:
@sergisiso This is ready for a look now, ITs all green.

LonelyCat124:
So I can restrict it to
```
if node.walk(Kern) and node.walk(Kern)[0].is_reduction:
   # Do OMPDo and OMPParallel
```

But I have some questions from reading the Kern documentation which I suspect isn't true:

```
    @property
    def reprod_reduction(self):
        '''
        :returns: whether this kernel/built-in is enclosed within an OpenMP
                  do loop. If so report whether it has the reproducible flag
                  set. Note, this also catches OMPParallelDo Directives but
                  they have reprod set to False so it is OK.
        :rtype: bool

        '''
        ancestor = self.ancestor(OMPDoDirective)
        if ancestor:
            return ancestor.reprod
        return False
```

This claims that ancestor.reprod is always False for `OMPParallelDo Directives`, but I'm not sure that this is true.
Its only defined on `OMPDoDirective`, and the only place I can see this value being set is in `OMPDoDirective.__init__` where its set with `self._reprod = Config.get().reproducible_reductions` (or `= reprod` if its passed to the constructor, but OMPParallelLoopTrans doesn't pass that argument), so maybe we're not doing the currently expected behaviour already? We had no tests that failed with the changes here apart from the one I added.

arporter:
Ooh, that doesn't look like it's right anymore. Can we remove the `reprod_reduction` method altogether perhaps?

LonelyCat124:
> Ooh, that doesn't look like it's right anymore. Can we remove the `reprod_reduction` method altogether perhaps?

I'm not sure, its tricky, in theory we need to check the ancestor's `reprod` behaviour, as it could be set to False in which case we need to know its not reproducible in places for determining whether some symbols are needed so I don't think so - I just think that part of the docstring is wrong.

examples/lfric/scripts/KGOs/lfric_psyclone.cfg
```diff
@@ -42,7 +42,7 @@
 # Settings common to all APIs
 [DEFAULT]
 DISTRIBUTED_MEMORY = true
-REPRODUCIBLE_REDUCTIONS = false
+REPRODUCIBLE_REDUCTIONS = true
```
sergisiso, line 45: I requested this change but, do we need this configuration option? The transformation already has a `reprod=True` option that the LFRic scripts explictly use. So this should not have any effect?

Should we maybe just delete the configuration option to make things simpler @arporter ?

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
sergisiso, line 2181: This is the initialisation of the final sequencial loop, could we bring this assignment there, in between the:
```
! sum the partial results sequentially\n"
do th_idx = 1, nthreads, 1\n"
```
statements?

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -2182,7 +2182,10 @@ def test_reprod_red_after_normal_real_do(tmpdir, monkeypatch, annexed,
             "    local_asum = 0.0_r_def\n"
             "\n"
             "    ! Call kernels and communication routines\n"
-            "    !$omp parallel default(shared) private(df,th_idx)\n"
+            "    !$omp parallel default(shared) private(df,th_idx,"
+            "local_temp_asum)\n"
+            "    ! Initialise thread-private reduction variable\n"
+            "    local_temp_asum = 0.0_r_def\n"
```
sergisiso: Could we rename:
local_temp_{var} -> thread_priate_{var}
local_asum -> per_thread_{var} or array_of_partial_{var}

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
sergisiso: If you move it, could we have the ALLOCATE and the initialisation of local_{var} under the same comment

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -2193,9 +2196,12 @@ def test_reprod_red_after_normal_real_do(tmpdir, monkeypatch, annexed,
             "    !$omp do schedule(static)\n"
             "    do df = loop1_start, loop1_stop, 1\n"
             "      ! Built-in: sum_X (sum a real-valued field)\n"
-            "      local_asum(1,th_idx) = local_asum(1,th_idx) + f1_data(df)\n"
+            "      local_temp_asum = local_temp_asum + f1_data(df)\n"
             "    enddo\n"
-            "    !$omp end do\n"
+            "    !$omp end do\n\n"
+            "    ! Store the thread private value of the reduction into the "
+            "shared array as required for reproducible reductions.\n"
```
sergisiso: Can we remove "as required for reproducible reductions."

(I think its clear by the context and it avoid line-breaks of long lines)

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -2385,21 +2397,32 @@ def test_two_reprod_reductions_real_do(tmpdir, dist_mem):
...
-            "      local_bsum(1,th_idx) = local_bsum(1,th_idx) + f1_data(df)\n"
+            "      local_temp_bsum = local_temp_bsum + f1_data(df)\n"
             "    enddo\n"
-            "    !$omp end do\n"
+            "    !$omp end do\n\n"
+            "    ! Store the thread private value of the reduction into "
+            "the shared array as required for reproducible reductions.\n"
+            "    local_bsum(1,th_idx) = local_temp_bsum\n\n"
+            "    ! Store the thread private value of the reduction into "
+            "the shared array as required for reproducible reductions.\n"
```
sergisiso: Not a big issue, but if it is simple, can we avoid repeating comments?, e.g. if the previous statement already have the exact same comment, don't add it agian.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -6630,3 +6731,96 @@ def test_deprecated_names(capsys):
...
+                                     dist_mem):
+    '''Test that when using reproducible reductions with
+    LFRicOMParallelLoopTrans it switches to using LFRicLoopTrans and
+    OMPParallelLoopTrans instead to ensure reproducibility.
+    '''
+    file_name = "15.17.2_one_standard_builtin_one_reduction.f90"
+    api_config = Config.get().api_conf(TEST_API)
+    monkeypatch.setattr(api_config, "_compute_annexed_dofs", annexed)
+    # Enabled reproducible reductions
+    Config.get()._reproducible_reductions = True
```
sergisiso, line 6752: Was this needed if we already have `otrans.apply(child, {"reprod": True})` in the apply?

src/psyclone/psyGen.py
```diff
@@ -967,6 +967,69 @@ def initialise_reduction_variable(self) -> None:
             insert_loc.addchild(assign, cursor)
         return new_node
 
+    def temp_to_array_assignment(self,
+                                 parent: Node,
+                                 table: SymbolTable) -> None:
```
sergisiso: Maybe called: store_thread_private_value_to_shared_array

And can we make 'parent' more specific, e.g. `parallel_region_body: Schedule`

src/psyclone/psyGen.py
```diff
@@ -967,6 +967,69 @@ def initialise_reduction_variable(self) -> None:
...
+
+        assign.preceding_comment = (
+            "Store the thread private value of the reduction into the "
+            "shared array as required for reproducible reductions."
+        )
+
+        # Where to put it?
+        parent.addchild(assign)
+
+    def initialise_and_privatise_scalar_store(
```
sergisiso: Maybe named: create_thread_private_variable ?

src/psyclone/psyGen.py
```diff
@@ -967,6 +967,69 @@ def initialise_reduction_variable(self) -> None:
...
+        self, parallel_region: Node, position: int, table: SymbolTable
```
sergisiso: Maybe instead of:
parallel_region: Node
use
parallel_directive: OMPParallelDirective

(So that static typing know that .dir_body and .private_clause are valid methods)

Also note that the "param" is currently "parent".

examples/lfric/scripts/KGOs/lfric_psyclone.cfg
sergisiso: I commented this below but there is a new test with both:
`Config.get()._reproducible_reductions = True`
and ` option={reprod:True}`

are we sure the later works without the former?

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -6630,3 +6731,96 @@ def test_deprecated_names(capsys):
         assert (f"Deprecation warning: the script uses the legacy name "
                 f"'{old_name}', please use new name "
                 f"'{new_name}' instead.") in out
+
+
+def test_reprod_red_with_parallel_do(tmpdir, monkeypatch, annexed,
+                                     dist_mem):
+    '''Test that when using reproducible reductions with
+    LFRicOMParallelLoopTrans it switches to using LFRicLoopTrans and
+    OMPParallelLoopTrans instead to ensure reproducibility.
```
sergisiso, line 6746: I am ok with this behaviour, what does @arporter think?

(the alterenative would be to make the transformation not validate?)

@LonelyCat124 Is this affcted/tested with https://github.com/stfc/PSyclone/issues/2648

src/psyclone/psyGen.py
```diff
@@ -967,6 +967,69 @@ def initialise_reduction_variable(self) -> None:
...
+    def temp_to_array_assignment(self,
+                                 parent: Node,
+                                 table: SymbolTable) -> None:
+        '''
+        Generate the appropriate code to transfer the private reduction
+        temporary into the shared array for reproducible reductions.
+        The assignment is always placed as the final child of parent.
+
+        This method is designed to be used *after* a Kern has been lowered
+        (and thus detached) and therefore does not use `self.scope`.
```
sergisiso, line 954: Could we explicitly check this with `if directive_body.parent.scope is not None: InternalError("")` if this is what it means?

Also the note and the check should also be in the method below.

examples/lfric/scripts/KGOs/lfric_psyclone.cfg
sergisiso: btw this is just to know the current behaviour and you opinion, if we want to drop the config, it should be another issue as this one is already big enough.

sergisiso (changes requested):
@LonelyCat124 See inline comments. FYI, I started https://github.com/stfc/PSyclone/compare/master...3207_lfric_kgo_reprod to see if we can more accurately check the KGOs for changes like this, but the generated code looks ok to me, I mostly commented on aspects about the clarity of the method and output code.

examples/lfric/scripts/KGOs/lfric_psyclone.cfg
LonelyCat124: This depends on the route of the code, my understanding is at the moment if `reprod` is provided as an option or kwarg, then it overrides the Config, while the config provides the default behaviour. From what I can see this is how it is setup in `OmpLoopTrans`.

However, the LFRicOMPParallelLoopTrans has its own behaviour, where the config overrides anything else (which I kept the behaviour in this PR, just had to refactor it to work with the new changes):
```
        if (Config.get().reproducible_reductions
                and self.omp_directive == "do"):
```

This is echoed in LFricOMPLoopTrans where the options are set by the transformation itself:

```
        options = options.copy() if options else {}
        # Make sure the default is set:
        options["reprod"] = options.get("reprod",
```

and always uses options to call its superclass. I think since this PR hasn't modified this transformation I won't try to refactor it more, but I will add a TODO to #2668 on that Transformation since its a larger required change than removing the dict.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
LonelyCat124: Not very easily I think, the current implementation has these blocks all created at different places in the code, and I don't think they're all super easy to combine (some even happen during lowering). I feel like this should be a separate PR or issue (or repurpose the original one) to rewrite this generation of the tree.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
LonelyCat124: Done.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
LonelyCat124: Done.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
LonelyCat124: I thought about this, and I think its not so straightforward to do for > 2 (which doesn't happen in these tests) as you'd have to check some arbitrary number of previous nodes to check if any of them have the comment which seems a bit messy. Happy to do that if you feel worthwhile though.

Checking the previous alone is not enough since if you have 3 reductions then you'd end up with the first and 3rd both having the comment but the second would not.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
LonelyCat124: Yes - actually the option is unnecessary (and seemingly ignored) so I removed the option being provided. See the explanation in a previous comment for how this works for LFric.

src/psyclone/psyGen.py
LonelyCat124: Done

src/psyclone/psyGen.py
LonelyCat124: Done.

src/psyclone/psyGen.py
LonelyCat124: Done.

src/psyclone/psyGen.py
LonelyCat124: I'm not sure there is a benefit to the check - I think we'll get an error when looking up by tag if lowering hasn't happened anyway since 
```
            local_var = table.find_or_create_tag(
                root_name="array_of_partial_"+self._reduction_arg.name,
                tag=f"{self.name}:{self._reduction_arg.name}:local",
                symbol_type=DataSymbol, datatype=array_type)
```

won't have happened yet, as its done during lowering (and the only place we create something with the appropriate tag.

I could do a try/catch block around the whole function to raise an `InternalError` if we get a failure if you want a cleaner error, but the documentation is mostly designed to just emphasise an order. Let me know what your preferred option is.

LonelyCat124:
@sergisiso This is ready for another look - I have a few questions in the comments that need will likely need me to address some changes again.
I also reverted the addition of transformation documentation wrapper and kwargs from LFRicOMPParallelLoopTrans, as there are issues with that implementation from master with the changes. Once we are closer to merging and if you feel it belongs in this PR we can add it back later.

~~Edit: Since its used in the examples I will add kwargs back into LFricOMPParallelLoopTrans after lunch as I don't want to revert changed scripts and things.~~ done

examples/lfric/scripts/KGOs/lfric_psyclone.cfg
sergisiso: Uh that's bad (but out of scope for this PR). Could you make the TODO to https://github.com/stfc/PSyclone/issues/3257 isntead, I think this is more specific than the options refactoring.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
sergisiso: I would re-introduce the option and instead add a TODO to https://github.com/stfc/PSyclone/issues/3257 saying that the option is currently ignored, but that issue will resolve the inconsitency.

src/psyclone/psyGen.py
```diff
@@ -943,6 +943,72 @@ def initialise_reduction_variable(self) -> None:
...
+        assign = Assignment.create(
+            arr_ref, Reference(local_var)
+        )
+
+        assign.preceding_comment = (
+            "Store the thread private value of the reduction into the "
+            "shared array."
+        )
+
+        # Where to put it?
```
sergisiso: Remove this comment?

src/psyclone/psyGen.py
```diff
@@ -943,6 +943,72 @@ def initialise_reduction_variable(self) -> None:
...
+        parallel_region_body.addchild(assign)
+
+    def create_thread_private_variable(
+        self, parallel_directive: OMPParallelDirective,
+        position: int, table: SymbolTable
+    ) -> None:
+        '''
+        Generate the appropriate code to initialise the scalar reduction
+        variable inside the parallel region, and add it to the private clause.
```
sergisiso, line 987: I am ok not adding the runtime check I mentioned in my previous review, but can you also add the:
```
This method is designed to be used *after* a Kern has been lowered
(and thus detached) and therefore does not use `self.scope`.
```
note here. Otherwise I keep wondering why the table is provided separately.

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
```diff
@@ -2175,14 +2175,17 @@ def test_reprod_red_after_normal_real_do(tmpdir, monkeypatch, annexed,
                     in result)
         assert "loop1_stop = f1_proxy%vspace%get_last_dof_owned()" in result
         expected_output = (
-            "    ALLOCATE(local_asum(8,nthreads))\n"
+            "    ALLOCATE(array_of_partial_asum(8,nthreads))\n"
```
sergisiso: Can we remove the extra-dimension with the 8 now?

- The access below are always: `array_of_partial_asum(1,`
- I assume it was introduced to achieve the same (avoid false sharing), but now we have a better solution.

sergisiso (changes requested):
@LonelyCat124 I resolved the remaining suggestions about the generated output, but see my response to the TODO comments above as I think we can be more specific.

The last bit remaining is whether the padding we had is still necessary or can we remove it now?

src/psyclone/psyGen.py
LonelyCat124: Done

src/psyclone/psyGen.py
LonelyCat124: Added

src/psyclone/tests/domain/lfric/transformations/lfric_transformations_test.py
LonelyCat124: Done, also removed all references of the pad variable from configs and tests, examples, docs (I think) and tutorials.

LonelyCat124:
@sergisiso Ready for another look.

sergisiso (approved):
@LonelyCat124 Thanks for removing the pad_size, this part looks simpler now (although the reproducible reductions implementation is still quite complex, but this is captured by other TODOs). I removed a few remaining mentions to the pad_size that I found. This is approved for merging.

PR #3458, opened 2026-06-15
(Closes #2721) Add `ArrayConstructor` to PSyIR

### LINKED ISSUE #2721: Fortran array syntax not recognized
```python
@pytest.mark.xfail
def test_psyclone_special_array_notation():
    code = """subroutine foo(a, n)
  integer :: n
  real*8, dimension(5) :: a

  a = (/2, 3, 4, 1/)

end subroutine foo\n"""
    reader = FortranReader()
    psyir_tree: Node = reader.psyir_from_source(code)
    assert len(psyir_tree.walk(CodeBlock)) == 0, psyir_tree.walk(CodeBlock)[
        0
    ].debug_string()
    writer = FortranWriter()
    output = writer(psyir_tree)
    assert code == output
```

mn416:
I noticed that `reference_accesses()` is very conservative on array constructors, e.g. `[x, y]` leads to both `x` and `y` being reported as read/write. This is because array constructors are treated as `CodeBlock`s and not converted to PSyIR.

This PR adds initial support for array constructors to PSyIR via a new `ArrayConstructor` node class. Currently, it only supports the most simple form of array constructor

  * `[<expr>, <expr>, ..., <expr>]`

not

  * `[<type-spec> :: <expr>, <expr>, ..., <expr>]` (type spec), or
  * `[(<expr>, <name> = <expr>, <expr>)]` (implied do)

which are still translated to `CodeBlock`, as before. I'm happy to add support for type specs and/or implied do, but I was wondering if there is any hunger for this PR before spending any more time on it.

I believe this PR addresses #2721.


arporter:
Thanks @mn416, I think this will be useful - we hit some limitations to do with this sometimes. Does your implementation support the old `(/x, y, z/)` syntax too?

mn416:
Hi @arporter, yes the old and new brackets are both supported. They are parsed to almost identical abstract syntax by `fparser`.

LonelyCat124:
I know we do hit some implied do in some codes, but I'm not sure we have a reason to need them to not be codeblocks unless we have issues with them so I think that can be skipped for now?

mn416:
I added support for type specs in the [mn416-array-ctr-type-spec](https://github.com/stfc/PSyclone/tree/mn416-array-ctr-type-spec) branch but I'm slightly less confident in it. I'm therefore inclined to agree with @LonelyCat124 that we may benefit from an incremental approach here, i.e. handle basic array constructors in this PR and add support for type specs and implied do later. An incremental approach should also simplify reviewing (several simple PRs versus one big one).

arporter:
Absolutely, I'm all in favour of simple PRs 😀

LonelyCat124:
Happy to review this then.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4975,6 +4977,43 @@ def _assignment_handler(self, node, parent):
 
         return assignment
 
+    def _array_constructor_handler(self, node, parent):
```
LonelyCat124: Can you add typehints and returntype here to the declaration rather than in the docstring?

src/psyclone/psyir/backend/fortran.py
```diff
@@ -1439,6 +1439,20 @@ def literal_node(self, node):
 
         return result
 
+    def arrayconstructor_node(self, node):
```
LonelyCat124: Use type hint for node and return type on the function instead of in the doc please.

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2017-2026, Science and Technology Facilities Council.
```
LonelyCat124: Just copyright 2026 please.

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+    _children_valid_format = "[DataNode]*"
+    _text_name = "ArrayConstructor"
+    _colour = "yellow"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    @staticmethod
+    def create(*elems: DataNode):
+        '''Create and ArrayConstructor instance representing an array
```
LonelyCat124: and should be an

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+        with the given elements.
+
+        :param elems: the elements of the array being constructed.
+        :returns: an ArrayConstructor instance.
+        :rtype: :py:class:`psyclone.psyir.nodes.IfBlock`
```
LonelyCat124: Use return type inline on the function as well - it may need to be `IfBlock` and import annotations or similar - if you have issues let me know and I'll find the usual solution.

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+    # Textual description of the node.
+    _children_valid_format = "[DataNode]*"
+    _text_name = "ArrayConstructor"
+    _colour = "yellow"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    @staticmethod
+    def create(*elems: DataNode):
```
LonelyCat124: Is there a reason for `*elems` vs `elems: list[DataNode]` here?
Also is ArrayConstructor.create used anywhere in the code? I don't think its required to have this function unless it provides significant benefit over just `ArrayConstructor(children=elems)` (approximately)

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+        '''
+        array_cons = ArrayConstructor()
+        for elem in elems:
+            array_cons.children.append(elem)
+        return array_cons
+
+    @staticmethod
+    def _validate_child(position, child):
+        '''
+        :param int position: the position to be validated.
```
LonelyCat124: For all of these in this file can you switch to inlining the argument/return types instead of defining them in the docstring.

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @property
+    def datatype(self):
+        '''
+        :returns: the type of this array constructor.
+        :rtype: :py:class:`psyclone.psyir.symbols.DataType`
+        '''
+        # The result of an array constructor is always a rank-1 array.
+        # We look through the children to find the array-element type.
+        elem_type = UnresolvedType()
```
LonelyCat124, line 94: I have a question here - are there any Fortarn rules w.r.t type promotion here? E.g. is `arr = (\ 2.0_4, 3.0_8\)` valid fortran, and if so what is the resultant type of this array?

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+            if isinstance(child.datatype, ArrayType):
+                elem_type = child.datatype.elemental_type
+                break
+            elif isinstance(child.datatype, ScalarType):
+                elem_type = child.datatype
+                break
+            elif isinstance(child.datatype, DataTypeSymbol):
+                elem_type = child.datatype
+                break
+        return ArrayType(elem_type, [ArrayType.Extent.ATTRIBUTE])
```
LonelyCat124, line 113: Can we do better for the size of the array? Is it just `len(self.children)` for these cases or is it not so straightforward>

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+        :rtype: str
+        '''
+        return f"{self.coloured_name(colour)}[]"
+
+    def get_all_accessed_symbols(self) -> set[Symbol]:
+        '''
+        :returns: a set of all the symbols accessed inside this
+            ArrayConstructor.
+        '''
+        return super().get_all_accessed_symbols()
```
LonelyCat124: We don't need this if it just calls super.

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def reference_accesses(self) -> VariablesAccessMap:
+        '''
+        :returns: a map of all the symbol accessed inside this node, the
+            keys are Signatures (unique identifiers to a symbol and its
+            structure accessors) and the values are AccessSequence
+            (a sequence of AccessTypes).
+        '''
+        return super().reference_accesses()
```
LonelyCat124: We don't need this if it just calls super.

src/psyclone/psyir/nodes/array_constructor.py
```diff
@@ -0,0 +1,153 @@
...
+        have matching names. If there is no match for a given Symbol then it
+        is left unchanged.
+
+        :param table_or_symbol: the symbol table from which to get replacement
+            symbols or a single, replacement Symbol.
+        :type table_or_symbol: :py:class:`psyclone.psyir.symbols.SymbolTable` |
+            :py:class:`psyclone.psyir.symbols.Symbol`
+
+        '''
+        super().replace_symbols_using(table_or_symbol)
```
LonelyCat124: We don't need this if it just calls super.

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
```diff
@@ -245,6 +245,17 @@ def validate(
             raise TransformationError(LazyString(
                 lambda: f"{message}, but found:\n{node.debug_string()}"))
 
+        # Do not allow to transform expressions with ArrayConstructors
+        # TODO: this could be relaxed in future
```
LonelyCat124: If we have a TODO we need a corresponding issue number. @sergisiso @arporter Is this worth opening an issue now or do we just drop the TODO and just have "PSyclone doesn't currenlt allow transformation of expressions with ArrayConstructors`" as the comment without the TODO.

src/psyclone/tests/psyir/frontend/fparser2_array_constructor_handler_test.py
```diff
@@ -0,0 +1,213 @@
+# -----------------------------------------------------------------------------
+# BSD 3-Clause License
+#
+# Copyright (c) 2019-2026, Science and Technology Facilities Council.
```
LonelyCat124: Copyright 2026

LonelyCat124 (changes requested):
@mn416 Have left some initial comments, mostly codestyle related - I didn't go through the tests yet, I'll do that on the next pass through.

src/psyclone/psyir/frontend/fparser2.py
mn416: Done (1ccb589).

src/psyclone/psyir/backend/fortran.py
mn416: Done (1ccb589).

src/psyclone/psyir/nodes/array_constructor.py
mn416: Done (5d8a9d0).

src/psyclone/psyir/nodes/array_constructor.py
mn416: Done (5d8a9d0)

src/psyclone/psyir/nodes/array_constructor.py
mn416: Done (5d8a9d0). The `IfBlock` was a copy/paste mistake and should of course be `ArrayConstructor`. Thanks for spotting this. I resolved the self-reference issue using the `from __future__ import annotations` solution that seems to be used elsewhere in the PSyclone code base.

src/psyclone/psyir/nodes/array_constructor.py
mn416: I've moved from `*elems` to `elems: list[DataNode]` in 5d8a9d0. I agree this is better as it allows different arguments should they be needed in future (e.g. when type specs are supported).

Regarding `ArrayConstructor.create()`, I've kept it for now.  I use `ArrayConstructor()` in `frontend/fparser2.py` to create an empty node, which is then populated by the parser. Then in the tests, I use `ArrayConstructor.create()` to make a populated node. This seems to be the pattern in other PSyIR nodes. However, I can remove it if you prefer. Just let me know.

src/psyclone/psyir/nodes/array_constructor.py
mn416: Done (5d8a9d0).

src/psyclone/psyir/nodes/array_constructor.py
mn416: That's a type error. My understanding is that PSyclone generally assumes that the program is well-typed, so we don't need to catch the error ourselves. Is that right?

src/psyclone/psyir/nodes/array_constructor.py
mn416: Removed (5d8a9d0).

src/psyclone/psyir/nodes/array_constructor.py
mn416: Removed (5d8a9d0).

src/psyclone/psyir/nodes/array_constructor.py
mn416: Removed (5d8a9d0). We will need to add these back if/when we add support for type-specs and implied do, but that's fine.

src/psyclone/tests/psyir/frontend/fparser2_array_constructor_handler_test.py
mn416: Done (5d8a9d0).

src/psyclone/psyir/transformations/arrayassignment2loops_trans.py
mn416: I've removed the TODO for now and left the comment. Happy to add it back with an issue number if desired.

src/psyclone/psyir/nodes/array_constructor.py
mn416: In general, with implied do, the size is not known until runtime. However, for constructors without implied-do, you're right, we can do better. The size is not quite `len(self.children)` which would be wrong for `[1,2,[3,4]]`. But I've added a recursive function to count it properly in commit d50bd60. Thanks for this suggestion.

mn416: Hmm, actually I just realised it's more complicated than that. For an array constructor such as `[x]`, we have to look at the type of `x` as it may be an array (of any rank). With this complexity, I'm not sure it's worth it. I've reverted the new commit d50bd60. Let me know if you want me to explore this further.

mn416:
Thanks @LonelyCat124. I've addressed all your comments for now and I think it's ready for another pass.

src/psyclone/psyir/nodes/array_constructor.py
LonelyCat124: Yeah, I think we're happy with bad input leading to bad output so happy to not catch that in PSyclone

src/psyclone/psyir/nodes/array_constructor.py
LonelyCat124: Yeah happy to leave this as is - can you add a comment explaining why we don't give the exact size and the challenges in case someone wants to change it in future.

src/psyclone/psyir/nodes/array_constructor.py
LonelyCat124: Yeah I'm fine with leaving it in, thanks for the change.

LonelyCat124 (changes requested):
Hi @mn416 - almost there now - one additional comment I'd liked to be added re: the unknown array size. I'll set any ITs running in the meantime.

mn416:
Thanks @LonelyCat124. I've added the comment.

LonelyCat124 (approved):
ITs all green, I'll proceed to merge.

Issue #3463, opened 2026-06-23
Improving MaximalOMPParallelRegionTrans for LFRic


LonelyCat124:
More info here when my head hurts less - main goal is to correctly handle assignments in such a way that they can be included.

LonelyCat124:
First PR ready for review for this - reversing the order of operations of the maximal region transformations.

Ok, so to update this, allowing assignments requires:
1. Scalar-only left-hand-side result. Probably also local-only variable, since otherwise we have to magically make assumptions about `call`s that will make this too hard (e.g. `omp_get_num_threads` is not pure so will cause issues with our assumptions for non-local calls).
2. No reads past the end of the current parallel region (or at least the first access after the parallel region must be a write always).
3. Must be written to first, i.e. no assignments like `a = a + X` but they must just be of the form `a = X` (where X is some statement that doesn't involve a).

LonelyCat124:
I think one other change that will be required is enabling `_node_allowed` to have more arguments, at the moment it assumes whether a node is allowed is independent of the current region, however this is no longer the case so it will need to take `current_block` as an additional argument (that is ignored in the default implementation). This also means `_can_be_in_region` will need this as well, and it will need to be passed in the base class.

LonelyCat124:
I've implemented an initial attempt, but I'm waiting on #3431 probably before I can actually do any real testing with it.

arporter:
@LonelyCat124 I have (or am about to) merge #3481 but I wasn't sure whether or not this issue should be closed?

LonelyCat124:
I'm not sure, I think we need to see some new results from LFRic with the new changes once they've been tested before we decide.

PR #3464, opened 2026-06-23
(Towards #3463) Reverse the order of operations to build the region in reverse

LonelyCat124:
Simple PR to reverse the order in which maximal parallel region trans build their regions. I expect to need this to work out if assignments can be in parallel regions and its probably a more sensible default way to build these regions.

Ready for a review from @arporter or @sergisiso . I considered using a `deque`, however I would have ended up needing to cast the `deque` to a `list` enough times I didn't think it would save anything.

### DIFF
--- a/src/psyclone/psyir/transformations/maximal_region_trans.py
+++ b/src/psyclone/psyir/transformations/maximal_region_trans.py
@@ -164,22 +164,22 @@ class MaximalRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
         # Find the largest sections we can surround with the transformation.
         all_blocks = []
         current_block = []
-        for child in node_list:
+        for child in node_list[::-1]:
             # If the child can be added to a transformed region then add it
             # to the current block of nodes.
             if self._can_be_in_region(child):
                 # Check that validation still succeeds if we add this child
                 # to the current block.
                 try:
-                    trans.validate(current_block + [child], **trans_kwargs)
-                    current_block.append(child)
+                    trans.validate([child] + current_block, **trans_kwargs)
+                    current_block.insert(0, child)
                 except TransformationError:
                     # If validation now fails, then don't add this to the
                     # current block and add the block to the allowed blocks
                     # if allowed.
                     if current_block:
                         if self._satisfies_minimum_region_rules(current_block):
-                            all_blocks.append(current_block)
+                            all_blocks.insert(0, current_block)
                         current_block = []
             else:
                 # Otherwise, if the current_block contains any children,
@@ -187,29 +187,29 @@ class MaximalRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
                 # the current_block.
                 if current_block:
                     if self._satisfies_minimum_region_rules(current_block):
-                        all_blocks.append(current_block)
+                        all_blocks.insert(0, current_block)
                     current_block = []
                 # Need to recurse on some node types
                 if isinstance(child, IfBlock):
-                    if_blocks = self._compute_transformable_sections(
-                            child.if_body, trans, trans_kwargs
-                    )
-                    all_blocks.extend(if_blocks)
                     if child.else_body:
                         else_blocks = self._compute_transformable_sections(
                             child.else_body, trans, trans_kwargs
                         )
-                        all_blocks.extend(else_blocks)
+                        all_blocks = else_blocks + all_blocks
+                    if_blocks = self._compute_transformable_sections(
+                            child.if_body, trans, trans_kwargs
+                    )
+                    all_blocks = if_blocks + all_blocks
                 if isinstance(child, (Loop, WhileLoop)):
                     loop_blocks = self._compute_transformable_sections(
                         child.loop_body, trans, trans_kwargs
                     )
-                    all_blocks.extend(loop_blocks)
+                    all_blocks = loop_blocks + all_blocks
         # If any nodes are left in the current block at the end of the
         # node_list, then add them to a transformed region
         if current_block:
             if self._satisfies_minimum_region_rules(current_block):
-                all_blocks.append(current_block)
+                all_blocks.insert(0, current_block)
 
         return all_blocks
 
--- a/src/psyclone/psyir/transformations/maximal_region_trans.py
+++ b/src/psyclone/psyir/transformations/maximal_region_trans.py
@@ -149,7 +149,7 @@ class MaximalRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
     def _compute_transformable_sections(
             self, node_list: Union[list[Node], Schedule],
             trans: Transformation,
-            trans_kwargs: dict[str, Any]
+            trans_kwargs: dict[str, Any],
     ) -> list[list[Node]]:
         '''
         Computes the sections of the input node_list to apply the
@@ -165,6 +165,22 @@ class MaximalRegionTrans(RegionTrans, metaclass=abc.ABCMeta):
         all_blocks = []
         current_block = []
         n_list = self.get_node_list(node_list)
+        # Initially test if the entire region can be transformed without
+        # breaking it up into pieces.
+        for child in n_list:
+            if not self._can_be_in_region(child):
+                break
+        else:
+            try:
+                # Test validation for the whole region, and if successful
+                # then we can finish.
+                trans.validate(n_list)
+                all_blocks.insert(n_list)
+                return all_blocks
+            except TransformationError:
+                pass
+        # Otherwise, loop over the region backwards and create the largest
+        # blocks possible.
         for child in reversed(n_list):
             # If the child can be added to a transformed region then add it
             # to the current block of nodes.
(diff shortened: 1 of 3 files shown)


src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -164,52 +164,52 @@ def _compute_transformable_sections(
         # Find the largest sections we can surround with the transformation.
         all_blocks = []
         current_block = []
-        for child in node_list:
+        for child in node_list[::-1]:
```
sergisiso: Maybe `reversed(node_list)` ?

sergisiso (changes requested):
@LonelyCat124 There is a small comment but I am ok with this change and it is already tested by all the maximal_region_trans_test examples. So this can be merged after the release.

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Yeah this is probably more efficient than making a new list since we don't modify, done.

LonelyCat124:
@sergisiso This is ready for another look - i couldn't just do reversed without checking I had a list and not a schedule, so I fixed that.

sergisiso (dismissed):
All the ITs came green, this is approved for merging @LonelyCat124

sergisiso:
@LonelyCat124 I approved the PR but before merging, I am seeing some differences to the NEMO profile, it goes from:
```
    CALL profile_psy_data % PreStart("stringop", "cmpblank-r0", 0, 0)
    lcc = LEN_TRIM(str)
    ipb = 1
    do while (.true.)
      if (ipb >= lcc) then
        ! PSyclone CodeBlock (unsupported code) reason:
        !  - Unsupported statement: Exit_Stmt
        EXIT
      end if
      if (str(ipb:ipb + 1) == '  ') then
        str(ipb + 1:) = str(ipb + 2:lcc)
        lcc = lcc - 1
      else
        ipb = ipb + 1
      end if
    end do
    CALL profile_psy_data % PostEnd
```
to
```
    CALL profile_psy_data % PreStart("stringop", "cmpblank-r0", 0, 0)
    lcc = LEN_TRIM(str)
    ipb = 1
    CALL profile_psy_data % PostEnd
    do while (.true.)
      if (ipb >= lcc) then
        ! PSyclone CodeBlock (unsupported code) reason:
        !  - Unsupported statement: Exit_Stmt
        EXIT
      end if
      if (str(ipb:ipb + 1) == '  ') then
        str(ipb + 1:) = str(ipb + 2:lcc)
        lcc = lcc - 1
      else
        ipb = ipb + 1
      end if
    end do
```
which is not quite right as the region should encapsulate the whole subroutine for this case. Can we maintain the previous behaviour?

(This happens in 10 files, all with similar patterns)

LonelyCat124:
@sergisiso I'll take a look later this week. Do you have an example file name (or are they ones in the PSyclone examples repo?) so I can investigate? I will remove your approval in the meantime so this can't be merged by accident.

sergisiso:
I tested it with the NEMOv5 BENCH from the integration tests. The files with differences are:
```
diff tests/BENCH_test_prof/BLD/tmp/bdylib.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/bdylib.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/diadct.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/diadct.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/fliocom.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/fliocom.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/getincom.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/getincom.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/obs_inter_h2d.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/obs_inter_h2d.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/obs_prep.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/obs_prep.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/sbcblk_skin_coare.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/sbcblk_skin_coare.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/stpctl.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/stpctl.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/stringop.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/stringop.psycloned.f90
diff tests/BENCH_test_prof/BLD/tmp/trcbdy.psycloned.f90 tests/BENCH_test_prof2/BLD/tmp/trcbdy.psycloned.f90
```
The example above is from `stringop.f90`

LonelyCat124:
The transformation is failing here:
```
                try:
                    trans.validate([child] + current_block, **trans_kwargs)
                    current_block.insert(0, child)
                except TransformationError as err:
                    # If validation now fails, then don't add this to the
                    # current block and add the block to the allowed blocks
                    # if allowed.
                    print(child.debug_string())
                    print(err)
                    if current_block:
                        if self._satisfies_minimum_region_rules(current_block):
                            all_blocks.insert(0, current_block)
                        current_block = []
```

This fails with the transformation error:
```
Transformation Error: Cannot apply the ProfileTrans to a code region containing a potential control flow jump, as these could skip the end of profiling caliper. Found:
'! PSyclone CodeBlock (unsupported code) reason:
!  - Unsupported statement: Exit_Stmt
EXIT
'
```

This is due to a conflict in ProfileTrans:
```
        if (isinstance(parent, Routine) and
                len(parent.children) == len(node_list)):
           return
```
This transformation has a special case where if you provide an entire routine it will skip over control flow jumps because they have to stay in the same scope (I think?). IF the function you provided instead was:

```
  lcc = LEN_TRIM(str)
  ipb = 1
  DO
    IF (ipb >= lcc)   EXIT
    IF (str(ipb:ipb+1) == '  ') THEN
      str(ipb+1:) = str(ipb+2:lcc)
      lcc = lcc-1
    ELSE
      ipb = ipb+1
    ENDIF
  ENDDO
  ipb = 1
```

Then we would see a similar issue and get the result you see here even processing forwards, since it would hit the "EXIT" before the full routine was passed into `validate` of ProfileTrans.

I'm not sure what a good solution would be. I could have _compute_transformable_sections have an option for forward or backward depending on the use-case, but we'll still see similar limitations with any slightly changed code for the maximal profile region transformation either way.

sergisiso:
Aha, that's true! In fact the previous version of the generated code was wrong as it could potentially leave profiling hooks unclosed?

arporter:
> Aha, that's true! In fact the previous version of the generated code was wrong as it could potentially leave profiling hooks unclosed?

It shouldn't - hopefully the code distinguishes between an early `return` and any loop control flow? In fact, I think it has `Return` in `excluded_nodes`? Of course, as @LonelyCat124 says, this only works if the transformation has been provided with the entire body of a routine. Otherwise any control-flow should be rejected.

LonelyCat124:
No, it only does this for codeblocks that meet this criteria:
```
            for child in walk(node, (Fortran2003.Goto_Stmt,
                                     Fortran2003.Exit_Stmt,
                                     Fortran2003.Cycle_Stmt,
                                     Fortran2003.StmtBase)):
```
and those control flow jumps always stay within the confines of the routine, so its safe to include them iff the entire routine is in the profile transformation.

In theory the PRofileTrans transformation could be smarter, if you have a Loop and the control flow jump is either an Exit_Stmt or Cycle_Stmt then the loop could also be contained (since those statements are limited in their ability to jump as well), which might be the best option to fix this?

sergisiso:
> any loop control flow.

You are right, it always confuses me that "EXIT" is a "loop break" in Fortran, not a exit the application. It is a self-contained jump inside the codeblock.

LonelyCat124:
@sergisiso Ready for another look with the change discussed this morning.

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -165,6 +165,22 @@ def _compute_transformable_sections(
...
+        # Initially test if the entire region can be transformed without
+        # breaking it up into pieces.
+        for child in n_list:
+            if not self._can_be_in_region(child):
+                break
+        else:
+            try:
+                # Test validation for the whole region, and if successful
+                # then we can finish.
+                trans.validate(n_list)
```
sergisiso, line 177: Almost, there is some cases where the whole region does not comply with the "_satisfies_minimum_region_rules", when the region should not be added at all.

sergisiso (changes requested):
@LonelyCat124 There are still changes from previous NEMO, looking at them some are correct, but some are due to missing some rules (see inline) when validating the whole region.

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done. If the full region doesn't satisfy the minimum region rules I return the empty list with the assumption that no sub region will satisfy the rules. I added a test of that functionality as well.

LonelyCat124:
@sergisiso Ready for another look. I removed the modifications you made to `changelog` due to conflicts and I felt I'd leave it to you to rewrite as appropriate (as more merges may happen before this).

sergisiso (approved):
@LonelyCat124 Doing a visual inspection of the remaining profiling differences all the cases now seem to be valid improvements. There is a small capitalisation change needed that I will do. This is approved for merging.

PR #3465, opened 2026-06-23
2757 script arguments

### LINKED ISSUE #3461: Import scripts with the same name more than once (and sys.path growing)
While debugging some odd behaviour in one of my tests, I realised that while running tests `sys.path` keeps on growing - each time a script is executed, a path is added, but never removed (not to mention we also add empty paths ... over and over :)  ).

I am wondering if it's worth avoiding this (i.e. keeping a copy of `sys.path` before calling `load_script`, then resetting it after execution of the script).

I also looked at removing the script symbol (`del trans_func` after calling it).

None of this is expected to have a real issue for an application, since they will call PSyclone only once with one script (even Fab uses subprocesses, i.e. you always get a clean `sys.path`).

The growing `sys.path` could in theory affect parallel testing (e.g. if we have one test expecting NOT to find a script of a given name, but accidentally a previous test adds a path that has a script with that name) - which is what I though was happening in my case:

I had three tests, each providing a script with the same name but slightly different behaviour. Each individual test worked, but when I executed them all (in parallel or not), some of them would fail. Reason of course is that the first executed script imported the script, and all further imports are fulfilled from the (Python internal) cached copy - no matter if we do a `del trans_func` (which removes the symbol, but NOT the cached import), and/or clean up `sys.path`. Fix for me was easy, just use three different names.

So, summary, we have thing with testing that we have to be careful about (different tests using the same script name), but can't fix it (there seems to be no way to affect the imported cache, unless we try using `reimport`??). 

The clean up (sys.path and removing the symbol) feels nice, but doesn't do anything. TBH, it feels like we should (at least) fix sys.path ;) But that's just me being ... peculiar about this :)  I am happy if we agree to just close this issue, but equally happy to clean up sys.path (and/or the symbol, but that makes probably even less sense ;) ). I will try to add a sentence or two to the docs as part of my script args PR otherwise.

hiker:
This is ready for a first review. It adds `--script-kwargs` command line option to pass command line arguments to scripts. It adds documentation, modifies one example and one tutorial lesson to use this feature.

Note that I originally intended to fix #3461, it turns out the fix to avoid modifying `sys.path` breaks tests in a non-obvious way, so have reverted this change.


hiker:
Ready for review. IT passed, the failure in linkinspector were caused by UKMO links not working, I've verified some of them, they are fine now.

doc/user_guide/user_scripts.rst
```diff
@@ -89,6 +89,26 @@ and/or kernels) contained within the provided tree.
 The :ref:`examples section<examples>` provides a list of psyclone user scripts
 and associated usage instructions for multiple applications.
 
+.. _script_kwargs:
+
+Arguments for Scripts
+---------------------
+Scripts can take optional keyword arguments specified on the command line
+using the option `--script-kwargs`. The keyword arguments are specified
+as separate `keyword:value` pairs, separate by `,`. For example:
```
arporter: Perhaps "in a string containing comma-delimited `keyword:value` pairs. For example..."

doc/user_guide/user_scripts.rst
```diff
@@ -89,6 +89,26 @@ and/or kernels) contained within the provided tree.
...
+
+.. code-block:: shell
+
+    psyclone -s ./optimise.py input_source.f90 \
+             --scripts-kwargs "omp: True, tiling: [4,4]"
```
arporter: "script-kwargs"

doc/user_guide/user_scripts.rst
```diff
@@ -89,6 +89,26 @@ and/or kernels) contained within the provided tree.
...
+
+This will result in the additional keyword arguments for any transformation call:
```
arporter, line 106: How are the specified arguments converted into Python? Presumably there are some restrictions/limitations? Also, what happens if an argument is specified that the `trans/trans_alg` routine doesn't accept? Worth mentioning this as I can imagine it being a common error.

examples/nemo/eg8/Makefile
```diff
@@ -60,7 +60,8 @@ PSYCLONE_PROFILING_LIB ?= ${PSYCLONE_PROFILING_DIR}/libsimple_timing.a
 PSYCLONE_PROFILING_LIBS ?= -L${PSYCLONE_PROFILING_DIR} -lsimple_timing
 
 transform:
-	ENABLE_PROFILING=1 ${PSYCLONE} -s ./omp_gpu_profile_trans.py ../code/tra_adv.F90 -o traadv_instrumented.F90
+	${PSYCLONE} -s ./omp_gpu_profile_trans.py --script-kwargs "profiling: True" \
```
arporter, line 63: Nicer :-)

examples/nemo/eg8/omp_gpu_profile_trans.py
```diff
@@ -94,8 +90,16 @@ def add_omp_region_profiling_markers(children: Union[List[Node], Schedule]):
             add_omp_region_profiling_markers(child.children)
 
 
-def trans(psyir):
-    """Apply OpenMP offloading and insert profiling around target regions."""
+def trans(psyir: FileContainer, profiling=False):
```
arporter: typehint for `profiling` please :-)

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+
+
+@pytest.mark.parametrize("kwargs, out", [("", "{}"),
+                                         ("'a':1", "{'a': 1}"),
+                                         ("'b': {1: 2}", "{'b': {1: 2}}"),
+                                         ("'l': [1,2]", "{'l': [1, 2]}"),
+                                         ("a:1", "{'a': 1}"),
+                                         ("b: {1: 2}", "{'b': {1: 2}}"),
+                                         ("l: [1,2]", "{'l': [1, 2]}")
+
```
arporter: Pls rm the blank line.

src/psyclone/generator.py
```diff
@@ -285,8 +303,10 @@ def generate(filename: str,
             .create(invoke_info)
         if script_name is not None:
             # Apply provided recipe to PSyIR
-            recipe, _, _ = load_script(script_name)
-            recipe(psy.container.root)
+            trans_func, _, _, kwargs = load_script(script_name, kwargs_str)
+            # trans_func is always defined, otherwise an exception is raised
+            assert trans_func
```
arporter: No bare assert please. You could explicitly pass `is_optional=False` to `load_script` if that helps and also extend the comment before the call to say an exception is raised if the transformation function isn't defined.

src/psyclone/generator.py
```diff
@@ -426,8 +448,10 @@ def generate(filename: str,
 
         if script_name is not None:
             # Call the optimisation script for psy-layer optimisations
-            recipe, _, _ = load_script(script_name)
-            recipe(psy.container.root)
+            trans_func, _, _, kwargs = load_script(script_name, kwargs_str)
+            # recipe is always defined, otherwise an exception is raised
```
arporter: Again, no bare assert please - you could just move the comment to before the call to `load_script`?

src/psyclone/generator.py
```diff
@@ -880,23 +911,26 @@ def add_builtins_use(fp2_tree, name):
...
+                             script_name: str,
+                             output_file: str,
+                             keep_comments: bool,
+                             keep_directives: bool,
                              keep_conditional_openmp_statements: bool,
-                             free_form: bool = True, line_length="off"):
-    ''' Process the input_file with the recipe_file instructions and
-    store it in the output_file.
+                             kwargs_str: Optional[str] = None,
+                             free_form: bool = True,
```
arporter: Thanks for updating the type hints. These two are also `Optional`.

src/psyclone/generator.py
arporter: Please document the `kwargs_str` argument too.

src/psyclone/utils.py
```diff
@@ -234,3 +236,82 @@ def wrapper(cls):
...
+
+# ----------------------------------------------------------------------------
+def parse_kwargs(kwargs: str) -> dict[str, Any]:
+    """
+    This function safely parses a user string provided on the command line
+    using '--kwargs ...` into a python dictionary. It especially simplifies
+    the syntax for the user by not requiring the keys to be escaped, e.g.
+    --kwargs "'a':1,'b':2" and --kwargs "a:1,b:2" will both work as expected.
+
+    This is done by using Python'kwargs ast parser, then adding a separate
```
arporter: "...Python's AST parser, then..."

I guess this answers my earlier question (in the User Guide): the argument values in the string have to be valid Python expressions?

src/psyclone/utils.py
```diff
@@ -234,3 +236,82 @@ def wrapper(cls):
...
+    :raises ValueError: if the string cannot be converted to a kwargs-style
+        dictionary.
+    """
+
+    # Parse as an expression. Note that various ast functions can
+    # raise different exceptions, so we catch all exceptions and
+    # re-raise them as a ValueError
+    try:
+        # Make it look like a dict literal
+        wrapped = "{" + kwargs.strip().rstrip(",") + "}"
```
arporter: Do you need the `rstrip(",")`? I can do:
```python
>>> mine = {"andy": True,}
>>> mine
{'andy': True}
```

arporter: It may be Python-version dependent?

src/psyclone/utils.py
```diff
@@ -234,3 +236,82 @@ def wrapper(cls):
...
+        # the modified tree, since the newly created nodes won't have
+        # this information
+        ast.fix_missing_locations(expr)
+
+        # Safely evaluate literals/containers
+        result = ast.literal_eval(expr)
+    # pylint: disable=broad-exception-caught
+    except Exception:
+        # This will trigger an exception in the next statement
+        result = None
```
arporter: It would be better to do `raise ValueError(...) from err` here so that we retain full information about what happened. You could then refine the error message in the second `raise` at L284, e.g. "the keyword arguments were parsed as a xxx instead of a dict"

src/psyclone/utils.py
```diff
@@ -234,3 +236,82 @@ def wrapper(cls):
...
+
+
+class NameKeysToStr(ast.NodeTransformer):
+    """
+    This is a helper class to convert dictionary keys that are
+    an ast.Name (i.e. not a string) into an ast.Constant (a string).
+
+    It will effectively change `{a:1}` to `{'a':1}`
+
+    :param node: the dictionary node.
```
arporter: There's no `__init__` so you don't need this ":param node..." line.

src/psyclone/utils.py
```diff
@@ -234,3 +236,82 @@ def wrapper(cls):
...
+        # This will trigger an exception in the next statement
+        result = None
+
+    if not isinstance(result, dict):
+        raise ValueError(f"Invalid syntax for keyword arguments '{kwargs}' ")
+
+    return result
+
+
+class NameKeysToStr(ast.NodeTransformer):
```
arporter: It might be best to make this class "private" as it's not expected to be used outside of this file?

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
         main([filename, "--config-opts", "DOES_NOT_EXIST=27"])
     assert ("Attempt to overwrite unknown configuration option: "
             "'DOES_NOT_EXIST=27'" in str(err.value))
+
+
+@pytest.mark.parametrize("kwargs, out", [("", "{}"),
```
arporter: These tests are mainly checking the parsing of the argument string? As such, they should be in `utils_test.py`. We could keep one simple test that the arguments are passed all the way through.

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+                                         ("'b': {1: 2}", "{'b': {1: 2}}"),
+                                         ("'l': [1,2]", "{'l': [1, 2]}"),
+                                         ("a:1", "{'a': 1}"),
+                                         ("b: {1: 2}", "{'b': {1: 2}}"),
+                                         ("l: [1,2]", "{'l': [1, 2]}")
+
+                                         ])
+def test_script_arguments_transmute(kwargs, out, tmp_path, capsys):
+    """Tests that script arguments are received as expected with the
+    transmute approach. This test creates a dummy script that prints the
```
arporter: Personally, I don't like "transmute" (although I appreciate this is what it is called in LFRic). Do you mean when transforming generic code?

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+                                         ("'a':1", "{'a': 1}"),
+                                         ("'b': {1: 2}", "{'b': {1: 2}}"),
+                                         ("'l': [1,2]", "{'l': [1, 2]}"),
+                                         ("a:1", "{'a': 1}"),
+                                         ("b: {1: 2}", "{'b': {1: 2}}"),
+                                         ("l: [1,2]", "{'l': [1, 2]}")
+                                         ])
+def test_script_arguments_lfric_testing(kwargs, out, tmp_path, capsys,
+                                        monkeypatch):
+    """Tests that script arguments are received as expected using the
```
arporter, line 2070: Is this because we currently have two ways of doing LFRic? If so, please could you add a TODO for the issue number (possibly #1618)?

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+    LFRic API. This test creates a dummy script that prints the arguments
+    for trans and trans_alg, which we check for. This uses LFRIC_TESTING,
+    which will also call trans_alg (which by LFRic otherwise would not do).
```
arporter: "..by default LFRIC..."?

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+    inputfile = Path(get_base_path("nemo")) / "afunction.f90"
+    outputfile = tmp_path / "output.f90"
+    main([str(inputfile), "-s", str(script_path),
+          "--script-kwargs", kwargs,
+          "-o", str(outputfile)])
+    stdout, _ = capsys.readouterr()
+    assert f"ARGS: {out}" in stdout
+
+
+@pytest.mark.parametrize("kwargs, out", [("", "{}"),
```
arporter: Again, I think all that is needed for this test is just one example that arguments are passed through as expected.

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+    main([str(inputfile), "-s", str(script_path), "--psykal-dsl", "lfric",
+          "--script-kwargs", kwargs,
+          "-opsy", str(psy_file),
+          "-oalg", str(alg_file)])
+    stdout, _ = capsys.readouterr()
+    assert f"trans args: {out}" in stdout
+    assert f"trans_alg args: {out}" in stdout
+
+
+@pytest.mark.parametrize("kwargs, out", [("", "{}"),
```
arporter: Again, just one example will do here I think.

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+                                         ("'a':1", "{'a': 1}"),
+                                         ("'b': {1: 2}", "{'b': {1: 2}}"),
+                                         ("'l': [1,2]", "{'l': [1, 2]}"),
+                                         ("a:1", "{'a': 1}"),
+                                         ("b: {1: 2}", "{'b': {1: 2}}"),
+                                         ("l: [1,2]", "{'l': [1, 2]}")
+                                         ])
+def test_script_arguments_lfric_default(kwargs, out, tmp_path, capsys):
+    """Tests that script arguments are received as expected using the
```
arporter, line 2104: Another TODO for #1618 (if that's correct)?

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+    inputfile = Path(get_base_path("lfric")) / "1_single_invoke.f90"
+    psy_file = tmp_path / "psy.f90"
+    alg_file = tmp_path / "alg.f90"
+    main([str(inputfile), "-s", str(script_path), "--psykal-dsl", "lfric",
+          "--script-kwargs", kwargs,
+          "-opsy", str(psy_file),
+          "-oalg", str(alg_file)])
+    stdout, _ = capsys.readouterr()
+    assert f"trans args: {out}" in stdout
+    # Default LFRic API does not call trans_alg!!! This line is here to
```
arporter, line 2129: #1618

src/psyclone/tests/generator_test.py
```diff
@@ -2110,3 +2110,147 @@ def test_config_overwrite() -> None:
...
+    # fail once we switch LFRic over.
+    assert f"trans_alg args: {out}" not in stdout
+
+
+@pytest.mark.parametrize("kwargs", ["1", "'a'", "[1,2]", "{1:2}",
```
arporter, line 2135: This test also belongs in utils_test I think? Perhaps just one example here to show the error gets propagated as expected?

src/psyclone/tests/utils_test.py
```diff
@@ -472,3 +472,38 @@ def apply(self, node, opt1: bool = False, opt2=None,
     :type opt3: int
     :param int opt3: (Option provided for SubTrans2) opt3 docstring."""
     assert correct in BaseTrans.apply.__doc__
+
+
+@pytest.mark.parametrize("kwargs, expected",
```
arporter, line 477: Great.

src/psyclone/tests/utils_test.py
```diff
@@ -472,3 +472,38 @@ def apply(self, node, opt1: bool = False, opt2=None,
...
+    """
+    result = parse_kwargs(kwargs)
+    assert result == expected
+
+
+@pytest.mark.parametrize("kwargs", ["1", "'a'", "[1,2]", "{1:2}",
+                                    "a=1", 123])
+def test_parse_kwargs_errors(kwargs):
+    """
+    Test that the parsing function for user-specific script options
```
arporter, line 502: "...raises the expected errors for malformed..."

src/psyclone/generator.py
```diff
@@ -470,8 +493,11 @@ def main(arguments):
         help='display version information')
     parser.add_argument('-c', '--config', help='config file with '
                         'PSyclone specific options')
-    parser.add_argument('-s', '--script', help='filename of a PSyclone'
-                        ' optimisation recipe')
+    scripts = parser.add_argument_group("Transformation scripts")
+    scripts.add_argument('-s', '--script',
+                         help='filename of a PSyclone optimisation recipe')
+    scripts.add_argument('--script-kwargs', help='Keyword arguments for the '
```
arporter, line 490: Seeing this reminds me that there are part(s?) of the UG which reproduce the help messages of the command line. These will need updating too.

tutorial/training/transformation/3.8-code-creation/solution/add_if.py
arporter, line 55: Please document the `parse_string` argument.

tutorial/training/transformation/3.8-code-creation/solution/add_if.py
```diff
@@ -71,7 +69,7 @@ def trans(psyir: FileContainer) -> None:
     # 1. Parsing a Fortran expression given as string
     # 2. Creating the tree representation using the PSyIR API
 
-    if os.environ.get("PARSE_STRING", False):
+    if parse_string:
         # Option 1: Create expression by parsing a Fortran string:
```
arporter, line 76: Not really this PR but this seems an odd thing to be doing in a transformation script?

tutorial/training/transformation/3.8-code-creation/add_if.py
```diff
@@ -50,7 +48,7 @@
                                   Literal, Loop, Routine, Schedule)
 
 
-def trans(psyir: FileContainer) -> None:
+def trans(psyir: FileContainer, parse_string: False) -> None:
     '''
     Create two versions of a loop, depending in iteration count.
```
arporter, line 55: Pls document the new argument.

arporter (changes requested):
Nice work Joerg, I like this new feature. The docs need a bit more work to explain the restrictions and what happens if things go wrong.
Apart from that, the only significant issue I have is that much of the parameterisation of the tests in `generator_test` is actually testing features that are implemented in `utils.py` (and, to be fair, are also tested in `utils_test.py`).
I'll run the ITs next time.

doc/user_guide/user_scripts.rst
hiker: Updated.

src/psyclone/tests/generator_test.py
hiker: Added.

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/tests/utils_test.py
hiker: Done.

src/psyclone/generator.py
hiker: All updated.

src/psyclone/utils.py
hiker: Yes :)

tutorial/training/transformation/3.8-code-creation/solution/add_if.py
hiker: Done.

tutorial/training/transformation/3.8-code-creation/solution/add_if.py
hiker: Fair enough. For training it just means trainees can pick which approach they prefer, and try both in the same setup, so no need to get familiar with a new environment.

tutorial/training/transformation/3.8-code-creation/add_if.py
hiker: Done.

doc/user_guide/user_scripts.rst
hiker: Done.

doc/user_guide/user_scripts.rst
hiker: Oops - we should look at doctesting this, not sure how to do this for shell scripts :(
But fixed.

examples/nemo/eg8/omp_gpu_profile_trans.py
hiker: Done

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/generator.py
hiker: That `assert` was for `mypy`, since `mypy` otherwise assumes that `trans_func` could be `None`. I've removed it, since atm we don't do `mypy` (sorry, problem with working on different code bases with slightly different coding styles). And I've updated/moved the comment.

src/psyclone/generator.py
hiker: Ditto - was for mypy.

src/psyclone/generator.py
hiker: Fixed.

src/psyclone/generator.py
hiker: Done.

src/psyclone/utils.py
hiker: No, you get different errors depending on which part of the parsing triggers it (ie. could be ValueError, SyntaxError, ...). I can't remember the details, but I could get at least two different errors (and I didn't find documentation listing all possible errors).

hiker: And indeed, the rstrip is not necessary. I've added a test for this elsewhere.

src/psyclone/utils.py
hiker: Done, and I also refactored the test two explicitly tests both kind of errors.

src/psyclone/utils.py
hiker: Oops.

src/psyclone/utils.py
hiker: Done.

src/psyclone/tests/generator_test.py
hiker: I've updated all `generator` tests to have only one kind of argument, the rest is indeed tested in `utils_test.py`.

src/psyclone/tests/generator_test.py
hiker: I hate that term, too - even though I believe it is technically correct. I've renamed it to `transform` :)

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/tests/generator_test.py
hiker: Done.

hiker:
I've triggered the IT. Extraction test failed while installing dependencies (gateway timeout), but compilation tests have already passed. The rest is running atm.

doc/user_guide/user_scripts.rst
```diff
@@ -109,6 +110,15 @@ This will result in the additional keyword arguments for any transformation call
     def trans(psyir, omp: bool, tiling: list[int]):
         # Modify psyir tree
 
+The arguments specified will be converted into a Python-style dictionary, e.g.
+``omp: True, tiling: [4,4]`` will be parsed as ``{omp: True, tiling: [4,4]}``.
+Therefore, the specified keyword arguments must be make a valid dictionary.
```
arporter: s/be//

doc/user_guide/user_scripts.rst
```diff
@@ -109,6 +110,15 @@ This will result in the additional keyword arguments for any transformation call
...
+For example, ``[1,2]`` is not a valid keyword argument, because ``{[1,2]}``
+is not valid Python (a list is not hashable), and ``1:2`` would not be valid,
+because the keys of keyword arguments must be strings.
+
+Obviously, the called transformation functions (``trans``, but also
+``trans_alg``, see below) must accept the specified keyword arguments.
```
arporter: Worth saying what happens if the user makes a mistake. This might make the error (and therefore solution) more discoverable.

src/psyclone/generator.py
arporter: Thanks :-)

src/psyclone/tests/generator_test.py
```diff
@@ -2222,35 +2121,42 @@ def trans_alg(psyir, **kwargs):
...
-    :param kwargs: the input string for the command line
-    """
-    recipe = '''
-def trans(psyir, **kwargs):
-    print("YARGS:", kwargs)
-    '''
-    script_path = tmp_path / "print_args_errors.py"
-    script_path.write_text(recipe)
+def test_script_arguments_errors(kwargs):
+    """Tests that script arguments errors are handled as correctly. We need
```
arporter: s/as//

src/psyclone/generator.py
```diff
@@ -675,6 +666,15 @@ def main(arguments: list[str]) -> None:
                   "specify the output destination of each psykal layer.")
             sys.exit(1)
 
+    if args.script_kwargs and not args.script:
+        msg = ("The '--script-kwargs' argument is only valid if a "
+               "script is specified using the '--script' option.")
+        logger.error(msg)
+        # By default the logger is off, so make sure the user actually sees
+        # this message by printing it as well:
+        print(msg)
```
arporter: Should print this to stderr? Not a biggie but it seems a bit odd to log it and print it. Perhaps just print it?

src/psyclone/utils.py
```diff
@@ -234,15 +242,14 @@ def wrapper(cls):
...
         return wrapper(*args)
-    else:
-        return wrapper
+    return wrapper
 
 
 # ----------------------------------------------------------------------------
 def parse_kwargs(kwargs: str) -> dict[str, Any]:
     """
     This function safely parses a user string provided on the command line
```
arporter: Reading this again, I realise this method doesn't need to mention how the user string is obtained - it just takes a string and converts it into a dict.

src/psyclone/utils.py
arporter: but you still have "Python'kwargs" :-)

arporter (changes requested):
Great stuff Joerg, just a little bit more tidying and then this can go on.
The ITs were all fine bar the LFRic test which failed because of a known issue on the test machine (soon to be fixed on master).

doc/user_guide/user_scripts.rst
hiker: Done.

doc/user_guide/user_scripts.rst
hiker: Added.

src/psyclone/tests/generator_test.py
hiker: Done.

src/psyclone/generator.py
hiker: Yes, it probably should. I looked, errors are handled somewhat inconsistently (though most or all? calls to a logger involve an exception?). Some print to stdout, some log, some do both :)  I thought logging might be useful in case that a caller registers their own handler, then again, by default logger doesn't emit messages.
So, KISS: print to stderr it is.

src/psyclone/utils.py
hiker: I've reworded it.

hiker:
IT is busy, I'll trigger it tomorrow.

arporter:
All looks good now, thanks Joerg. I'll re-run the ITs just to be on the safe side.

arporter (approved):
ITs were all green. This is good to go.

PR #3467, opened 2026-06-24
Update compilation and nemo 4 tests to nvfortran 26.5

nmnobre:


sergisiso (approved):
@nmnobre Looks good, this is approved for merging

Issue #1629, opened 2022-02-21
Use algorithm PSyIR when generating PSy-layer code


rupertford:
At the moment the PSy-layer uses algorithm-layer information gathered by the `parse` function in `algorithm/parse.py`. This should be replaced with use of the algorithm PSyIR. This will probably be part of a larger restructuring of psy-layer generation as psy-layer generation also moves to using PSyIR.

PR #3468, opened 2026-06-25
(Closes #3461) duplicated script names

### LINKED ISSUE #3461: Import scripts with the same name more than once (and sys.path growing)

hiker:
Fixes #3461.


hiker:
I've fixed the unit test failures by slightly changing the test (instead of setting sys.path, I `chdir` into the script directory).

We do get additional failure in example, since without modifying sys.path, helper scripts in the same directory as the psyclone scripts are NOT found anymore (unless you set `PYTHONPATH=.`). We could still modify sys.path in generator.py to accommodate that use case, I have started the CI to see if nemo/lfric relies on this.

hiker:
Failures in one example and nemo (lfric is fine) indicate that we should modify sys.path when importing and executing a script (so that helper scripts in the same directory as the script are found). That was done in ab86dd3, so once I've updated the documentation this should be ready to go.

hiker:
IT passed, ready for review.

doc/user_guide/user_scripts.rst
```diff
@@ -44,23 +44,22 @@ PSyclone User Scripts
...
-Alternatively, script files may be specified with a path. In this case
-the file must exist in the specified location. This location is then added to
-the Python search path ``PYTHONPATH`` as before. For example::
-
     > psyclone -s ./optimise.py input_source.f90
     > psyclone -s ../scripts/optimise.py input_source.f90
     > psyclone -s /home/me/PSyclone/scripts/optimise.py input_source.f90
 
+PSyclone will not take ``PYTHONPATH`` into account when importing the script,
+so the user must ensure that the script is found: For this, it must either be
```
arporter: "...found: for this,..."

doc/user_guide/user_scripts.rst
```diff
@@ -44,23 +44,22 @@ PSyclone User Scripts
...
+in the current working directory (if no path is specified at all), or be
+specified with a valid (relative or absolute) path. PSyclone will add the
+directory of the script to its Python's ``sys.path``, meaning the PSyclone
```
arporter: s/its//

arporter: and perhaps s/PSyclone/transformation/

src/psyclone/tests/generator_test.py
```diff
@@ -79,37 +78,18 @@
 GOCEAN_BASE_PATH = Path(get_base_path("gocean"))
 
 
-@pytest.fixture(name="script_factory", scope="function")
-def create_script_factor(tmpdir):
-    ''' Fixture that creates a psyclone optimisation script given the string
+def script_factory(tmpdir, code: str) -> Path:
```
arporter: Since we're here, upgrade to `tmp_path`?

arporter: and throughout this file?

src/psyclone/generator.py
```diff
@@ -163,12 +164,18 @@ def load_script(
...
-    # already exist elsewhere in the system path
+
+    # Add the script directory to sys.path, so scripts can easily import
+    # helper scripts in the same directory (this step is not needed to
+    # import the script itself, but it maintains backwards compatibility).
     sys.path.insert(0, filepath)
-    recipe_module = importlib.import_module(module_name)
+
+    # This will import the module, but not make it part of the
+    # system list of all modules, i.e. it can be used elsewhere.
```
arporter: The last bit after the "i.e" - are you trying to say that a module of the same name can then safely be used elsewhere or something else?

arporter: Might also be worth giving some context here about why we don't modify the system module path. (As I'm sure we'll forget in future.)

src/psyclone/generator.py
```diff
@@ -298,6 +305,7 @@ def generate(filename: str,
             # defined, otherwise an exception is raised.
             trans_func, _, _, kwargs = load_script(script_name, kwargs_str)
             trans_func(psy.container.root, **kwargs)
+            del sys.path[0]
```
arporter, line 288: What if algorithm optimisations are also being performed? i.e. the `load_script` call at L358. Not your fault at all but this whole routine is large and not very well structured. I'm a bit nervous about the multiple places you've had to add this `del`. However, any re-structuring might be best waiting for #1629 which I think @sergisiso is working on (when he has time).

src/psyclone/generator.py
arporter: This docstring needs updating. Also, you'll see from my comments below that I'm uneasy about having to manually clean-up after the call to this method. Although this was already a "feature" of the implementation, would it be possible to make this method define a context so we could do something like `with load_script(...) as blah:` and then we could automatically clean-up once done?

arporter (changes requested):
Thanks very much @hiker, this feels like a significant improvement. While reviewing it, I got very uncomfortable that our `load_script` routine (already) modifies global state and that this then has to be manually undone in multiple places after the script(s) have been used. I'm wondering whether there's a better way to do this that will prevent us making mistakes in the clean-up.

hiker:
@arporter , I agree with the refactoring (including using a `with`). I was originally thinking of doing this, but that would have made this change (which changing the rather essential feature of importing the PSyclone script ;) ) much larger, and potentially more difficult to debug if something should break (e.g. in LFRic).

But if #1629 is being worked on, that would create huge conflicts, since this refactoring will need to introduce additional functions, i.e. moving code around quite a bit to avoid even deeper nesting. 

I see three options:
1. I fix everything else you mentioned, and open a new ticket to refactor this properly. Once #1629 is done, I'll fully implement this. Ideal if #1629 is close to ready and has already refactored code in generator.py.
2. I think I can add a `with` statement without much refactoring, but nesting levels might become deep (and the code even harder to understand). This might have only a small impact on #1629 - the deep nesting is mostly in the `code_transformation_mode` function, so this should not affect LFRic work at all.
3. I do it properly, and @sergisiso needs to sort it out 👿  Not ideal, that would be quite messy and I don't like it. But if not a lot of work in #1629 has been done on `generator.py` (but elsewhere so far), that might(!) work.

Can you let me know what you prefer?

src/psyclone/generator.py
hiker: Not sure that I understand the comment about L308 and L358, as far as I can see they are in different branches and should not affect each other. If alg and kernel transformations are done, `load_script` is called twice (L358 and L454).

If I am missing something, could you please clarify? If it's just confusing because of the deeply nested and long if statements ... then we agree that it needs some refactoring :)

sergisiso:
@hiker My PR won't be ready anytime soon, I found some blockers, but once completed it will simplify the generate.py.

Andy and me discussed this and we decided you can keep the current implementation with the "del" statements, but add a new issue/TODOs, as we want to come back to clean this up at some point.

src/psyclone/generator.py
hiker: I've clarified the sentence (I hope), and added context.

src/psyclone/generator.py
hiker: I believe this all works as expected (though I agree with the del being bad :) ). Alg. and kernel transformation calls both call load_script (and add the cleanup after).

src/psyclone/generator.py
hiker: Done, and I've created #3514, and corresponding #TODOs in the code

src/psyclone/tests/generator_test.py
hiker: All done, I've replaced all of os.path, and using pathlib everywhere now.

doc/user_guide/user_scripts.rst
hiker: Both done.

doc/user_guide/user_scripts.rst
hiker: Done.

hiker:
Ready for next review, IT triggered (issue #3514 for cleanup refactoring added).

hiker:
Documentation failure is caused by the NEMO wiki now requiring an account. IT triggered

src/psyclone/tests/generator_test.py
```diff
@@ -951,9 +950,7 @@ def test_config_flag():
         file references in the environment variable.
     '''
     filename = str(LFRIC_BASE_PATH / "1_single_invoke.f90")
-    config_name = os.path.join(os.path.dirname(os.path.abspath(__file__)),
-                               "test_files", "dummy_config.cfg")
-
+    config_path = Path(get_base_path("lfric")) / ".." / "dummy_config.cfg"
```
arporter: Path(get_base_path("lfric")) is just LFRIC_BASE_PATH?

src/psyclone/tests/generator_test.py
arporter: Many thanks :-)

src/psyclone/generator.py
arporter: Yes, I think I failed to parse the code correctly :-(

arporter (changes requested):
Thanks Joerg. ITs were all green. Link failure (as you say) is not your fault and I've opened #3539 for that.
I just have one small comment on one of the test changes and then this is good to go.

src/psyclone/tests/generator_test.py
hiker: Indeed, thanks!

hiker:
Issue addressed, updated to master, IT triggered.

hiker:
The failing build is caused by a timeout of a NEMO link (which worked fine for me just now).

hiker:
IT passed.

arporter (approved):
Requested change has been made (thanks). ITs were all green (thanks).
Will proceed to merge.

PR #3469, opened 2026-06-25
Control over Fortran file extensions recognised by `ModuleManager`

mn416:
Basic stuff, but I started using PSyclone in a code base with `.f95` files and wondered why imports were not getting resolved. Turns out that `ModuleManager` has a limited set of harded coded file extensions that doesn't include `.f95` so this PR adds the ability to add/remove Fortran file extensions recognised by `ModuleManager`. It also abstracts out a commonly used "doesn't need preprocessing" check into a method, which makes the code a bit cleaner.

Edit: I believe these changes preserve existing behaviour exactly; they just introduce methods to allow modification of the default settings.

### DIFF
--- a/src/psyclone/parse/module_manager.py
+++ b/src/psyclone/parse/module_manager.py
@@ -235,7 +235,8 @@ class ModuleManager:
         if not isinstance(exts, set):
             raise TypeError(
                 f"'fortran_file_exts' must be a set, but found "
-                f"{type(exts).__name__}")
+                f"{type(exts).__name__}"
+            )
         self._fortran_file_exts = exts
 
     # ------------------------------------------------------------------------
--- a/src/psyclone/parse/module_manager.py
+++ b/src/psyclone/parse/module_manager.py
@@ -524,8 +524,10 @@ class ModuleManager:
         # with '.x'. (The latter condition is present to preserve
         # previous behaviour, although it's unclear if that behavior
         # is indeed desired.)
-        base, ext = os.path.splittext(filename)
-        return ext.islower() and not ext.beginswith(".x")
+        base, ext = os.path.splitext(filename)
+        return (ext in self._fortran_file_exts and
+                ext.islower() and
+                not ext.startswith(".x"))
 
     def get_module_info(self, module_name: str) -> Optional[ModuleInfo]:
         """This function returns the ModuleInfo for the specified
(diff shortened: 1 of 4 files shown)


src/psyclone/parse/module_manager.py
```diff
@@ -141,6 +141,10 @@ def __init__(self):
         self._module_pattern = re.compile(r"^\s*module\s+([a-z]\S*)\s*$",
                                           flags=re.IGNORECASE | re.MULTILINE)
 
+        # Files with an extension from this set will be considered
+        # when searching for Fortran files
+        self._fortran_file_exts = {".F90", ".f90", ".X90", ".x90"}
```
LonelyCat124: I thnk we're probably happy to just include F95 and F03 in this list now - the default is to tell fparser to support up to f2008 (though incomplete) so having the default supported files closer to reflecting that is probably best.

LonelyCat124: Also f95 and f03.

LonelyCat124 (changes requested):
@mn416 This mostly seems fine - I have one slight request to change the defaults as well. I will set the ITs running after that.

src/psyclone/parse/module_manager.py
mn416: Done (636f73c).

mn416:
Thanks @LonelyCat124. Change made.

LonelyCat124:
Setting ITs running and then will proceed to merge unless they flag any issues.

LonelyCat124 (approved):
Ready to merge once the new release is done.

src/psyclone/parse/module_manager.py
```diff
@@ -210,6 +216,28 @@ def resolve_indirect_imports(self, value: Union[bool, Iterable[str]]):
...
+        '''
+        :param exts: the set of file extensions that are considered when
+            searching for Fortran files.
+
+        :raises TypeError: if the provided value is not a set.
+        '''
+        if not isinstance(exts, set):
+            raise TypeError(
+                f"'fortran_file_exts' must be a set, but found "
+                f"{type(exts).__name__}")
```
LonelyCat124: Actually @mn416 one minor formatting thing - would you mind adjust this to
```
raise TypeErrror(
    f"...."
    f"...."
)
```
before I merge? I can't merge this until after our release anyway

LonelyCat124 (changes requested):
One minor change requested actually.

src/psyclone/parse/module_manager.py
mn416: Done (53b2398).

mn416:
Thanks @LonelyCat124. Formatting done.

LonelyCat124 (approved):
This is ready for merge once the release is done.

PR #3470, opened 2026-06-26
(Closes #3364) Prepare for release 3.3.0

arporter:
Just updates the version number in the two locations.
At merge time we'll need to update the changelog too.
As the 'update to fparser 0.2.4' has just gone on, we know that all of the fparser settings in the various workflow files are correct already.

### DIFF
--- a/src/psyclone/version.py
+++ b/src/psyclone/version.py
@@ -37,14 +37,16 @@
 
 ''' Single location for the current version number of PSyclone. This is
     used in setup.py and
-    doc/{user_guide,developer_guide,reference_guide/source}/conf.py '''
+    doc/{user_guide,developer_guide,reference_guide/source}/conf.py
+
+    It is NOT used in doc/reference_guide/doxygen.config'''
 
 __MAJOR__ = 3
 __MINOR__ = 3
-__MICRO__ = 1
+__MICRO__ = 0
 
-# Version suffix (e.g. -rc1 or -dev)
-_VERSION_SUFFIX = "-dev"
+# Version suffix e.g. "-rc1", "-dev" or "" (for a full release)
+_VERSION_SUFFIX = ""
 
 __SHORT_VERSION__ = f"{__MAJOR__:d}.{__MINOR__:d}{_VERSION_SUFFIX}"
 __VERSION__ = f"{__MAJOR__:d}.{__MINOR__:d}.{__MICRO__:d}{_VERSION_SUFFIX}"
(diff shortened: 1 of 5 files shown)


arporter:
Ready for review (@hiker or @sergisiso) and for final testing at the Met Office (@andrewcoughtrie). Sergi tells me that he's also brought the associated `lfric_apps` PR up-to-date: https://github.com/MetOffice/lfric_apps/pull/545

src/psyclone/version.py
```diff
@@ -37,14 +37,16 @@
...
     used in setup.py and
-    doc/{user_guide,developer_guide,reference_guide/source}/conf.py '''
+    doc/{user_guide,developer_guide,reference_guide/source}/conf.py
+
+    It is NOT used in doc/reference_guide/doxygen.config'''
 
 __MAJOR__ = 3
 __MINOR__ = 3
-__MICRO__ = 1
+__MICRO__ = 0
```
sergisiso, line 46: It's a bit unfortunate that we have to move a number down. How did that happen? Was it because we did a release candidate and didn't do the associated release?

sergisiso (changes requested):
@arporter Since this cycle we moved to pyproject so I attempted the build manually and everything worked, but it has the same warning that we fixed in fparser:
```
SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
!!

        ********************************************************************************
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2027-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
```

Could you fix it here as well please?

src/psyclone/version.py
arporter: Yes. We did the rc release and then I bumped the working version to 3.3.1-dev. In retrospect this should probably have been `3.3.0-rc2-dev`.

arporter: (I was expecting that we'd make the release shortly after the rc but in the end, it didn't make sense to do that.)

arporter:
License string updated (GH wouldn't let me respond to the comment directly) and I've pushed the changes.
Ready for another look/to hear back from the MO.

sergisiso (approved):
@arporter All good now, just waiting for the green light from the MO.

PR #3472, opened 2026-06-29
(Closes #3471) get option sub transformation fix

### LINKED ISSUE #3471: Sub transformations options cause an error with get_option
If you have a transformation with sub transformations, but call `get_option` from the main transformation, it gives an error due to `get_valid_options` not supporting the sub transformations. I don't think get_valid_options *should* support the sub transformations (we have the split_kwargs functionality for this), but the `get_option` code probably needs fixing.

LonelyCat124:
Small PR that fixes an issue where using get_option to check if a sub transformation option has a specific value (e.g. to control a metatransformation's behaviour) fails on current implementation.

Ready for review from @sergisiso


sergisiso (approved):
@LonelyCat124 This is a small fix to metatransformations, so it can be merge now for the upcoming micro release. The implementation and test look good. I triggered ITs just in case.

sergisiso:
The ITs failed but I think it is related to fetching Jules, not the changes in this PR. I will investigate.

sergisiso:
@LonelyCat124 and also @arporter This is now ready to merge and it fixes the LFRic ITs. But since UKMO is evaluating the head of master as a release candidate, I will block it and wait for a possible micro release before merging more things.

PR #3473, opened 2026-06-30
(towards #3124) Make the loop variable as child of Loop instead of an attribute

sergisiso:


sergisiso:
@arporter @LonelyCat124 This is ready for review. Switching to in-tree loop references has been relatively painless, with the only oddity being the lfric null loops (which are loops that are not loops?) and have a dummy symbol as they don't exist after lowering anyway.

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -965,15 +965,16 @@ def test_node_dag(tmpdir, have_graphviz):
     assert isinstance(dag, graphviz.Digraph)
 
     result = my_file.read()
+    print(result)
```
LonelyCat124: delete.

LonelyCat124 (changes requested):
I had a very quick scan through and it looks to make sense. There's a leftover print statement and this opens up one other issue that wasn't previously addressed in the definition use chains.

Can you create a test using `loop.variable.next_accesses()` and check what the behaviour is? Once we know that I'll have a more detailed look through.

src/psyclone/tests/psyir/nodes/node_test.py
sergisiso: Done

sergisiso:
> Can you create a test using loop.variable.next_accesses() and check what the behaviour is?

@LonelyCat124 `loop.variable` won't have a next_access() method because it points to the symbol `self._children[0].symbol` but next_access is a reference method. I kept it this way to keep all the existing `loop.variable` code unmodified (but I was considering adding a `variable_reference` or similar for the typical syntactic navigation methods that we offer).

Is this what you wanted to check or do you mean a test with `loop.children[0].next_access()`?

LonelyCat124:
>Is this what you wanted to check or do you mean a test with loop.children[0].next_access()?

@sergisiso I guess a test with `loop.children[0].next_accesses()` if `.variable` isn't the Reference (i didn't check that detail yet).

src/psyclone/domain/lfric/lfric_loop.py
```diff
@@ -1104,6 +1106,32 @@ def independent_iterations(self,
...
+            structure accessors) and the values are AccessSequence
+            (a sequence of AccessTypes).
+
+        '''
+        var_accesses = VariablesAccessMap()
+
+        if self.variable.name != "null":
+            var_accesses.add_access(Signature(self.variable.name),
+                                    AccessType.WRITE, self)
+            # This re
```
LonelyCat124: This comment is unfinished.

src/psyclone/psyir/nodes/loop.py
```diff
@@ -53,11 +54,13 @@
...
-    the first one represents the loop lower bound, the second one represents
-    the loop upper bound, the third one represents the step value and the
-    fourth one is always a PSyIR Schedule node containing the statements inside
-    the loop body.
+    '''Node representing a loop within the PSyIR. It has 5 mandatory children:
+     - the first one represents the loop variable,
+     - the second one represents the loop lower bound,
+     - the third one represents the loop upper bound,
+     - the forth one represents the step value
+     - and the fith one is always a PSyIR Schedule node containing the
```
LonelyCat124: fith/fifth

sergisiso:
@LonelyCat124 I added the suggested check in references_test.py. Ready for another look

src/psyclone/psyir/nodes/loop.py
```diff
@@ -280,12 +267,37 @@ def _check_completeness(self):
         # We cannot just do str(self) in this routine we can end up being
         # called as a result of str(self) higher up the call stack
         # (because loop bounds are evaluated dynamically).
-        if len(self.children) < 4:
+        if len(self.children) < 5:
             raise InternalError(
                 f"Loop is incomplete. It should have exactly 4 "
```
LonelyCat124: Error message is out of date.

src/psyclone/tests/psyir/nodes/reference_test.py
```diff
@@ -230,6 +230,9 @@ def test_reference_next_accesses(fortran_reader):
     assert len(b.next_accesses()) == 1
     assert b.next_accesses()[0] == b
 
+    # Check the next_access of the loop variable
+    assert loop.children[0].next_accesses() == [loop.loop_body.children[0].rhs]
```
LonelyCat124: Thanks for this - can you add a test what happens if we have:
```
do i = 1, 100
   a(1) = a(1) + 1
end do
```

I'm not sure what will happen with the next_access for `i` and want to make sure it makes sense and we understand it.

Also in `definition_use_chains_forward_dependence_test.py` line 664 we have a test that returns a `Loop` (instead of its variable). Perhaps the best thing to do is to create an issue "Definition use chains return a Loop instead of the Loops' variable reference" (or similar) and we add a TODO to all the tests that return these loops (and maybe a note in the DUC docs with the TODO also?) since there is now a reference that could be pointed at.

LonelyCat124 (commented):
@sergisiso A few more changes
I think a variable_reference (or some better name) getter (rather than `children[0]`) is probably worthwhile to avoid random potential accesses to `children[0]`.

Also I think `Reference.is_write` should be updated for these new References also?

src/psyclone/domain/lfric/lfric_loop.py
sergisiso: Ups, filled the rest now

src/psyclone/psyir/nodes/loop.py
sergisiso: Fixed

src/psyclone/tests/psyir/nodes/reference_test.py
sergisiso: The example above seems to work as expected. What didn't is the previous_access of 'i' as it does not consider it the beginning of the DUC, I added another test for this.

I also added the requested TODO everywhere as I thought was related.

sergisiso:
@LonelyCat124 This is ready for another look. I tried to capture in the TODO the rough edges with the previous/next access that now can potentially be improved.

src/psyclone/tests/psyir/nodes/reference_test.py
LonelyCat124: We discussed this and decided what the answer of this should be is a DUC problem so happy to resolve this for now.

sergisiso:
>I think a variable_reference (or some better name) getter (rather than children[0]) is probably worthwhile to avoid random potential accesses to children[0].

Done

>Also I think Reference.is_write should be updated for these new References also?

Done

@LonelyCat124 These are also addressed now, ready for another look

LonelyCat124:
Going to run ITs but otherwise everything looks good

LonelyCat124 (approved):
ITs were all green, proceeding to merge.

PR #3476, opened 2026-07-01
(closes #3475) Make OMPParallelLoopTrans respect the force_private kwarg

### LINKED ISSUE #3475: OMPParallelLoopTrans does not respect the force_private kwarg

sergisiso:
This fixes one of the issues that LFRic_apps is having to adopt psyclone 3.3.

@LonelyCat124 I got a bit nervous about modifying the existing 'enable_reductions' option because it does a local `options.copy()` that I don't understand. So for the time being I left it as an option. I propose we unify this after the release.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+   35) PR #3476 for #3475. Fixes OMPParallelLoopTrans to support the
+   force_private keyword argument.
+
    34) PR #3440. Update to use release 0.2.4 of fparser.
  
    33) PR #3467. Update CI compilers.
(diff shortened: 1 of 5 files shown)


LonelyCat124 (approved):
Ah I think this is why this escaped my updates before. This transformation just force overrides specified options even if they've been provided. I think we can make this nicer later - essentially now that we have `inherit=False` working, we can avoid exposing the `reduction_ops` kwarg in the docstring, or directly on the function. This means we can always just ignore it (perhaps with a warning if `reduction_ops` is in **kwargs)

The `options.copy()` is just because the transformation can modify the dictionary and this needs to be modified, but as you say we don't need to fix this now.

I'll build the docs and check those quick, but otherwise this is ready to merge.

The only other thing is that this transformation currently won't match rules w.r.t options vs kwargs, but again not essential right now but we should fix as soon as possible after.

@sergisiso I assume this one needs merging now before the release?

LonelyCat124:
All is ready to be merged if the CI passes.

PR #3480, opened 2026-07-02
(Closes #3478) Fixes the OMPParallelLoopTrans to use modern kwargs approach

### LINKED ISSUE #3478: OMPParallelLoopTrans doesn't fully use **kwargs
Ah I think this is why this escaped my updates before. This transformation just force overrides specified options even if they've been provided. I think we can make this nicer later - essentially now that we have `inherit=False` working, we can avoid exposing the `reduction_ops` kwarg in the docstring, or directly on the function. This means we can always just ignore it (perhaps with a warning if `reduction_ops` is in **kwargs)

The `options.copy()` is just because the transformation can modify the dictionary and this needs to be modified, but as you say we don't need to fix this now.

I'll build the docs and check those quick, but otherwise this is ready to merge.

The only other thing is that this transformation currently won't match rules w.r.t options vs kwargs, but again not essential right now but we should fix as soon as possible after.

_Originally posted by @LonelyCat124 in https://github.com/stfc/PSyclone/pull/3476#pullrequestreview-4609976799_

LonelyCat124:
Small improvement to @sergisiso's work on this to use modern kwargs approach now I understand how the transformation works better. Its weird because the call to super.validate() does a bunch of the actual work so its confusing why the options matter at first glance.

### DIFF
--- a/src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
+++ b/src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
@@ -116,11 +116,13 @@ class OMPParallelLoopTrans(OMPLoopTrans):
                 if "reduction_ops" in kwargs:
                     del kwargs["reduction_ops"]
                     logger.warning(
-                        f"{self.name} overriddes the provided reduction_ops "
+                        f"{self.name} overrides the provided reduction_ops "
                         f"keyword argument to those supported by PSyclone."
                     )
                 reduction_ops = list(MAP_REDUCTION_OP_TO_OMP.keys())
 
+        # reduction_ops is the argument used by the superclass to determine
+        # whether to allow reductions, so we don't pass enable_reductions.
         self.validate(node, options=local_options, force_private=force_private,
                       reduction_ops=reduction_ops, **kwargs)
 
--- a/src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
+++ b/src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
@@ -43,7 +43,7 @@
 import logging
 from typing import Iterable
 from psyclone.psyir.nodes import (
-    OMPParallelDoDirective, OMPReductionClause)
+    OMPParallelDoDirective, OMPReductionClause, Loop)
 from psyclone.psyir.nodes.omp_directives import MAP_REDUCTION_OP_TO_OMP
 from psyclone.psyir.transformations.omp_loop_trans import OMPLoopTrans
 from psyclone.utils import transformation_documentation_wrapper
@@ -73,8 +73,9 @@ class OMPParallelLoopTrans(OMPLoopTrans):
     def __str__(self):
         return "Add an 'OpenMP PARALLEL DO' directive"
 
-    def apply(self, node,
+    def apply(self, node: Loop,
               force_private: Iterable[str] = tuple(),
+              enable_reductions: bool = False,
               options=None, **kwargs):
         ''' Apply an OMPParallelLoop Transformation to the supplied node
         (which must be a Loop). In the generated code this corresponds to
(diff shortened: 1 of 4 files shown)


src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
```diff
@@ -89,20 +90,39 @@ def apply(self, node,
...
+                # it as needed.
+                if "reduction_ops" in kwargs:
+                    del kwargs["reduction_ops"]
+                    logger.warning(
+                        f"{self.name} overriddes the provided reduction_ops "
+                        f"keyword argument to those supported by PSyclone."
+                    )
+                reduction_ops = list(MAP_REDUCTION_OP_TO_OMP.keys())
 
         self.validate(node, options=local_options, force_private=force_private,
```
arporter, line 140: Should we be passing `enable_reduction` in to `validate` too?

arporter: If not, please add a comment explaining.

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -915,6 +915,50 @@ def test_parallellooptrans_refuse_codeblock():
...
+    else:
+        # If not using the options dict we should have a logged message too.
+        caplog.clear()
+        with caplog.at_level(logging.WARNING,
+                             "psyclone.psyir.transformations."
+                             "omp_parallel_loop_trans"):
+            OMPParallelLoopTrans().apply(loop, enable_reductions=True,
+                                         reduction_ops=[])
+            assert caplog.records[0].levelname == "WARNING"
+            assert ("OMPParallelLoopTrans overriddes the provided "
```
arporter: "overrides" :-)

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -915,6 +915,50 @@ def test_parallellooptrans_refuse_codeblock():
             "by a OMPParallelLoopTrans transformation" in str(err.value))
 
 
+@pytest.mark.parametrize("use_options_dict", [True, False])
```
arporter: I've just noticed that although `OMPParallelLoopTrans` is in its own module, its tests are still in this file. Would you mind moving them out into their own test file if possible?

src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
arporter: I guess we don't need to because the presence of `reduction_ops` signals the same thing?

src/psyclone/tests/psyir/transformations/transformations_test.py
```diff
@@ -915,6 +915,50 @@ def test_parallellooptrans_refuse_codeblock():
...
+def test_ompparallellooptrans_reductions(fortran_reader, fortran_writer,
+                                         caplog, use_options_dict):
+    '''Check that OMPParallelLoopTrans behaves correctly with reductions.'''
```
arporter: It would help me a bit if you fleshed this out: e.g. that it inherits/overides (whichever is correct) some property from OMPLoopTrans?

arporter (changes requested):
Thanks Aidan. One day I will understand this. In the meantime I need a couple more comments...
I've also noticed that we could do with moving the tests for this class. Since this is a small PR I've requested that you do that if it is straightforward.

src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
LonelyCat124: Yeah this is a weird one, the superclass (that implements validate) doesn't care about this enable_reduction option, its unique to this subclass. I'll add a comment.

LonelyCat124: The superclass validate also doing a bunch of "work" and not just validation is strange to me, but not something designed for this quick PR.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/transformations/transformations_test.py
LonelyCat124: Added

LonelyCat124:
@arporter Addressed the comments, ready for another look. This is a bit of a messy section of code all around, but this was just to fix up using kwargs after Sergi's changes to the functionality for the last release.

src/psyclone/psyir/transformations/omp_parallel_loop_trans.py
arporter: Perhaps we should discuss this separately. I've had a look at `ParallelLoopTrans.validate()` and can't see anything that isn't validation? That said, I see it does an LFRic-specifc check which probably should be done elsewhere.

LonelyCat124: I think the thing I was thinking of (probably doesn't qualify as a bunch of work) is this section:
```
                if (reduction_ops and
                        message.code == DTCode.WARN_SCALAR_REDUCTION):
                    if len(message.var_infos) == 1:
                        (sig, access_info) = message.var_infos[0]
                        red_tool = ReductionInferenceTool(reduction_ops)
                        clause = red_tool.attempt_reduction(sig, access_info)
                        if clause:
                            self.inferred_reduction_clauses.append(clause)
                            continue
```
The `self.inferred_reduction_clauses` are used by apply to create reduction clauses I think, which makes me feel like something here isn't a good separation of concerns. Its not a major issue I think.

arporter: Ah! I didn't spot that. I agree; it's setting some state and that's not what a `validate` should be doing.

arporter: I found (and updated) #2648 for this.

arporter (commented):
Thanks @LonelyCat124. All looks good now. I really didn't want to open a new Issue but happily I found an existing one (#2648) so I've updated that.
Just waiting on the ITs now.

arporter:
The LFRic extraction test failed yesterday (looked like an SRS problem) so I'll re-run.

arporter (approved):
ITs were all green. Will proceed to merge.

PR #3481, opened 2026-07-06
(Towards/Closes? #3463) max parallel assignments

LonelyCat124:
This is probably waiting on #3464 before it can be merged.


LonelyCat124:
@sergisiso @arporter This is ready for a first look now.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
```diff
@@ -67,6 +70,8 @@ class MaximalOMPParallelRegionTrans(MaximalRegionTrans):
     _transformation = OMPParallelTrans
     _SUB_TRANSFORMATIONS = [OMPParallelTrans]
     # Tuple of statement nodes allowed inside the _transformation
+    # Note that Assignments may also be allowed, but they are a special
+    # case not included in this selection.
     _allowed_contiguous_statements = (
```
arporter, line 48: Not strictly part of this PR but why "contiguous"?

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
```diff
@@ -86,6 +91,101 @@ class MaximalOMPParallelRegionTrans(MaximalRegionTrans):
...
+        # Find the abs_position of the final node in the current_block.
+        last_element = current_block[-1]
+        # For reviewer - this could be implemented using
+        # following_node.abs_position-1, but if there is no following_node
+        # we'd still need to return to this (or I guess some recursive
+        # loop over children[-1]. Let me know if you have a preference
+        # w.r.t performance as this is probably the worst option but
+        # easy to read.
+        start_position = node.lhs.abs_position
+        end_position = last_element.walk(Node)[-1].abs_position
```
arporter: This does feel inefficient. It might be worth adding a new method to `Node` that just finds the 'last' node in a subtree? Something like:
```python
def  last_node(self):
    if self.children:
        return self.children[-1].last_node()
    return self
```
perhaps?

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
```diff
@@ -86,6 +91,101 @@ class MaximalOMPParallelRegionTrans(MaximalRegionTrans):
...
+        # transformation.
+        return True
+
+    def _apply_transformation(self, block: list[Node],
+                              **kwargs):
+        '''
+        Applies the transformation defined by self._transformation to the
+        block supplied, with the kwargs provided.
+
+        The default implementation does nothing else, however the
```
arporter: Please update/change this to explain what this subclass reimplementation brings to the party.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
```diff
@@ -86,6 +91,101 @@ class MaximalOMPParallelRegionTrans(MaximalRegionTrans):
...
+        '''
+        # If we have any assignments directly in the block then we need to
+        # do privatisation of the lhs variable
+        force_private = []
+        if "force_private" in kwargs:
+            force_private.extend(kwargs["force_private"])
+            del kwargs["force_private"]
+        for node in block:
+            if isinstance(node, Assignment):
+                if node.lhs.symbol.name not in force_private:
```
arporter, line 149: Would it be possible to add a comment to the output code to say that PSyclone has chosen to privatise this variable?

src/psyclone/psyir/transformations/maximal_region_trans.py
```diff
@@ -108,24 +110,27 @@ def _satisfies_minimum_region_rules(self, region: list[Node]) -> bool:
...
+        if self._node_allowed(node, current_block):
             return True
 
         if isinstance(node, (Loop, WhileLoop)):
             # Check that all contents of the loop body can be part
             # of the region.
             for child in node.loop_body:
-                if not self._can_be_in_region(child):
+                if not self._can_be_in_region(child, current_block):
                     break
```
arporter: Can you just `return False` here rather than `break` and then you don't need the `else` on the `for` loop?

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
```diff
@@ -86,6 +91,101 @@ class MaximalOMPParallelRegionTrans(MaximalRegionTrans):
...
+                    force_private.append(node.lhs.symbol.name)
+
+        self._transformation().apply(block, force_private=force_private,
```
arporter: It's probably a bit better to call `super()._apply_transformation(...)` here.

src/psyclone/tests/psyir/transformations/maximal_omp_parallel_region_trans_test.py
```diff
@@ -153,3 +154,155 @@ def test_maximal_ompparallel_region_trans_apply(fortran_reader):
...
+    # from it.
+    assert maxpartrans._node_allowed(assignment, [loop])
+
+    psyir = fortran_reader.psyir_from_source('''
+    subroutine test
+
+    integer :: a
+    integer :: b, i
+
+    a = a + 1
```
arporter, line 260: You could perhaps add this Fortran into the previous fragment rather than parse a new  fragment here?

src/psyclone/tests/psyir/transformations/maximal_omp_parallel_region_trans_test.py
```diff
@@ -153,3 +154,155 @@ def test_maximal_ompparallel_region_trans_apply(fortran_reader):
...
+    assert not maxpartrans._node_allowed(assign, [loop])
+
+    # Assignment is not allowed if the next access is a read.
+    psyir = fortran_reader.psyir_from_source('''
+    subroutine test
+
+    integer :: a
+    integer :: b, i
+
+    a =  1
```
arporter, line 289: Again, you could extend the previous Fortran to include this case too (i.e. so it has three separate loops etc.).

src/psyclone/tests/psyir/transformations/maximal_region_trans_test.py
```diff
@@ -91,7 +91,7 @@ def test_can_be_in_region(fortran_reader, statement, expected):
     psyir = fortran_reader.psyir_from_source(code)
     routine = psyir.walk(Routine)[0]
     trans = MaxParTrans()
-    assert trans._can_be_in_region(routine.children[0]) == expected
```
arporter, line 65: This file does not 100% cover the corresponding source file - L111 is missed.

src/psyclone/tests/psyir/transformations/maximal_omp_parallel_region_trans_test.py
```diff
@@ -37,6 +37,7 @@
```
arporter, line 9: Not strictly this PR but, since I noticed, pls could you rename this test file to match the corresponding source file? (Latter has "..omp_parallel..." rather than "...ompparallel..".)

arporter (changes requested):
Nice job Aidan. For finding the last node in a region, I'm wondering whether it's worth having a new method to do that? There are probably other places where it would be useful? (I checked PSyTran but it doesn't have such a thing.)

Not this PR but I think we really could do with testing OMP compilation. I think we have an issue somewhere that I'll look for now.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: You can have non-contiguous regions, i.e.
```
Region A of contiguous allowed statements
Something not allowed
Region B of contiguous allowed statements
```

Allowed statements could be allowed but since it was describing neighbouring statements we had said contiguous. It also explains the expected behaviour without needing to check the superclass method better.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: Added

LonelyCat124: There are probably a lot of places throughout PSyclone that this pattern has been used, I'll keep an eye out for it as I work on other sections of the code - is it worth making an issue?

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/maximal_region_trans.py
LonelyCat124: Done

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: Done

src/psyclone/tests/psyir/transformations/maximal_omp_parallel_region_trans_test.py
LonelyCat124: Changed into 3 tests.

src/psyclone/tests/psyir/transformations/maximal_region_trans_test.py
LonelyCat124: Added a new test to cover this.

src/psyclone/tests/psyir/transformations/maximal_omp_parallel_region_trans_test.py
LonelyCat124: Done I think

LonelyCat124:
@arporter ready for another look.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
arporter: Aha, thanks.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
arporter: Let's not make an issue but we should publicise this new functionality at the weekly meetings.

arporter: You've added the new method to Node but not used it here? ;-)

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -2088,3 +2088,26 @@ def test_get_all_accessed_symbols(fortran_reader):
...
+
+
+def test_last_node(fortran_reader):
+    '''Test the last_node functionality of the Node class.'''
+    code = """subroutine test_sub()
+    integer :: i, j, k
+
+    k = 12
+    i = j + k
+    end subroutine test_sub
```
arporter, line 2069: Worth testing with an `if..else if...else` block too?

src/psyclone/tests/psyir/nodes/node_test.py
```diff
@@ -2088,3 +2088,26 @@ def test_get_all_accessed_symbols(fortran_reader):
...
+    """
+    psyir = fortran_reader.psyir_from_source(code)
+
+    # The last_node of the psyir statement is the k in i = j + k
+    assigns = psyir.walk(Assignment)
+    assert psyir.last_node() is assigns[1].rhs.children[1]
```
arporter: I wasn't sure about the naming of the new method but I googled and then thought that `get_last_node()` is more consistent with how we normally do things - it's not a simple property that we're looking up and it involves some work so we normally have a method with a name beginning with `get_`.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
arporter: In double checking, I grepped and found:
```
psyir/tools/definition_use_chains.py:                    last_node = ancestor.walk(Node)[-1]
```
which looks like a good candidate to use this functionality too.

arporter (changes requested):
Nice, thanks Aidan. Only snag is that I think you've omitted to actually use the new 'last_node' method in anger :-) I've  also had a think and concluded that it should probably be named `get_last_node` if you and @sergisiso agree?
Once it is being used, I'll launch the ITs (or you could).

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: Yeah. Will add a unit test of the function too.

src/psyclone/tests/psyir/nodes/node_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/nodes/node_test.py
LonelyCat124: I did `get_last_descendent_node`

LonelyCat124:
@arporter Ready for another look.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
arporter: This comment can be removed now?

arporter (changes requested):
Nice job Aidan. Bonus points for finding the extra places that the new method could be used.
Only snag is that I think we should use "descendant" (with an 'a') as that seems to be the commonly-accepted spelling. There's also a comment about the previous implementation that can be removed now.
I've fired off the ITs as those are only minor things that need doing.

src/psyclone/psyir/transformations/maximal_omp_parallel_region_trans.py
LonelyCat124: Removed yeah.

LonelyCat124:
Sorted those 2 issues @arporter

arporter (approved):
All good now, thanks Aidan. I want a clear IT run before I merge this and GHA has got some technical troubles at the moment.

arporter:
ITs were finally all green. I'm going to put Sergi's big #3508 on first and then this one.

arporter:
(The link that failed in the linkcheck works fine for me today.)

arporter:
The linkcheck failure is still a page on OpenMP and it still works fine for me so I shall ignore it.

PR #3483, opened 2026-07-06
(Closes #3482) Test how docs look without sphinx.stringify_annotations

### LINKED ISSUE #3482: Test failures if sphinx is not installed.
Sphinx is only a pre-requisit for `doc` insatallation, not for `test`. There is therefore an implementation of `utils.stringify_annotation`, but this returns incorrect values (at least for python 3.14 and 3.10):
```
from sphinx.util.typing import stringify_annotation
from psyclone.utils import stringify_annotation as sa
stringify_annotation(int)
'int'
sa(int)
"<class 'int'>"
```

Fix is trivial:
```
joerg@Joerg-Surface:~/work/psyclone/src/psyclone$ git diff utils.py
diff --git a/src/psyclone/utils.py b/src/psyclone/utils.py
index a7153d6854..07ecc536f0 100644
--- a/src/psyclone/utils.py
+++ b/src/psyclone/utils.py
@@ -88,7 +88,7 @@ def stringify_annotation(annotation) -> str:
 
     :returns: The string representation of annotation.
     '''
-    return str(annotation)
+    return annotation.__name__
 
 
 def transformation_documentation_wrapper(*args,

```
The only question is, do we instead want to make sphinx a requirement for testing (and avoid our own implementation of `stringify_annotation`?)

sergisiso:
They seem to look alright, an even preserve some links?

image


sergisiso:
@LonelyCat124 One for you to have a look.

LonelyCat124:
@sergisiso I had a quick look at sphinx's implementation. What we probably don't support with this is forward referencing, e.g. when we have `"MyType"`, as we probably want to remove the quotes in that case. There is also the potential of checking with `annotation, type.ForwardRef` but that might only happen with 3.14.

Are the `reduction_ops` an inherited parameter in the case you show? My concern is whether we keep the linking for inherited options on transformations with this change.

sergisiso:
> Are the reduction_ops an inherited parameter in the case you show? My concern is whether we keep the linking for inherited options on transformations with this change.

Yes, I should have said that the example figure above is from OMPLoopTrans and it inherits all those options. Also it has kept the link when needed.

The operator.Operator is strange as this is not the type on the annotation! But it turns out sphinx has the same issue?
```
from sphinx.util.typing import stringify_annotation
>>> stringify_annotation(list[BinaryOperation.Operator])
'list[psyclone.psyir.nodes.operation.Operator]'
```

sergisiso:
Do we have some example with an option and a forward reference I could look at?

LonelyCat124:
> Do we have some example with an option and a forward reference I could look at?

It doesn't look like we do this anywhere in Transformations at current, but I could see it one day being required. It'd probably need its own test just to be safe (mostly that we don't get spurious `"` in the actual docstring. My own quick test does show our current behaviour is the same as sphinx's stringify_annotation for a string though:
```
>>> from sphinx.util.typing import stringify_annotation
>>> stringify_annotation("BinaryOperation")
'BinaryOperation'
>>> stringify_annotation("\"BinaryOperation\"")
'"BinaryOperation"'
>>> stringify_annotation("BinaryOperation")
'BinaryOperation'
>>> from psyclone.utils import stringify_annotation as sa
>>> sa("BinaryOperation")
'BinaryOperation'
```



I'm going to ignore any requirement for us to handle `...` for now as I don't expect us to use it.

sergisiso:
This is ready for anther look @LonelyCat124  I added the extra test, the type elipsis worked for 3.14 (which is the one that generated the documentation)

doc/developer_guide/pip_requirements.txt
```diff
@@ -1,43 +0,0 @@
...
-# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-# POSSIBILITY OF SUCH DAMAGE.
-# -----------------------------------------------------------------------------
-# Author: A. R. Porter, STFC Daresbury Lab
-
-# pip requirements file specifying the packages required to build
-# the PSyclone documentation. Used by readthedocs.
```
LonelyCat124, line 37: Why is it safe to remove this file? Don't we still have the requirement on sphinx to build the documentation on RTD?

LonelyCat124 (changes requested):
@sergisiso I think this is mostly fine, I had one question that I wasn't sure about - can we test that this branch builds correctly on RTD before we merge?

doc/developer_guide/pip_requirements.txt
sergisiso: This file is a leftover from when the developers_guide was a separate documentation.

I temporarily added this branch on RDT, the link is here: https://psyclone.readthedocs.io/en/3482_remove_sphinx_stringify_annotations/

sergisiso:
Back to you @LonelyCat124

sergisiso:
@LonelyCat124 The RTD build with updated versions succeeded, is the same link as in the comment above.

src/psyclone/tests/utils_test.py
```diff
@@ -338,22 +338,38 @@ def apply(self, node, opt3: int = 1, **kwargs):
...
+            # truncates the resulting string
+            assert "tuple[int, ...]" == anno or "tuple" == anno
+
+        if "custom_class" == k:
+            anno = stringify_annotation(v.annotation)
+            assert "Transformation" == anno
+
+        if "forwardref" == k:
+            anno = stringify_annotation(v.annotation)
+            assert "Transformation" == anno
```
LonelyCat124: I had a question now regarding this test - I think it can just be simpler:

`assert stringify_annotation(Union[bool | int]) in ["typing.Union[bool, int]", "bool | int"]` for each of these instead of needing to create a function and then disentangling the annotations.

LonelyCat124 (changes requested):
One question regarding the test -  I think we could make i much simpler.

src/psyclone/tests/utils_test.py
LonelyCat124: Also on 3.12 I have failing tests on this PR:

`tests/domain/lfric/transformations/lfric_redundant_computation_trans_test.py::test_rc_invalid_depth_type`

```
        else:
>           assert (f"'depth' option expects type 'int | "
                    f"psyclone.psyir.nodes.datanode.DataNode | None' but received "
                    f"'2' of type '{type('2').__name__}'" in str(excinfo.value))
E           assert "'depth' option expects type 'int | psyclone.psyir.nodes.datanode.DataNode | None' but received '2' of type 'str'" in "'LFRicRedundantComputationTrans' received options with the wrong types:\n'depth' option expects type 'typing.Union[in...anode.DataNode, NoneType]' but received '2' of type 'str'.\nPlease see the documentation and check the provided types."
E            +  where "'LFRicRedundantComputationTrans' received options with the wrong types:\n'depth' option expects type 'typing.Union[in...anode.DataNode, NoneType]' but received '2' of type 'str'.\nPlease see the documentation and check the provided types." = str(TypeError("'LFRicRedundantComputationTrans' received options with the wrong types:\n'depth' option expects type 'typin...node.DataNode, NoneType]' but received '2' of type 'str'.\nPlease see the documentation and check the provided types."))
E            +    where TypeError("'LFRicRedundantComputationTrans' received options with the wrong types:\n'depth' option expects type 'typin...node.DataNode, NoneType]' but received '2' of type 'str'.\nPlease see the documentation and check the provided types.") = <ExceptionInfo TypeError("'LFRicRedundantComputationTrans' received options with the wrong types:\n'depth' option expe...Node, NoneType]' but received '2' of type 'str'.\nPlease see the documentation and check the provided types.") tblen=4>.value
```

and

`tests/psyir/transformations/transformations_test.py::test_omploop_trans_new_options`

 ```
        if sys.version_info >= (3, 10):
>           assert ("'OMPLoopTrans' received options with the wrong types:\n"
                    "'collapse' option expects type 'int | bool' but "
                    "received 'x' of type 'str'.\n"
                    "Please see the documentation and check the provided types."
                    in str(excinfo.value))
E           assert "'OMPLoopTrans' received options with the wrong types:\n'collapse' option expects type 'int | bool' but received 'x' of type 'str'.\nPlease see the documentation and check the provided types." in "'OMPLoopTrans' received options with the wrong types:\n'collapse' option expects type 'typing.Union[int, bool]' but received 'x' of type 'str'.\nPlease see the documentation and check the provided types."
E            +  where "'OMPLoopTrans' received options with the wrong types:\n'collapse' option expects type 'typing.Union[int, bool]' but received 'x' of type 'str'.\nPlease see the documentation and check the provided types." = str(TypeError("'OMPLoopTrans' received options with the wrong types:\n'collapse' option expects type 'typing.Union[int, bool]' but received 'x' of type 'str'.\nPlease see the documentation and check the provided types."))
E            +    where TypeError("'OMPLoopTrans' received options with the wrong types:\n'collapse' option expects type 'typing.Union[int, bool]' but received 'x' of type 'str'.\nPlease see the documentation and check the provided types.") = <ExceptionInfo TypeError("'OMPLoopTrans' received options with the wrong types:\n'collapse' option expects type 'typing.Union[int, bool]' but received 'x' of type 'str'.\nPlease see the documentation and check the provided types.") tblen=3>.value
```

LonelyCat124: It looks like a change of behaviour where before at 3.10 (until 3.13 I assume) the stringify_annotation from sphinx would give `int | bool` but with this new version we instead get `typing.Union[int, bool]`

sergisiso: I removed the specific type from the assert, let me know if 3.12 is happy with this version.

sergisiso:

LonelyCat124:
Messed up when merging master so will need to rerun the ITs.

LonelyCat124:
ITs all green, will proceed towards merging.

LonelyCat124 (approved):

PR #3490, opened 2026-07-08
(Closes #3488) Fixes an issue with where statements if they are preceeded by comments

### LINKED ISSUE #3488: Internal error when parsing WHERE construct with --keep-comments or --keep-directives
PSyclone reports an internal error when parsing a valid Fortran WHERE construct, but only when either the --keep-comments or --keep-directives option is enabled. The same source is processed successfully when neither option is used.

```((.psyclone_ENES-Rise) ) vka83-nxn02@rupert:~/PSycloneHorizon/other_applications$ psyclone --keep-comments nemo-5.0/MULE/BLD/ppsrc/nemo/diaptr.f90
Failed to create PSyIR from file 'nemo-5.0/MULE/BLD/ppsrc/nemo/diaptr.f90' due to:
PSyclone internal error: Failed to find opening where construct statement in: !
  WHERE (zmsk1 /= 0._wp)
  zstr(:, jn, jp_tem) = SUM(pvtr_int(:, :, jn, jp_vtr), 2) * SUM(pvtr_int(:, :, jn, jp_tem), 2) / zmsk1
  zstr(:, jn, jp_sal) = SUM(pvtr_int(:, :, jn, jp_vtr), 2) * SUM(pvtr_int(:, :, jn, jp_sal), 2) / zmsk1
ELSEWHERE
  zstr(:, jn, jp_tem) = 0._wp
  zstr(:, jn, jp_sal) = 0._wp
END WHERE. File was treated as free form.
```

LonelyCat124:
This fixes #3488. The implementation is a bit of a mess since the expectation from `fparser2.py` is to use the same behaviour for both `Fortran2003.Where_Construct` and `Fortran2003.Where_Stmt` and these have different structures for how they are output by fparser itself.

There are tests with preceding comments for both types of statement in the test now, but the structure of how to create the children is dependent on where the `Where_Construct_Stmt` is for the `Where_Construct`. This still keeps the check of the first non-comment in the `Where_Construct` must be the `Where_Construct_Stmt`

Assuming no issues with testing this is ready for a look from @arporter or @sergisiso


src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4697,10 +4695,19 @@ def _contains_intrinsic_reduction(pnodes):
             was_single_stmt = True
             annotations = ["was_where", "was_single_stmt"]
             logical_expr = [node.items[0]]
+            # There are no comments in the items list for a
```
arporter: Perhaps s/are/can be/

src/psyclone/psyir/frontend/fparser2.py
arporter, line 4737: Is it possible for `node.content[-1]` to be a comment too? e.g. what if the `END WHERE` has an inline comment on it?

arporter: Should be `node.content[first_non_comment:-1]` ?

arporter: Implies we need to extend the test for a WHERE containing a reduction intrinsic?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4918,6 +4927,14 @@ def _contains_intrinsic_reduction(pnodes):
                 self.process_nodes(sched, node.content[1:-1])
         # Convert all uses of array syntax to indexed accesses
         self._array_syntax_to_indexed(ifblock, loop_vars)
+
+        # Compute the preceding comments
```
arporter: Rather than "Compute" how about "Collect all of the preceding comments and associate them with the outermost loop"?

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -718,6 +719,29 @@ def test_where_stmt(fortran_reader):
...
+      WHERE( at_i(:,:) > rn_amax_2d(:,:) ) &
+            a_i(:,:,jl) = a_i(:,:,jl) * rn_amax_2d(:,:) / at_i(:,:)
+    end program where_test
+    '''
+    fortran_reader = FortranReader(ignore_comments=False)
+    psyir = fortran_reader.psyir_from_source(code)
+    routine = psyir.walk(Routine)[0]
+    assert len(routine.children) == 2
+    assert isinstance(routine[1], Loop)
+    assert routine[1].preceding_comment == "Here is a comment."
```
arporter, line 741: Pls add a compilation check too.

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
```diff
@@ -1355,3 +1379,28 @@ def test_nested_where(fortran_reader, fortran_writer, tmp_path):
...
+    end where
+    end subroutine test"""
+    fortran_reader = FortranReader(ignore_comments=False)
+    psyir = fortran_reader.psyir_from_source(code)
+    routine = psyir.walk(Routine)[0]
+    assert len(routine.children) == 2
+    assert isinstance(routine[1], Loop)
+    assert routine[1].preceding_comment == (
+        "Here is a comment\nHere is another comment"
+    )
```
arporter, line 1407: Compilation check pls.

arporter (changes requested):
Thanks Aidan - looks pretty good. I did debate whether it would be more pythonic to iterate over the members of `content` directly but I don't think it would improve the code. It looks like there's a small oversight in the code that checks for an array reduction so I think we need an additional test (or an extension of one).
This is a bug fix so no doc update required.
I will check the ITs next time around (although I don't think we test `keep-comments` yet?).

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Done.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Seemingly not, I added a new test and it just worked out of the bag - I think it ends up elsewhere in the tree and some other code we already have handles putting it *somewhere* (I think as an inline comment on the where which goes at the end).

If we have this input:
```
    code = """subroutine test()
    real, dimension(100) :: arr1, arr2
    arr1 = 0.0
    where (arr1 == 0.0) ! Here is a comment on where
        arr2 = arr1 * 3
    elsewhere ! Here is a comment on elsewhere
        arr2 = 0
    end where ! Here is a comment on end where
    end subroutine test"""
```

then the fparser tree is:
```
[Where_Construct_Stmt(Level_4_Expr(Name('arr1'), '==', Real_Literal_Constant('0.0', None))), Comment('! Here is a comment on where'), Assignment_Stmt(Name('arr2'), '=', Add_Operand(Name('arr1'), '*', Int_Literal_Constant('3', None))), Elsewhere_Stmt('ELSEWHERE', None), Comment('! Here is a comment on elsewhere'), Assignment_Stmt(Name('arr2'), '=', Int_Literal_Constant('0', None)), End_Where_Stmt('WHERE', None)]
```

and the resulting output code from PSyclone is:
```
subroutine test()
  real, dimension(100) :: arr1
  real, dimension(100) :: arr2
  integer :: widx1

  arr1 = 0.0
  do widx1 = 1, 100, 1
    if (arr1(widx1) == 0.0) then
      ! Here is a comment on where
      arr2(widx1) = arr1(widx1) * 3
    else
      ! Here is a comment on elsewhere
      arr2(widx1) = 0
    end if
  enddo  ! Here is a comment on end where

end subroutine test
```

The comments on the where/elsewhere get moved down slightly to be preceding comments of the first element in their schedule (and if they're schedule is empty they are lost). I'll clean up this test and add it in.

LonelyCat124: Test (and compilation test) added.

LonelyCat124: Fixed and xfailing test  (no compliation because xfailing) added.

src/psyclone/psyir/frontend/fparser2.py

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/frontend/fparser2_where_handler_test.py
LonelyCat124: Added

LonelyCat124:
@arporter Should be ready for another look now I think.

> I will check the ITs next time around (although I don't think we test keep-comments yet?).
I don't think so, but the compilation tests don't run locally for me so the ITs will be needed.

arporter:
Thanks @LonelyCat124 - could you take a look at the CodeCov complaint for me?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -4697,24 +4695,35 @@ def _contains_intrinsic_reduction(pnodes):
...
+            if not isinstance(node.content[first_non_comment],
                               Fortran2003.Where_Construct_Stmt):
                 raise InternalError(f"Failed to find opening where construct "
                                     f"statement in: {node}")
             if not isinstance(node.content[-1], Fortran2003.End_Where_Stmt):
                 raise InternalError(f"Failed to find closing end where "
                                     f"statement in: {node}")
-            if _contains_intrinsic_reduction(node.content[1:-1]):
+            if _contains_intrinsic_reduction(
+                    node.content[first_non_comment:-1]
```
arporter: From looking at the uncovered line in CodeCov, I think this should be `first_non_comment+1:-1` (so that we recover the original behaviour when `first_non_comment` is 0).

arporter (changes requested):
Almost there now. I took the liberty of adding yet another comment to your new test, just to satisfy my curiosity.
The only thing remaining is a drop in coverage and that is due to a slight error in the code that checks for array reductions - see inline.
Once that's fixed, please could you fire off the ITs (assuming the machine has come back up)?

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Thanks - fixed.

LonelyCat124:
@arporter ITs are green, ready for another look.

arporter (approved):
All good now, thanks Aidan.

PR #3495, opened 2026-07-14
(towards #3207)  Update LFRic integration tests

sergisiso:
- Update to LFRic vn3.2
- Use Jules from repo instead of file
- Use https url by a config change instead of patching the dependencies.yaml
- Use full numerical reproducibility comparisons for the cases that work, and add TODOs when it does not.


sergisiso:
@arporter This is ready for review. It is still a 'towards PR' because not all targets fully match the baseline yet. Also, contrary to what we discussed I left the does_nothing.py because it does match the minimum script that lfric has (it was my mistake the previous mismatch)

examples/lfric/scripts/gpu_offloading.py
```diff
@@ -124,16 +124,16 @@ def trans(psyir):
...
         # Use OpenACC offloading
         loop_offloading_trans = ACCLoopTrans()
         kernels_trans = ACCKernelsTrans()
         gpu_region_trans = ACCParallelTrans(default_present=False)
         gpu_annotation_trans = ACCRoutineTrans()
     else:
         print(f"The PSyclone transformation script expects the "
-              f"LFRIC_OFFLOAD_DIRECTIVES to be set to 'omp' or 'acc' "
-              f"but found '{OFFLOAD_DIRECTIVES}'.")
+              f"LFRIC_OFFLOAD_DIRECTIVES to be set to 'omp', 'omp_uniform', "
```
arporter, line 135: I was expecting to see the new (?) `acc_uniform` and `omp_uniform` options used elsewhere or is that future work?

arporter (changes requested):
Thanks Sergi. All looks good and I see that this test only modifies the LFRic side of things and the corresponding IT succeeded yesterday.
Only issue is that you seem to have introduced new "omp/acc_uniform" options but the script doesn't actually use them. Is that deliberate?

examples/lfric/scripts/gpu_offloading.py
sergisiso: I should have explained the context better, they are set in .github/workflows/lfric_test.yml and used by lfric_core's nvfortran.mk (https://github.com/MetOffice/lfric_core/compare/main...stfc:lfric_core:psyclone_next)

It is a step towards trying to match the CPU results but not sufficient. From NEMOs experience we also need to avoid certain intrinsics but here is more complicated because they can happen inside infrastructure calls that psyclone cannot see. I will explore this on a separate PR.

sergisiso:
@arporter Ready for next review.

examples/lfric/scripts/gpu_offloading.py
arporter: Aha. Thanks.

arporter (approved):
All good now.

Issue #2905, opened 2025-02-20
Refactor PSyKAl workflow


sergisiso:
With #1010 we now have PSyIR nodes and PSyIR symbols throughout the LFRic workflow, which currently consists of:

1. parsing
2. generate psy-layer (parses kernel metadata and creates initial loops and calls)
3. transformation script (some just enable booleans, .e.g. colouring, distributed mem ...)
4. generate declarations
5. generate initialisations
6. lowering (includes argordering and second stage of some transfomrations)

This has 2 problem:
 - It is hard to deal with symbols, when operating with them it is unclear if we are at a point before or after each symbol is created, so we have "find_or_create" all over the place, duplicating the definition of the types. Having declarations and initialisation separated also means we have almost symetric implementations (all declarations but the loop variables need initialisations) however we need to do many lookups to find them.
 - It is hard to do transformations because if a previous transformation is just a boolean and not reflected in the PSyIR, the new transformation also needs to be a boolean to place it later, or worst, encode knowledge about all possible transformations and what changes the implicate for itself (which is not scalable with the number of transformations)

I propose simplifying the workflow to:

parsing
generate psy-layer (loops and kernels)
generate declarations + initialisations
transformation script (do it in-place when possible)
lowering

Changing the location of the declarations+inits is realtively simple, the big change here is in some transformations because they now have the responsibility of cleaning up the symbols they replace and add the new ones, including their initialisations (e.g. colourmaps). But as a benefit:
- they become more composable, they don't need to know about any other transformations, the PSyIR they recieve is the truth.
- they are better encapsulated: the symbols, initialisations, and changes they introduce are all in the same file, and not scattered through the code

Note: ArgOrdering also has duplications with "generate declarations", it does "find_or_create" again and interface setting again, I will leave this for a second stage.

@arporter @hiker @TeranIvy we have discussed this before, but any comments on this proposal? (the change won't come quick, specially moving the distributed memory logic inside a transformation could be hard, but I will start exploring this)

arporter:
All (?) PSy-layer symbols are now created when the `Invoke` is created. The most obvious blocker for #3398 remains the handling of loop bounds - at the moment we use labels for these that are replaced at lowering time. As Sergi says earlier, changing this is going to be involved because Transformations will have to manage the associated symbols and assignments, rather than delaying this to final code gen. Most obviously, this will affect the LFRic and GOcean colouring transformations.

[...]

Issue #3416, opened 2026-04-27
Issue to track dead code after Python 3.9 support is removed.


LonelyCat124:
Various tests involving treesitter are marked to be skipped pre 3.10.

LonelyCat124:
As I understand, PEP604 overrides PEP 484, so we should stop using `Optional[X]` as a typehint and instead use `X | None` after 3.9 support is concluded.

sergisiso:
Are you sure?, there is not deprecation warning for typing.Optional in the docs. And I (completely subjectively) like it more than the alternatives.

LonelyCat124:
That was the discussion I read online - `Optional` is somewhat misleading imo as a typehint also (as it doesn't imply the parameter is optional, merely that the input type is either the parameterised type hint or `None`, so I personally also find `int | None` clearer than `Optional[int]` in that regard.

Edit: Since `Optional[int]` and `Union[int, None]` are the same and the former is mostly a convenience, `int | None` is more clean and convenient I think.

sergisiso:
Ok, it's not a big deal to me but I think is important to be consistent, so I am happy to move to `int | None` once we drop 3.9.

PR #3499, opened 2026-07-20
(Towards #2905) mv lfric symbol creation

arporter:
Move the Symbol creation into the LFRicInvoke constructor. This then affects some tests that were having to pre-populate a table.


arporter:
A relatively simple PR that modernises some of the PSy-layer generation. 
The LFRic ITs have passed.
One for @sergisiso @LonelyCat124 or @hiker :-)

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -641,14 +641,16 @@ def fs_common(self, function_space,
...
-            self, function_space,
-            var_accesses: Optional[VariablesAccessMap] = None):
-        '''Add compulsory arguments associated with this function space to
+            self,
+            function_space: FunctionSpace,
+            var_accesses: Optional[VariablesAccessMap] = None
+    ) -> None:
+        '''
+        Add compulsory arguments associated with this function space to
         the list. If supplied it also stores this access in var_accesses.
```
LonelyCat124:
```suggestion
        this argument list. If supplied it also stores this access in var_accesses.
```

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -661,16 +663,7 @@ def fs_compulsory_field(
...
-        dtype = UnsupportedFortranType(
-            f"{intrinsic_type.intrinsic.name}("
-            f"kind={intrinsic_type.precision.name}), pointer, "
-            f"dimension(:,:) :: {map_name} => null()",
-            partial_datatype=ArrayType(
-                intrinsic_type,
-                [ArrayType.Extent.DEFERRED, ArrayType.Extent.DEFERRED]))
-        sym = self._symtab.find_or_create_tag(
-            map_name, symbol_type=DataSymbol, datatype=dtype)
+        sym = self._symtab.lookup_with_tag(map_name)
```
LonelyCat124, line 636:
```suggestion
        sym = self._symtab.lookup_with_tag(function_space.map_name)
```

src/psyclone/domain/lfric/kern_call_arg_list.py
```diff
@@ -661,16 +663,7 @@ def fs_compulsory_field(
         self.append(sym.name, var_accesses)
 
         map_name = function_space.map_name
```
LonelyCat124, line 635:
```suggestion

```

src/psyclone/domain/lfric/kern_call_arg_list.py
LonelyCat124: This can just be a single line now, I don't think the temporary is necessary for readability any more.

src/psyclone/domain/lfric/lfric_dofmaps.py
```diff
@@ -222,12 +222,16 @@ def invoke_declarations(self):
         super().invoke_declarations()
         # Function space dofmaps
         for dmap in sorted(self._unique_fs_maps):
-            if dmap not in self.symtab:
-                dmap_sym = DataSymbol(
-                    dmap, UnsupportedFortranType(
-                        f"integer(kind=i_def), pointer :: {dmap}(:,:) "
-                        f"=> null()"))
-                self.symtab.add(dmap_sym, tag=dmap)
+            intrinsic_type = LFRicTypes("LFRicIntegerScalarDataType")()
```
LonelyCat124: Certainly not one for this PR, but LFRicTypes being dependent on strings feels a bit messy somehow though I think it would require some significant rewriting and object-orientation to do much to.

src/psyclone/domain/lfric/lfric_dofmaps.py
```diff
@@ -222,12 +222,16 @@ def invoke_declarations(self):
...
+            dtype = UnsupportedFortranType(
+                f"{intrinsic_type.intrinsic.name.lower()}("
+                f"kind={intrinsic_type.precision.name}), pointer "
+                f":: {dmap}(:,:) => null()",
+                partial_datatype=ArrayType(
+                    intrinsic_type,
+                    [ArrayType.Extent.DEFERRED, ArrayType.Extent.DEFERRED]))
+            dmap_sym = self.symtab.find_or_create_tag(
+              dmap, symbol_type=DataSymbol, datatype=dtype)
```
LonelyCat124: Small thing - if we process code like this through the Frontend:
```
program x
  integer, pointer, dimension(:, :) :: a => null()


end program x
```

then the symbol we come across has an initial value set:
```
  a: DataSymbol<UnsupportedFortranType('INTEGER, POINTER, DIMENSION(:, :) :: a => null()'), Unknown, initial_value=IntrinsicCall[name='NULL']>
```

Could you add the initial value into this find_or_create_tag to match? If this causes issues though then we probably don't have to.

src/psyclone/domain/lfric/lfric_invoke.py
```diff
@@ -223,6 +223,19 @@ def __init__(self,
...
+                         self.fields, self.lma_ops,
+                         self.stencil, self.meshes,
+                         self.function_spaces, self.dofmaps, self.cma_ops,
+                         self.boundary_conditions, self.evaluators,
+                         self.halo_depths,
+                         self.proxies, self.cell_iterators,
+                         self.reference_element_properties,
+                         self.mesh_properties, self.loop_bounds,
+                         self.run_time_checks]:
+            entities.invoke_declarations()
```
LonelyCat124, line 205: I assume all LFRicCollections need to have their own implementation of `invoke_declarations`? Can you test making the one define in the `LFRicCollections` base class `abstractmethod` to ensure we can't miss it in subclasses?

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -839,27 +837,23 @@ def _is_scalar_parallelisable(self,
...
 
     # -------------------------------------------------------------------------
-    def can_loop_be_parallelised(self, loop,
-                                 test_all_variables=False,
-                                 signatures_to_ignore=None):
+    def can_loop_be_parallelised(
+            self,
+            loop: Loop,
+            test_all_variables: bool = False,
+            signatures_to_ignore: Optional[Signature] = None) -> bool:
```
LonelyCat124: I did a bit of looking at this, and its another thing we'll need to update post-3.9, `Optional` is a bit of a misnomer, which means "can be None" and not "optional". As of 3.10 we should use `Signature | None` for typehints like this, I added a comment to #3416

src/psyclone/psyir/tools/dependency_tools.py
```diff
@@ -839,27 +837,23 @@ def _is_scalar_parallelisable(self,
...
         :param loop: the loop node to be analysed.
-        :type loop: :py:class:`psyclone.psyir.nodes.Loop`
-        :param bool test_all_variables: if True, it will test if all variable
-                                        accesses can be parallelised,
-                                        otherwise it will stop after the first
-                                        variable is found that can not be
-                                        parallelised.
+        :param test_all_variables: if True, it will test if all variable
+            accesses can be parallelised, otherwise it will stop after the
+            first variable is found that can not be parallelised.
```
LonelyCat124: Can this docstirng be rewritten slightly to remove the need to specify "if True" - since its a bool I think we can do something like "whether to continue to test if all variable accesses can be parallelised once a variable has been found that can not be parallelised"?

src/psyclone/tests/core/variables_access_map_test.py
```diff
@@ -323,10 +323,10 @@ def test_variables_access_map_domain_loop() -> None:
     _, invoke = get_invoke("25.1_kern_two_domain.f90", "lfric", idx=0)
     vam = invoke.schedule.reference_accesses()
     assert str(vam) == (
-        "a: READ, b: READ, f1_data: READWRITE, f2_data: READWRITE, field_type:"
-        " CONSTANT, i_def: CONSTANT, map_w3: READ, ncell_2d_no_halos: READ, "
-        "ndf_w3: READ, nlayers_f1: READ, nlayers_f2: READ, r_def: CONSTANT, "
-        "undf_w3: READ")
+        "a: READ, b: READ, f1_data: READWRITE, f2_data: READWRITE, "
+        "field_proxy_type: CONSTANT, field_type: CONSTANT, i_def: CONSTANT, "
```
LonelyCat124, line 298: Just to make sure I understand - the change to these tests is because we now create the symbols at construction time, which means they exist when these tests are being checked (wheras before they would only have been created later in the psyclone chain, so weren't visible to these tests)?

LonelyCat124 (changes requested):
Hi @arporter I think I mostly understand this, have a couple of questions and a couple of small fixes requested, and one I'm not 100% on so I'd like to rerun ITs again after just in case.

src/psyclone/domain/lfric/kern_call_arg_list.py
arporter: Done.

src/psyclone/domain/lfric/kern_call_arg_list.py
arporter: If I've understood correctly then I can't do this: `map_name` is used twice below in calls to `append_array_reference`?

src/psyclone/domain/lfric/kern_call_arg_list.py
arporter: As previous.

src/psyclone/domain/lfric/lfric_dofmaps.py
arporter: Yes, LFRicTypes is a bit of an issue (e.g. #2659).

src/psyclone/domain/lfric/lfric_dofmaps.py
arporter: I started doing this (`initial_value = IntrinsicCall.create(Intrinsic.NULL)`) but then realised that will create a standard assignment rather than a pointer assignment (#2577). Therefore, I don't think we can do it currently but I'll add a TODO. I've also added partial_datatype to the other sorts of dofmap that this method creates since I'm here.

src/psyclone/domain/lfric/lfric_invoke.py
arporter: They do but the base one is not empty - it contains a check that the LFRicInvoke has been setup correctly.

src/psyclone/psyir/tools/dependency_tools.py
arporter: Thanks.

src/psyclone/psyir/tools/dependency_tools.py
arporter: Yes, done.

src/psyclone/tests/core/variables_access_map_test.py
arporter: Yes, that's right.

arporter:
Thanks @LonelyCat124, ready for another look now (CI permitting).

src/psyclone/domain/lfric/kern_call_arg_list.py
LonelyCat124: Ah yes, my mistake I missed those somehow.

src/psyclone/domain/lfric/lfric_dofmaps.py
LonelyCat124: See comment on Teams, I'd still like this to happen

src/psyclone/domain/lfric/lfric_invoke.py
LonelyCat124: Yeah, thats not an issue, you can still call `super().func()` up to an abstractmethod, see this example:
```
import abc

class base:

    @abc.abstractmethod
    def func(self):
        print("Hello from base")


class sub(base):
    def func(self):
        print("Hello from sub")
        super().func()

sub().func()
```

LonelyCat124 (changes requested):
I've left a couple of things still to be resolved from last time as I think they are both possible and would improve the code.

src/psyclone/domain/lfric/lfric_dofmaps.py
arporter: Done now - no issues so far.

src/psyclone/domain/lfric/lfric_invoke.py
arporter: Aha! Done now.

arporter:
Took me a while to get back to this but should be ready for another look now.

arporter:
Coverage cleaned up. I've set the ITs going again. Ready for another look @LonelyCat124 :-)

src/psyclone/domain/lfric/lfric_dofmaps.py
LonelyCat124: Can you still add the TODO so if/when we have support for pointer assignments in declarations we can easily find this code to update it appropriately?

LonelyCat124 (changes requested):
@arporter One more TODO I'd like adding and then we're good to merge (not sure if I can just add it when I update the changelog or whether I have to pass it back to you?) assuming ITs are green.

src/psyclone/domain/lfric/lfric_dofmaps.py
arporter: Good idea - done.

arporter:
Ready for another look now @LonelyCat124.

LonelyCat124 (approved):
Thanks Andy, all good now I think, will proceed to merge.

Issue #3460, opened 2026-06-18
Caught in infinite recursion


schreiberx:
Given this code:
```
MODULE timing
   IMPLICIT NONE
   PRIVATE

   TYPE timer
      TYPE(timer), POINTER :: s_next, s_prev, s_parent
   END TYPE timer

CONTAINS
   SUBROUTINE timing_finalize(sd_root)
      TYPE(timer), POINTER, INTENT(inout) :: sd_root
   END SUBROUTINE timing_finalize

END MODULE timing
```
Making a copy of the psyir tree with `psyir.copy()` results in an infinite recursion where I'm only posting the last few lines:
```
  File "/home/martin/workspace/neptune_REPOS/psyclone_REPOS/src/psyclone/psyir/symbols/datatypes.py", line 1470, in replace_symbols_using
    component.datatype.replace_symbols_using(table_or_symbol)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/martin/workspace/neptune_REPOS/psyclone_REPOS/src/psyclone/psyir/symbols/datatypes.py", line 315, in replace_symbols_using
    self.partial_datatype.replace_symbols_using(table_or_symbol)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/martin/workspace/neptune_REPOS/psyclone_REPOS/src/psyclone/psyir/symbols/data_type_symbol.py", line 157, in replace_symbols_using
    self.datatype.replace_symbols_using(table_or_symbol)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/martin/workspace/neptune_REPOS/psyclone_REPOS/src/psyclone/psyir/symbols/datatypes.py", line 1470, in replace_symbols_using
    component.datatype.replace_symbols_using(table_or_symbol)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/martin/workspace/neptune_REPOS/psyclone_REPOS/src/psyclone/psyir/symbols/datatypes.py", line 315, in replace_symbols_using
    self.partial_datatype.replace_symbols_using(table_or_symbol)
[...]

PR #3507, opened 2026-07-28
(Closes #3502) Allow intialisers for variables of a derived type

### LINKED ISSUE #3502: Derived types and initialisers
Consider the program:

```fortran
module pair
  type :: pair
    integer :: x, y
  end type
  type(pair) :: p = pair(1, 2)
end module
```

In the PSyIR symbol table for this module, the variable `p` is an `UnsupportedFortranType`. The reason is that the datatype of `p` does not pass the following check in the frontend:

  https://github.com/stfc/PSyclone/blob/2c736380cc11bd5138629e34734acc76342e794e/src/psyclone/psyir/frontend/fparser2.py#L2207-L2210

It's not clear to me whether the behaviour is intentional or not. If we wanted to allow `p` to have an initial value then we'd have to allow allow `DataTypeSymbol` in the check and also allow `Call` as a valid node inside the `DataSymbol.initial_value` setter. Would this be somehow bad?  It would permit some invalid Fortran, such as calling a function rather than a data constructor, but that could be potentially checked by the frontend or permitted under the assumption that the input code is valid fortran without leading to major issues?

I encountered this issue while trying to understand why the following program gives an error:

```fortran
module pair
  type :: pair
    integer :: x, y
  end type
  type(pair) :: ps(1) = [pair(1, 2)]
end module
```

This program passes the frontend check (because it's an array not a `DataTypeSymbol`) but then hits the lack of support for `Call` in  `DataSymbol.initial_value`. I'm trying to figure out what might be the right current solution for PSyclone:

1. allow `Call` nodes in initial values, or
2. disallow initial values for derived types.

mn416:
Closes #3502.

This PR allows variable declarations such as

```fortran
type(some_type) :: x = some_type(...)
```

to be parsed to a `DataSymbol` with a `DataTypeSymbol` type rather than an `UnsupportedFortranType`. It also allows

```fortran
type(some_type) :: x(1) = [some_type(...)]
```

which currently leads to a PSyclone internal error.

By relaxing what is allowed inside a variable initialiser, this PR allows both "good" code that was disallowed before but also "bad" code that is not valid Fortran. In particular, it allows arbitrary `Call` in an initialiser even though Fortran doesn't allow arbitrary functions to be called there.

Please note that there was a test `test_unsupported_decln_structure_type()` indicating that structures with initialisers were intentionally unsupported. I worry that the developer who wrote this test had a good reason why it was unsupported, but I don't know what that reason was/is.


LonelyCat124:
> Please note that there was a test test_unsupported_decln_structure_type() indicating that structures with initialisers were intentionally unsupported. I worry that the developer who wrote this test had a good reason why it was unsupported, but I don't know what that reason was/is.

Took me a bit of time to find the source of this, its #779 , so it looks like its just a way to enable unrecognised declarations (@arporter I don't know if you have any other memory if there was a reason to not support this other than just getting stuff working?)

Otherwise, this PR is missing some coverage indirectly as that test previously covered a `NotImplementedError` in the fparser2 frontend that is not covered. @mn416 Can you take a look - it may be now unreachable unless there is somehow a way to reach it with a StructureType maybe?

arporter:
> Took me a bit of time to find the source of this, its #779 , so it looks like its just a way to enable unrecognised declarations (@arporter I don't know if you have any other memory if there was a reason to not support this other than just getting stuff working?)

I think it was simply the fact that we hadn't got that far - it took a while for us to support derived types. Therefore I wouldn't worry about removing the test unless, for coverage sake, you need to change it so that it contains something that is still unsupported.

mn416:
Thanks @LonelyCat124 and @arporter. I think it's impossible to reach the body of this statement now:

```python
if not isinstance(
        datatype,
        (ScalarType, ArrayType, UnsupportedType,
                  DataTypeSymbol)):
    raise NotImplementedError
```

Looing at the code, `datatype` is whatever is returned by `_process_type_spec()` (possibly wrapped in an `ArrayType`) which, according to the type hints for that method, is `Union[ScalarType, DataTypeSymbol]`. In my latest commit, I've removed this statement. It's possible that there is another statement that is not longer reachable. Let's see what the coverage tool tells us.

mn416:
@LonelyCat124, looks like the coverage tool is happy now.

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -218,9 +218,9 @@ def initial_value(self, new_value):
         # pylint: disable=import-outside-toplevel
         from psyclone.psyir.nodes import (
             Assignment, Node, Literal, Operation, Reference,
-            CodeBlock, IntrinsicCall, ArrayConstructor)
-        from psyclone.psyir.symbols.datatypes import (ScalarType, ArrayType,
-                                                      UnsupportedType)
+            CodeBlock, Call, ArrayConstructor)
+        from psyclone.psyir.symbols.datatypes import (
+            ScalarType, ArrayType, UnsupportedType, DataTypeSymbol)
```
LonelyCat124:
```suggestion
            ScalarType, ArrayType, UnsupportedType, DataTypeSymbol
        )
```

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -229,24 +229,25 @@ def initial_value(self, new_value):
                     f"A DataSymbol with an ArgumentInterface can not have an "
                     f"initial value.")
             if not isinstance(self.datatype,
-                              (ScalarType, ArrayType, UnsupportedType)):
+                              (ScalarType, ArrayType,
+                               UnsupportedType, DataTypeSymbol)):
                 raise ValueError(
                     f"Error setting initial value for symbol '{self.name}'. "
                     f"A DataSymbol with an initial value must be a scalar or "
```
LonelyCat124:
```suggestion
                    f"A DataSymbol with an initial value must be a scalar, "
```

src/psyclone/psyir/symbols/datasymbol.py
```diff
@@ -229,24 +229,25 @@ def initial_value(self, new_value):
...
-                    f"an array or of UnsupportedType but found "
+                    f"an array or DataTypeSymbol or UnsupportedType but found "
```
LonelyCat124:
```suggestion
                    f"an array, a DataTypeSymbol, or an UnsupportedType but found "
```

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -944,9 +944,7 @@ def test_unsupported_decln_structure_type(fortran_reader):
     psyir = fortran_reader.psyir_from_source(code)
     routine = psyir.walk(Routine)[0]
     assert isinstance(routine.symbol_table.lookup('x').datatype,
-                      UnsupportedFortranType)
-    assert (routine.symbol_table.lookup('x').datatype.declaration ==
-            "TYPE(some_type), PARAMETER :: x = func()")
+                      DataTypeSymbol)
```
LonelyCat124, line 947: Can you add a similar test into the fortran backend to check we still get the correct output.

src/psyclone/tests/psyir/symbols/datasymbol_test.py
```diff
@@ -286,7 +286,8 @@ def test_datasymbol_initial_value_setter_invalid():
     with pytest.raises(ValueError) as error:
         sym.initial_value = 1.0
     assert ("Error setting initial value for symbol 'a'. A DataSymbol with an"
-            " initial value must be a scalar or an array or of UnsupportedType"
+            " initial value must be a scalar or an array or DataTypeSymbol"
+            " or UnsupportedType"
             " but found 'UnresolvedType'." in str(error.value))
```
LonelyCat124: Can you check if this will all fit on the line above instead of 2 lines?

src/psyclone/tests/psyir/frontend/fparser2_test.py
LonelyCat124: Also a frontend test for the case that was previously causing an InternalError but no longer does.

LonelyCat124 (changes requested):
Thanks @mn416 , I've asked for a couple of small tweaks to formatting and text and also a couple of additional tests.

src/psyclone/psyir/symbols/datasymbol.py
mn416: Done (bf34d14).

src/psyclone/psyir/symbols/datasymbol.py
mn416: Done (bf34d14).

src/psyclone/psyir/symbols/datasymbol.py
mn416: Done (bf34d14).

src/psyclone/tests/psyir/symbols/datasymbol_test.py
mn416: It doesn't quite fit, but I justified the text a bit (bf34d14).

src/psyclone/tests/psyir/frontend/fparser2_test.py
mn416: Done (5642913 and 477f243 respectively).

LonelyCat124:
Happy in principle with the implementation, however the ITs are all failing immediately for NEMO due to #3460 .
This will have to sit on hold for now until that is fixed @mn416  - we need to fix that issue at some point for the ENES-RISE project so we can try to prioritise it if anyone has free cycles @sergisiso @arporter ?

mn416:
Thanks @LonelyCat124, no rush from my point of view.

LonelyCat124:
ITs are now green after Sergi fixed the other issue

LonelyCat124 (approved):

PR #3510, opened 2026-07-29
(Closes #868) implement support for kernels with field arguments that specify nlayers/ndata

### LINKED ISSUE #868: Support custom function spaces in LFRic (multi-data fields, different number of layers and data)

arporter:


arporter:
This is very close to being ready. I need to finish doing a 'self review' and deal with the various TODOs, esp. where I'm using `LFRicTypes("MeshHeightDataSymbol")` for `ndata`.

arporter:
This is ready for a first review now. It's slightly bigger (in terms of files touched) than I would like because I realised that at some point during the last PR, I mixed 'nlevels' with 'nlayers'. I've gone through and fixed this so that we only use 'nlayers' now. I've also altered the naming used for *all* `nlayers` arguments and that modified a bunch of tests too.

arporter:
ITs were all green. @stevemullerworth and @christophermaynard it would be good if someone at the MO could also test this branch out to check that I haven't missed anything.

examples/lfric/README.md
```diff
@@ -390,6 +390,13 @@ the algorithm subroutine generated by this example:
  5. Run ``../bin/skeleton ./configuration.nml`` in the ``miniapps/skeleton/example`` directory.
 
 
+## Example 21: Kernels with arguments that specify ndata and/or nlayers
```
sergisiso, line 393: We need something similar in `doc/tutorials_and_examples/lfric_examples.rst`, although we really should unify both files.

examples/common.mk
```diff
@@ -66,6 +66,12 @@ ${NOTEBOOK_FILES}:
 
 notebook: ${NOTEBOOK_FILES}
 
+%.o: %.F90
+	$(F90) $(F90FLAGS) -c $<
+
+%.o: %.f90
+	$(F90) $(F90FLAGS) -c $<
+
```
sergisiso, line 74: Adding these to common.mk may eliminate them from many inner of the Makefiles. Grep returns lots of these, but I don't know if all of them import common.mk

src/psyclone/domain/lfric/kernel/field_arg_metadata.py
```diff
@@ -50,12 +50,12 @@ class FieldArgMetadata(ScalarArgMetadata):
 
     def __init__(self, datatype: str, access: str, function_space: str,
                  stencil: Optional[str] = None,
-                 nlevels: Optional[str] = None,
-                 ndata: str = "1"):
+                 nlayers: Optional[str] = None,
+                 ndata: Optional[str] = "1"):
```
sergisiso: Is this really optional (we can provide a None value) or is just a str with a default value.

The docstring of these 2 is also missing, and the types removed from the docstring.

src/psyclone/domain/lfric/function_space.py
```diff
@@ -34,9 +40,15 @@ class FunctionSpace():
     ## and field names.
     MAX_NAME_LEN = 21
 
-    def __init__(self, name: str, kernel_args: "LFRicKernelArguments"):
+    def __init__(self,
+                 name: str,
+                 kernel_args: "LFRicKernelArguments",
+                 nlayers: Optional[str] = None,
+                 ndata: Optional[str] = "1"):
```
sergisiso: Is it Optional[]?

src/psyclone/domain/lfric/function_space.py
```diff
@@ -76,20 +88,35 @@ def mangled_name(self) -> str:
...
+                raise FieldNotFoundError(
+                    f"No kernel argument found for function "
+                    f"space '{self._orig_name}'")
+
+        parts = [base_name]
+        if self._nlayers:
+            parts.append(self._nlayers)
+        if self._ndata != "1":
+            parts.append(self._ndata)
+        return self._shorten_name("_".join(parts))
```
sergisiso, line 119: [Codex comment] When fields share an original space but differ in nlayers or ndata, this code creates distinct mangled spaces, but FunctionSpace.field_on_space() still matches fields using only orig_name. Consequently LFRicDofmaps associates every variant with the first field on that original space—for example, the new test generates map_w2_shallow => f2_proxy%vspace%get_whole_dofmap() although shallow belongs to f3—and the kernel receives an incorrect dofmap. Field selection must compare the mangled space as well

src/psyclone/domain/lfric/lfric_cell_iterators.py
```diff
@@ -22,84 +27,148 @@ class LFRicCellIterators(LFRicCollection):
...
+        # We must also check for any subsequent arguments that have
+        # a number of layers or ndata specified by a label in the metadata.
+        for arg in kern.arguments.args:
+            if arg.nlayers and not arg.nlayers.isnumeric():
+                sym = self.symtab.find_or_create_tag(
+                    f"nlayers_{arg.nlayers}",
+                    symbol_type=LFRicTypes("MeshHeightDataSymbol"))
+                if sym.name not in self._nlayers_names:
+                    self._nlayers_names[sym.name] = arg
+                    _update_arg_properties(sym)
```
sergisiso, line 102: [Codex comment] If an invoke contains two calls to the same kernel with different actual fields, each metadata instance has the same named nlayers/ndata labels, but these label-only tags cause the second call to reuse the first call's symbol and map entry. Only the first field is queried, so both calls receive its runtime size even though separate actual fields need not have equal sizes; the key must distinguish calls or representative actual arguments while still deduplicating equal labels within one call.

src/psyclone/domain/lfric/kernel_interface.py
```diff
@@ -113,36 +112,66 @@ def generate(self, var_accesses=None):
...
+        :param var_accesses: an unused optional argument for interface
+                             consistency.
+        '''
+        for arg in self._kern.arguments.args:
+            if arg.ndata and not arg.ndata.isnumeric():
+                sym = self._symtab.find_or_create_tag(
+                    f"ndata_{arg.ndata}",
+                    symbol_type=LFRicTypes("NumberOfValuesPerDofDataSymbol"))
+                if sym not in self._arglist:
+                    self._arglist.append(sym)
```
sergisiso, line 174: [Codex comment] For any kernel with a named NDATA value, KernelInterface.generate() passes this symbol to SymbolTable.specify_argument_list() with its default AutomaticInterface, which raises ValueError because every listed argument must have an ArgumentInterface.

This makes LFRicKern.validate_kernel_code_args() fail before validation; create the symbol with the same read-only interface used for the named nlayers symbols.

src/psyclone/domain/lfric/function_space.py
```diff
@@ -76,20 +88,35 @@ def mangled_name(self) -> str:
...
+                if base_name:
+                    break
+            else:
+                # Raise an error if there are no kernel arguments on this
+                # function space
+                raise FieldNotFoundError(
+                    f"No kernel argument found for function "
+                    f"space '{self._orig_name}'")
+
+        parts = [base_name]
```
sergisiso, line 114: [Codex comment] The documented metadata permits arbitrary labels such as NLAYERS="some name", but appending the raw label produces invalid Fortran identifiers such as w3_some name, nlayers_some name, and corresponding dofmap names. PSyclone therefore emits unparsable code for a documented input; labels need a consistent, collision-free conversion to valid identifier components before being used in generated names.

sergisiso (changes requested):
@arporter I haven't reach the tests yet but see inline some comments from myself and Codex.

examples/lfric/README.md
arporter: Oh yes. And yes, we really should.

examples/common.mk
arporter: I've done all the ones I can in lfric. GOcean lacks a `gocean_common.mk` so would require more refactoring which I felt was going a bit far here.

src/psyclone/domain/lfric/kernel/field_arg_metadata.py
arporter: Oops, good spots.

src/psyclone/domain/lfric/function_space.py
arporter: No it isn't.

src/psyclone/domain/lfric/function_space.py
arporter: Impressive. Dead right. Code and test updated.

src/psyclone/domain/lfric/lfric_cell_iterators.py
arporter: I don't think this one is correct. If two different fields are on the same function space and have the same number of layers and `ndata` then they *will* be of the same size.

src/psyclone/domain/lfric/kernel_interface.py
arporter: That's also a good spot. However, obviously we don't see any failures so I need to check precisely why that is.

arporter: Test extended too.

src/psyclone/domain/lfric/function_space.py
arporter: Blimey, it's pretty good isn't it? I've updated the metadata validation to check that any label is itself a valid Fortran name. I've also updated the docs to specify this.

arporter:
Thanks Sergi (and Codex). Ready for another look now.

src/psyclone/domain/lfric/function_space.py
sergisiso: Thanks but Codex is still not happy with this, consider:

- For valid metadata containing one W2 field with nlayers='shape' and another with ndata='shape', both function spaces are mangled to w2_shape. Since unique function spaces, DoF maps, NDF and UNDF arguments are keyed by this value, one field's distinct layout is silently merged and its argument triple disappears from the kernel call.

I think space, nlayers, ndata should be mandatory parts? Maybe with a "" if not is given, so there will be an extra underscore to show the difference?

- Custom labels now pass through the lossy _shorten_name() operation, so distinct valid layouts can acquire the same identity; for example, W2 labels abcdefghijklmnopqr and azzzzzzzzzzzzzzzzr both shorten to w2_ar. This again merges their DoF-map and dimension arguments, so shortening must retain deterministic uniqueness, such as through a collision-resistant suffix.

That's also a good point, a better way would be to hash the long name, but that still can have rare collisitons. Maybe the best would be to store a global (or scopped) dictonary with the full and short names. There is also the possibility to use the long mangled names only as tags an let the symbol table manage naming. But that would require a lot of changes. I let to decide how deep you want to go to fix this.

src/psyclone/domain/lfric/lfric_cell_iterators.py
sergisiso: > If two different fields are on the same function space and have the same number of layers and ndata then they will be of the same size.

Can there be a kernel with ANY_SPACE and a NLAYERS or NDATA label? Then different invocations of the kernel may have different sizes?

Regardless, I also think there is a collision with how we use `f"nlayers_{first_arg.name}"`. For instance:  if the first algorithm field is named f1 and another field specifies nlayers='f1', both values use the tag nlayers_f1? Maybe this should be "labelled_nlayers_{arg.nlayers}"? I think ndata doesn't have a collision, but maybe we want to use the same naming.

sergisiso (changes requested):
@arporter Most of the PR looks good now but I believe there may still be name/tag collisions that are wrong. See inline comments.

arporter:
I'm working on switching to tags for robust naming of symbols in the PSy layer.

`lfric.py:qr_basis_alloc_args` currently returns a list of str for the arguments to an allocate.
This is only called from `lfric.py:LFRicBasisFunctions._basis_fn_declns`. This returns a list of dimensioning variable names and "a dict of basis arrays" - this is keyed by array names and entries are a list of the dimensionsing variables (str).
In all of these cases, it would be better if we returned lists of `Symbol` s instead.

arporter:
Am currently passing the suggested name for the basis/diff basis array as the first item in the list of its dimensioning variables. This is clearly bad but it lets me get something working. Once I have more tests passing I can look at improving it.

PR #3519, opened 2026-07-31
#1312 fix scalar array functionality

mo-alistairp:
The `gh_scalar_array` functionality is not working as intended. The generated kernel subroutine includes an array of the extents of the dimensions of the scalar array in the the list of arguments. This isn't being generated in the corresponding invoke call. As a result the mismatch in arguments causes build errors when `gh_scalar_array` is used.

This was likely because `kern_call_invoke_arg_list.py` was missed. This wasn't picked up by testing because the algorithm layer doesn't get compiled. As a way to verify that this functionality does work I'll also add an example using `gh_scalar_array` that gets compiled to ensure that it builds correctly. I'll probably also add a bit more documentation to aid users.


mo-alistairp:
This mistakenly has changes from another branch in it so I'm closing this PR and the correct changes are in #3520

Issue #1770, opened 2022-06-24
Make lfric eg20 compile


arporter:
Currently eg20 just generates an algorithm *subroutine* however @hiker says:

"I could get this to compile with our infrastructure files by adding function_space_collection_mod.F90 to our infrastructure (with a small change to get around code differences between our version and the current lfric infrastructure) and a dummy main program It certainly won't run, but just being able to compile this would be great."

so it would be good to do this.

arporter:
In #3520 I've added support for ScalarArrays and Sergi pointed out that I was generating un-compilable code. I thought I would prevent this from happening in future by adding compilation to eg20. However, it turns out that we generate (algorithm code) like:
```Fortran
    vector_space_w1_ptr => function_space_collection % get_fs(mesh, element_order_h, element_order_v, w1)
    vector_space_w2_ptr => function_space_collection % get_fs(mesh, element_order_h, element_order_v, w2)
    vector_space_w3_ptr => function_space_collection % get_fs(mesh, element_order_h, element_order_v, w3)
```
where we expect to be able to import `element_order_?` from `finite_element_config_mod`. However, this module no longer seems to exist. Therefore, the Algorithm layer we are generating is no longer consistent with the LFRic infrastructure. Presumably this means that the generated LFRic adjoint test harnesses are also wrong.

arporter:
I looked at the "simple diffusion" application in lfric_apps and that uses `components/driver/source/driver_fem_mod.f90` to setup its function spaces:
https://github.com/stfc/lfric_core/blob/dbd9fbda3fe89db3ad7e40fda3fea7889366c2a4/components/driver/source/driver_fem_mod.f90#L186-L208
where `this_coord_order` depends on:
`coord_order          = config%finite_element%coord_order()`
and:
https://github.com/stfc/lfric_core/blob/dbd9fbda3fe89db3ad7e40fda3fea7889366c2a4/components/driver/source/driver_fem_mod.f90#L59-L64
i.e. `config` is of `config_type` and is passed in.

arporter:
However, having said all of that, the element order values are only either 0 or 1 and, for compilation purposes, we don't care about the actual values. However, in the PSyAD test harness we should use the 'right' values but that probably requires new command-line flags.

hiker:
That file is generated by `core/infrastructure/build/tools/GenerateNamelistLoader`. We can see if we can add a copy to our version of the infrastructure.

hiker:
I copied 
```
[...]

PR #3520, opened 2026-07-31
(closes #1312) fix scalar array functionality

### LINKED ISSUE #1312: Pass read-only arrays of scalars to kernels

mo-alistairp:
The `gh_scalar_array` functionality is not working as intended (#1312). The generated kernel subroutine includes an array of the extents of the dimensions of the scalar array in the the list of arguments. This isn't being generated in the corresponding invoke call. As a result the mismatch in arguments causes build errors when `gh_scalar_array` is used.

This was likely because `kern_call_invoke_arg_list.py` was missed. This wasn't picked up by testing because the algorithm layer doesn't get compiled. As a way to verify that this functionality does work I'll also add an example using `gh_scalar_array` that gets compiled to ensure that it builds correctly. I'll probably also add a bit more documentation to aid users.


arporter:
I've taken this on from Alistair as he has moved roles.

This is ready for a first look now. It's quite LFRic heavy so one for @sergisiso or @hiker. (Although I'm now scared that Sergi will use Codex...) We now have an example of the functionality that compiles and links so are in much better shape.

@christophermaynard or @stevemullerworth - the one thing I omitted to ask Alistair for was a 'science' contact to help with testing of this functionality. Could you help here?

examples/lfric/eg1/Makefile
```diff
@@ -5,20 +5,53 @@
...
-transform:
-	@echo "Sequential code:"
-	${PSYCLONE} -api lfric -d ../code -nodm ./single_invoke.x90
+transform: single_invoke_alg.f90
 	@echo "With distributed-memory support:"
 	${PSYCLONE} -api lfric -d ../code ./single_invoke.x90
 
-compile: transform
-	@echo "No compilation supported for lfric/eg1"
+single_invoke_alg.f90 single_invoke_psy.f90: single_invoke.x90
```
sergisiso, line 31: The only comment from Codex was that modifying `../code/testkern_mod.F90` should retrigger this target as the psy-layer could be different. It is right (and very smart again, the kernel is not explicitly mentioned here so it knows what single_invoke.x90 have and that kernel metadata changes can produce changes in single_invoke_psy.f90 which is the output of this target).

However, this is a mistake that we have consistently across all examples and I would not explictly mention the specific kernel in the Makefile rule, which would fix the issue, but this would duplicate what is encoded inside the alg file. We could maybe mention the whole directory since it is explictly mentioned in the command, but make it worse the "-d" parameter is recursive. So maybe the right solution is something like https://stackoverflow.com/questions/25005637/makefile-rule-depend-on-directory-content-changes ?

I don't know. I will let you decide if its worth fixing, or leave it be, or just create a separate issue.

src/psyclone/domain/lfric/kern_call_invoke_arg_list.py
sergisiso, line 113: Apparently we don't keep a separate list of scalar_arrays arguments here, so Codex recons they still go through the scalar method and creates a scalar DataSymbol, ignoring scalar_arg._array_ndims

However the generated code in the output is correct. I am always confuse by this class but I cannot see scalar arrays handled anywhere. How come this doesn't matter?

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
```diff
@@ -578,32 +578,39 @@ def test_scalar_array(tmpdir, dist_mem):
     generated_code = str(psy.gen)
     expected_subroutine = (
        "  subroutine invoke_0(f1, real_array, logical_array, integer_array, "
-       "dims_integer_array, a, f2, f3, f4, b, dims_real_array, "
-       "dims_logical_array, dims_integer_array_1)\n"
+       "dims_integer_array, a, f2, f3, f4, b)\n"
```
sergisiso, line 581: This "dim_integer_array" here had me confused until I read the algorithm file which has:

```
  ! Include dims_integer_array as a scalar value to check that the
  ! generated code names do not clash
  call invoke(                                                                                       &
       testkern_scalar_array_type(f1, real_array, logical_array, integer_array, dims_integer_array), &
```
Maybe we can repeat the comment here to make clear why we recieve this name.

sergisiso (changes requested):
@arporter The changes look good and the generated code looks correct.

One problem I see is that the array_scalars handling is missing in some places for instance in: `src/psyclone/domain/lfric/kernel/meta_args_metadata.py:81` and the `kern_call_invoke_arg_list.py` that I mentioned in an inline comment. I am wondering now if this is a matter of the old/new metadata path. In which case I might prefer to keep the new path as is, since I have changed it substancially in another PR, but we may want to document the divergence.

examples/lfric/eg1/Makefile
arporter: Oh yes, that is impressive. I've followed that stackoverflow post as it's not a big change.

src/psyclone/domain/lfric/kern_call_invoke_arg_list.py
arporter: Codex was right (again). It seems we only use this in LFRicAlg which is used (in PSyAD) to construct algorithm-layer code. I've extended the appropriate test (which also showed that repeated calls to 'generate' were resulting in Symbols being repeatedly added to the table) and fixed it.

src/psyclone/tests/domain/lfric/lfric_scalar_codegen_test.py
arporter: Good idea, done.

arporter:
Ready for another look now @sergisiso. (It occurs to me we should possibly just get Codex to analyse e.g. symbol_table.py and check for edge case we have missed.)

sergisiso (changes requested):
@arporter All the previous comments have been adressed, I am running the ITs now but while they run I have a final question.

But if I do:
`psyclone-kern -api lfric -gen alg examples/lfric/code/testkern_mod.F90`
(or `LFRicAlg.create_from_kernel()`)

We get:
```
subroutine test_alg(mesh, chi, panel_id)
  ...
  real(kind=r_def), dimension(:,:) :: rscalar_array_2
  ...
  rscalar_array_2 = 1_i_def
  ..., testkern_type(rscalar_1, rscalar_array_2, field_3, field_4, field_5, field_6))
```

This won't compile as it is an assumed-shape which is not an argument. Can we give concrete shape (or make it allocatable?) to the `psyclone-kern -api lfric -gen alg` generated scalar_array?

arporter:
> This won't compile as it is an assumed-shape which is not an argument. Can we give concrete shape (or make it allocatable?) to the `psyclone-kern -api lfric -gen alg` generated scalar_array?

Good spot. I've arbitrarily decided to give each rank an extent of 3. I've updated the documentation and added a comment to the generated code to spell this out.

arporter:
Now that I write that, I should update eg20 to exercise this. I'll do that shortly...
...unfortunately I hit #1770. This must have been broken for a while and is bigger than this Issue so I propose deferring the problem.

arporter:
Therefore, this is ready for another look @sergisiso.

sergisiso (approved):
Thanks @arporter. Now all comments have been addressed and the IT pass (the last change in psyclone-kern is not covered by ITs), this is approved for merging.

PR #3542, opened 2026-08-05
Fixes lost comments on interface blocks with limitations due to fparser output.

LonelyCat124:
See fparser issue 521: https://github.com/stfc/fparser/issues/521

Fixes lost comments missed in the last PR. Still WIP as I'm now getting a duplicated interface for some reason.


LonelyCat124:
@sergisiso @arporter This is ready for a look now - there are limitations due to fparser that are discussed in comments.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2512,21 +2512,20 @@ def _process_parameter_stmts(self, nodes, parent):
                     # Ensure the interface to this Symbol is static
                     symbol.interface = StaticInterface()
 
-    def _process_interface_block(self, node, symbol_table, visibility_map):
+    def _process_interface_block(self, node: Fortran2003.Interface_Block,
+                                 symbol_table: SymbolTable,
+                                 visibility_map: dict[str, Symbol.Visibility],
+                                 preceding_comments: Iterable[str] = ()):
```
sergisiso, line 2478: The docstring param is missing

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
```diff
@@ -447,8 +449,12 @@ def test_interface_with_comments_is_supported(fortran_writer):
 
     correct = """module test
   implicit none
+  !Preceding comment
```
sergisiso: Can we preserve the space between the `!` and the comment?

sergisiso: Yes src/psyclone/psyir/backend/fortran.py:gen_interfacedecl should use _COMMENT_PREFIX

And it is important because without it, it could convert a `! $omp` into an `!$omp`, chaning the functionality.

src/psyclone/psyir/backend/fortran.py
```diff
@@ -696,17 +696,28 @@ def gen_interfacedecl(self, symbol):
             raise InternalError(
                 f"gen_interfacedecl only supports 'GenericInterfaceSymbol's "
                 f"but got '{type(symbol).__name__}'")
-
-        decln = f"{self._nindent}interface {symbol.name}\n"
+        decln = ""
+        if symbol.preceding_comment:
+            for line in symbol.preceding_comment.splitlines():
+                decln += f"{self._nindent}!{line}\n"
```
sergisiso: This pattern starts to be repeated a lot, but with slighly different conditions:
`if symbol.preceding_comment:`
`if symbol.preceding_comment != "":`
`if len(symbol.preceding_comment) > 0:`

It's just a few lines but I think this would be more redable is we make it 
`decln = self.gen_comments(symbol)`

In fact, I would even make it a `def gen_comments(obj: CommentableMixin)` and search for all `_COMMENT_PREFIX` or `preceding_comments` in the backend (symbols and nodes) and make sure we pass them all by this function, as I am not sure we apply this splitline consistently.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2583,6 +2589,10 @@ def _process_interface_block(self, node, symbol_table, visibility_map):
                             f"Expected '{rsym.name}' referenced by generic "
                             f"interface '{name}' to be a Symbol or a "
                             f"RoutineSymbol but found '{type(rsym).__name__}'")
+                    rsym.preceding_comment = self._comments_list_to_string(
+                            current_comments
+                    )
+                    current_comments = []
```
sergisiso: I find this a bit strange, we store the interface comment in the symbol, but the interface is not the declaration of the 'symbol'. First I thought maybe that's ok (with some comments explaining it) because the comments in the delcaration of the symbol as stored in the Routine node:
```
! Comment
suborutine mysub()
````

But I think there is a problem of overwriting it if we have multiple instances of the symbol (since this are not the declaration statements I belive it can be used several times?)

```
interface int1
    ! Comment 1
    procedure :: mysub
end interface
interface int2
    ! Comment 2
    procedure :: mysub
end interface
```

File: src/psyclone/psyir/frontend/fparser2.py [hunk 5]
```diff
@@ -2595,8 +2605,14 @@ def _process_interface_block(self, node, symbol_table, visibility_map):
...
                 # although there will be no corresponding implementation
                 # with that name.)
-                symbol_table.add(GenericInterfaceSymbol(
-                    name, rsymbols, visibility=vis))
+                interface_sym = GenericInterfaceSymbol(
+                    name, rsymbols, visibility=vis)
+                symbol_table.add(interface_sym)
+                interface_sym.preceding_comment = (
+                    self._comments_list_to_string(
+                        preceding_comments
```
sergisiso, line 2575: Should we also add it to the two unsupportedfortrantype in the fallback cases below?

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -242,9 +242,13 @@ def copy(self):
         # The constructors for all Symbol-based classes have 'name' as the
         # first positional argument.
         rt_info = [(rt.symbol, rt.from_container) for rt in self.routines]
-        return type(self)(self.name, rt_info,
+        copy = type(self)(self.name, rt_info,
                           visibility=self.visibility,
                           interface=self.interface.copy())
+        # Copy the preceding comment as well.
+        # Inline comments are not yet supported, see fparser issue #521.
```
sergisiso: Its the frontend/backend that don't support it, but does this really matter for the symbol?, it can already be:
`copy.inline_comment = self.inline_comment`

I wouldn't bother adding specific tests yet, but we don't need to leave it just partially implemented.

(same in the other similar statements in other symbol types)

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
```diff
@@ -462,6 +468,8 @@ def test_interface_with_comments_is_supported(fortran_writer):
 end module test"""
     out = fortran_writer(psyir)
     assert correct in out
-    assert "Some comment here" not in out
-    pytest.xfail(reason="TODO #3517 Comments in declarations are lost "
+    # We lost the inline comment at the end
+    assert "inline_comment" not in out
+    pytest.xfail(reason="Fparser #521 Inline comments in interface and "
+                        "module procedure declarations are out of order "
                         "inside Interface blocks.")
```
sergisiso: I would still use 3517, but with the new description and maybe textually saying that it is an fparser limitation.

Unfortunately referencing fparser issue like this now confuses the automatic closed tickets checking. If we really need to we can use the whole issue url.

sergisiso (changes requested):
@LonelyCat124 See comments inline, my main concern is if this overwrites other interfaces comments.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Added

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
LonelyCat124: Added, will probably be removed when refactoring soon.

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Did this for `gen_preceding_comments`. Most of the comments are generated by the visitor but I'm not sure if this is generalisable to be in the visitor in general? There are some restrictions in the implementation in the visitor itself which I'm not sure are necessarily valid (i.e. the valid_locations? I'm not sure where a preceding comment is not allowed?).

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: This doesn't work, I've removed comments inside interfaces with a comment in the frontend about this.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I should add it to the one that doesn't create a `_PSYCLONE_INTERNAL` at least.

src/psyclone/psyir/symbols/generic_interface_symbol.py
LonelyCat124: Sure, done.

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
LonelyCat124: Done.

LonelyCat124:
@sergisiso Comments addressed, ready for another look.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2674,7 +2698,9 @@ def process_declarations(self, parent, nodes, arg_list,
                         f"declarations not supported but found '{node}'")
             elif isinstance(node, Fortran2003.Interface_Block):
                 self._process_interface_block(node, parent.symbol_table,
-                                              visibility_map)
+                                              visibility_map,
+                                              preceding_comments)
+                preceding_comments = []
```
sergisiso: What happens with the big list of fparser nodes that we skip at line 2712. Would we accomulate their comments in the wrong place?

I am wondering if it would be better to change this loop to have:
```
for node in nodes:
  if is comment:
     process_comment
     continue
  <other nodes processing>
  #Flush the preceding comment buffer
  preceding_comments = []
```

src/psyclone/psyir/symbols/generic_interface_symbol.py
```diff
@@ -213,9 +213,15 @@ def copy(self):
...
-        return type(self)(self.name, rt_info,
+        copy = type(self)(self.name, rt_info,
                           visibility=self.visibility,
                           interface=self.interface.copy())
+        # Copy the preceding comment as well.
+        copy.preceding_comment = self.preceding_comment
+        # Inline comments are not yet supported by the backend but we can
+        # still copy them.
+        copy.inline_comment = self.inline_comment
+        return copy
```
sergisiso: Remove both comments, the code is already clear. (Also in all similar places)

src/psyclone/psyir/backend/fortran.py
```diff
@@ -249,6 +250,18 @@ def _reverse_map(reverse_dict, op_map):
             if mapping_key not in reverse_dict:
                 reverse_dict[mapping_key] = mapping_value.upper()
 
+    def gen_preceding_comments(self, obj: CommentableMixin) -> str:
```
sergisiso, line 253: obj param docstring missing

src/psyclone/psyir/backend/fortran.py
```diff
@@ -664,17 +675,18 @@ def gen_interfacedecl(self, symbol):
...
+        decln = ""
+        decln += self.gen_preceding_comments(symbol)
+        decln += f"{self._nindent}interface {symbol.name}\n"
         self._depth += 1
         # Any module procedures.
-        routines = ", ".join([rsym.name for rsym in symbol.container_routines])
-        if routines:
-            decln += f"{self._nindent}module procedure :: {routines}\n"
+        for routine in symbol.container_routines:
+            decln += self.gen_preceding_comments(routine)
```
sergisiso: Didn't we say that we can't use routine symbol comments for interfaces as this are repeated/overwritten (because the interface use is not the symbol declaration point)

src/psyclone/psyir/backend/fortran.py
```diff
@@ -664,17 +675,18 @@ def gen_interfacedecl(self, symbol):
...
+            decln += f"{self._nindent}module procedure :: {routine.name}\n"
         # Any other (external) procedures.
-        routines = ", ".join([rsym.name for rsym in symbol.external_routines])
-        if routines:
-            decln += f"{self._nindent}procedure :: {routines}\n"
+        for routine in symbol.external_routines:
+            decln += self.gen_preceding_comments(routine)
```
sergisiso: Same here?

src/psyclone/tests/psyir/backend/fortran_gen_decls_test.py
```diff
@@ -500,6 +500,34 @@ def test_gen_interfacedecl(fortran_writer):
 end interface subx
 ''')
 
+    # Check we get the preceding comment as expected
+    isub.preceding_comment = "Here is my preceding comment"
+    out = fortran_writer.gen_interfacedecl(isub)
+    assert (out == """! Here is my preceding comment
```
sergisiso: I prefer: 
```
= """\
! Here is
```

src/psyclone/tests/psyir/backend/fortran_gen_decls_test.py
```diff
@@ -500,6 +500,34 @@ def test_gen_interfacedecl(fortran_writer):
...
+interface subx
+  module procedure :: sub2
+  procedure :: sub1
+end interface subx
+""")
+    # Add a preceding comment to one of the routines.
```
sergisiso: I believe from here to the end of the test should be removed, see backend comment.

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
```diff
@@ -431,7 +435,36 @@ def test_interface_with_comments_is_supported(fortran_writer):
...
+    # We lost the inline comment at the end
+    assert "inline_comment" not in out
+    pytest.xfail(reason="#3517 Inline comments in interface and "
+                        "module procedure declarations are not supported "
+                        "inside Interface blocks. This is a limitation "
+                        "partly due to Fparser issue 521.")
+
+
+def test_unsupported_interface_keep_comment(fortran_writer):
+    ''' Test that an unsupported interface keeps the proceeding comments.'''
```
sergisiso: preceding

sergisiso (changes requested):
@LonelyCat124 see below some more doubts/suggestions.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Yeah I think this is probably better.

LonelyCat124: Ok, this also revealed an issue with how its currently implemented, since the comment isn't node the continue statement is currently ignored entirely, I think I can fix it with for/else.

LonelyCat124: Done.

src/psyclone/psyir/symbols/generic_interface_symbol.py
LonelyCat124: Done.

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Added

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Yes, I only removed the creation in the frontend, I'll remove it in the backend too.

src/psyclone/psyir/backend/fortran.py
LonelyCat124: Done.

src/psyclone/tests/psyir/backend/fortran_gen_decls_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/backend/fortran_gen_decls_test.py
LonelyCat124: Yeah, removed.

src/psyclone/tests/psyir/frontend/fparser2_interface_block_test.py
LonelyCat124: Fixed.

LonelyCat124:
@sergisiso Thanks, I think I've fixed all the issues you found - also realised the code was doing something badly when finding preceding comments, I think the new implementation makes more sense for what that section is meant to be doing.

LonelyCat124:
codecov is reporting missing coverage but also not showing any lines missing...
I'll try to run it locally

sergisiso:
image

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2704,11 +2730,15 @@ def process_declarations(self, parent, nodes, arg_list,
                     root_name="_PSYCLONE_INTERNAL_NAMELIST",
                     symbol_type=DataSymbol,
                     datatype=UnsupportedFortranType(str(node)))
-
+                # Reset preceding_comments to avoid placing comments
+                # on following nodes.
```
sergisiso: I think that's a leftover, the comment is already below

sergisiso (changes requested):
@LonelyCat124 There is just the missing coverage and the a small comment.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Removed.

LonelyCat124:
@sergisiso Yeah thanks, I also found the coverage, codecov was just initially displaying it weird - I assume this was an actual example of preceding_comments not being reset and making to where they shouldn't, as we no longer were hitting the line. I added a test for it now.

sergisiso (approved):
@LonelyCat124 I modified a bit the new test with direct asserts as I don't like frontend tests calling the backend. I also ported the linkcheck fix here as I am tired of seeing red CIs :)

This is approved for merging if ITs are green.

Issue #2554, opened 2024-04-22
InlineTrans does not accept a 'clash' between two identical COMMON blocks


arporter:
Hugo has reported:

## Common Block

* Common block can be well inlined if there is no conflict.

```python
code = """subroutine caller()
integer :: istr
integer :: iend

call foo(istr, iend)

end

subroutine foo(istr, iend)
integer :: istr
integer :: iend

call sub_foo(istr,iend)

end

subroutine sub_foo(istr,iend)
integer :: istr
integer :: iend
integer :: i
COMMON /ocean/ zeta
real zeta(10)

do i=istr, iend, 1
  zeta(i)=zeta(i+1)
enddo

end
"""
```

```python
def test_sub_foo_in_foo_and_inline_foo_in_caller():
    "Contains common block"
    psyir = FortranReader().psyir_from_source(code)
    inline_trans = InlineTrans()

    call = psyir.walk(Call)[1]
    inline_trans.apply(call)
    call = psyir.walk(Call)[0]
    inline_trans.apply(call)

    code_inlined = """subroutine caller()

  implicit none
  integer :: istr
  integer :: iend
  integer :: i
  real, dimension(10) :: zeta
  COMMON /ocean/ zeta

  do i = istr, iend, 1
    zeta(i) = zeta(i + 1)
  enddo

end subroutine caller"""

    [...]
```

* If there is a name clash between the Common block, the validation raises an exception.
```bash
PSyclone SymbolTable error: There is a name clash for symbol 'zeta' that cannot be resolved by renaming one of the instances because:
E                   - PSyclone SymbolTable error: Cannot rename symbol 'zeta' because it has a CommonBlock interface.
E                   - PSyclone SymbolTable error: Cannot rename symbol 'zeta' because it has a CommonBlock interface.

## code examples

### Common Block
```python

simple_code_with_common_clash = """subroutine caller()
integer :: istr
integer :: iend
integer :: i
real, dimension(10) :: zeta
common /ocean/ zeta
istr = 0
iend = 4
do i = istr, iend, 1
  zeta(i) = i
enddo
istr = 4
iend = 8
call foo(istr, iend)

end subroutine caller

subroutine foo(istr, iend)
integer :: istr
integer :: iend
integer :: i
real, dimension(10) :: zeta
common /ocean/ zeta

do i = istr, iend, 1
  zeta(i) = i + 1
enddo

end subroutine foo
"""

def test_simple_code_with_common_clash():
[...]

PR #3543, opened 2026-08-05
(closes #3392) Store commonblock name inside its interface

### LINKED ISSUE #3392: Extend CommonBlockInterface so that it holds the name of the common block
See https://github.com/stfc/PSyclone/pull/3389#pullrequestreview-4038078159

sergisiso:
This felt as a very natural way to implement the CommonBlocks, I don't know why it wasn't done like this in the first place.


sergisiso:
@arporter @LonelyCat124 This is ready for review, it is the ground work to then address https://github.com/stfc/PSyclone/issues/2554

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2925,22 +2924,16 @@ def _process_common_blocks(nodes, psyir_parent):
...
-                    root_name="_PSYCLONE_INTERNAL_COMMONBLOCK",
-                    symbol_type=DataSymbol,
-                    datatype=UnsupportedFortranType(str(node)))
-
                 # Get the names of the symbols accessed with the commonblock,
                 # they are already defined in the symbol table but they must
                 # now have a common-block interface.
                 try:
                     # Loop over every COMMON block defined in this Common_Stmt
                     for cb_object in node.children[0]:
```
LonelyCat124, line 2920: Since we've been having regular issues when comments appear on/before blocks we've not tested, can you have a go and see what happens if you have comemnts inline or immediately preceding a codeblock with `--keep-comments`? Am just increasingly wary of our hard coded `node.children[0]` code.

LonelyCat124: Also applies to cb_object[0] - probably need to double check in fparser that its always [0].

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -2904,10 +2904,9 @@ def _process_data_statements(nodes, psyir_parent):
 
     @staticmethod
     def _process_common_blocks(nodes, psyir_parent):
```
LonelyCat124: Can you add type hinting to this since it's being modified.

src/psyclone/tests/psyir/symbols/interfaces_test.py
```diff
@@ -91,12 +91,30 @@ def test_static_interface():
 
 
 def test_commonblockinterface():
-    '''Test we can create an CommonBlockInterface instance and check its
-    __str__ value
+    '''Test we can create an CommonBlockInterface instance and
+    __str__, __eq__, copy, and get its name.
```
LonelyCat124: "and tests __str__, __eq__, copy and can get its name"?

LonelyCat124 (changes requested):
@sergisiso This mostly makes sense and seems reasonable. I've asked for a couple of things, one is to just test behaviour regarding comments around/in common blocks declaration to check it doesn't also break since it uses hard coded indices.

src/psyclone/tests/psyir/symbols/interfaces_test.py
sergisiso: Rephrased

src/psyclone/psyir/frontend/fparser2.py
sergisiso: Done

src/psyclone/psyir/frontend/fparser2.py
sergisiso: See if you are happy with the new test.

I agree the `children[0]` doesn't look good and I was expecting it to fail but it didn't.

I checked if the received fparser had comment nodes and initially it didn't because it was not using the right fparser arguments because we were calling the reader directly (we have a lot of tests like this that need updating).

But after that I can see fparser comment nodes, but the test still passed (?). Is that what you expected to see?

sergisiso:

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Yeah I'm ok with the comments here being discarded, as I don't think there is anywhere for the frontend to put them for now. The comments would only be kept (I assume) if they were attached to the common block declaration by fparser itself. You could add the same 
```
if isisntance(node, Fortran2003.comment):
   self._process_comment(node, preceding_comments)
   continue
if isinstance(node, Fortran2003.Common_Stmt):
    ....
preceding_comments = []
```
code into `_process_common_blocks`, but I'll leave that decision up to you @sergisiso, from my point of view I'm happy to just lose comments on/around common blocks, as any implementation would also need to add `CommentableMixin` to CommonBlockInterface and support it in the backend.

LonelyCat124 (approved):

PR #3550, opened 2026-08-10
(closes #3544) Fix UnresolveInterface propagation issue, and add support for recursive

### LINKED ISSUE #3544: Symbols in a `CodeBlock` subroutine escape their scope
This issue was recently introduced in commit 34bcdaa.

Fortran code:

```fortran
module foo
contains
  recursive subroutine sub()
    use omp_lib
    integer :: x
  end subroutine
end module
```

Printing out the symbol table of module `foo` gives:

```
Symbol Table of Container 'foo':
--------------------------------
RoutineSymbol:
  sub: RoutineSymbol<NoType, pure=unknown, elemental=False>
Symbol:
  omp_lib: Symbol<Unresolved>
  x: Symbol<Unresolved>
```

Here, `omp_lib` and `x` have escaped their scope, seemingly due to `sub()` being converted to a `CodeBlock`.

sergisiso:
I probably made this PR have too many responsibilities, but I was not happy how things worked and implemented a few fixes while I was tackling this:
1) I consider the fronted resolve_imports to mean imports from external files. When there are two modules in the same file and we have all the information there already parsed, so there is not reason not to use it.
2) I added support for 'recursive', in modern Fortran it is just an optional hint, so there is no reason to fall over it.
3) I made the UnresolvedInterface symbols behave as private symbols. If we don't know where a symbol originates, it is not safe to propagate it to other modules (because there is no shadowing), while it is safe to propagate them upwards on the same module (because there is shadowing if we get it wrong, and we reserve the name to avoid future clashes).


sergisiso:
@LonelyCat124 This is ready for a first review.

src/psyclone/psyir/nodes/routine.py
```diff
@@ -45,6 +45,7 @@ class Routine(Schedule, CommentableMixin):
     # Textual description of the node.
     _children_valid_format = "[Statement]*"
     _text_name = "Routine"
+    valid_annotations = ("had_recursive_hint",)
```
LonelyCat124: @sergisiso Is there a reason to do this as an annotation as opposed to a symbol property like we do for other routine properties? We could have `is_recursive` as True/False/None as we do for pure? I suppose it less useful for the internals, but I'm concerned we're more likely to (unexpectedly) lose it as an annotation?

It also limits it's usability if someone specifically wanted to do something to things they know are explicitly declared as recursive (even though its an optional keyword).

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -6096,7 +6108,7 @@ def _module_handler(self, node, parent):
 
     def _program_handler(self,
                          node: Fortran2003.Program,
-                         parent: Node) -> Node:
+                         _parent: Node) -> Node:
```
LonelyCat124, line 6138: Why the rename? The argument is no longer used and the docstring isn't updated. If its no longer required is there a neater way to do this? If not just update the docs (maybe remove the param from the docstring entirely).

src/psyclone/tests/psyir/frontend/fparser2_test.py
```diff
@@ -1843,6 +1843,34 @@ def test_process_resolving_modules_give_correct_types(
...
+    prog = f2008_parser(reader)
+    processor = Fparser2Reader()
+    root = processor.generate_psyir(prog)
+    main = [m for m in root.children if m.name == "main"][0]
+    # The 'main' module has the local symbols
+    assert "test" in main.symbol_table
+    assert "b" in main.symbol_table
+    # And also the imported ones
+    assert "a" in main.symbol_table
+    assert "sub1" in main.symbol_table
```
LonelyCat124, line 1871: Does this fix the issue when we have multiple files and resolve_imports with the same symbol? If so, can we have a test for that as well so we can't break it again in the future?

LonelyCat124 (changes requested):
@sergisiso A couple of comments, I think we definitely at least need an additional test and some docstring fixes.

src/psyclone/psyir/nodes/routine.py
sergisiso: Ok, I will convert it to a property.

I first when for an annotation because I know in modern Fortran it is just an optional hint, losing or adding it does not change the behaviour/semantics and I thought this mapped well with the 'annotation' concept. But after looking at this a bit more I see that this change was only introduced in the F2018 standard, and losing it is meaningful in previous standards, so I agree that a property is the way to go for now.

src/psyclone/psyir/frontend/fparser2.py
sergisiso: The rename is because pylint complains about an UnusedArgument, instead of adding a pylint specific directive to disable the warning, the underscore seems to be the tool agnostic way to skip it. I updated the docstring.

The signature cannot really be changed because it is a handler, and they are called automatically with the two parameters.

LonelyCat124: Yeah, I had considered suggesting `*args` but that felt no better, so I'm happy with `_parent`, just needs the docstring updating.

src/psyclone/tests/psyir/frontend/fparser2_test.py
sergisiso: I will take this to mean the next test as this one is very focused on just testing that modules in filecontainers are connected. Correct me if I misunderstood what you meant.

For the next test, I now added the "modules in separate files" as a new case. They behave the same. In fact the reason why I made the other modifications to filecontainer is because before the behaviour was different depending on where the module was.

sergisiso:

src/psyclone/tests/psyir/frontend/fparser2_test.py
LonelyCat124: I think so yeah, I just didn't understand the test before.

LonelyCat124:
This is ok assuming ITs pass.

LonelyCat124 (approved):

PR #3551, opened 2026-08-10
(Closes #3549) Fixes issues with comments before labelled statement, where PSyclone would fail to recognise the label

### LINKED ISSUE #3549: Label for do loop dropped if ignore_comments=True
It seems that labels for do loops are dropped if parsing Fortran code with `ignore_comments=True`.

This can lead to other follow-up errors since the label is simply missing.

A reproducer can be found here:

[bug_reproducer.py](https://github.com/user-attachments/files/30854829/bug_reproducer.py)

Bringing @LonelyCat124 into the loop for this.

LonelyCat124:


LonelyCat124:
@sergisiso Ready for a first look.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -603,3 +603,61 @@ def test_directives_not_comments(directive):
     reader = FortranReader(ignore_comments=False)
     psyir = reader.psyir_from_source(code)
     assert directive not in psyir.debug_string()
+
+
+def test_labels_found_with_comments(fortran_reader, fortran_writer):
+    """ Test that the FortanReader correctly finds labels and produced a
+    CodeBlock when the labelled statement is preceded by comments."""
```
sergisiso: I am confused, the code below does not have comments? what is exactly being tested?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3190,13 +3190,21 @@ def _create_child(self, child, parent=None):
...
+            # there), so we have to examine the first non-comment statement
+            # within it. We must allow for the cases where the block is empty
+            # or only consists of comments though.
+            if child.content:
+                first_non_comment = 0
+                while (len(child.content) > first_non_comment and
+                       isinstance(child.content[first_non_comment],
+                                  Fortran2003.Comment)):
+                    first_non_comment = first_non_comment + 1
+            if (child.content and child.content[first_non_comment] and
```
sergisiso: I find this difficult to read, I would suggest (if I understand the purpose correctly):
```
  if isinstance(child, BlockBase):
      # Examine the first non-comment statement
      first_non_comment = next(
          (node for node in child.content
           if not isinstance(node, Fortran2003.Comment)),
          None)
      # If its starts with a label, the whole block is not supported
      if (first_non_comment is not None and
              not isinstance(first_non_comment, Fortran2003.Directive)
              and first_non_comment.item
              and first_non_comment.item.label):
          raise NotImplementedError("Unsupported labelled block")
```

and move the top `# We don't support statements with labels.` to the elif, as it just refers to that.

(I tested it and seems to work after fixing a couple of the updated error messages)

sergisiso (changes requested):
Thanks @LonelyCat124 but I am struggling to see what is happening, see if my updated code suggestion make sense and explain the test a bit more in its docstring.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Aha but it does! it has the secret comment, "\n\n" :P.

I just took this from Martin's reproducer, I'd like to leave the blank line "comment" as its a less obvious use case, but I will comment it better and explain better in the documentation.

LonelyCat124: Also added a test with a non-blank line comment.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Yeah, though I realised it should also check for Directives when ignoring nodes when looking for labels, so I've fixed that as well.

LonelyCat124:
@sergisiso This is ready for another look, hopefully clears up any confusion.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -3190,14 +3190,19 @@ def _create_child(self, child, parent=None):
...
-            # must allow for the case where the block is empty though.
-            if (child.content and child.content[0] and
-                    (not isinstance(child.content[0],
-                                    (Fortran2003.Comment,
-                                     Fortran2003.Directive))) and
-                    child.content[0].item and child.content[0].item.label):
-                raise NotImplementedError("Unsupported labelled statement")
+            # there), so we have to examine the first non-comment statement
+            # within it. We must allow for the cases where the block is empty
+            # or only consists of comments though.
```
sergisiso: of "comments or directives".

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -603,3 +603,119 @@ def test_directives_not_comments(directive):
...
+  logical :: cond2
+
+  if (cond1) then
+    a = 1
+  end if
+
+  ! PSyclone CodeBlock (unsupported code) reason:
+  !  - Unsupported labelled block
+  !  - Unsupported statement: Goto_Stmt\n  \n  100\
+ DO i = 1, 10
```
sergisiso, line 660: The preceding comment has merged the label '100' , that will be invalid code?

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -603,3 +603,119 @@ def test_directives_not_comments(directive):
...
+    code = """
+   subroutine foo()
+      integer :: i, j
+      integer :: a
+      logical :: cond1, cond2
+
+      if (cond1) then
+          a = 1
+      end if
+      !Comment line
```
sergisiso, line 684: Maybe also add a directive here (and the corresponding argument in the FortranReader below) since now they are accounted

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -603,3 +603,119 @@ def test_directives_not_comments(directive):
     reader = FortranReader(ignore_comments=False)
     psyir = reader.psyir_from_source(code)
     assert directive not in psyir.debug_string()
+
+
+def test_labels_found_with_comments(fortran_reader, fortran_writer):
```
sergisiso: Unused fortran_reader

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
```diff
@@ -603,3 +603,119 @@ def test_directives_not_comments(directive):
...
+    """ Test that the FortanReader correctly finds labels and produces a
+    CodeBlock when the labelled statement is preceded by comments, in this
+    test the comment in the first code segment is a blank line (which is
```
sergisiso: Maybe `in this test fparser returns a comment in the first code segment (as blank lines are parsed as Fortran2003.Comment("")).` ?

sergisiso (changes requested):
@LonelyCat124 Ah, I understand now, thanks. There is just one remaining concern with a test and then minor suggestions.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: It hasn't I don't think, I had some issue formatting this string, the result as it appears in the output is:
```
!  - Unsupported statement: Goto_Stmt
  
  100 DO i = 1, 10
```
I couldn't do this because i wasn't allowed the blank line with whitespace, nor trailing whitespace, so I did the above formatting to avoid the linter complaining. If you'd prefer I can add the comments that disable `flake8` from the errors instead.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Yeah, removed.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: Yeah, changed.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Yes, also non-comment, non-directive above.

LonelyCat124:
@sergisiso Comments addressed, ready for another look.

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
sergisiso: Aha, not its fine, I didn't understand it. I will just touch the comment above to be more specific about what are theformatting issue.

sergisiso: On second thought, why do we keep the empty line here. It seems in other places (e.g. if the do was not a codeblock) we discard these? Can we discard this as well?

sergisiso (changes requested):
@LonelyCat124 See last comment above

src/psyclone/tests/psyir/frontend/fparser2_comment_test.py
LonelyCat124: @sergisiso Not easily, I think this comes out of fparser itself inside the Labelled_Do_Loop_Block (or whatever the node type is called), so to remove it before making the codeblock we'd have to actually modify the fparser tree (since we have a CodeBlock here, the blank line isn't a preceding comment in PSyclone, but part of the CodeBlock ast), which I think isn't really PSyclone's job.

sergisiso: Ok, so just adjust the comment above this to make clear what the formatting issue is.

PS: I think you added the coverage commit here by accident

LonelyCat124: I did, I wasn't sure if i could revert it safely but I will do that now.

LonelyCat124: Fixed.

LonelyCat124:
@sergisiso Updated the comment to better explain and removed the coverage that is for the other PR.

sergisiso (approved):
@LonelyCat124 This is approved for merging, I will fix the conflict and update changelog but I will wait until the other "Ready to merge" PR is in first.

Issue #3538, opened 2026-08-04
Precision literals should be stored as Literal, not int


sergisiso:
This will add methods to work as if it where any other expression , e.g. copy(), without having to have edge cases.

mn416:
I noticed this issue while working on PR #3554.

```f90
subroutine sub()
  real(4) :: r
  r = r + 1.0_4
end subroutine
```

The value of `psyir.walk(Assignment)[0].rhs.datatype` is `Scalar<REAL, UNDEFINED>`. This is because the types of the arguments to `+` have slightly different representations:

* `Scalar<REAL, 4>`
* `Scalar<REAL, Literal[value:'4', Scalar<INTEGER, UNDEFINED>]>`

I think the source of the problem is that the `compute_precision()` function does not understand `Literal` precisions.

PR #3554, opened 2026-08-12
(Closes #1590) Add support for `complex` numbers

### LINKED ISSUE #1590: [PSyIR] Add support for complex (imaginary) numbers
Simon is using PSyclone to optimise the global-sum routine in NEMO and that routine uses complex arithmetic for handling numerical errors. AFAIK, this is the *only* routine in NEMO and MEDUSA that does so so this functionality is not high priority.

There would probably also need to be support for the `cmplx` intrinsic.

mn416:
Closes #1590.

This PR:

1. Adds a new `ScalarType.Intrinsic.COMPLEX` enumeration value.

2. Adds a new `ComplexLiteral` PSyIR node, as complex literals are not a good fit for the existing `Literal` class (they are a composite structure which can contain named constants, and are not suitable for representation as a textual string like other literal values).

3. Updates the frontend and backend to use/handle these new constructs.

4. Fixes all the tests that previously assumed `complex` types would lead to `UnsupportedFortranType`. In all cases I simply replace `complex` with `byte`.

5. Adds tests for the new functionality.


LonelyCat124:
Before reviewing I'll just set CI running in case it flags anything. I had a quick look to check that we hadn't missed anything obvious.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -576,9 +581,11 @@ def get_literal_precision(fparser2_node, psyir_literal_parent):
 
     :param fparser2_node: the fparser2 literal node.
     :type fparser2_node: :py:class:`Fortran2003.Real_Literal_Constant` or \
+        :py:class:`Fortran2003.Signed_Real_Literal_Constant` or \
```
LonelyCat124: Can remove the \ from this docs now. Also move the type information from the docstring to the declaration please :).

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -595,9 +602,11 @@ def get_literal_precision(fparser2_node, psyir_literal_parent):
     '''
     if not isinstance(fparser2_node,
                       (Fortran2003.Real_Literal_Constant,
+                       Fortran2003.Signed_Real_Literal_Constant,
                        Fortran2003.Logical_Literal_Constant,
                        Fortran2003.Char_Literal_Constant,
-                       Fortran2003.Int_Literal_Constant)):
+                       Fortran2003.Int_Literal_Constant,
+                       Fortran2003.Signed_Int_Literal_Constant)):
```
LonelyCat124, line 607: This is probably some lack of understanding from me - what are these `Signed_XXXX_Literal_Constant` classes representing from the tree (that we missed before?). I tried to see if we failed anywhere on these before but I didn't find anything.

Is it just something like `integer :: a = -1.0` would hit this (as an example)/

LonelyCat124: If this is now supported and you didn't already add a test to demonstrate this support could some be added?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5538,9 +5551,10 @@ def _number_handler(self, node, parent):
             # If the value has a "." without a digit before it then
             # add a "0" as the PSyIR does not allow this
             # format. e.g. +.3 => +0.3
-            if value[0] == "." or value[0:1] in ["+.", "-."]:
+            if value[0] == "." or value[0:2] in ["+.", "-."]:
```
LonelyCat124, line 5550: Good catch.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5605,6 +5619,35 @@ def _bool_literal_handler(self, node, parent):
             f"Expected to find '.true.' or '.false.' as fparser2 logical "
             f"literal, but found '{value}' instead.")
 
+    def _cmplx_literal_handler(self,
```
LonelyCat124: I think I'd slightly prefer `_complex_literal_handler` as the name here.

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+        if not isinstance(child, (Literal, Reference)):
+            return False
+        if (isinstance(child.datatype, (ScalarType, ArrayType)) and
+                child.datatype.intrinsic != ScalarType.Intrinsic.REAL):
+            return False
+        return True
+
+    def __eq__(self, other) -> bool:
+        is_eq = super().__eq__(other)
+        is_eq = is_eq and self.children == other.children
```
LonelyCat124: The superclass (on Node) equality already does this so we probably don't need to reimplement __eq__ other than maybe `super(DataNode).__eq__` (or similar) if we inherit from Literal.

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+from psyclone.psyir.nodes import Node
+from psyclone.psyir.nodes.datanode import DataNode
+from psyclone.psyir.nodes.literal import Literal
+from psyclone.psyir.nodes.reference import Reference
+from psyclone.psyir.symbols.symbol import Symbol
+from psyclone.psyir.symbols.symbol_table import SymbolTable
+from psyclone.psyir.symbols.datatypes import ScalarType, ArrayType, DataType
+
+
+class ComplexLiteral(DataNode):
```
LonelyCat124, line 22: I think probably the biggest concern I have with this implementation is the superclass choice (presumably due to some methods on Literal that don't work for ComplexLiteral?).

This wouldn't be a big deal, but I think that `.walk(Literal)` should probably find both types of `Literal`, so I think this needs to inherit from Literal.

I think value can be implemented to return the (fortran) repr of `(a, b)`. node_str should also (I think) contain a complex representation (even `x + y i` with some formatting on the `i` if possible? Not sure if possible though...) in the string. `value_as_python` should also be implementable.

The issue I think have is there may be places int he code that can't handle compelx numbered values, but we probably need to hit them and fix them?

Tagging @sergisiso and @arporter for their opinions as well.

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+        precision = compute_precision(precisions)
+        return ScalarType(ScalarType.Intrinsic.COMPLEX, precision)
+
+    def node_str(self, colour: bool = True) -> str:
+        '''
+        Construct a text representation of this node, optionally containing
+        colour control codes.
+        '''
+        return (f"{self.coloured_name(colour)}"
+                f"[datatype: {self.datatype}]")
```
LonelyCat124, line 95: See class comment.

LonelyCat124 (changes requested):
@mn416 I didn't get to reviewing all the tests, I have some concerns about whether or not `ComplexLiteral` should subclass `Literal`. I've asked for input from Sergi and Andy as they have time - I don't think we should change it just from my current thoughts as without the usecases for `.walk` hitting Litearl I'm not sure. Its perhaps enough to have some Mixin that both inherit (so they are generally separate but you can do `.walk(LiteralMixin)` if you really want them all.

src/psyclone/psyir/nodes/complex_literal.py
mn416: Hi @LonelyCat124, a few (minor) arguments against subclassing:

* The children of `ComplexLiteral` can be `Literal` so a subclass would naturally allow representation of nested `ComplexLiteral`, which is invalid Fortran. Of course, it could be explicitly disallowed (actually I already disallow non-`real` literals as children) but it feels a bit confusing.

* There would be no shared code between `Literal` and `ComplexLiteral`, as far as I can see.

* I'm not sure how `value_as_python()` would work for a complex literal containing a named constant, e.g. `(1.0, foo)`.

The alternatives are:

* Another new class `BaseLiteral`, which is a superclass of both.

* Write `.walk((Literal, ComplexLiteral))`.

Yeah, not conclusive arguments, just sharing my thinking.

src/psyclone/psyir/frontend/fparser2.py
mn416: I believe the `Signed_` nodes are only used in complex literals so weren't needed before this PR. In normal expressions, the unary `+` and `-` operators can be used to get signed literals. However, the components of a complex literal are not expressions, so we need a `Signed_` literal node to be able to capture `(-1.0, +2.0)`.

Yes, I will add a test for this.

src/psyclone/psyir/frontend/fparser2.py
mn416: Done (e9d69ad).

src/psyclone/psyir/frontend/fparser2.py
mn416: Test added/modified in 0e089ce.

src/psyclone/psyir/frontend/fparser2.py
mn416: Done (446a221).

src/psyclone/psyir/nodes/complex_literal.py
mn416: Done (ed8cae1).

mn416:
Thanks @LonelyCat124. I'll pass it back to "ready for review" while we decide what to do about the subclass question.

src/psyclone/psyir/nodes/complex_literal.py
sergisiso: I don't have a definitive answer but to expand on the alternatives:

If we expect `.walk(Literal)` to return them all we could rename the leaf literals rather then the base:
```
          |-> LeafLiteral
Literal --|
          |-> ComplexLiteral
```
          
If we don't expect `.walk(Literal)` to return compound literals we could make the PSyIR nomenclature diverge slightly from Fortran and call them `ComplexConstructor` (which would parallel ArrayConstructor). 

I suppose the Fortran standard distinguishes constructors vs literals based on whether their components must be constant, but PSyIR cannot fully guarantee that distinction. So the choice really comes down to what behaviour we want `.walk(Literal)` to imply and how close to the standard nomenclature we want to be.

Perhaps we can vote on it in the telco. @mn416 We usually meet Mondays at 8:30am (though the next two are sparse due to holidays). Would you be available or interested in joining occasionally?

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @property
+    def datatype(self) -> DataType:
+        from psyclone.psyir.tools.type_info_computation \
+            import compute_precision
+        precisions = []
+        for child in self.children:
+            if isinstance(child.datatype, (ScalarType, ArrayType)):
+                precisions.append(child.datatype.precision)
+        precision = compute_precision(precisions)
```
sergisiso, line 86: When both components are use-associated named constants, their `Reference` nodes have `UnresolvedType`, so no precision is appended and `compute_precision([])` raises `IndexError` during processing (for example, `use m, only: x, y; c = (x, y)`). Unknown component types should produce undefined precision rather than being omitted.

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+        '''
+        :param position: the position to be validated.
+        :param child: a child to be validated.
+        :return: whether the given child and position are valid for this node.
+        '''
+        if position < 0 or position >= 2:
+            return False
+        if not isinstance(child, (Literal, Reference)):
+            return False
+        if (isinstance(child.datatype, (ScalarType, ArrayType)) and
```
sergisiso, line 62: For a valid literal such as `c = (iparam, 1.0)`, `_name_handler` creates an integer Reference, which this validation rejects even though Fortran permits integer named constants as complex-literal components. The reader consequently raises `GenerationError` for valid source.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5605,6 +5618,35 @@ def _bool_literal_handler(self, node, parent):
...
+
+        :param node: node in fparser2 parse tree.
+        :param parent: parent node of the PSyIR node we are constructing.
+        :returns: PSyIR representation of node.
+        '''
+        # Convert real and imaginary parts to PSyIR nodes
+        lit = ComplexLiteral(parent=parent)
+        parts = []
+        for part in node.items:
+            if isinstance(part, Fortran2003.Name):
```
sergisiso, line 5634: When an integer component has an explicit kind, such as `(1_8, 2)`, this conversion creates a `REAL(kind=8)` child and the writer emits `(1.0_8, 2.0)`. That changes the complex value from default kind to kind 8 because an integer component's kind does not determine the resulting complex kind.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -744,8 +758,11 @@ class Intrinsic(IAttr, Enum):
                 types=DataNode,
                 arg_names=(("z",),)),
             optional_args={},
-            # TODO #1590 Complex numbers' precision unsupported.
-            return_type=lambda node: UnsupportedFortranType(""),
+            return_type=lambda node:
```
sergisiso, line 791: When `z` is a complex array, elemental `AIMAG(z)` is a real array of the same shape, but this helper always returns a scalar `ScalarType`. This gives incorrect datatype and rank information for valid array calls.

src/psyclone/psyir/nodes/intrinsic_call.py
```diff
@@ -1526,8 +1543,11 @@ class Intrinsic(IAttr, Enum):
                 types=DataNode,
                 arg_names=(("x",),)),
             optional_args={"y": DataNode, "kind": DataNode},
-            # TODO #1590 Complex numbers unsupported.
-            return_type=lambda node: UnsupportedFortranType(""),
+            return_type=lambda node:
```
sergisiso, line 1576: When either elemental argument to `CMPLX` is an array, the result is a complex array of the conforming shape, but `_type_of_scalar_with_optional_kind` always returns a scalar. For example,` c = CMPLX(real_array)` is therefore assigned an incorrect scalar datatype.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -404,7 +405,8 @@ class CharLengthParameter(ScalarTypeAttribute):
         Intrinsic.INTEGER: int,
         Intrinsic.CHARACTER: str,
         Intrinsic.BOOLEAN: bool,
-        Intrinsic.REAL: float}
+        Intrinsic.REAL: float,
+        Intrinsic.COMPLEX: Tuple[float, float]}
```
sergisiso: When a `DataSymbol` with complex datatype is initialized from a Python tuple, its setter passes this mapping to `isinstance`, but `typing.Tuple[float, float]` is a subscripted generic and raises `TypeError: Subscripted generics cannot be used with class and instance checks`. Complex native initializers therefore cannot use the newly added mapping.

src/psyclone/psyir/symbols/datatypes.py
```diff
@@ -666,6 +668,34 @@ def real8_type() -> "ScalarType":
...
+    @staticmethod
+    def complex_double_type() -> "ScalarType":
+        ''' :returns: a COMPLEX double scalartype '''
+        return ScalarType(ScalarType.Intrinsic.COMPLEX,
+                          ScalarType.Precision.DOUBLE)
+
+    @staticmethod
+    def complex4_type() -> "ScalarType":
+        ''' :returns: a COMPLEX 4-byte scalartype '''
+        return ScalarType(ScalarType.Intrinsic.COMPLEX, 4)
```
sergisiso, line 691: The new `complex4_type()` and `complex8_type()` constructors produce integer precisions, but `FortranWriter.gen_datatype()` rejects integer precision for complex, so symbols using either constructor fail with `VisitorError`. The same incompatibility affects parsed old-style `complex*8`/`complex*16` declarations that now enter the supported complex path.

sergisiso (changes requested):
@mn416 @LonelyCat124 I just copy pasted a Codex review, so take this critically and feel free to ignore them. But I think Codex is getting quite good at reviewing psyclone code.

src/psyclone/psyir/nodes/complex_literal.py
mn416: Thanks @sergisiso. Please do add me to the telco meeting, it would be nice to join occasionally (though my availability is also sporadic for the next month or so).

My vote would be for the `.walk(Literal)` to not return compound literals as the `.value` and `.value_as_python()` methods make less sense due to the fact that complex literals can contain named constants. It seems less likely to break existing code, which would keep this PR simple, and existing code can be adapted to support complex literals over time where desired. The important thing is that complex literals will no longer sit behind a `CodeBlock`.

I had a similar thought regarding complex literals versus complex constructors. My only concern about calling the class `ComplexConstructor` is that some methods assume the Fortran complex literal constraints, e.g. `reference_accesses()` says that all accesses have `CONSTANT` access type, and `.validate_child()` checks that children are `Literal` or `Reference`.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: This one definitely needs fixing - I missed it when looking at it myself.

src/psyclone/psyir/nodes/intrinsic_call.py
LonelyCat124: This one too - I missed that these are elemental.

src/psyclone/psyir/nodes/complex_literal.py
LonelyCat124: Yeah I think i'm leaning towards not subclassing now. I thought we walked for Literal (or Reference + Literal) more often than we do, and neither of the places we walk for Reference + Literal need to are about Complex types (since they're checking for strings) so I think I was overvaluing finding these alongside Literals, happy to leave this if everyone else agrees.

src/psyclone/psyir/nodes/complex_literal.py
mn416: Good point. I had assumed `compute_precision()` would handle the empty list. Fixed in 8620c5d.

src/psyclone/psyir/nodes/complex_literal.py
mn416: Good point. Fixed in 9c0c026.

src/psyclone/psyir/frontend/fparser2.py
mn416: That also seems right. Fixed in c8e8333.

src/psyclone/psyir/nodes/intrinsic_call.py
mn416: Yes, this seems to be a common problem across intrinsics using the `_type_of_intrinsic_with_precision_of_named_arg()` helper. I've modified this helper (in 019b63c) to also propagate the array shape of the argument type to the result type, if it is indeed an array.

src/psyclone/psyir/nodes/intrinsic_call.py
mn416: Fixed in c8b2702.

src/psyclone/psyir/symbols/datatypes.py
mn416: This line was leftover from a previous attempt to make complex literals part of the Literal class. I've removed it now as it doesn't make sense for complex literals (see 98f1c4c).

src/psyclone/psyir/symbols/datatypes.py
mn416: These new types just mirror the existing  `real4_type()` and `real8_type()` helpers, so I've left them alone. This sounds related to issue #3538.

mn416:
@LonelyCat124 @sergisiso I think I've addressed all your comments so far (thank you) and brought the branch up-to-date with master. I'll be around for the rest of the day but, after that, not until next Monday.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: There are still a number of \ in the docstring (e.g. :param psyir_literal_parent:) to be remove - just end each line with a newline and start the next one with 4 spaces indentation - the \ are outdated now.

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+    _text_name = "ComplexLiteral"
+    _colour = "yellow"
+
+    def __init__(self, parent: Node = None):
+        super().__init__(parent=parent)
+
+    @staticmethod
+    def create(re_part: Union[Literal, Reference],
+               im_part: Union[Literal, Reference],
+               parent: Optional[Node] = None):
```
LonelyCat124: -> ComplexLiteral

src/psyclone/psyir/nodes/complex_literal.py
```diff
@@ -0,0 +1,145 @@
...
+        dt = self.datatype
+        if isinstance(dt.precision, DataNode):
+            precision_symbols = dt.get_all_accessed_symbols()
+            for symbol in precision_symbols:
+                access_info.add_access(
+                    Signature(symbol.name), AccessType.CONSTANT, dt.precision)
+        return access_info
+
+    def replace_symbols_using(self,
+                              table_or_symbol: Union[SymbolTable, Symbol]):
```
LonelyCat124: -> None

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I'm not sure how the referenced commit contains the fix here, we probably should have a test but not sure.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: If its related to 3538 can we get an xfail test / TODO to demonstrate the issue here as well?

LonelyCat124 (changes requested):
Hi @mn416 A couple more things to sort, hoping this succesfully puts all my pending comments through...

src/psyclone/psyir/frontend/fparser2.py
mn416: Fixed (4a3ff929d9ff3c5bf003311b2cc0d20d58daa839).

src/psyclone/psyir/frontend/fparser2.py
mn416: The fix was to not convert integer literals to real literals inside complex literals in the frontend. After the fix, a `Literal` child of a `ComplexLiteral` could have `INTEGER` or `REAL` type (whereas before it was always `REAL`). The end result is that code is emitted in exactly the same format as it was parsed.

I added a test in commit b3eca289da3c411a37f3b38e97011b22634bfad7.

src/psyclone/psyir/symbols/datatypes.py
mn416: Apologies, I read the comment as a concern about the definition of `complex4_type()` when it was actually a concern about the definition of `FortranWriter.gen_datatype()`. Fixed in commit 5da2721e665d9ba4358d0502dd881d5a575f3398.

src/psyclone/psyir/nodes/complex_literal.py
mn416: Done (982be6d44469907721015df1bf1ed7677617c287).

src/psyclone/psyir/nodes/complex_literal.py
mn416: Done (982be6d44469907721015df1bf1ed7677617c287).

src/psyclone/psyir/symbols/datatypes.py
mn416: This commit poorly chosen test name and a typo. Fixed in 0f88e91fc8cd598c26ebfb6af84f1a47f4267107.

mn416:
Thank-you very much @LonelyCat124. I think I've addressed all outstanding comments. I'll pass it back to you.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: @mn416 Can you add a test for `(1_8, 2)` as well, and make sure to check the output kind is the same as well. I think it should be, would just be good to sanity check.

src/psyclone/psyir/symbols/datatypes.py
LonelyCat124: Maybe `test_fw_complex_int_kind`? Otherwise this seems good.

src/psyclone/psyir/nodes/complex_literal.py
LonelyCat124: Can you adjust the formatting to:
```
def replace_symbols_using(
    self, table_or_symbol: Union[SymbolTable, Symbol]
) -> None:
```

to avoid the need for the \?

LonelyCat124 (changes requested):
@mn416 Almost there, a couple more small changes.

src/psyclone/psyir/nodes/complex_literal.py
mn416: Done (b0b6ff431f157d60d743649d69e4d043f8a99bfe).

src/psyclone/psyir/frontend/fparser2.py
mn416: Good catch. This check uncovered a bug. Fixed in 3a6486893b0d44977fd78fe970f54c6a73f6ca9e.

src/psyclone/psyir/symbols/datatypes.py
mn416: Done (3538269b742b8cf1bd4a29e7546a1c62fd70ef2a).

mn416:
Thanks @LonelyCat124, another bug squashed. Back to you.

LonelyCat124:
I'm going to set the ITs going again. I was wondering whether we should have some compliation tests, but I didn't find anywhere appropriate for them so I think we can leave them for now - hopefully with NEMO having some complex requirements soon we can build that into the ITs.

LonelyCat124:
Its all green, I'll merge in the morning, feel free to go ahead if anything else is ready before.

LonelyCat124 (approved):

PR #3555, opened 2026-08-12
(Closes #3553) Moves `find_break_point` to use regex, and ensure comments will never cause line length failures.

### LINKED ISSUE #3553: NEMOv5 integration test with --keep-comments produces internal error with line_length transformation due to long comment containing table formatting
In NEMOv5 integration test (gfortran OpenMP for CPUs BENCH)

`./makenemo -r BENCH -m linux_spack -n ${TEST_DIR} add_key "key_nosignedzero" -j ${NUM_PARALLEL} -v 1` produces an internal error. This occurs in `psyclone.line_length.find_break_point()`, which cannot find a suitable break point in:

```
'+--------+--------------+----+----+----+----+-----+-------+-----+------+------+------+----+------------+----------------+--------' and keys '[' ', '.', ',']'
```
 causing the build to fail:

```
  File "/home/HCHPSE01/nxn02/vka83-nxn02/.psyclone_ENES-Rise/bin/psyclonefc", line 45, in <module>
    compiler_wrapper(sys.argv[1:])
  File "/home/HCHPSE01/nxn02/vka83-nxn02/PSyclone/src/psyclone/psyclonefc_cli.py", line 124, in compiler_wrapper
    main(psyclone_args)
  File "/home/HCHPSE01/nxn02/vka83-nxn02/PSyclone/src/psyclone/generator.py", line 724, in main
    code_transformation_mode(
  File "/home/HCHPSE01/nxn02/vka83-nxn02/PSyclone/src/psyclone/generator.py", line 969, in code_transformation_mode
    output = fll.process(output)
             ^^^^^^^^^^^^^^^^^^^
  File "/home/HCHPSE01/nxn02/vka83-nxn02/PSyclone/src/psyclone/line_length.py", line 180, in process
    break_point = find_break_point(
                  ^^^^^^^^^^^^^^^^^
  File "/home/HCHPSE01/nxn02/vka83-nxn02/PSyclone/src/psyclone/line_length.py", line 45, in find_break_point
    raise InternalError(
psyclone.errors.InternalError: PSyclone internal error: Error in find_break_point. No suitable break point found for line '+--------+--------------+----+----+----+----+-----+-------+-----+------+------+------+----+------------+----------------+--------' and keys '[' ', '.', ',']'
fcm_internal compile failed (256)
make: *** [/home/HCHPSE01/nxn02/vka83-nxn02/UKMO-NEMOv5/tests/11_08_BENCH_OMP_THREADING_GCC/BLD/Makefile:3068: tide_mod.o] Error 1
make -s -f /home/HCHPSE01/nxn02/vka83-nxn02/UKMO-NEMOv5/tests/11_08_BENCH_OMP_THREADING_GCC/BLD/Makefile -j 1 all failed (2) at /home/HCHPSE01/nxn02/vka83-nxn02/UKMO-NEMOv5/ext/FCM/bin/../lib/Fcm/Build.pm line 597.
```
This occurs in tide_mod.f90:

Image

LonelyCat124:


LonelyCat124:
@sergisiso This is ready for a look.

LonelyCat124:
I did just realise i could (instead of reversing the string) use findall and then rfind the last returned element, I'll leave it to the reviewer for the preferred method.

src/psyclone/tests/line_length_test.py
```diff
@@ -127,7 +148,7 @@ def test_fail_to_wrap():
     ''' Tests that we raise an error if we can't find anywhere to wrap
     the line'''
     input_file = "!$OMPPARALLELDO"
-    with pytest.raises(Exception) as excinfo:
+    with pytest.raises(InternalError) as excinfo:
```
sergisiso, line 156: Not yours, but there are a few print statements in this file that you could delete.

src/psyclone/line_length.py
```diff
@@ -38,10 +38,23 @@ def find_break_point(line: str, max_index: int, key_list: list[str]) -> int:
     # We should never break the line before the first element on the
     # line.
     first_non_whitespace = len(line) - len(line.lstrip())
+    # We create the appropriate substring and then index it backwards.
+    search_string = (line[first_non_whitespace+1:max_index])[::-1]
```
sergisiso: What happens if `first_non_whitespace+1 > max_index` ? Can you add a test?

src/psyclone/line_length.py
```diff
@@ -38,10 +38,23 @@ def find_break_point(line: str, max_index: int, key_list: list[str]) -> int:
...
     for key in key_list:
-        idx = line.rfind(key, first_non_whitespace+1, max_index)
-        if idx > 0:
-            return idx+len(key)
+        if isinstance(key, re.Pattern):
+            match = re.search(key, search_string)
```
sergisiso: How come we don't have to reverse this? I know they are precompiled re but would it be problematic for the ordered ones such as INTEGER, REAL, ... ?

> I did just realise I could (instead of reversing the string) use findall and then rfind the last returned element, I'll leave it to the reviewer for the preferred method.

Maybe not reversing will be easier? But whatever you prefer.

src/psyclone/line_length.py
```diff
@@ -38,10 +38,23 @@ def find_break_point(line: str, max_index: int, key_list: list[str]) -> int:
...
+            # For string input keys, we reverse them in case they are of
+            # length >1 as we are searching backwards.
+            match = re.search(re.escape(key[::-1]), search_string)
+        if match:
+            # If max_index is larger than the line length, we need to compute
+            # the matched index from the the end of the line.
+            end = min(max_index, len(line))
+            # Since we're working backwards we can return the start of the
+            # match (which is the end of the matched section of the forward
+            # string).
```
sergisiso: "the the"

Maybe this part will also be clear going forward instead of backwards?

sergisiso (changes requested):
Thanks @LonelyCat124, I think its worth exploring the version without reversing if it makes the code easier?

src/psyclone/line_length.py
sergisiso: The key_list typehint also needs updating to add the "re" expressions.

src/psyclone/line_length.py
LonelyCat124: We just get an empty search string so will always hit the internal error. I can add a test if there isn't one already.

LonelyCat124: Added

src/psyclone/tests/line_length_test.py
LonelyCat124: Removed.

src/psyclone/line_length.py
LonelyCat124: Done.

We didn't have to before as there was a note saying that any multi-length regex would need to be declared backwards (which is no longer needed)

LonelyCat124:
@sergisiso Should be ready for another look now, the new implementation is probably enough simpler to keep it, and I suspect probably isn't even more expensive since we don't have to reverse the loop and any input keys again.

src/psyclone/line_length.py
```diff
@@ -38,10 +40,20 @@ def find_break_point(line: str, max_index: int, key_list: list[str]) -> int:
     # We should never break the line before the first element on the
     # line.
     first_non_whitespace = len(line) - len(line.lstrip())
+
+    search_string = line[first_non_whitespace+1:max_index]
     for key in key_list:
-        idx = line.rfind(key, first_non_whitespace+1, max_index)
-        if idx > 0:
-            return idx+len(key)
```
sergisiso, line 44: I am not sure that I understand why this previous version was not working. I understand the branch to escape special characters, but why do we need to separate it with a findall and then a rfind of the same match?

sergisiso (changes requested):
@LonelyCat124 Just one more clarification.

src/psyclone/line_length.py
LonelyCat124: @sergisiso string `rfind` doesn't work with regex so we can't use it for the regular expressions we need to allow the more general behaviours introduced in this PR for comments (and also generalises this functionality). There is also no regex operation to match the last element in a string that I could find, so findall and rfind on the last match (e.g. you have some regex that matches `[a-z]` for the string `abc` would give you `['a', 'b', 'c']` from findall and then rfind on the last will find the 'c' at the end).

I could seperate it out to only do the re.findall when we get a `re.Pattern` input class if you'd like?

sergisiso: No that's ok, I just wanted to understand why this was needed.

sergisiso (approved):
@LonelyCat124 The last comment has been resolved and this passed the ITs yesterday. This is approved for merging.

PR #3560, opened 2026-08-17
(Closes #3367) statement next_accesses

LonelyCat124:


LonelyCat124:
I've not implemented the tests for this yet - I'm not sure if there should be any specific order to the result (or whether that should matter). I'm wondering if it should be sorted by abs_position or not.

I do think that for some blocks the way this has to be implemented at current will be prohibitively expensive, but the only solution to that will be a rewrite of the DUCs.

However, for the LFRic halo exchange infrastructure that requires this code I'm hopeful its not too expensive.

LonelyCat124:
Looking at #3399, the way I implemented this for Assignments was to return a dictionary of `Signature, List[Node]` (so multiple lists for each variable in the Assignment).

In this PR I thought that distinguishing between them is not so relevant, the caller should be worrying about the result, and next/previous_accesses should always return just `list[Node]` to be consistent regardless of node type (where supported).

@arporter @sergisiso Do you agree? If so I'll modify `Assignment` before I continue with this implementation (as its currently stuck in testing due to this).

sergisiso:
It would be good to have a reminder/plan about how we are going to use this to do the HEx with the DUC.

Does completing https://github.com/stfc/PSyclone/issues/3124 changes the need for these? For instance in the loop implementation you have `self.variable_reference.next_accesses() ...` but this should not be necessary anymore now that this is a children?

Are this methods equivalent to:
`accesses = {ref.next_accesses() for ref in self.walk(Reference)}`
?

LonelyCat124:
> It would be good to have a reminder/plan about how we are going to use this to do the HEx with the DUC.
> 
> Does completing #3124 changes the need for these? For instance in the loop implementation you have `self.variable_reference.next_accesses() ...` but this should not be necessary anymore now that this is a children?

I don't remember to be honest, I think we need Kern to be a `Call` and then it inherits next_access and should "just work" to reimplement `forward_dependence` and `backward_depedence` (that I understood are the functions used by the HEx?) to use this functionality.

I could just do `for child in loop.children` probably, I currently didn't consider the start/stop/step which is probably wrong, so I should fix that up.

> Are this methods equivalent to: `accesses = {ref.next_accesses() for ref in self.walk(Reference)}` ?

Not quite, it should be similar, but with node equality rules that wouldn't quite work I think. It also doesn't return any accesses contained inside the `Statement` itself, which for Loops would occur for accessing those references individually.


Swapping Assignment to just return a singular list was easy and didn't cause any side effects anywhere that I can see (as it was done in preperation for this PR I believe) so I've made that change and will continue to implement tests, as I think they'll make it clearer.

sergisiso:
> Not quite, it should be similar, but with node equality rules that wouldn't quite work I think. It also doesn't return any accesses contained inside the Statement itself, which for Loops would occur for accessing those references individually.

I'm a bit lost by this, could you show an example of an statement where it would have different results?

LonelyCat124:
> > Not quite, it should be similar, but with node equality rules that wouldn't quite work I think. It also doesn't return any accesses contained inside the Statement itself, which for Loops would occur for accessing those references individually.
> 
> I'm a bit lost by this, could you show an example of an statement where it would have different results?

```
do i = 1, 100
   j = 4
end do
j = 3
```

If you just do `psyir.walk(Assignment)[0].next_accesses()` you'll get a self-reference to `j=4` and then `j=3`. The loop.next_accesses() will just give the reference in `j=3`.

sergisiso:
> I don't remember to be honest, I think we need Kern to be a Call and then it inherits next_access and should "just work" to reimplement forward_dependence and backward_depedence (that I understood are the functions used by the HEx?) to use this functionality.

My worry is that next_accesses would return the ReadAfterRead without a way to differenciate them? In the Reference.next_accesses it is not a problem because we can do the is_read/is_write of both the location and target. But in a next_accesses of the statement we cannot do that.

> If you just do psyir.walk(Assignment)[0].next_accesses() you'll get a self-reference to j=4 and then j=3. The loop.next_accesses() will just give the reference in j=3.

Then maybe something like:
```
accesses = {ref.next_accesses() for ref in self.walk(Reference)}
accesses = filter(lambda x: not x.is_descendant_of(self), accesses)
```

LonelyCat124:
> > I don't remember to be honest, I think we need Kern to be a Call and then it inherits next_access and should "just work" to reimplement forward_dependence and backward_depedence (that I understood are the functions used by the HEx?) to use this functionality.
> 
> My worry is that next_accesses would return the ReadAfterRead without a way to differenciate them? In the Reference.next_accesses it is not a problem because we can do the is_read/is_write of both the location and target. But in a next_accesses of the statement we cannot do that.
> 
> > If you just do psyir.walk(Assignment)[0].next_accesses() you'll get a self-reference to j=4 and then j=3. The loop.next_accesses() will just give the reference in j=3.


I guess I don't remember if we have some way to check if a Statement has read/write access to a variable with reference_accesses anyway? I think thats sufficient to distinguish between WaR and RaR.

> 
> Then maybe something like:
> 
> ```
> accesses = {ref.next_accesses() for ref in self.walk(Reference)}
> accesses = filter(lambda x: not x.is_descendant_of(self), accesses)
> ```

Because nodes aren't hashable this doesn't work, we can't put them in sets.

LonelyCat124:
Also calling ref.next_accesses() repeatedly is currently inefficient (until the DUC rewrite) as assignment.next_accesses() is more efficient where possible, and I can probably do a similar trick for some of loop's nodes later.

LonelyCat124:
Been a bit slow with this sorry, a first implementation is ready for a first look @sergisiso @arporter . I do (at some point during the review process) want to look at potentially improving the performance of the loop implementation, and maybe look at some of the DSL nodes implementations (may need some advice there), but a first review to see how it works would be better first I think.

sergisiso (changes requested):
> accesses = {ref.next_accesses() for ref in self.walk(Reference)}
> accesses = filter(lambda x: not x.is_descendant_of(self), accesses)

> Because nodes aren't hashable this doesn't work, we can't put them in sets.

Sorry to linger on this, my concern is not so much the implementation, but that I don't see anything node-specific that requires that we have a separate implementation for each one*. In all of them we get all the references and then merge a unique list of next_accesses.

If I make the Statement.next_access be the following:
```
    def next_accesses(self) -> list[Node]:
        uniq_next = {}
        from psyclone.psyir.nodes import Reference
        for ref in self.walk(Reference):
            for next_ref in ref.next_accesses():
                if not next_ref.is_descendant_of(self):
                    uniq_next[id(next_ref)] = next_ref
        return list(uniq_next.values())
```
and then I remove the custom implementation in Call and Loop. I still have all their associated tests passing.

Is this not sufficient (and the same for the previous accesses)?

*I can see the Directive and PSyData nodes are special as they have `return []`. Is there a reason these don't behave as the other nodes? I believe they should behave as the other but even if you disagree, we could just reimplement these two with the empty return instead of all of them.

(the current implementation has problems with providing empty list to `DefinitionUseChain(refs)` as these generate IndexError, at least for calls and codeblocks `call foo(1)` or `print *, "hello"`. The proposed implementation bypasses this issue, but this could also be easily fixed in the current implementation)

LonelyCat124:
> > accesses = {ref.next_accesses() for ref in self.walk(Reference)}
> > accesses = filter(lambda x: not x.is_descendant_of(self), accesses)
> 
> > Because nodes aren't hashable this doesn't work, we can't put them in sets.
> 
> Sorry to linger on this, my concern is not so much the implementation, but that I don't see anything node-specific that requires that we have a separate implementation for each one*. In all of them we get all the references and then merge a unique list of next_accesses.
> 
> If I make the Statement.next_access be the following:
> 
> ```
>     def next_accesses(self) -> list[Node]:
>         uniq_next = {}
>         from psyclone.psyir.nodes import Reference
>         for ref in self.walk(Reference):
>             for next_ref in ref.next_accesses():
>                 if not next_ref.is_descendant_of(self):
>                     uniq_next[id(next_ref)] = next_ref
>         return list(uniq_next.values())
> ```
> 
> and then I remove the custom implementation in Call and Loop. I still have all their associated tests passing.
> 
> Is this not sufficient (and the same for the previous accesses)?
> 
> *I can see the Directive and PSyData nodes are special as they have `return []`. Is there a reason these don't behave as the other nodes? I believe they should behave as the other but even if you disagree, we could just reimplement these two with the empty return instead of all of them.
> 
> (the current implementation has problems with providing empty list to `DefinitionUseChain(refs)` as these generate IndexError, at least for calls and codeblocks `call foo(1)` or `print *, "hello"`. The proposed implementation bypasses this issue, but this could also be easily fixed in the current implementation)

@sergisiso The main reason at this stage is performance optimisation. You're implementation will call the (currently very expensive) DUCs once for every reference in the region. My implementation will minimise the calls to the DUCs as much as possible by merging multiple references into a single call  (e.g. assignment.next_accesses rather than reference.next_accesses N times) where possible (except maybe on loop variable/start/stop/step, as I'm not sure if that will work). If we can improve the DUCs to be a single expensive call and then reuse the results everywhere then your implementation is reasonable, assuming we did want all `Statement`s to have that implementation, but we are some time away from any DUC changes.

Regarding `Directive` and `PsyData` nodes, I was less sure. Directive I felt would be better avoided since their `dir_body` children (if they have them) will return the relevant information, and we need to avoid looking at their `Clause` children, and the DUCs ignore Directives anyway. Directives are also not exactly code with any side effects directly so they're a bit funny.

I also didn't see a use-case for PSyData `next/previous_accesses` so I left it empty (though at best it would just be:
```
for child in self.psy_data_body.children
    self._merge_accesses(...)
```
If I'm wrong and you think we would need that I can add it.

LonelyCat124:
It could be argued the base statement could just do:
```
for child in self.children:
   merge_accesses(..., children.next_accesses())
```
would be ok, but I still prefer having the abstract method on the abstract class here to force thought on what the result should be for each Statement class (and whether the default behaviour is appropriate).

sergisiso:
> My implementation will minimise the calls to the DUCs as much as possible by merging multiple references into a single call (e.g. assignment.next_accesses rather than reference.next_accesses N times) where possible (except maybe on loop variable/start/stop/step, as I'm not sure if that will work).

Aha, here may be the part I was missing. I saw that assignment.next_accesses() was still doing a walk for refs and calling with a list of all refs `DefinitionUseChain(refs)` and I thougth that was the same than doing `[DefinitionUseChain(ref) for ref in refs]`, but reading the DefinitionUseChain init I see that there is a check validating that they are all in the same Schedule. So I assume all if all the refs are in the same 'basic block' there are some simplifications that make the former cheaper? If this is the case it may be good to add this consideration in the ":param references:" as I don't think its obvious.

> Regarding Directive and PsyData nodes, I was less sure. Directive I felt would be better avoided since their dir_body children (if they have them) will return the relevant information

That depends on how you call it right? If you call it from flat list of statements (e.g. the children between two halo exchanges) it will never recurse down to the body, but if you call it from a .walk(statements) it would. I think my expectation is that a node.next_accesses recurses down itself to find its dependencies (some other nodes do). For example currently if we do a single directive node: `this_directive.next_accesses()` it returns nothing even if it has a body we many references that have forward accesses. I find this unexpected.

> It could be argued the base statement could just do: ...
> but I still prefer having the abstract method on the abstract class here to force thought on what the result should be for each Statement class (and whether the default behaviour is appropriate).

Could we attempt the base Statement implementation (ideally with Directive and PSyData) with the same behaviour. One of the reasons why we said dependencies belong to a separate tool: DependencyTools of DUC, was because it was quite difficult to understand by themself, and even worst with the implementation spread out in multiple files. If we don't have a consistent behaviour here then we will have part of the logic spread out and part of it in a complex file. And I already find quite hard to follow the dependencies logic as it is. A unified behaviour would make it much simpler.

> > (the current implementation has problems with providing empty list to DefinitionUseChain(refs) as these generate IndexError, at least for calls and codeblocks call foo(1) or print *, "hello". The proposed implementation bypasses this issue, but this could also be easily fixed in the current implementation)

I still think this is a problem, can you add tests for these.

LonelyCat124:
>    Regarding Directive and PsyData nodes, I was less sure. Directive I felt would be better avoided since their dir_body children (if they have them) will return the relevant information

> That depends on how you call it right? If you call it from flat list of statements (e.g. the children between two halo exchanges) it will never recurse down to the body, but if you call it from a .walk(statements) it would. I think my expectation is that a node.next_accesses recurses down itself to find its dependencies (some other nodes do). For example currently if we do a single directive node: this_directive.next_accesses() it returns nothing even if it has a body we many references that have forward accesses. I find this unexpected.

Yes, this is actually possibly an issue for a loop containing a directive I hadn't thought about, if you had a loop containing a `RegionDirective` as a child and did next_accesses on the loop you would miss any next accesses from the nodes contained in the RegionDirective - I will leave the default Directive behaviour (to cover standalone directives) but I will need to do an implementation for RegionDirective.


>        (the current implementation has problems with providing empty list to DefinitionUseChain(refs) as these generate IndexError, at least for calls and codeblocks call foo(1) or print *, "hello". The proposed implementation bypasses this issue, but this could also be easily fixed in the current implementation)

> I still think this is a problem, can you add tests for these.

I missed this comment before, this is a major issue yes I will take a a look.

LonelyCat124:
> Aha, here may be the part I was missing. I saw that assignment.next_accesses() was still doing a walk for refs and calling with a list of all refs DefinitionUseChain(refs) and I thougth that was the same than doing [DefinitionUseChain(ref) for ref in refs], but reading the DefinitionUseChain init I see that there is a check validating that they are all in the same Schedule. So I assume all if all the refs are in the same 'basic block' there are some simplifications that make the former cheaper? If this is the case it may be good to add this consideration in the ":param references:" as I don't think its obvious.

Also its less a case of its "cheaper" to do the DUC call, but you should only need to do 1 DUC call and thus 1 tree walk rather than 1 for each. I've not measured the gain, its mostly algorithmic theory at this point, since the cost of the DUCs is mostly traversing the tree my assumption has been its cheaper. I'll check this now if I can.

LonelyCat124:
Put together a simple test to check this assumption:
```
from psyclone.psyir.tools import DefinitionUseChain
from psyclone.psyir.nodes import Assignment, Reference

import time

code = """subroutine test
integer :: i, j, k, l
logical :: cond
j = 1
k = 3
l = 4
do i = 1, 100
   j = k + l
end do
if(cond) then
   l = 4
else
   k = 4
end if
i = j
do i = 1, 100
   j = j + i
end do
k = l * j
j = j - l
end subroutine test"""

psyir = FortranReader().psyir_from_source(code)

assigns = psyir.walk(Assignment)

start = assigns[3]

# Preload/setup all the infrastructure needed for the DUCs.
start.next_accesses()

start_t = time.time()
for i in range(10000):
    for ref in start.walk(Reference):
        chain = DefinitionUseChain(ref)
        x = chain.find_forward_accesses()
end = time.time()
print(f"Individual runs: {end-start_t}s")

start_t = time.time()
for i in range(10000):
    refs = []
    for ref in start.walk(Reference):
        refs.append(ref)
    chain = DefinitionUseChain(refs)
    x = chain.find_forward_accesses()
end = time.time()
print(f"Combined runs: {end-start_t}s")
```

This doesn't even account for any cost of merging the results from multiple DUCs that the first case would require. The results on my laptop felt fairly conclusive to me:
```
(base-env) aidan@DIN-43888:~$ python test2.py
Individual runs: 6.930823802947998s
Combined runs: 3.044456720352173s
(base-env) aidan@DIN-43888:~$ python -OO test2.py
Individual runs: 7.020304918289185s
Combined runs: 3.136359453201294s
(base-env) aidan@DIN-43888:~$ python -OO test2.py
Individual runs: 6.967965126037598s
Combined runs: 3.1583425998687744s
(base-env) aidan@DIN-43888:~$ python test2.py
Individual runs: 6.992641448974609s
Combined runs: 3.089935064315796s
```

It also furthers my assumption that a single (reusable) call to the DUCs for an entire routine would probably be best for most use-cases, however we're a long way from that right now.

(Also I had a bug in the initial test here and only computed next_accesses for the last reference in the first case, in which case it took 2.67s to only create the chains and execute the last one as opposed to 3.08s for the combined one, which suggests that addition additional references is a fraction of the basic cost).

The last thing I tried is just adding a bunch more (m to r) variables to the assignment, and also some assignments to those variables at the end to just a constant. The resulting runtimes were:
```
(base-env) aidan@DIN-43888:~$ python test2.py
Individual runs: 28.45047950744629s
Combined runs: 8.506452798843384s
(base-env) aidan@DIN-43888:~$ python test2.py
Individual runs: 29.341212272644043s
Combined runs: 8.31400465965271s
```

I may make a discussion thread about DUC performance, as it would be good to know what is driving the various costs here.

LonelyCat124:
@sergisiso Not everything is fixed (far from it) but can you look at the new implementations of Statement and Loop and let me know if you're ahppy with these changes? If so, I'll continue with those changes.

LonelyCat124:
@sergisiso I think this is ready for a first proper look now.

src/psyclone/psyir/nodes/routine.py
```diff
@@ -567,6 +567,20 @@ def replace_with(self, node, keep_name_in_context=True):
                 "replaced.")
         super().replace_with(node, keep_name_in_context=keep_name_in_context)
 
+    def next_accesses(self) -> list[Node]:
```
sergisiso: That's unexpected, Routine does not subclass Statement. Why was this needed?

sergisiso (changes requested):
Sorry for the slow progress on reviewing this @LonelyCat124 , I'm fine with the finer details of the implementation but my main problem is that I don't know what to decide about:

- Having two different behaviours of these methods, some find their children's references (with their own logic for performance reasons) and some don't (they simply return [] even when they have children). If it were up to me I would prefer a single behaviour. But without use cases its difficult to say what will be useful in practice.

- I don't particularly like that these are abstract methods but many of their implementations are just `return super().previous_accesses()`. I think I would make them not abstract, and delete all the empty implementations. But again this makes more sense with a single behaviour. I think you decided to use an abstract method to force each node to choose which behaviour they have? But this adds lots of verbosity.

Maybe it would be good to have a third opinion from @arporter about these points. I'm happy to compromise if needed so that this PR doesn't remain paused any longer.

src/psyclone/psyir/nodes/assignment.py
```diff
@@ -229,9 +229,15 @@ def previous_accesses(self) -> dict[Signature, list[Node]]:
...
         from psyclone.psyir.tools import DefinitionUseChain
         chain = DefinitionUseChain(refs)
-        return chain.find_backward_accesses()
-
-    def next_accesses(self) -> dict[Signature, list[Node]]:
+        accesses = chain.find_backward_accesses()
+        results = []
+        for sig in accesses:
+            for access in accesses[sig]:
+                if all(x is not access for x in results):
```
arporter, line 236: A bit nit-picky but, since I'm looking, is there a reason this can't be `if access not in results`?

LonelyCat124:
> Having two different behaviours of these methods, some find their children's references (with their own logic for performance reasons) and some don't (they simply return [] even when they have children). If it were up to me I would prefer a single behaviour. But without use cases its difficult to say what will be useful in practice.

Some of the domain-specific nodes I wasn't sure whether it made sense to look at their children or not, so my initial implementation was left as `[]`, whilst some need to specifically avoid their children (e.g. `StandaloneDirective`). I suspect the Kern implementation can either be improved or is wrong, but I felt that might be a decision better made later when we have a more concrete plan for how they should be integrated fully with PSyIR functionality (e..g Kern subclassing Call). I'm also never quite sure if the Kern can use the domain-specific knowledge or not?

I went for an abstractmethod as a defensive programming measure mostly - it prevents the creation of Statement subclasses without this method being implemented and forces a specific decision to be made about their behaviour, even if that behaviour is just to use the parent class' implementation.

src/psyclone/psyir/nodes/assignment.py
LonelyCat124: Yeah - `access` is a node (likely `Reference`) and these checks need to use `is` not `==` for `Node`s, and `in` checks equality not identity.

arporter: I realise (from reading the discussion on the PR) that `Node`s aren't hashable and we've overridden the `__eq__` method so this does have to be done this way.

arporter: (Sorry Aidan, you beat me to it.)

src/psyclone/psyir/nodes/while_loop.py
```diff
@@ -125,3 +127,55 @@ def reference_accesses(self) -> VariablesAccessMap:
         var_accesses = self.condition.reference_accesses()
         var_accesses.update(self.loop_body.reference_accesses())
         return var_accesses
+
+    def next_accesses(self) -> list[Node]:
+        '''
+        :returns: the combined next_accesses for the children of this
+            WhileLoop
+        '''
+        next_accesses = self._get_next_accesses(self.loop_body[:])
```
arporter, line 136: From my reading of the discussion, this can't be as simple as `self.children[:]` because we'd end up with accesses within the loop body?
That said, what's the reason that `Loop` and `WhileLoop` have distinct implementations?

src/psyclone/psyGen.py
```diff
@@ -1131,6 +1147,22 @@ def lower_to_language_level(self):
 
         return super().lower_to_language_level()
 
+    def next_accesses(self) -> list[Node]:
+        '''
+        next_accesses on base Kern class returns an empty list.
+
+        :returns: an empty list.
+        '''
+        return []
```
arporter, line 1156: In the existing PSy-layer DA, these lists comprise `Kern` and `HaloExchange` nodes (I think, perhaps `PSyLoop` as well?).

src/psyclone/psyir/nodes/while_loop.py
LonelyCat124: Performance and node structure - I combine the DUC calls where I can (as they're expensive), but these need to be for `self.condition.walk(Reference)` vs variable/start/stop/step.walk(Reference). Also note that using `_get_next_accesses` is essentially a recursive call to `next_accesses` on all the nodes (which will all be Statements and thus will combine calls to the DUC wherever possible by implementation), whilst the "condition" (for lack of a combining term) is done directly to the DUC for the combined set of all references.


I guess in principle you could implement a shared method for both, however WhileLoop and Loop share no ancestor other than Statement its not really feasible.

src/psyclone/psyGen.py
arporter: Perhaps these methods should just raise `NotImplementedError` for now and we can think about them in a next step?

arporter: It is the `PSyLoop` objects that hold information on the depth of halo access and are (currently) responsible for creating `HaloExchange` objects in the Schedule. I can't remember why it was implemented in that way. Thinking about it now, it feels like something that should be done as a final step in the creation of the initial Schedule of the PSy layer.

src/psyclone/psyir/nodes/while_loop.py
arporter: Ah. I hadn't appreciated that `WhileLoop` is not a `Loop`. That doesn't seem to be ideal. I also hadn't grasped the distinction between `condition` and the `start`/`stop`/`step` properties.

arporter:
> * I don't particularly like that these are abstract methods but many of their implementations are just `return super().previous_accesses()`. I think I would make them not abstract, and delete all the empty implementations. But again this makes more sense with a single behaviour. I think you decided to use an abstract method to force each node to choose which behaviour they have? But this adds lots of verbosity.

I can see both arguments (defensive programming versus verbosity) but am inclined to agree with Sergi here - i.e. remove the abstract property and remove the empty implementations. We don't add new Node types very often after all.

arporter:
Regarding the "returns different things", I'm more relaxed. However, the DSL-level nodes should probably raise a NotImplementedError rather than silently do the wrong thing. I appreciate that this might get triggered unexpectedly though so it's a suck-it-and-see.

PR #3562, opened 2026-08-19
(closes #3559) Resolve all imports when we have multiple use statements targeting the same module

### LINKED ISSUE #3559: Performance regression in the NEMOv5 OpenMP GPU Async
It was introduced when adding recursive functions and stop propagating UnresolvedInterface, but there was also changes regarding comments that I didn't submit to ITs because in theory they don't use them. Surprisingly only affects the BENCH Async implementation.
Image

sergisiso:


sergisiso:
@LonelyCat124 @arporter This is ready for review. It ended up being a very small fix, even if it was hard to find out what the problem was.

arporter (approved):
Thanks @sergisiso. I can't find anything wrong with this :-) The ITs were all green.
Will proceed to merge.

PR #3565, opened 2026-08-21
(closes #3564) Bump setuptools

### LINKED ISSUE #3564: Issues installing with old setuptools and packaging
Lucas from EPCC reported that psyclone 3.1.1 has issues being installed in pre-existing Spack repositories that have old version of setuptools/packaging. I assume pypi have the same problem but we haven't seen it because we usually do fresh installs.
He suggested:
```
depends_on("py-setuptools@75:", type=("build"),when="@3.3.1:") # PEP 639 implementation , see https://github.com/pypa/setuptools/issues/4629
depends_on("py-packaging@24.2:", type=("build"), when="@3.3.1:") # required to interpret the pyproject.toml file
```

sergisiso:


sergisiso:
@arporter @LonelyCat124 @hiker  This is ready for review, I just updated setuptools to match the proposed update in spack to fix EPCC installation issues. Contrary than in https://github.com/spack/spack-packages/pull/6184 I have not updated 'packaging' here because it is not a direct dependency and pypi doesn't seem to have the same issue.

LonelyCat124:
Gonna run ITs just to be sure but otherwise this seems good to go.

LonelyCat124 (approved):
Waiting for ITs but ready to merge unless one of the remaining ones fails to install somehow.

PR #3573, opened 2026-08-27
Codestyle doc updates and a few small fixes to misuse of Optional

LonelyCat124:
Tiny PR just following up on things I've seen previously. In general I do much prefer `Union[X, None]` over `Optional[None]`, as I think the `Optional` typehint is poorly named (and likely led to the mistakes in this PR).

I only checked for uses on actual type hints. There were some where currently ` = None` is the default, but I suspect could use `= ""`, but I didn't try those modifications as they may need more substantial code changes.

Also just changing some of the docs to update our coding style as it seems very out of date.

### DIFF
--- a/doc/developer_guide/interface_example.py
+++ b/doc/developer_guide/interface_example.py
@@ -5,7 +5,11 @@
 # See the full LICENSE file in the project root for details.
 # -----------------------------------------------------------------------------
 
-def some_function(filename, kernel_path, node=None):
+from typing import Union
+from psyclone.psyir.nodes import Node
+
+def some_function(filename: str, kernel_path: str,
+                  node: Union[Node, None] = None) -> Node:
     '''The description starts with a capital letter and must have
     proper punctuation. Use for example :func:`parse.algorithm.parse`
     to reference functions, or :py:class:`psyclone.psyir.nodes.Node` for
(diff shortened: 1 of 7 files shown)


LonelyCat124:
@arporter small update, ended up being mostly to the docs but cleaned up some old `Optional` misuses too.

LonelyCat124:
@arporter I'm just also adjusting a couple of the docs that haven't come out quite as I wanted but I'll wait to push them until you've reviewed.

doc/developer_guide/interface_example.py
```diff
@@ -5,41 +5,38 @@
 # See the full LICENSE file in the project root for details.
 # -----------------------------------------------------------------------------
 
-def some_function(filename, kernel_path, node=None):
+from typing import Union
```
arporter, line 8: This is good, thanks. We should also mention the need to use `from __future__ import annotations` if a type-spec needs to refer to a class that is defined only after its use in a given module. (Presumably this is an older Python thing?)
We should also mention that if it's not possible to have a 'proper' import of a class referred to in a typehint (because of a circular dep) then its import should be protected by a `if TYPE_CHECKING:` and its actual use in the typehint put in double quotes.

doc/developer_guide/interface_example.py
```diff
@@ -5,41 +5,38 @@
...
-    :param node: the parameter type can also be declared in a stand-alone\
-                 line which follows immediately after the corresponding\
-                 :param: line. The actual type should only contain the\
-                 type, no filler words like 'return type is integer'.\
-                 Type information should be specified according to PEP 483\
-                 (https://peps.python.org/pep-0483/).\
-                 Notice the empty line between parameter and return\
-                 documentation.
-    :type node: Optional[:py:class:`psyclone.psyir.nodes.Node`]
+    :param str filename: start lower case, but add full stop.
```
arporter: You need to remove the 'str' type here and below.

arporter (changes requested):
Looks good Aidan, thanks. It's nice to have these things written down (correctly!).
I've realised that we also need to talk about what do do when things get difficult with getting the necessary class def for a typehint.

doc/developer_guide/interface_example.py
LonelyCat124: Removed.

doc/developer_guide/interface_example.py
LonelyCat124: I'll add them into the Coding and Documentation style list, as they're difficult to show examples int his file I think.

LonelyCat124: I'm not sure if/when `from __future__ import annotations` has become standard, apparently it was originally scheduled for 3.10 but that apparently changed and it may no longer never become standard behaviour, see: https://peps.python.org/pep-0749/  At some point (minimum 3.14 or later) we'll have to move to the new behaviour specified in PEP 649 and 749 I think.

LonelyCat124: Took me a while to get the docs to actually build correctly and look how I wanted, but this is done now too.

LonelyCat124:
@arporter Should be ready for another look now.

doc/developer_guide/interface_example.py
LonelyCat124: (Is there a faster way of building the docs? I do `build html` but its very slow)

arporter: Not AFAIK - it's slow for me too although quicker on subsequent builds as it doesn't have to generate all the (dot) graphs.

arporter: re. the PEPs - I learn something new everyday. I like the additions to the docs.

arporter (approved):
All good now, thanks for these improvements.
Since the changes are exclusively docs and typehints I won't run the ITs.

PR #3574, opened 2026-08-28
(Closes #3568) Removes references to fparser directly in the DUCs

### LINKED ISSUE #3568: DefinitionUseChains still have direct accesses to fparser2 classes
When finding codeblocks, the DUCs check for specific classes for certain code paths, e.g. Goto_Stmt etc.

These checks should be migrated into the Codeblock subclasses, both for Fparser2CodeBlock and TreeSitterCodeBlock. Once this is done we can update the DUCs to use these calls instead of accessing fparser classes.

LonelyCat124:
I decided this was a small change and better to do now, and the TreeSitter changes can come later.

The only things I'm unsure about are why the base CodeBlock class isn't abstract now, and whether its correct to do `-> bool` on an abstractmethod with no code? It conceptually seems correct but I'm not sure about the reality.

Assuming coverage is correct this is ready for a review from @arporter @sergisiso


mn416:
A related method to add here would be `contains_stop_stmt()` (which checks for `Fortran2003.Stop_Stmt` or `Fortran2008.Error_Stop_Stmt`).

It also might be useful to have a method `is_singleton()`, which tells whether the `CodeBlock` contains a single parse node. That way, `cb.is_singleton() and cb.contains_exit()` tells you not only that the `CodeBlock` _contains_ an `exit` statement but that it _is_ an `exit` statement.

LonelyCat124:
I'll add those as well.

LonelyCat124:
Added @mn416 's suggestions (I think), so ready for review again.

src/psyclone/psyir/tools/definition_use_chains.py
```diff
@@ -542,22 +536,15 @@ def _check_forward_codeblock(self, reference: CodeBlock) -> bool:
...
             )
         # If we find an Exit or Cycle statement, we can't
         # reach further in this code region so we can return.
-        # TODO #3568: Codeblocks should have implemented methods
-        # to check for the presence of these statements.
-        if any([isinstance(
-            ref, (Exit_Stmt, Cycle_Stmt)) for ref in
-            reference.parse_tree_nodes]
-        ):
+        if (reference.contains_exit_stmt() or
```
arporter, line 546: I'm wondering whether we should have some sort of `contains_loop_control_flow` method that combines these two. However, presumably a CodeBlock is never (?) going to contain a large chunk of AST so repeatedly walking it shouldn't be expensive?

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -387,6 +419,59 @@ def get_fortran_lines(self) -> list[str]:
...
+    def contains_stop_stmt(self) -> bool:
+        '''
+        :returns: whether the Codeblock contains a STOP statement.
+        '''
+        # Purposely inlined to lazily load this modules only when needed
+        # pylint: disable=import-outside-toplevel
+        from fparser.two import Fortran2003, Fortran2008
+        from fparser.two.utils import walk
+        for node in self._parse_tree_nodes:
+            if walk(node, (Fortran2003.Stop_Stmt,
```
arporter: It's a shame that the fparser2 `walk` doesn't have a "stop once you find one" option but we can always add that as a future optimisation if we need to.

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -387,6 +419,59 @@ def get_fortran_lines(self) -> list[str]:
             output.extend(node.tofortran().split("\n"))
         return output
 
+    def contains_goto_stmt(self) -> bool:
```
arporter: Rather than repeat this same pattern a lot, you could have:
```python
Fparser2CodeBlock.contains_stmt(self, stmt_cls):
    from fparser.two.utils import walk
    for node in self...
    ...
```

and then:

```python
Fparser2CodeBlock.contains_goto_stmt(self):
    return self.contains_stmt(Fortran2003.Goto_Stmt)
```

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -301,26 +301,56 @@ def test_codeblock_ref_accesses(parser):
...
+    assert not codeblocks[1].is_singleton()
+
+    # Test for TreeSitter
+    # TODO #3083: Treesitter support is incomplete so there is always
+    # only one CodeBlock at the moment
+    processor = FortranTreeSitterReader()
+    ptree = processor.generate_parse_tree_from_source(code)
+    psyir = processor.generate_psyir(ptree)
+    codeblocks = psyir.walk(TreeSitterCodeBlock)
+    assert codeblocks[0].is_singleton()
```
arporter: But it therefore presumably contains multiple lines of code and isn't a singleton?

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -346,3 +376,179 @@ def test_codeblock_has_potential_control_flow_jump(fortran_reader):
...
+    assert not codeblocks[1].contains_exit_stmt()
+
+
+def test_fparser_codeblock_contains_stop_stmt(fortran_reader):
+    """Test the contains_stop_stmt function works correctly for Fparser
+    codeblocks."""
+    code = """subroutine test()
+    integer :: i
+    if (i == 10) then
+        STOP 3
```
arporter: Pls add a comment saying this is F2008 syntax (I was wondering why this test had two STOP statements).

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -346,3 +376,179 @@ def test_codeblock_has_potential_control_flow_jump(fortran_reader):
...
+    write(*,*) "Hello"
+    1234 i = 3
+    end subroutine test"""
+    processor = FortranTreeSitterReader()
+    ptree = processor.generate_parse_tree_from_source(code)
+    psyir = processor.generate_psyir(ptree)
+    codeblocks = psyir.walk(TreeSitterCodeBlock)
+    with pytest.raises(NotImplementedError) as excinfo:
+        assert codeblocks[0].contains_goto_stmt()
+        assert not codeblocks[1].contains_goto_stmt()
```
arporter: This second assert will never be hit?

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -346,3 +376,179 @@ def test_codeblock_has_potential_control_flow_jump(fortran_reader):
...
+    write(*,*) "Hello"
+    1234 i = 3
+    end subroutine test"""
+    processor = FortranTreeSitterReader()
+    ptree = processor.generate_parse_tree_from_source(code)
+    psyir = processor.generate_psyir(ptree)
+    codeblocks = psyir.walk(TreeSitterCodeBlock)
+    with pytest.raises(NotImplementedError) as excinfo:
+        assert codeblocks[0].contains_cycle_stmt()
+        assert not codeblocks[1].contains_cycle_stmt()
```
arporter: Also won't be hit (could comment it out and add a TODO?)

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -346,3 +376,179 @@ def test_codeblock_has_potential_control_flow_jump(fortran_reader):
...
+    write(*,*) "Hello"
+    1234 i = 3
+    end subroutine test"""
+    processor = FortranTreeSitterReader()
+    ptree = processor.generate_parse_tree_from_source(code)
+    psyir = processor.generate_psyir(ptree)
+    codeblocks = psyir.walk(TreeSitterCodeBlock)
+    with pytest.raises(NotImplementedError) as excinfo:
+        assert codeblocks[0].contains_exit_stmt()
+        assert not codeblocks[1].contains_exit_stmt()
```
arporter: Same here.

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -346,3 +376,179 @@ def test_codeblock_has_potential_control_flow_jump(fortran_reader):
...
+    write(*,*) "Hello"
+    1234 i = 3
+    end subroutine test"""
+    processor = FortranTreeSitterReader()
+    ptree = processor.generate_parse_tree_from_source(code)
+    psyir = processor.generate_psyir(ptree)
+    codeblocks = psyir.walk(TreeSitterCodeBlock)
+    with pytest.raises(NotImplementedError) as excinfo:
+        assert codeblocks[0].contains_stop_stmt()
+        assert codeblocks[1].contains_stop_stmt()
```
arporter: Same here. Once treesitter support is implemented, hopefully we could parameterise the existing fparser2 tests so that the tests are all just run twice, once for each frontend?

arporter (changes requested):
Thanks for fixing this Aidan. You'll see I worried a bit about efficiency but decided that might well prove to be premature optimisation so left it (and it would require an fparser change anyway). Possibly the implementation could be tidied slightly - see what you think of my suggestion.
I tried building the docs and looking at the API reference and discovered that `codeblock.py` is missing an `__all__` at the bottom so we don't get Sphinx AutoAPI docs for it. Please could you add this?
I'll check the ITs next time.

mn416:
Thanks @LonelyCat124 for adding my suggestions. The `is_singleton()` definition is kind of what I had in mind but combining it with, say, `is_exit()` would still match a non-exit statement that contains an exit statement as a child. I can't think of anything better than:

```python
def contains_exit_stmt(self, only: bool = False) -> bool:
    if only:
        return (len(self._parse_tree_nodes) == 1 and
                isinstance(self._parse_tree_nodes[0],
                           Fortran2003.Exit_Stmt))
    else:
        for node in self._parse_tree_nodes:
            if walk(node, Fortran2003.Exit_Stmt):
                return True
    return False
```

This is quite verbose, but Andy already suggested in his review that the general structure of these methods could be abstracted out and not repeated for each node class, so it might look quite neat after that refactoring. With this approach we could then drop `is_singleton()`.

src/psyclone/psyir/tools/definition_use_chains.py
LonelyCat124: Yeah I felt so - its also not totally clear to me when `contains_loop_control_flow` would be valid and `has_potential_control_flow_jump` wouldn't be sufficient (since goto is almost loop control flow anyway).

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Yeah - I did check fparser for that but unfortunately not at the moment. If we hit performance issues with this then we can add it, but I suspect usage of these functions will be quite sparse so its not likely to cause much issue.

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Added

src/psyclone/tests/psyir/nodes/codeblock_test.py
LonelyCat124: The base implementation of singleton checkks if `len(self._parse_tree_nodes) == 1`, and this does only contain a single node (that has many children). I think @mn416 might just supercede this "is_singleton" approach as its otherwise too messy to implement well.

src/psyclone/tests/psyir/nodes/codeblock_test.py
LonelyCat124: Added

src/psyclone/tests/psyir/nodes/codeblock_test.py
LonelyCat124: Yeah, I left it in as something that would be correct once implemented, but I'll comment the second asserts out and add a TODO for all of these.

src/psyclone/tests/psyir/nodes/codeblock_test.py
LonelyCat124: Possibly - I think we will need a fixture at some point for treesitter, but its too early for now for it to exist.

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: I didn't feel like the has_potential_control_flow jump could quite use this implementation, but the others all do now.

LonelyCat124:
All comments addressed, ready for another look @arporter

src/psyclone/tests/psyir/nodes/codeblock_test.py
LonelyCat124: (Or if there's some way for the current fortran_reader to take a choice as an option somehow)

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -229,35 +229,40 @@ def get_fortran_lines(self) -> list[str]:
         raise NotImplementedError("Use appropriate CodeBlock subclass")
 
     @abc.abstractmethod
-    def contains_goto_stmt(self) -> bool:
+    def contains_goto_stmt(self, only: bool = False) -> bool:
         '''
+        :param only: whether to check if the only statement containined
```
arporter: "contained"

arporter: and for the other modified docstrings below...

src/psyclone/psyir/nodes/codeblock.py
```diff
@@ -419,58 +424,81 @@ def get_fortran_lines(self) -> list[str]:
             output.extend(node.tofortran().split("\n"))
         return output
 
-    def contains_goto_stmt(self) -> bool:
+    def _contains_stmt(self, stmt_cls: Union[type, Iterable[type]],
+                       only: bool = False) -> bool:
         '''
-        :returns: whether the Codeblock contains a GOTO statement.
+        :param stmt_cls: the class or classes to check if is contains in this
```
arporter: s/if is contains/for/

src/psyclone/tests/psyir/nodes/codeblock_test.py
```diff
@@ -378,6 +348,30 @@ def test_codeblock_has_potential_control_flow_jump(fortran_reader):
...
+    print *, i
+    end subroutine test"""
+    psyir = fortran_reader.psyir_from_source(code)
+    codeblocks = psyir.walk(Fparser2CodeBlock)
+    # The first CodeBlock contains both a Print and Goto stmt.
+    assert codeblocks[0]._contains_stmt((Fortran2003.Print_Stmt,
+                                         Fortran2003.Stop_Stmt))
+    assert not codeblocks[0]._contains_stmt(Fortran2003.Print_Stmt, only=True)
+    # The second CodeBlock only contains a Print Stmt
+    assert codeblocks[1]._contains_stmt(Fortran2003.Print_Stmt, only=False)
```
arporter, line 372: Repeat with `only=True` too?

LonelyCat124:
@arporter You've marked this as reviewed with actions but I don't see any new comments, did something go wrong?

arporter (changes requested):
Very nearly there now, thanks @LonelyCat124. Just a bit of tidying and `codeblock.py` still needs an `__all__` at the end of it.
I've set the ITs running...

arporter:
Sorry @LonelyCat124, must have pressed the wrong button - my review was still waiting to be submitted.

arporter:
(ITs were green.)

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Fixed.

src/psyclone/psyir/nodes/codeblock.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/nodes/codeblock_test.py
LonelyCat124: Yeah, I suspect this was meant to be `only=True` originally, but testing both seems good.

LonelyCat124:
@arporter Addressed those comments and move the `__all__` for Codeblock into the file from `__init__.py`. Should be ready for another look.

arporter:
Thanks Aidan. Documentation generates fine. ITs were green. I'll proceed to merge.

arporter (approved):

arporter:
Just waiting on #3520.

PR #3577, opened 2026-09-01
(Closes #1957) Added an Intrinsic2CodeTrans metatransformation

### LINKED ISSUE #1957: MetaTransformation for applying all possible Intrinsic2Code transformations
In both LFRic and Nemo I need to expand the intrinsic calls to its explicit/loop form. Some of this transformations already exists e.g. Sum2Code, Matmul2Code, ... But it would be convenient to have a metatransformation that does applies all of then for a given region of code.

LonelyCat124:
Simple additional metatransformation to cleanup the nemo script.


LonelyCat124:
Should be ready now for a first look from @arporter or @sergisiso . I had a name conflict which I changed, so this transformation does have `MetaTrans` in its name - happy to change the name if the reviewer has a better idea.

examples/nemo/scripts/out.F90
```diff
@@ -0,0 +1,33 @@
+module stringop
```
arporter: Is this file supposed to be included? Similarly there a `out` with just whitespace changes?

examples/nemo/scripts/stringop.f90
```diff
@@ -0,0 +1,33 @@
+MODULE stringop
```
arporter: I don't think we should have a .f90 file in the `nemo/scripts` directory?

examples/nemo/scripts/utils.py
```diff
@@ -186,14 +186,7 @@ def normalise_loops(
...
             try:
-                if intr.intrinsic.name == "MAXVAL":
-                    Maxval2LoopTrans().apply(intr, verbose=True)
-                elif intr.intrinsic.name == "SUM":
-                    Sum2LoopTrans().apply(intr, verbose=True)
-                elif intr.intrinsic.name == "MINVAL":
-                    Minval2LoopTrans().apply(intr, verbose=True)
-                elif intr.intrinsic.name == "PRODUCT":
-                    Product2LoopTrans().apply(intr, verbose=True)
+                Intrinsic2CodeMetaTrans().apply(intr, verbose=True)
```
arporter: Nice.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_metatrans.py
```diff
@@ -0,0 +1,77 @@
...
+    import Minval2LoopTrans
+from psyclone.psyir.transformations.intrinsics.sum2loop_trans\
+    import Sum2LoopTrans
+from psyclone.psyir.transformations.intrinsics.product2loop_trans\
+    import Product2LoopTrans
+from psyclone.utils import transformation_documentation_wrapper
+
+
+@transformation_documentation_wrapper
+class Intrinsic2CodeMetaTrans(Transformation):
```
arporter: Regarding naming, if @sergisiso and @hiker agree, I would suggest re-naming the existing `Intrinsic2CodeTrans` to `Intrinsic2CodeBaseTrans` (or `SingleIntrinsic2CodeTrans` or similar) and then having this one as `Intrinsic2CodeTrans` as it's the one we'd expect people to use.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,77 @@
...
+    transformations to the provided input. The available transformations are
+    Maxval2LoopTrans, Sum2LoopTrans, Minval2LoopTrans, or Product2LoopTrans.
+
+    '''
+    _SUB_TRANSFORMATIONS = [Maxval2LoopTrans, Sum2LoopTrans,
+                            Minval2LoopTrans, Product2LoopTrans]
+
+    # Create a map of intrinsic names to the appropriate Intrinsic2Code
+    # transformation.
+    intrinsic_to_trans = {"MAXVAL": Maxval2LoopTrans,
```
arporter: Rather than use strings, can we use members of the `IntrinsicCall.Intrinsic` enum instead?

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_metatrans.py
```diff
@@ -0,0 +1,77 @@
...
+        kwargs_dict = {}
+        local_kwargs, kwargs_dict["MAXVAL"], kwargs_dict["SUM"], \
+            kwargs_dict["MINVAL"], kwargs_dict["PRODUCT"] = \
+            self.split_kwargs(**kwargs)
+
+        self.validate(node, **local_kwargs)
+
+        # If the intrinsic is one of the supported intrinsics then
+        # apply the relevant transformation.
+        if node.intrinsic.name in Intrinsic2CodeMetaTrans.intrinsic_to_trans:
```
arporter: If we can use the members of the IntrinsicCall.Intrinsic enum as keys, we can remove the various `.name` here.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,77 @@
...
+            kwargs_dict["MINVAL"], kwargs_dict["PRODUCT"] = \
+            self.split_kwargs(**kwargs)
+
+        self.validate(node, **local_kwargs)
+
+        # If the intrinsic is one of the supported intrinsics then
+        # apply the relevant transformation.
+        if node.intrinsic.name in Intrinsic2CodeMetaTrans.intrinsic_to_trans:
+            Intrinsic2CodeMetaTrans.intrinsic_to_trans[node.intrinsic.name]().\
+                apply(node, **kwargs_dict[node.intrinsic.name])
```
arporter: We could add some logging here. I'm never sure what an appropriate level of detail would be but I'm thinking that, for debugging purposes, it might be good if a user could discover that a given intrinsic has not been transformed because this transformation doesn't support it?

src/psyclone/psyir/transformations/__init__.py
```diff
@@ -173,4 +175,5 @@
     "OMPCriticalTrans",
     "MaximalOMPParallelRegionTrans",
     "OMPParallelTrans",
+    "Intrinsic2CodeMetaTrans",
```
arporter: I *think* that having this `__all__` list in this `__init__.py` file results in the annoying situation where the AutoAPI docs will create an empty page for e.g. `Intrinsic2CodeMetaTrans` (and all the others here). If you could put an `__all__` at the end of the appropriate source file instead the that seems to work better.

src/psyclone/tests/psyir/transformations/metatransformations/intrinsic2code_trans_test.py
```diff
@@ -0,0 +1,73 @@
...
+    Tests the validate method of the Intrinsic2CodeMetaTrans
+    metatransformation.
+    '''
+    with pytest.raises(TypeError) as err:
+        Intrinsic2CodeMetaTrans().validate(123)
+    assert ("Input node to Intrinsic2CodeMetaTrans must be an IntrinsicCall "
+            "but received 'int'." in str(err.value))
+
+
+@pytest.mark.parametrize("code, expected", [
```
arporter: I think all we need to test here is that the appropriate `apply` method is being called - the unit tests for each individual transformation obviously cover their functionality. You might be able to do that by monkeypatching the `apply` method on the class so that it raises an exception and then checking for that.

arporter: (You'd need to monkeypatch the instance of the individual transfomration classes that are stored within the meta class.)

src/psyclone/psyGen.py
```diff
@@ -2372,7 +2372,11 @@ def split_kwargs(self, **kwargs) -> tuple[dict[str, Any]]:
                 if key in trans.get_valid_options():
                     other_dicts[idx][key] = kwargs[key]
                     if key not in type(self).get_valid_options():
-                        del first_dict[key]
+                        # Sometimes we may have the same option in multiple
+                        # subtransformations, so we only delete the key
+                        # from the first_dict if its still present.
```
arporter: "it's"

arporter (changes requested):
Thanks Aidan, I think this is a good simplification to the user interface. If @hiker and @sergisiso agree with my naming suggestion, it may be best to do that first as a separate PR as files and test files will need renaming/moving.

A few strange files have appeared in `examples/nemo/scripts` and probably shouldn't be there?

Finally, please update the UG section on transformations to include this functionality.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
sergisiso: Also, we have a few more: abs2code, matmul2code, ... is it a problem for NEMO is we add them all here?

LonelyCat124: I'm not sure - I just mirrored the behaviour in the NEMO scripts, I think it would be better to do a separate performance analysis to check?

examples/nemo/scripts/out.F90
LonelyCat124: Oops, I think I did `git add .` as I would for the src dir and ended up with bonus files, I'll make sure to remove them.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_metatrans.py
sergisiso: I agree with this suggestion, but another alternative is to use something like: `AllArrayIntrinsics2Code` or something that make more clear that this transformation is aiming to loopify intrinsics that apply to all items of the array.

src/psyclone/tests/psyir/transformations/metatransformations/intrinsic2code_trans_test.py
LonelyCat124: @arporter I'm not sure I quite understand, does this not test that anyway? Or I guess the point is if we change the functionality/output from the other tests we break this test as an unintended side-effect and we shouldn't care about the actual result in the tests for the MetaTransformation as they're an implementation detail of the sub transformations?

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: I guess these intrinsics are all array -> scalar result intrinsics, I could do a base one (this one) and nanother `ArrayToScalarIntrinsics2Code` or similar that just does these 4? I think we can in theory only subclass by overriding `intrinsic_to_trans`, but there may be some funny doc/options that appear in the docstring or are accepted but would be ignored.

examples/nemo/scripts/out.F90
LonelyCat124: Removed in 9189de5

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_metatrans.py
LonelyCat124: Naming redone in [56215e5](https://github.com/stfc/PSyclone/pull/3577/commits/56215e58711d02ce194366ce6c145b39c000fd8f)

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_metatrans.py
LonelyCat124: Fixed in [af1fbb7](https://github.com/stfc/PSyclone/pull/3577/commits/af1fbb7e1bf08ed15e276d89cf30250ffb7196f5)

src/psyclone/psyir/transformations/__init__.py
LonelyCat124: Done during renaming.

src/psyclone/psyGen.py
LonelyCat124: Done

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: Added in [0aa1c65](https://github.com/stfc/PSyclone/pull/3577/commits/0aa1c65a120272ce5d9955777b998f1e2c60701f)

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: Done in [7987734](https://github.com/stfc/PSyclone/pull/3577/commits/7987734a9a0efec9888a761f1330ed3a6ff1002d)

LonelyCat124:
@arporter All comments addressed and the transformations are added to the UG, should be ready for another look.

src/psyclone/tests/psyir/transformations/metatransformations/intrinsic2code_trans_test.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,134 @@
...
+    import Sum2LoopTrans
+from psyclone.utils import transformation_documentation_wrapper
+
+
+@transformation_documentation_wrapper
+class Intrinsic2CodeTrans(Transformation):
+    '''This metatransformation applies any of the Intrinsic2Code
+    transformations to the provided input.
+    '''
+    _SUB_TRANSFORMATIONS = [Maxval2LoopTrans, Sum2LoopTrans,
```
arporter: Could you make use of your new `ArrayIntrinsic2LoopTrans` here rather then separately listing all the loop ones?

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,134 @@
...
+                            DotProduct2CodeTrans, Abs2CodeTrans,
+                            Max2CodeTrans, Min2CodeTrans,
+                            Sign2CodeTrans, Matmul2CodeTrans]
+
+    # Create a map of intrinsic names to the appropriate Intrinsic2Code
+    # transformation. This should be in the same order as the
+    # _SUB_TRANSFORMATIONS else the _split_kwargs on this Transformation
+    # may not work correctly.
+    intrinsic_to_trans = {
+        IntrinsicCall.Intrinsic.MAXVAL: Maxval2LoopTrans,
```
arporter: Again `ArrayIntrinsic2LoopTrans` perhaps?

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,134 @@
...
+    def _split_kwargs(self, **kwargs) -> \
+            tuple[dict[str, Any],
+                  dict[IntrinsicCall.Intrinsic, dict[str, Any]]]:
+        '''
+        :returns: the kwargs for this transformation and the kwargs dict for
+           the sub transformations indexed by appropriate Intrinsic.
+        '''
+        # The split_kwargs function returns a tuple containing the
+        # kwargs for this transformation as the first entry and then
+        # the kwargs for the SUB_TRANSFORAMTIONS in order as the following
```
arporter, line 91: "...in the same order as ...?" perhaps?

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def apply(self, node: IntrinsicCall, **kwargs) -> None:
+        '''
+        Applies the appropriate Intrinsic2Code transformation to the provided
+        input node.
+
+        :param node: the IntrinsicCall to be transformed.
+        '''
+        # Split the options for the subtransformations. The options are
+        # returned in the order of the _SUB_TRANSFORMATIONS list.
```
arporter, line 108: Are they not in a dict, indexed by the target Intrinsic now?

src/psyclone/psyir/transformations/__init__.py
```diff
@@ -117,7 +117,9 @@
 from psyclone.psyir.transformations.metatransformations.omp_cpu_routine_trans\
     import OMPCPURoutineTrans
 from psyclone.psyir.transformations.metatransformations.\
-    intrinsic2code_metatrans import Intrinsic2CodeMetaTrans
+    intrinsic2code_trans import Intrinsic2CodeTrans
```
arporter, line 120: Alpha order pls.

doc/user_guide/transformations.rst
```diff
@@ -270,6 +270,18 @@ can be found in the API-specific sections).
 
 ####
 
+.. autoclass:: psyclone.psyir.transformations.Intrinsic2CodeTrans
+      :members: apply
+      :no-index:
+
+####
+
+.. autoclass:: psyclone.psyir.transformations.ArrayIntrinsic2LoopTrans
```
arporter, line 279: Pls put this one in alpha order.

src/psyclone/tests/psyir/transformations/metatransformations/arrayintrinsic2loop_trans_test.py
```diff
@@ -0,0 +1,71 @@
...
+@pytest.mark.parametrize("code, transformation", [
+    ("j = MAXVAL(i)", Maxval2LoopTrans),
+    ("j = MINVAL(i)", Minval2LoopTrans),
+    ("j = PRODUCT(i)", Product2LoopTrans),
+    ("j = SUM(i)", Sum2LoopTrans),
+    ])
+def test_arrayintrinsic2loop_trans_apply(fortran_reader, monkeypatch,
+                                         code, transformation):
+    '''Test the apply function of the ArrayIntrinsic2LoopTrans
+    metatransformation.
```
arporter, line 32: ... calls the apply method of the expected sub transformation. (or similar?)

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
```diff
@@ -0,0 +1,134 @@
...
+                node, **kwargs_dict[node.intrinsic]
+            )
+        else:
+            # Setup the logger.
+            logger = logging.getLogger(__name__)
+            supported_intrinsics = []
+            if logger.isEnabledFor(logging.INFO):
+                supported_intrinsics = [intrinsic.name for intrinsic in
+                                        self.intrinsic_to_trans.keys()]
+            logger.info(
```
arporter, line 127: Should this be indented one further level?

src/psyclone/tests/psyir/transformations/metatransformations/intrinsic2code_trans_test.py
```diff
@@ -0,0 +1,87 @@
...
+    ("j = ABS(i)", Abs2CodeTrans),
+    ("j = MAX(i(0), i(1))", Max2CodeTrans),
+    ("j = MIN(i(0), i(1))", Min2CodeTrans),
+    ("j = SIGN(i(0), +0.5)", Sign2CodeTrans),
+    ("j = MATMUL(i, i)", Matmul2CodeTrans),
+    ])
+def test_intrinsic2code_trans_apply(fortran_reader, monkeypatch,
+                                    code, transformation):
+    '''Test the apply function of the Intrinsic2CodeTrans
+    metatransformation.
```
arporter, line 50: ...calls the apply method of the correct sub-transformation.

arporter (changes requested):
Thanks @LonelyCat124 - looking pretty good now. Just a minor bit of tidying and I think there may be a mistake with indentation in the new logging code.
I've set the ITs going.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: It doesn't need to be, I think best practice is to let the logger itself determine whether to log a message (`logger.info` determines whether to output a logging message or not anyway, so it doesn't need to be inside the if`. 

The code above it is has an optimisation to avoid computing the `supported_intrinsics` if the wrong logging level is used, but I'm happy to just always do this to make the code cleaner?

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: Added

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: Yeah, updated.

src/psyclone/psyir/transformations/__init__.py
LonelyCat124: Done.

doc/user_guide/transformations.rst
LonelyCat124: Done.

src/psyclone/tests/psyir/transformations/metatransformations/arrayintrinsic2loop_trans_test.py
LonelyCat124: Done.

src/psyclone/tests/psyir/transformations/metatransformations/intrinsic2code_trans_test.py
LonelyCat124: Done.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
LonelyCat124: I could probably (I'm not sure) use lazy string instead (though `[intrinsic.name for intrinsic in self.intrinsic_to_trans.keys()]` is quite long to fit in an fstring). Let me know if you have a preference.

LonelyCat124:
@arporter Should be ready for another look - let me know if you want me to change anything about the logging as its correct but I'm not sure what the best approach to avoid unnecessary computation of the list is.

src/psyclone/psyir/transformations/metatransformations/intrinsic2code_trans.py
arporter: No it's fine. Thanks for adding the comment.

arporter (approved):
Thanks very much Aidan. All looks good now. The ITs were green (last week) and I don't think it's worth re-running them.
I'll proceed to merge.

arporter (approved):
All good now.
I will re-run the ITs, just to be sure.

PR #3582, opened 2026-09-03
(closes #3392) Preserve order of common block sequences

### LINKED ISSUE #3392: Extend CommonBlockInterface so that it holds the name of the common block

sergisiso:
We do it by keeping a index in the CommonBlockInterface:
- the frontend will keep track of the index number when inserting multiple symbols to the same common interface looking at the previous inserted symbols.
- the backend will order them using this value

I also tweaked symbol removal to ensure we don't remove symbols that would leave a gap in the sequence.

### DIFF
--- a/src/psyclone/psyir/backend/fortran.py
+++ b/src/psyclone/psyir/backend/fortran.py
@@ -726,7 +726,7 @@ class FortranWriter(LanguageWriter):
                 name = symbol.interface.name.lower()
                 common_blocks.setdefault(name, []).append(symbol)
 
-        # Order the symbols by their commonblock interface possition
+        # Order the symbols by their commonblock interface position
         declarations = ""
         for name, members in common_blocks.items():
             positions = [symbol.interface.position for symbol in members]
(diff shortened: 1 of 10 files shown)


sergisiso:
@arporter @LonelyCat124 This is ready for review

src/psyclone/psyir/backend/fortran.py
```diff
@@ -703,14 +703,44 @@ def gen_vardecl(self,
...
+        '''
+        # Create a dict of common block names and the symbols that belong to
+        # that common block
+        common_blocks = {}
+        for symbol in symbols:
+            if symbol.is_commonblock:
+                name = symbol.interface.name.lower()
+                common_blocks.setdefault(name, []).append(symbol)
+
+        # Order the symbols by their commonblock interface possition
```
LonelyCat124:
```suggestion
        # Order the symbols by their commonblock interface position
```

LonelyCat124 (approved):
One comment typo, but otherwise I'm happy with this subject to ITs. I'll fix the comment myself.

PR #3584, opened 2026-09-04
#3583 Preserve Symbol tag when swapping symbols in a SymbolTable

### LINKED ISSUE #3583: SymbolTable.swap() does not account for tags
The `swap` method does not check whether the original Symbol is associated with a tag and therefore its use can result in us silently losing tags. I think that this is probably not what we want.

ump45nose:
## Summary

Fixes #3583: `SymbolTable.swap()` silently drops any tag associated with the
old symbol.

## Root cause

`swap()` calls `remove(old_symbol)` — which disassociates any tag pointing at
the symbol — and then `add(new_symbol)` without a `tag` argument, so the tag is
never re-established on the new symbol.

## Changes

`src/psyclone/psyir/symbols/symbol_table.py`:
- Before removal, record the tag associated with `old_symbol` (at most one
  exists per symbol, per the `get_reverse_tags_dict()` invariant), then pass it
  to `add(new_symbol, tag=old_tag)` so the tag follows the replacement.

`src/psyclone/tests/psyir/symbols/symbol_table_test.py`:
- Added `test_swap_symbol_preserves_tag`, which swaps a tagged symbol and
  asserts the tag now resolves to the new symbol via `lookup_with_tag`.

## Verification

- `flake8 --count --show-source --statistics src/psyclone/psyir/symbols/symbol_table.py` — 0 issues
- `pylint src/psyclone/psyir/symbols/symbol_table.py` — only pre-existing
  warnings (TODO comments, `too-many-branches`/`too-many-statements` in `add`
  and other existing methods); the changed `swap` method is clean
- `pytest src/psyclone/tests/psyir/symbols/symbol_table_test.py -q` — 105 passed, 1 xfailed

## AI assistance disclosure

This change was implemented with AI assistance (Pi coding agent). The author
reviewed and verified every line, reproduced the bug by reading the
`remove`/`add`/`swap` code paths, and ran the checks above to confirm the
reported results.

### DIFF
--- a/src/psyclone/psyir/symbols/symbol_table.py
+++ b/src/psyclone/psyir/symbols/symbol_table.py
@@ -1253,13 +1253,20 @@ class SymbolTable():
             raise SymbolError(
                 f"Cannot swap symbols that have different names, got: "
                 f"'{old_symbol.name}' and '{new_symbol.name}'")
+        # Preserve any tag associated with old_symbol so that it is carried
+        # over to new_symbol rather than being silently dropped.
+        old_tag = None
+        for tag, symbol in self._tags.items():
+            if symbol is old_symbol:
+                old_tag = tag
+                break
         for sym in self.symbols:
             sym.replace_symbols_using(new_symbol)
         if self.node:
             # Update the PSyIR tree associated with this table.
             self.node.replace_symbols_using(new_symbol)
         self.remove(old_symbol)
-        self.add(new_symbol)
+        self.add(new_symbol, tag=old_tag)
 
     def _validate_remove_routinesymbol(self, symbol):
         '''
--- a/src/psyclone/psyir/symbols/symbol_table.py
+++ b/src/psyclone/psyir/symbols/symbol_table.py
@@ -1228,16 +1228,15 @@ class SymbolTable():
         return [symbol for symbol in self.imported_symbols if
                 symbol.interface.container_symbol is csymbol]
 
-    def swap(self, old_symbol, new_symbol):
+    def swap(self, old_symbol: Symbol, new_symbol: Symbol) -> None:
         '''
         Remove the `old_symbol` from the table and replace it with the
         `new_symbol`. Any references to `old_symbol` in the PSyIR tree
-        associated with this table (if any) will also be updated.
+        associated with this table (if any) will also be updated. Any tag
+        associated with `old_symbol` is moved to `new_symbol`.
 
         :param old_symbol: the symbol to remove from the table.
-        :type old_symbol: :py:class:`psyclone.psyir.symbols.Symbol`
         :param new_symbol: the symbol to add to the table.
-        :type new_symbol: :py:class:`psyclone.psyir.symbols.Symbol`
 
         :raises TypeError: if either old/new_symbol are not Symbols.
         :raises SymbolError: if `old_symbol` and `new_symbol` don't have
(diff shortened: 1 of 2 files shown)


src/psyclone/psyir/symbols/symbol_table.py
LonelyCat124: The docs should probably say the tag is moved from old_symbol to new_symbol (or some similar wording).

LonelyCat124: We need to move type hints from the docs onto the declaration on this method if we're changing it, please add them here (and also `-> None` for the return type). Note these type hints must be compatible with Python 3.9.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
```diff
@@ -662,6 +662,21 @@ def test_swap_symbol():
...
+    old symbol by re-associating it with the new symbol. '''
+    sym_table = symbols.SymbolTable()
+    symbol1 = symbols.Symbol("var1")
+    sym_table.add(symbol1, tag="var1_tag")
+    symbol2 = symbols.Symbol("Var1")
+
+    sym_table.swap(symbol1, symbol2)
+
+    assert sym_table.lookup("var1") is symbol2
+    # The tag must follow the new symbol rather than being dropped.
```
LonelyCat124:
```suggestion
    # Search for tag must find the new symbol rather than being dropped.
```

changelog
```diff
@@ -1,3 +1,6 @@
+   47) PR #3584 for #3583. Preserve the tag of a symbol when it is
```
LonelyCat124: Revert the changes to the changelog - this is done by the reviewer before merging.

LonelyCat124 (commented):
Quick review while waiting on testing, a few small issues (some of which are just code modernisation requirements for changed code) and some comment/doc fixes.

I'm going to hold off on any approval, I think @arporter would be best to look at this as I'm not 100% sure on where the original issue was hit.

LonelyCat124 (changes requested):
See changes from previous review, once those are done I'll launch ITs to check all is ok

src/psyclone/psyir/symbols/symbol_table.py
ump45nose: Done - the docstring now states that the tag is moved from `old_symbol` to `new_symbol`.

ump45nose: Done - moved the type hints onto the declaration (`old_symbol: Symbol, new_symbol: Symbol`) and added `-> None`. They are Python 3.9 compatible.

src/psyclone/tests/psyir/symbols/symbol_table_test.py
ump45nose: Applied the suggested comment wording.

changelog
ump45nose: Done - reverted the changelog change.

ump45nose:
Thanks for the review. I've addressed all four points and rebased onto current `master` to resolve the conflict:

- Moved type hints from the docstring onto the `swap()` declaration (`old_symbol: Symbol, new_symbol: Symbol`) and added `-> None` (Python 3.9 compatible).
- Updated the docstring to state the tag is *moved* from `old_symbol` to `new_symbol`.
- Applied the suggested test-comment wording.
- Reverted the changelog change.

Tests: `symbol_table_test.py` passes (107 passed, 1 xfailed).

LonelyCat124:
I've set integration running, but otherwise this looks fine other than some small fixes I'll need to do as part of/after merging.

LonelyCat124 (approved):
ITs were green - I'll update the changelog, fix the remaining comment (as it was my mistake) in `symbol_table_test.py` and update the contributers and authors docs on master once merging is done.

PR #3585, opened 2026-09-07
#3575 Don't walk unnecessary if a path was already given.

### LINKED ISSUE #3575: Avoid duplicated search paths in module manager
UKMO is intending to run PSyclone from the Fab build script directly (i.e. without starting psyclone as subprocess).

While this has been shown to work, I believe it might cause some unnecessary overhead internally: each call will add to the module manager's search path, and it could be that this triggers (when calling PSyclone's `generate` explicitly) repeatedly searching the same paths over and over.

This ticket is just to assess this (and fix it if it if required).

hiker:
Closes #3575.
Avoids walking the file system repeatedly if the same (or already visited) directories are specified.
Useful if UKMO intends to call PSyclone repeatedly from a Python apps (instead of starting it over and over in subprocesses)

### DIFF
--- a/src/psyclone/parse/module_manager.py
+++ b/src/psyclone/parse/module_manager.py
@@ -232,6 +232,8 @@ class ModuleManager:
             directories = [directories]
 
         for directory in directories:
+            if directory in self._original_search_paths:
+                continue
             if not isinstance(directory, (str, Path)):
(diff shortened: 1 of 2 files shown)


hiker:
Closing this, since it's not clear if this kind of optimisation is useful, and it breaks potential use cases when a script creates additional files in directories that have already been parsed.

PR #3587, opened 2026-09-07
(towards #2216) Add KernelInlineTrans to do a deferred InlineTrans after lowering the psy-layer.

sergisiso:


sergisiso:
There are a couple of issues remaining:
-  After deferred-inlining the OpenMP clauses are outdated, variables that were internal to the kernel should now be private.
- Sometimes "arguments that depend on uninitialised symbols" end up inlined. For example `rhs_sample_eos_code` produces:
```
    integer(kind=i_def) :: ndf_w3
    real(kind=r_def), dimension(ndf_w3) :: rho_e
    real(kind=r_def), dimension(ndf_w3) :: exner_e
    ndf_w3 = rhs_proxy%vspace%get_ndf()
```

sergisiso:
> Sometimes "arguments that depend on uninitialised symbols" end up inlined. For example rhs_sample_eos_code produces:
>    integer(kind=i_def) :: ndf_w3
>    real(kind=r_def), dimension(ndf_w3) :: rho_e
>    real(kind=r_def), dimension(ndf_w3) :: exner_e
>    ndf_w3 = rhs_proxy%vspace%get_ndf()

The reason it fails is because `previous_accesses()` of 'ndf_w3' stops at an ALLOCATE statement, and all "previous.is_write" are false:
```
print(list(reversed([x.ancestor(Statement).debug_string() for x in actual_arg.previous_accesses()])))
[
'ALLOCATE(basis_wtheta_on_w3(dim_wtheta,ndf_wtheta,ndf_w3))\n',
 '\n'
 '! Compute basis/diff-basis arrays\n'
 'do df_nodal = 1, ndf_w3, 1\n'
 '  do df_w3 = 1, ndf_w3, 1\n'
 '    basis_w3_on_w3(:,df_w3,df_nodal) = rhs_proxy%vspace%call_function(BASIS, '
 'df_w3, nodes_w3(:,df_nodal))\n'
 '  enddo\n'
 'enddo\n',
 'do df_w3 = 1, ndf_w3, 1\n'
 '  basis_w3_on_w3(:,df_w3,df_nodal) = rhs_proxy%vspace%call_function(BASIS, '
 'df_w3, nodes_w3(:,df_nodal))\n'
 'enddo\n',
 'do df_nodal = 1, ndf_w3, 1\n'
 '  do df_wtheta = 1, ndf_wtheta, 1\n'
 '    basis_wtheta_on_w3(:,df_wtheta,df_nodal) = '
 'theta_proxy%vspace%call_function(BASIS, df_wtheta, nodes_w3(:,df_nodal))\n'
 '  enddo\n'
 'enddo\n',
```
But in reality above the ALLOCATE there is a `'ndf_w3 = rhs_proxy%vspace%get_ndf()\n'`. @LonelyCat124 I am wondering if the  DUC stops at any allocate? Even if it is:
```
"IntrinsicCall[name='ALLOCATE']\n"
 "    Reference[name:'ALLOCATE']\n"
 "    ArrayReference[name:'basis_wtheta_on_w3']\n"
 "        Reference[name:'dim_wtheta']\n"
 "        Reference[name:'ndf_wtheta']\n"
 "        Reference[name:'ndf_w3']\n"
```
Instead of continue looking backwards for 'ndf_w3'?

LonelyCat124:
It probably stops because it believes its a write since its not an inquiry function, it will get to this in `_check_backward_matched_reference`:
```
        elif reference.ancestor((Call, CodeBlock)):
            # Otherwise we assume read/write access as PSyclone does not
            # know information about intents or CodeBlock accesses.
            if self._defsout[sig]:
                self._killed[sig].extend(self._defsout[sig])
                self._defsout[sig] = []
            self._defsout[sig].append(reference)
```

This is probably bad both in general and for IntrinsicCall - can you try changing this to `elif isinstance(reference.parent,(Call, CodeBlock))` and see if tests fail and if your code works better? It should only be doing this for arguments, otherwise it should just be a read.

We can also do better for `IntrinsicCall` arguments I think now using their reference_accesses, but we should also fix that in is_write first probably.

sergisiso:
Does this make sense:
```
        elif reference.ancestor((Call, CodeBlock)):
            if reference.is_write:
                if self._defsout[sig]:
                    self._killed[sig].extend(self._defsout[sig])
                    self._defsout[sig] = []
                self._defsout[sig].append(reference)
```
I want to guarantee that the beginning of the chain is "is_write" (or there is not prior is_write at all), otherwise it is inconsistent with the logic on the inline transformation. This appears to have worked for my case, but I don't know if its problematic somewhere else.

LonelyCat124:
> Does this make sense:
> 
> ```
>         elif reference.ancestor((Call, CodeBlock)):
>             if reference.is_write:
>                 if self._defsout[sig]:
>                     self._killed[sig].extend(self._defsout[sig])
>                     self._defsout[sig] = []
>                 self._defsout[sig].append(reference)
> ```
> 
> I want to guarantee that the beginning of the chain is "is_write" (or there is not prior is_write at all), otherwise it is inconsistent with the logic on the inline transformation. This appears to have worked for my case, but I don't know if its problematic somewhere else.

The behaviour of the is_write here should be the same as the behaviour of `isinstance(reference.parent, (Call, CodeBlock)`, see:
```
        # pure or inquiry IntrinsicCall nodes do not write to their arguments.
        if (isinstance(parent, IntrinsicCall) and (parent.is_inquiry or
                                                   parent.is_pure)):
            return False
        # All other arguments of all other Calls are assumed to write to their
        # arguments. This could be improved in the future by looking at
        # intents where available.
        if isinstance(parent, Call):
            return True
        # The reference that is the LHS of an assignment is a write.
        if isinstance(parent, Assignment) and parent.lhs is self:
            return True
        # Assume the worst for CodeBlocks
        if isinstance(parent, CodeBlock):
            return True
```

It will still get it wrong for a read-only IntrinsicCall at the moment either way.

sergisiso:
> It will still get it wrong for a read-only IntrinsicCall at the moment either way.

I am not too worried if it is wrong (in worst-case scenario direction), I'm more worried that it could be inconsistent with "is_write" since this confuses the logic in the transformation. This is why I would prefer to use "is_write" is its the same. To make sure it is always consistent.

The good news is that it now finally pass the OpenMP offloading IT with matching results, and stats are:
```
 --- First we need to be able to modify kernels ---
Module-inline successful: 283
Module-inline failed: 16
   -> because it accesses data from its outer scope: 16

 --- Then we need to inline them, or fallback to GPU routine annotations ---
Annotation successful: 128
   Deferred-Inline successful: 73
   Deferred-Inline failed: 56
   -> Inlining polymorphic kernels is not supported: 34
   -> contains one or more Return statements: 7
   -> is passed by argument and is assigned to before the call: 14
Annotation failed: 171
   -> accesses the imported symbol: 148
   -> calls another routine: 7
   -> calls intrinsic: 21
   -> only supports the transformation of a MATMUL operation when: 1

 --- Then offload each loop with kernels inside ---
Offload independent loop: 83
Offload with dof loop: 35
Offload with atomics: 0
Offload with cell colouring: 40
Offload with cell tile-colouring: 0
Failed to offload: 3
Added inner loop nested parallelism: 0
Added OMP threading: 171
```

LonelyCat124:
> I am not too worried if it is wrong (in worst-case scenario direction), I'm more worried that it could be inconsistent with "is_write" since this confuses the logic in the transformation. This is why I would prefer to use "is_write" is its the same. To make sure it is always consistent.

Probably we could replace the non-assignment stuff with `is_write` at this point, I'm not sure it existed when i originally implemented this stuff. Can you make an issue and I can take a look at some point potentially.

sergisiso:
@arporter @LonelyCat124 This is ready to review. As mentioned above it manges to inline 73 offloaded lfric kernels, and it reports in the stats the reasons why other kernels are not inlined so it can guide future GPU work.

src/psyclone/domain/common/transformations/kernel_inline_trans.py
```diff
@@ -0,0 +1,81 @@
...
+                 node: CodedKern,
+                 options: Optional[dict[str, Any]] = None,
+                 **kwargs) -> None:
+        """Validate that the supplied kernel may be marked for inlining.
+
+        Argument-dependent and routine-body validation is deliberately
+        deferred until the complete language-level Call is constructed.
+
+        :param node: the kernel to mark for inlining.
+        :param options: a deprecated dictionary of transformation options.
```
arporter: Since this is a brand new transformation, do we need to keep `options` @LonelyCat124 ?

src/psyclone/domain/common/transformations/kernel_inline_trans.py
```diff
@@ -0,0 +1,81 @@
...
+        :raises TransformationError: if ``node`` is not a CodedKern.
+        :raises TransformationError: if the kernel PSyIR cannot be found.
+        :raises TransformationError: if the kernel is polymorphic.
+        :raises TransformationError: if the kernel implementation is not in
+            the same Container as the call site.
+
+        """
+        if not options:
+            self.validate_options(**kwargs)
+
```
arporter, line 49: For completeness, ought to call `super().validate(...)`?

src/psyclone/domain/common/transformations/kernel_inline_trans.py
```diff
@@ -0,0 +1,81 @@
...
+        """Validate that the supplied kernel may be marked for inlining.
+
+        Argument-dependent and routine-body validation is deliberately
+        deferred until the complete language-level Call is constructed.
+
+        :param node: the kernel to mark for inlining.
+        :param options: a deprecated dictionary of transformation options.
+
+        :raises TransformationError: if ``node`` is not a CodedKern.
+        :raises TransformationError: if the kernel PSyIR cannot be found.
```
arporter: It doesn't raise any of these last three.

src/psyclone/domain/common/transformations/kernel_inline_trans.py
```diff
@@ -0,0 +1,81 @@
...
+        """Mark the supplied kernel to be inlined when it is lowered.
+
+        :param node: the kernel to mark for inlining.
+        :param options: a deprecated dictionary of transformation options.
+
+        """
+        self.validate(node, options=options, **kwargs)
+        # pylint: disable=protected-access
+        # TODO #2216: This is a temporary solution, longer-term we could
+        # inline during transformation time (instead of deferring it) in
```
arporter: s/in/and/

src/psyclone/domain/common/transformations/kernel_inline_trans.py
```diff
@@ -0,0 +1,81 @@
...
+        # keep the contents in an InlinedKern node that preserve the metadata
+        node._inline = True
```
arporter: pylint might not like this as `CodedKern` doesn't have an `_inline` attribute? Probably best to add it with a suitably descriptive comment?

src/psyclone/domain/lfric/lfric_loop.py
arporter: Please update to say the "first node" if lowering results in multiple siblings. Also please update for type hints.

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -195,6 +199,11 @@ def apply(self,
         new_stmts = []
         refs = []
         for child in routine.pop_all_children():
+            if isinstance(child, (OMPDeclareTargetDirective,
+                                  ACCRoutineDirective)):
+                # Skip directives that apply to the routine, as this do not
```
arporter: s/this/these/

src/psyclone/psyir/transformations/inline_trans.py
```diff
@@ -218,10 +227,26 @@ def apply(self,
...
             self._replace_formal_args_in_expr(
                 ref, node, formal_args, routine_node=routine,
-                use_first_callee_and_no_arg_check=(
-                    use_first_callee_and_no_arg_check)
+                allow_no_args_check_if_only_one_callee=(
+                    allow_no_args_check_if_only_one_callee)
             )
 
+        # If this is inside a region that differentiates between private
+        # and shared symbols, the ones inside the callee are definetely
```
arporter: "definitely"

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
```diff
@@ -0,0 +1,229 @@
...
+from psyclone.tests.lfric_build import LFRicBuild
+from psyclone.tests.utilities import get_invoke
+from psyclone.transformations import LFRicColourTrans
+
+
+def _add_lfric_kernel_body(kern, use_map=False, statement_count=1):
+    """Add a simple executable body to an otherwise empty test kernel.
+
+    The test kernel source uses wildcard imports for the LFRic kind symbols.
+    Make these imports explicit so that InlineTrans can prove that the body
```
arporter, line 31: We ought to have an Issue for this (making all of our LFRic test Fortran follow LFRic standards by removing wildcard imports) but we don't seem to. Please could you create one and add a TODO.

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
```diff
@@ -0,0 +1,229 @@
...
+    has no unresolved outer-scope accesses.
+
+    :param kern: the module-inlined LFRic kernel.
+    :param bool use_map: whether to index the output field using its dofmap.
```
arporter, line 37: typehints in interface definition please.

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
```diff
@@ -0,0 +1,229 @@
...
+    """ in code
+
+    # Code generation lowers a copy and must preserve the DSL-level original.
+    assert isinstance(invoke.schedule.walk(CodedKern)[0], CodedKern)
+    assert invoke.schedule.walk(CodedKern)[0].inline
+    assert LFRicBuild(tmpdir).code_compiles(psy)
+
+
+def test_kernel_inline_trans_gocean(capsys):
+    """The deferred transformation is generic across CodedKern APIs."""
```
arporter, line 155: Whoop! :-)

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # The code is inlined inside the GOcean loop
+    code = str(psy.gen)
+    assert "call compute_cu_code_inlined_" not in code
+    assert """
+      do i = cu_fld%internal%xstart, cu_fld%internal%xstop, 1
+        ! Deferred-Inline successful for kernel 'compute_cu_code_inlined_'
+        cu_fld%data(i,j) = 0.5d0 * (p_fld%data(i + 1,j) + """ in code
+    assert capsys.readouterr().out == (
+        "Deferred-Inline successful for kernel 'compute_cu_code_inlined_'\n")
```
arporter, line 170: Can we test the compilation of GOcean output?

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def test_kernel_inline_trans_empty_body_does_not_skip_sibling():
+    """Removing an empty marked kernel must not skip a following kernel."""
+    _, invoke = get_invoke("4.2_multikernel_invokes.f90", "lfric",
+                           idx=0, dist_mem=False)
+    schedule = invoke.schedule
+    # This has to lfric loops with to kernels, fuse them in the first loop
```
arporter: s/to/two/g

src/psyclone/tests/psyir/transformations/inline_trans_test.py
```diff
@@ -78,6 +106,83 @@ def test_apply_empty_routine(fortran_reader, fortran_writer, tmp_path):
...
+    psyir = fortran_reader.psyir_from_source(
+        """module test_mod
+            contains
+              subroutine caller()
+                integer :: value
+                call callee(value)
+              end subroutine caller
+              subroutine callee(value)
+                integer, intent(out) :: value
+                integer :: work
```
arporter, line 144: It would be good to extend this so that the routine also reads from an imported symbol.

That then made me wonder what would happen if the routine was naughty and wrote to a global symbol. Should we attempt to catch this case (although that would mean that the original code was wrong before we did the inlining)?

src/psyclone/psyGen.py
```diff
@@ -1314,6 +1325,57 @@ def lower_to_language_level(self) -> Node:
...
+                        inlined_node = candidate
+            except (TransformationError, InternalError) as err:
+                # If inline failes, we still continue with the non-inlined
+                # version. We report the issues in stdout. Even if the most
+                # natural reporting would be to use logging, the lfric call
+                # to psyclone is hardcoded in their build system, but we
+                # want to count this errors in our gpu offloading report.
+                message = (f"Deferred-Inline failed for kernel '{self.name}' "
+                           f"due to: {err.value}")
+                print(message)
```
arporter: strictly, `..., file=sys.stdout)`

src/psyclone/psyGen.py
```diff
@@ -1314,6 +1325,57 @@ def lower_to_language_level(self) -> Node:
...
+                call_node.append_preceding_comment(message)
+                return call_node
+
+            message = f"Deferred-Inline successful for kernel '{self.name}'"
+            print(message)
```
arporter: and again here.

src/psyclone/psyGen.py
```diff
@@ -1295,13 +1305,14 @@ def node_str(self, colour: Optional[bool] = True) -> str:
         return (self.coloured_name(colour) +
                 f" {self.name}({self.arguments.names})")
 
-    def lower_to_language_level(self) -> Node:
+    def lower_to_language_level(self) -> Optional[Node]:
         '''
         In-place replacement of CodedKern concept into language level
         PSyIR constructs. The CodedKern is implemented as a Call to a
         routine with the appropriate arguments.
```
arporter: Please add that this method will attempt to inline the subroutine if the `_inline` flag is set.

arporter (changes requested):
Good stuff Sergi. It's nice that it 'just works' for GOcean too. Just minor tidying to do.

src/psyclone/domain/common/transformations/kernel_inline_trans.py
sergisiso: Ups, removed.

src/psyclone/domain/common/transformations/kernel_inline_trans.py
sergisiso: Added

src/psyclone/domain/common/transformations/kernel_inline_trans.py
sergisiso: Removed `options` unless @LonelyCat124 says otherwise

src/psyclone/domain/common/transformations/kernel_inline_trans.py
sergisiso: Done

src/psyclone/domain/lfric/lfric_loop.py
sergisiso: Done

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: Fixed

src/psyclone/psyir/transformations/inline_trans.py
sergisiso: Done

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Done

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Done

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Fixed

src/psyclone/tests/psyir/transformations/inline_trans_test.py
sergisiso: I added a local global variable to the test (this should not be private - and they aren't). Is this what you meant? Imports would not validate. 

I'm unsure about the second part, how we deal with imports hasn't changed and is already tested in: test_validate_non_local_import and test_validate_unresolved_import. Or is it something different what you are proposing?

src/psyclone/psyGen.py

src/psyclone/psyGen.py

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Done

src/psyclone/psyGen.py

sergisiso:

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
arporter: You've left them in the docstring but I can fix that if it's the only thing.

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
arporter: Second 'to' still needs replacing but I can do this.

src/psyclone/tests/psyir/transformations/inline_trans_test.py
arporter: No, I think I got confused - a written global variable should still be `shared` so it makes no difference. Thanks for extending the test.

src/psyclone/psyGen.py
```diff
@@ -1309,7 +1310,10 @@ def lower_to_language_level(self) -> Optional[Node]:
         '''
         In-place replacement of CodedKern concept into language level
         PSyIR constructs. The CodedKern is implemented as a Call to a
-        routine with the appropriate arguments.
+        routine with the appropriate arguments. If the CodedKern has the
+        'inline' flag set, it signifies that a deferred InlineTrans for this
+        kernel was requested. It will be applied it here after converting the
```
arporter: "...to it..."

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
```diff
@@ -148,7 +152,7 @@ def test_kernel_inline_trans_lfric_colouring(tmpdir):
     assert LFRicBuild(tmpdir).code_compiles(psy)
 
 
-def test_kernel_inline_trans_gocean(capsys):
+def test_kernel_inline_trans_gocean(capsys, tmpdir):
```
arporter: `tmp_path` is the new way to do this.

arporter (changes requested):
Thanks @sergisiso. It just about crossed the threshold for what I was prepared to do myself so back to you for some minor corrections.

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Done

src/psyclone/psyGen.py

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Done

src/psyclone/tests/domain/common/transformations/kernel_inline_trans_test.py
sergisiso: Done

sergisiso:
@arporter Ready for another look

arporter (approved):
All good now. Thanks for bearing with my picky requests.

arporter:
Doc failure was zenodo again.

PR #3593, opened 2026-09-10
(Closes #3503) add LFRic meta-transformation for colouring and OpenMP

### LINKED ISSUE #3503: New LFRic meta-transformation to apply colouring and OpenMP parallelism
Currently the Met Office have `psyclone_tools.py` which contains `colour_loops`:
https://github.com/MetOffice/lfric_core/blob/52b9aa7d188688b5329717b4d2486a60a3c02ee6/infrastructure/build/psyclone/psyclone_tools.py#L93-L124
and `openmp_parallelise_loops`. It would be good if PSyclone provided a meta-transformation that first coloured loops and then parallelised them. This would reduce the size of `psyclone_tools.py` and move the functionality into PSyclone, allowing it to be fully tested.

haanahfrost:
Adds `LFRicColourAndOMPTrans`, a meta-transformation that colours the loops
in an LFRic `Routine` and then applies OpenMP parallelism to them. This
replaces `colour_loops` and `openmp_parallelise_loops` from the Met Office's
`psyclone_tools.py`, so that the functionality lives in PSyclone and is
properly tested.

WIP:
- [x] Colouring half
- [x] OpenMP half
- [x] Full test coverage

### DIFF
--- a/src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
+++ b/src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
@@ -1,0 +1,50 @@
+# -----------------------------------------------------------------------------
+# SPDX-FileCopyrightText: Copyright (c) 2026 Science and Technology
+#                         Facilities Council
+# SPDX-License-Identifier: BSD-3-Clause
+# See the full LICENSE file in the project root for details.
+# -----------------------------------------------------------------------------
+
+'''This module provides a meta-transofmation that applies colouring and
+then OpenMP parallelisation to the loops in an LFRic routine.'''
+
+from psyclone.psyGen import Transformation
+from psyclone.psyir.nodes import Routine
+from psyclone.psyir.transformations.omp_parallel_trans import OMPParallelTrans
+from psyclone.transformations import LFRicColourTrans, LFRicOMPLoopTrans
+from psyclone.utils import transformation_documentation_wrapper
+
+from typing import Union
+
+
+@transformation_documentation_wrapper
+class LFRicColourAndOMPTrans(Transformation):
+    '''
+    Colours the loops in an LFRic routine and then applies OpenMP
+    parallelisation to them.
+    '''
+    _SUB_TRANSFORMATIONS = [LFRicColourTrans, OMPParallelTrans,
+                            LFRicOMPLoopTrans]
+
+    def validate(self, node: Routine,
+                 reprod: Union[bool, None] = None, **kwargs):
+        '''
+        Validates the input options of the LFRicColourAndOMPTrans.
+
+        :param node: the Routine node to transform
+        :param reprod: whether to use reproducible form of OpenMP reduction.
+        If none, the default value from the configuration is used.
+        '''
+        raise NotImplementedError("TODO #3503")
+
+    def apply(self, node: Routine, reprod: Union[bool, None] = None, **kwargs):
+        '''
+        :param node: the Routine node to transform
+        :param reprod: whether to use reproducible form of OpenMP reduction.
+        If none, the default value from the configuration is used.
+        '''
+        raise NotImplementedError("TODO #3503")
+
+
+# For Sphinx AutoAPI documentation generation
+__all__ = ["LFRicColourAndOMPTrans"]
--- a/src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
+++ b/src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
@@ -12,7 +12,7 @@ from typing import Union
 
 from psyclone.domain.lfric.lfric_constants import LFRicConstants
 from psyclone.psyGen import Transformation
-from psyclone.psyir.nodes import Routine, Loop, ProfileNode
+from psyclone.psyir.nodes import Directive, Loop, ProfileNode, Routine
 from psyclone.psyir.transformations.omp_parallel_trans import OMPParallelTrans
 from psyclone.psyir.transformations.transformation_error import (
     TransformationError)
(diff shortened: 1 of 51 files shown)


src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
```diff
@@ -0,0 +1,122 @@
...
+    '''
+    Colours the loops in an LFRic routine and then applies OpenMP
+    parallelisation to them.
+    '''
+    _SUB_TRANSFORMATIONS = [LFRicColourTrans, OMPParallelTrans,
+                            LFRicOMPLoopTrans]
+
+    def validate(self, node: Routine,
+                 reprod: Union[bool, None] = None, **kwargs):
+        # pylint: disable=arguments-renamed
```
LonelyCat124, line 34: can you add a TODO around this pylint as if options are removed it should no longer be necessary.

src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
```diff
@@ -0,0 +1,122 @@
...
+        '''
+        Validates the input options of the LFRicColourAndOMPTrans.
+
+        :param node: the Routine node to transform
+        :param reprod: whether to use reproducible form of OpenMP reduction.
+        If none, the default value from the configuration is used.
```
LonelyCat124, line 40: This line of docstring should be indented.

src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
```diff
@@ -0,0 +1,122 @@
...
+                f"the supplied Routine '{node.name}' already contains a "
+                f"ProfileNode.")
+
+    def _colour_loops(self, node: Routine, **colour_kwargs):
+        '''
+        Applies colouring to every loop over cell-columns that is on a
+        continuous function space.
+
+        :param node: the Routine whose loops are to be coloured.
+        :param colour_kwargs: keyword arguments for the LFRicColourTrans
```
LonelyCat124, line 65: If we're documenting this it needs to be type hinted - I think it should be `dict[str, Any]` but double check. Also for _parallelise_loops

src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def _parallelise_loops(self, node: Routine,
+                           reprod: Union[bool, None] = None, **par_kwargs):
+        '''
+        Applies OpenMP parallelisation to every loop that is not a loop
+        over colours and it not already a directive.
+
+        :param node: the Routine whose loops are to be parallelised.
+        :param reprod: whether to use reproducible form of OpenMP reduction.
+        If none, the default value from the configuration is used.
```
LonelyCat124, line 85: Indent this line please.

src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
```diff
@@ -0,0 +1,122 @@
...
+        :param par_kwargs: keyword arguments for the OMPParallelTrans.
+        '''
+        otrans = LFRicOMPLoopTrans()
+        oregtrans = OMPParallelTrans()
+
+        # TODO #2668: LFRicOMPLoopTrans has not yet been migrated to kwargs. It
+        # it gives options inherrited from OMPLoopTrans but discards any
```
LonelyCat124, line 92: s/inherrited/inherited

src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
```diff
@@ -0,0 +1,122 @@
...
+                continue  # Skip if an outer loop is already parallelised
+            oregtrans.apply(loop, **par_kwargs)
+            otrans.apply(loop, options=options)
+
+    def apply(self, node: Routine, reprod: Union[bool, None] = None, **kwargs):
+        # pylint: disable=arguments-renamed
+        '''
+        :param node: the Routine node to transform
+        :param reprod: whether to use reproducible form of OpenMP reduction.
+        If none, the default value from the configuration is used.
```
LonelyCat124, line 110: Indent this line.

src/psyclone/tests/domain/lfric/transformations/lfric_colour_and_omp_trans_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def test_apply_skips_loop_already_in_directive():
+    ''' Check that a loop that is already inside a directive is not
+    parallelised a second time. '''
+    _, invoke = get_invoke("1_single_invoke_w3.f90", TEST_API, idx=0,
+                           dist_mem=False)
+    sched = invoke.schedule
+    # w3 is discontinuous so no colouring happens. Parallelise the loop by
+    # hand first so that the transformation finds it already in a directive.
```
LonelyCat124, line 110: What happens if we have a continuous space and we already parallelise the loop? Can you test that behaviour as well?

LonelyCat124 (changes requested):
@haanahfrost Couple of change, most are just documentation or type hint related.
I've asked for one additional thing in the tests because I'm not entirely sure I know what will happen and I'd like it to be explicit in the tests as well.

src/psyclone/domain/lfric/transformations/lfric_colour_and_omp_trans.py
haanahfrost: checked this with mypy. For `**kwargs` the annotation applies to each value rather than the dict itself so `**colour_kwargs: Any` reveals as `dict[str, Any]`
Annotating it `dict[str, Any]` gives `dict[str, dict[str, Any]]` and mypy rejects passing tiling=True.
Gone with `Any` for both `_colour_loops` and `_parallelise_loops` which also matches `lfric_invoke_schedule.py`

src/psyclone/tests/domain/lfric/transformations/lfric_colour_and_omp_trans_test.py
haanahfrost: Tested this and it silently does nothing. Once `OMPParallelTrans` has been applied the loop is no longer a direct child of the Routine (children go from `['LFRicLoop']` to `['OMPParallelDirective']`), so `_colour_loops` never sees it, and `_parallelise_loops` skips it because it has a `Directive` ancestor. This results in an uncoloured loop on a continuous space inside a parallel region, which is the race condition colouring is there to prevent.

Do you want `validate` to reject this? (raise if any loop with a `Directive` ancestor is on a continuous space)
Or should the test just document current behaviour?

LonelyCat124 (changes requested):
@haanahfrost One thing left I forgot last time and one other change - can you add this transformation into the user guide (`doc/user_guild/transformations.rst`) in the long list of transformations.

Also I think `examples/lfric/eg3/colouring_and_omp.py` can be updated to use this new metatransformation - can you give that a try as well? Edit: Also double check the readme/docs in the eg3 folder and the description in the base lfric examples folder.

LonelyCat124 (approved):
Looks good, I'm going to change one thing I got wrong in the last review and then set the ITs going.

PR #3594, opened 2026-09-10
Added support for providing multiple scripts with psyclone

schreiberx:


sergisiso:
I am happy to review this one (once the tests pass) but @arporter @LonelyCat124 @hiker haven't we discussed this one before (having multiple `-s script1.py -s script2.py`)? I can't find an issue but I would have sworn we had it and we agreed it was a good idea.

hiker:
> I am happy to review this one (once the tests pass) but @arporter @LonelyCat124 @hiker haven't we discussed this one before (having multiple `-s script1.py -s script2.py`)? I can't find an issue but I would have sworn we had it and we agreed it was a good idea.

Not that I remember. I am a big fan of using more OO approach (i.e. one script inheriting from another) - which we can do already (i.e. `trans` just creating an instance ...), and we have discussed hierarchical search paths in the context of the build system (e.g. when we decided not to integrate the psyclone yaml specification from lfric).

What is the use case of multiple script - especially what can be achieved with that that we can't already do? FWIW,  I already have an example in my tutorial (e.g. `tutorial/training/gocean/2.10-GameOfLife-mpi/solution/openmp.py`) where I use one transformation script in another:
```
from copy_kernels_and_fuse_loops import trans as fuse_trans
...
def trans(psyir: FileContainer)
...
    # Call the existing fuse transformation, which will also module inline.
    fuse_trans(psyir)
...
```
Together with arguments for scripts (which can e.g. be used to specify which scripts to tun), I can't see anything that multiple scripts can do that we can't already do, or am I missing something?

I don't really object, but am a big fan of KISS :)

hiker:
And I just notice, how does this work together with keyword arguments? I assume they would be provided for both scripts?

hiker:
> And I just notice, how does this work together with keyword arguments? I assume they would be provided for both scripts?

Ah, I had a quick look now (sorry, should have checked before writing) - each script get its own kwargs.

schreiberx:
> > I am happy to review this one (once the tests pass) but @arporter @LonelyCat124 @hiker haven't we discussed this one before (having multiple `-s script1.py -s script2.py`)? I can't find an issue but I would have sworn we had it and we agreed it was a good idea.
> 
> Not that I remember. I am a big fan of using more OO approach (i.e. one script inheriting from another) - which we can do already (i.e. `trans` just creating an instance ...), and we have discussed hierarchical search paths in the context of the build system (e.g. when we decided not to integrate the psyclone yaml specification from lfric).
> 
> What is the use case of multiple script - especially what can be achieved with that that we can't already do? FWIW, I already have an example in my tutorial (e.g. `tutorial/training/gocean/2.10-GameOfLife-mpi/solution/openmp.py`) where I use one transformation script in another:
> 
> ```
> from copy_kernels_and_fuse_loops import trans as fuse_trans
> ...
> def trans(psyir: FileContainer)
> ...
>     # Call the existing fuse transformation, which will also module inline.
>     fuse_trans(psyir)
> ...
> ```
> 
> Together with arguments for scripts (which can e.g. be used to specify which scripts to tun), I can't see anything that multiple scripts can do that we can't already do, or am I missing something?
> 
> I don't really object, but am a big fan of KISS :)

You're asking why this would be nice? Here we go:

1. I would first like to use Poseidon to rewrite the code.
2. Then I like to use, e.g., the OpenACC or OpenMP backend of PSyclone.
3. Maybe I would like to do some further injection of debugging features in between or after it.

For this, I don't like to mess around with calling scripts from other scripts, but just pipeline all of them on the command line.

sergisiso:
Another reason is that scripts are not always properly packaged to guarantee that imports always work.

This is not now we use NEMO, but there is a way to let the NEMO build system call psyclone that copies the scripts from the source directory to a build directory with a different structure. Currently any "imports" in those scripts become broken. You could say that it is the build system problem, but the problem also goes away if they are "independent" scripts that are applied one after the other.

In any case this doesn't remove the current functionality, it just provides an alternative.

sergisiso:
@schreiberx Have a look at the coverage misses before I do the review.

We can also have a discussion in the next telco to make sure everybody agrees.

hiker:
> You're asking why this would be nice? Here we go:
> 
> 1. I would first like to use Poseidon to rewrite the code.
> 2. Then I like to use, e.g., the OpenACC or OpenMP backend of PSyclone.
> 3. Maybe I would like to do some further injection of debugging features in between or after it.
> 
> For this, I don't like to mess around with calling scripts from other scripts, but just pipeline all of them on the command line.

One alternative to a PR:
```
#!/usr/bin/env python3

import sys
from psyclone.generator import load_script

def trans(psyir: FileContainer, scripts: list[str]):
    for script in scripts:
        print("LOADING", script)
        transformation_recipe, _, _, _ = load_script(script)
        transformation_recipe(psyir)
        del sys.path[0]
```
Usage:
```
psyclone  ... -s ./multi_script.py   --script-kwargs scripts:'["copy_kernels_and_fuse_loops.py", "openmp.py"]'
LOADING copy_kernels_and_fuse_loops.py
LOADING openmp.py
```
Though this of course needs some additional work to support different command line options per script, and to use the `files_to_skip` option.

TBH, in your case, I would just use command line options (`"parallel='openmp',debug=['phase1', 'phase3']"`).

Re Sergi's comment about the old-style(?) NEMO build system moving files around: since this is using the PSyclone loader, I think this should work with whatever NEMO is doing there.

I don't object to adding this, but I feel that PSyclone is moving away from it's core ability, and instead handles issues with build systems, script design etc. (and the above `del sys.path[0]` actually shows why `load_script` needs some refactoring ;) ).

sergisiso:
What do @arporter @LonelyCat124 think about this? I still find it relatively natural that a psyclone command accepts multiple `-s`.

@schreiberx We still need some clarity/tests about what happens with global script options: https://psyclone.readthedocs.io/en/latest/user_guide/user_scripts.html#script-global-variables
Imo, FILES_TO_SKIP should only apply to the script that have them, but RESOLVE_IMPORTS (which affects the parsing) should accomulate, with True having priority above others, and otherwise aggregating a set of all modules listed in any scripts.

schreiberx:
I think that neither `FILES_TO_SKIP`, nor `RESOLVE_IMPORTS` should be used here. This should be part of the transformation script. But that should be part of a different PR. As an interim solution, they should all be accumulated until a better solution is found.

E.g., to avoid this, a script could look like this:
```
def trans(psyir, control: PsycloneControl):
    # Could also be done through .ancector(FileContainer)
    if control.get_filename() in ["asdf.f90", "asdf.f90"]:
        return        # Skip processing this file

    # Resolve all imports for the current psyir file
    control.resolve_imports(psyir)
```

sergisiso:
@schreiberx This is not possible:

> if control.get_filename() in ["asdf.f90", "asdf.f90"]:
>        return        # Skip processing this file

This is already done when the issue is the script, the FILES_TO_SKIP is to skip files where the issue is in the frontend or backend, so an empty passthrough (skipping the script) also generated invalid code.

> #Resolve all imports for the current psyir file
> control.resolve_imports(psyir)

This works but it is much slower (because changing symbol information after parsing needs to traverse the tree updating some nodes with the new information to keep things consistent), also some logic during the frontend, e.g. resolving WHERE constructs is worse without the symbol information.

schreiberx:
OK. This now makes all sense to me.

Then, concatenating all FILES_TO_SKIP and OR-ing RESOLVE_IMPORTS would be the approach we'd go for?

sergisiso:
> Then, concatenating all FILES_TO_SKIP and OR-ing RESOLVE_IMPORTS would be the approach we'd go for?

Yes for Or-ing (or Union-ing is Set notation terms), with a 'True' taking precedence for RESOLVE_IMPORTS.

src/psyclone/generator.py
```diff
@@ -933,16 +966,33 @@ def code_transformation_mode(input_file: str,
...
+    for script_name, kwargs_str in script_kwargs_pairs:
+        if script_name:
+            (trans_recipe, files_to_skip,
+             resolve_mods, kwargs, inserted_path) = load_script(
+                 script_name, kwargs_str)
+            if trans_recipe:
+                trans_recipes.append((trans_recipe, kwargs))
+            all_files_to_skip.extend(files_to_skip)
+            if resolve_mods:
+                final_resolve_mods = resolve_mods
```
schreiberx, line 988: @sergisiso "Or-ing" is done here. It could also be a one-liner, but I prefer this one (old school).

src/psyclone/generator.py
```diff
@@ -933,16 +966,33 @@ def code_transformation_mode(input_file: str,
...
+    # Load all script files
+    trans_recipes = []
+    for script_name, kwargs_str in script_kwargs_pairs:
+        if script_name:
+            (trans_recipe, files_to_skip,
+             resolve_mods, kwargs, inserted_path) = load_script(
+                 script_name, kwargs_str)
+            if trans_recipe:
+                trans_recipes.append((trans_recipe, kwargs))
+            all_files_to_skip.extend(files_to_skip)
```
schreiberx, line 986: @sergisiso  Files to skip are appended here.

schreiberx:
I think we're ready to go.

Issue #3590, opened 2026-09-09
LoopFuseTrans (and subclasses) my not work correctly with TransformationDocumentationWrapper


LonelyCat124:
I seem to remember the documentation wrapper (and related suites) doing stuff related on there being one `node` or `node_list` argument, however these appear to take `node1, node2` so I need to check the behaviour at some point.

PR #3595, opened 2026-09-10
(Towards/Close #3315, Closes #3590) New implementation of lfric loop fuse trans

### LINKED ISSUE #3315: Allow loop fusion of builtins on the same ANY_SPACE_

LonelyCat124:
Initial implementation for the new lfric loop fuse trans setup.
There are still for sure some things to sort out but it seems to work for our tests and solves ANY_SPACE things.

I'd like to have an investigate of being able to fuse more than 2 LFRicLoops, provided they all meet some criteria (i.e. some_space is provided, they're all on the same field etc., or that they all have the same space at runtime) but I'd need to know if that is useful @christophermaynard @MetBenjaminWent ? Its probably a more substantial implementation but I think its possible.

### DIFF
--- a/src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
+++ b/src/psyclone/domain/lfric/transformations/lfric_loop_fuse_trans.py
@@ -10,6 +10,12 @@
 
 from psyclone.core.access_type import AccessType
 from psyclone.domain.lfric import LFRicConstants, LFRicLoop
+from psyclone.psyGen import BuiltInCall
+from psyclone.psyir.nodes import (
+    ArrayOfStructuresReference, BinaryOperation, Call, IfBlock,
+    StructureReference, Literal
+)
+from psyclone.psyir.symbols import ScalarType
 from psyclone.psyir.transformations import LoopFuseTrans, TransformationError
 from psyclone.transformations import check_intergrid
 from psyclone.utils import transformation_documentation_wrapper
@@ -50,11 +56,6 @@ class LFRicLoopFuseTrans(LoopFuseTrans):
         :param node2: the second Loop to fuse.
         :param options: a dictionary with options for transformations.
         :type options: Optional[Dict[str, Any]]
-        :param bool options["same_space"]: this optional flag, set to `True`, \
-            asserts that an unknown iteration space (i.e. `ANY_SPACE`) \
-            matches the other iteration space. This is set at the user's own \
-            risk. If both iteration spaces are discontinuous the loops can be \
-            fused without having to use the `same_space` flag.
 
         :raises TransformationError: if either of the supplied loops contains \
                                      an inter-grid kernel.
@@ -64,9 +65,6 @@ class LFRicLoopFuseTrans(LoopFuseTrans):
                                      does not apply because neither field \
                                      is on `ANY_SPACE` or the spaces are not \
                                      the same.
-        :raises TransformationError: if one or more of the iteration spaces \
-                                     is unknown (`ANY_SPACE`) and the \
-                                     `same_space` flag is not set to `True`.
         :raises TransformationError: if the loops are over different spaces \
                                      that are not both discontinuous and \
                                      the loops both iterate over cells.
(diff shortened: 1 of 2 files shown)


sergisiso:
Here is the example of min and max builtins that could be easy targets to fuse: `src/psyclone/tests/test_files/lfric/15.10.9_min_max_X_builtin.f90`

arporter:
Some notes from our meeting.
I think the check list for whether or not two kernels can be fused is:
 * Do the two kernels `OPERATE_ON` the same thing (dofs or cell columns)?
 * Is their iteration space determined by the same field/operator?
 * Or, do their iteration spaces correspond to the same function space (static check, will need to resolve 'ANY_SPACE' to an actual space)?
 * [Optionally] do the fuse but protect it with a runtime check on the function spaces.
 
We think most of the the benefit will come from fusing built-ins (particularly those that do MIN/MAX on the same field).
For user-provided kernels that operate on cell-columns, I *think* they can be fused following the rules above, provided that the DA doesn't determine that a halo-exchange is required between them. (This allows for e.g. a kernel that increments a field on W2 that is also written to in the first kernel and also for stencil accesses.) @christophermaynard  thinks the opportunities for doing this kind of fusion will be very limited. 

We also discussed 'constant propagation' but realised that often, a `setval_c` is used to initialise a field that is then incremented by a kernel that loops over cell-columns. Since, for continuous function spaces, different cell-columns will increment the same dof it will not be possible to do 'constant propagation' in this case.

LonelyCat124:
I think the new implementation does most of this, what I couldn't work out is why is this necessary:
```
Is their iteration space determined by the same field/operator?
```

This can be either:
1. to save searching for what the space is if they're both built ins?
2. If we have 2 any_space `Kern`s without a non-any_space kern on the field to be able to work out the iteration space?

Does that make sense @arporter ?

I am also unsure I've done anything re: " I think they can be fused following the rules above, provided that the DA doesn't determine that a halo-exchange is required between them". I'll have to work that out next.

arporter:
> This can be either:
> 
> 1. to save searching for what the space is if they're both built ins?
> 2. If we have 2 any_space `Kern`s without a non-any_space kern on the field to be able to work out the iteration space?

Yes, sorry, I was a bit slack in how I phrased it. If the iteration space is defined by the same field/operator then we don't have to look at function spaces. We probably still need to worry about stencil accesses. In fact, now that I write that, we need to be careful about kernels that update more than one argument (they are permitted to do this). A first step might be to refuse to fuse such cases.

LonelyCat124:
I *think* I've added a check in validate for writing to only one field (I think).
I think it would be good to have more things in validate than apply, however there are challenges due to the different behaviour with `conditional_fusion`, which we need a lot of information to determine whether its required or not.

LonelyCat124:
I think I have most of the basics done now, but there are some remaining hurdles that I'm not sure I understand.

The first is this test (not implemented in the pushed version) that fails, but I think we wanted to be able to do:
```
def test_loop_fuse_min_max_same_var():
    '''Test that we fuse two builtins on the same field without any
    other invoke elements.'''
    _, invoke = get_invoke(
        "15.10.9_min_max_X_builtin.f90",
        TEST_API, name="invoke_0", dist_mem=False)
    schedule = invoke.schedule
    ftrans = LFRicLoopFuseTrans()

    ftrans.apply((schedule.children[0], schedule.children[1]))
    ftrans.apply((schedule.children[0], schedule.children[1]))
```

This fails due to an existing test in validate, where we aren't allowed to fuse 2 reduction kernels (we have a setval_c, a max and a min in the input invoke).
Is there an appropriate way to determine whether this should be allowed?

The other thing I'm not sure about was working out how to avoid fusing loops if there is some dependency between them. I assume I can check if `loop1.forward_dependency == loop2` (or something similar)? I'm also unclear how to test that case.

@arporter @sergisiso If you have any idea about either of these let me know.

LonelyCat124:
I've allowed reductions to be fused if the iteration space variable is the same - not sure if this is correct in general?

PR #3599, opened 2026-09-11
(Closes #3596, closes #3195) Fix for creating Intrinsics to imported symbols that shadow intrinsic names

### LINKED ISSUE #3195: Handling existing code that overrides Fortran intrinsics
We know that both NEMO and NEMOVAR override (different) Fortran intrinsics. PSyclone simply does not handle this situation correctly at present.

When attempting to match an intrinsic function call, fparser *does* check that it hasn't already got a Symbol defined with the same name.
If it has, then it does not match:
https://github.com/stfc/fparser/blob/eee1f4fd40b0b339e6190b709b853227e5b2d857/src/fparser/two/Fortran2003.py#L12526-L12536

However, if an application is using wildcard imports then there will be no Symbol already defined and fparser will match an intrinsic function call unless we are lucky and the overridden version has an argument count inconsistent with the standard-defined version. If fparser does match an intrinsic function then PSyclone simply converts this into an IntrinsicCall node - fparser2._intrinsic_handler does not check for any existing Symbols of the same name:
https://github.com/stfc/PSyclone/blob/c9c20b1ee96c10352b31463276408ad33ab84752/src/psyclone/psyir/frontend/fparser2.py#L4994

A relatively simple(?) first step would be to extend `fparser2._intrinsic_handler` to attempt to lookup a symbol with the same name as the intrinsic. If it finds one, then it would have to call the appropriate handler (e.g. Call or ArrayReference). This would work provided:
 1. The symbol is not a transitive import;
 2. We have the originating module listed in `RESOLVE_IMPORTS` in the transformation script.

### LINKED ISSUE #3596: AttributeError: 'GenericInterfaceSymbol' object has no attribute 'intrinsic' bug
The problem is related to `RESOLVE_IMPORTS = True`.

Only if this is enabled in trans.py, the error shows up:

`AttributeError: 'GenericInterfaceSymbol' object has no attribute 'intrinsic'`

A reproducer is attached.

[reproducer.zip](https://github.com/user-attachments/files/32065563/reproducer.zip)

LonelyCat124:


LonelyCat124:
@arporter @sergisiso This should be ready for review. I'll set the ITs going.

arporter:
This may also fix or be towards #3195.

src/psyclone/psyir/frontend/fparser2.py
arporter, line 5428: Pls update to use type hints.

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -5436,6 +5436,20 @@ def _intrinsic_handler(self, node, parent):
...
         try:
+            # If we already have a RoutineSymbol with the name of the supposed
+            # intrinsic, then it is shadowed by the declared RoutineSymbol and
+            # we should instead create a Call.
+            existing_symbol = parent.scope.symbol_table.lookup(
+                node.items[0].string, otherwise=None
+            )
+            if (existing_symbol is not None and
+                    isinstance(existing_symbol, RoutineSymbol)):
+                call = Call(parent=parent)
```
arporter, line 5445: I think it would be better to call `_call_handler(node, parent)` at this point rather than duplicate what it does.

src/psyclone/tests/psyir/frontend/fparser2_intrinsic_handler_test.py
```diff
@@ -318,3 +322,44 @@ def test_handling_nested_intrinsic():
...
+        integer:: a, b
+
+        b = real(a)
+    end subroutine test
+    end module"""
+    fortran_reader = FortranReader(resolve_modules=["mod_a"])
+    monkeypatch.setattr(Config.get(), "_include_paths", [tmp_path])
+
+    psyir = fortran_reader.psyir_from_source(code)
+    # abs is imported from mod_a so we get a Call instead of an
```
arporter: s/abs/real/ ?

src/psyclone/tests/psyir/frontend/fparser2_intrinsic_handler_test.py
```diff
@@ -318,3 +322,44 @@ def test_handling_nested_intrinsic():
...
+    # IntrinsicCall
+    assert len(psyir.walk(IntrinsicCall)) == 0
+    test = psyir.walk(Routine)[0]
+    assert type(test.children[0].rhs) is Call
```
arporter, line 365: Just for safety, please could you check that the target of the call is a RoutineSymbol with the expected name.

arporter (commented):
Thanks very much for fixing this Aidan. To save on duplication, I think the intrinsic-call handler should call the standard call handler when it encounters this situation. Once that's done (provided it works), I'll check the ITs.

arporter:
I got an error message from GH when I tried to submit my review as a 'request changes' so I did it as a 'comment' instead.

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: I mentioned on teams, an `Intrinsic_Function_Reference` doesn't share enough structure with a `Call_Stmt` for this to work - when I attempted it the code broke. It seems to result in  the creation of a Codeblock when attempting to create the first child node of the Call, which must be a Reference.

src/psyclone/tests/psyir/frontend/fparser2_intrinsic_handler_test.py
LonelyCat124: Fixed.

src/psyclone/tests/psyir/frontend/fparser2_intrinsic_handler_test.py
LonelyCat124: Added

src/psyclone/psyir/frontend/fparser2.py
LonelyCat124: Done.

LonelyCat124:
@arporter Should be ready for another look.

src/psyclone/psyir/frontend/fparser2.py
arporter: Oh yes, I should have realised that. Thanks for adding the comment.

arporter (approved):
Thanks Aidan, this all looks good now.
I'll proceed to merge.

Issue #2666, opened 2024-07-19
Copying a tree containing shadowed symbols may change semantics


arporter:
`ScopingNode._refine_copy()` uses `replace_symbols_using` to ensure that Symbols referred to in a sub-tree are updated. However, since `replace_symbols_using` only uses symbol *names*, this won't get the correct symbol if the PSyIR happens to have symbols shadowed in nested scopes. In practise, this isn't often a problem but it is easy to imagine a situation where it will break, e.g.:

      subroutine test
        integer :: a
        integer :: b = 1
        if condition then
          ! PSyIR declares a shadowed, locally-scoped a'
          a' = 1
          if condition2 then
            ! PSyIR declares a shadowed, locally-scoped b'
            b' = 2
            a = a' + b'

Here, the final assignment will end up being `a' = a' + b'` and thus the semantics of the code are changed. (Note that Fortran code is only used to illustrate the problem - it's not possible to create such a situation directly from actual Fortran because it doesn't support nested scopes*.)

* apart from ASSOCIATE constructs and the PSyIR doesn't support those yet.

Issue #2331, opened 2023-09-24
problems with SymbolTable resolve_imports()


rupertford:
I think `resolve_imports()` needs improving and can lead to incorrect code in its current form. It is currently used in cases where there are subroutines with no use statements and a parent module with wildcard use statements and it attempts to determine the datatype information of variables that are declared in these use statements, not declared in the subroutines. It has had to be extended a little as variables without declarations were previously declared in the module symbol table as Symbols but are now declared in the subroutine symbol table as Symbols.

The problems will occur when the symbol should actually be declared in the subroutine symbol table e.g. because there is a wilcard use statement there for the symbol, or because its declaration is hidden via an UnknownFortranType or something similar.

I would suggest that `resolve_imports()` is not just limited to module symbol tables (which it seems to be) and does not just act on the current symbol table, but is run recursively up the hierarchy of symbol tables from bottom to top. I guess this latter bit could be a different method/function. That would allow symbols to be resolved with the correct use statements.

However, there are likely to be cases when it is not possible to determine what the datatype of a variable is and these symbols should be left as they are (I think), which is not the case now as a new datasymbol is currently added to the module symbol table when there is already a symbol with the same name in the subroutine

A simple example of a current problem is:

```
module mod1
  avar=1
end module
module mod2
  avar=2
end module
module work
use mod1
contains
subroutine compute()
use mod2
print *,avar
end subroutine
end module
```

In the above case we end up with a symbol in the subroutine symbol table and a datasymbol in the module symbol table which would give the wrong value if the backend allowed it to be written to Fortran.

Issue #1540, opened 2021-12-07
Resovle symbol information from nested wildcard imports


sergisiso:
Currently we only resolve wildcard imports from directly referenced modules. We don't fail if we don't find a symbol (in fact we don't know for sure it must be there), we just fill as much information as we find.

In reality we could have nested wildcard imports and we would need to traverse all dependency tree looking for information.

PR #3600, opened 2026-09-11
(closes #3598) SymbolTable hardening

### LINKED ISSUE #3598: Audit SymbolTable implementation
(requested by @arporter done by Codex)

`SymbolTable` is central to PSyIR correctness: each `ScopingNode` owns one, frontends populate it, transformations merge and mutate it, and backends use it to generate declarations and `USE` statements. The overall design is sound, but I found several uncovered cases that can silently corrupt symbol identity or produce incorrect import information.

### How it is used

- Every PSyIR scope creates or attaches a table in [scoping_node.py](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/nodes/scoping_node.py:38).
- The Fortran frontend populates symbols and formal arguments, e.g. [fparser2.py](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/frontend/fparser2.py:2874).
- Transformations create temporaries, rename symbols, and merge scopes. Inlining depends directly on `merge()` in [inline_trans.py](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/transformations/inline_trans.py:201).
- The Fortran backend uses symbols for declarations and import renaming in [fortran.py](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/backend/fortran.py:484).
- PSyIR references retain symbol-object identity, not just names. Analyses and transformations therefore depend on references and tables agreeing.

## Confirmed problems

1. **Critical: import resolution can corrupt symbols in sibling or shadowing scopes.**

   In `_update_with_resolved_symbol()`, a matching unresolved descendant triggers:

   [symbol_table.py:1934](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1934)

   ```python
   self.node.replace_symbols_using(symbol)
   ```

   This replaces matching names throughout the entire outer scope. Replacement with a single symbol is name-based ([reference.py:388](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/nodes/reference.py:388)), so it ignores lexical shadowing.

   I reproduced a module containing:

   - routine `s1` with local integer `x`;
   - routine `s2` with unresolved `x`;
   - outer wildcard import providing `x`.

   After resolution, `s1`’s table still contained its local `x`, but its `Reference` pointed to the imported outer `x`. This creates internally inconsistent PSyIR and can mislead dependency analysis and transformations.

2. **High: ambiguous wildcard imports produce a hybrid symbol instead of an error.**

   When an existing local symbol already has an `ImportInterface`, `_import_symbols_from()` accepts it without checking which container it comes from:

   [symbol_table.py:1830](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1830)

   I reproduced wildcard imports from modules `a` and `b`, both defining `x` with different types. After resolving both, the resulting symbol:

   - retained its interface saying it came from `a`;
   - acquired the datatype from `b`.

   This should be treated as ambiguous or verified to be the same origin.

3. **High: formal-argument invariants can easily be broken.**

   `_validate_arg_list()` checks only type and interface, not table membership or uniqueness:

   [symbol_table.py:1493](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1493)

   Consequently:

   - `specify_argument_list([arg, arg])` accepts duplicates.
   - It accepts an argument that is not present in the table.
   - `argument_list` returns the mutable internal list directly.
   - `remove(arg)` leaves the removed symbol in `_argument_list`.
   - `insert_argument()` and `append_argument()` mutate `_argument_list` before operations that may fail ([symbol_table.py:1444](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1444)).

   `merge()` also adds symbols with `ArgumentInterface` but does not merge argument-list metadata. Merging a valid table containing one argument produced a destination whose `argument_list` property raised `InternalError`.

4. **High: copying an aliased import loses its original name.**

   `copy_external_import()` replaces the copied interface with:

   [symbol_table.py:1721](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1721)

   ```python
   ImportInterface(container_ref)
   ```

   It does not preserve `orig_name`. Thus `local => remote` becomes an import of `local`. The backend explicitly relies on `orig_name` to generate `local=>remote` ([fortran.py:487](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/backend/fortran.py:487)).

   This path is used by [transformations.py:1950](/home/centos/workspace/psyclone/PSyclone/src/psyclone/transformations.py:1950), so it can affect generated code. Constructing a fresh `ContainerSymbol` also loses properties such as `is_intrinsic` and any custom interface.

5. **High: targeted resolution of renamed imports fails.**

   With `use a, only: local => remote`, calling `resolve_imports(symbol_target=local)` compares the target’s local name with the external symbol’s original name:

   [symbol_table.py:1806](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1806)

   I confirmed that this raises “target symbol `local` was not found”, although ordinary untargeted resolution succeeds.

6. **Medium: failed `add()` can leave a dangling tag.**

   `add()` inserts the tag before validating the import interface:

   [symbol_table.py:591](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:591)

   If import validation then fails, the symbol is absent from `_symbols`, but `lookup_with_tag()` still returns it. Tags are also not type-checked; non-string truthy tags can be inserted but cannot later be retrieved through `lookup_with_tag()`.

7. **Medium: APIs claiming to accept `Iterable` mishandle generators.**

   `resolve_imports()` validates `container_symbols` by iterating over it and then iterates it again ([symbol_table.py:1964](/home/centos/workspace/psyclone/PSyclone/src/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1964)). A generator is exhausted during validation, so no imports are resolved.

   Similarly, `merge()` and `check_for_clashes()` consume `symbols_to_skip` more than once. With a clashing skipped symbol supplied through a generator, I observed that it was renamed and merged instead of skipped.

   Additionally, `symbols_to_skip` is ignored for `ContainerSymbol`s because container merging is unconditional at [symbol_table.py:1022](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1022).

8. **Medium: tag multiplicity contradicts internal assumptions.**

   `copy_external_import()` can associate additional tags with an existing symbol at [symbol_table.py:1737](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1737). However:

   - `get_reverse_tags_dict()` assumes one tag per symbol and silently retains only the last ([symbol_table.py:1564](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1564)).
   - `swap()` preserves only the first tag it finds.
   - Transformations using the reverse map can therefore lose metadata.

## Existing acknowledged weaknesses

The source already documents several related limitations:

- `remove()` does not generally check PSyIR references, TODO #898 ([symbol_table.py:1319](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:1319)).
- Nested imports are not recursively processed, TODO #1540.
- Equality ignores tags, argument order and visibility, TODO #1698 ([symbol_table.py:2252](/home/centos/workspace/psyclone/PSyclone/src/psyclone/psyir/symbols/symbol_table.py:2252)).
- Name-based replacement during copied-scope handling is already documented as unsafe with shadowing, TODO #2666.
- The current test suite contains an expected failure for child-scope wildcard resolution, issue #2331.

## Recommended priority

1. Make descendant import resolution identity- and scope-aware.
2. Detect conflicting import provenance before copying symbol properties.
3. Define and enforce a single table invariant covering argument membership, uniqueness, tags and import containers.
4. Make all mutations validate-before-commit or roll back on failure.
5. Preserve complete import metadata, including `orig_name` and intrinsic-container properties.
6. Materialise iterable arguments once and honour `symbols_to_skip` consistently.
7. Return copies or read-only views for `argument_list`, `symbols_dict`, and `tags_dict`.

The existing symbol-table tests pass: **107 passed, 1 expected failure**. The reproduced cases above are not covered by them.

sergisiso:
- Scope-aware import resolution now preserves shadowed symbols.
- Conflicting wildcard imports raise SymbolError; legitimate re-exports remain supported.
- Renamed imports preserve orig_name, intrinsic-module status, and container metadata.
- Generator arguments work correctly in merge() and resolve_imports().
- Merge now preserves arguments and tags and honours skipped containers.
- Argument lists enforce membership and uniqueness, return copies, and roll back failed mutations.
- Tags are validated, one-to-one, transactional, and managed through add_tag()/remove_tag().
- Public symbol/tag mappings are read-only.
- swap() performs identity/import checks and preserves formal arguments.
- Equality now considers visibility, tags, and argument ordering.
- Visibility filters are fully validated.

### DIFF
--- a/src/psyclone/psyir/frontend/fparser2.py
+++ b/src/psyclone/psyir/frontend/fparser2.py
@@ -388,7 +388,7 @@ def _find_or_create_psyclone_internal_cmp(node):
             container.symbol_table.add(symbol)
             symbol = container.symbol_table.lookup(name_interface)
             # Add the appropriate tag to find it regardless of the name
-            container.symbol_table.tags_dict['psyclone_internal_cmp'] = symbol
+            container.symbol_table.add_tag(symbol, 'psyclone_internal_cmp')
             return symbol
 
     raise NotImplementedError(
--- a/src/psyclone/psyir/symbols/symbol_table.py
+++ b/src/psyclone/psyir/symbols/symbol_table.py
@@ -15,6 +15,7 @@ from collections.abc import Iterable
 import inspect
 import copy
 import logging
+from types import MappingProxyType
 from typing import Any, List, Optional, Union, TYPE_CHECKING
 
 from psyclone.configuration import Config
@@ -573,6 +574,9 @@ class SymbolTable():
                     tag is given.
 
         :raises InternalError: if the new_symbol argument is not a symbol.
+        :raises TypeError: if the tag is not a string.
+        :raises ValueError: if the tag is empty or `new_symbol` is already
+                            associated with another tag.
         :raises KeyError: if the symbol name is already in use.
         :raises KeyError: if a tag is supplied and it is already in use.
         :raises SymbolError: if the supplied symbol has an ImportInterface that
@@ -588,14 +592,11 @@ class SymbolTable():
             raise KeyError(f"Symbol table already contains a symbol with "
                            f"name '{new_symbol.name}'.")
 
-        if tag:
-            if tag in self.get_tags():
-                raise KeyError(
-                    f"This symbol table, or an outer scope ancestor symbol "
-                    f"table, already contains the tag '{tag}' for the symbol"
-                    f" '{self.lookup_with_tag(tag).name}', so it can not be "
-                    f"associated with symbol '{new_symbol.name}'.")
-            self._tags[tag] = new_symbol
+        # Validate the tag before changing either of the table's mappings.
+        # In particular, an error while validating an imported Symbol must not
+        # leave a dangling tag behind.
+        if tag is not None:
+            self._validate_new_tag(tag, new_symbol)
 
         if new_symbol.is_import:
             csym = new_symbol.interface.container_symbol
@@ -615,6 +616,85 @@ class SymbolTable():
                     f"the import interface refers to.")
 
         self._symbols[key] = new_symbol
+        if tag is not None:
+            self._tags[tag] = new_symbol
+
+    def _validate_new_tag(self, tag: str, symbol: Symbol) -> None:
+        '''Validate a proposed association between a tag and a Symbol.
+
+        A tag must be a non-empty string and both tags and Symbols have a
+        one-to-one relationship within a SymbolTable hierarchy.
+
+        :param tag: the proposed tag.
+        :param symbol: the Symbol to which the tag will refer.
+
+        :raises TypeError: if the tag is not a string.
+        :raises ValueError: if the tag is empty or the Symbol already has a
+                            different tag in this table.
+        :raises KeyError: if the tag is already associated with another
+                          Symbol in this or an ancestor table.
+        '''
+        if not isinstance(tag, str):
+            raise TypeError(
+                f"A SymbolTable tag must be a str but found "
+                f"'{type(tag).__name__}'.")
+        if not tag:
+            raise ValueError("A SymbolTable tag must not be an empty string.")
+
+        existing = self.get_tags()
+        if tag in existing:
+            if existing[tag] is symbol:
+                return
+            raise KeyError(
+                f"This symbol table, or an outer scope ancestor symbol "
+                f"table, already contains the tag '{tag}' for the symbol "
+                f"'{existing[tag].name}', so it can not be associated with "
+                f"symbol '{symbol.name}'.")
+
+        for old_tag, tagged_symbol in existing.items():
+            if tagged_symbol is symbol:
+                raise ValueError(
+                    f"Symbol '{symbol.name}' is already associated with tag "
+                    f"'{old_tag}' and cannot also be associated with tag "
+                    f"'{tag}'.")
+
+    def add_tag(self, symbol: Symbol, tag: str) -> None:
+        '''Associate an existing, local Symbol with a tag.
+
+        :param symbol: the Symbol to tag.
+        :param tag: the tag to associate with it.
+
+        :raises TypeError: if `symbol` is not a Symbol.
+        :raises KeyError: if `symbol` is not a local entry in this table.
+        '''
+        if not isinstance(symbol, Symbol):
+            raise TypeError(
+                f"The symbol to tag must be a Symbol but found "
+                f"'{type(symbol).__name__}'.")
+        local_symbol = self._symbols.get(self._normalize(symbol.name))
+        if local_symbol is not symbol:
+            raise KeyError(
+                f"Symbol '{symbol.name}' is not a local entry in this "
+                f"SymbolTable.")
+        self._validate_new_tag(tag, symbol)
+        self._tags[tag] = symbol
+
+    def remove_tag(self, tag: str) -> None:
+        '''Remove a tag defined locally in this SymbolTable.
+
+        :param tag: the tag to remove.
+
+        :raises TypeError: if `tag` is not a string.
+        :raises KeyError: if the tag is not defined locally.
+        '''
+        if not isinstance(tag, str):
+            raise TypeError(
+                f"A SymbolTable tag must be a str but found "
+                f"'{type(tag).__name__}'.")
+        if tag not in self._tags:
+            raise KeyError(
+                f"Tag '{tag}' is not defined locally in this SymbolTable.")
+        del self._tags[tag]
 
     def check_for_clashes(self, other_table, symbols_to_skip=()):
         '''
@@ -636,6 +716,7 @@ class SymbolTable():
             when importing symbols from `other_table` into this table.
 
         '''
+        # pylint: disable=too-many-branches,too-many-statements
         # pylint: disable-next=import-outside-toplevel
         from psyclone.psyir.nodes import IntrinsicCall
 
@@ -664,6 +745,12 @@ class SymbolTable():
             # the same Container.
             if (isinstance(this_sym, ContainerSymbol) and
                     isinstance(other_sym, ContainerSymbol)):
+                if (this_sym.is_intrinsic != other_sym.is_intrinsic or
+                        type(this_sym.interface) is not
+                        type(other_sym.interface)):
+                    raise SymbolError(
+                        f"ContainerSymbols named '{this_sym.name}' have "
+                        f"incompatible intrinsic or interface information.")
                 continue
 
             # If they are both IntrinsicSymbol then that's fine.
@@ -745,7 +832,8 @@ class SymbolTable():
                         f"that cannot be resolved by renaming "
                         f"one of the instances because:\n- {err1}\n- {err2}")
 
-    def _add_container_symbols_from_table(self, other_table):
+    def _add_container_symbols_from_table(self, other_table,
+                                          symbols_to_skip=()):
         '''
         Takes container symbols from the supplied symbol table and adds them to
         this table. All references to each container symbol are also updated.
@@ -758,6 +846,8 @@ class SymbolTable():
 
         '''
         for csym in other_table.containersymbols:
+            if csym in symbols_to_skip:
+                continue
             outer_sym = self.lookup(csym.name, otherwise=None)
             if not outer_sym:
                 # No clash so add the symbol.
@@ -788,6 +878,9 @@ class SymbolTable():
         # from this ContainerSymbol so that they now point to the one in
         # scope in this table instead.
         for isym in other_table.imported_symbols:
+
+            if isym in symbols_to_skip:
+                continue
 
             other_sym = self.lookup(isym.name, otherwise=None)
             if other_sym:
@@ -1003,6 +1096,7 @@ class SymbolTable():
         :raises SymbolError: if name clashes prevent the merge.
 
         '''
+        # pylint: disable=too-many-branches,too-many-locals
         if not isinstance(other_table, SymbolTable):
             raise TypeError(f"SymbolTable.merge() expects a SymbolTable "
                             f"instance but got '{type(other_table).__name__}'")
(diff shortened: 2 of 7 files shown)


sergisiso:
The 8 problems reported in the audit (#3598) are fixed:
1. Import resolution corrupts sibling/shadowing scopes: Fixed through scope-aware table replacement
2. ~~Ambiguous wildcard imports create hybrid symbols:  Fixed; conflicting provenance raises SymbolError before property copying~~ I decided this is not a problem and I added a test to showcase it
3. Formal-argument invariants can be broken: Fixed; membership, uniqueness, completeness, defensive copying, removal, rollback, and merge ordering are enforced
4. Copied aliased import loses orig_name and container properties: Fixed; full import and container metadata is preserved
5. Targeted resolution of renamed imports fails: Fixed; matching uses the external/original name
6. Failed add() leaves a dangling or invalid tag: Fixed through validate-before-commit
7. Iterable APIs consume generators and skip containers incorrectly: Fixed in resolve_imports(), merge(), and now direct check_for_clashes(); container dependencies are checked
8. Multiple tags contradict reverse-map and swap assumptions: Fixed through one-tag-per-symbol enforcement, read-only mappings, ancestor-aware merge, and tag-preserving swap

The older acknowledged limitations concerning general reference-safe removal (#898), recursive nested imports (#1540), copied-scope shadowing (#2666), and child wildcard resolution (#2331) remain outside these eight fixes.

sergisiso:
@arporter This heavily AI-assisted PR that comes from an audit of the symbol table that you suggested and I just had time to do a light review, but since I won't be able to continue next week and you have been looking at symbol tables recently I am already putting it for your review.

Feel free to let me know if some parts of these are not really necessary.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1011,6 +1106,56 @@ def merge(self, other_table, symbols_to_skip=()):
...
+            return
+
+        tags_to_merge = [(tag, sym)
+                         for tag, sym in other_table.tags_dict.items()
+                         if sym not in symbols_to_skip]
+        visible_tags = self.get_tags()
+        for tag, tagged_symbol in tags_to_merge:
+            if tag in visible_tags and visible_tags[tag] is not tagged_symbol:
+                raise SymbolError(
+                    f"Cannot merge tag '{tag}' because it is associated with "
```
sergisiso, line 1123: I am not sure if this is the correct solution when the merge is done during the backend. At that point we don't care about tag clashes. But I still left it to discuss: what is the correct merge semantics for tags during normal operations. Should it converge the symbols into a single one? Let them be 2 symbols and lose the tag in one of them? Fail but add an argument that says ignore tag clashes for then we don't care like in the backend?

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -6271,15 +6271,18 @@ def _module_handler(self, node, parent):
         try:
             subprog_part = _first_type_match(
                 node.children, Fortran2003.Module_Subprogram_Part)
+        except ValueError:
+            pass
```
arporter: Comments please, "No module subprogram found"

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -6271,15 +6271,18 @@ def _module_handler(self, node, parent):
...
+        else:
             module_subprograms = \
```
arporter: "Create a list holding all subprogram parse trees."

src/psyclone/psyir/frontend/fparser2.py
```diff
@@ -6271,15 +6271,18 @@ def _module_handler(self, node, parent):
...
             if module_subprograms:
-                self.process_nodes(parent=container, nodes=module_subprograms)
-        except SymbolError as err:
-            raise NotImplementedError(str(err.value))
-        except ValueError:
-            pass
+                try:
+                    self.process_nodes(parent=container,
+                                       nodes=module_subprograms)
+                except SymbolError as err:
```
arporter: Would it be better to call `process_nodes` separately for each entry in `module_subprograms` or is a SymbolError here fatal to any ambition to have PSyIR for the program unit as a whole?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -615,6 +612,85 @@ def add(self, new_symbol: Symbol, tag: Optional[str] = None):
...
+                          Symbol in this or an ancestor table.
+        '''
+        if not isinstance(tag, str):
+            raise TypeError(
+                f"A SymbolTable tag must be a str but found "
+                f"'{type(tag).__name__}'.")
+        if not tag:
+            raise ValueError("A SymbolTable tag must not be an empty string.")
+
+        existing = self.get_tags()
```
arporter: Presumably tags are case sensitive (I think that makes sense)?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -615,6 +612,85 @@ def add(self, new_symbol: Symbol, tag: Optional[str] = None):
...
+        if tag in existing:
+            if existing[tag] is symbol:
+                return
+            raise KeyError(
+                f"This symbol table, or an outer scope ancestor symbol "
+                f"table, already contains the tag '{tag}' for the symbol "
+                f"'{existing[tag].name}', so it can not be associated with "
```
arporter: "cannot"

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -644,6 +721,11 @@ def check_for_clashes(self, other_table, symbols_to_skip=()):
                 f"check_for_clashes: 'symbols_to_skip' must be an instance of "
                 f"Iterable but got '{type(symbols_to_skip).__name__}'")
 
+        # This argument is queried for every clashing Symbol below. Materialise
+        # it once so generators and other one-shot Iterables behave like
```
arporter: I need this translating into plain English please :-)

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -664,6 +746,12 @@ def check_for_clashes(self, other_table, symbols_to_skip=()):
             # the same Container.
             if (isinstance(this_sym, ContainerSymbol) and
                     isinstance(other_sym, ContainerSymbol)):
+                if (this_sym.is_intrinsic != other_sym.is_intrinsic or
+                        type(this_sym.interface) is not
+                        type(other_sym.interface)):
+                    raise SymbolError(
+                        f"ContainerSymbols named '{this_sym.name}' have "
```
arporter: Does printing the ContainerSymbol (e.g. f"{this_sym}") give useful informatino here?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1011,6 +1106,56 @@ def merge(self, other_table, symbols_to_skip=()):
                 f"SymbolTable.merge() expects 'symbols_to_skip' to be an "
                 f"Iterable but got '{type(symbols_to_skip).__name__}'")
 
+        # This argument is documented as any Iterable and is used repeatedly
```
arporter: I'm still confused by this - what affect does documenting the type have on the code (I thought none)?

src/psyclone/psyir/symbols/symbol_table.py
arporter: My feeling is that if they are both tagged but have different tags then we raise an error. If only one of them is tagged then we can converge the symbols into a single one and retain the tag. Do you have examples of where this causes failures in the backend?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1011,6 +1106,56 @@ def merge(self, other_table, symbols_to_skip=()):
...
+        visible_tags = self.get_tags()
+        for tag, tagged_symbol in tags_to_merge:
+            if tag in visible_tags and visible_tags[tag] is not tagged_symbol:
+                raise SymbolError(
+                    f"Cannot merge tag '{tag}' because it is associated with "
+                    f"Symbol '{visible_tags[tag].name}' in this scope and "
+                    f"Symbol '{tagged_symbol.name}' in the supplied table.")
+            local_match = self.lookup(tagged_symbol.name, otherwise=None)
+            combines_with_local = local_match and (
+                (isinstance(local_match, ContainerSymbol) and
```
arporter: I suggest putting this logic inside its own method.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1020,13 +1165,34 @@ def merge(self, other_table, symbols_to_skip=()):
...
         # Copy each Symbol from the supplied table into this one, excluding
         # ContainerSymbols and any listed in `symbols_to_skip`.
         self._add_symbols_from_table(other_table,
                                      symbols_to_skip=symbols_to_skip)
 
+        # Preserve formal-argument ordering for any arguments included in the
+        # merge. Without this, adding an ArgumentInterface produces an
+        # internally inconsistent destination table.
+        merged_arguments = self._argument_list[:]
+        for argument in other_table.argument_list:
```
arporter: I wasn't sure about this at first but, having written a long response, I've come around to it. Please could you update the method docstring to emphsise that this is what happens.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1020,13 +1165,34 @@ def merge(self, other_table, symbols_to_skip=()):
...
+            if (argument not in symbols_to_skip and
+                    argument not in merged_arguments):
+                merged_arguments.append(argument)
+        self.specify_argument_list(merged_arguments)
+
+        # Tags are part of the identity-preserving metadata of a Symbol and
+        # must follow it into the destination table. A clashing Symbol may have
+        # been renamed or combined with an existing equivalent Symbol above.
```
arporter: but its object identity remains unchanged?

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1082,7 +1248,29 @@ def specify_argument_list(self, argument_symbols):
             the existing entries in the SymbolTable.
 
         '''
-        self._validate_arg_list(argument_symbols)
+        # Historically this method also accepted new Symbols. Preserve that
```
arporter: Please modernise the docstring and use typehints.

src/psyclone/psyir/symbols/symbol_table.py
```diff
@@ -1082,7 +1248,29 @@ def specify_argument_list(self, argument_symbols):
...
+                if local_symbol is None:
+                    self.add(symbol)
+                    added_symbols.append(symbol)
+                elif local_symbol is not symbol:
+                    raise ValueError(
+                        f"DataSymbol '{symbol.name}' is listed as a kernel "
+                        f"argument but a different Symbol with that name is "
+                        f"the local entry in this SymbolTable.")
+            # Validate the other half of the invariant before committing the
+            # new list: every local argument Symbol must be present in it.
```
arporter: s/it/the table/

src/psyclone/psyir/symbols/symbol_table.py
arporter: By "accepted new Symbols" presumably it means that we just didn't check? I can't see anything in the old implementation that checked whether or not they were new? Also, a comment about what this method used to do isn't that useful - it just needs to say wahat this does.

PR #3603, opened 2026-09-15
(Closes #1770) make eg20 compile

### LINKED ISSUE #1770: Make lfric eg20 compile

arporter:
Adds a couple of extra files to our test copy of LFRic infrastructure and extends eg20 so that it compiles and links the example.


arporter:
A small PR that just extends the LFRic test infrastructure and then uses it to get eg20 to compile. One for @hiker :-)

hiker (approved):
Simple change. The code compiles and cleans as expected. I've triggered the IT, and will merge when this is green (CT linkcheck failures are caused by known issues with zeonodo atm).

PR #3607, opened 2026-09-16
Removes references to Dynamo transformationsg

LonelyCat124:


src/psyclone/tests/psyGen_test.py
```diff
@@ -500,8 +500,6 @@ def test_find_subclasses():
     # Check that the method does not include the legacy names for the
```
arporter: This comment needs removing.

arporter (changes requested):
Thanks Aidan. One comment needs removing and, in doc/user_guide/lfric.rst we still talk about "early transformations" having Dynamo in their name:
```
doc/user_guide/lfric.rst:are given below. Early transformations include "Dynamo0p3" or "Dynamo"
```
To be completely on the safe side, I'll also run the ITs.

src/psyclone/tests/psyGen_test.py
LonelyCat124: Removed.

LonelyCat124:
@arporter Fixed the comments and docs, should be ready for another look. ITs look fine (NEMO5 still going but shouldn't be affected).

arporter (approved):
All good now, thanks Aidan (and the ITs were all green).

Issue #3454, opened 2026-06-11
Module-inline with callee scope symbols


sergisiso:
Currently Module-inline fails the validation for subroutines that contain references to other callee scope symbols.

There are 2 solutions to this:
- Add imports in the caller scope to these symbols (but this only works if they are public)
- Localise all the symbols inside the module-inlined routine. E.g. copying if it is a parameter, or in-lining it if it is a call to a subroutine. 

This is the GPU bottleneck in the timesteping region after annotating the LFRic infrastructure:
```
! Transformation Error: Cannot apply KernelModuleInlineTrans to Kernel 'ffsl_flux_xy_panel_remap_code' because it accesses data 
!& from its outer scope: PSyclone SymbolTable error: Kernel 'ffsl_flux_xy_panel_remap_code' contains accesses to 
!& 'ffsl_flux_xy_panel_remap_1d' which is declared in the callee module scope.
```

arporter:
Hit #3608 while adding tests for this. If I add `--f90flags='-Werror=implicit-procedure'` then I get a couple of extra (new) test failures in kernel_module_inline_trans_test.py.

arporter:
If a Kernel calls a routine that itself calls another routine then:
 * Check that that routine can also be module inlined - this needs to happen in `validate`;
 * Add it to the list of routines to module inline;
 * Finally, module inline the full list of routines.

PR #3610, opened 2026-09-18
(Towards #3454) extend module inline to recursively copy in the targets of calls to local routines.

arporter:


arporter:
The GungHo with ACC offload IT failed because:

```
Error trying to create PSyIR for 'psyclone-tests/latest-run/lfric_apps/working-gh-accoffload/working/build_gungho_model/utilities/constants_mod.F90': 'PSyclone SymbolTable error: Symbol 'r_def' already present in SymbolTable with a defined interface (Static).'
```
and `constants_mod.F90` has preprocessor directives in it.

arporter:
This gives:
```
Module-inline successful: 293
Module-inline failed: 6
   -> because it accesses data from its outer scope: 3

 --- Then we need to inline them, or fallback to GPU routine annotations ---
Annotation successful: 128
   Deferred-Inline successful: 73
   Deferred-Inline failed: 56
   -> Inlining polymorphic kernels is not supported: 0
   -> contains one or more Return statements: 7
   -> is passed by argument and is assigned to before the call: 14
Annotation failed: 171
   -> accesses the imported symbol: 150
   -> calls another routine: 4
   -> calls intrinsic: 21
   -> only supports the transformation of a MATMUL operation when: 1

 --- Then offload each loop with kernels inside ---
Offload independent loop: 83
Offload with dof loop: 35
Offload with atomics: 0
Offload with cell colouring: 40
Offload with cell tile-colouring: 0
Failed to offload: 3
Added inner loop nested parallelism: 0
Added OMP threading: 171
```
This is an increase of 10 kernels module inlined. Unfortunately this hasn't had any effect on the number of routines that are deferred-inlined. I also need to look to see why 3 of them fail but aren't picked up under 'accesses data from its outer scope'. Might be because I've changed the message that we emit. Similarly,  "Inlining polymorphic kernels is not supported: 0" was previously 34 so I've probably changed something there too.

arporter:
```
Transformation Error: Cannot apply KernelModuleInlineTrans to Kernel 'polyh_w3_koren_code' because it contains static data symbol(s): 'point'
```
where that kernel routine contains:
```fortran
integer(kind=i_def), dimension(3,4) :: point=reshape((/4,1,2,5,1,3,2,1,4,3,1,5/),shape(point))
```
Presumably it can't be a `parameter`? We could use the DA to check that it is never written to. I've just tested and gfortran is happy for it to be made a parameter (in a test program):
```bash
code_fragments$ gfortran array_parameter.f90
code_fragments$ cat array_parameter.f90 
program andy
  implicit none
  integer, dimension(3,4), parameter :: point=reshape((/4,1,2,5,1,3,2,1,4,3,1,5/),shape(point))
end program andy
```

arporter:
```
Transformation Error: Cannot apply KernelModuleInlineTrans to Kernel 'rtheta_bd_code' because it contains static data symbol(s): 'upwind'
```
this has:
```fortran
    logical(kind=l_def) :: upwind = .false.
```
and `upwind` is only read therefore it could definitely be a `parameter`.

arporter:
> Similarly, "Inlining polymorphic kernels is not supported: 0" was previously 34 so I've probably changed something there too.

This is a puzzle. I can't find any code on master that would emit this message. Can @sergisiso remember what it is supposed to be checking for?
```bash
psyclone$ grep -inr 'polymorphic' --include \*.py --exclude-dir tests .
./psyGen.py:1197:        By default, a Kern is not polymorphic and therefore has no interface
./transformations.py:1865:        :raises TransformationError: if the supplied node is a polymorphic
./domain/common/transformations/kernel_module_inline_trans.py:115:        :raises TransformationError: if the call is to a polymorphic routine
./domain/common/transformations/kernel_module_inline_trans.py:163:                    f"it is a polymorphic routine (i.e. an interface) and the "
./domain/gocean/transformations/gocean_opencl_trans.py:759:        # validate() has checked that the kernel is not polymorphic.
./domain/lfric/lfric_kern.py:789:        :returns: the interface symbol for this kernel if it is polymorphic,
./psyir/nodes/scoping_node.py:35:    # Polymorphic parameter to initialize the Symbol Table of the ScopingNode
./gocean1p0.py:1084:        For consistency with LFRic kernels (which may be polymorphic), this
./gocean1p0.py:2123:    # Polymorphic parameter to initialize the Symbol Table of the Schedule
```

arporter:
I need to check test coverage. Also need to look at the functionality I added to routine.py and check whether it needs to check for unresolved symbols.

sergisiso:
@arporter That is a leftover! I previously had some pre-validations in KernelInlineTrans that I decided to remove in the last minute because InlineTrans already validates the same conditions, and I forgot to update the counting script. I suspect this equivalent to the InlineTrans:
```
            callee_count = len(node.get_callees())
            if callee_count != 1:
                raise TransformationError(
                    f"Cannot inline routine '{name}' because its call has "
                    f"{callee_count} possible callees. The "
                    "'allow_no_args_check_if_only_one_callee' option requires "
                    "exactly one callee.")
````

Issue #3601, opened 2026-09-14
LFRic on GPU next steps


sergisiso:
After https://github.com/stfc/PSyclone/pull/3587 the LFRic on GPU produces this stats
```
 --- First we need to be able to modify kernels ---
Module-inline successful: 283
Module-inline failed: 16
   -> because it accesses data from its outer scope: 16

 --- Then we need to inline them, or fallback to GPU routine annotations ---
Annotation successful: 128
   Deferred-Inline successful: 73
   Deferred-Inline failed: 56
   -> Inlining polymorphic kernels is not supported: 34
   -> contains one or more Return statements: 7
   -> is passed by argument and is assigned to before the call: 14
Annotation failed: 171
   -> accesses the imported symbol: 148
   -> calls another routine: 7
   -> calls intrinsic: 21
   -> only supports the transformation of a MATMUL operation when: 1

 --- Then offload each loop with kernels inside ---
Offload independent loop: 83
Offload with dof loop: 35
Offload with atomics: 0
Offload with cell colouring: 40
Offload with cell tile-colouring: 0
Failed to offload: 3
Added inner loop nested parallelism: 0
Added OMP threading: 171
```

Some steps we can do are:

Annotate Infrastructure:
- [ ] Can we add annotations to declarations and to module declarations.
- [ ] https://github.com/stfc/PSyclone/issues/3449 A pass that append "subroutines" that need GPU annotations into a file, and a pass that applies them.
- Alternatives: check if imported calls already have annotations, module-inline calls inside other module-inline.

Inlining polymorphic kernels:
- [ ] Fails at argument datatype comparisons (i_def, pointers)

Module-inline:
- [ ] #3454 Access-data from outer scope, mostly other subroutines. We could copy them. What we do if they access a module variable.

Full kernel inline:
- [ ] Automatic arrays that depend on and argument. Can we hoist them? Otherwise need allocation 

Topics left to discuss
- What is the necessary domain size/node to take advantage of GPUs. Create an example with this size for accurate profiling.
- Atomics.
- Fusing.

sergisiso:
After inlining kernels that have automatic arrays converting them to allocatables, there are 2 kernels that fail at runtime with:
```
Accelerator Fatal Error: call to cuStreamSynchronize returned error 700 (CUDA_ERROR_ILLEGAL_ADDRESS): Illegal address during kernel execution
```
These are:
```
    !$omp target
    !$omp teams distribute parallel do default(shared) private(cell)
[...]

PR #3611, opened 2026-09-21
(towards #3601) Inline automatic arrays by converting them to allocatables

sergisiso:


sergisiso:
@arporter This is ready for a first look

